PR #22, opened 2018-09-27
List all images concurrently

Zeniuus:
There are two points that I'm not sure about my code.

1. It seems that aiojobs module is not needed here currently...
2. Should I do more Exception handling?


achimnol:
And yes, I agree with that `aiojobs` is not strongly required here. But it is nice to have a knob to limit cuncurrenty when required as long as using `aiojobs` does not incur significant code maintainence overheads. Feel free either to remove it or leave it.

PR #24, opened 2018-10-25
Update docs: tag for session

Zeniuus:
Pull request for [manager#102](https://github.com/lablup/backend.ai-manager/issues/102).

### DIFF
--- a/docs/user-api/kernels.rst
+++ b/docs/user-api/kernels.rst
@@ -30,6 +30,12 @@ Parameters
      - The kernel runtime type, usually in the form of the language name and its version tag connected with a colon.
        (e.g., ``"python:latest"``)
 
+   * - ``tag``
+     - ``str``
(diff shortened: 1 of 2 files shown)


docs/user-api/kernels.rst
```diff
@@ -30,6 +30,10 @@ Parameters
      - The kernel runtime type, usually in the form of the language name and its version tag connected with a colon.
        (e.g., ``"python:latest"``)
 
+   * - ``tag``
+     - ``str``
+     - An optional per-session, user-provided tag to distinguish which sessions are from which users.
```
achimnol:
```suggestion
     - An optional per-session, user-provided tag for administrators to keep track of additional information of each session, such as which sessions are from which users.
```

achimnol: I think changing explanation like this would be more comprehensive.

Issue #128, opened 2020-02-03
API v5 GQL Revision


achimnol:
In API v5 (along with Backend.AI v19.12 series), we renamed most `kernel` wording to `session`.
This has introduced backward-incompatible changes in clients.

To prepare with Backend.AI v20.03's multi-container session support (#78), the GraphQL query schema for `compute_session` objects also must be changed in a backward-incompatible way.

To minimize the breaking changes, we decided to bring the necessary GQL changes in Backend.AI v19.12, with an emulated query resolver.

achimnol:
Now the remaining work is documentation update.

achimnol:
Docs update will be handled separately.

PR #132, opened 2020-02-04
Update docs for v20.03 release

achimnol:
refs #128


achimnol:
Internal ticket: OP#100

Issue #211, opened 2020-08-26
When installing install-dev.sh in macOS environment, if pyenv is not installed, the installation will not be completed.


zel0rd:
**Describe the bug**
- When running scripts/install-dev.sh on macOS, if pyenv and virtualenv are not installed, the installation cannot be completed.

**To Reproduce**
1. [Setting same environment] brew uninstall pyenv / brew uninstall pyenv-virtualenv
2. [execute scripts/install-dev.sh] ./install-dev.sh
3. [See error1] WARNING: Can not proceed with installation. Kindely remove the '/User/./.pyenv' directory first.
4. [Fix error1] brew install pyenv
5. [See error2] [INFO] Creating virtualenv on pyenv... / pyenv: no such command 'virtualenv'
6. [Fix error2] brew install pyenv-virtualenv

**Expected behavior**
If pyenv or pyenv-virtual is not installed, automatically install it using brew.

**Screenshots**
<img width="758" alt="issue_when_not_installed_pyenv" src="https://user-images.githubusercontent.com/28799597/91275467-2c847400-e7bb-11ea-82d3-36fddef78eee.png">
<img width="468" alt="issue_when_not_installed_virtualenv" src="https://user-images.githubusercontent.com/28799597/91275497-36a67280-e7bb-11ea-80f3-f0e2d11ae1c1.png">

**Additional context**
- [ ] I will be revising pyenv and pyenv-virtual within this week so that they can be automatically installed.

zel0rd:
https://github.com/lablup/backend.ai/pull/212

achimnol:
The root problem is here: between your reproduction step 4 and 5, **there is a missing step**.
You must (re-)initialize pyenv for your current shell, by executing `exec "$(pyenv init -)"` or restarting the shell after setting up your shell configuration such as `.bashrc`, `.bash_profile`, `.profile`, etc. [as guided by the pyenv manual](https://github.com/pyenv/pyenv#basic-github-checkout).

achimnol:
The correct fix would be to improve the error messages, so that users can properly "reinstall both pyenv and pyenv-virtualenv" and/or "fix the shell config", not to silenty brew-uninstall as in #212.
Please rewrite #212.

achimnol:
Closing due to long inactivity.

PR #212, opened 2020-08-26
Fixed pyenv install error

zel0rd:
Fixed an error that occurred when installing pyenv because the .pyenv path already exists on mac

### DIFF
--- a/scripts/install-dev.sh
+++ b/scripts/install-dev.sh
@@ -98,6 +98,11 @@ show_error() {
   echo "${RED}[ERROR]${NC} ${LRED}$1${NC}"
 }
 
+show_check() {
+  echo " "
+  echo -n "${CYAN}[CHECK]${NC} $1"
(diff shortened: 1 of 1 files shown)


zel0rd:
resolved: #211

achimnol:
Thanks for the PR, but I think we need to *explicitly* ask the user whether to perform "brew uninstall" because it's a destructive operation.

Currently it says just removal of pyenv and silently invoke "brew uninstall", but it could jeopardize the user setup!
(e.g., I have a few tens of virtualenvs created with pyenv for various projects with pyenv, but if a program silently/implicitly removes them then I will blame it forever)
At least, the message should include that it will remove brew-installed pyenv and all its environments with a strong warning, or it should provide choices for *how* to remove the existing pyenv setup.

achimnol:
Closing due to long inactivity.

PR #215, opened 2020-09-02
Update install-dev.sh to install and run storage-proxy

yongkyunlee:
- Install storage-proxy and get it to run.


inureyes (approved):
I tested it and it works as expected.

Issue #224, opened 2020-09-14
When installing install-dev.sh in ubuntu environment, if postgreSQL client is not installed, the installation is break.


KHJcode:
**Describe the bug**
When installing install-dev.sh in ubuntu environment, if postgreSQL client is not installed, the installation is break.

**To Reproduce**
1. Run the install-dev.sh file.

**Expected behavior**
The install-dev.sh file use psql inside of the pgsql container.

**Screenshots**
![image](https://user-images.githubusercontent.com/60125944/93086883-c19bce00-f6d2-11ea-8d45-992c736d8b22.png)

**Client:**
 - OS: Ubuntu 20.04 LTS
 - App version: -
 - Browser version: Chrome
 - Python version: 3.8.3
 - NodeJS version: 12.13.1

PR #226, opened 2020-09-15
Support install without Postgres installed

yongkyunlee:
Update psql command to be run with docker_exec.

Solving issue lablup/backend.ai#224
Tested with clean Ubuntu 16.04 instance (only pyenv, pyenv-virtualenv, docker, docker-compose installed)


inureyes (approved):
LGTM.

PR #261, opened 2021-04-22
fix: change the python execution command in the installation script

### LINKED ISSUE #260: Change the python execution command in the installation script
The execution command in manager needs to be changed because the location changed from gateway to manager

Sujin-Kim1:
modify how manager runs in execution script
this PR closes #260


achimnol (approved):
Good catch!

PR #294, opened 2021-10-07
fix: docs build warning

Sujin-Kim1:
Fix
- Title level inconsistent.
- The document is not included in any toctree.


achimnol:
Those excluded documents were falling behind, because they were written for very very old versions of Backend.AI.
For instance, we don't use supervisord for deployment at all.
To include them, we first need to update them.

achimnol (changes requested):
First need to update the contents of excluded documents before including them, or remove them and replace with more fresh explanation.

Sujin-Kim1:
> First need to update the contents of excluded documents before including them, or remove them and replace with more fresh explanation.

Okay. I just deleted them now, and I'll write the update PR again later when the schedule is okay.

Issue #290, opened 2021-09-19
Let's make a new markdown file to save a install information


mindgitrwx:
**Is your feature request related to a problem? Please describe.**
At the very beginning of the Backend.ai installation, developers use this script.
- [install-dev.sh] https://github.com/lablup/backend.ai/blob/main/scripts/install-dev.sh

This prints out critical notes for development like ID, password, keypair, environment ID at the end of the installation. 
But we always copy this information manually in the terminal.

**Describe the solution you'd like**
Automating extract this information to another file - like 'INSTALL_INFO.md'

**Describe alternatives you've considered**
I wonder markdown file format is the best option. If there is a better way, please let me know.

**Additional context**
![Selection_254](https://user-images.githubusercontent.com/37054916/133930330-39b0c365-fcb8-47ad-aece-b6a26f17f82e.png)

lizable:
@mindgitrwx How about providing other file extensions(.txt) for convenience sake?

PR #304, opened 2021-10-20
feat: Generate INSTALL-INFO after backend.ai install using script (#290)

mindgitrwx:
**Issue:** https://github.com/lablup/backend.ai/issues/290

This PR adds Install_info generate logic in install-dev.sh

When users run ./install-dev.sh, It generates a basic documentation markdown file that has critical information to use backend.ai

image

(I tested this on ubuntu 18,04. I didn't test on other platforms.)

### DIFF
--- a/scripts/install-dev.sh
+++ b/scripts/install-dev.sh
@@ -130,6 +130,10 @@ has_python() {
   fi
 }
 
+info_to_md(){
+    sed 's/\x1b\[[0-9;]*m//g' | tee -a "INSTALL-INFO.md"
+}
+
 if [[ "$OSTYPE" == "linux-gnu" ]]; then
   if [ $(id -u) = "0" ]; then
     docker_sudo=''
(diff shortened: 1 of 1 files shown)


achimnol (commented):
Q. Does it preserve the original terminal output (with colors)?
Q. Does the generated result actually conform with the Markdown syntax?

mindgitrwx:
What I tested so far is here.

Answer 

Q1: I tested this code on ubuntu 18.04 bash, zsh, Macbook terminal, eshell etc. And it preserve the original terminal output. If the colors don't come out as expected, the problem is not with the code, but with the nature of the terminal.

Q2: I think this code is not perfectly conform with the markdown syntax. This picture shows how output comes out. I just simply redirect the code the markdown to reduce complexity of the code.

스크린샷 2021-11-12 오후 5 38 19

achimnol:
I think this should be reworked to provide either Markdown-compliant output or just a plain-text dump.

achimnol (changes requested):
I think this should be reworked to provide either Markdown-compliant output or just a plain text dump.

achimnol:
Closing this as lack of interest and now we rely on Pants for both first-party and 3rd-party dependency mgmt.

PR #305, opened 2021-11-01
Fix/typo in install dev.sh

yujung7768903:
During the installation process, the following message appeared because POSTGRES_CONTAINER_ID is ''.
`Error: No such container: psql`
Screen Shot 2021-11-01 at 5 46 19 PM

In Docker container list, Name is "backendai-half-db-1" but there's a typo with "backendai-half-db_1" in install-dev.sh.
Screen Shot 2021-11-01 at 5 50 52 PM

Screen Shot 2021-11-01 at 5 56 32 PM

### DIFF
--- a/docker-compose.halfstack-2109.yml
+++ b/docker-compose.halfstack-2109.yml
@@ -35,7 +35,7 @@ services:
       retries: 10
 
   backendai-half-etcd:
-    image: quay.io/coreos/etcd:v3.4.15
+    image: quay.io/coreos/etcd:v3.5.0
     restart: unless-stopped
     volumes:
       - "./tmp/backend.ai-halfstack/${DATADIR_PREFIX:-.}/etcd-data:/etcd-data:rw"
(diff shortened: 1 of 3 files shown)


achimnol:
Hmmm. The container naming in docker compose v1/v2 (and probably Docker Desktop for Mac / Docker Engine on Linux) differs. Could we generalize?

yujung7768903:
I think we need to correct the typo in the 'install-dev.sh' file without changing the docker container name. The docker's name is `dev2109-backendai-half-db-1`, but `${ENV_ID}_backendai-half-db_1` in install.dev.sh file. 
So when I checked with "POSTGRES_CONTAINER_ID", it came out as a blank. Therefore, there seems to be a problem during the installation process.

achimnol:
> I think we need to correct the typo in the 'install-dev.sh' file without changing the docker container name. The docker's name is `dev2109-backendai-half-db-1`, but `${ENV_ID}_backendai-half-db_1` in install.dev.sh file. So when I checked with `echo $POSTGRES_CONTAINER_ID`, it came out as a blank. Therefore, there seems to be a problem during the installation process.

I mean, that may not be a typo in specific environments. Could we generalize the pattern?

yujung7768903:
I understand. Then, why don't we change "${ENV_ID}-backendai-half-db-1" only if "POSTGRES_CONTAINER_ID" is blank?

inureyes (approved):
LGTM

PR #313, opened 2021-11-28
setup: change the value of num-proc to 1 of manager.toml

adrysn:
When a user tries to install in a low-end PC or a VM which have cores less than 4, `num-proc = 4` causes an installation error.


inureyes (approved):

inureyes:
LGTM

Issue #331, opened 2021-12-28
update-to-latest-repo.sh does not work.


mindgitrwx:
**Describe the bug**
update-to-latest-repo.sh does not work.

**To Reproduce**
1. Go to 'backend.ai-dev/backend.ai/scripts'
2. execute script ./update-to-latest-repo.sh
3. Some errors have been generated.
```
./update-to-latest-repo.sh: line 3: cd: ./backend.ai-dev: No such file or directory
ERROR: Could not open requirements file: [Errno 2] No such file or directory: './requirements-dev.txt'

```
**Expected behavior**
Update backend.ai project properly

**Screenshots**

![image](https://user-images.githubusercontent.com/37054916/147607297-52da0357-5a44-4ec9-82a0-8112d9bd0090.png)

**Enviroment**
 - Tested on macOS Monterey 12.0.1, ubuntu 18.04, zsh, eshell, bash
 
**Additional context**
추가적으로 문제 상황을 설명할 수 있는 다른 맥락을 기술해주세요.

achimnol:
This script is no longer valid as we migrated to the mono-repo. (#417)
`update-to-latest-repo.sh` is deprecated now.

PR #332, opened 2021-12-28
fix: add HOME variable and modify requirements path #331

mindgitrwx:
This script was tested on several environment Intel mac and ubuntu-desktop shell, vscode shell, emacs shell.


inureyes:
This is fixed via 9fcf76f1ec3ffa8ccb08e5c62abb75e13fbc6492 

Thank you :D

PR #353, opened 2022-01-22
docs: translate index.po

studioego:
translate index.po into Korean

This repository is a meta repository where we only keep compatible version
dependencies to individual components of Backend.AI.

Please head to appropriate Backend.AI sub-projects (e.g., manager, agent, kernels) to
contribute your code unless you really have contribution here.


achimnol (approved):
As the documentation structure is slightly changing, this requires some of rework. But I will first merge the PR to avoid conflicts.

PR #388, opened 2022-03-15
fix: install non binary grpcio and set up db schema early

fregataa:
update install script due to change of image management


achimnol:
Thanks! Just closing in favor of #391.

PR #398, opened 2022-03-21
fix: install grpcio on Apple M1 & python version notation

studioego:
1. fix to install grpcio on Apple M1 platform
read the link https://github.com/grpc/grpc/issues/25082
Some environment variables need to be set for build on apple M1 platform.
```
export GRPC_PYTHON_BUILD_SYSTEM_OPENSSL=1
export GRPC_PYTHON_BUILD_SYSTEM_ZLIB=1
```

However, I found that the script doesn't set some environment variables on Apple M1.
So, I fix shell script code.

2. fix the python version notation on shell script
`3.9.10` to `3.10.2`

This repository is a meta repository where we only keep compatible version
dependencies to individual components of Backend.AI.

Please head to appropriate Backend.AI sub-projects (e.g., manager, agent, kernels) to
contribute your code unless you really have contribution here.

### DIFF
--- a/scripts/install-dev.sh
+++ b/scripts/install-dev.sh
@@ -51,7 +51,7 @@ usage() {
   echo ""
   echo "  ${LWHITE}--python-version VERSION${NC}"
   echo "                       Set the Python version to install via pyenv"
-  echo "                       (default: 3.9.10)"
+  echo "                       (default: 3.10.2)"
   echo ""
   echo "  ${LWHITE}--install-path PATH${NC}  Set the target directory"
   echo "                       (default: ./backend.ai-dev)"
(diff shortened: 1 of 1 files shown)


achimnol:
In my case, those flags caused build failure. T_T

achimnol:
It turns out that having one of the following conditions may incur this error with explicit `GRPC_PYTHON_BUILD_*` flags:
* Xcode beta version is installed.
* `CFLAGS`, `LDFLAGS` are configured to refer homebrew-provided OpenSSL.

If we remove Xcode beta and those flags, then it installs without `GRPC_PYTHON_BUILD_*` flags on macOS 12.3 and Xcode 13.3 with M1 Macs.

achimnol:
So I'm closing this PR as the original report has been affected by local setup.

luisvicentesanchez:
I don't have xcode install and although I have the homebrew-provided openssl, I don't have set the CFLAGS and LDFLAGS in my environment and I still get this error.

PR #423, opened 2022-05-18
fix: Update default target branch to 21.09

rapsealk:
Update default `--server-branch` and `--client-branch` on `install-dev.sh` to keep compatibility.


achimnol:
These arguments will be deprecated after #417 and it is intended to keep them "main" as the default development setup should be based on "main".

PR #430, opened 2022-06-01
Add Session DB table and migrate scheduling code to use session table

### LINKED ISSUE #325: Split session and kernel database tables
Currently we have one single big `kernels` table used for queueing and tracking multiple containers of cluster sessions, with complicated unique constraint and indexes.

Let's refactor this as follows:
* The **`sessions`** table keeps track of the session-level information and represents **the desired state** and **what to do**.
  - Along with the pipeline's "manager-task" type modules, there may be session objects with *no* kernel objects.
  - Now we could distinguish different session types: "interactive", "batch", "service", "system", "manager-task", etc.
* The **`kernels`** table will remain as almost same, but with less duplicate columns, and represents **the current state** of provisioned resources.
  - Now we could distinguish different kernel types: "container", "process", "k8s-pod", etc.

This will simplify queries against sessions by removing `kernels.c.role == DEFAULT_ROLE` conditions.

We need to define an aggregate function F: `kernel[].status` -> `session.status` to apply on `sessions.status` when updating `kernels.status` via the event bus.

This work will require major rewriting of the scheduler and dispatcher, but now we could take advantage of SQLAlchemy's ORM as it now supports asyncio.
The GraphQL API will remain mostly same, because it already reflects the separation of compute sessions and compute containers.

fregataa:
resolves #325 

## 1. Add Session DB table.
- This table is used to schedule and manage compute sessions.
- It seems to have many columns that overlaps with kernel DB table. We need to normalize those tables according to our spec.

## 2. Migrate to SQLAlchemy ORM partially
- Migrate manager's scheduler codes from SQLAlchemy Core to ORM. We still are able to use Core tables and SQLAlchemy 1.x query styles.

## 3. Change compute session scheduler to use session table and APIs.
- We have managed compute session with `kernel` table which cause blocking issues -e.g) cannot implement session tag table- and make us confused. From now, data of scheduling jobs such as enqueuing, preparing and destroy compute session will be inserted to session table. And we can handle those scheduling job data by APIs implemented in `sessions.py`.

## 4. Optimize DB query
- Thanks to the new session table, we are able to optimize some DB query statements especially which are related to Manager's scheduler. 
- But there are a few obstacles since SQLAlchemy does not fully support Asynchronous tasks -e.g) Unable lazy loading- and usage of SQLAlchemy's `orm.Session` is not concrete -e.g) boundary of `orm.Session` is arbitrary. We can optimize such tasks much more.

## Plus
- We need to refactor this scheduler codes since this migrated code followed some legacy codes. e.g) it assumes that there is no heterogeneous kernels in one session such as no different images or etc, Updating status of compute sessions and kernels process is not clear yet. Need to specify the steps and interactions between session and kernel through updating status.
- some of manager API still use SQLAlchemy Core based queries. We can update the codes.

## TODO
1. Normalize Session DB table and Kernel DB table. Mainly drop duplicated and legacy columns.
2. Optimize the usage of SQLAlchemy Session.


fregataa:
closed due to a number of conflicts and commits with bad readability

PR #432, opened 2022-06-03
add GraphQL query to query vfolder with id

fregataa:
migrate from https://github.com/lablup/backend.ai-manager/pull/597

~~blocked by #574~~


achimnol (changes requested):
Please apply the dataloader to batch multiple vfolder queries.

achimnol (approved):

achimnol (changes requested):
I think the current implementation looks nice.

Though, let's extend it to allow non-admin users to query about their own vfolders with IDs while limiting query other user's vfolders.
Could we use something like `ai.backend.manager.models.base.scoped_query()`?

Since we need a separate multi-query implementation `ai.backend.manager.models.vfolder.query_accessible_vfolders()`, it might not be possible to directly reuse `scoped_query()` as-is.  How about adding an ID-based version of `query_accessible_vfolders()`?  For instance, we could fetch vfolders based on the IDs with privilege check conditions.

fregataa:
> Since we need a separate multi-query implementation `ai.backend.manager.models.vfolder.query_accessible_vfolders()`, it might not be possible to directly reuse `scoped_query()` as-is. How about adding an ID-based version of `query_accessible_vfolders()`? For instance, we could fetch vfolders based on the IDs with privilege check conditions.

privilege checking for vfolders is too complicated to simplify, I implemented ID-based version of `query_accessible_vfolders()` as very similar with `query_accessible_vfolders()`. Do I need to abstract privilege checking for RBAC?

achimnol:
This should be updated to work with #1386.

achimnol (approved):

achimnol (commented):
`scoped_query` will provide user-level ownership check.

But, we also need to check scenarios like:
- Querying project-owned vfolders by a normal user
- Querying either user-owned/project-owned vfolders by a superadmin

Ideally we need to have a separate `vfolder_ownership` relation table, but to narrow the scope of this PR, we could declare that "vfolder GraphQL API is only for user-owned vfolders for now" or "superadmin-only" as an alternative.

achimnol (approved):

PR #433, opened 2022-06-03
ci: Automate creation of GitHub releases

achimnol:
Let's automatically create the GitHub release when we make tags.


achimnol:
NOTE: The reason for CI failure by timeline-check is the test release has consumed the news fragment.

Issue #395, opened 2022-03-21
Add `session watch` command like `gh run watch`


achimnol:
For better CLI scriptability, let's add `backend.ai session watch` command to let it display the event stream of the target session and terminates with a proper exit code (e.g., 0 for task-success and 1 for others for batch-type sessions) when it finishes.

We could implement this feature entirely in the client-side, using the existing codebase of `backend.ai events`.

Reference: https://cli.github.com/manual/gh_run_watch

rapsealk:
For better understanding, `gh run watch` works like this:

<img width="448" alt="스크린샷 2022-05-23 오후 2 44 41" src="https://user-images.githubusercontent.com/14137676/169753266-8a7282b3-cb04-42c9-a3b4-9056327a2ff5.png">
<img width="574" alt="스크린샷 2022-05-23 오후 2 44 50" src="https://user-images.githubusercontent.com/14137676/169753268-54bc1aa0-0d91-4044-8c03-5c145b165743.png">

rapsealk:
https://user-images.githubusercontent.com/14137676/170036190-681f8112-b1c2-49d3-930a-671ff338d40f.mov

Is this what it expected to be?

PR #440, opened 2022-06-07
feat: Add `session watch` command to client-py

rapsealk:
This PR resolves lablup/backend.ai#395.

This PR implements a `backend.ai session watch` command, or `backend.ai watch` in short, to track and display the event stream of target session.

How to use:
- `backend.ai watch [SESSION_ID]`
- `backend.ai session watch [SESSION_ID]`

image

Limitations (Next things to do):
- It cannot know in which state the session is until it receives the first event.
- It does not check whether the output device is tty or not.


src/ai/backend/client/cli/session.py
```diff
@@ -861,3 +866,154 @@ async def _run_events():
...
+        async def _run_events_with_timeout(timeout: int):
+            assert timeout > 0
+            done, pending = await asyncio.wait({
+                asyncio.create_task(_run_events()),
+                asyncio.create_task(asyncio.sleep(timeout)),
+            }, return_when=asyncio.FIRST_COMPLETED)
+            for task in pending:
+                task.cancel()
+                with suppress(asyncio.CancelledError):
+                    await task
```
achimnol: This is clever. :wink:
But I think using `async_timeout` would make it more readable and intuitive.

achimnol (changes requested):

achimnol:
To merge, please regenerate the lockfiles after merging the main branch.

achimnol (approved):
I'm approving this PR as is for now, but I'd like to add support for `--output=json` with this command by printing a json object line-by-line. Let's work on it in a new follow-up PR.

rapsealk:
@achimnol How would it be different from each other; printing a json object line-by-line and run `backend.ai events` command?

rapsealk:
Now it supports both; console and json outputs.

* Console(default)

https://user-images.githubusercontent.com/14137676/173993900-418a4115-1498-4b2b-ac69-3e99a41c8bfb.mov

* JSON

https://user-images.githubusercontent.com/14137676/173993934-c0470a6a-8df7-47b3-aeab-f59da510ca72.mov

achimnol (changes requested):
Could we transparently pass the exit code of the session's startup command if the session is a batch-type session?

rapsealk:
Now we can retrieve `exitCode` from event data.

스크린샷 2022-06-22 오전 10 29 49

src/ai/backend/client/cli/session.py
```diff
@@ -926,16 +933,15 @@ def watch(session_name_or_id: str, owner_access_key: str, scope: str, max_wait:
         session_names = _fetch_session_names()
         if not session_names:
             click.echo('No matching items.')
-            return
+            sys.exit(0)
```
achimnol: This should be a non-zero exit code.

src/ai/backend/client/cli/session.py
```diff
@@ -926,16 +933,15 @@ def watch(session_name_or_id: str, owner_access_key: str, scope: str, max_wait:
         session_names = _fetch_session_names()
         if not session_names:
             click.echo('No matching items.')
```
achimnol: I think this echo should target stderr, not to confuse the JSON parse when `--output=json` is used.

src/ai/backend/client/cli/session.py
```diff
@@ -954,33 +960,40 @@ def watch(session_name_or_id: str, owner_access_key: str, scope: str, max_wait:
                     break
             else:
                 click.echo('No matching items.')
-                return
+                sys.exit(0)
```
achimnol: Same here.

src/ai/backend/client/cli/session.py
```diff
@@ -926,16 +933,15 @@ def watch(session_name_or_id: str, owner_access_key: str, scope: str, max_wait:
...
-            'session_started',
-            'kernel_terminating',
-            'kernel_terminated',
-            session_terminated,
+            KernelCreatingEvent.name,
+            KernelStartedEvent.name,
+            SessionStartedEvent.name,
+            KernelTerminatingEvent.name,
+            KernelTerminatedEvent.name,
+            SessionTerminatedEvent.name,
```
achimnol, line 944: 👍🏼

achimnol (changes requested):
Minor fixes please!

achimnol (changes requested):
How about using `ai.backend.client.output.print_fail()`?

achimnol:
Please resolve the merge conflicts!

rapsealk:
@achimnol Conflicts resolved.

src/ai/backend/client/cli/session.py
```diff
@@ -1,20 +1,38 @@
...
+    KernelCancelledEvent,
+    KernelCreatingEvent,
+    KernelStartedEvent,
+    KernelTerminatedEvent,
+    KernelTerminatingEvent,
+    SessionFailureEvent,
+    SessionStartedEvent,
+    SessionSuccessEvent,
+    SessionTerminatedEvent,
+)
```
achimnol: Currently `ai.backend.client` should not depend on `ai.backend.common`.
Here, the event bus is specific to the server-side components.
The client should treat the event names as specific string constants.

achimnol (changes requested):

src/ai/backend/client/cli/session.py
rapsealk: Thank you for the feedback! Now the event names are defined separately in `ai.backend.client`.

src/ai/backend/client/cli/session.py
```diff
@@ -995,23 +993,22 @@ def print_state(session_name_or_id: str, current_state_idx: int):
...
-                            break
+                    if ev.event == session_success:
+                        print_done(session_success)
+                        sys.exit(json.loads(ev.data).get('exitCode', 0))
+                    elif ev.event == session_failure:
+                        print_fail(session_failure)
+                        sys.exit(json.loads(ev.data).get('exitCode', 1))
+                    elif ev.event == kernel_cancelled:
+                        print_fail(kernel_cancelled)
+                        break
```
achimnol: We may use the `match case` statement. Why not?

src/ai/backend/client/cli/session.py
```diff
@@ -950,13 +938,23 @@ def watch(session_name_or_id: str, owner_access_key: str, scope: str, max_wait:
...
 
+        kernel_cancelled: Final[str] = 'kernel_cancelled'
+        kernel_creating: Final[str] = 'kernel_creating'
+        kernel_started: Final[str] = 'kernel_started'
+        kernel_terminated: Final[str] = 'kernel_terminated'
+        kernel_terminating: Final[str] = 'kernel_terminating'
+        session_failure: Final[str] = 'session_failure'
+        session_started: Final[str] = 'session_started'
+        session_success: Final[str] = 'session_success'
+        session_terminated: Final[str] = 'session_terminated'
```
achimnol: In Python, it is recommended to use uppercase names for constants. I think this may be defined in the top-level scope of the module or a new module.

src/ai/backend/client/cli/session.py
```diff
@@ -995,23 +993,22 @@ def print_state(session_name_or_id: str, current_state_idx: int):
             print_state(session_name_or_id, current_state_idx=-1)
```
achimnol: Could you explain what is the purpose of using the state index?

achimnol (changes requested):

src/ai/backend/client/cli/session.py
rapsealk: When comparing with any variable in `match case`, it seems that we have to wrap the constant(pattern) with string and I was not sure if it would be okay. Here is an example below:
```python
match ev.event:
    case str(session_success):
        # Do something here.
    case _:
        # Do something else here.
```
Reference: https://stackoverflow.com/questions/67525257/capture-makes-remaining-patterns-unreachable

If the session/kernel related constants are migrated into another package, then the problem would be solved.

src/ai/backend/client/cli/session.py
rapsealk: Since the state changes in serial order, we can use index to check if one kernel/session has reached to specific state. In this line, `current_state_idx=-1` means that the session is not reached to any state yet; even to the initial state.
Would it be better to set the default value of `current_state_idx` to `-1` and make it implicit?

rapsealk:
Now it works as below:

https://user-images.githubusercontent.com/14137676/177474929-19c1b05c-cd72-4412-9d7f-e18792f9f69a.mov

src/ai/backend/manager/api/events.py
```diff
@@ -254,7 +256,9 @@ async def _fetch():
     if row is None:
         return
     for q in app_ctx.session_event_queues:
-        q.put_nowait((event.name, row._mapping, event.reason))
+        exit_code = event.exit_code if isinstance(event, (KernelTerminatingEvent, KernelTerminatedEvent)) \
+                    else None
```
achimnol: For readability:
```suggestion
        exit_code = (
            event.exit_code
            if isinstance(event, (KernelTerminatingEvent, KernelTerminatedEvent))
            else None
        )
```

rapsealk:
Sorry, not if-statement but an if-expression.

achimnol (approved):

Issue #426, opened 2022-05-30
`backend.ai-tester` fails due to response mismatch


rapsealk:
**Is your feature request related to a problem? Please describe.**
기능 요청이 특정 문제에 연관된 것이라면 여기에 그 문제를 설명해주세요.
When I run `backend.ai test run-cli admin`, some cases fail because their assertion do not match to the response from the client of the latest version. For example, one case tries to find `User testaccount1@lablup.com is created` in the message to check if the user is created.
```python
# https://github.com/lablup/backend.ai-test/blob/c3258f2db4a2629157bb9c9ecdf18cfe99ec23ba/src/ai/backend/test/cli_integration/admin/test_user.py#L36
with closing(run(add_arguments)) as p:
    p.expect(EOF)
    assert 'User testaccount1@lablup.com is created' in p.before.decode(), 'Account add error'
```
However, the response from the latest version client is like below:
```json
{
  "ok": true,
  "msg": "success",
  "user": {
    "domain_name": "default",
    "email": "testaccount1@lablup.com",
    "username": "testaccount1",
    "uuid": "4d4f7dd5-9ae4-4b0e-aae2-c9c7bbe4ea72"
  }
}
```
Therefore, it should be changed to:
```python
with closing(run(add_arguments)) as p:
    p.expect(EOF)
    response = json.loads(p.before.decode())
    assert (response.get('ok') is True) and (response.get('msg') == 'success'), 'Account creation failed: Account#1'
```

**Describe the solution you'd like**
Change the condition to compare with the latest version of response, to verify if the function works.

**Describe alternatives you've considered**
...

**Additional context**
...

rapsealk:
Another found in `ai.backend.cli_integration.admin.test_user`.
Current assertion:
```python
# https://github.com/lablup/backend.ai-test/blob/c3258f2db4a2629157bb9c9ecdf18cfe99ec23ba/src/ai/backend/test/cli_integration/admin/test_user.py#L177
with closing(run(['admin', 'user', 'purge', 'testaccount1@lablup.com'])) as p:
    p.sendline('y')
    p.expect(EOF)
    assert 'User is deleted:' in p.before.decode(), 'Account deletion failed: Account#1'
```
Actual response:
```
Field    Value
-------  -----------------------
ok       True
msg      success
email    testaccount1@lablup.com
```

PR #442, opened 2022-06-07
fix: Update integration test to work with the new repo structure

rapsealk:
Some test cases fail because of their outdated assertions; the responses from the latest client do not contain the expected message anymore. I found such mismatches from methods below:
- `ai.backend.test.cli_integration.admin.test_user#test_add_user`
- `ai.backend.test.cli_integration.admin.test_user#test_delete_user`

and they are fixed in this PR.

For details, please check the referenced issue below.

Refs:
- lablup/backend.ai#426

### DIFF
--- a/changes/442.fix.md
+++ b/changes/442.fix.md
@@ -1,0 +1,1 @@
+Change formats in test assertions to JSON for fixing mismatches and future consistency.
--- a/changes/442.fix.md
+++ b/changes/442.fix.md
@@ -1,1 +1,1 @@
-Change formats in test assertions to JSON for fixing mismatches and future consistency.
+Update test assertions to utilize the JSON output of mutation commands for forward compatibility
--- a/scripts/install-dev.sh
+++ b/scripts/install-dev.sh
@@ -619,10 +619,8 @@ sed_inplace "s/https:\/\/api.backend.ai/http:\/\/127.0.0.1:${MANAGER_PORT}/" ./w
 sed_inplace "s/ssl-verify = true/ssl-verify = false/" ./webserver.conf
 sed_inplace "s/redis.port = 6379/redis.port = ${REDIS_PORT}/" ./webserver.conf
 
-cp configs/testers/sample-env-tester.sh ./env-tester-admin.sh
-echo "BACKENDAI_TEST_CLIENT_ENV=$( cd -- "$( dirname -- "${BASH_SOURCE[0]:-$0}"; )" &> /dev/null && pwd 2> /dev/null; )/env-local-admin-api.sh" >> ./env-tester-admin.sh
-cp configs/testers/sample-env-tester.sh ./env-tester-user.sh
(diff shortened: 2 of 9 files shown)


src/ai/backend/test/cli_integration/conftest.py
```diff
@@ -16,18 +16,8 @@
...
-    return Path(p)
-
-
-@pytest.fixture(scope="session")
-def client_bin(
-    client_venv: Path,
-) -> Path:
-    return client_venv / 'bin' / 'backend.ai'
+def client_bin() -> Path:
+    return Path(os.path.abspath(os.path.dirname(__file__))) / '..' / '..' / '..' / '..' / '..' / 'backend.ai'
```
achimnol: It is better to utilize `pathlib.Path` APIs because `..` may be an OS-specific special name though practically we don't have to worry about.

```suggestion
    return Path(__file__).parent.parent.parent.parent.parent.parent.parent / 'backend.ai'
```

But, instead of hard-coding the source directory hierarchy, I'd like to suggest that let's first [search `BUILD_ROOT` file using `ai.backend.plugin.entrypoint.find_build_root()`](https://github.com/lablup/backend.ai/blob/main/src/ai/backend/plugin/entrypoint.py#L107) and use it as the reference point.

scripts/install-dev.sh
```diff
@@ -619,9 +619,8 @@ sed_inplace "s/https:\/\/api.backend.ai/http:\/\/127.0.0.1:${MANAGER_PORT}/" ./w
 sed_inplace "s/ssl-verify = true/ssl-verify = false/" ./webserver.conf
 sed_inplace "s/redis.port = 6379/redis.port = ${REDIS_PORT}/" ./webserver.conf
 
-# TODO
-## cp configs/testers/sample-env-tester.sh ./env-tester-admin.sh
-## cp configs/testers/sample-env-tester.sh ./env-tester-user.sh
+echo "BACKENDAI_TEST_CLIENT_ENV=$( cd -- "$( dirname -- "${BASH_SOURCE[0]:-$0}"; )" &> /dev/null && pwd 2> /dev/null; )/env-local-admin-api.sh" > ./env-tester-admin.sh
+echo "BACKENDAI_TEST_CLIENT_ENV=$( cd -- "$( dirname -- "${BASH_SOURCE[0]:-$0}"; )" &> /dev/null && pwd 2> /dev/null; )/env-local-user-api.sh" > ./env-tester-user.sh
```
achimnol, line 623: 👍🏼

changes/442.fix.md
```diff
@@ -0,0 +1 @@
+Change formats in test assertions to JSON for fixing mismatches and future consistency.
```
achimnol:
```suggestion
Update test assertions to utilize the JSON output of mutation commands for forward compatibility
```

achimnol (changes requested):

src/ai/backend/test/cli_integration/conftest.py
rapsealk: Thank you for the feedback! It's good to know more about `pathlib.Path` and our plugins.

achimnol (approved):

PR #443, opened 2022-06-07
feat: add support for weka.io backend

kyujin-cho:
This PR is migrated from lablup/backend.ai-storage-proxy#45.


src/ai/backend/storage/weka/weka_client.py
```diff
@@ -0,0 +1,383 @@
...
+        if method == "GET":
+            func = sess.get
+        elif method == "POST":
+            func = sess.post
+        elif method == "PUT":
+            func = sess.put
+        elif method == "PATCH":
+            func = sess.patch
+        else:
+            func = sess.delete
```
achimnol: Just FYI: Now we can use the `match` statement. :wink:

src/ai/backend/storage/weka/weka_client.py
```diff
@@ -0,0 +1,383 @@
...
+            data["total_budget"],
+        )
+
+
+def error_handler(inner):
+    async def outer(*args, **kwargs):
+        try:
+            return await inner(*args, **kwargs)
+        except web.HTTPNotFound:
+            raise WekaNotFoundError
```
achimnol, line 98: Wouldn't it better to handle other HTTP errors and wrap them as `WekaAPIError`?

kyujin-cho: Can you tell me more about "wrapping"? Do you mean handling other error cases (400, 500, ...) by creating new exception classes which inherits from `WekaAPIError`?

src/ai/backend/storage/weka/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+
+        try:
+            cluster_info = await self.api_client.get_cluster_info()
+            quotas = await self.api_client.list_quotas(self._fs_uid)
+            return {
+                "status": health_status,
+                "status_info": None,
+                "metadata": {
+                    "quota": json.dumps([q.to_json() for q in quotas]),
+                    "cluster_info": cluster_info,
```
lizable: Since the return type "[HardwareMetaData](https://github.com/lablup/backend.ai/blob/96033aa276534c1f7b14eef3932d88fb0f68c485/src/ai/backend/common/types.py#L224-L227)"  definition says that key `metadata` allows string type only, not map type, Therefore `cluster_info` should be converted to stringified JSON unless we update the type definition to allow multi-level key-value.

lizable: Like this:
`"cluster_info": json.dumps(cluster_info),`

src/ai/backend/storage/weka/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+        vfpath = self.mangle_vfpath(vfid)
+        inode_id = await self._get_inode_id(vfpath)
+        weka_path = (
+            vfpath.absolute()
+            .as_posix()
+            .replace(self.mount_path.absolute().as_posix(), "")
+        )
+        if not weka_path.startswith("/"):
+            weka_path = "/" + weka_path
+        await self.api_client.set_quota_v1(weka_path, inode_id, hard_limit=size_bytes)
```
lizable, line 169: I found that we can apply the latest API of Weka, instead of using the previous(v1) one. Although getting the absolute path of virtual folder(`vfpath`) is necessary, we may not need to get `weka_path`.
Please refer to code description [here]().

lizable: Instead of calling v1 API in [L173](https://github.com/lablup/backend.ai/pull/443/files#diff-02f55cb868aea9b9275c6fd5b7ee34d552813e4b749c158094cc05f0ca0258f8R173), I would like to suggest using v2 API as following statement below:
`await self.api_client.set_quota(vfid, inode_id, hard_limit=size_bytes)`

src/ai/backend/storage/weka/weka_client.py
```diff
@@ -0,0 +1,388 @@
...
+
+    @error_handler
+    async def set_quota(
+        self,
+        fs_uid: str,
+        inode_id: int,
+        soft_limit: BinarySize,
+        hard_limit: BinarySize,
+    ) -> WekaQuota:
+        def _format_size(s: BinarySize) -> str:
```
lizable, line 272: We shall use `finite_from_str` function defined in `BinarySize` class instead of using inner function `_format_size`, since both `soft_limit_bytes` and `hard_limit_bytes` requires numbers in byte unit, not human-readable value, such as 2.5 GiB(2684354560 bytes).

src/ai/backend/storage/weka/weka_client.py
```diff
@@ -0,0 +1,388 @@
...
+        }
+
+        async with aiohttp.ClientSession(
+            base_url=self.api_endpoint,
+        ) as sess:
+            response = await self._build_request(
+                sess,
+                "PUT",
+                f"/fileSystems/{fs_uid}/quota/{inode_id}",
+                body,
```
lizable, line 292: In [L287](https://github.com/lablup/backend.ai/pull/443/files#diff-36cfe4b50d0a1c978df8cb77d6e45162c6df438ed66997fc94e8d231218b00cbR287), sending Mapping type value triggers parsing error. Let's wrap with `json.dumps()` function so that it will automatically be stringified.

src/ai/backend/storage/weka/weka_client.py
```diff
@@ -0,0 +1,388 @@
...
+            )
+            data = await response.json()
+
+        quota_id = data["data"].keys()[0]
+        return WekaQuota(
```
lizable, line 297: Since we adopted "directory-based" quota strategy from weka integration, quota_id received from weka API server would be only one.
There's no need to parsing "quota_id" from response. Instead, Let's just leave quota_id as None.(or It may be better to get removed from member variable in [WekaQuota](https://github.com/lablup/backend.ai/pull/443/files#diff-36cfe4b50d0a1c978df8cb77d6e45162c6df438ed66997fc94e8d231218b00cbR21-R26) class).

lizable: In that case, `quota_id` key in [L294-L297 ](https://github.com/lablup/backend.ai/pull/443/files#diff-36cfe4b50d0a1c978df8cb77d6e45162c6df438ed66997fc94e8d231218b00cbR294-R297)should be removed likewise.

lizable (changes requested):
There are several feedbacks from E2E test. Please refer to the comments below.
And also, we need to decide whether we should execute Weka CLI on quota setting or not. Because setting a quota for the virtual folder via API if the virtual folder is not empty is not allowed, but is only available in CLI.

There are two approaches to solving this problem.
* **Approach 1. Set default quota when creating virtual folder.**
   In this case, the default quota would be the min value among two values described below:
   - Max size of vfolder in keypair resource policy that current user's crendential are belonged.
   - Overall size (allowed) of weka storage
   
   Pros: Easy to apply and implement.
   Cons: May ruin the "degree of freedom" of quota setting, since It always set quota.


* **Approach 2. Set folder empty checking condition, so when quota via API if vfolder is empty, and execute CLI if it's not empty.**
   This approach will require login (with session) to weka cluster when executing storage-proxy.
   
   Pros: Users (superadmin role) can set a quota whenever they want, they don't need to set quota every time.
   Cons: A bit trickier compare to Approach 1.

lizable:
@kyujin-cho 
One more request: 
str type need to be changed into UUID in [here](https://github.com/lablup/backend.ai/blob/96033aa276534c1f7b14eef3932d88fb0f68c485/src/ai/backend/storage/api/manager.py#L410).
According to [ABC function of `set_quota`](https://github.com/lablup/backend.ai/blob/96033aa276534c1f7b14eef3932d88fb0f68c485/src/ai/backend/storage/abc.py#L143-L145), the type of parameter "vfid" is UUID. so the type should be changed to "UUID" instead of "str".

src/ai/backend/storage/weka/__init__.py
kyujin-cho: v1 API lacks support of setting quota only with hard/soft limit.

src/ai/backend/storage/weka/weka_client.py
kyujin-cho: Well, I guess `finite_from_str()` function is for converting `BinarySize`-like objects to `BinarySize`.  We need an opposite logic; converting `BinarySize` to `str`.

kyujin-cho:
> @kyujin-cho One more request: str type needs to be changed into UUID in [here](https://github.com/lablup/backend.ai/blob/96033aa276534c1f7b14eef3932d88fb0f68c485/src/ai/backend/storage/api/manager.py#L410). According to [ABC function of `set_quota`](https://github.com/lablup/backend.ai/blob/96033aa276534c1f7b14eef3932d88fb0f68c485/src/ai/backend/storage/abc.py#L143-L145), the type of parameter "vfid" is UUID. so the type should be changed to "UUID" instead of "str".

Hmm, that seems like quite irrelevant to this PR. Please create a separate issue.

src/ai/backend/storage/weka/weka_client.py
kyujin-cho: I think we can solve this easily, by changing `body` keyword parameter of `func()` call in L193 to `json`. Check https://docs.aiohttp.org/en/stable/client_advanced.html#custom-request-headers.

src/ai/backend/storage/weka/weka_client.py
kyujin-cho: Writing `weka_client.py` like is to easily adapt and expand its code to other projects. Of course, ignoring `quota_id`  also will be a good if we're going to use `WekaAPIClient` class only in storage-proxy, but I don't think that approach won't be a extensible one. Please add your opinion about this.

src/ai/backend/storage/weka/weka_client.py
lizable: You're right. We don't need to change it. let's just leave them as is.

src/ai/backend/storage/weka/weka_client.py
lizable: It makes sense to me. How about converting `body` to string with using `json.dumps()` like [here](https://github.com/lablup/backend.ai/pull/443/files#diff-36cfe4b50d0a1c978df8cb77d6e45162c6df438ed66997fc94e8d231218b00cbR332)? 
If not, I get parsing error like this.
Screen Shot 2022-06-22 at 8 35 02 PM

lizable (changes requested):
I added a new PR(#487) based on your suggestion since it's related but not deeply involved in the main issue. 
Also, I found that there's one more thing to be better fixed although we are not using it for now. Except that everything is working fine.
Please refer to the comments below.

lizable:
@achimnol Btw, I wonder if we shall add test code for weka storage as we did to other storage hosts (e.g. `xfs`, `purestorage`, `netapp`, etc.)

src/ai/backend/storage/weka/weka_client.py
kyujin-cho: :( my bad. There was a typo on the code. Fixed, so it should work now.

src/ai/backend/storage/weka/exceptions.py
```diff
@@ -0,0 +1,39 @@
...
+    name = 'WekaAPIError'
+    message = ''
+
+    def __init__(self, message: str = None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.message = message or 'Unknown Error'
+
+    def __str__(self) -> str:
+        return f'{self.name}: {self.message}'
```
achimnol: This could be replaced with `type(self).__name__` instead of duplicating exception names as string attributes.
```python
>>> class MyAwesomeError(Exception):
...   def __str__(self):
...     return type(self).__name__
...
>>> str(MyAwesomeError())
'MyAwesomeError'
```

src/ai/backend/storage/weka/exceptions.py
```diff
@@ -0,0 +1,39 @@
...
+class WekaInitError(WekaError):
+    pass
+
+
+class WekaAPIError(WekaError):
+    name = 'WekaAPIError'
+    message = ''
+
+    def __init__(self, message: str = None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
achimnol: It should transparently pass the `message` argument as the part of `args` passed to the super consturctor (See also: [`args` of `BaseException`](https://docs.python.org/3/library/exceptions.html#BaseException.args), [PEP-352](https://peps.python.org/pep-0352/)).
[Stripping arguments for the super constructor affects *picklability* of exceptions.](https://stackoverflow.com/questions/16244923/how-to-make-a-custom-exception-class-with-multiple-init-args-pickleable)  If exceptions are not picklable, it will make it hard to read the stderr exception logs when the service uses `ai.backend.common.logging` with a multi-process architecture.

```python
class WekaAPIError(WekaError):
    message: str

    def __init__(self, message: Optional[str] = None, *args):
        # Unless we explicitly utilize keyword arguments,
        # there is no need to pass them to the super constructor.
        super().__init__(message, *args)
        # just additionally define typed attributes
        self.message = message or ""
```

src/ai/backend/storage/weka/exceptions.py
achimnol: Also, if you utilize `self.args`, you could just use `repr()` to get a canonical string representation of the exception class with all its arguments.

```python
>>> class MyXXError(Exception):
...   pass
...
>>> x = MyXXError('abc', 123)
>>> str(x)
"('abc', 123)"
>>> repr(x)
"MyXXError('abc', 123)"
```

src/ai/backend/storage/weka/weka_client.py
achimnol: Hm, please ignore the above comment. Since this is a server-side code, so I think we don't need to mixin `aiohttp.HTTPException` classes here but we need to wrap these internal exceptions in the manager-facing or client-facing **API handlers**.

src/ai/backend/storage/README.md
```diff
@@ -160,3 +162,12 @@ Please DO NOT remove Backend.AI mapped qtree in NetApp ONTAP Sysmgr GUI. If not,
 
 > NOTE:   
 Qtree name in configuration file(`storage-proxy.toml`) must have the same name created in NetApp ONTAP Sysmgr.
+
+### Weka.IO
+
+#### Prerequisites
+
+* Weka.IO agent installed and running
+* Weka.IO filesystem mounted under `/vfroot`
```
achimnol: `/vfroot` is an arbitrary sample text rather than a strict prerequisite, though it is used as the default config in the relative form by `install-dev.sh`.
Please rephrase it to indicate the general configuration option.

achimnol (changes requested):

achimnol (approved):

achimnol:
> @achimnol Btw, I wonder if we shall add test code for weka storage as we did to other storage hosts (e.g. `xfs`, `purestorage`, `netapp`, etc.)

Currently I don't have ideas to handle such integration tests with external vendor devices, without writing a fully mocked API server, when the vendor devices are only available for a specific period.  Let's handle this with separate acceptance tests when delivered to customer sites.

Issue #309, opened 2021-11-24
Metadata URL support for security and rich features


achimnol:
Metadata URLs in clouds (e.g., 169.254.169.254 in AWS EC2) exposes many system detail information to containers.We need to restrict some of those information and also need to provide a more flexible way of interacting with the agent and containers and user programs.

Using the following command in the agent host, access to `http://169.254.169.254` is redirected to `172.17.0.1:8000`, where 172.17.0.1 is the IP address of the `docker0` interface, while still allowing the host's metadata access:

```shell
$ iptables -t nat \
    -I PREROUTING \
    -p tcp \
    -d 169.254.169.254 \
    --dport 80 \
    -j REDIRECT \
    --to-ports 8000 \
    -i docker0
```

Let's make the agent to listen on a specific local port on the docker0 interface and redirect the metadata URL traffic to it.After then, we could expand features based on container-agent interactions.We need to detect `docker0` IP address at runtime or at least it should be configurable, because it may be different site-by-site.

JIRA Issue: BA-285

achimnol:
@kyujin-cho The internal port number 40128 introduced in lablup/backend.ai#319 should be configurable in the local config and by `install-dev.sh` to allow multiple agents in a single host (#496).

PR #447, opened 2022-06-08
feat: make container metadata server listening port configurable

kyujin-cho:
refs #309, #496 

This PR is migrated from lablup/backend.ai-agent#361.


achimnol (approved):

achimnol:
Need to investigate why `tests/agent/docker/test_agent.py` hangs forever due to infinite retries on Redis connections, while it seems that the Redis container fixture works fine.

PR #448, opened 2022-06-08
feat: make container metadata server extensible

kyujin-cho:
This PR is migrated from lablup/backend.ai-agent#333.


achimnol (changes requested):
Let's just rename "WebappPlugin" to "MetadataPlugin" for clarity.
I think other parts look good. :)

achimnol (approved):

PR #451, opened 2022-06-08
fix: agent startup error due to unbound local error in dumping the last registry

adrysn:
- fix: always dump kernel registry information upon agent termination
- Add news fragment
- fix: typo
- fix: UnboundLocalError in starting the agent server


adrysn:
Close this PR since it mis-reference the branch.

Issue #464, opened 2022-06-12
Test cases are being interfered by errors from plugin checking


rapsealk:
**Describe the bug**
1. There is a line in `./py`, which is to get a list of plugins. If there is no plugins, then it throws `ls: cannot access 'plugins/*/': No such file or directory` message.
https://github.com/lablup/backend.ai/blob/38db104af84e4360571ad75cdd38afa5b65330ce/py#L15
2. Each test case is based on standard output message; it has a belief that when given `--output=json` option, each message is printed in JSON format. Therefore, it there is any other message that is not JSON format, test case cannot work properly.

**To Reproduce**
문제를 재현하기 위한 순서를 자세히 적어주세요:
1. Run `./backend.ai mgr start-server --debug`
2. Run `./backend.ai ag start-server --debug`
3. Run `./backend.ai test run-cli admin.test_user`

**Expected behavior**
Each test case successfully parses(JSONify) standard output messages.

**Screenshots**
![image](https://user-images.githubusercontent.com/14137676/173235594-927717ad-cbce-4fdb-96a7-f29993e76ed4.png)
https://github.com/lablup/backend.ai/blob/38db104af84e4360571ad75cdd38afa5b65330ce/py#L15

**Client:**
 - OS: Windows 11 21H2 (WSL2 Ubuntu 20.04)
 - App version: -
 - Browser version: -
 - Python version: 3.10.4
 - NodeJS version: -

**Server:**
 - OS: Windows 11 21H2 (WSL2 Ubuntu 20.04)
   - Output of `uname -a`: Linux DESKTOP-78S6JQL 5.4.72-microsoft-standard-WSL2#1 SMP Wed Oct 28 23:40:43 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
 - Backend.AI version: main
 - Python version: 3.10.4
 - Installation method: scripts/install-dev.sh

**Additional context**
-

PR #465, opened 2022-06-12
fix: Ignore plugins-not-found message

rapsealk:
Refs:
- lablup/backend.ai#464

Changed it to ignore messages from `ls -d plugins/*/` when there is no plugin.


achimnol (approved):
👍🏼

Issue #352, opened 2022-01-19
Let's replace hard coded test data and make use of fixture function


fregataa:
[Backend.ai test](https://github.com/lablup/backend.ai-test) uses pytest module.
it create dummy data to test but it depends on hard coded data such as username or email.
Let's use Faker or something similar.
And some dummy data which is used as a dependency -such as foreign key in db- is created and destroyed repeatedly during testing.
Let's make use of fixture to avoid such repeated codes.

PR #466, opened 2022-06-13
fix: Generate dummy data using `faker`

rapsealk:
This PR resolves lablup/backend.ai#352.

For some data such as `users`, the same data are used in several test cases/modules, but defined in each source file separately. Therefore, it would be better to make them as `pytest.fixture` to maintain easily.
Also, since current test data are hard-coded, they can cause problem on database integrity constraint in case some test cases fail to run properly and retry to run with the same data.

### DIFF
--- a/changes/462.doc
+++ b/changes/462.doc
@@ -1,0 +1,1 @@
+Migrate unmerged doc translation to monorepo, which is about FAQ and Key concept.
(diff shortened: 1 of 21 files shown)


achimnol (approved):
Great! Thanks for introducing `faker`.

TIP: When you regenerate lockfiles, use `--resolve=python-default` option so that Pants updates the main lock file only without modifying tool lockfiles, unless necessary.

achimnol:
Please resolve the conflict!

achimnol:
image
Backporting this reveals that I've missed backporting the prior integration test suite update PR.

Issue #404, opened 2022-03-24
support to install backend.ai for openSUSE/SUSE Linux distribution 


studioego:
**Is your feature request related to a problem? Please describe.**
기능 요청이 특정 문제에 연관된 것이라면 여기에 그 문제를 설명해주세요.

For openSUSE/SUSE Linux users, It doesn't install backend.ai using `install-dev.sh`.
Because `install-dev.sh` file is supported only both `debian-based distributions`(such as. Ubuntu, Debian, etc) & `redhat-based distributions`(such as Centos, Redhat etc).

However, openSUSE & SUSE Linux distribution uses a unique system(yast2 - system manager, zypper - software manager, etc) unlike Redhat based linux distributions

As a result, I need to modify `install-dev.sh` for opensuse/SUSE Linux distribution.

**Describe the solution you'd like**
어떤 기능이 있으면 좋겠는지 자세히 설명해주세요.

support to install  backend.ai for openSUSE/SUSE Linux Distribution users 
https://github.com/lablup/backend.ai/blob/main/scripts/install-dev.sh#L149

 modify `install-dev.sh` for opensuse/SUSE Linux distribution.

**Describe alternatives you've considered**
혹시 다른 대안들을 생각해본 적이 있다면 함께 적어주세요.

**Additional context**
기능 요청에 대해 보다 잘 이해할 수 있는 다른 맥락을 기술해주세요.

studioego:
opensuse doesn't support to Docker CE for Linux installation script https://get.docker.com/

studioego:
AWS's SUSE Linux Enterprise Server 12 instance doesn't have jq package

studioego:
AWS's SUSE Linux Enterprise Server 15 SP3 instance doesn't have git-lfs package.
```
ec2-user@ip-10-122-3-27:~> sudo zypper search git
'Basesystem_Module_x86_64' 서비스를 새로 고치는 중입니다.
'Containers_Module_x86_64' 서비스를 새로 고치는 중입니다.
'Desktop_Applications_Module_x86_64' 서비스를 새로 고치는 중입니다.
'Development_Tools_Module_x86_64' 서비스를 새로 고치는 중입니다.
'Legacy_Module_x86_64' 서비스를 새로 고치는 중입니다.
'Public_Cloud_Module_x86_64' 서비스를 새로 고치는 중입니다.
'Python_2_Module_x86_64' 서비스를 새로 고치는 중입니다.
[...]

PR #474, opened 2022-06-14
feat: Support installation on openSUSE

studioego:
Feat: Support for openSUSE Tumbleweed installation
Only the openSUSE Tumbleweed supports docker-compose v2.2, so only this
distribution is supported.
The lastest openSUSE Leap v15.4's docker-compose version is 1.25.
Also, The lastest SLES 15 SP3's docker-compose version is 1.25.
As a result, It only supported "openSUSE Tumbleweed", rolling release version.
close https://github.com/lablup/backend.ai/issues/404

### DIFF
--- a/changes/474.feature.md
+++ b/changes/474.feature.md
@@ -1,1 +1,1 @@
-Support for openSUSE Tumbleweed installation
+Support for openSUSE (Leap and Tumbleweed) installation
--- a/changes/474.feature.md
+++ b/changes/474.feature.md
@@ -1,0 +1,1 @@
+Support for openSUSE Tumbleweed installation
--- a/scripts/check-docker.py
+++ b/scripts/check-docker.py
@@ -1,8 +1,8 @@
 import functools
+import json
 import re
 import subprocess
 import sys
-import os
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


achimnol (changes requested):
Let's use `$docker_compose` variable in `install-dev.sh` so that we don't have to make branches everywhere we invoke `docker compose`.
I think we can make `check-docker.py` to print the detetced docker compose v2 command to stdout and let `install-dev.sh` to assign it to the variable.

studioego:
I tested backend.ai installing on both openSUSE Tumbleweed x86_64 platform and the one aarch64 platform.
on aarch64 architecture, Docker compose installed version is v2.2.3
``` Docker Compose version v2.2.3```
However, x86_64 platform, Docker compose installed version is 1.28.5
```docker-compose version 1.28.5```

So, I hold on issue  https://github.com/lablup/backend.ai/issues/404 and PR https://github.com/lablup/backend.ai/pull/474


If the docker-compose version is version 2 on all platforms of openSUSE Tumbleweed, I will re-submit the PR.

studioego (commented):
changed scripts

PR #478, opened 2022-06-16
fix: skip measuring the stat for an agent registry item if it has not yet assigned container ID to prevent the occasional unhandled `UnboundLocalError`

adrysn:
Sometimes, the `UnboundLocalError` is raised due to the missing container ID in one of the agent registry items. To avoid this unhandled exception, this PR skips adding, or referencing, the container ID from the `kernel_id_map`.

image

### DIFF
--- a/src/ai/backend/agent/stats.py
+++ b/src/ai/backend/agent/stats.py
@@ -366,7 +366,7 @@ class StatContext:
                     cid = info['container_id']
                 except KeyError:
                     log.warning('collect_container_stat(): no container for kernel {}', kid)
-                    continue
+                else:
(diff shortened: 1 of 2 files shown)


src/ai/backend/agent/stats.py
```diff
@@ -366,6 +366,7 @@ async def collect_container_stat(
                     cid = info['container_id']
                 except KeyError:
                     log.warning('collect_container_stat(): no container for kernel {}', kid)
+                    continue
```
achimnol: How about this for better readability?
```suggestion
                else:
```

PR #479, opened 2022-06-17
fix: Add detection logic of HT sibling core count

hydroxyde:
This PR refers to: lablup/backend.ai-internal#51
Current implementation does not concider hyperthreaded cores. Hwtool reports the number of physical cores only and license server activate physical cores without hyperthreaded cores. However, license server detect a higher total number of cores if hyperthread is activated and this raises an invalidation error.
With this PR we can detect hyperthreading. Make Activator check for hyperthreaded cores in order to remove them from the cores license has to activate.

### DIFF
--- a/src/ai/backend/agent/vendor/linux.py
+++ b/src/ai/backend/agent/vendor/linux.py
@@ -41,8 +41,23 @@ class libnuma:
                     data = await resp.json()
                     return {idx for idx in range(data['NCPU'])}
         except aiohttp.ClientError:
+            file_path = os.path.join("/sys/devices/system/cpu/cpu0/topology/thread_siblings_list")
+            cpu_count = os.cpu_count()
+            num_htSiblings = 0
             try:
-                return os.sched_getaffinity(os.getpid())
+                with open(file_path) as siblings:
+                    htSiblings = len(siblings.read().split(','))
+                    if cpu_count is not None:
+                        num_htSiblings = cpu_count // htSiblings
+            except (IOError, OSError) as e:
+                raise Exception(
+                    "{}\nCould not read thread siblings list".format(e))
+            try:
+                restricted_pid_cpu = os.sched_getaffinity(os.getpid())
+                for cpu in list(restricted_pid_cpu):
+                    if cpu > num_htSiblings - 1:
+                        restricted_pid_cpu.discard(cpu)
+                return restricted_pid_cpu
             except AttributeError:
                 return {idx for idx in range(os.cpu_count())}
 
--- a/src/ai/backend/agent/vendor/linux.py
+++ b/src/ai/backend/agent/vendor/linux.py
@@ -31,6 +31,23 @@ class libnuma:
             return 1
 
     @staticmethod
+    def htcpu_info():
+        file_path = os.path.join("/sys/devices/system/cpu/cpu0/topology/thread_siblings_list")
+        cpu_count = os.cpu_count()
+        num_htSiblings = 0
+        try:
+            with open(file_path) as siblings:
+                htSiblings = len(siblings.read().split(','))
+                if htSiblings > 0:
+                    num_htSiblings = cpu_count // htSiblings
+                    return({'isActive': True, 'num_htcpu': num_htSiblings})
+                else:
+                    return({'isActive': False, 'num_htcpu': 0})
+        except (IOError, OSError) as e:
+            raise Exception(
+                "{}\nCould not read thread siblings list".format(e))
+
+    @staticmethod
     @aiotools.lru_cache(maxsize=1)
     async def get_available_cores():
         try:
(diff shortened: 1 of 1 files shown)


src/ai/backend/agent/vendor/linux.py
```diff
@@ -30,6 +30,23 @@ def num_nodes():
         else:
             return 1
 
+    @staticmethod
+    def htcpu_info():
+        file_path = os.path.join("/sys/devices/system/cpu/cpu0/topology/thread_siblings_list")
```
adrysn: This is platform-dependent code. How about using a more platform-independent code by utilizing `psutil` like this reference: https://stackoverflow.com/a/62653110.

adrysn (changes requested):

src/ai/backend/agent/vendor/linux.py
```diff
@@ -30,6 +31,20 @@ def num_nodes():
         else:
             return 1
 
+    @staticmethod
+    async def htcpu_info():
```
achimnol, line 35: Let's rename it to `get_core_thread_count()` as "hyper-thread" is an Intel-specific term.

src/ai/backend/agent/vendor/linux.py
```diff
@@ -41,8 +56,13 @@ async def get_available_cores():
                     data = await resp.json()
                     return {idx for idx in range(data['NCPU'])}
         except aiohttp.ClientError:
+            num_htSiblings = await libnuma.htcpu_info()['num_htcpu']
```
achimnol, line 59: Let's keep Pythonic naming: `num_htSiblings` -> `num_core_threads` (aligning with the above comment).

src/ai/backend/agent/vendor/linux.py
```diff
@@ -30,6 +31,20 @@ def num_nodes():
         else:
             return 1
 
+    @staticmethod
+    async def htcpu_info():
+        num_htSiblings = 0
```
achimnol, line 36: Same here: `num_htSiblings` -> `num_core_threads`.

src/ai/backend/agent/vendor/linux.py
```diff
@@ -30,6 +31,20 @@ def num_nodes():
...
+        try:
+            htSiblings = psutil.cpu_count(logical=False)
+            if htSiblings > 0:
+                num_htSiblings = psutil.cpu_count() // htSiblings
+                return({'isActive': True, 'num_htcpu': num_htSiblings})
+            else:
+                return({'isActive': False, 'num_htcpu': 0})
```
achimnol, line 43: I think we could just return 1 if HT is disabled or there is no multiple threads per core since any CPU core has the capability of running at least one thread anyway.  There is no need to define arbitrary data-transfer objects like this dict.

Also remove the parenthesis around the return statement unless you are returning a tuple or a complex expression that does not fit in a single line.

src/ai/backend/agent/vendor/linux.py
```diff
@@ -30,6 +31,20 @@ def num_nodes():
...
+        except OSError as e:
+            raise Exception(
+                "{}\nCould not read thread siblings".format(e))
```
achimnol, line 46: Don't put arbitrary newlines in the exception messages.
Also avoid raising `Exception` but use more specific concrete exception class such as `RuntimeError`.

src/ai/backend/agent/vendor/linux.py
```diff
@@ -41,8 +56,13 @@ async def get_available_cores():
...
                     return {idx for idx in range(data['NCPU'])}
         except aiohttp.ClientError:
+            num_htSiblings = await libnuma.htcpu_info()['num_htcpu']
             try:
-                return os.sched_getaffinity(os.getpid())
+                restricted_pid_cpu = os.sched_getaffinity(os.getpid())
+                for cpu in list(restricted_pid_cpu):
+                    if cpu > num_htSiblings - 1:
+                        restricted_pid_cpu.discard(cpu)
+                return restricted_pid_cpu
```
achimnol, line 65: This routine is for when the agent runs inside a container.
For licensing we should exclude HT logical cores but there is no need to do so with myself available cores.

src/ai/backend/agent/vendor/linux.py
```diff
@@ -30,6 +31,20 @@ def num_nodes():
         else:
             return 1
 
+    @staticmethod
+    async def htcpu_info():
+        num_htSiblings = 0
+        try:
+            htSiblings = psutil.cpu_count(logical=False)
```
achimnol, line 38: ~~I'm not sure `psutil.cpu_count()` handles the case where the agent runs inside a container, where we should use the process' CPU core affinity mask to get the available number of cores.
In that case, I think we should not use the entire logic for excluding HT cores.~~

Let's write the code to branch by whether to call `psutil.cpu_count(logical=True)` or `psutil.cpu_count(logical=False)` depending on the context: reporting the available resource slots vs. reporting the CPU cores to be activated, without complex logic.

achimnol (changes requested):

src/ai/backend/agent/vendor/linux.py
achimnol: I think we should consider the case that the agent runs inside a container.
In such case, we should skip the entire logic for HT core detection and just rely on the CPU affinity mask of the process (`os.sched_getaffinity(os.getpid())`).

achimnol: Hmmm. But in that case, we should consider whether we expose the host's resources or the container's resources.  In the context of Backend.AI, we should use the host's resources and then we should detect the host-side CPU count and apply HT core exclusion logic.

src/ai/backend/agent/vendor/linux.py
achimnol: I think we should just remove use of `os.sched_getaffinity(os.getpid())` entirely because this routine is for detecting the number of cores in containers.

Actually the activator and agent's `CPUPlugin` uses `libnuma.get_available_cores()`.
As `psutil.cpu_count()` provides `logical` option, I think we could just use it.

We also need to explicitly distinguish the number of cores reported for resource slots (including logical cores and multiplied by the overcommit factor) and license server (only physical cores) in the activator's `CPUPlugin`.

achimnol:
I think this should be reworked in a separate issue!

Issue #469, opened 2022-06-14
A feature to query the actual resource allocation time for compute sessions


adrysn:
Compute sessions have three status-related time fields: `created_at`, `status_changed`, and `terminated_at`. With these fields, we can query the total session lifetime, the time between `PENDING` and `TERMINATED` statuses, but unable to get the actual resource allocation time (`PREPARING` ~ `TERMINATED`). If a session stays long in `PENDING` status, the actual resource allocation time can greatly differ compared to the result of `terminated_at - created_at`.

Since the actual resource allocation time is useful to calculate the billing metric in cloud and on-premise sites, it would be good to support this feature.

rapsealk:
Work in progress @ branch feature/469-actual-resource-allocation-time.
I temporarily added a command `backend.ai session resource-allocation-time` and it gives back an elapsed time in status transition from `PREPARING` to `TERMINATED`.

<img width="665" alt="스크린샷 2022-06-20 오후 3 43 35" src="https://user-images.githubusercontent.com/14137676/174541776-b82c6fc2-8f57-40aa-99c7-9c19296f55e7.png">

adrysn:
GUI counterpart should also be implemented.

PR #480, opened 2022-06-17
feat: Add `kernels.status_history` JSONB column for tracking status transition timestamp

rapsealk:
This PR resolves lablup/backend.ai#412.

It this PR, `status_history` column, which stores JSONB data, is added to `kernels` table. When manager notifies status change on kernels/sessions and tries to update their `status`, they also try to update `status_history` from now on, so that they can track the timestamp of status transition.

PR #483, opened 2022-06-20
feat: Add `GET_ACTUAL_RESOURCE_ALLOC_TIME` API

rapsealk:
This PR resolves lablup/backend.ai#469.

**WARNING: This PR is based on previous PR lablup/backend.ai#480, which has to be merged before this.**

In this PR, both an endpoint and an api function are added to compute actual resource allocation time of session. Suppose we already have timestamps for each session status, thanks to lablup/backend.ai#480, we can easily get an elapsed time from `PREPARING` to `TERMINATED`.

스크린샷 2022-06-20 오후 3 43 35

### DIFF
--- a/src/ai/backend/client/cli/session.py
+++ b/src/ai/backend/client/cli/session.py
@@ -662,6 +662,26 @@ def logs(session_id):
             sys.exit(1)
 
 
+@session.command('resource-allocation-time')
+@click.argument('session_id', metavar='SESSID')
+def resource_allocation_time(session_id):
+    '''
+    Shows the resource allocation time (PREPARING ~ TERMINATED) of a compute session.
+
+    \b
+    SESSID: Session ID or its alias given when creating the session.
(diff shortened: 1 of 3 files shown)


rapsealk:
Update: Manager only provides `status_history` value and client has a responsibility for computing actual resource allocation time.

src/ai/backend/client/cli/session.py
```diff
@@ -10,9 +10,13 @@
 import uuid
 
 import click
+from dateutil.parser import isoparse
+from dateutil.tz import tzutc
 from humanize import naturalsize
 from tabulate import tabulate
 
+from ai.backend.manager.models import KernelStatus
```
achimnol: The client SDK should not import the manager source!!

achimnol (changes requested):

achimnol (approved):

PR #484, opened 2022-06-20
feat: selective payload decryption logic on client-webserver

inureyes:
Add selective payload decryption logic between client and webserver

 In the http protocol environment, packets can be exposed as is. Although not normally required, adding the ability to encode and transmit selected information will help enhance security.

 1. Encrypted payload is sent with 'X-BackendAI-Encoded': 'true' header.
 2. Payload uses AES256-CBC as cipher algorithm with IV and signature.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -5,7 +5,7 @@ python_requirements(
         "aiodataloader-ng": ["aiodataloader"],
         "attrs": ["attr", "attrs"],
         "aiohttp-session": ["aiohttp_session"],
-        "Crypto": ["pycryptodome"],
+        "pycryptodome": ["Crypto"],
         "python-dateutil": ["dateutil"],
         "python-json-logger": ["pythonjsonlogger"],
         "python-snappy": ["snappy"],
--- a/BUILD
+++ b/BUILD
@@ -5,6 +5,7 @@ python_requirements(
         "aiodataloader-ng": ["aiodataloader"],
         "attrs": ["attr", "attrs"],
         "aiohttp-session": ["aiohttp_session"],
+        "Crypto": ["pycryptodome"],
         "python-dateutil": ["dateutil"],
         "python-json-logger": ["pythonjsonlogger"],
         "python-snappy": ["snappy"],
--- a/src/ai/backend/web/proxy.py
+++ b/src/ai/backend/web/proxy.py
@@ -23,7 +23,7 @@ from ai.backend.client.request import Request
 from .auth import get_api_session, get_anonymous_session
 from .logging import BraceStyleAdapter
 
-log = BraceStyleAdapter(logging.getLogger('ai.backend.console.proxy'))
+log = BraceStyleAdapter(logging.getLogger('ai.backend.web.proxy'))
 
 HTTP_HEADERS_TO_FORWARD = [
     'Accept-Language',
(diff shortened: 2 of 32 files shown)


src/ai/backend/web/server.py
```diff
@@ -291,7 +291,12 @@ async def login_handler(request: web.Request) -> web.Response:
             'type': 'https://api.backend.ai/probs/generic-bad-request',
             'title': 'You have already logged in.',
         }), content_type='application/problem+json')
-    creds = await request.json()
+    secure_context = request.headers.get('X-BackendAI-Encoded', None)
```
fregataa: I check the value is string 'true'.
maybe it can be improved by a validator such as trafaret or something similar.

fregataa (approved):
the payload of login process is successfully encoded and decoded.

src/ai/backend/web/server.py
inureyes: Good point!

BUILD
```diff
@@ -5,6 +5,7 @@ python_requirements(
         "aiodataloader-ng": ["aiodataloader"],
         "attrs": ["attr", "attrs"],
         "aiohttp-session": ["aiohttp_session"],
+        "Crypto": ["pycryptodome"],
```
achimnol: This should be reversed.

src/ai/backend/web/proxy.py
```diff
@@ -20,7 +23,7 @@
 from .auth import get_api_session, get_anonymous_session
 from .logging import BraceStyleAdapter
 
-log = BraceStyleAdapter(logging.getLogger('ai.backend.console.proxy'))
+log = BraceStyleAdapter(logging.getLogger('ai.backend.web.proxy'))
```
achimnol: Let's change this to `__name__`.

src/ai/backend/web/proxy.py
```diff
@@ -115,6 +118,24 @@ async def close_upstream(self) -> None:
...
+async def decode_payload(request):
+    config = request.app['config']
+    scheme = config['service'].get('force-endpoint-protocol')
+    if not request.content:
+        return request
+    if scheme is None:
+        scheme = request.scheme
+    api_endpoint = f'{scheme}://{request.host}'
+    payload = await request.text()
+    iv, real_payload = payload.split(':')  # Extract IV and real_payload from payload
```
```suggestion
    iv, real_payload = payload.split(':')  # Extract initial vector and actual payload
```
To give the code readers an opportunity to learn the acronym automatically.

tools/flake8.lock
```diff
@@ -180,18 +180,18 @@
...
           "requires_python": ">=3.7",
-          "version": "62.3.2"
+          "version": "62.5"
         }
       ],
       "platform_tag": [
         "cp310",
         "cp310",
-        "manylinux_2_31_aarch64"
+        "macosx_12_0_x86_64"
```
achimnol, line 189: FYI: No need to fix but this platform-tag alternating depending on where the lock file is generated is a currently known issue in pex and pants (https://pantsbuild.slack.com/archives/C046T6T9U/p1654207727719299, https://github.com/pantsbuild/pex/issues/1685)

src/ai/backend/web/proxy.py
```diff
@@ -127,6 +148,12 @@ async def web_handler(request, *, is_anonymous=False) -> web.StreamResponse:
             # but need to keep the client's version header so that
             # the final clients may perform its own API versioning support.
             request_api_version = request.headers.get('X-BackendAI-Version', None)
+            secure_context = request.headers.get('X-BackendAI-Encoded', None)
```
achimnol: I think the naming "Encoded" is misleading because encryption != encoding. How about "X-BackendAI-Encrypted"?

src/ai/backend/web/proxy.py
```diff
@@ -115,6 +118,24 @@ async def close_upstream(self) -> None:
             await self.up_conn.close()
 
 
+async def decode_payload(request):
```
achimnol: Here, it should be also `decrypt_payload()`.

achimnol (changes requested):

achimnol (approved):

achimnol:
We need a follow-up PR to implement the same feature on the client SDK.

achimnol (approved):

PR #487, opened 2022-06-22
fix: Casting UUID like str to UUID

lizable:
## Description 
This PR resolves AttributeError triggered by unmatching type between UUID-like string and UUID object itself during updating the quota limit. Simply casting string value to UUID would workaround this issue.
It's been confirmed while reviewing #443, so It should be merged with it, too.
For more information, please check the [UUID module definition in official python documentation (version 3.10.5)](https://docs.python.org/3/library/uuid.html#uuid.UUID.hex). 

## Screenshot
Screen Shot 2022-06-22 at 7 14 08 PM


achimnol (changes requested):
I think this sould be handled using Trafaret, e.g., `ai.backend.common.validators.UUID`.

achimnol (approved):

PR #488, opened 2022-06-23
fix: Type correction on renaming file api

lizable:
## Description
This PR resolves two things; setting the default value for `is_dir` parameter in the rename_file function, and providing more precise message in exception handling when virtual folder operation errors occurred on the same function.
I confirmed this issue during E2E testing for #443.

## Screenshot(s)
Screen Shot 2022-06-23 at 12 16 18 PM
Screen Shot 2022-06-23 at 12 17 53 PM

### DIFF
--- a/src/ai/backend/manager/models/storage.py
+++ b/src/ai/backend/manager/models/storage.py
@@ -31,7 +31,7 @@ import yarl
 from .base import (
     Item, PaginatedList,
 )
-from ..api.exceptions import VFolderOperationFailed
+from ..api.exceptions import InvalidAPIParameters, VFolderOperationFailed
 from ..exceptions import InvalidArgument
 if TYPE_CHECKING:
     from .gql import GraphQueryContext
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/models/storage.py
```diff
@@ -165,15 +164,18 @@ async def request(
                 try:
                     error_data = await client_resp.json()
                     raise VFolderOperationFailed(
-                        extra_msg=error_data.pop("msg"),
```
achimnol, line 168: 👍🏼 Good catch!

src/ai/backend/storage/api/manager.py
```diff
@@ -485,7 +485,7 @@ async def rename_file(request: web.Request) -> web.Response:
                 t.Key("vfid"): tx.UUID(),
                 t.Key("relpath"): tx.PurePath(relative_only=True),
                 t.Key("new_name"): t.String(),
-                t.Key("is_dir"): t.ToBool(),  # ignored since 22.03
```
achimnol, line 488: You don't have to remove parenthesis of `t.ToBool()` because Trafaret automatically handles the difference of type reference and instance reference when there is no argument of a trafaret class.

PR #491, opened 2022-06-23
feat: NUMA-aware allocation maps

achimnol:
Add support for NUMA-aware affinitization and ordering of devices when assigning devices in AllocMap classes.

This will resolve lablup/backend.ai-internal#35.

## New configuration options

* `agent.toml`
  - `[resource]`
    - `allocation-order`: `["cuda", "rocm", "tpu", "cpu", "mem"]` (default)
      - The name is taken from `slot_name.partition["."][0]`
        (e.g., `cuda.device` -> `cuda`, `cuda.shares` -> `cuda`)
    - `affinity-policy`: `"INTERLEAVED"` (default) or `"PREFER_SINGLE_NODE"`

## Implementation detail

First, the agent builds a matrix (`AffinityMap`) which represents the distances between all possible pairs of compute devices upon startup.

When the agent receives kernel creation requests, previously it just sorted the devices by the remaining resource slots.  Now it calculates the connected components of devices with zero distances from the affinity map and try the largest component groups first while sorting the devices inside each group by the remaining resource slots.  Then it iterates over all device types as specified in `allocation_order`. (Previously, there was no explicit configuration for the allocation order!)

During iteration over device types, now an `AffinityHint` object is passed through the `AllocMap.allocate()` methods, which carries the information about which devices were allocated in the prior device type, the affinity map, and the policy.  Using this, subsequent device type's `alloc_map` can utilize the information of the previous `alloc_map`'s allocation result.

Depending on the affinity policy, when it allocates the first device type (`device_name` in the code; "cuda" by default), it either uses `itertools.chain()` for `PREFER_SINGLE_NODE` or `more_itertools.interleave_longest()` for `INTERLEAVED`. Even with `PREFER_SINGLE_NODE`, if there is not enough room to allocate in the devices from the first NUMA node, it will allocate devices from the other NUMA nodes as well ("spill over").

From the second and other subsequent device types, it always applies the interleaving policy because there may be a mixture of devices from multiple different NUMA nodes and there should be at least one or two CPU cores for accelerators in the same NUMA node as required by Weka.io or other NUMA-aware resource provisioners.


kyujin-cho (approved):
LGTM.

PR #492, opened 2022-06-23
fix: Raise explicit error if multiple entries found when deleting a vfolder

lizable:
This PR resolves to prevent manager from deleting unexpected vfolder on deletion requests by raising error when there's the same name of vfolder in multiple accessible storage hosts.

## Screenshots
Screen Shot 2022-06-23 at 11 54 01 PM
Screen Shot 2022-06-24 at 12 46 54 AM

### DIFF
--- a/changes/492.fix.md
+++ b/changes/492.fix.md
@@ -1,1 +1,1 @@
-Do not delete a virtual folder if there are other folders with the same name (in other folder hosts), rather than blindly and dangerously deleting the first-queried one.
+Do not delete a virtual folder if there are other folders with the same name (in other folder hosts) and handle by new relevant exception, `TooManyVFoldersFound`, rather than blindly and dangerously deleting the first-queried one.
--- a/changes/492.fix.md
+++ b/changes/492.fix.md
@@ -1,0 +1,1 @@
+Raise an error on deleting vfolder by its name if there is the same name of vfolder in multiple accessible storage hosts.
--- a/changes/492.fix.md
+++ b/changes/492.fix.md
@@ -1,1 +1,1 @@
-Raise an error on deleting vfolder by its name if there is the same name of vfolder in multiple accessible storage hosts.
+Do not delete a virtual folder if there are other folders with the same name (in other folder hosts), rather than blindly and dangerously deleting the first-queried one.
--- a/src/ai/backend/manager/api/exceptions.py
+++ b/src/ai/backend/manager/api/exceptions.py
@@ -256,9 +256,9 @@ class VFolderCreationFailed(BackendError, web.HTTPBadRequest):
     error_title = 'Virtual folder creation has failed.'
 
 
-class VFolderDeletionFailed(BackendError, web.HTTPBadRequest):
-    error_type = 'https://api.backend.ai/probs/vfolder-deletion-failed'
-    error_title = 'Virtual folder deletion has failed.'
+class TooManyVFoldersFound(BackendError, web.HTTPNotFound):
+    error_type = 'https://api.backend.ai/probs/too-many-vfolders'
+    error_title = 'Thare are two or more matching vfolders.'
 
 
 class VFolderNotFound(ObjectNotFound):
--- a/src/ai/backend/manager/api/exceptions.py
+++ b/src/ai/backend/manager/api/exceptions.py
@@ -257,7 +257,7 @@ class VFolderCreationFailed(BackendError, web.HTTPBadRequest):
 
 
 class VFolderDeletionFailed(BackendError, web.HTTPBadRequest):
-    error_type = "https://api.backend.ai/probs/vfolder-deletion-failed"
+    error_type = 'https://api.backend.ai/probs/vfolder-deletion-failed'
     error_title = 'Virtual folder deletion has failed.'
 
 
--- a/src/ai/backend/manager/api/exceptions.py
+++ b/src/ai/backend/manager/api/exceptions.py
@@ -258,7 +258,7 @@ class VFolderCreationFailed(BackendError, web.HTTPBadRequest):
 
 class TooManyVFoldersFound(BackendError, web.HTTPNotFound):
     error_type = 'https://api.backend.ai/probs/too-many-vfolders'
-    error_title = 'Thare are two or more matching vfolders.'
+    error_title = 'There are two or more matching vfolders.'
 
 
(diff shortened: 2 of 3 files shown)


changes/492.fix.md
```diff
@@ -0,0 +1 @@
+Raise an error on deleting vfolder by its name if there is the same name of vfolder in multiple accessible storage hosts.
```
adrysn: Suggestion:
> Do not delete a virtual folder if there are other folders with the same name (in other folder hosts), rather than blindly and dangerously deleting the first-queried one.

src/ai/backend/manager/api/exceptions.py
```diff
@@ -256,6 +256,11 @@ class VFolderCreationFailed(BackendError, web.HTTPBadRequest):
     error_title = 'Virtual folder creation has failed.'
 
 
+class VFolderDeletionFailed(BackendError, web.HTTPBadRequest):
+    error_type = "https://api.backend.ai/probs/vfolder-deletion-failed"
```
adrysn: Let's be consistent with the single or double quotes.

src/ai/backend/manager/api/vfolder.py
```diff
@@ -1608,36 +1608,57 @@ async def delete(request: web.Request) -> web.Response:
...
+        entry = entries[0]
+        if len(entries) > 1:
+            log.error('VFOLDER.DELETE(folder name:{}, hosts:{}', folder_name, [entry['host'] for entry in entries])
+            raise VFolderDeletionFailed(
+                extra_msg="Found same name of vfolder to delete on multiple storage hosts.",
+                extra_data=None,
+            )
+        elif len(entries) == 0:
             raise InvalidAPIParameters('No such vfolder.')
+        elif entry['name'] == folder_name:
```
adrysn: Since the folders are filtered with the `folder_name` already, this condition is not necessary.

src/ai/backend/manager/api/vfolder.py
```diff
@@ -1608,36 +1608,57 @@ async def delete(request: web.Request) -> web.Response:
...
+        #                 'Cannot delete the vfolder '
+        #                 'that is not owned by myself.')
+        #         break
+        # FIXME: For now, multiple entries on delete vfolder will raise an error. Will be fixed in 22.06
+        # query_accesible_vfolders returns list
+        entry = entries[0]
+        if len(entries) > 1:
+            log.error('VFOLDER.DELETE(folder name:{}, hosts:{}', folder_name, [entry['host'] for entry in entries])
+            raise VFolderDeletionFailed(
+                extra_msg="Found same name of vfolder to delete on multiple storage hosts.",
```
adrysn: My suggestion:
> Multiple folders with the same name.

src/ai/backend/manager/api/vfolder.py
```diff
@@ -1608,36 +1608,57 @@ async def delete(request: web.Request) -> web.Response:
...
+        proxy_name, volume_name = root_ctx.storage_manager.split_host(folder_host)
+        async with root_ctx.storage_manager.request(
+            proxy_name, 'POST', 'folder/delete',
+            json={
+                'volume': volume_name,
+                'vfid': str(folder_id),
+            },
+        ):
+            pass
+        return web.Response(status=204)
```
adrysn: Why do you indent this block of code? It seems like there is no operation which requires DB connection.

adrysn (changes requested):

src/ai/backend/manager/api/vfolder.py
lizable: My mistakes. I'll fix it right away.

src/ai/backend/manager/api/vfolder.py
lizable: I agree. I'll remove that condition.

src/ai/backend/manager/api/vfolder.py
```diff
@@ -1608,20 +1608,41 @@ async def delete(request: web.Request) -> web.Response:
...
+        if len(entries) > 1:
+            log.error('VFOLDER.DELETE(folder name:{}, hosts:{}', folder_name, [entry['host'] for entry in entries])
+            raise VFolderDeletionFailed(
+                extra_msg="Multiple folders with the same name.",
+                extra_data=None,
+            )
+        elif len(entries) == 0:
             raise InvalidAPIParameters('No such vfolder.')
+        else:
+            pass
```
adrysn: No op lines.

adrysn (changes requested):

adrysn (approved):
LGTM.

src/ai/backend/manager/api/vfolder.py
```diff
@@ -1608,20 +1608,38 @@ async def delete(request: web.Request) -> web.Response:
...
+        #             not entry['is_owner']
+        #             and entry['permission'] != VFolderPermission.RW_DELETE
+        #         ):
+        #             raise InvalidAPIParameters(
+        #                 'Cannot delete the vfolder '
+        #                 'that is not owned by myself.')
+        #         break
+        # FIXME: For now, multiple entries on delete vfolder will raise an error. Will be fixed in 22.06
+        # query_accesible_vfolders returns list
+        entry = entries[0]
```
achimnol: Let's move this line to 1635 to avoid reusing in the for-loop in 1628.

src/ai/backend/manager/api/exceptions.py
```diff
@@ -256,6 +256,11 @@ class VFolderCreationFailed(BackendError, web.HTTPBadRequest):
     error_title = 'Virtual folder creation has failed.'
 
 
+class VFolderDeletionFailed(BackendError, web.HTTPBadRequest):
```
achimnol: image

Let's subclass `web.HTTPNotFound` and rename to `TooManyVFoldersFound` to keep consistency.

achimnol (changes requested):

src/ai/backend/manager/api/exceptions.py
achimnol: The rationale behind subclassing "not found" for "too many" exceptions is that it should represent that the precondition for "only one match wanted" is not satisfied and this is not a kind of user's fault nor system error.

achimnol (approved):

PR #497, opened 2022-06-27
git ignore kernel scratches

fregataa:
let git ignore `/scratches` directory which is created for kernels.
git keeps tracking the directory like below

image


achimnol (approved):
Great!

Issue #148, opened 2018-04-13
User activity logs


achimnol:
Some customers want to keep detailed command logs executed by their users due to their security audit policies.Let's add an option to store all query-mode and batch-mode code snippets.

1st phase:

- [ ] Add an etcd configuration option
- [ ] Add the `activity_logs` database table
- [ ] Add an admin API to retrieve activity logs
  2nd phase:

- [ ] Extend jail to log system call arguments and return values\* This will incur non-negligible performance overheads. We should implement this very efficiently, and buffer & compress the logs on the fly as hundreds of syscalls may be issued per second.

JIRA Issue: BA-334

Yaminyam:
https://github.com/lablup/backend.ai/pull/1769
The created_user field has been deleted from that pr.
This field existed to record cases where the created_user and the owner user of the session are different, and it was decided to record created_user as an event in the audit_log.

PR #502, opened 2022-06-29
feature: Add audit logs

hydroxyde:
This pull request refers to: lablup/backend.ai-internal#17 and #148.

This is the implementation of the audit log feature's model which keeps records of creation, modification and deletion, etc., of users, keypairs and vfolders. 
### Audit Logs model
Each record in the audit log is as follow:

**The subject** (i.e., who did the action?)
It contains: the API access key, user UUID, user email (with respective types, all indexed)

**The action**

- Action type: CREATE | CHANGE | DELETE

- Action data: (JSONB field)

{"before": {"data1": "val1", ...}, "after": {"data1": "val2", ...}}
For passwords and secret keys, we do not record the original/hash values.
CREATE: record only "after", leaving "before" as null
CHANGE: record both "before", "after" and include only the changed data.
DELETE: record only "before", leaving "after" as null

**The target object**
It contains the values of users.uuid, keypairs.access_key or vfolder.name

**The date/time** of action (indexed, this should be the default descending order key of audit log queries)

> Screenshot 2022-05-04 at 17 29 36

### Periodic removal of audit logs
This command extends https://github.com/lablup/backend.ai-manager/pull/498: 
- Option to remove audit logs, time based
   - e.g., remove all logs older than 3 months, 1 year, etc...

Screenshot 2022-07-03 at 18 13 45

**Extended options are as follow:**

- **-t, --table [kernels|audit_logs]** : Which table to run the operation on. Options available are kernels  and audit_logs. If not set or set to kernels, it will run the operation on kernels table. If set to audit_logs, it will run on audit_logs table (default: kernels). 
- **--target** : For audit_logs you can chose which audit target you would like the operation to run on. You can provide user.uuid, keypairs.access_key or vfolder_name. If not set, it will run on all targets.

### List
This command list audit logs of a user by user_id

Screenshot 2022-05-04 at 18 30 52

Screenshot 2022-05-09 at 21 09 03

This PR uses revamp from https://github.com/lablup/backend.ai-manager/pull/449

- `filter` uses Python like query expression with SQL-specific operators
  - e.g., `email contains "@lablup.com"`

- Use of mini-language for expressing query ordering `order` as the list of multiple columns with ascending/descending order marks
  - e.g., `-created_date`


achimnol (changes requested):
For the record of face-to-face review:
- Let's use `vfolders.id` (UUID) column to refer vfolders as target objects.
- Let's add a string column to distinguish the target object type, because vfolder UUIDs and user UUIDs are indistinguishable.

I'll do a follow-up code-level review soon.

achimnol:
@hydroxyde Please resolve merge conflicts and re-run black (`./pants fmt --changed-since=main`) on your code.

achimnol:
Need to add an audit log when #522 is merged.

achimnol:
@hydroxyde Please update the PR if you have any updates on this!

PR #503, opened 2022-06-29
Feature/add pipeline server handler

lizable:
## Description
This PR provides a separate handler for pipeline server to sending up requests from WebUI.
Related PR: [backend.ai-webui#1221](https://github.com/lablup/backend.ai-webui/pull/1221)

### DIFF
--- a/src/ai/backend/web/proxy.py
+++ b/src/ai/backend/web/proxy.py
@@ -24,6 +24,7 @@ log = BraceStyleAdapter(logging.getLogger(__name__))
 
 HTTP_HEADERS_TO_FORWARD = [
     'Accept-Language',
+    'Authorization'
 ]
 
 extra_config_headers = t.Dict({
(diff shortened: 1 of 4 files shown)


lizable:
By suggestion of @achimnol, I would like to refactor pipeline specific web handler into web_handler since the code flow is quite similar.

src/ai/backend/web/server.py
```diff
@@ -590,6 +590,11 @@ async def server_main(
     cors.add(app.router.add_route('POST', '/func/{path:.*$}', web_handler))
     cors.add(app.router.add_route('PATCH', '/func/{path:.*$}', web_handler))
     cors.add(app.router.add_route('DELETE', '/func/{path:.*$}', web_handler))
+    cors.add(app.router.add_route('GET', '/{name:flow}/{path:.*$}', web_handler))
+    cors.add(app.router.add_route('PUT', '/{name:flow}/{path:.*$}', web_handler))
+    cors.add(app.router.add_route('POST', '/{name:flow}/{path:.*$}', web_handler))
+    cors.add(app.router.add_route('PATCH', '/{name:flow}/{path:.*$}', web_handler))
+    cors.add(app.router.add_route('DELETE', '/{name:flow}/{path:.*$}', web_handler))
```
achimnol: I think we need a precaution step to prevent overlapping of the flow name and `"func"`.  The best way to do is to prefix the URL routes, e.g., `'/pipeline/{name:flow}/{path:.*$}'`.

achimnol (changes requested):

src/ai/backend/web/proxy.py
```diff
@@ -141,13 +141,13 @@ async def decrypt_payload(request):
 
 async def web_handler(request, *, is_anonymous=False) -> web.StreamResponse:
     path = request.match_info.get('path', '')
-    name = request.match_info.get('name', '')
+    first_path = request.path.strip(r'\/').split('/')[0]  # extract the first path
```
```suggestion
    first_path = request.path.lstrip('/').partition('/')[0]  # extract the first path
```
- Simplify the string pattern and use `lstrip()` only
- `str.partition()` works even when the trailing part does not contain any slash.

achimnol (changes requested):

src/ai/backend/web/proxy.py
lizable: it's more safer than the previous change. I'll apply it right away!

achimnol (approved):

PR #468, opened 2022-06-13
feat: migrate to redis-py and add support for redlock

kyujin-cho:
This PR replaces underlying redis API library from aioredis to redis-py, as redis-py now supports asyncio OOTB.    
This PR also includes support for redis' lock mechanism, `redlock`. Users can try new Redis Lock by setting manager.toml's`manager.distributed-lock` config to `redlock`.

PR #512, opened 2022-07-04
fix: Use `aioredis.client.Redis.ping()` to ping redis server

rapsealk:
When I was working on `ping_redis_connection()` for the first time(https://github.com/lablup/backend.ai-common/pull/139), I used [`aioredis.client.Redis.time()`](https://aioredis.readthedocs.io/en/latest/api/high-level/#aioredis.client.Redis.time) since it seemed to be a reasonable alternative one.
Now I found [`aioredis.client.Redis.ping()`](https://aioredis.readthedocs.io/en/latest/api/high-level/#aioredis.client.Redis.ping) from the official document and I suggest to use it rather than the old one.


achimnol:
NOTE: After merging #468, this should be rewritten.

achimnol (approved):

PR #514, opened 2022-07-05
fix: add timeout to `prepare()`'s `PersistentTaskGroup`

kyujin-cho:
This PR resolves `prepare()` hanging infinitely when individual session's `start_session()` task inside `TaskGroup` hangs.    
This PR also changes `TaskGroup` to `PersistentTaskGroup`, as every `start_session()` call needs execution result.

### DIFF
--- a/changes/514.fix
+++ b/changes/514.fix
@@ -1,2 +1,1 @@
 Fix `prepare()` not running when `start_session()` call hangs without raising Exception
-
--- a/changes/514.fix
+++ b/changes/514.fix
@@ -1,0 +1,2 @@
+Fix `prepare()` not running when `start_session()` call hangs without raising Exception
+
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -800,6 +804,8 @@ async def _mark_session_preparing() -> Sequence[PendingSession]:
                          "maybe another prepare() call is still running")
                 raise asyncio.CancelledError()
             raise
+        except asyncio.TimeoutError:
+            log.warn('timeout while executing start_session()')
```
achimnol: Let's attach which scaling group is observing the session start timeout.

changes/514.fix
```diff
@@ -0,0 +1,2 @@
+Fix `prepare()` not running when `start_session()` call hangs without raising Exception
+
```
achimnol: Please remove this extra trailing newline to prevent line-separation of PR number reference during auto-generation of changelogs at release.

achimnol (changes requested):
Minor fixes please!

src/ai/backend/manager/scheduler/dispatcher.py
achimnol: This will be removed in #515, so let's skip over for now.

achimnol (approved):

achimnol:
This will not be backported to 21.09 as it is no longer shipped to customer sites.

PR #516, opened 2022-07-05
feat: update codespaces bootstrap script

### LINKED ISSUE #499: Revisit codespace auto-setup
We have set up a codespace auto configuration at #280.
Now we have migrated to mono-style repository at #417, it should be updated.

kyujin-cho:
This PR updates codespace bootstrap script to reflect updated monorepo setup.    
This resolves #499.


Yaminyam (approved):
I checked the installation and server normal operation in the codespace environment.

Yaminyam (commented):
image
An error occurs while performing postCreateCommand during setup

achimnol (changes requested):
Let's split install-dev.sh into two parts that can be either executed at once (with no option) or separately (with `--codespace-on-create`, `--codespace-post-create` so that we can use prebuilt containers.

- In "on create", run `docker pull compose` and `./pants export ::` to pre-download and populate external dependencies that takes long time to fetch and install.
- In "post create", run any configuration and cloning of additional repositories because in this stage we have fully configured GitHub access tokens.
- If run inside Codespace (we could check this by environment variables), we may skip several prerequisite checks (such as availability of Docker).
- Let's explicitly add a handful range (about 10?) of forwarded TCP ports for wsproxy and explicitly configure the wsproxy worker port range as well (for v1).

kyujin-cho:
> * Let's explicitly add a handful range (about 10?) of forwarded TCP ports for wsproxy and explicitly configure the wsproxy worker port range as well (for v1).

This can't be achieved within current webui's wsproxy implementation:    
- wsproxy randomly selects listening ports (lablup/backend.ai-webui#1389)
- codespace distinguishes traffic to designated port from subdomain, not by port

kyujin-cho:
Relevant document updates will be followed after this PR is merged, on a separate PR.

scripts/install-dev.sh
```diff
@@ -540,37 +545,44 @@ echo "${LGREEN}Backend.AI one-line installer for developers${NC}"
...
-  show_error "When run as the root, 'docker compose' seems not to be a compatible version (v2)"
-  show_info "Please check the following link: https://docs.docker.com/compose/install/compose-plugin/#install-the-plugin-manually to install Docker Compose CLI plugin on /usr/local/lib/docker/cli-plugins"
-  exit 1
-fi
-if [ "$DISTRO" = "Darwin" ]; then
-  echo "validating Docker Desktop mount permissions..."
-  docker pull alpine:3.8 > /dev/null
-  docker run --rm -v "$HOME/.pyenv:/root/vol" alpine:3.8 ls /root/vol > /dev/null 2>&1
+if [ $CODESPACES = "true" ]; then
+  $bpython scripts/check-docker.py
```
achimnol, line 546: Shouldn't we check this when we are *not* on codespace?

.devcontainer/devcontainer.json
```diff
@@ -1,7 +1,40 @@
...
 // https://github.com/microsoft/vscode-dev-containers/tree/v0.187.0/containers/ubuntu
 {
-	"name": "Ubuntu",
-	"postCreateCommand": "bash /workspaces/backend.ai/.devcontainer/bootstrap.sh",
-	"forwardPorts": [6022, 8090, 8091]
-}
\ No newline at end of file
+  "name": "Ubuntu",
+  "onCreateCommand": "cd /workspaces/backend.ai && bash /workspaces/backend.ai/scripts/install-dev.sh --codespaces-on-create >> install-oncreate.log 2>&1",
+  "postCreateCommand": "cd /workspaces/backend.ai && bash /workspaces/backend.ai/scripts/install-dev.sh --codespaces-post-create >> install-postcreate.log 2>&1",
```
achimnol: I think it is a bad practice to hide the log. Do we have another way to see the output?

achimnol: Probably we need to skip the completion guide message depending on whether we're on codespace or not, but I think we don't have to omit it.

achimnol (changes requested):

scripts/install-dev.sh
kyujin-cho: Gosh, a huge mistake 🤦‍♂️

.devcontainer/devcontainer.json
kyujin-cho: Initial design of this command was to retain installation log to a file (`/workspaces/backend.ai/install.log`). Will it be better to just print installation outputs on stdout? In that way, users can see current status through the option under Command Palette. I'll follow your decision.

achimnol: I think it's better to utilize existing & officially provided interfaces to inspect the logs whenever possible.

kyujin-cho: Got it. I'll update the configuration.

achimnol:
image
Please take a look at this.

achimnol:
Also let's add the ssh feature config: https://github.com/cli/cli/pull/6019/files

achimnol (approved):

PR #521, opened 2022-07-07
feat: Rename `shared_by` to `owner` and add a `type` field to the `list_shared_vfolders` API

Sujin-Kim1:
- Since the creator of the vfolder shares it with others, `shared_by` should be set to the creator, not the person who requested the query.
- Set default value if `vfolder_id` to get the list of shared folders of all users.

- Related PR: https://github.com/lablup/backend.ai-control-panel/pull/287


adrysn (approved):
LGTM.

achimnol (approved):

achimnol (changes requested):
When `vfolder_id is None`, shouldn't we allow the operation only for superadmins?  Also does it work as the join condition?

achimnol (approved):
When the superadmin performs sharing of a user's vfolder to another user, this PR will show the "shared_by" field as the user, instead of the superadmin.  I think that this is what we intend, because the fact that the superadmin did sharing will be left in the audit log while users would observe it as a pre-configuration.

So I think it is okay to keep "shared_by" field to be the creator of the vfolder.

achimnol (changes requested):
Let's put projetc ID to `shared_by` if the vfolder is a project vfolder.

Sujin-Kim1:
> Let's put projetc ID to `shared_by` if the vfolder is a project vfolder.

I just applied it 😉.

achimnol (changes requested):
I think... we would also need to specify the type of the subject of sharing (i.e., is it a group or a user?) because we cannot distinguish it by just looking at the `shared_by` field.

How about just rename the field to `owner` and add a `type` field (to indicate whether it is a user vfolder or a project vfolder)?

achimnol (commented):
The change itself looks sane, but I'm worride about the backward-compatibility with prior versions of clients & webui.
Is it okay?

Sujin-Kim1:
> The change itself looks sane, but I'm worride about the backward-compatibility with prior versions of clients & webui. Is it okay?

Yes. Even though webui uses this API, it doesn't use the `shared_by` option. And clients do not use this API.

achimnol (approved):

PR #522, opened 2022-07-07
feat: Allow superadmins to remove users from shared vfolders

Sujin-Kim1:
The existing `leave` API allows the requesting users to leave the invited vfolders.
This PR allows superadmins to call this API on behalf of a specific user, effectively "removing" the user from the vfolder permission list as the superadmin want.

- Related PR: lablup/backend.ai-control-panel#287


src/ai/backend/manager/api/vfolder.py
```diff
@@ -2387,6 +2422,7 @@ def create_app(default_cors_options):
     cors.add(add_route('GET',    r'/{name}/files', list_files))
     cors.add(add_route('POST',   r'/{name}/invite', invite))
     cors.add(add_route('POST',   r'/{name}/leave', leave))
+    cors.add(add_route('POST',   r'/{name}/force-leave', force_leave))
```
adrysn: The `force_leave` handler is almost the same as the existing `leave` handler. How about taking out the common logic to a separate method?

adrysn (changes requested):

src/ai/backend/manager/api/vfolder.py
agatha197: I reflected it, so please review it when you're okay :)

src/ai/backend/manager/api/vfolder.py
```diff
@@ -1805,6 +1805,30 @@ async def delete(request: web.Request) -> web.Response:
     return web.Response(status=204)
 
 
+async def leave_request(request: web.Request, params: Any) -> web.Response:
```
achimnol: Why is this function named "request"? It just deletes the permission entry "blindly". If this is an internal database model manipulation, this should not take `web.Request` for the principle of separation of concerns, and should moved to `ai.backend.manager.models.vfolder`.

src/ai/backend/manager/api/vfolder.py
```diff
@@ -1819,23 +1843,55 @@ async def leave(request: web.Request, row: VFolderRow) -> web.Response:
...
+@server_status_required(ALL_ALLOWED)
+@vfolder_permission_required(VFolderPermission.READ_ONLY)
+@check_api_params(
+    t.Dict(
+        {
+            tx.AliasedKey(["shared_user_id", "sharedUserId"]): t.String,
+        }
+    ),
+)
+async def force_leave(request: web.Request, params: Any, row: VFolderRow) -> web.Response:
```
achimnol: Could we make "force" as an option to the existing API instead of adding a new API?

achimnol (changes requested):

src/ai/backend/manager/api/vfolder.py
```diff
@@ -1805,93 +1805,55 @@ async def delete(request: web.Request) -> web.Response:
...
     perm = row["permission"]
-    rqst_params = {
-        "vfolder_id": row["id"],
-        "user_uuid": params["shared_user_id"],
-    }
+
+    if shared_user_uuid:
+        # Allow only superadmin to leave the shared vfolder of others.
+        if (rqst_user_uuid != shared_user_uuid) and (user_role not in (UserRole.SUPERADMIN)):
+            raise GenericForbidden("Insufficient permission")
```
achimnol: How about just using `manager.api.exceptions.InsufficientPrivilege`? 😉

src/ai/backend/manager/api/vfolder.py
```diff
@@ -1805,93 +1805,55 @@ async def delete(request: web.Request) -> web.Response:
...
-
-
-@superadmin_required
-@server_status_required(ALL_ALLOWED)
-@vfolder_permission_required(VFolderPermission.READ_ONLY)
 @check_api_params(
     t.Dict(
         {
-            tx.AliasedKey(["shared_user_id", "sharedUserId"]): t.String,
+            t.Key("shared_user_uuid", default=None): t.String | t.Null,
```
achimnol: Is it safe to remove the alias?
We've left it for the webui's camel casing.
As a frontend developer, if you feel it's safe, it's fine to remove it though.

src/ai/backend/manager/api/vfolder.py
```diff
@@ -1808,22 +1809,40 @@ async def delete(request: web.Request) -> web.Response:
...
+    user_role = request["user"]["role"]
+    rqst_user_uuid = request["user"]["uuid"]
+    shared_user_uuid = params["shared_user_uuid"]
     vfolder_id = row["id"]
-    perm = None
+    perm = row["permission"]
+
+    if shared_user_uuid:
+        # Allow only superadmin to leave the shared vfolder of others.
+        if (rqst_user_uuid != shared_user_uuid) and (user_role not in (UserRole.SUPERADMIN)):
```
kyujin-cho: Uh... Is there a specific reason using `in` operator instead of `==`? Of course this will work, but still not a perfect way to do so. If you were intending to use `in` to check if `user_role` exists inside tuple containing single `UserRole.SUPERADMIN` element, then please add comma at the end of tuple element. The braces on this line will work as a brace to wrap code, not for defining tuple. 


```suggestion
        if (rqst_user_uuid != shared_user_uuid) and (user_role not in (UserRole.SUPERADMIN,)):
```

kyujin-cho (changes requested):

src/ai/backend/manager/api/vfolder.py
agatha197: It's superadmin only. So I changed to use `==` as you mentioned. Thanks!

kyujin-cho (approved):
LGTM.

PR #528, opened 2022-07-08
migrate keypair resource policy api to sqlalchemy orm partially

fregataa:
The first step of migration of SQLAlchemy Core API to ORM API.
add Declarative `KeyPairResourcePolicyRow` table which is provided by `keypair_resource_policy` imperative table.
https://docs.sqlalchemy.org/en/14/orm/declarative_tables.html#declarative-with-imperative-table-a-k-a-hybrid-declarative


achimnol (changes requested):
I think you just updated the selected column list to be taken from ORM, instead of rewriting SQLAlchemy Core APIs with ORM-based queries.
Could we go just one step further?

Note: Since this PR is a small demonstration of the ORM migration changes, let's keep prior table-based queries when the table is accessed from or joined with other querie.

fregataa:
> I think you just updated the selected column list to be taken from ORM, instead of rewriting SQLAlchemy Core APIs with ORM-based queries. Could we go just one step further?
> 
> Note: Since this PR is a small demonstration of the ORM migration changes, let's keep prior table-based queries when the table is accessed from or joined with other querie.

I have a question about *keeping prior table-based queries* and * rewriting SQLAlchemy Core APIs with ORM-based queries*. I thought that rewriting Core APIs with ORM-based queries should be like [this](https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#orm-query-unified-with-core-select), therefore there would be no huge change of queries if there is no optimization using `sqlalchemy.Session` or  ORM-based relationships.
e.g)
```
# before
query = sa.select([users]).select_from(users).where(users.c.name == username)

# after
query = sa.select(Users).select_from(Users).where(Users.name == username)
```
am I misunderstanding between *migrating to 2.0* and *rewriting with ORM-based queries*?

Issue #489, opened 2022-06-23
Relocate `last_registry.dat` to a persistent directory


achimnol:
Currently the agent keeps the memory of containers using a pickled database stored under `/tmp`.
Let's relocate this database to a more persistent location such as `/var/lib`.

This will become more important for #416 and handle node failures in customer sites.

The implementation should have a migration from the old location to the new location seamlessly.

PR #529, opened 2022-07-08
feat(agent): Introduce `var-base-path` config and relocate last registry db

KangJuSeong:
resolve issue #489


tests/agent/conftest.py
```diff
@@ -24,6 +24,8 @@ def test_id():
 def local_config(test_id, etcd_container, redis_container):  # noqa: F811
     ipc_base_path = Path.cwd() / f'.tmp/{test_id}/agent-ipc'
     ipc_base_path.mkdir(parents=True, exist_ok=True)
+    var_base_path = Path.cwd() / 'var/lib/backend.ai'
```
achimnol: For testing, we need to use an isolated configuration other than the current development sources, like the above `ipc_base_path`'s using the `test_id` fixture.

```suggestion
    var_base_path = Path.cwd() / '.tmp/{test_id}/agent-var'
```

achimnol (changes requested):
Good job!
Check out a minor comment please.

PR #530, opened 2022-07-08
fix: Background async tasks disappearing due to unhandled exceptions

kyujin-cho:
This PR fixes some asyncio tasks disappearing without error due to unhandled exceptions inside task implementation.


src/ai/backend/manager/api/stream.py
```diff
@@ -626,36 +627,43 @@ async def stream_conn_tracker_gc(root_ctx: RootContext, app_ctx: PrivateContext)
...
-                    log.debug(f"conn_tracker: gc {session_id} "
-                              f"removed/remaining = {removed_count}/{remaining_count}")
-                    if prev_remaining_count > 0 and remaining_count == 0:
-                        await root_ctx.idle_checker_host.update_app_streaming_status(
-                            session_id,
-                            AppStreamingStatus.NO_ACTIVE_CONNECTIONS,
+            try:
+                no_packet_timeout: timedelta = tx.TimeDuration().check(
+                    await shared_config.etcd.get('config/idle/app-streaming-packet-timeout') or '5m',
+                )
```
achimnol, line 633: The outer try-catch block should embrace this part only if there no etcd access below to minimize the change.

achimnol (changes requested):

achimnol (approved):

Issue #415, opened 2022-04-19
Leader election-based distributed timer and image rescan rate limiting


achimnol:
Recently we have replaced our distributed locks and global timers to use etcd's concurrency API.

However, there are still edge cases that require a global coordination of all manager processes, such as rate-limited container registry access (e.g., Docker Hub with anonymous user).  Since there are many manager processes that receives the API requests in a load-balanced fashion, it is difficult to share the rate-limit states between different manager processes.  This is why lablup/backend.ai-manager#501 is on hold.

Let's localize such globally coordinated states to a single manager process, or a leader.
To keep high availability, we should perform periodic checks on the liveness the leader and re-elect it, and fortunately etcd provides the facilities to implement this.

* [ ] manager: Implement leader election of manager processes with periodic leader status checks.
* [ ] manager: Rewrite global timer to run on the leader manager process. (When a new leader is elected, the new one should start global timers while the old one should stop, of course when the old one is still alive.)
* [ ] manager: Rewrite lablup/backend.ai-manager#501 to use a local `aiolimiter` state to implement its own rate-limiting to the container registries.  Assuming that leader changes rarely happen, we can just keep the state per process without explicit migration upon leader elections.

rapsealk:
It seems that [Etcd Election API](https://etcd.io/docs/v3.3/dev-guide/api_concurrency_reference_v3/#service-election-etcdserverapiv3electionv3electionpbv3electionproto) is only for their own cluster. Therefore, I wrote [raft-py](https://github.com/lablup/raft-py) which can run in standalone mode and can be attached to our `backend.ai` manager.

PR #532, opened 2022-07-10
feat: Implement `Raft` consensus algorithm for distributed managers

rapsealk:
This PR is related to lablup/backend.ai#415.


rapsealk:
Using `Raft` algorithm, a cluster cannot proceed anymore when the majority of managers are malfunctioning, because it lacks the quorum in election. However, in our case, a cluster should keep working on leader election and log replication even in such condition referred above. Therefore, we should consider a way to use both `quorum` and `majority` methods.

rapsealk:
Migrated to #697

PR #537, opened 2022-07-11
refactor: separate duplicate decorators from start and run

Yaminyam:
Due to the reduced readability due to overlapping decorators, the commonly needed decorators have been integrated into one decorator.


changes/537.misc.md
```diff
@@ -0,0 +1,4 @@
+Add `click_start_option` function to create common start decorator
+Separate `app.py, ssh.py, args.py` into session submodules
+Rename `session.py` to `lifecycle.py` and separate it into session submodule
+Rename `run.py` to `execute.py` and separate it into session submodule
```
achimnol: The changelog should be in one single sentence written in a single line. If there are multiple notable changes, please summarize the intention and the collective goal.

achimnol (changes requested):
How about rename `args` module to `_args` to distinguish it (a utility-containing module) from other command-containing modules?

src/ai/backend/client/cli/session/args.py
```diff
@@ -0,0 +1,154 @@
...
+def decorator_group(*decs):
+    def d(f):
+        for decorator in decs[::-1]:
+            f = decorator(f)
+        return f
+
+    return d
+
+
+def click_start_option():
```
fregataa: How about adding type hint of return value?
And I suggest to declare the options in one list as a constant.
like the below.
```
START_OPTION = [
    click.option(),
    click.option(),
    ...
]
```

fregataa (commented):
looks good!
I leave comments

achimnol (changes requested):
Let's leave comments after the command handler functions' args to indicate that which args are from which decorator.

e.g.,

```python
    @some_other_option()
    @click_start_option()
    def create(
        # base args
        image: str,                             # some_other_option
        name: str | None,                       # click_start_option
        owner: str | None,                      # some_other_option
        type: Literal["batch", "interactive"],  # click_start_option
        starts_at: str | None,                  # click_start_option
        startup_command: str | None,            # click_start_option
        enqueue_only: bool,                     # click_start_option
        max_wait: int,                          # click_start_option
        no_reuse: bool,                         # click_start_option
...
```

achimnol (approved):
It's too late, but I think the refactoring of client SDK's session CLI module/subpkg is a nice thing to have. I see no apparent regression in this PR, so let's first merge it.

PR #539, opened 2022-07-12
fix: logging error when `log_record` has a key `level`, but its value is `None`

adrysn:


adrysn:
Close this PR since this is a duplicate of https://github.com/lablup/backend.ai/pull/540.

PR #541, opened 2022-07-13
setup: Set the minimum redis-py version to 4.3.4

achimnol:
As a follow-up to #468, we need to set the minimum redis-py version to 4.3.4 because 4.3.3 and lower versions have different return values of the `XAUTOCLAIM` API wrapper.
(refs redis/redis-py#2252)


kyujin-cho (approved):
LGTM!

Yaminyam (approved):
LGTM!

PR #543, opened 2022-07-14
feat: Add `keypair_resource.max_session_lifetime` to client

rapsealk:
While I was running unit tests, I faced an error and found it is because of inconsistency in schema. Before it was merged into a monorepo, there was a commit that added `KeypairResourcePolicy.max_session_lifetime` field to `backend.ai-manager`. However, it was not updated to `backend.ai-client-py`. Therefore, now I reflected the change to our client as well.

Related commit: https://github.com/lablup/backend.ai-manager/commit/74bfaf47aeb15b55afda9cffbecc2584253c2960

Before | After
------ | -----
스크린샷 2022-07-14 오전 11 03 54 | 스크린샷 2022-07-14 오전 11 02 11


src/ai/backend/test/cli_integration/conftest.py
```diff
@@ -128,3 +129,8 @@ def new_keypair_options() -> Tuple[KeypairOption, ...]:
             is_active=False, is_admin=False, rate_limit=100000, resource_policy="default"
         ),
     )
+
+
+@pytest.fixture(scope="module")
+def keypair_resource_policy() -> str:
+    return str(uuid.uuid4())
```
achimnol: Currently keypair resource policy does not have UUID-style IDs but just human-readable names as its primary key.
Let's keep this consistent in the test fixtures.

achimnol (changes requested):

src/ai/backend/test/cli_integration/conftest.py
rapsealk: Thank you for the feedback! Then I will fix it using faker.

achimnol (approved):
LGTM

PR #545, opened 2022-07-14
setup: Introduce our own PyPI to workaround upstream issues

achimnol:
For infinite stream of questions about installing grpcio on M1 Mac and until this problem is ultimately fixed by Google, we decided to have our own PyPI distribution service: https://dist.backend.ai/pypi/simple/

**WARNING:** The exact list of packages served there may change over time depending on our needs, so we advise the general audience *not* to rely on this package index.

Special thanks to @jsirois for quickly patching pantsbuild/pex#1849!


kyujin-cho:
🎉

jsirois (commented):
Excellent.

PR #546, opened 2022-07-14
ci: Workaround potential pants bug when running some commands in parallel

achimnol:
After merging #545, this began to happen in both local setups and GitHub Actions.


achimnol:
This was identified as a cache issue.
[Clearing the cache for Pants using the GitHub CLI](https://docs.github.com/en/rest/actions/cache#delete-github-actions-caches-for-a-repository-using-a-cache-key) and the cache key retrieved from the action run logs resolved the problem.

PR #548, opened 2022-07-14
fix(install-dev): A typo in sed expression while installing backend.ai-webui

studioego:
fix: set the backend.ai-webui bug on install-dev.sh


achimnol (approved):

achimnol:
I'll merge this directly without the news fragment.

Issue #544, opened 2022-07-14
New environmental variable to check the service ports of a compute session


adrysn:
Currently, there is no way for a user to check the pre-open ports inside a container. So, let's add a new environment variable, such as `BACKENDAI_SERVICE_PORTS` by reading the image's label, `service-ports`. It may be appended under the following code block:
https://github.com/lablup/backend.ai/blob/39c292eff9ff0978e8371d57349f298516335f99/src/ai/backend/manager/registry.py#L1235-L1242.

PR #549, opened 2022-07-14
New environmental variable to check the service ports of a compute session #544 (solved)

sonarypt:
Issue: New environmental variable to check the service ports of a compute session #544
Solution: `'BACKENDAI_SERVICE_PORTS': ",".join(str(port) for port in scheduled_session.preopen_ports)`


achimnol:
Please make a new branch after fork to make the CI run correctly -- recreate the PR please.
It is a not recommended practice to directly work on the main branches even in forks, because it would confuse your commit history when there are new commits in the origin/main and you are going to merge back them.  Since lablup/backend.ai repository evolves very fast, you will need to frequently merge back the main branch.

PR #551, opened 2022-07-14
fix: fix typo & check file on install-dev.sh

studioego:
Fix typo & check file on install-dev.sh
1. fix typo & check file on install-dev.sh

### DIFF
--- a/scripts/install-dev.sh
+++ b/scripts/install-dev.sh
@@ -521,13 +521,9 @@ install_editable_webui() {
     # In the former case, you may live-edit the webui sources while running them in the electron app.
     sed_inplace "s@webServerURL =@webServerURL = "'"'"http://127.0.0.1:${WEBSERVER_PORT}"'"@' config.toml
     sed_inplace "s@proxyURL =@proxyURL = "'"'"http://127.0.0.1:${WSPROXY_PORT}"'"@' config.toml
-    env_file=".env"
-    if [ ! -f $env_file ]; then
-      touch $env_file
-    fi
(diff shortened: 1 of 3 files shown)


achimnol (changes requested):
entrypoint.py is already fixed in #550.
Please remove the changes here. :wink:

achimnol (approved):

scripts/install-dev.sh
```diff
@@ -515,15 +515,20 @@ install_editable_webui() {
...
-    echo "PROXYBASEHOST=localhost" >> src/ai/backend/webui/.env
-    echo "PROXYBASEPORT=${WSPROXY_PORT}" >> src/ai/backend/webui/.env
+    env_file="src/ai/backend/webui/.env"
+    env_dir=`dirname $env_file`
+    if [ ! -f $env_file ]; then
+      mkdir -p $env_dir; touch $env_file
+    fi
+    echo "PROXYLISTENIP=0.0.0.0" >> $env_file
+    echo "PROXYBASEHOST=localhost" >> $env_file
+    echo "PROXYBASEPORT=${WSPROXY_PORT}" >> $env_file
```
achimnol: We have already `cd`'ed into `src/ai/backend/webui`. So here `$env_file` must be just `.env`. The original path was my mistake.

achimnol (changes requested):

scripts/install-dev.sh
studioego: I fixed it. remove env_dir and checking only env_file

achimnol (approved):

PR #556, opened 2022-07-15
Env var for service ports of a compute session #544

sonarypt:
Issue: New environmental variable to check the service ports of a compute session #544
Solution: 'BACKENDAI_SERVICE_PORTS': ",".join(str(port) for port in scheduled_session.preopen_ports)`

### DIFF
--- a/src/ai/backend/manager/registry.py
+++ b/src/ai/backend/manager/registry.py
@@ -1246,6 +1246,7 @@ class AgentRegistry:
             'BACKENDAI_CLUSTER_HOSTS':
                 ",".join(binding.kernel.cluster_hostname for binding in kernel_agent_bindings),
             'BACKENDAI_ACCESS_KEY': scheduled_session.access_key,
+            'BACKENDAI_SERVICE_PORTS': ",".join(str(port) for port in scheduled_session.preopen_ports),
         })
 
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/registry.py
```diff
@@ -1246,6 +1246,7 @@ async def start_session(
             'BACKENDAI_CLUSTER_HOSTS':
                 ",".join(binding.kernel.cluster_hostname for binding in kernel_agent_bindings),
             'BACKENDAI_ACCESS_KEY': scheduled_session.access_key,
+            'BACKENDAI_SERVICE_PORTS': ",".join(str(port) for port in scheduled_session.preopen_ports),
```
adrysn, line 1249: What happens when `scheduled_session.preopen_ports` is `None`?

adrysn: Also, how about add two environment variables:
- `BACKENDAI_SERVICE_PORTS`: Service ports defined in the container image label (`ai.backend.service-ports`).
- `BACKENDAI_PREOPEN_PORTS`: Pre-open ports delivered via the request parameters (`scheduled_session.preopen_ports`).

adrysn (changes requested):

Issue #486, opened 2022-06-21
Let's define a new `ReturnCode` enum to clarify its meaning


rapsealk:
**Is your feature request related to a problem? Please describe.**
There are some methods returning numeric values(e.g. 0, -1, 1, etc.) which can be used as exit codes. However, it is hard to understand their intended meaning since they are put in [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)). Please read the code below as an example:
https://github.com/lablup/backend.ai/blob/17216c821325a55e5d86c45f090ecde71fd4ac33/src/ai/backend/client/cli/session.py#L1018-L1027
If there occurs an error and the process exits with code `1`, it might be hard to find out the actual reason; it can be a `TimeoutError`, or possibly a `RuntimeError`.

**Describe the solution you'd like**
Therefore, I suggest to make a new `ReturnCode` enum class to make their meaning clear.
```py
class ReturnCode(enum.Enum):
    OK = 0
    # Add more here..
```
When it is done, all the exit codes whose written in magic number have to be changed to the `ReturnCode` values.

**Describe alternatives you've considered**
-

**Additional context**
-

rapsealk:
Closed as PR559 is merged.

PR #559, opened 2022-07-15
refactor: Express CLI exitcode using semantic enum values

Yaminyam:
#486 As mentioned in the issue, it is expressed as an enum so that the exitcode expressed as a number can be explicitly recognized.


src/ai/backend/client/cli/types.py
```diff
@@ -20,3 +20,10 @@ class CLIContext:
     api_config: APIConfig = attr.field()
     output_mode: OutputMode = attr.field()
     output: BaseOutputHandler = attr.field(default=None)
+
+
+class ExitCode(enum.Enum):
+    OK = 0
+    ERROR = 1
+    TIMEOUT = 2
+    INVALID_ARGUMENTS = 3
```
achimnol: Um, I'd like to put this on `ai.backend.cli.types`, not `ai.backend.client.cli.types` because it could be reused across all other CLI modules, even in manager/agent/etc.

Also, how about renaming `ERROR` to `FAILURE` becaues 'error' means a narrower scope of failures such as bug or system failures while 'failure' generally means "we could not accomplish what you wanted" for whatever reasons.

Others look good.

achimnol (changes requested):

src/ai/backend/client/cli/types.py
Yaminyam: I was worried because there were behaviors that were fail rather than errors, but I think it's good that both can be included if it is marked as `FAILURE`.

Yaminyam:
> Nice work! Good to see this issue is finally in progress now. Could you please rewrite the changelog to make its meaning more clear and concrete?

`To clarify the meaning of the exit code expressed as a magic number, it is expressed as an enum ReturnCode.`
How about an expression of degree?

changes/559.misc.md
```diff
@@ -0,0 +1 @@
+Expressed as an enum so that the exitcode expressed as a number can be explicitly recognized.
```
achimnol: The sentence is not complete. Where is the objective of the verb "expressed"? Also, in changelog we primarily use the imperative form.

achimnol (changes requested):

changes/559.misc.md
Yaminyam: How about the above comment that was rewritten because others pointed it out?
`To clarify the meaning of the exit code expressed as a magic number, it is expressed as an enum ReturnCode.`

achimnol: I'd suggest:
> Introduce `ExitCode` enum to give concrete semantics to numeric CLI exit codes

achimnol (approved):

PR #565, opened 2022-07-19
refactor: move scaling group check

### LINKED ISSUE #500: Move `check_scaling_group()` predicate check to a precondition check in `enqueue_session()`
Just like lablup/backend.ai-manager#504.

Then, there will be no more `permanent=True` predicate check results and all predicates will become "temporary", which means that cluster state changes such as termination of existing sessions may flip the results.

fregataa:
resolves #500 
move validator which check scaling group by session type from predicate to enqueue_session.


achimnol (approved):
Great!

PR #566, opened 2022-07-19
feat(client): Support `.env` file to populate configurations

### LINKED ISSUE #523: Add support for `.env` to the client SDK
Let's make our client SDK and its CLI to automatically detect and read `.env` file if exists.

KangJuSeong:
resolve #523 
auto detect .env file


.git-blame-ignore-revs
```diff
@@ -1 +1 @@
-685c50b017a2c2b9d5d5722efa5ceffd2235992a
+ebcd16b478349949a6aef1dae478a892c1a6c57b
```
achimnol: This should be kept as-is.

changes/566.feature.md
```diff
@@ -0,0 +1 @@
+Using python-dotenv for auto detect .env file
```
achimnol: Please use the imperative form of sentence and describe not just what you did but what it changes in the perspective of users.

achimnol (changes requested):

changes/566.feature.md
```diff
@@ -1 +1 @@
-Using python-dotenv for auto detect .env file
+Can add environmental variables through files that were previously set as a bash file
```
achimnol: Now it is much better in the sense of expressing the intention, but still there are more improvements required.

First, "Can add ~~" is not an imperative form but an incomplete fragment.

Second, since the changelog should include technical details to make other developers aware of, when introducing new dependencies and the new dependency takes a key role like this PR, I believe it's better to mention `python-dotenv` explicitly.

achimnol (approved):

PR #569, opened 2022-07-20
feat: option to always enqueue compute session when there is not enough resources

adrysn:


achimnol (changes requested):
For the sake of semantics, this should not be a boolean flag but a two-value enum choice because "always" is a subjective word for new code readers and requires a background context before the change. Boolean flags should be used for "allow/disallow/forbid/enable/disable" stuffs.
I'd suggest to use something like `session_creation_with-insufficient_resource` with choices of `enqueue` or `forbid-creation`.

adrysn:
Duplication of https://github.com/lablup/backend.ai/pull/732

PR #571, opened 2022-07-21
feat: add a paginated query for shared vfolder list

Sujin-Kim1:
Add a paginated query for the shared vfolder list for superadmin.
This shows the shared vfolder list of all users.

Related PR: https://github.com/lablup/backend.ai-control-panel/pull/287


src/ai/backend/manager/models/gql.py
```diff
@@ -394,6 +399,15 @@ class Queries(graphene.ObjectType):
         access_key=graphene.String(),  # must be empty for user requests
     )
 
+    # super-admin only
+    shared_vfolder_list = graphene.Field(
```
achimnol: I'd like to rename it as `vfolder_share_list` or more directly, `vfolder_permission_list`.

- "Shared vfolder" -> the target object is vfolder.
- "VFolder share" -> the target object is the permission or the fact that it is shared.

achimnol (changes requested):

src/ai/backend/manager/models/gql.py
agatha197: I renamed it to `vfolder_permission_list`.

PR #574, opened 2022-07-21
feat: add shared_by value when query accessible vfolders

### LINKED ISSUE #568: feat: Update `query_accessible_vfolders()` to include shared-by/shared-with flags in the result
Users cannot easily distinguish between private folders, folders shared by me, or folders shared by others.

Update query_accessible_vfolders() to include shared-by/shared-with flags in the result and apply them to the virtual folder's query filter.

**Additional context**
- If there is a row with a `vfolder_id` whose id matches in the `vfolder_permissions` table, the `shared_by_me` is true.
- If there is a row with a `user_id` in the `vfolder_permissions` table that matches the person who requests `user_id`, you can reversely refer to the vfolder table to get the `shared_with_me` information.

fregataa:
resolves #568 

and additionally refactor `query_accessible_vfolders()` function.
`query_accessible_vfolders()` function calls another function based on user role,
and each function divided by user role queries owned vfolders and group vfolders.

blocked by https://github.com/lablup/backend.ai/pull/521


fregataa:
close this PR since we will use graphQL and https://github.com/lablup/backend.ai/pull/1437 will satisfy this need.

PR #576, opened 2022-07-22
refactor: Split the session & kernel tables with ORM migration

### LINKED ISSUE #325: Split session and kernel database tables

fregataa:
resolves #325

## 1. Add Session DB table.
- This table is used to schedule and manage compute sessions.
- It seems to have many columns that overlaps with kernel DB table. We need to normalize those tables according to our spec.

## 2. Migrate to SQLAlchemy ORM partially
- Migrate manager's scheduler codes from SQLAlchemy Core to ORM and use SQLAlchemy 2.0 query styles. We still are able to use Core tables and SQLAlchemy 1.x query styles.

## 3. Change compute session scheduler to use session table and APIs.
- We have managed compute session with `kernel` table which cause blocking issues -e.g) cannot implement session tag table- and make us confused. From now, data of scheduling jobs such as enqueuing, preparing and destroy compute session will be inserted to session table. And we can handle those scheduling job data by APIs implemented in `sessions.py`.

## 4. Optimize DB query
- Thanks to the new session table, we are able to optimize some DB query statements especially which are related to Manager's scheduler. 
- But there are a few obstacles since SQLAlchemy does not fully support Asynchronous tasks -e.g) Unable lazy loading- and usage of SQLAlchemy's `orm.Session` is not concrete -e.g) boundary of `orm.Session` is arbitrary. We can optimize such tasks much more.

## Plus
- We need to refactor this scheduler codes since this migrated code followed some legacy codes. e.g) it assumes that there is no heterogeneous kernels in one session such as no different images or etc, Updating status of compute sessions and kernels process is not clear yet. Need to specify the steps and interactions between session and kernel through updating status.
- some of manager API still use SQLAlchemy Core based queries. We can update the codes.

## TODO
1. Normalize Session DB table and Kernel DB table. Mainly drop duplicated and legacy columns.
2. Optimize the usage of SQLAlchemy Session.

## Question
- Which one seems better to [prevent Implicit IO when Using AsyncSession](https://docs.sqlalchemy.org/en/14/orm/extensions/asyncio.html?highlight=async#preventing-implicit-io-when-using-asyncsession)??
[no loading strategy](https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html#sqlalchemy.orm.noload) vs [raise loading strategy](https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html#sqlalchemy.orm.raiseload)
They have different result when we try to access the un-loaded relationship. `noload()` let the relationship returns `None` or `[]`, and `raiseload()` raise [`InvalidRequestError`](https://docs.sqlalchemy.org/en/14/core/exceptions.html#sqlalchemy.exc.InvalidRequestError)


src/ai/backend/manager/models/base.py
```diff
@@ -177,13 +178,15 @@ def process_bind_param(self, value: Union[Mapping, ResourceSlot], dialect):
...
+    def process_result_value(self, raw_value: Dict[str, str] | None, dialect):
+        if raw_value is None:
+            return None
         # legacy handling
         interim_value: Dict[str, Any] = raw_value
         mem = raw_value.get("mem")
         if isinstance(mem, str) and not mem.isdigit():
             interim_value["mem"] = BinarySize.from_str(mem)
-        return ResourceSlot.from_json(interim_value) if raw_value is not None else None
+        return ResourceSlot.from_json(interim_value)
```
fregataa: need to handle when `raw_value` is None.

src/ai/backend/manager/models/kernel.py
```diff
@@ -818,378 +813,13 @@ async def batch_load_detail(
             )
 
 
-class ComputeSession(graphene.ObjectType):
```
fregataa, line 932: move `ComputeSession` to session.py

src/ai/backend/manager/models/session.py
```diff
@@ -0,0 +1,1066 @@
...
+    KernelStatus.RESIZING: SessionStatus.RUNNING,
+    KernelStatus.SUSPENDED: SessionStatus.RUNNING,
+    KernelStatus.TERMINATING: SessionStatus.TERMINATING,
+    KernelStatus.TERMINATED: SessionStatus.TERMINATED,
+    KernelStatus.ERROR: SessionStatus.ERROR,
+    KernelStatus.CANCELLED: SessionStatus.CANCELLED,
+}
+
+
+def aggregate_kernel_status(
```
kyujin-cho: Name of the function doesn't clearly shows what this function does.
```suggestion
def determine_session_status_by_kernels(
```

src/ai/backend/manager/models/session.py
```diff
@@ -0,0 +1,1066 @@
...
+            if s.cluster_role == DEFAULT_ROLE:
+                return SessionStatus.ERROR
+            return SessionStatus.RUNNING_DEGRADED
+        elif s.status in priority_statuses:
+            candidate = KERNEL_SESSION_STATUS_MAPPING[s]
+        elif s.status == target_status:
+            continue
+        else:
+            # TODO: handle unexpected kernel status
+            pass
```
kyujin-cho: I think we can just raise `ValueError` instead of leaving this as `TODO`.

```suggestion
            raise ValueError(f"Unexpected status {s.status}")
```

src/ai/backend/manager/models/session.py
```diff
@@ -0,0 +1,1066 @@
...
+    access_key: AccessKey | None = None,
+    *,
+    max_matches: int | None,
+    allow_stale: bool = True,
+    for_update: bool = False,
+    do_ordering: bool = False,
+    eager_loading_op: Iterable | None = None,
+):
+    cond = base_cond
+    if access_key is not None:
```
kyujin-cho: Just to match our coding style:
```suggestion
    if access_key:
```

src/ai/backend/manager/models/session.py
```diff
@@ -0,0 +1,1066 @@
...
+        .execution_options(populate_existing=True)
+    )
+    if max_matches:
+        query = query.limit(max_matches).offset(0)
+    if for_update:
+        query = query.with_for_update()
+    if do_ordering:
+        query = query.order_by(SessionRow.created_at)
+
+    if eager_loading_op is not None:
```
kyujin-cho, line 456: Same here.
```suggestion
    if eager_loading_op:
```

Issue #581, opened 2022-07-23
Unable to apply version 2 WSProxy when the coordinator's user-accessible URL is different from the manager-accessible URL


adrysn:
If a user accesses Backend.AI service via NAT, the user-facing address of the WSProxy (Proxy Coordinator) can be different from the manager-facing one. In this case, we cannot apply version 2 WSProxy since user's browser will try to connect to the WSProxy with the address saved at `scaling_groups.wsproxy_addr`, which is only accessible behind the NAT.

So, I suggest to add a feature to report advertise address of the WSProxy to users when opening an app. This will increase the coverage of the version 2 WSProxy.

PR #582, opened 2022-07-23
feat: support advertise URL for WSProxoy v2

adrysn:
Return WSProxy advertise URL, if exists, by querying the value from the coordinator in starting an app service. Since the coordinator's `status` handler now returns the advertise address, the `query_wsproxy_version` method is renamed to `query_wsproxy_status`.

This PR resolves https://github.com/lablup/backend.ai/issues/581.

### DIFF
--- a/src/ai/backend/manager/api/scaling_group.py
+++ b/src/ai/backend/manager/api/scaling_group.py
@@ -33,7 +33,7 @@ class WSProxyVersionQueryParams:
 @aiotools.lru_cache(expire_after=30)  # expire after 30 seconds
 async def query_wsproxy_status(
     wsproxy_addr: str,
-) -> str:
+) -> dict[str, Any]:
     async with aiohttp.ClientSession() as session:
         async with session.get(wsproxy_addr + "/status") as resp:
             return await resp.json()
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/api/scaling_group.py
```diff
@@ -31,13 +31,12 @@ class WSProxyVersionQueryParams:
 
 
 @aiotools.lru_cache(expire_after=30)  # expire after 30 seconds
-async def query_wsproxy_version(
+async def query_wsproxy_status(
     wsproxy_addr: str,
 ) -> str:
```
achimnol: The return type should be a `dict[str, Any]` as it seems to directly return the JSON object response.

src/ai/backend/manager/api/session.py
```diff
@@ -1245,6 +1246,11 @@ async def start_service(request: web.Request, params: Mapping[str, Any]) -> web.
     wsproxy_addr = sgroup["wsproxy_addr"]
     if not wsproxy_addr:
         raise ServiceUnavailable("No coordinator configured for this resource group")
+    wsproxy_status = await query_wsproxy_status(wsproxy_addr)
+    if wsproxy_status.get("advertise_address"):
+        wsproxy_advertise_addr = wsproxy_status["advertise_address"]
```
achimnol: We can avoid dupliacte lookups it using the walrus operator:
```suggestion
    if advertised_addr := wsproxy_status.get("advertise_address"):
        wsproxy_advertise_addr = advertised_addr
```

achimnol (changes requested):

achimnol (approved):

PR #588, opened 2022-07-24
fix: Incorrect warning message is output when allocating a size small…

hotkimho:
…er than the size of the image you are trying to create

Modify the json SmallerResourceThenImageRequires value "자원가" to "자원이"


achimnol:
Seems to have fixed inside the submodule from [backend.ai-app](https://github.com/lablup/backend.ai-app) repository.
The correct place to contribute to the web UI is [backend.ai-webui](https://github.com/lablup/backend.ai-webui) repository.

When a new version of webui is released, the transpiled and minified version is pushed to the app repository and then we can update the submodule reference here.

Issue #563, opened 2022-07-18
Support committing and pushing a container image from a running session


achimnol:
There are several customer requests for supporting "freezing" their current compute sessions.

Historically, we _intentionally_ have not added this feature due to _inherent volatility_ and _security restrictions_ of Docker containers.  The filesystem that users see in the sessions is **a composition of multiple local/remote filesystems**, such as the container image, vfolders, and scratch directories.  For security, we don't expose the root user inside the container by default, so there is usually no way to modify the container filesystem provided by the image because all system packages and directories are owned by root.

To allow installation of additional packages using user-site paths in Python (`pip install --user`) and make it persistent across different compute sessions, we have added the following features:

- #98
- #99
- auto-mounting dot-prefixed vfolders (such as `~/.config`, `~/.local`)
  Nevertheless, many HPC/AI customers want to use the containers like VMs, and it is not easy to fill the conceptual gap between the volatile & hermetic nature of containers and the full ownership of volume data of VMs.

So, despite whatever additional cautions required when committing a Backend.AI compute session, let's make it technically available.

JIRA Issue: BA-259

adrysn:
Requirements for this feature:
- [ ] (Web-UI) Add a "commit" button in the "Controls" column per compute session.
  - The commit button should be disabled or displayed as a spinning icon if the commit task for the session is on-going.
- [ ] (Manager) API handler that accepts the request from the client and then relays it to the Agent that runs the target container.
  - Only the owner of the compute session can issue the commit request.
  - A new config parameter that specifies the save location of the `.tar.gz` file. The location should usually be on a network filesystem.
  - API handler that returns the status of the commit task.
- [ ] (Agent) RPC handler that executes the actual commit operation for the target container.
  - The resulting `.tar.gz` should be saved on the specified location.
  - RPC handler that returns the status of the commit task.

PR #601, opened 2022-07-25
feat: Commit and store running session containers as tarball images

fregataa:
notes: #563 

this PR does not fully covers the issue.

TODO:
- [x] Use toml file or `config` to set commit destination
- [ ] ~~performance.. Exporting Docker container from a running session is time consuming.~~ -> asynchronous reading chunk and writing to tarball file.


src/ai/backend/common/bgtask.py
```diff
@@ -218,15 +219,16 @@ async def _pipe_builder(r: Redis) -> Pipeline:
 
         await redis_helper.execute(redis_producer, _pipe_builder)
 
-        task = asyncio.create_task(self._wrapper_task(func, task_id, name))
+        task = asyncio.create_task(self._wrapper_task(func, task_id, name, **kwargs))
         self.ongoing_tasks.add(task)
-        return task_id
+        return str(task_id)
```
fregataa: image

change `task_id` into str since rpc cannot return this `task_id` in UUID form

src/ai/backend/common/bgtask.py
```diff
@@ -194,13 +194,14 @@ async def start(
...
         name: str = None,
+        **kwargs,
     ) -> uuid.UUID:
         task_id = uuid.uuid4()
         redis_producer = self.event_producer.redis_client
 
         async def _pipe_builder(r: Redis) -> Pipeline:
             pipe = r.pipeline()
-            tracker_key = f"bgtask.{task_id}"
+            tracker_key = f"bgtask.{str(task_id)}"
```
achimnol: f-string literals already automatically calls `str()`.
No need to do it again.

src/ai/backend/agent/docker/kernel.py
```diff
@@ -24,6 +26,9 @@
 from ..resources import KernelResourceSpec
 from ..utils import closing_async, get_arch_name
 
+# TODO: set the commit tags directory from toml or cfg
+COMMIT_DIR = Path("/tmp/backend.ai/commit/")
```
achimnol: This should be taken from the configuration.

src/ai/backend/agent/docker/kernel.py
```diff
@@ -127,6 +132,52 @@ async def get_service_apps(self):
...
+        tag_path = commit_path / "tags" / container_id
+
+        tag_path.parent.mkdir(parents=True, exist_ok=True)
+        try:
+            with open(tag_path, mode="x") as f:
+                if get_lock:
+                    f.write("")
+        except FileExistsError:
+            return int(False)
+        return int(True)
```
achimnol: Why do such non-trivial type conversion?
Make the method name more standing out with the intention and/or use enum or just plain bool values.

src/ai/backend/agent/docker/kernel.py
```diff
@@ -127,6 +132,52 @@ async def get_service_apps(self):
...
+            raise AssertionError("malformed committed path.")
+
+        loop = current_loop()
+        async with closing_async(Docker()) as docker:
+            async with docker._query(
+                f"containers/{container_id}/export",
+                method="GET",
+            ) as response:
+
+                data: bytes = await response.read()
```
achimnol: Blank lines should be minimized unless it is better to group statements as separate paragraphs. (It's just like writing.)

src/ai/backend/agent/kubernetes/kernel.py
```diff
@@ -202,6 +202,43 @@ async def get_service_apps(self):
         result = await self.runner.feed_service_apps()
         return result
 
+    async def check_commit_tag(self, commit_path, get_lock):
+        log.error("Committing in Kubernetes is not supported yet.")
+
+    async def commit(self, image_commit_path, path):
+        log.error("Committing in Kubernetes is not supported yet.")
+        # assert self.runner is not None
```
achimnol: Add `TODO:` or `FIXME:` comment so that future developers could refer there is something to do here.
Also, it's better to explicitly raise `NotImplementedError` instead of just logging it and doing nothing, particularly when the function is expected to return some values.

src/ai/backend/agent/kubernetes/kernel.py
```diff
@@ -202,6 +202,43 @@ async def get_service_apps(self):
         result = await self.runner.feed_service_apps()
         return result
 
+    async def check_commit_tag(self, commit_path, get_lock):
+        log.error("Committing in Kubernetes is not supported yet.")
```
achimnol, line 206: Same here.

src/ai/backend/client/func/session.py
```diff
@@ -574,6 +574,23 @@ async def rename(self, new_id):
         async with rqst.fetch():
             pass
 
+    @api_function
+    async def commit(self):
+        """
+        Commit a running session to tar file.
```
achimnol:
```suggestion
        Commit a running session to a tar file in the agent host.
```

src/ai/backend/common/bgtask.py
```diff
@@ -250,11 +252,11 @@ async def _wrapper_task(
 
             async def _pipe_builder(r: Redis):
                 pipe = r.pipeline()
-                tracker_key = f"bgtask.{task_id}"
+                tracker_key = f"bgtask.{str(task_id)}"
```
achimnol: Same here about the duplicate `str()` in f-string literals.

src/ai/backend/agent/docker/kernel.py
```diff
@@ -127,6 +132,52 @@ async def get_service_apps(self):
...
 
+    async def check_commit_tag(self, commit_path: Path, get_lock: bool = False):
+        container_id: str = str(self.data["container_id"])
+        tag_path = commit_path / "tags" / container_id
+
+        tag_path.parent.mkdir(parents=True, exist_ok=True)
+        try:
+            with open(tag_path, mode="x") as f:
+                if get_lock:
+                    f.write("")
```
achimnol: Let's use `common.lock.FileLock` for consistent operation.

src/ai/backend/manager/registry.py
```diff
@@ -2982,6 +2985,61 @@ async def mark_session_terminated(
...
+        async with self.db.begin_readonly() as db_conn:
+            query = (
+                sa.select([users.c.email])
+                .select_from(users)
+                .where(users.c.uuid == kernel["user_uuid"])
+            )
+            result = await db_conn.execute(query)
+            user_email = result.scalar()
+        # TODO: get path from toml or cfg
+        path = Path(str(user_email), str(kernel["session_id"]), filename)
```
achimnol: Is it safe to directly use email and ISO-formatted datetime string as path components?

achimnol: How is this value related with the agent-side commit path configuration as you noted as comments?

achimnol (changes requested):

configs/agent/halfstack.toml
```diff
@@ -15,6 +15,7 @@ pid-file = "./agent.pid"
 event-loop = "uvloop"
 var-base-path = "./var/lib/backend.ai"
 # allow-compute-plugins = []
+image-commit-path = "/tmp/backend.ai/commit/"
```
achimnol: If the default value is `/tmp`, it would break with Snap-based Docker installations. I'd recommend using a directory under the current working directory of agent.

src/ai/backend/manager/registry.py
fregataa: Id like to make sure admin can check whose commits they are.
agent-side commit path is the root of commit path and this email and ISO formatted datetime path is underlying path.

the path would be like the.
./tmp/backend.ai/commit/myemail@email/20220801

src/ai/backend/agent/docker/kernel.py
```diff
@@ -127,6 +132,51 @@ async def get_service_apps(self):
...
+                    method="GET",
+                ) as response:
+                    log.debug("Container commit data is being written on {}", filepath)
+                    data: bytes = await response.read()
+
+                    def _save_tar():
+                        tar_info = tarfile.TarInfo(filename)
+                        tar_info.size = len(data)
+                        with tarfile.open(name=filepath, mode="x:gz") as tar:
+                            tar.addfile(tar_info, fileobj=BytesIO(data))
```
achimnol: This will load all the output in the memory and write it at once, and is highly likely to cause out-of-memory of the host system.
Make it read a chunk and write a chunk. Refer the sample codes (this code is the server-side upload receiver but you can implement chunking in the same way in the client-side download receiver):
https://github.com/lablup/backend.ai/blob/141844d383970b7274b9861b625d7f51a6601998/src/ai/backend/storage/api/client.py#L245-L266

achimnol (changes requested):

src/ai/backend/agent/docker/kernel.py
```diff
@@ -149,31 +151,26 @@ async def commit(self, path: Path):
...
+                docker = Docker()
+                container = docker.containers.container(container_id)
+                response: Mapping[str, Any] = await container.commit()
+                image_id = response["Id"]
+                async with docker._query(f"images/{image_id}/get") as tb_resp:
+                    with tarfile.open(str(filepath), "w|gz") as tar:
+                        async for chunk in tb_resp.content.iter_chunked(DEFAULT_CHUNK_SIZE):
+                            assert tar.fileobj is not None
+                            tar.fileobj.write(chunk)
+                        tar.add(str(filepath))
```
achimnol: Isn't this a recursion in the tar file (tar file is added to the tar file)?

src/ai/backend/agent/docker/kernel.py
```diff
@@ -149,31 +151,26 @@ async def commit(self, path: Path):
...
-                    await loop.run_in_executor(None, _save_tar)
+                docker = Docker()
+                container = docker.containers.container(container_id)
+                response: Mapping[str, Any] = await container.commit()
+                image_id = response["Id"]
+                async with docker._query(f"images/{image_id}/get") as tb_resp:
+                    with tarfile.open(str(filepath), "w|gz") as tar:
+                        async for chunk in tb_resp.content.iter_chunked(DEFAULT_CHUNK_SIZE):
+                            assert tar.fileobj is not None
+                            tar.fileobj.write(chunk)
```
achimnol: This should reside in a separate thread. Simply we could use `asyncio.run_in_executor()`, or employ the structure of `common.files.AsyncFileWriter` which uses `janus` to stream the chunks via a queue shared by an executor task and the asyncio world.
For now, let's wrap it up using `asyncio.run_in_executor()` for simpler implementation.

src/ai/backend/agent/docker/kernel.py
```diff
@@ -149,31 +151,26 @@ async def commit(self, path: Path):
...
+                        tar.add(str(filepath))
+                await docker.images.delete(image_id)
+                await docker.close()
```
achimnol: Could we put these two methods in a `finally` block to guarantee their execution?

achimnol (changes requested):

src/ai/backend/agent/docker/kernel.py
fregataa: if multiple requests reach the `finally` block, those guaranteed codes will raise error.
should I do like this?
```
finally:
    try:
        await docker.images.delete(image_id)
        await docker.close()
    except:
        pass
```

src/ai/backend/agent/docker/kernel.py
fregataa: but the agent can process multiple requests at the same time with this code.
it can check the duplicated(or on-going) session commit and can create new session while the file write streaming is ongoing.

besides, I am not sure how to implement `asyncio.run_in_executor()`. since `tarfile` does not support appending mode for opening file.

achimnol: I will fix it by myself this part.

src/ai/backend/agent/docker/kernel.py
achimnol: It's already protected by `FileLock` and the finally block should be inside the lock region.  I will fix by myself.

PR #608, opened 2022-07-27
feat: add API to get shared vfolders information and API to update/delete permission

Sujin-Kim1:
- Description: To show the shared vfolder info and update the permission, I added the client-py API.
`shared_vfolders_info` shows the information such as folder name, folder id, owner, permission, folder type, shared to, etc.
`update_shared_vfolder` updates permission for the given shared vfolders.

- Related PR: https://github.com/lablup/backend.ai-control-panel/pull/287
---

**[Questions]**
- I'm not sure which one is better to set as a name, `shared vfolders_info` / `update_vfolder permission` and `update_shared_vfolder` / `vfolder_permission_info`.

- To prevent normal users from modifying other people's shared folders, the following options are possible. Which one is the best?
  1. Edit manager
  2. Set as admin-only API
  3. Add logic to determine whether client-py is admin or not


src/ai/backend/client/cli/vfolder.py
```diff
@@ -765,3 +765,61 @@ def update_options(name, permission, set_cloneable):
...
+    """Show the vfolder permission information of the given virtual folder.
+
+    \b
+    VFOLDER_ID: ID of a virtual folder.
+    """
+    with Session() as session:
+        try:
+            resp = session.VFolder.shared_vfolder_info(vfolder_id)
+            result = resp.get("shared", [])
+            if len(result) > 0:
```
fregataa:
```suggestion
            if result:
```
let's use Python truthiness.

src/ai/backend/client/cli/vfolder.py
```diff
@@ -765,3 +765,61 @@ def update_options(name, permission, set_cloneable):
...
+                print(
+                    'Virtual folder "{0}" (ID: {1})'.format(
+                        _result["vfolder_name"], _result["vfolder_id"]
+                    )
+                )
+                print("- Owner: {0}".format(_result["owner"]))
+                print("- Permission: {0}".format(_result["perm"]))
+                print("- Folder Type: {0}".format(_result["type"]))
+                shared_to = _result.get("shared_to", [])
+                if len(shared_to) > 0:
```
fregataa:
```suggestion
                if shared_to:
```
Also this one

src/ai/backend/client/cli/vfolder.py
```diff
@@ -765,3 +765,61 @@ def update_options(name, permission, set_cloneable):
...
+            sys.exit(ExitCode.FAILURE)
+
+
+@vfolder.command()
+@click.argument("vfolder_id", type=str)
+@click.argument("user_id", type=str)
+@click.option(
+    "-p", "--permission", type=str, metavar="PERMISSION", help="Folder's innate permission."
+)
+def update_shared_vfolder(vfolder_id, user_id, permission):
```
fregataa: how about rename it to `update_shared_vf_permission` or something else?
because it only allows to update permission.

fregataa (commented):
does admin allow to update permission of **any** vfolder even though the admin does not own the vfolder?
then it will be good to set the cli function as admin only

src/ai/backend/client/cli/vfolder.py
agatha197: good point! I changed it.

src/ai/backend/client/cli/vfolder.py
agatha197: You're right. Thanks for suggestion!

src/ai/backend/client/cli/admin/vfolder.py
```diff
@@ -203,3 +203,61 @@ def umount_host(name, edit_fstab):
...
+    "-p", "--permission", type=str, metavar="PERMISSION", help="Folder's innate permission."
+)
+def update_shared_vf_permission(vfolder_id, user_id, permission):
+    """
+    Update permission for shared vfolders.
+    If permission is None, remove user's permission for the vfolder.
+
+    \b
+    VFOLDER_ID: ID of a virtual folder.
+    USER_ID: ID of user who shared the vfolder.
```
kyujin-cho: This description can be misread as "the user who owns the vFolder".
```suggestion
    USER_ID: ID of user who have been granted access to shared vFolder.
```

src/ai/backend/client/cli/admin/vfolder.py
```diff
@@ -203,3 +203,61 @@ def umount_host(name, edit_fstab):
...
+            sys.exit(ExitCode.FAILURE)
+
+
+@vfolder.command()
+@click.argument("vfolder_id", type=str)
+@click.argument("user_id", type=str)
+@click.option(
+    "-p", "--permission", type=str, metavar="PERMISSION", help="Folder's innate permission."
+)
+def update_shared_vf_permission(vfolder_id, user_id, permission):
```
kyujin-cho, line 246: Serving both permission update and removal feature on one command can confuse user. How about splitting those features as a separate command (`update-shared-vf-permission` and `remove-shared-vf-permission`)?

kyujin-cho (changes requested):

kyujin-cho (approved):

Issue #607, opened 2022-07-27
If Linux distribution system Python version is 3.6 and older, show the 'capture_output' parameter error on  'check-docker.py'.


studioego:
**Describe the bug**
If Linux distribution system Python version is 3.6 and older, show the 'capture_output' parameter error on  'check-docker.py'.

Some Linux distribution's default system Python version is 3.6. such as  CentOS7, 8, 9 and  openSUSE Leap 15.x versions. 

After checking "install_script_deps", checking docker using "scripts/check-docker.py"
https://github.com/lablup/backend.ai/blob/main/scripts/install-dev.sh#L543

```console
Detected Docker installation: System package (20.10.17)
Traceback (most recent call last):
  File "scripts/check-docker.py", line 108, in <module>
    main()
  File "scripts/check-docker.py", line 90, in main
    proc = run(['docker', 'compose', 'version'], capture_output=True, check=True)
  File "/usr/lib64/python3.6/subprocess.py", line 423, in run
    with Popen(*popenargs, **kwargs) as process:
TypeError: __init__() got an unexpected keyword argument 'capture_output'
```

https://github.com/lablup/backend.ai/blob/main/scripts/check-docker.py#L90
Python3's subprocess.run() parameter `capture_output` added in Python version 3.7

Below is the details of Python3's subprocess module.
https://docs.python.org/3.10/library/subprocess.html#using-the-subprocess-module
"Changed in version 3.7: Added the text parameter, as a more understandable alias of universal_newlines. Added the capture_output parameter."

**To Reproduce**
1. try to install Backend.AI using 'script/install-dev.sh'
2. See error

**Expected behavior**
Successfully installed Backend.AI

**Screenshots**
![openSUSE-Leap-15 4_aOUde2lGqG](https://user-images.githubusercontent.com/20002/181193847-fceedb2a-3880-4ed1-8045-3e9d7fd91eae.png)

```console
Detected Docker installation: System package (20.10.17)
Traceback (most recent call last):
  File "scripts/check-docker.py", line 108, in <module>
    main()
  File "scripts/check-docker.py", line 90, in main
    proc = run(['docker', 'compose', 'version'], capture_output=True, check=True)
  File "/usr/lib64/python3.6/subprocess.py", line 423, in run
    with Popen(*popenargs, **kwargs) as process:
TypeError: __init__() got an unexpected keyword argument 'capture_output'
dhsung@DESKTOP-042UROI:~/backend.ai> pyenv install 3.10.5
[...]

PR #612, opened 2022-07-28
fix: checking for python 3.7 and greater before calling checking-docker.py

studioego:
Checking for python 3.7 and greater before calling  `script/checking-docker.py`

The commit will make the Backend.AI installation script(script/install-dev.sh) more robust.

close https://github.com/lablup/backend.ai/issues/607


kyujin-cho (changes requested):
Refactoring script to check system's bundled python version and raising error if version is lower than 3.6 seems more clearer.

achimnol:
Closing in favor of #1122.

PR #613, opened 2022-07-28
fix: Accept both `str` and `FieldSpec` in Client SDK's functional APIs

fregataa:
fix cli's api methods which has wrong `field` types.
change the type from str to FieldSpec and usage of join of them


src/ai/backend/client/func/user.py
```diff
@@ -253,14 +253,19 @@ async def create(
...
         Creates a new user with the given options.
         You need an admin privilege for this operation.
         """
         if fields is None:
-            fields = ("domain_name", "email", "username", "uuid")
+            fields = (
+                FieldSpec("domain_name"),
+                FieldSpec("email"),
+                FieldSpec("username"),
+                FieldSpec("uuid"),
```
achimnol: I think we should refer `ai.backend.output.fields.user_fields`.

achimnol (changes requested):
For backward-compatibility and brevity, could we patch it to accept `Iterable[FieldSpec | str]` and automatically resolve the str values into FieldSpec by referring appropraite `output.fields` pre-defined tables?

src/ai/backend/client/func/domain.py
```diff
@@ -119,7 +119,10 @@ async def create(
             }
         """
         )
-        query = query.replace("$fields", " ".join(f.field_ref for f in fields))
+        resolved_fields = tuple(
+            f.field_ref if isinstance(f, FieldSpec) else domain_fields[f].field_ref for f in fields
+        )
```
achimnol: Though this is a simple logic, it is repeated many times.
Let's refactor this!

achimnol (changes requested):

src/ai/backend/client/func/domain.py
fregataa: refactor this as a new function named `resolve_field()`.
additionally, implement `field_resovler` decorator but it is not used yet.
I will not adapt the decorator if using `resolve_field()` is enough.

src/ai/backend/client/func/base.py
```diff
@@ -41,6 +46,38 @@ def api_function(meth):
     return meth
 
 
+def resolve_field(
```
achimnol: Since this function works on a list of fields, let's rename it to `resolve_fields`.

achimnol (changes requested):

src/ai/backend/client/func/base.py
achimnol: Also please fix up test failures.

achimnol (approved):

Issue #614, opened 2022-07-28
ImageNotFound message typo error when creating session with non-exist image references


soheeeeP:
**Describe the bug**
When creating session with non-exist image references, ImageNotFound 404 error is raised.
Here, I found ImageNotFound message typo error. 'Unknown' is incorrectly spelled like 'Unkown'.

**To Reproduce**
1. Run `./backend.ai mgr start-server --debug`
2. Run `./backend.ai ag start-server --debug`
3. Run `./py -m ai.backend.storage.server` &&   `./py -m ai.backend.web.server`
4. Log-in to the console API proxy
5. Create session  with non-exist image references and see typo error message

**Expected behavior**
See correctly spelled error message.

**Screenshots**
![image](https://user-images.githubusercontent.com/55832904/181655640-88e89092-6e99-44a0-a244-cff83fd0bfe2.png)
![image](https://user-images.githubusercontent.com/55832904/181655670-8d55bcab-ab36-4539-9fb1-734c17d54405.png)

**Client:**
 - OS: [e.g. macOS 10.14, Windows 10 (1903)]
 - App version: [e.g. 19.08.3] (if you are using a GUI client app)
 - Browser version: [e.g. Chrome 76, Safari, Edge] (if you are using a GUI web console)
 - Python version: [e.g. 3.7.4] (if you are using the Python SDK)
 - NodeJS version: [e.g. 12.0] (if you are using the Javascript SDK)

**Server:**
 - OS: Linux linux-dev 5.4.0-100-generic #113-Ubuntu SMP Thu Feb 3 18:43:29 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
 - Backend.AI version: main
 - Python version: 3.8.10
 - Installation method: scripts/install-dev.sh

PR #615, opened 2022-07-29
fix: Incorrectly spelled word in ImageNotFound exception message

soheeeeP:
fix: incorrectly spelled word in ImageNotFound exception message.

Change & Correct word from 'Unkown' to 'Unknown'.

fix issue #614

### DIFF
--- a/changes/615.fix.md
+++ b/changes/615.fix.md
@@ -1,0 +1,1 @@
+correct misspelled word in ImageNotFound exception message.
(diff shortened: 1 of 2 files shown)


rapsealk:
Thank you for the reporting!

rapsealk (commented):
It would be great if the changelog starts with a capital letter.
I tried to fix it by myself, but it seems that I'm not allowed to access your repository.

PR #617, opened 2022-08-01
fix: Repeated access error fix for stale containers

### LINKED ISSUE #555: Agent's kernel registry sees stale containers when they fail at docker API
When there are configuration issues such as enabling the experimental VirtioFS in Docker Desktop for Mac which does not support mounting UNIX domain sockets into containers and thus making the socket-relay container to repeat restarting infinitely, the agent observes an immediate failure upon the Docker's container start API.

When this happens, the agent's stat collector tries to access those non-created containers and keeps saying:
image

There may be other errors when creating containers even after validating all the configurations in our side, such as the network storage mount is gone away for unknown hardware/network failures (sometimes happened in our on-premise customer sites).  In that case the failure should be handled gracefully.

Let's properly retire such stale entries.

KangJuSeong:
resolve #555


achimnol:
I think this PR should be rewritten as the diff shows so many unncessary file additions and changes unrelated to what it does.

achimnol:
Closing as stale. Please reopen with a clean history based on the latest main if you're available.

PR #618, opened 2022-08-01
docs: translate for 2022 OSSCA

fregataa:
This is a translation branch for 2022 OSSCA


fregataa:
close this PR since OSSCA finished.

Issue #609, opened 2022-07-27
hiredis-py wheel build error with gcc 4.8.5 on CentOS 7 due to missing Python 3.10 wheels


studioego:
**Describe the bug**
During Backend.AI installation, a c99 error occurred in Hiredis while proceeding with the wheel build of the pantsbuild.

on Cent OS7, gcc version is 4.8.5.

```console
[dhsung@localhost backend.ai]$ gcc --version
gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

```console
  gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -I/usr/include/openssl11 -I/usr/include/openssl11 -fPIC -Ivendor -I/home/dhsung/.cache/pants/named_caches/pex_root/venvs/8319bbb11ce5c15b9ed667feb4c6427d31617cf3/5bf7713a1151cad07d23c23a677807ba4604ef8a/include -I/home/dhsung/.pyenv/versions/3.10.5/include/python3.10 -c vendor/hiredis/read.c -o build/temp.linux-x86_64-3.10/vendor/hiredis/read.o
  vendor/hiredis/read.c: In function ‘redisReaderFree’:
  vendor/hiredis/read.c:646:9: error: ‘for’ loop initial declarations are only allowed in C99 mode
           for (int i = 0; i < r->tasks; i++) {
           ^
  vendor/hiredis/read.c:646:9: note: use option -std=c99 or -std=gnu99 to compile your code
  error: command '/usr/bin/gcc' failed with exit code 1
  ----------------------------------------
  ERROR: Failed building wheel for hiredis
ERROR: Failed to build one or more wheels
```

**To Reproduce**
1. Installing Backend.AI on CentOS 7
2. See error

**Expected behavior**
Successfully installed Backend.AI 

**Screenshots**
![powershell_JvPf39pMHs](https://user-images.githubusercontent.com/20002/181288431-25198849-4038-42be-9825-f11deb2a15ab.png)

Below is full log for build error on CentOS 7.
```console
[...]

PR #620, opened 2022-08-01
hotfix: Update corresponding location of `grpcio`-related packages

rapsealk:
Since there is an issue about installing `grpcio` on ARM64 macOS, we are currently using our own wheel repository. However, after the wheel location has changed due to git-lfs, it is not updated on our repository, which is actually using the wheel file.
In this PR, I changed the `url` of `grpcio`-related packages **manually** and tested the install process.
You may compare the location from both [backend.ai-oven](https://github.com/lablup/backend.ai-oven/blob/8dcc73e1f99dccdce36d3b70562afbf2244d92b3/pypi/package/index.html) and [backend.ai](https://github.com/lablup/backend.ai/blob/f05b4362c490390c5f32029893b53e4d64115ff6/python.lock#L1584).

### DIFF
--- a/python.lock
+++ b/python.lock
@@ -1581,7 +1581,7 @@
             {
               "algorithm": "sha256",
               "hash": "c6345a0b8188ff915bf33cce6e445d9fd3738faa32640ca6267bd1344249ef6f",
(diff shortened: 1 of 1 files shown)


studioego:
I found  backend.ai-oven's hiredis wheels. good!
If the PR is merged, I close the issue https://github.com/lablup/backend.ai/issues/609

achimnol (approved):

PR #621, opened 2022-08-01
doc: Add README to the changes directory

achimnol:
Let's add more concrete and detailed guide on how to write good commit messages, news fragments, and PR titles.


inureyes:
@achimnol It will be better if example sentences are added. :D

PR #624, opened 2022-08-02
feat: allow passing blocklist to compute plugin ctx read from local config

fregataa:
~~correct wrong type of `compute_plugins` allowlist and blocklist which is list in toml files.~~
~~additionally, pass blocklist argument to compute_plugin_ctx when it inits.~~

since type of `allowlist` is correct as `set`, this PR changes to feature for allow plugin ctx to get blocklist.


kyujin-cho (changes requested):
Please define corresponding schema of `block-compute-plugins` in `src.ai.backend.config.agent_local_config_iv`, so that Trafaret can parse and validate value.

configs/agent/sample.toml
```diff
@@ -56,11 +56,12 @@ scaling-group = "default"
 # The port number for aiomonitor.
 # aiomonitor-port = 50200
 
-# Explicitly set allowed compute plugins to avoid conflicts of mock &
+# Explicitly set allowed(or blocked) compute plugins to avoid conflicts of mock &
 # open-source CUDA plugins in the mono-repo source installation.
 # Regardless of this config, CPUPlugin and MemoryPlugin are always enabled.
 # example: ["ai.backend.accelerator.cuda_mock"]
 # allow-compute-plugins = []
+# block-compute-plugins = []
```
kyujin-cho, line 64: Please fix agent config validation to raise validation error when duplicate value is in both arrays.

kyujin-cho (changes requested):

kyujin-cho (approved):
Looks good to me.

changes/624.feature.md
```diff
@@ -0,0 +1 @@
+Allow reading compute plugin blocklists from toml file.
```
kyujin-cho: How about specifying "which" configuration to edit, instead of just toml?
```suggestion
Allow reading compute plugin blocklists from agent local configuration.
```

kyujin-cho (changes requested):
Just a quick change!

kyujin-cho (approved):
LGTM!

PR #626, opened 2022-08-04
refactor: Change decrypt function to middleware

### LINKED ISSUE #598: Refactor `decrypt_payload()` in webserver as a middleware
The manager's optional payload encryption API should allow encryption of arbitrary API requests by the client's choice.

Currently the webserver detects and tries decryption in `login_handler()` only.
Let's refactor this logic out as a middleware so that it could be applied to any request handlers.

KangJuSeong:
resolve #598

### DIFF
--- a/changes/626.feature.md
+++ b/changes/626.feature.md
@@ -1,1 +1,1 @@
-Refactoring `decrypt_payload()` as a middleware so that applied to other request handlers
+Refactor `decrypt_payload()` as a middleware so that applied to `web_handler()` and `login_handler()`
--- a/changes/626.feature.md
+++ b/changes/626.feature.md
@@ -1,0 +1,1 @@
+Refactoring `decrypt_payload()` as a middleware so that applied to other request handlers
--- a/src/ai/backend/web/proxy.py
+++ b/src/ai/backend/web/proxy.py
@@ -133,7 +133,7 @@ async def decrypt_payload(request: web.Request, handler) -> web.StreamResponse:
     secure_context = request_headers.get("X-BackendAI-Encoded", None)
     if secure_context:
         if not request.content:
-            request.app["config"]["payload"] = ""
+            request["payload"] = ""
             return await handler(request)
         config = request.app["config"]
         scheme = config["service"]["force_endpoint_protocol"]
@@ -141,18 +141,20 @@ async def decrypt_payload(request: web.Request, handler) -> web.StreamResponse:
             scheme = request.scheme
         api_endpoint = f"{scheme}://{request.host}"
         payload = await request.text()
-        iv, real_payload = payload.split(":")  # Extract initial vector and actual payload
-        key = (base64.b64encode(api_endpoint.encode("ascii")).decode() + iv + iv)[0:32]
-        crypt = AES.new(bytes(key, encoding="utf8"), AES.MODE_CBC, bytes(iv, encoding="utf8"))
+        initial_vector, real_payload = payload.split(":")
+        key = (base64.b64encode(api_endpoint.encode("ascii")).decode() + initial_vector * 2)[0:32]
+        crypt = AES.new(
+            bytes(key, encoding="utf8"), AES.MODE_CBC, bytes(initial_vector, encoding="utf8")
+        )
         b64p = base64.b64decode(real_payload)
         dec = unpad(crypt.decrypt(bytes(b64p)), 16)
         result = dec.decode("UTF-8")
         if not result:
-            request.app["config"]["payload"] = {}
+            request["payload"] = {}
         else:
-            request.app["config"]["payload"] = json.loads(result)
+            request["payload"] = json.loads(result)
     else:
-        request.app["config"]["creds"] = request.json()
+        request["payload"] = ""
     return await handler(request)
 
 
@@ -179,7 +181,7 @@ async def web_handler(request, *, is_anonymous=False) -> web.StreamResponse:
             secure_context = request_headers.get("X-BackendAI-Encoded", None)
             decrypted_payload_length = 0
             if secure_context:
-                payload = request.app["config"]["payload"]
+                payload = request["payload"]
                 decrypted_payload_length = len(payload)
             else:
                 payload = request.content
--- a/src/ai/backend/web/proxy.py
+++ b/src/ai/backend/web/proxy.py
@@ -127,22 +127,33 @@ class WebSocketProxy:
             await self.up_conn.close()
 
 
-async def decrypt_payload(request: web.Request) -> str:
-    if not request.content:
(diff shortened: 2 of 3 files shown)


src/ai/backend/web/proxy.py
```diff
@@ -127,22 +127,33 @@ async def close_upstream(self) -> None:
...
+        if not request.content:
+            request.app["config"]["payload"] = ""
+            return await handler(request)
+        config = request.app["config"]
+        scheme = config["service"]["force_endpoint_protocol"]
+        if scheme is None:
+            scheme = request.scheme
+        api_endpoint = f"{scheme}://{request.host}"
+        payload = await request.text()
+        iv, real_payload = payload.split(":")  # Extract initial vector and actual payload
```
rapsealk: I wonder if it would be right to place an additional comment here.
Maybe we can set the variable name to `initial_vector` rather than `iv` and remove the comment?

changes/626.feature.md
```diff
@@ -0,0 +1 @@
+Refactoring `decrypt_payload()` as a middleware so that applied to other request handlers
```
rapsealk: I suggest to write in the imperative form. Maybe we can say a verb "Refactor" than "Refactoring".
Also, it would be great if it is clearly specified which is applied to other request handlers.

src/ai/backend/web/proxy.py
KangJuSeong: Should I replace the existing `iv` with `initial_vector`?

src/ai/backend/web/proxy.py
```diff
@@ -127,22 +127,33 @@ async def close_upstream(self) -> None:
...
+        key = (base64.b64encode(api_endpoint.encode("ascii")).decode() + iv + iv)[0:32]
+        crypt = AES.new(bytes(key, encoding="utf8"), AES.MODE_CBC, bytes(iv, encoding="utf8"))
+        b64p = base64.b64decode(real_payload)
+        dec = unpad(crypt.decrypt(bytes(b64p)), 16)
+        result = dec.decode("UTF-8")
+        if not result:
+            request.app["config"]["payload"] = {}
```
kyujin-cho: `request.app` is not suitable for storing request-specific data; it's designed for storing global variables (ref: https://docs.aiohttp.org/en/stable/web_advanced.html#application-s-config). Request data can be stored directly in `request`.
```suggestion
            request["payload"] = {}
```

src/ai/backend/web/proxy.py
```diff
@@ -127,22 +127,33 @@ async def close_upstream(self) -> None:
...
+        else:
+            request.app["config"]["payload"] = json.loads(result)
+    else:
+        request.app["config"]["creds"] = request.json()
```
kyujin-cho: `request.json()` returns Coroutine, not `dict` (ref: https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.BaseRequest.json). You need to add `await` to retrieve actual `dict` data.
```suggestion
        request["creds"] = await request.json()
```

kyujin-cho (changes requested):

src/ai/backend/web/proxy.py
rapsealk: In my personal opinion, yes, if the comment is only for describing what `iv` is.

src/ai/backend/web/proxy.py
KangJuSeong: When I add await I get that error.
image

changes/626.feature.md
```diff
@@ -0,0 +1 @@
+refactor(webserver): `decrypt_payload()` as a middleware so that applied to `web_handler()` and `login_handler()`
```
rapsealk: This format is for commit messages, not news fragments. Please refer changelog/README.md
(https://github.com/lablup/backend.ai/blob/main/changes/README.md#in-the-middle-news-fragments)

KangJuSeong: I wasn't familiar with the format for commit messages and changeslog. thank you for telling me.

src/ai/backend/web/proxy.py
KangJuSeong: When the variable name of iv is changed, it is thought that the variable name is too long when used in the next line. Would it be okay?

rapsealk: That is why you chose to write in abbr and it is quite reasonable.
How about replacing `initial_vector + initial_vector` to `initial_vector * 2`?

changes/626.feature.md
rapsealk: Furthermore, it is not recommended to use **force** push. Don't worry, all the commits will be squashed and rebased when this PR is merged into the main branch.

src/ai/backend/web/proxy.py
KangJuSeong: I didn't think about that part because I used it without changing the logic of the existing decrypt_payload function. I think it would be better to change it like you said.

src/ai/backend/web/proxy.py
kyujin-cho: Looks like it's not issue of `await`; it's more like unnoticed bug being revealed. What `aiohttp` complaing about is that request body is not a json. Current implementation makes `decrypt_payload()` middleware to do its job on all the requests coming into webserver. And since your middleware is trying to treat all request bodies (which aren't empty) as json, when a request which isn't json (maybe graphql requests?), boom. It'll fail like this.    
IMO there is no reason to try to decode & re-encode request body as json. Let's just keep it as `str`.     
Also, dict key which holds decrypted payload should be some other self-explanatory name, not `creds`. 
```suggestion
        request["payload"] = request
```

src/ai/backend/web/proxy.py
```diff
@@ -127,22 +127,35 @@ async def close_upstream(self) -> None:
...
+        crypt = AES.new(
+            bytes(key, encoding="utf8"), AES.MODE_CBC, bytes(initial_vector, encoding="utf8")
+        )
+        b64p = base64.b64decode(real_payload)
+        dec = unpad(crypt.decrypt(bytes(b64p)), 16)
+        result = dec.decode("UTF-8")
+        if not result:
+            request["payload"] = {}
+        else:
+            request["payload"] = json.loads(result)
```
kyujin-cho: Hmm. As we aren't confident that request body always will be type of json, I think it's better to just payload as-is.

```suggestion
            request["payload"] = ""
        else:
            request["payload"] = result
```

kyujin-cho (changes requested):

kyujin-cho (approved):
Looks good to me.

achimnol (approved):
Great work!

Issue #265, opened 2021-05-31
Track "oom killed"


achimnol:
Let's add "oom-killed" status info for OOM-killed kernels (in addition to "self-terminated").

Q: `docker inspect` against an exited container reveals the `State.OOMKilled` field. Could we get the same information from the Docker event stream?

achimnol:
We first need to migrate lablup/backend.ai-agent#304 to the mono-repo, preserving the original git commit author information.

PR #629, opened 2022-08-05
feat: Add OOM status info

Yaminyam:
https://github.com/lablup/backend.ai-agent/pull/304
I took the work of the pr and proceeded with it.

#265 The contents of the work before the monorepo integration have been transferred to the monorepo.
The logic works as it is, and only the parts where errors occur have been resolved.

### DIFF
--- a/changes/629.feature.md
+++ b/changes/629.feature.md
@@ -1,0 +1,1 @@
+Put 'out-of-memory' as the termination reason when we detect the containers OOM death via our docker event subscriber.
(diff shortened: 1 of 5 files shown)


achimnol:
Hmmm... Pausing when OOM occurs seems not to be a valid strategy to let users resolve the problem by themselves, because the OOM state won't be changed but just frozen.  (The review & discussion for this point has been missed due to other urgent issues.)

Could we split the introduction of PAUSE/UNPAUSE lifecycle event addition as a separate PR?
In this PR, let's just make the agent to recognize OOM events and report them to the manager, and make the manager to store those as `status_info` and `status_data`.

achimnol:
Replaced with #1373.

PR #632, opened 2022-08-07
setup: Remove an explicit dependency on `grpcio` from requirements

rapsealk:
In this PR, dependencies on `grpcio` and `grpcio-tools` are removed from the `requirements` field from `python.lock`.
There was a mistake on the previous PR; I added `grpcio` and `grpcio-tools` to requirements.txt explicitly before generating lock files. However, I notified that we used an implicit dependency from `etcetra` and therefore `grpcio`-families were not included directly in the `requirements` field.

Reference
- lablup/backend.ai#620

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -145,8 +145,7 @@ jobs:
         else
           BASE_REF="HEAD~1"
         fi
-        ./pants test ::
-      # --changed-since=$BASE_REF --changed-dependees=transitive
+        ./pants test --changed-since=$BASE_REF --changed-dependees=transitive
     - name: Upload pants log
       uses: actions/upload-artifact@v3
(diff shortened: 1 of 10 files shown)


kyujin-cho (approved):

achimnol (changes requested):
Where is the update for `requirements.txt`?

rapsealk:
@achimnol `grpcio` and `grpcio-tools` were never been in `requirements.txt`. I temporarily put `grpcio` and `grpcio-tools` in `requirements.txt` and ran `pants generate-lockfiles`, but didn't update them on commit.

achimnol:
Removal of explicit dependency to grpcio 1.47.0 is already done in #636.
Closing as already-done.

PR #637, opened 2022-08-08
feat: Generalize an interface for watchdog timer on DistributedLock

rapsealk:
In this PR, watchdog-related properties are added to `AbstractDistributedLock`.
- `enqueue_watchdog_task()` and `dequeue_watchdog_task()` are for abstraction; these methods provide capsulization of actual task handling process.
- `_watchdog_timer()` is an abstract method; each class inherits `AbstractDistributedLock` can define its own specified lock-related logics in this method, and this will be used as an actual handler when watchdog timer rings.

### DIFF
--- a/src/ai/backend/common/lock.py
+++ b/src/ai/backend/common/lock.py
@@ -75,7 +75,7 @@ class FileLock(AbstractDistributedLock):
         if self._file is not None:
             self._debug = False
             self.release()
-            log.debug("file lock implicitly released: {}", self._path)
+            log.debug(f"{self.__class__.__name__} implicitly released: {self._path}")
 
     async def acquire(self) -> None:
         assert self._file is None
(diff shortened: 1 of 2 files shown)


src/ai/backend/common/lock.py
```diff
@@ -44,6 +45,20 @@ async def __aenter__(self) -> Any:
     async def __aexit__(self, *exc_info) -> Optional[bool]:
         raise NotImplementedError
 
+    def enqueue_watchdog_task(self, ttl: float) -> None:
```
achimnol: Hm... Why did you choose "enqueue" and "dequeue" for naming these methods?

src/ai/backend/common/lock.py
```diff
@@ -44,6 +45,20 @@ async def __aenter__(self) -> Any:
...
+
+    def dequeue_watchdog_task(self) -> None:
+        if task := self._watchdog_task:
+            if not task.done() and not task.cancelled():
+                task.cancel()
+        self._watchdog_task = None
+
+    @abc.abstractmethod
+    async def _watchdog_timer(self, ttl: float) -> None:
+        await asyncio.sleep(ttl)
```
achimnol, line 60: The concrete method of setting a TTL may not be represented as tasks depending on the backend implementation. In such case, we should be able to customize how we cancel the watchdog in subclasses. How could we abstract this as well?

src/ai/backend/common/lock.py
rapsealk: I couldn't find an appropriate pair. How about `register_watchdog_task()` and `unregister_watchdog_task()`?

src/ai/backend/common/lock.py
fregataa: maybe we can pass a cancellation function as an argument to watchdog timer?

HyeockJinKim:
In my opinion, ABC, which is the Distributed Lock, should only perform the locking action, and watchdog tasks should be implemented in a different layer.
I think that the existing GlobalTimer, and now the Leader Election Task, are somewhat replacing this role, so I will close this PR.

Issue #641, opened 2022-08-10
Elaborate "no-available-instances" messaging


achimnol:
Usually "no-available-instances" error happens when there is no contiguous chunk of resources in a single agent to schedule a session container, *after* passing the resource policy checks in the scheduler predicates.  This problem is also known as "resource fragmentation".

We are getting reports from users that this is still confusing because they feel it like an error in the system.

There are two approaches possible:
- Better visualize the current agent resource capacity (#640)
- Elaborate the failure message (it's a failure, not an error!)

This PR handles the latter.

* [x] **manager**: Raise `InstanceNotAvailable` exceptions with more concrete cases and messages, and preserve the message in `kernels.status_data`
* [ ] **webui**: When the session's `status_info` is `no-available-instances`, take the concrete message from `status_data` and display it to the user. (Probably similarly to the predicate check result hover panel?)

adrysn:
I think Manager part (lablup/backend.ai#643) is working well. We need to display the newly added `scheduler['msg']` response field in the Web-UI. For reference, I'll append the entire response content when `InstanceNotAvailable` is raised from the server:

```json
{
  "scheduler": {
    "msg": "Could not find a contiguous resource region in any agent big enough to host the session (d536c3f1-aa41-4d70-8656-9fa57424b74d)",
    "retries": 3,
    "last_try": "2022-08-20T15:49:37.971944+00:00",
    "failed_predicates": [],
    "passed_predicates": [
      {
        "name": "reserved_time"
      },
      {
        "name": "concurrency"
      },
      {
        "name": "dependencies"
      },
      {
        "name": "keypair_resource_limit"
      },
      {
        "name": "user_group_resource_limit"
      },
      {
        "name": "domain_resource_limit"
      }
    ]
  }
}
```

PR #643, opened 2022-08-10
feat: Elaborate messaging and logging of `InstanceNotAvailable` error

achimnol:
refs #641

Now it reports `InstanceNotAvailable` error with detailed `extra_msg` when:
- For manually-assigned agents, the target agent does not have enough resource to host the kernel or the session.
- The scheduler could not assign the agent.
- There is no agents compatible with the per-kernel image architecture.

It reports `GenericBadRequest` if the manually assigned agent is an unknown one or there are multiple architectures detected for a single-node multi-container session.

Also the dispatcher stores `extra_msg` as the `scheduler.msg` JSON field in the `status_data` column.


adrysn (approved):
LGTM.

PR #644, opened 2022-08-10
ci: Fix workflows for forks

achimnol:


achimnol:
This is replaced with #1356.
Let's revisit the issue when we see similar things in the future.

PR #647, opened 2022-08-12
feat: Add service `configure` commands

### LINKED ISSUE #267: CLI-based initial configuration command for manager and agent when installed as packages
Currently installing BackendAI using PyPI packages does not give any hints for the required TOML configuration, though our GitHub repositories contain `config/sample.toml` file with detailed comments.

Let's add `backend.ai mgr configure` and `backend.ai ag configure` with interactive prompts so that users can easily start up a new cluster interactively. Those commands should:
 - Take necessary inputs from the user (e.g., the address and credentials for etcd, Redis, and PostgreSQL server / the network storage mount path(s) / etc.)
 - Validate the user inputs (e.g., check the input formats, test connections)
 - Generate the TOML file
 - (manager)
   - Populate the initial etcd config
   - Generate `alembic.ini` and run alembic migration

As the next step, we could extend their functionality to edit existing configuration interactively as well, maybe simply opening up a text editor against the TOML file and `alembic.ini`.

achimnol:
resolves #267

* [x] `mgr configure`
  - [ ] Include the template TOML in the packaging
* [x] `ag configure`
  - [ ] Include the template TOML in the packaging
* [ ] `web configure`
* [ ] `storage configure`


achimnol:
This is deprecated in favor of the TUI-based installer.

PR #648, opened 2022-08-12
feat: Add env var for service ports, preopen ports

Yaminyam:
#544 
An environment variable has been added to check the port inside the container.


src/ai/backend/manager/registry.py
```diff
@@ -1289,6 +1289,15 @@ async def start_session(
                     binding.kernel.cluster_hostname for binding in kernel_agent_bindings
                 ),
                 "BACKENDAI_ACCESS_KEY": scheduled_session.access_key,
+                "BACKENDAI_SERVICE_PORTS": ",".join(
+                    image_infos[binding.kernel.image_ref].labels.get("ai.backend.service-ports")
+                    for binding in kernel_agent_bindings
```
achimnol: Is it necessary to iterate over `kernel_agent_bindings`?
I think we should put the *per-kernel* environment variable instead of session-level ones here.
https://github.com/lablup/backend.ai/blob/3aedf4149d3e43914d073205195ad9f405474f7e/src/ai/backend/manager/registry.py#L1503

achimnol (changes requested):

src/ai/backend/manager/registry.py
Yaminyam: also added a service port value to Kernel Env.

achimnol: The `BACKENDAI_SERVICE_PORTS` should be *moved* there (as you did).
Please remove it from here as well.

PR #653, opened 2022-08-16
feat: Add `status_history` to the result of `get_container_stats_for_period()`

Sujin-Kim1:
I added `status_history` to the query field of `get_container_stats_for_period`. This is to check when resources are allocated by adding `scheduled_at` (one of the session history) to the Usage Logs page of the control panel.

**Related issue:**
- https://github.com/lablup/backend.ai-control-panel/issues/299
- https://github.com/lablup/backend.ai/issues/469

### DIFF
--- a/changes/653.feature.md
+++ b/changes/653.feature.md
@@ -1,0 +1,1 @@
+Add `status_history` to the query field of `get_container_stats_for_period` to know when the status of the session within a given period has changed.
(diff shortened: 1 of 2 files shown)


achimnol (commented):
Shouldn't we also include the field in the return value?

Sujin-Kim1:
> 

ah..! you're right. I'll apply it right away.

achimnol (approved):

PR #654, opened 2022-08-16
feat: Add new abstract methods to accelerator plugins

kyujin-cho:
This PR defines new API for accelerator plugin interface (`AbstractComputePlugin`). These interfaces will help future compute interfaces requiring additional networks (e.g. RDMA) work as intended.
### Newly introduced interfaces
- `generate_mounts(source: Path, device_alloc: Mapping) -> List[MountInfo]`
- `get_docker_networks(device_alloc: Mapping) -> List[str]`


achimnol (approved):
LGTM.
We also need to update and release our enterprise-only plugins as well.

inureyes:
@kyujin-cho Plugins should be also updated following this PR. @adrysn FYI

PR #655, opened 2022-08-16
refactor: Fix a missing change of numeric exitcode to an enum value

studioego:
Following the #486 As mentioned in the issue and #559  Merged PR,
I found a missing item about changing numeric exit code to enum value.

As a result, I modify missing item about changing numeric exitcode to enum value.


achimnol (approved):
Good catch!

achimnol:
I'm gonna skip the changelog because it's more like a typo fix and has no significant impact to the users.

rapsealk (approved):
LGTM!

PR #656, opened 2022-08-16
feat: return one result value using SQL union query in `query_accessible_vfolders`.

Sujin-Kim1:
- migrate from https://github.com/lablup/backend.ai-manager/pull/573 to mono-repo  

  > To successfully support the pagination query, the vfolder list is united into one.
Previously, `query_accessible_vfolders()` was implemented with Python's list concat, but now only one query containing all vfolder lists is returned.


achimnol:
This is superseded by #1386 and #1387.

PR #658, opened 2022-08-17
ci: pr auto author assign

Yaminyam:
Most prs are assigned by the author, but there are some that are missing.
Automate this through github actions

It does not work in the following situations.
- Someone is already assigned as an assignee
- The author is a bot


inureyes (approved):
LGTM

achimnol (approved):

PR #670, opened 2022-08-22
feat: Automate force-termination of hanging sessions

### LINKED ISSUE #603: Automatic force-termination of hanging sessions
Thanks to #412 (PR: #480), now we can determine how long time has passed after a session has entered PREPARING or TERMINATING status.

Let's add a periodic check to scan such sessions and automatically force-terminate (with concrete `reason` logs) them.

rapsealk:
This PR resolves #603.
To automate force-termination of hanging sessions, it will be processed in the order below:
1. Query kernels
  - where `status` is `PREPARING` or `TERMINATING`
  - and time has passed more than given `datetime.timedelta` since `status_history[status]`
2. Force termination and cleanup.


src/ai/backend/manager/server.py
```diff
@@ -454,6 +454,84 @@ async def monitoring_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
...
+            _ = await asyncio.gather(
+                *[
+                    asyncio.create_task(
+                        root_ctx.registry.destroy_session(
+                            functools.partial(
+                                root_ctx.registry.get_session_by_session_id,
+                                session_id,
+                            ),
+                            forced=True,
+                            reason="automatic-force-termination",
```
adrysn: How about distinguishing the termination reason for the `PREPARING` and `TERMINATING` sessions? This will help to know in which status the system is failing easily. Perhaps, `hang-preparing` / `hang-terminating`?

adrysn (commented):
How about making this feature optional with a config flag (turning it on by default will be okay)? Since this is an invasive feature that force-terminates `PREPARING` and `TERMINATING` sessions without prior notification to users, I think some sites may not want to activate this feature.

src/ai/backend/manager/server.py
rapsealk: Thank you for the feedback! Now it provides a distinguishable reason for each preparing/terminating session. Please check the updated code below:
https://github.com/lablup/backend.ai/blob/9470b6907659124b8fbdeb3535d93cff8173a805/src/ai/backend/manager/server.py#L508-L521

rapsealk:
@adrysn I also placed a additional config flag `force-terminate-hanging-sessions` in `manager.toml`. The default value is `true` and an user can set it `false` to disable the feature.
https://github.com/lablup/backend.ai/blob/ccea5c63de2a9696fb378b13dda003075a800dfe/configs/manager/halfstack.toml#L16-L24
https://github.com/lablup/backend.ai/blob/ccea5c63de2a9696fb378b13dda003075a800dfe/src/ai/backend/manager/server.py#L688-L689

rapsealk:
Maybe we can add a config flag for setting an interval to sleep?

configs/manager/halfstack.toml
```diff
@@ -21,6 +21,7 @@ service-addr = { host = "0.0.0.0", port = 8081 }
 ssl-enabled = false
 #ssl-cert = "/etc/backend.ai/ssl/apiserver-fullchain.pem"   # env: BACKNED_SSL_CERT
 #ssl-privkey = "/etc/backend.ai/ssl/apiserver-privkey.pem"  # env: BACKNED_SSL_KEY
+enabled-session-managing-contexts = { "PREPARING" = "1h", "TERMINATING" = "30m" }
```
kyujin-cho: Please write down briefly about this value (what this value means, available values, ...) on `configs/manager/sample.toml` and define scheme on manager's config.py

kyujin-cho (changes requested):

configs/manager/halfstack.toml
rapsealk: @kyujin-cho Thanks for the feedback!
https://github.com/lablup/backend.ai/blob/632ca06e46f966f24f87805b66118207e8bd4eda/configs/manager/sample.toml#L116-L119

configs/manager/halfstack.toml
```diff
@@ -30,6 +30,8 @@ disabled-plugins = []
 
 hide-agents = true
 
+max-session-hangtime = { "PREPARING" = "1h", "TERMINATING" = "30m" }
+
```
kyujin-cho: Characteristics of this config is more suitable with Etcd (shared) config, rather than `manager.toml`. Let's refactor the code so that we can find this config at the shared config.

kyujin-cho: Also, name of this config looks a little bit vague; how about renaming it?
```suggestion
session-hang-toleration-threshold = { "PREPARING" = "1h", "TERMINATING" = "30m" }

```

kyujin-cho (changes requested):

achimnol:
I'd like to add this feature! Please merge the latest main branch.

.gitignore
achimnol, line 1: Since we now store the whole etcd configuration right after running `install-dev.sh`, I don't think we need to add another one here. It is just sufficient to `etcd put-json` the given example file.

src/ai/backend/common/events.py
```diff
@@ -200,6 +200,7 @@ class KernelLifecycleEventReason(str, enum.Enum):
     EXEC_TIMEOUT = "exec-timeout"
     FAILED_TO_START = "failed-to-start"
     FORCE_TERMINATED = "force-terminated"
+    HANGTIME_EXCEEDED = "hangtime-exceeded"
```
achimnol: How about renaming this to `HANG_TIMEOUT`, just like `IDLE_TIMEOUT`?

src/ai/backend/manager/server.py
```diff
@@ -490,6 +491,113 @@ async def monitoring_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
         await ectx.cleanup()
 
 
+@actxmgr
+async def hanging_sessions_scanner_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
```
achimnol: It's a minor thing, but we could just name it `hanging_session_scanner_ctx` (no plural for "session" in the middle of the composite word).

src/ai/backend/manager/server.py
```diff
@@ -490,6 +491,113 @@ async def monitoring_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
...
+            case "m":
+                threshold = timedelta(minutes=int(threshold_fmt[:-1]))
+            case "s":
+                threshold = timedelta(seconds=int(threshold_fmt[:-1]))
+            case _:
+                log.error(
+                    f'Skip "{status}:{threshold_fmt}" as it is an invalid argument. Supported'
+                    ' postfixes: "h", "m", "s"'
+                )
+                continue
```
achimnol: I think we could reuse `ai.backend.common.validators.TimeDuration`.
If we use it, it is no longer required to assert `isinstance(threshold_fmt, str)` as well.

src/ai/backend/manager/server.py
```diff
@@ -490,6 +491,113 @@ async def monitoring_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
...
+                ]
+            )
+
+            await asyncio.sleep(threshold.seconds)
+
+    raw_session_config = await root_ctx.shared_config.etcd.get_prefix("config/session")
+    assert isinstance(raw_session_config, Mapping)
+
+    hang_tolerance_threshold_dict = raw_session_config.get("hang-tolerance-threshold", {})
+    assert isinstance(hang_tolerance_threshold_dict, Mapping)
```
achimnol: We can avoid this boilerplate by splitting out the inner `t.Dict()` trafaret in manager/config.py as a separate top-level variable to reuse it in both the whole `shared_config` input validation and a typed deserializer when reading the etcd on the fly.

src/ai/backend/manager/server.py
```diff
@@ -490,6 +491,113 @@ async def monitoring_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
...
+                _force_terminate_hanging_sessions(status=session_status, threshold=threshold)
+            )
+        )
+
+    yield
+
+    for task in session_force_termination_tasks:
+        if not task.done():
+            if not task.cancelled():
+                task.cancel()
```
achimnol: We may just use `if not task.done()` only here.

src/ai/backend/manager/server.py
```diff
@@ -490,6 +491,113 @@ async def monitoring_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
...
+            _ = await asyncio.gather(
+                *[
+                    asyncio.create_task(
+                        root_ctx.registry.destroy_session(session, forced=True, reason=reason),
+                    )
+                    for session in sessions
+                ]
+            )
+
+            await asyncio.sleep(threshold.seconds)
```
achimnol: The check interval should be shorter than the timeout threshold, but it shouldn't be too frequent to avoid database overheads.
There is no golden rule here, but I think at least we need to use an interval less than *the half* of the threshold to achieve a sufficient frequency. (i.e., The wavelength to get a desired resolution should be the half of the target resolution. In the same way, usually we set the heartbeat timeout as a double of heartbeat interval.)

achimnol (changes requested):

src/ai/backend/common/events.py
rapsealk: Resolved: 59622bdb036db0d035ffa4b130f3e541dfc94f98

.gitignore
rapsealk: Resolved: 39f77f528718bc57454dce825b5688880b203a5e

src/ai/backend/manager/server.py
rapsealk: Resolved: 8be780b530fb0c5b268ea00df67a74cac8200284

src/ai/backend/manager/server.py
rapsealk: Resolved: 20d878b5c4f2eadb538eeef4f577a011f7fb5887, c666327407e85fc27dc5dabd6cf2e426b318af3d, 9604d3c9a12970745df5c558d9c05064a4541eb2

src/ai/backend/manager/server.py
rapsealk: Resolved: c666327407e85fc27dc5dabd6cf2e426b318af3d

src/ai/backend/manager/server.py
rapsealk: Resolved: f2ebf3d4e8277618731f8c01e967ef1de283d087

src/ai/backend/manager/server.py
rapsealk: Resolved: 395ea3b7d946e743422a630b71c413707cb786d9, 9604d3c9a12970745df5c558d9c05064a4541eb2, 76c311cb70c00ea3b5cbd225e6e65af2b390f4e1
I set the interval to 40% of the given threshold.

src/ai/backend/manager/server.py
```diff
@@ -495,6 +496,119 @@ async def monitoring_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
...
+                            session, forced=True, reason=KernelLifecycleEventReason.HANG_TIMEOUT
+                        ),
+                    )
+                    for session in sessions
+                ],
+                return_exceptions=True,
+            )
+            for result_or_exception in results_and_exceptions:
+                if isinstance(result_or_exception, (BaseException, Exception)):
+                    log.error(result_or_exception)
```
achimnol: * The logger interface requires an explicit message format and arguments. This is an discouraged pattern.
* We need to pass `exc_info` kwarg to specify the exception object details if we're not inside an `except` block scope.
* Also we need to consider picklability of the errors, so it might be better to wrap them with `repr()`.

src/ai/backend/manager/server.py
```diff
@@ -495,6 +496,119 @@ async def monitoring_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
...
+
+            await asyncio.sleep(heuristic_interval)
```
achimnol: It would be better to use aiotools timer to split out what it does on each interval and how we invoke it upon each interval. (e.g., separation of the unhandled exception from the timer loop)

achimnol (changes requested):

src/ai/backend/manager/server.py
```diff
@@ -495,6 +496,122 @@ async def monitoring_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
...
+            continue
+
+        if isinstance(threshold, relativedelta):  # months, years
+            if months := threshold.months:
+                threshold = timedelta(seconds=months * 30 * 24 * 60 * 60)
+            elif years := threshold.years:
+                threshold = timedelta(seconds=years * 365 * 24 * 60 * 60)
+            else:
+                log.warning("Skipping {}:{}...", status, threshold)
+                continue
```
achimnol: `relativedelta` considers leap years and the difference of the number of days in months, while `timedelta` is a single absolute value.
So, we should keep the original form of `relativedelta` until we do the actual comparison instead of converting it to `total_seconds`.

See https://stackoverflow.com/a/49021582.
Could we remove this and do the comparison inside the timer function?

achimnol (changes requested):

src/ai/backend/manager/server.py
achimnol: Note that if the user's intention is to specify an absolute value of time difference, still one can achieve the goal using a single-character suffixes when specifying the threshold value (e.g., `365d` instead of `1yr`).

src/ai/backend/manager/server.py
```diff
@@ -495,6 +496,116 @@ async def monitoring_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
...
+
+    async def _force_terminate_hanging_sessions(
+        status: SessionStatus,
+        threshold: relativedelta | timedelta,
+        interval: float,
+    ) -> None:
+        try:
+            sessions = await _fetch_hanging_sessions(root_ctx.db, status, threshold)
+        except asyncio.CancelledError:
+            pass
```
achimnol: Shouldn’t we just raise up the cancellation because the outer context already uses `suppress`.

achimnol (approved):

PR #672, opened 2022-08-22
feat: Add pause, unpause type to kernel lifecycle

Yaminyam:
https://github.com/lablup/backend.ai/pull/629 isolated from pr
Added pause and unpause to kernel lifecycle types.


fregataa (commented):
good!
how about let pause_kernel and unpause_kernel methods in kubernetes.agent module raise `NotImplementedError`?

src/ai/backend/agent/agent.py
```diff
@@ -927,6 +928,39 @@ async def _handle_clean_event(self, ev: ContainerLifecycleEvent) -> None:
...
+    ) -> None:
+        pass
+
+    @abstractmethod
+    async def unpause_kernel(
+        self,
+        kernel_id: KernelId,
+        container_id: Optional[ContainerId],
+    ) -> None:
+        pass
```
kyujin-cho, line 945: Please implement these methods on Kubernetes backend also. If you don't have any knowledge about Kubernetes, it's okay to let implementations as blank and create separate issue stating we have to implement it later.

kyujin-cho (approved):

src/ai/backend/agent/agent.py
kyujin-cho, line 962: Please implement these methods on Kubernetes backend also. If you don't have any knowledge about Kubernetes, it's okay to let implementations as blank and create separate issue stating we have to implement it later.

kyujin-cho (changes requested):

s4shantanu (approved):

src/ai/backend/agent/docker/agent.py
```diff
@@ -1501,18 +1548,35 @@ async def handle_action_die(kernel_id: KernelId, evdata: Mapping[str, Any]) -> N
...
+                                        self.docker_ptask_group.create_task(
+                                            handle_action_start(kernel_id, evdata),
+                                        )
                                     )
-                                )
-                            elif evdata["Action"] == "die":
-                                await asyncio.shield(
-                                    self.docker_ptask_group.create_task(
-                                        handle_action_die(kernel_id, evdata),
+                                    break
```
achimnol: I remember that Python's match syntax does not 'fall-through' to the subsequent case blocks. Please remove these unnecessary `break` statements, unless you are intending to exit the entire subscription loop!

achimnol (changes requested):

s4shantanu (commented):
Looks good to me!!

achimnol:
As #629 is superseded by #1373, I'm closing this as well.
Currently it's not practical to forcibly pause a container (i.e., excluding the container processes from the Linux scheduler) as this will make it unresponsive at all while the container still occupies the agent resources.

PR #675, opened 2022-08-22
ci: automatic db migration labeling

Yaminyam:
Added auto labeling ci for efficient pr management
If there is a change in the migration file that is created when the db schema is changed, create a require:db-migration tag
It is thought that auto labeling will be possible by checking the change of manager, agent, etc. composed of monorepo using this workflow.

In the future, I think it would be nice to be able to import the label assigned to the issue linked to Development.


achimnol (approved):
Thanks!

Issue #619, opened 2022-08-01
Add GitHub Actions workflow to preview documentation builds


achimnol:
Using a new GitHub Actions workflow which is executed only when a PR in this repository has the `docs` label, let's build the previews of both English/Korean versions of the documentation and serve it on a publicly accessible static hosting using auto-generated branch-based names, just like Netlify.

* [x] Explore possibility to use Netlify for building our Sphinx docs. &rarr; ❌ it does not support Python 3.10 runtime. (#669 is merged and reverted)
* [x] Configure an authentication token for [the RTD API](https://docs.readthedocs.io/en/stable/api/v3.html) and register it to the repository secrets. @achimnol 
  - Use the `RTD_AUTH_TOKEN` repository secret.
* [x] Write a new GitHub workflow (`docs-preivew.yml`) @soheeeeP 
  - Automatically **activate** and **deactivate** the branch if the PR has the "docs" label and when it is open/reopen/closed/labeled/unlabeled.
    - The project IDs to use: `sorna` (English ver.) and `sorna-ko` (Korean ver.)
  - When the branch is activated, we may need to manually **trigger the first build**.  Once activated, subsequent builds will be automatically picked up by RTD.

Some references:
- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
- https://docs.github.com/en/actions/learn-github-actions/expressions
- https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow
- https://docs.readthedocs.io/en/stable/api/v3.html#build-triggering
- https://docs.readthedocs.io/en/stable/api/v3.html#project-update
  - Where to update: <img width="238" alt="image" src="https://user-images.githubusercontent.com/555156/185850889-4c29913d-bafc-4808-b976-9fc3565e3630.png">
  - https://docs.readthedocs.io/en/stable/api/v3.html#project-details
- https://github.com/lablup/backend.ai/blob/3a255783f30c70d95d65e564a2c182f35c321629/.github/workflows/default.yml#L7
- https://github.com/lablup/backend.ai/blob/3a255783f30c70d95d65e564a2c182f35c321629/.github/workflows/timeline-check.yml#L5

soheeeeP:
Can I try this issue?

achimnol:
Tried to implement this using Netlify but unfortunately Netlify does not support Python 3.10 runtime.

[...]

PR #687, opened 2022-08-25
ci: Add `docs-preview.yml` to preview documentation builds when PR contains `docs` label

soheeeeP:
issue #619 

- Description
   - Add workflow to preview documentation builds which is executed only when a PR contains `docs` label.

- Process
    1) Update a project and fetch activate/deactivate versions
    2) Activate PR branch: update a `version_slug` of its project
    3) Trigger a new build for the `version_slug`
    4) Deactivate PR branch: update a `version_slug` of its project

- Question
   - I have a question about RTD build process.
      I wonder if when the English version of the documentation is successfully build/served,
the process for Korean version documentation will be automatically triggered or it should be build/served manually.


soheeeeP:
Could anyone assign `docs` label to this PR?

soheeeeP:
@achimnol 
RTD api requests fails with `401(unauthorized)` status code.
It seems something is wrong with repository secret(`RTD_AUTH_TOKEN`).

image

achimnol:
Hm... I think secret access is not allowed from PRs by forks.
https://github.com/openmainframeproject/feilong/issues/367
Let me think how we could workaround it.

achimnol:
Currently this is being handled by the updated RTD webhooks.

PR #688, opened 2022-08-25
feat: Support the asyncio debug mode and apply aiomonitor-ng

achimnol:
If [the asyncio debug mode](https://docs.python.org/3/library/asyncio-dev.html#asyncio-debug-mode) is enabled, it additionally keeps track of the stack trace of where the tasks are created though this behavior is *undocumented*.

Chaining of full task creation/cancellation tracebacks is handled in [our modified aiomonitor](https://github.com/achimnol/aiomonitor-ng).

This PR also fixes missing sample configurations for aiomonitor ports and changes the default value of them to avoid overlapping of the aiomonitor ports in manager and agent when they run in a same host (e.g., development setups).


fregataa:
the port number of aiomonitors changes due to better management of ports?

adrysn (approved):
It works in my environment. LGTM.

achimnol:
I will make separate PRs for future updates of aiomonitor-ng.

PR #689, opened 2022-08-25
fix: unable to login from Web-UI since the decoded payload is not JSON loaded

adrysn:
This is a follow-up fix for https://github.com/lablup/backend.ai/pull/626.

### DIFF
--- a/src/ai/backend/web/server.py
+++ b/src/ai/backend/web/server.py
@@ -187,7 +187,12 @@ async def login_handler(request: web.Request) -> web.Response:
             ),
             content_type="application/problem+json",
         )
-    creds = request["payload"]
+    raw_creds = json.loads(request["payload"])
+    try:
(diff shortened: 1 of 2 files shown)


src/ai/backend/web/server.py
```diff
@@ -187,7 +187,12 @@ async def login_handler(request: web.Request) -> web.Response:
             ),
             content_type="application/problem+json",
         )
-    creds = request["payload"]
+    raw_creds = json.loads(request["payload"])
+    try:
+        creds = json.loads(raw_creds)
```
rapsealk: It seems that `json.loads()` is applied twice. Maybe `creds = json.loads(request["payload"])` should come inside the try block?

adrysn: Yes. It's my mistake. Thanks.

achimnol (approved):

PR #695, opened 2022-08-30
feat: IP based client restriction

### LINKED ISSUE #693: Per-user IP-based access blocking
Let's implement a per-user IP-based access blocking as follows.

* [ ] **webserver**
  - Attach the client's IP address when proxying API requests to the manager.
* [ ] **manager**
  - Add a new nullable `allowed_client_ip` column to store a set of IP addresses or prefixes to `keypairs`.
  - Update the `auth_middleware`:
    - to fetch the client IP address from webserver-attached header with a fallback to the client IP address retrieved from the aiohttp request object
    - to take account into the the `allowed_client_ip` field when checking the determined client IP address
* [ ] Let's write test cases to match arbitrary IP addresses and prefixes.

fregataa:
resolve #693 

Get `X-BackendAI-IP` header from request and validate the client IP.
The client IP should be stored in `User` table by admin.
Wild-card pattern and CIDR are allowed to input, customized `CIDR` column type converts it into valid CIDR type.
```
# Admin's input
allowed_client_ip = [192.0.0.1/32, 192.32.0.*]

# User table
allowed_client_ip = [192.0.0.1/32, 192.32.0.1/24]
```


fregataa:
Should I change `allowed_client_ip` field to JSON field and use `sql_json_merge()` to update the field?
or just simply *append* when it is updated?

src/ai/backend/manager/api/auth.py
```diff
@@ -499,6 +499,19 @@ async def _pipe_builder(r: Redis) -> RedisPipeline:
             },
             "is_admin": row["keypairs_is_admin"],
         }
+
+        def validate_ip(user: Mapping[str, Any]):
+            raw_ip = user.get("allowed_client_ip", None)
+            if not raw_ip or raw_ip is None:
+                # raw_ip is None or "" - empty string
+                return
+            _ip = request.headers.get("X-BackendAI-IP")
```
fregataa: Should `AuthorizationFailed` be raised if the header does not have client ip?

src/ai/backend/manager/api/auth.py
```diff
@@ -499,6 +500,21 @@ async def _pipe_builder(r: Redis) -> RedisPipeline:
...
+            if not allowed_client_ip or allowed_client_ip is None:
+                # raw_ip is None or [] - empty list
+                return
+            assert isinstance(allowed_client_ip, list)
+            raw_client_addr: str | None = request.headers.get("X-BackendAI-IP") or request.remote
+            if raw_client_addr is None:
+                raise AuthorizationFailed("Not allowed IP address")
+            client_addr: ReadableCIDR = ReadableCIDR(raw_client_addr)
+            if client_addr not in allowed_client_ip:
+                raise AuthorizationFailed(f"'{client_addr}' is not allowed IP address")
```
kyujin-cho: Is there a specific reason for this function to be defined inside `auth_middleware()`? I think it's safe to split out `validate()` to a global function.

src/ai/backend/client/cli/admin/user.py
```diff
@@ -42,6 +45,7 @@ def info(ctx: CLIContext, email: str) -> None:
         user_fields["created_at"],
         user_fields["domain_name"],
         user_fields["groups"],
+        user_fields["allowed_client_ip"],
```
kyujin-cho, line 48: In my understanding, `allowed_client_ip` have multiple CIDRs. How about adding plural to this variable (and all other keywords referencing allowed CIDRs)?

src/ai/backend/common/types.py
```diff
@@ -335,6 +338,72 @@ def __str__(self) -> str:
...
+        str_val = str(value)
+        if "*" in str_val:
+            _ip, _, given_cidr = str_val.partition("/")
+            filtered = _ip.replace("*", "0")
+            if given_cidr:
+                return ip_network(f"{filtered}/{given_cidr}", strict=False)
+            octets = _ip.split(".")
+            cidr = octets.index("*") * 8
+            return ip_network(f"{filtered}/{cidr}", strict=False)
+        return ip_network(str_val, strict=False)
```
kyujin-cho: This function is more like a class method than an instance method...

src/ai/backend/common/types.py
```diff
@@ -335,6 +338,72 @@ def __str__(self) -> str:
...
+    Convert wild-card based IP address into CIDR.
+
+    e.g)
+    192.10.*.* -> 192.10.0.0/16
+    """
+
+    def __init__(self, address: str | None) -> None:
+        self._address: _Address | None = (
+            cast(_Address, self._convert_to_cidr(address)) if address is not None else None
+        )
```
kyujin-cho: How about defining type of `_address` at class defintion? 
```suggestion
class ReadableCIDR(Generic[_Address]):
    """
    Convert wild-card based IP address into CIDR.

    e.g)
    192.10.*.* -> 192.10.0.0/16
    """
    _address: _Address | None

    def __init__(self, address: str | None) -> None:
        self._address = (
            cast(_Address, self._convert_to_cidr(address)) if address is not None else None
        )
```

src/ai/backend/common/types.py
```diff
@@ -335,6 +338,72 @@ def __str__(self) -> str:
...
+        def ip_to_list(val: str) -> list[str]:
+            _ip, _, cidr = val.partition("/")
+            octets = _ip.split(".")
+            return octets if not cidr else octets[: int(cidr) // 8]
+
+        ip1, ip2 = ip_to_list(str(self)), ip_to_list(str(other))
+        for octet1, octet2 in zip(ip1, ip2):
+            if octet1 != octet2:
+                return False
+        return True
```
kyujin-cho: `ReadableCIDR` stores given network info as a `IPv4Network` or `IPv6Network`. `_BaseNetwork`, superclass of both classes, is also capable of being compared as it defines custom comparator (`__gt__`, `__lt__`, `__eq__`, ...). How about using these instead of redefining?

src/ai/backend/manager/api/auth.py
```diff
@@ -499,6 +500,21 @@ async def _pipe_builder(r: Redis) -> RedisPipeline:
...
+
+        validate_ip(auth_result["user"])
```
kyujin-cho: Is there a reason to let this function as an inner function? I think it's okay to split out this function as a separate global function.

src/ai/backend/manager/models/keypair.py
```diff
@@ -513,7 +513,7 @@ async def mutate(
         root,
         info: graphene.ResolveInfo,
         access_key: AccessKey,
-        props: ModifyUserInput,
+        props: ModifyKeyPairInput,
```
kyujin-cho: Oh, kudos for catching this error. But it's a bit off-topic to this PR. Can you create a new pull request solely referencing this issue?

src/ai/backend/manager/api/auth.py
```diff
@@ -499,6 +500,21 @@ async def _pipe_builder(r: Redis) -> RedisPipeline:
...
             "is_admin": row["keypairs_is_admin"],
         }
+
+        def validate_ip(user: Mapping[str, Any]):
+            allowed_client_ip = user.get("allowed_client_ip", None)
+            if not allowed_client_ip or allowed_client_ip is None:
+                # raw_ip is None or [] - empty list
+                return
+            assert isinstance(allowed_client_ip, list)
+            raw_client_addr: str | None = request.headers.get("X-BackendAI-IP") or request.remote
```
kyujin-cho: In current implementation, when user who has IP restriction enabled tries to access Backend.AI API through webserver, will always fail because request.remote value will always point to webserver's IP address. Thankfully webserver already provides original IP address under `X-Forwarded-For` header, we can benefit by using that header to determine original value. Also there seems to be no any other logic adding this header to request on client, so I think it's okay to remove parsing `X-BackendAI-IP` header.

kyujin-cho (changes requested):

src/ai/backend/manager/api/auth.py
fregataa: I thought this is a part of authentication.
maybe `validate()` can be used in other module?

src/ai/backend/client/cli/admin/user.py
fregataa: I didnt add plural because @achimnol suggested that the word ips can be confusing (since ip is a abbreviation).
But after I saw the field name `groups`, ips seems okay.

src/ai/backend/client/request.py
```diff
@@ -267,6 +268,10 @@ def _sign(
                 pass
         else:
             raise ValueError("unsupported endpoint type")
+        bai_ip = self.session.aiohttp_session.headers.get("X-BackendAI-IP") or socket.gethostbyname(
+            socket.gethostname()
+        )
+        self.headers.update({"X-BackendAI-IP": bai_ip})
```
kyujin-cho: As manager can extract client's IP address automatically (via request.remote and `X-Forwarded-For` header) I think it's useless to bundle arbitrary header representing originated IP address.
```suggestion
        bai_ip = self.session.aiohttp_session.headers.get("X-BackendAI-IP") or socket.gethostbyname(
            socket.gethostname()
        )
        self.headers.update({"X-BackendAI-IP": bai_ip})
```

src/ai/backend/manager/api/auth.py
```diff
@@ -375,6 +377,27 @@ async def sign_request(sign_method: str, request: web.Request, secret_key: str)
...
+    allowed_client_ip = user.get("allowed_client_ip", None)
+    if not allowed_client_ip or allowed_client_ip is None:
+        # allowed_client_ip is None or [] - empty list
+        return
+    assert isinstance(allowed_client_ip, list)
+    raw_client_addr: str | None = (
+        request.headers.get("X-BackendAI-IP")
+        or request.headers.get("X-Forwarded-For")
+        or request.remote
+    )
```
kyujin-cho: Same here. I think we won't need `X-BackendAI-IP` header anymore.
```suggestion
        request.headers.get("X-Forwarded-For")
        or request.remote
    )
```

kyujin-cho (changes requested):

kyujin-cho (approved):
LGTM.

PR #697, opened 2022-08-31
feat: Leader election based GlobalTimer

rapsealk:
This PR resolves 415.

Using [aioraft-ng](https://github.com/lablup/aioraft-ng), this PR removes a dependency on`DistributedLock` from `GlobalTimer` and ensures that `GlobalTimer` ticks only run on the leader process.


rapsealk:
When running a cluster with 5 nodes, it tolerates up to 2 failures.

https://user-images.githubusercontent.com/14137676/188081676-2f0d7204-a936-455c-b7a7-b258b95e3ec8.mov

rapsealk:
Need to find a way to transfer `RootContext` and `PrivateContext` objects to the leader process using Redis.

src/ai/backend/manager/server.py
```diff
@@ -281,6 +292,83 @@ async def manager_status_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
     yield
 
 
+@actxmgr
+async def leader_election_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
+    configuration = root_ctx.local_config["manager"]["raft-configuration"]
+    raft_id = configuration[root_ctx.pidx]
+    grpc_port = int(raft_id.split(":")[-1])
```
fregataa: let's use [partition()](https://docs.python.org/3/library/stdtypes.html#str.partition) instead of `split()` if you want to split the string into 2.

src/ai/backend/manager/server.py
```diff
@@ -281,6 +292,83 @@ async def manager_status_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
     yield
 
 
+@actxmgr
+async def leader_election_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
+    configuration = root_ctx.local_config["manager"]["raft-configuration"]
+    raft_id = configuration[root_ctx.pidx]
```
fregataa: Is `pidx` different from every manager process although they are running on a different machine?
and how about handling index error here? just in case of wrong configuration.

src/ai/backend/common/distributed.py
```diff
@@ -69,3 +72,100 @@ async def leave(self) -> None:
...
+    async def join(self) -> None:
+        self._stopped = False
+        if self._tick_task is None:
+            self._tick_task = asyncio.create_task(self.generate_tick())
+
+    async def leave(self) -> None:
+        self._stopped = True
+        await asyncio.sleep(0)
+        if (task := self._tick_task) is None:
+            return
```
fregataa, line 166: assigned `task` is not used.

src/ai/backend/manager/server.py
```diff
@@ -281,6 +292,83 @@ async def manager_status_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
...
+        client=client,
+        configuration=filtered_configuration,
+        on_state_changed=_on_state_changed,
+    )
+
+    _cleanup_coroutines: List[Coroutine] = []
+
+    _ = (
+        asyncio.create_task(
+            GrpcRaftServer.run(
```
fregataa: I am confused here because `GrpcRaftServer.run()` method is not a classmethod or staticmethod.
https://github.com/lablup/aioraft-ng/blob/main/aioraft/server.py#L28
Is this intended?

src/ai/backend/manager/server.py
```diff
@@ -281,6 +292,83 @@ async def manager_status_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
...
+        if root_ctx.node_id == event.node_id:
+            global_timer_ids.append(event.timer_id)
+            if reigns:
+                await root_ctx.event_producer.produce_event(
+                    GlobalTimerJoinEvent(node_id=root_ctx.node_id, timer_id=event.timer_id)
+                )
+
+    global_timer_created_handler = root_ctx.event_dispatcher.subscribe(
+        GlobalTimerCreatedEvent, None, _on_global_timer_created
+    )
```
fregataa, line 313: are global timers registered by an event handler because the timers should produce event as soon as registered if the manager is leader?

fregataa (changes requested):
Please update `manager_local_config_iv` in `manager.config.py` file to validate `raft-configuration` configuration.

And it seems that Raft and manager components are communicating through event handlers and a callback function which changes raft state. Is it correct?

src/ai/backend/manager/server.py
rapsealk: Thanks for introducing `str.partition()`!
However, I found that it would be better to use `HostPortPair` type, which has a `port` field, for "raft-configuration".  Maybe I can have a chance to use `str.partition()` in other tasks. 😢

About the raft component and manager, it is using a callback for getting notified of its raft status change.
Event handlers are used between a manager server and its `GlobalTimer` objects.
```
Raft ---(callback)---> Manager <---(event)---> GlobalTimer(a)
                                      ㄴ-----> GlobalTimer(b)
```

src/ai/backend/manager/server.py
rapsealk: Good point. It would be better to set its address manually in `manager.toml`.

rapsealk: But we still have to distinguish each process on the same machine since they share the same `manager.toml`.
How about separating them into `local-raft-configuration` and `remote-raft-configuration`? Each process can concatenate them before passing them to a `Raft` instance.

fregataa: hmm, then each manager-runner machine has different `remote-raft-configuration` configuration.
It seems error prone cause we need to believe users set a number of different configuration correctly.
I think that we need to have *fixed* machine id or ip addresses of machines.

src/ai/backend/manager/server.py
rapsealk: Ouch my mistake 😢 I forgot to update it after changing `GrpcRaftServer.run()` to an instance method (It was originally a staticmethod in the first version).
Thank you for the comment!

src/ai/backend/common/distributed.py
rapsealk: It is used in right below the line (L153-159):
```python
self._tick_task = None
if not task.done():
    try:
        task.cancel()
        await task
    except asyncio.CancelledError:
        pass
```

The purpose of this is to remove a reference from `self._tick_task` before assigning `None` and move it to local level.
https://github.com/lablup/backend.ai/pull/697/files/2105ff832f650165a47243b4b357c54cd2416f81#diff-5ea3f42af9928fe8f8d81bbfb581f39b93592ae3922b6da07e036d7b97a629a9R153-R159

achimnol:
[aioraft-ng](https://github.com/lablup/aioraft-ng) is superseded by [raftify](https://github.com/lablup/raftify).

PR #698, opened 2022-08-31
feat: Add audit logs

hydroxyde:
This pull request refers to: lablup/backend.ai-internal#17 and #148.

This is the implementation of the audit log feature's model which keeps records of creation, modification and deletion, etc., of users, keypairs and vfolders. 
### Audit Logs model
Each record in the audit log is as follow:

**The subject** (i.e., who did the action?)
It contains: the API access key, user UUID, user email (with respective types, all indexed)

**The action**

- Action type: CREATE | CHANGE | DELETE

- Action data: (JSONB field)

{"before": {"data1": "val1", ...}, "after": {"data1": "val2", ...}}
For passwords and secret keys, we do not record the original/hash values.
CREATE: record only "after", leaving "before" as null
CHANGE: record both "before", "after" and include only the changed data.
DELETE: record only "before", leaving "after" as null

**The target object**
It contains the values of users.uuid, keypairs.access_key or vfolder.name

**The date/time** of action (indexed, this should be the default descending order key of audit log queries)

> Screenshot 2022-05-04 at 17 29 36

### Periodic removal of audit logs
This command extends https://github.com/lablup/backend.ai-manager/pull/498: 
- Option to remove audit logs, time based
   - e.g., remove all logs older than 3 months, 1 year, etc...

Screenshot 2022-07-03 at 18 13 45

**Extended options are as follow:**

- **-t, --table [kernels|audit_logs]** : Which table to run the operation on. Options available are kernels  and audit_logs. If not set or set to kernels, it will run the operation on kernels table. If set to audit_logs, it will run on audit_logs table (default: kernels). 
- **--target** : For audit_logs you can chose which audit target you would like the operation to run on. You can provide user.uuid, keypairs.access_key or vfolder_name. If not set, it will run on all targets.

### List
This command list audit logs of a user by user_id

Screenshot 2022-05-04 at 18 30 52

Screenshot 2022-05-09 at 21 09 03

This PR uses revamp from https://github.com/lablup/backend.ai-manager/pull/449

- `filter` uses Python like query expression with SQL-specific operators
  - e.g., `email contains "@lablup.com"`

- Use of mini-language for expressing query ordering `order` as the list of multiple columns with ascending/descending order marks
  - e.g., `-created_date`


hydroxyde:
@achimnol I didn't notice I was not working on audit_logs branch. I made a new branch/PR which includes  target object type, vfolder.id instead of name and black style.

achimnol:
Does this replace #502?

hydroxyde:
Yes, it replaces #502 sorry for the confusion.

src/ai/backend/client/func/audit_logs.py
```diff
@@ -0,0 +1,115 @@
...
+
+class AuditLog(BaseFunction):
+    """
+    Provides management of audit logs.
+    """
+
+    @api_function
+    @classmethod
+    async def list(cls, list_all=False):
+        rqst = Request("GET", "/audit_logs")
```
kyujin-cho, line 37: Hmm, where's handler definition for this API located at? I only can find graphql implementations.

src/ai/backend/manager/api/vfolder.py
```diff
@@ -521,8 +540,56 @@ async def delete_by_id(request: web.Request, params: Any) -> web.Response:
...
+            "action": "DELETE",
+            "target_type": "vfolder",
+            "target": prev_data["id"],
+        }
+        try:
+            insert_auditlog_query = sa.insert(audit_logs).values(auditlog_data)
+            await conn.execute(insert_auditlog_query)
+        except sa.exc.DataError:
+            raise InvalidAPIParameters
+
```
kyujin-cho: Those two audit logs look identical. Is there a reason creating duplicate logs?

src/ai/backend/manager/models/audit_logs.py
```diff
@@ -0,0 +1,240 @@
...
+
+audit_logs = sa.Table(
+    "audit_logs",
+    metadata,
+    sa.Column("user_id", sa.String(length=256), index=True),
+    sa.Column("access_key", sa.String(length=20), index=True),
+    sa.Column("email", sa.String(length=64), index=True),
+    sa.Column(
+        "action",
+        sa.Enum("CREATE", "CHANGE", "DELETE", name="auditlogs_action", create_type=False),
```
kyujin-cho, line 39: Please define `action` enum as python enum type, so that we can handle these values more pythonic- way. Ref: https://github.com/lablup/backend.ai/blob/main/src/ai/backend/manager/models/user.py#L74

src/ai/backend/manager/models/audit_logs.py
```diff
@@ -0,0 +1,240 @@
...
+        "target_type",
+        sa.Enum("user", "keypairs", "vfolder", name="auditlogs_targettype", create_type=False),
+        index=True,
+    ),
+    sa.Column("target", sa.String(length=64), index=True),
+    sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.func.now(), index=True),
+)
+
+
+class AuditLog(graphene.ObjectType):
```
kyujin-cho, line 53: There should be a security consideration to prevent unprivileged users from accessing audit log of others. Let's refactor this object to raise error if the requester is non-admin but variables referencing each user (`user_id`, `access_key` and `email`) does not match requester's identity.

src/ai/backend/manager/models/gql.py
```diff
@@ -839,6 +890,24 @@ async def resolve_user_from_uuid(
...
+    @staticmethod
+    @scoped_query(autofill_user=True, user_key="email")
+    async def resolve_user_from_email(
+        executor: AsyncioExecutor,
+        info: graphene.ResolveInfo,
+        *,
+        domain_name: str = None,
+        email: str = None,
+    ) -> User:
+        print("into resolve")
```
kyujin-cho, line 947: This looks like only for debugging.
```suggestion
```

src/ai/backend/manager/models/audit_logs.py
```diff
@@ -0,0 +1,240 @@
...
+    sa.Column("email", sa.String(length=64), index=True),
+    sa.Column(
+        "action",
+        sa.Enum("CREATE", "CHANGE", "DELETE", name="auditlogs_action", create_type=False),
+        index=True,
+    ),
+    sa.Column("data", pgsql.JSONB()),
+    sa.Column(
+        "target_type",
+        sa.Enum("user", "keypairs", "vfolder", name="auditlogs_targettype", create_type=False),
```
kyujin-cho, line 45: Same here.

kyujin-cho (changes requested):

src/ai/backend/manager/api/vfolder.py
hydroxyde: Yes, that was duplicate. thank you

src/ai/backend/client/func/user.py
```diff
@@ -242,6 +251,39 @@ async def detail_by_uuid(
...
+            query = textwrap.dedent("""\
+                query($email: String!) {
+                    user_from_email(email: $email) {$fields}
+                }
+            """)
+        query = query.replace("$fields", " ".join(f.field_ref for f in fields))
+        print("query{0}", query)
+        variables = {"email": email}
+        data = await api_session.get().Admin._query(query, variables if email is not None else None)
+        print("data from func", data["user_from_email"])
```
achimnol, line 284: The functional APIs in the client should avoid using direct stdout outputs. Please remove debugging-purpose print() calls and/or use the standard logging so that the caller can control the logs.

src/ai/backend/manager/cli/__main__.py
achimnol, line 1: The implementation should be rewritten following the latest style (see `_clear_redis_history()` and `_clear_terminated_sessions()` for reference), as we no longer depend on `psycopg2` but only `asyncpg`.

I think it would be better to merge `entry` and `target` CLI arguments into a single `target` expression argument to specify the following options (example):
- `all`
- `stats`
- `kernels`
- `audit-logs`
  - A subset of audit logs using a dot-separated category expression like `audit-logs.user`, `audit-logs.vfolder`, etc.
- a comma-separated combination of above keywords

src/ai/backend/manager/models/alembic/versions/b5866ae16fc1_update_auditlogs_table.py
achimnol, line 1: You will need to generate a merge revision using `./py -m alembic merge <rev1> <rev2>` where `<rev1>` is the current revision `b5866ae16fc1` and `<rev2>` is the current head revision (you may check it out using `./py -m alembic heads`).

achimnol (changes requested):
A batch of comments after merging the latest main branch.

kyujin-cho:
Closing in favor of #1822

PR #645, opened 2022-08-10
feat: Add `AgentSummary` and list queries for normal users to get agent info

fregataa:
refs #640 

Allows non-admin users to query the agent list with a limited view.

PR #699, opened 2022-08-31
feat: Do not return AgentSummary if hide-agents is true

### LINKED ISSUE #640: Implement an option to display the agent list to non-admin users
We are getting frequent requests from the enterprise customers and open-source users to check the agent resource capacity even as a normal user, not super-admins, when the tenants are semi-trusted users in the same organization.

Currently, the manager has `[manager].hide-agents` boolean flag in `manager.toml` to include the agent information in session, kernel, and installed image GraphQL queries.

Let's extend this feature as follows:
* [x] **manager**: ~~Allow querying `Agent` and `AgentList` GraphQL objects from normal users if `hide-agents` flag is set `false`.~~ Add a new `AgentSummary` and `AgentSummaryList` GraphQL queries to proivde a limited view when `hide-agent` flag is set `false`. (#645)
* [ ] **web**: Add a new option `hide-agents` (default: `true`) to `webserver.conf` and include it in the rendered `config.toml`
* **webui**: (https://github.com/lablup/backend.ai-webui/pull/1439)
  - [x] Read the new `hide-agents` option (assuming `true` if not specified/present) from `/config.toml`, and if set `false`, show the `Agent Summary` menu in the sidebar.
  - [x] Add the `Agent Summary` list page using the `AgentSummaryList` query.

fregataa:
do not return AgentSummary and AgentSummaryList gql objects if `hide-agents` is true.

Follow-up of #645.
Resolves #640.

### DIFF
--- a/src/ai/backend/manager/models/gql.py
+++ b/src/ai/backend/manager/models/gql.py
@@ -558,6 +558,9 @@ class Queries(graphene.ObjectType):
         scaling_group: str | None = None,
     ) -> AgentSummary:
         ctx: GraphQueryContext = info.context
+        if ctx.local_config["manager"]["hide-agents"]:
+            raise ObjectNotFound(object_name="agent")
+
         loader = ctx.dataloader_manager.get_loader(
             ctx,
             "Agent",
@@ -583,8 +586,12 @@ class Queries(graphene.ObjectType):
         scaling_group: str | None = None,
         status: str | None = None,
     ) -> AgentSummaryList:
+        ctx: GraphQueryContext = info.context
+        if ctx.local_config["manager"]["hide-agents"]:
+            raise ObjectNotFound(object_name="agent")
+
         total_count = await AgentSummary.load_count(
-            info.context,
+            ctx,
             access_key=access_key,
             scaling_group=scaling_group,
             domain_name=domain_name,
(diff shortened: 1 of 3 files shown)


adrysn:
I'm not sure, but wouldn't it be better to raise 404 NotFound error instead of returning the empty results to hide the queries themselves?

fregataa:
> I'm not sure, but wouldn't it be better to raise 404 NotFound error instead of returning the empty results to hide the queries themselves?

that would be nice cause handling query at model interface by reading config seems inappropriate.

adrysn (approved):
Tested and works for me.

src/ai/backend/manager/models/gql.py
```diff
@@ -583,16 +586,20 @@ async def resolve_agent_summary_list(
         scaling_group: str | None = None,
         status: str | None = None,
     ) -> AgentSummaryList:
+        ctx: GraphQueryContext = info.context
+        if ctx.local_config["manager"]["hide-agents"]:
```
kyujin-cho, line 590: Please add corresponding config definition on manager's local config scheme, `manager_local_config_iv`.

kyujin-cho (changes requested):

kyujin-cho (approved):

Issue #678, opened 2022-08-23
Let's group string-literal `KernelEvent.reason` values into a manageable format


rapsealk:
**Is your feature request related to a problem? Please describe.**
Each **Kernel-Lifecycle-EventArgs** class has a `reason` field to briefly describe the source of the lifecycle event. For example, when terminating a kernel, a `KernelTerminatingEvent` object is published with the `reason` field inherited from its parent class `KernelTerminationEventArgs`.
https://github.com/lablup/backend.ai/blob/c33063ec7f567eb9f36821a19348d910d2e27c22/src/ai/backend/common/events.py#L194-L197
https://github.com/lablup/backend.ai/blob/c33063ec7f567eb9f36821a19348d910d2e27c22/src/ai/backend/common/events.py#L266-L269

There are several values used as a default reason such as `user-requested` or `self-terminated`, and the problem is that it is hard to manage since they are just string literals.
https://github.com/lablup/backend.ai/blob/c33063ec7f567eb9f36821a19348d910d2e27c22/src/ai/backend/manager/api/session.py#L1491-L1498

**Describe the solution you'd like**
I suggest managing each of the default reason values as a constant variable. Having an enum class can be an option, but it will cause a problem in understanding raw data from a database that stores enum values in index-like formats(0, 1, ...).

**Describe alternatives you've considered**
-

**Additional context**
Here is an example:
```python
class KernelLifecycleEventReason:
    USER_REQUESTED = "user-requested"
    FORCE_TERMINATED = "force-terminated"
    ...

await root_ctx.registry.destroy_session(
    functools.partial( 
        root_ctx.registry.get_session_by_session_id, 
        event.session_id, 
    ), 
    forced=False, 
    reason=event.reason or KernelLifeCycleEventReason.KILLED_BY_EVENT, 
) 
```

qkoo0833:
Dear rapsealk,
I'm finding string-literal `KernelEvent.reason` for this issue.
Meanwhile, I saw several reasons about SessionEvent, AgentEvent, etc.
If I make a PR for this issue, Could I include reasons for not only KernelEvent, but other Event?

rapsealk:
@qkoo0833 Oh, I just found your comment now. First of all, thanks for your support. I think it would be better to separate those tasks into different issues for each.

qkoo0833:
@rapsealk Thank you for your review. I made additional commit that include only `KernelEvent.reason` and add change log.

qkoo0833:
[...]

PR #700, opened 2022-09-01

qkoo0833:
1. Add class `KernelLifecycleEventReason` in common/events.py to manage `kernelEvent.reason`

#678

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -38,10 +38,14 @@ jobs:
           key: lfs-${{ hashFiles('.lfs-assets-id') }}
     - name: Git LFS Pull
       run: git lfs pull
+    - name: Extract Python version from pants.toml
+      run: |
+        PYTHON_VERSION=$(grep -oP '(?<=CPython==)([^"]+)' pants.toml)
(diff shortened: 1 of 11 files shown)


changes/700.feature.md
```diff
@@ -0,0 +1 @@
+feat: Add class KernelLifecycleEventReason to manage KernelEvent.reason
```
rapsealk: This is not a valid style for changelogs. Please refer [changes/README](https://github.com/lablup/backend.ai/blob/main/changes/README.md).

rapsealk (changes requested):

src/ai/backend/common/events.py
```diff
@@ -529,6 +529,15 @@ def __call__(
 ]
 
 
+class KernelLifecycleEventReason:
```
kyujin-cho: The recommended way to group multiple values as a single category is to use `enum.Enum`. 
https://github.com/lablup/backend.ai/blob/bede3c289c0a2b3620d07e061d986f52898654d9/src/ai/backend/common/types.py#L232-L235. Please update this class to inherit `enum.Enum`. Don't forget to update type definition on code which accepts `KernelLifecycleEventReason` also!

kyujin-cho (changes requested):
Thank you for your contribution! Here are some points which can be enhanced.

PR #702, opened 2022-09-02
setup: Remove dependency to psycopg

achimnol:
`psycopg2-binary` is making some difficulties for installation on Apple Silicon Macs.
Let's remove the dependency to it.

### Included in this PR
* Apply [alembic's official guide to use it with asyncio](https://alembic.sqlalchemy.org/en/latest/cookbook.html#using-asyncio-with-alembic) by wrapping the alembic APIs with `await connection.run_sync()`
  - Use a boolean context variable to make both `alembic` CLI commands and programmatic invocation of alembic APIs to work seamlessly.
* Replace all occurrences of "postgresql://" with "postgresql+asyncpg://"
  - Remove the no longer used `scripts/agents/ci` directory and its contents

### Deprecated
* The official `psycopg-binary` referenced from `psycopg[binary]` does not provide official M1 (macos-arm64) wheel builds.  So I've built our own wheel at: https://dist.backend.ai/pypi/simple/psycopg-binary
  - It is built using [`cibuildwheel`](https://github.com/pypa/cibuildwheel), so `libpq.so` / `libpq.dylib` is **_embedded inside_** the packaged wheel. No longer headaches.


achimnol:
I've confirmed the API updates in `mgr clear-histry` work:
image

studioego (approved):
I tried to install Backend.AI's the PR #702 
Successfully installed and there is no problem.
image

studioego:
> I tried to install Backend.AI's the PR #702 Successfully installed and there is no problem. 
Also, I checked to install Backend.AI on Fedora core 35 (x86_64) with Nvidia GTX 1650Ti.
There is also no problem.
image

kyujin-cho (approved):

PR #707, opened 2022-09-05
fix: Remove a migrated decode operation when decrypting payload

lizable:
This PR resolves [webui#1453](https://github.com/lablup/backend.ai-webui/issues/1453).


src/ai/backend/web/proxy.py
```diff
@@ -131,31 +131,31 @@ async def close_upstream(self) -> None:
...
+            )
+            b64p = base64.b64decode(real_payload)
+            request["payload"] = unpad(crypt.decrypt(bytes(b64p)), 16)
         else:
-            request["payload"] = result
-    else:
-        request["payload"] = ""
-    return await handler(request)
+            request["payload"] = ""
+        return await handler(request)
```
lizable: @achimnol I found that there's only one explicit description about exception that can occur during decoding operation([`base64.b64decode(...)`](https://docs.python.org/3.10/library/base64.html#base64.b64decode)), not in encoding operation.
I also add a general Exception class just in case, so that It would handle another exception (not from `b64decode`) when something went wrong during executing decrypt_payload with the request text.

achimnol: It is best to determine the exact possible list of exceptions.
Or, as an alternative, you could check if they use library-level parent exception class or Python's intrinsic exceptions such as `ValueError`.
In either cases, it is highly recommended to avoid using the catch-call `Exception` class.

src/ai/backend/web/proxy.py
```diff
@@ -131,31 +131,31 @@ async def close_upstream(self) -> None:
...
+    except (Exception, base64.binascii.Error):
```
```python
>>> import binascii
>>> binascii.Error.__mro__
(<class 'binascii.Error'>, <class 'ValueError'>, <class 'Exception'>, <class 'BaseException'>, <class 'object'>)
>>>
```
I think it would be best to just use `ValueError`.

src/ai/backend/web/proxy.py
```diff
@@ -131,31 +131,31 @@ async def close_upstream(self) -> None:
...
+            if not request.content:
+                request["payload"] = ""
+                return await handler(request)
+            config = request.app["config"]
+            scheme = config["service"]["force_endpoint_protocol"]
+            if scheme is None:
+                scheme = request.scheme
+            api_endpoint = f"{scheme}://{request.host}"
+            payload = await request.text()
+            initial_vector, real_payload = payload.split(":")
```
achimnol: I'd like to move the beginning of `try` block to here, as the core parsing logic starts here.

src/ai/backend/web/proxy.py
achimnol: Also, even though `base64` internally refers `binascii`, it is not recommended to indirectly refer a module imported inside another module.  If you really want to use `binascii.Error`, just use `binascii.Error` instead of `base64.binascii.Error`.
In this case, the type error occurs because `binascii` is not included in `base64.__all__` which serves a similar role as `export` in Javascript.

achimnol (changes requested):

lizable:
I think we need to make a decision since we are now facing the need for exception handling across the component, which seems a bit too big to handle here.
Also, I think it's better to delegate this issue to someone who contributes to Backend.AI Core.
I'm afraid that I cannot handle this in prompt because I'm getting much more diverse issues than before.

Anyway, there are two options and of course, both of them have pros and cons.
- Option 1.
Wrap up with minimal changes (with exception handling) and leave the changing stuff as a new issue.

- Option 2.
Change the architecture first (as a separate issue and branch.), and then apply the exception to this issue.

It would be great to decide on one of the options above and make a progress. What do you think @achimnol?

achimnol (approved):

PR #708, opened 2022-09-05
fix: Commit db session depends on the option of session context

fregataa:
remove `session.commit()` in `models.utils.begin_session()` method which commits DB implicitly.
Implicit commit is helpful when we use Session API such as `Session.add()` or `Session.delete()`, but confusing and unnecessary when we use `Connection.begin()` with SQLAlchemy 2.0 style query.
https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#library-level-but-not-driver-level-autocommit-removed-from-both-core-and-orm

```
async with db.begin_session() as session:
    await session.execute(
        update(User).where(User.name == "lee").values(email="hello@world.com")
    )

    # the commit below is unnecessary.
    # await session.commit()
```

Add an `do_commit` option to `begin_session()` method which let session commits after the session context is returned.


src/ai/backend/manager/models/utils.py
```diff
@@ -93,7 +93,6 @@ async def begin_session(self) -> AsyncIterator[SASession]:
             session = SASession(bind=conn)
             try:
                 yield session
```
kyujin-cho, line 95: Instead of calling `session.commit()` every time in the async with context, how about adding keyword parameter to `begin_session()` which determines to do the commit or not? 
```suggestion
                yield session
                if do_commit:
                    await session.commit()
```

src/ai/backend/manager/models/image.py
```diff
@@ -848,6 +851,7 @@ async def mutate(
                     sa.delete(ImageAliasRow).where(ImageAliasRow.image_id.in_(image_ids))
                 )
                 await session.execute(sa.delete(ImageRow).where(ImageRow.registry == registry))
+                await session.commit()
```
kyujin-cho: Well... There's a `session.execute()` call here, so I guess we don't need to call additional `session.commit()`. Is it correct?

kyujin-cho (changes requested):

src/ai/backend/manager/models/image.py
fregataa: my mistake. `session.commit()` is not needed here.

achimnol:
No longer required.

PR #709, opened 2022-09-05
feature: allocate random host port when creating test db container

kyujin-cho:


achimnol (changes requested):
We should remove the web/static changes!

PR #710, opened 2022-09-05
Feature/filebrowser in storage proxy

leksikov:
Moves file browsers to the storage proxy. This will greatly reduce the burden of manager and webserver to mediate large file transfers.

* [x] Implement filebrowser container mgmt (with a configurable maximum number of containers)
* [x] Add new manager-facing storage-proxy APIs to create/destroy filebrowser sessions
* [x] **manager**: Add new vfolder APIs to map the new manager-facing storage-proxy APIs
* [x] **client-py**: Add new functional wrappers and `vfolder` subcommands to use the filebrowser session APIs
* [x] **webui**: Update the current filebrowser UIs to use the new filebrowser session mgmt APIs


leksikov:
- Implemented main feature such as the communication between Client, Manager and Storage Proxy to support FileBrowser session where user can attach multiple folders and copy/move data using UI.
- Implemented the CLI commands on client side to start filebrowser with attached folders.
- Implemented Manager API interface to redirect requests from Client to Storage Proxy and receive the response from Storage Proxy about the url, port number and container id.
- The user using CLI can destroy the filebrowser session using the container id information printed out in cli.

While implementing the filebrowser, I decoupled filebrowser from Vfolder as a separate app. 
This will allow having filebrowser its own routing management for more complex commands. 
Decoupling will help to improve maintenance and advanced feature implementation of Filebrowser without affecting Vfolder.

Previously, did the implementation on the server of preventing multiple monitors run with _common.FileLock_.

Implemented UID and GID settings in a startup.sh script after the containers are set. The input variables are defined in the .toml settings file.

Refactoring, in order to deal with multiple volumes and their path in the settings file.

Added a _host:volume_ option in client-py. This is needed for the following things: to specify the volume on which folders are located. The same name folder can occur on multiple volumes. Then based on given argument options such as volume and name the vfid of vfolder is infered. Then filebrowser based on volume path and vfid is able to mount those directories into the container.

On client-py command execution example:
**backend.ai filebrowser create -host local:volume1 -vf mydata1 -vf mydata2**

src/ai/backend/manager/api/filebrowser.py
```diff
@@ -0,0 +1,149 @@
...
+    app["api_versions"] = (
+        2,
+        3,
+        4,
+    )
+    app.on_startup.append(init)
+    app.on_shutdown.append(shutdown)
+    cors = aiohttp_cors.setup(app, defaults=default_cors_options)
+    cors.add(app.router.add_route("POST", r"/create", create_or_update_filebrowser))
+    cors.add(app.router.add_route("DELETE", r"/destroy", destroy_filebrowser))
```
kyujin-cho: If you're not intending to represent regex literal, I think we need to remove leading `r`.
```suggestion
    cors.add(app.router.add_route("POST", "/create", create_or_update_filebrowser))
    cors.add(app.router.add_route("DELETE", "/destroy", destroy_filebrowser))
```

src/ai/backend/storage/config.py
```diff
@@ -13,6 +14,14 @@
...
 
 
+@asynccontextmanager
+async def closing_async(thing):
+    try:
+        yield thing
+    finally:
+        await thing.close()
+
+
```
kyujin-cho: Now that we've migrated into mono-repo setup, I think it's a good chance to remove merge duplicate declarations of `closing_async` into one, maybe on `ai.backend.common.utils`. Please replace `closing_async` usage of cuda plugins and agent also.

src/ai/backend/storage/filebrowser/BUILD
```diff
@@ -0,0 +1,5 @@
+python_sources()
+
+python_tests(
+    name="tests",
+)
```
kyujin-cho: No test case defined inside this package, so this should be removed.
```suggestion
```

src/ai/backend/manager/api/filebrowser.py
kyujin-cho: These regex literals are not used to represent regex literal, so I think it's better to replace with plain string literals.
```suggestion
    cors.add(app.router.add_route("POST", "/create", create_or_update_filebrowser))
    cors.add(app.router.add_route("DELETE", "/destroy", destroy_filebrowser))
```

src/ai/backend/manager/api/filebrowser.py
```diff
@@ -0,0 +1,149 @@
...
+                "vfid": await get_vfid(root_ctx, host, vfolder_name),
+            },
+        )
+    proxy_name, _ = root_ctx.storage_manager.split_host(host)
+    try:
+        proxy_info = root_ctx.storage_manager._proxies[proxy_name]
+    except KeyError:
+        raise InvalidArgument("There is no such storage proxy", proxy_name)
+    headers = {}
+    headers["X-BackendAI-Storage-Auth-Token"] = proxy_info.secret
```
kyujin-cho: Please refrain from using raw string itself rather than referencing `AUTH_TOKEN_HDR` constant under `ai.backend.manager.models.storage`.
```suggestion
    headers = {AUTH_TOKEN_HDR: proxy_info.secret}
```

src/ai/backend/manager/api/filebrowser.py
```diff
@@ -0,0 +1,149 @@
...
+    try:
+        async with proxy_info.session.request(
+            "POST",
+            proxy_info.manager_api_url / "storage/filebrowser/create",
+            headers=headers,
+            json={"host": host, "vfolders": vfolders},
+        ) as client_resp:
+            print(headers, host, proxy_info.manager_api_url / "storage/filebrowser/create")
+            print(web.json_response(await client_resp.json()))
```
kyujin-cho: Are these for debugging? If then please remove it. Otherwise please use logger object to print output.

src/ai/backend/manager/api/filebrowser.py
```diff
@@ -0,0 +1,149 @@
...
+    vfolders = []
+    host = params["host"]
+    for vfolder_name in params["vfolders"]:
+        vfolders.append(
+            {
+                "name": vfolder_name,
+                "vfid": await get_vfid(root_ctx, host, vfolder_name),
+            },
+        )
+    proxy_name, _ = root_ctx.storage_manager.split_host(host)
```
kyujin-cho: There should be a security measure to check if requesting user actually has granted to alter contents inside the vFolder.

src/ai/backend/manager/api/filebrowser.py
```diff
@@ -0,0 +1,149 @@
...
+    # search for volume among available volumes which has file browser container id in order to destroy
+    for volume in volumes:
+        proxy_name = volume[0]
+        try:
+            proxy_info = root_ctx.storage_manager._proxies[proxy_name]
+        except KeyError:
+            raise InvalidArgument("There is no such storage proxy", proxy_name)
+
+        headers = {}
+        headers["X-BackendAI-Storage-Auth-Token"] = proxy_info.secret
```
kyujin-cho: Same here.
```suggestion
        headers = {AUTH_TOKEN_HDR: proxy_info.secret}
```

src/ai/backend/storage/filebrowser/config_browser_app.py
```diff
@@ -0,0 +1,91 @@
...
+
+async def prepare_filebrowser_app_config(
+    settings_path: Path,
+    service_port: int,
+) -> None:
+    filebrowser_config = {
+        "settings": {
+            "key": (
+                "cdsnV117a7KLMN3GAbNWIB/b4w/P1zsVxmAZxAU"
+                "AKidhZG9418Fn5wjE+Zcv6C9eCEDlajcINFkvV+BFSWtVZw=="
```
kyujin-cho: Just for curiosity: Is is okay to leave access key hard-coded like this?

src/ai/backend/storage/filebrowser/database.py
```diff
@@ -0,0 +1,70 @@
...
+        self._engine = sa.engine
+        self.engine = sa.create_engine(self.url)
+
+        insp = sa.inspect(self.engine)
+        if "containers" not in insp.get_table_names():
+            self.meta.create_all(self.engine)
+
+    async def get_all_containers(self) -> Any:
+        with self.engine.connect() as connection:
+            rows = connection.execute(self.containers.select())
```
kyujin-cho: These calls use blocking I/O, so should be executed with `loop.run_in_executor()`. Or, even better, we can utilize async version of sqlite, such as `aiosqlite`... Please make up your mind and either replace all sqlite calls to use `aiosqlite` or wrap it under `run_in_executor()`.

src/ai/backend/storage/filebrowser/filebrowser.py
```diff
@@ -0,0 +1,208 @@
...
+                },
+            )
+        container_name = f"ai.backend.container-filebrowser-{service_port}"
+        container = await docker.containers.create_or_replace(
+            config=config,
+            name=container_name,
+        )
+        container_id = container._id
+        await container.start()
+    tracker_db = FilebrowserTrackerDB(db_path)
```
kyujin-cho: `FilebrowserTrackerDB`'s initializer calls some blocking calls (related to sqlite) and this does isn't good for asyncio codes. We need to split out blocking I/O calls onto separate function (maybe `__ainit__()`?).

src/ai/backend/storage/filebrowser/filebrowser.py
```diff
@@ -0,0 +1,208 @@
...
+                name=container_name,
+            )
+            await container.start()
+        except Exception as e:
+            print("Failure to recreate container ", e)
+
+
+async def destroy_container(ctx: Context, container_id: str) -> None:
+    db_path = ctx.local_config["filebrowser"]["db_path"]
+    tracker_db = FilebrowserTrackerDB(db_path)
```
kyujin-cho: Same here.

src/ai/backend/storage/filebrowser/filebrowser.py
```diff
@@ -0,0 +1,208 @@
...
+) -> tuple[str, int, str]:
+    image = ctx.local_config["filebrowser"]["image"]
+    service_ip = ctx.local_config["filebrowser"]["service_ip"]
+    service_port = ctx.local_config["filebrowser"]["service_port"]
+    max_containers = ctx.local_config["filebrowser"]["max_containers"]
+    user_id = ctx.local_config["filebrowser"]["user_id"]
+    group_id = ctx.local_config["filebrowser"]["group_id"]
+    cpu_count = ctx.local_config["filebrowser"]["max_cpu"]
+    memory = ctx.local_config["filebrowser"]["max_mem"]
+    memory = int(BinarySize().check_and_return(memory))
```
kyujin-cho: `filebrowser.max_mem` is already declared as `tx.BinarySize` on trafaret scheme initiation, so I think we can just skip this step.

src/ai/backend/storage/filebrowser/filebrowser.py
```diff
@@ -0,0 +1,208 @@
...
+    db_path = ctx.local_config["filebrowser"]["db_path"]
+    p = Path(pkg_resources.resource_filename(__name__, ""))
+    storage_proxy_root_path_index = p.parts.index("storage")
+    settings_path = Path(*p.parts[0 : storage_proxy_root_path_index + 1]) / "filebrowser_dir/"
+    _, requested_volume = host.split(":")
```
kyujin-cho: Just for safety!
```suggestion
    _, requested_volume = host.split(":", maxsplit=1)
```

src/ai/backend/storage/filebrowser/filebrowser.py
```diff
@@ -0,0 +1,208 @@
...
+    for volume_name in volumes.keys():
+        if requested_volume == volume_name:
+            volume_cls: Type[AbstractVolume] = BACKENDS[volumes.get(volume_name)["backend"]]
+            mount_path = Path(volumes.get(volume_name)["path"])
+            volume_obj = volume_cls(
+                local_config=ctx.local_config,
+                mount_path=mount_path,
+                fsprefix=None,
+                options={},
+            )
```
kyujin-cho, line 80: What happens if desired volume does not exist actually?

src/ai/backend/storage/filebrowser/filebrowser.py
```diff
@@ -0,0 +1,208 @@
...
+        yield thing
+    finally:
+        await thing.close()
+
+
+async def create_or_update(
+    ctx: Context,
+    host: str,
+    vfolders: list[dict],
+) -> tuple[str, int, str]:
```
kyujin-cho: Returning just tuple can make code readers confused about what the components do. Please replace it with `NamedTuple` or `TypedDict` so that readers can easily understand its role.

src/ai/backend/storage/filebrowser/filebrowser.py
```diff
@@ -0,0 +1,208 @@
...
+                ],
+            },
+        }
+        for vfolder in vfolders:
+            filebrowser_mount_path = str(
+                volume_obj.mangle_vfpath(UUID(vfolder["vfid"])),
+            )
+            config["HostConfig"]["Mounts"].append(
+                {
+                    "Target": f"/data/{str(vfolder['name'])}",
```
kyujin-cho: `vfolder['name']` will be automatically converted to `str` when it's in f-strings.
```suggestion
                    "Target": f"/data/{vfolder['name']}",
```

src/ai/backend/storage/filebrowser/filebrowser.py
```diff
@@ -0,0 +1,208 @@
...
+    async with closing_async(aiodocker.Docker()) as docker:
+        try:
+            docker = aiodocker.Docker()
+            container = await docker.containers.create_or_replace(
+                config=config,
+                name=container_name,
+            )
+            await container.start()
+        except Exception as e:
+            print("Failure to recreate container ", e)
```
kyujin-cho: We need to notify user that the operation is failed, instead of just eating up raised Exception.

src/ai/backend/storage/filebrowser/filebrowser.py
```diff
@@ -0,0 +1,208 @@
...
+    tracker_db = FilebrowserTrackerDB(db_path)
+    async with closing_async(aiodocker.Docker()) as docker:
+        for container in await docker.containers.list():
+            if container._id == container_id:
+                try:
+                    await container.stop()
+                    await container.delete()
+                    await tracker_db.delete_container_record(container_id)
+                except Exception as e:
+                    print(f"Failure to destroy container {container_id[0:7]} ", e)
```
kyujin-cho: Same here: please alert user.

kyujin-cho (changes requested):
Initial batch of reviews. Also, since new file browser feature requires Docker to run, I think we need to describe it on our storage proxy's README. Please take that in mind. cc: @achimnol

src/ai/backend/manager/api/filebrowser.py
```diff
@@ -0,0 +1,295 @@
...
+
+VFolderRow = Mapping[str, Any]
+
+
+async def get_vfid(root_ctx: RootContext, host: str, name: str) -> str:
+    async with root_ctx.db.begin_readonly() as conn:
+        query = (
+            sa.select([vfolders.c.id])
+            .select_from(vfolders)
+            .where(vfolders.c.host == host and vfolders.c.name == name)
```
fregataa:
```suggestion
            .where((vfolders.c.host == host) & (vfolders.c.name == name))
```
This part seems suspicious

fregataa (commented):
Is there any infra set up with multiple storage proxies?
I am curious of DB in storage-proxy.
Is storage-proxy using any DB like sqlite currently or this be the first time to use DB in storage-proxy?
And is there any reason to use aiosqlite rather SQLAlchemy?

src/ai/backend/manager/api/filebrowser.py
leksikov: I am not aware of the infra setup with multiple storage proxies and how it is configured.

Before this PR, the Storage Proxy didn't use any Docker containers or databases, with this PR, I added a SQLite database to track created Docker containers for FileBrowser app. This database is temporarily created for the moment of the duration of usage for FileBrowser containers. After Storage Proxy is shut down all containers and this database are destroyed.

aiosqlite was chosen based on the suggestion mentioned in the comment here for the purpose of async operations:
https://github.com/lablup/backend.ai/pull/710#discussion_r991754787

leksikov: Even, if there are multiple storage setup then I think there is no issues. Let's say in SP1 user created a session for localhost:mydata1 and another user created session for localhost:mydata1 through SP2. Those FileBrowser sessions are independent, they have its own local sqlite database and managed individually by each Storage Proxy instance without conflicting with each other.

achimnol:
Now we are going to merge this PR with the SFTP resource group running on storage proxy nodes. (related: #1246, #1250)

Issue #575, opened 2022-07-22
Introduce vfolder status


achimnol:
For storage-proxy operations that takes a long time, such as vfolder clones and deletion, we need to prevent user access in the meanwhile.

Let's introduce vfolder status to allow or deny user's access to individual vfolders.  For example: `READY`, `CLONING`, `DELETING`.

We also need to consider preventing start of the deletion operation when the target vfolder is being mounted to one or more sessions.

Q: When a vfolder is being cloned, should we block the access to the source vfolder as well?

soheeeeP:
@achimnol Can I try this issue if no one has assigned it yet?

PR #713, opened 2022-09-06
feat: Introduce vfolder status

soheeeeP:
issue #575 

- add column to the `vfolders` table to define its status.
- add the decorator to control user's access while vfolder is implementing storage-proxy operations. (such as clones and deletion)


src/ai/backend/manager/api/vfolder.py
```diff
@@ -82,6 +84,60 @@
 VFolderRow = Mapping[str, Any]
 
 
+def vfolder_status_check(allowed_status: VFolderOperationStatus):
```
fregataa: It is good to implement wrapper method.
How about rename the method and the parameter?
Because I could not understand what this method actually does at the first glance.
Such as `vfolder_filter_by_status()`? This is just an example, so please do not use it if you do not think the name represents what this method does.

src/ai/backend/manager/api/vfolder.py
```diff
@@ -82,6 +84,60 @@
...
+                        VFolderOperationStatus.DELETING,
+                    ]
+                )
+            elif allowed_status == VFolderOperationStatus.PREPARED:
+                vf_status_conds = vfolders.c.status.in_(
+                    [
+                        VFolderOperationStatus.READY,
+                        VFolderOperationStatus.PREPARED,
+                    ]
+                )
```
fregataa: vfolders with any status are filtered if `VFolderOperationStatus.READY` is given?
What does `allowed_status` mean?

src/ai/backend/manager/api/vfolder.py
```diff
@@ -1802,6 +1877,14 @@ async def delete(request: web.Request) -> web.Response:
         },
     ):
         pass
+
+    async with root_ctx.db.begin() as conn:
+        query = (
+            sa.update(vfolders)
+            .values(status=VFolderOperationStatus.READY)
+            .where(vfolders.c.name == folder_name)
+        )
```
fregataa: Why does the status of vfolder become READY? 
How about add DELETED in status?

fregataa (commented):
Good!
I left some comments, please ask me back if you have any question.

src/ai/backend/manager/api/vfolder.py
soheeeeP: `allowed_status` is a parameter which is used to define vfolder statuses (values listed in `vf_status_conds`) to determine whether to lock the vfolder operation or not.

As it seems not intuitive, I modified it to `perm` and listed all cases of `perm`(`VFolderOperationStatus`) by using elif statements.

src/ai/backend/manager/api/vfolder.py
```diff
@@ -101,25 +102,28 @@ async def _wrapped(request: web.Request, *args, **kwargs) -> web.Response:
...
-                    ]
-                )
-            else:
+            elif perm in [
+                VFolderOperationStatus.PERFORMING,
+                VFolderOperationStatus.CLONING,
+                VFolderOperationStatus.DELETING,
+            ]:
+                # if PERFORMING, CLONING, DELETING is requested, only READY is accepted.
                 vf_status_conds = vfolders.c.status == VFolderOperationStatus.READY
```
soheeeeP: @fregataa I remove `PREPARED` and add `PERFORMING` to VFolderOperationStatus.
`PERFORMING` is requested for PATCH/POST operations to make sure that there is no on-going operations (such as cloning, deleting ) of requested vfolder.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -115,9 +115,10 @@ class VFolderOperationStatus(str, enum.Enum):
     """
 
     READY = "ready"
-    PREPARED = "prepared"
+    PERFORMING = "performing"
     CLONING = "cloning"
     DELETING = "deleting"
+    DELETED = "deleted"
     MOUNTED = "mounted"
```
soheeeeP, line 122: @fregataa I can't define clear approach about how to prevent clone/delete operation when the target vfolder is being mounted to one or more sessions. Can I get some hints?

src/ai/backend/manager/models/vfolder.py
```diff
@@ -108,6 +109,19 @@ class VFolderInvitationState(str, enum.Enum):
...
+    """
+    Introduce virtual folder status for storage-proxy operations.
+    """
+
+    READY = "ready"
+    PERFORMING = "performing"
+    CLONING = "cloning"
+    DELETING = "deleting"
+    DELETED = "deleted"
+    MOUNTED = "mounted"
```
kyujin-cho: Your implementation accepts desired vFolder status as an enum, which is also used to represent vFolder current status. The problem is that desired vFolder status and current vFolder status slightly differs its meaning. "current" vFolder status should represent as-is, e.g. Copying something, Cloning to new vFolder, being deleted, and so on. In contrast to previous term, "desired" condition should only distinguish between "ready to read" (superset of all created vFolders) or "ready to alter" (this should mean all vFolders which are only in "READY" status). Using same enum to represent two different categories can be confused to other readers reading this code. How about creating new Enum to represent desired vFolder condition? This value should be determined by current vFolder status (`vfolder.status` column), so it won't be added to database scheme.

```suggestion
class VFolderOperationStatus(str, enum.Enum):
    """
    Introduce virtual folder status for storage-proxy operations.
    """

    READY = "ready"
    PERFORMING = "performing"
    CLONING = "cloning"
    DELETING = "deleting"
    DELETED = "deleted"
    MOUNTED = "mounted"

class VFolderAccessStatus(str, enum.Enum):
    READABLE = "readable"
    WRITABLE = "writable"
```

src/ai/backend/manager/api/vfolder.py
```diff
@@ -82,6 +85,63 @@
...
+        @functools.wraps(handler)
+        async def _wrapped(request: web.Request, *args, **kwargs) -> web.Response:
+            root_ctx: RootContext = request.app["_root.context"]
+            domain_name = request["user"]["domain_name"]
+            user_role = request["user"]["role"]
+            user_uuid = request["user"]["uuid"]
+            allowed_vfolder_types = await root_ctx.shared_config.get_vfolder_types()
+            folder_name = request.match_info["name"]
+            vf_name_conds = vfolders.c.name == folder_name
+            if perm == VFolderOperationStatus.READY:
```
kyujin-cho: Please see the comment above.

kyujin-cho (changes requested):
Thank you for your contribution on Backend.AI! Here some opinions about your code.

src/ai/backend/manager/api/vfolder.py
soheeeeP: @kyujin-cho 
Thanks for the review! I've updated the commits below for the requested changes. 
I have an additional question about using this vfolder filter wrapper method. 
Would it be necessary to add the decorator to all of the operations defined in vfolder api? (commits above declare decorator only for "ready to alter" related operations but not for "ready to read" operations like `get_*`)

kyujin-cho: Finding simillar definitions on previous approaches can be a good starting point. In Backend.AI manager, we have a decorator function `server_status_required()`, which limits access from user to API by filtering user’s permission. When you look at the usages of `server_status_required()`, you can see that we have added this to all of the API handlers which requires authentication, even though it’s allowed to all users, not only for previliged users (admin or superadmin). I think this will be a good reference for `vfolder_filter_by_status_required()`.      
So, in short: yes, please add it to all vFolder API handlers!

soheeeeP: updated this in c14de5e!

soheeeeP (commented):
@kyujin-cho updated this on c14de5e!

src/ai/backend/manager/models/vfolder.py
```diff
@@ -108,6 +110,29 @@ class VFolderInvitationState(str, enum.Enum):
...
+
+
+class VFolderAccessStatus(str, enum.Enum):
+    """
+    Introduce virtual folder desired status for storage-proxy operations.
+    Not added to db scheme  and determined only by current vfolder status.
+    """
+
+    READABLE = "readable"
+    WRITABLE = "writable"
```
kyujin-cho: Hmm, how about changing name of this enum to `UPDATABLE` instead of `WRITABLE`? There can be a single chance that others can misunderstand this status as a 'status that user can write file to this vFolder'.

kyujin-cho (changes requested):
Just a small suggestion!

changes/713.feature.md
```diff
@@ -0,0 +1 @@
+Introduce vfolder status by adding enumtype column to control user access to individual vfolders
```
fregataa: just a little comment.
it is good to write only **what** changes not **how** in news fragment.
it would be better to remove "by adding enumtype column".

src/ai/backend/manager/models/vfolder.py
fregataa: I guess we can change the status in `manager.api.vfolder.clone()` to `CLONING` and restore it in `_clone_bgtask()` method. then the status will change after background task finishes.
but I am not sure whether we should block the access or not.
you can try to apply it and let reviewers consider of it.

fregataa (commented):
I left comments.
How do you think?

src/ai/backend/manager/models/vfolder.py
soheeeeP: @fregataa Does it simply mean to restore the vfolder status update query in `clone_bgtask()` method, so the status can be updated right after the background task is completed?

```python
async def _clone_bgtask(reporter: ProgressReporter) -> None:
    async with root_ctx.storage_manager.request(
        source_folder_host,
        "POST",
        "folder/clone",
        json={
            "src_volume": source_volume_name,
            "src_vfid": str(row["id"]),
            "dst_volume": target_volume_name,
            "dst_vfid": str(folder_id),
        },
    ):
        pass
    async with root_ctx.db.begin() as conn:
        query = (
            sa.update(vfolders)
            .values(status=VFolderOperationStatus.READY)
            .where(vfolders.c.name == row["name"])
        )
        await conn.execute(query)   

task_id = await root_ctx.background_task_manager.start(_clone_bgtask)
```

fregataa: yes. and you can relocate the db query context block into request context

src/ai/backend/manager/api/vfolder.py
```diff
@@ -536,11 +603,19 @@ async def delete_by_id(request: web.Request, params: Any) -> web.Response:
         },
     ):
         pass
+    async with root_ctx.db.begin() as conn:
+        query = (
+            sa.update(vfolders)
+            .values(status=VFolderOperationStatus.DELETED)
+            .where(vfolders.c.id == folder_id)
+        )
+        await conn.execute(query)
```
kyujin-cho: This line won't do anything, as we've already removed all rows having `folder_id` as vfolder id on L593.
```suggestion
```

kyujin-cho (changes requested):
Almost there! A small suggestion.

src/ai/backend/manager/api/vfolder.py
soheeeeP: Updated on 478e951.
Plus, I removed `DELETED` enum value as there is no vfolder status related query using `DELETED`.

fregataa (changes requested):
DB schema has been updated in this PR https://github.com/lablup/backend.ai/pull/695.
Please check alembic history and you will need to merge the migrations.
ref: https://alembic.sqlalchemy.org/en/latest/branches.html#merging-branches

fregataa (approved):
LGTM

kyujin-cho (approved):
Looks good to me!

PR #714, opened 2022-09-07
feature: retry vfolder file download

### LINKED ISSUE #339: Implement retry feature in client-py download files from storage proxy
기능 요청이 특정 문제에 연관된 것이라면 여기에 그 문제를 설명해주세요.

client-py can download files from storage proxy server. the server offers ranged request feature but the client does not offer any retry feature about it.

어떤 기능이 있으면 좋겠는지 자세히 설명해주세요.

let the client use ranged requests to continue interrupted downloads with automatic retries.

혹시 다른 대안들을 생각해본 적이 있다면 함께 적어주세요.

**Additional context**
기능 요청에 대해 보다 잘 이해할 수 있는 다른 맥락을 기술해주세요.

storage proxy server use [FileResponse of aiohttp](https://docs.aiohttp.org/en/latest/web_reference.html#fileresponse) which supports ranged requests with specific headers. there is no need to update any code of server side.

fregataa:
resolve #339 
apply retry using tenacity to vfolder download


src/ai/backend/client/func/vfolder.py
```diff
@@ -203,37 +218,83 @@ def _write_file(file_path: Path, q: janus._SyncQueueProxy[bytes]):
...
-                    finally:
-                        await q.async_q.put(b"")
-                        await writer_fut
-                        q.close()
-                        await q.wait_closed()
+            range_start = 0
+            if_range: str | None = None
+            file_unit = "bytes"
+            file_mode = "wb"
+            file_req_hdrs = {}
```
kyujin-cho: Wouldn't it be better to just define type of `file_req_hdrs` dictionary here rather than calling `cast` later?

```suggestion
            file_req_hdrs: Dict[str, str] = {}
```

src/ai/backend/client/func/vfolder.py
```diff
@@ -203,37 +218,83 @@ def _write_file(file_path: Path, q: janus._SyncQueueProxy[bytes]):
...
+            session_max_attempts = 20
```
kyujin-cho: It'll be great if user can set this value, rather than letting this as constant. How about adding `--max-retries` parameter to vfolder download command?

kyujin-cho (changes requested):

kyujin-cho (approved):
LGTM.

Issue #606, opened 2022-07-26
Allow using an alternate Docker bridge network


achimnol:
In some VM-based setups, using the default Docker bridge network results in inaccessibility to external networks due to the missing gateway.

Though the root cause is still unknown, we discovered that this could be workarounded by using another manually created bridge network.

Let's add an agent configuration option to specify the name of this alternative bridge network, as this should be applied only for containers of single-node cluster sessions and `agent-docker-container-opts.json` (lablup/backend.ai-agent#301) cannot distinguish such conditions.

Ask @adrysn for concrete code locations and conditions to apply this bridge network name.

PR #715, opened 2022-09-07
feat: add alternative docker bridge

kmkwon94:
Handling issue #606 
Add alternative docker bridge configuration option in `sample.toml` (it would be using as `agent.toml`).
 
If you create new docker bridge with this command `docker network create {new_network}`, the new network is created. Before running the agent, uncomment `container-alternative_bridge` in agent.toml. Then compute session would be created as applying new docker network. This logic would be run in singe-node cluster.


kyujin-cho (changes requested):
Please define corresponding config scheme on `src/ai/backend/agent/config.py`, just like other values.  Ref: https://github.com/lablup/backend.ai/blob/dcf198c44b07aa348cf719772000c51e0ba5f6ac/src/ai/backend/agent/config.py#L18

src/ai/backend/agent/docker/agent.py
```diff
@@ -434,6 +434,15 @@ async def apply_network(self, cluster_info: ClusterInfo) -> None:
                         },
                     }
                 )
+        else:
+            if "alternate_bridge" in self.local_config["container"]:
```
adrysn: We can merge these lines to reduce the indentation level.

kmkwon94: Thank you. I applied it

src/ai/backend/agent/docker/agent.py
```diff
@@ -434,6 +434,14 @@ async def apply_network(self, cluster_info: ClusterInfo) -> None:
                         },
                     }
                 )
+        elif "alternate-bridge" in self.local_config["container"]:
```
kyujin-cho: Please add non-null check here, as `config["container"]["alternate-bridge"]` value will always exist (either as a string or `None`).
```suggestion
        elif self.local_config["container"].get("alternate-bridge") is not None:
```

kyujin-cho (changes requested):

src/ai/backend/agent/docker/agent.py
kmkwon94: Thank you. I applied it.

kyujin-cho (approved):
LGTM!

kyujin-cho:
closing #606.

PR #718, opened 2022-09-07
fix: Fix user update in client method and manager's mutation method

### LINKED ISSUE #717: Cannot update admin

macOS (Apple Silicon)

### Backend.AI version

dcf198c44b07aa348cf719772000c51e0ba5f6ac

### Describe the bug

If the user is admin or superadmin, it cannot update any option because of the below error.
image

### To Reproduce

1. Add an admin or superadmin user.
2. Update info of that user using `./backend.ai admin user update [OPTIONS] EMAIL`.
3. See an error.

### Expected Behavior

Update option without error (admin or superadmin previleged).

### Anything else?

I think [these codes](https://github.com/lablup/backend.ai/blob/main/src/ai/backend/manager/models/user.py#L696-L697) are the problem.

fregataa:
resolves #717 
change default parameters of user update method to `None`
fix and refactor user mutate method


Sujin-Kim1 (approved):
LGTM.

Sujin-Kim1 (approved):

src/ai/backend/manager/models/user.py
```diff
@@ -655,13 +655,14 @@ async def _post_func(conn: SAConnection, result: Result) -> Row:
...
                     # User's becomes admin. Set the keypair as active admin.
+                    # TODO: Should we update the role of all users related to keypair?
                     kp = result.first()
                     kp_data = dict()
                     if not kp.is_admin:
                         kp_data["is_admin"] = True
                     if not kp.is_active:
                         kp_data["is_active"] = True
-                    if len(kp_data.keys()) > 0:
+                    if kp_data:
```
kyujin-cho, line 665: Treating non-bool values as bool decreases code readability. Unless you do not have specific reason to do so, let's revert changes on this line.
```suggestion
                    if len(kp_data.keys()) > 0:
```

src/ai/backend/manager/models/user.py
```diff
@@ -670,36 +671,34 @@ async def _post_func(conn: SAConnection, result: Result) -> Row:
...
                             kp_updates.append(kp_data)
-                        cnt += 1
-                    await conn.execute(
-                        sa.update(keypairs)
-                        .values(
-                            {
-                                "is_admin": bindparam("is_admin"),
-                                "is_active": bindparam("is_active"),
-                            }
+                    if kp_updates:
```
kyujin-cho, line 690: Oops! Same here. 
```suggestion
                    if len(kp_updates) > 0:
```

kyujin-cho (changes requested):

src/ai/backend/manager/models/user.py
fregataa: This one is Python's truthness
https://docs.python.org/3/library/stdtypes.html#truth-value-testing
https://github.com/lablup/backend.ai-manager/pull/563/commits/69d8467d6d064e704199ab51c01a55e150e9ff41

src/ai/backend/manager/models/user.py
fregataa: I already updated this one, but how about using Python truthness?

src/ai/backend/manager/models/user.py
kyujin-cho: Oh, I couldn’t catch the latter one. Looks like our current convention is to use Python’s truthness… So I guess you can just ignore this review and revert already changed one. Sorry for inconvenience!

fregataa: thank you. I reverted the commit and you can re-review this.

kyujin-cho (approved):
LGTM. We still need to decide what to do when user reactivates inactive keypair which was created with previous user role, but it's out of this PR's scope.

Issue #691, opened 2022-08-30
Add the commit history between releases in the release notes


achimnol:
Let's append the actual commit history for each release by updating `scripts/extract-release-changelog.py`.
Now we can skip changelogs for specific PRs, but sometimes we may still want to inspect the full history of commits.

The generated URL should look like:
`https://github.com/lablup/backend.ai/compare/22.09.0b4...22.09.0b5` (The result: https://github.com/lablup/backend.ai/compare/22.09.0b4...22.09.0b5)

This requires determining the *previous* release tag name in the script.

100sun:
@kyujin-cho You might close this issue because the above PR has already merged.

PR #721, opened 2022-09-11
feat: Add commit history between releases

soheeeeP:
issue #691 

- Add commit history link between releases in release notes(`CHANGELOG_RELEASE.md`) by updating `scripts/extract-release-changelog.py`

### DIFF
--- a/changes/721.feature.md
+++ b/changes/721.feature.md
@@ -1,0 +1,1 @@
+Add commit history link between releases in release notes(`CHANGELOG_RELEASE.md`) by updating `scripts/extract-release-changelog.py`
--- a/changes/721.feature.md
+++ b/changes/721.feature.md
@@ -1,1 +1,1 @@
-Add commit history link between releases in release notes(`CHANGELOG_RELEASE.md`) by updating `scripts/extract-release-changelog.py`
+Add commit history link between releases in release notes(`CHANGELOG_RELEASE.md`).
--- a/scripts/extract-release-changelog.py
+++ b/scripts/extract-release-changelog.py
@@ -14,6 +14,13 @@ def get_tag():
     revision = p.stdout.decode().strip()
     return revision
 
+def get_prev_tag():
+    rev = subprocess.run(['git', 'rev-list', '--tags', '--skip=0', '--max-count=1'], capture_output=True)
+    rev = rev.stdout.decode().strip()
+
+    tag = subprocess.run(['git', 'describe', '--abbrev=0', '--tags', f'{rev}'], capture_output=True)
+    tag = tag.stdout.decode().strip()
+    return tag
 
 def main():
     parser = argparse.ArgumentParser()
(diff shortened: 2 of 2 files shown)


rapsealk:
Towncrier does not recognize the changelog since the tag `ci-cd` is not supported. You can choose one from `["breaking", "feature", "fix", "deprecation", "doc", "misc"]`.
You may check details here: https://github.com/lablup/backend.ai/blob/6636e3db145a9778cfed60e7b0ccd9578f0ff69b/pyproject.toml#L11-L39

changes/721.feature.md
```diff
@@ -0,0 +1 @@
+Add commit history link between releases in release notes(`CHANGELOG_RELEASE.md`) by updating `scripts/extract-release-changelog.py`
```
kyujin-cho: Sinse changelog is for informing "what" has changed - not "how" it's changed -, it'll be okay to skip which script you updated.
```suggestion
Add commit history link between releases in release notes(`CHANGELOG_RELEASE.md`).
```

kyujin-cho (changes requested):

changes/721.feature.md
soheeeeP: Thanks. I've updated the changelog on the below commit! (dfe35e1)

scripts/extract-release-changelog.py
```diff
@@ -14,6 +14,13 @@ def get_tag():
     revision = p.stdout.decode().strip()
     return revision
 
+def get_prev_tag():
+    rev = subprocess.run(['git', 'rev-list', '--tags', '--skip=0', '--max-count=1'], capture_output=True)
+    rev = rev.stdout.decode().strip()
```
fregataa: since the first `rev` is subprocess object, how about rename it?
this is just a suggestion.

scripts/extract-release-changelog.py
```diff
@@ -14,6 +14,13 @@ def get_tag():
...
+
+    tag = subprocess.run(['git', 'describe', '--abbrev=0', '--tags', f'{rev}'], capture_output=True)
+    tag = tag.stdout.decode().strip()
```
fregataa: same here. The first `tag` is subprocess object.

scripts/extract-release-changelog.py
```diff
@@ -23,8 +30,11 @@ def main():
     )
     args = parser.parse_args()
 
-    tag = get_tag()
-    url = f"https://github.com/lablup/backend.ai/blob/{tag}/CHANGELOG.md"
+    prev_tag, tag = get_prev_tag(), get_tag()
+    url = {
+        "commit_log": f"https://github.com/lablup/backend.ai/compare/{prev_tag}...{tag}",
+        "change_log": f"https://github.com/lablup/backend.ai/blob/{tag}/CHANGELOG.md"
+    }
```
fregataa: just a question.
why did you change `url` into dictionary?
we can just make multiple `..._url` strings.

fregataa (commented):
Thank you for your contribution!
I left some comments for review. But it does not mean that you need to change them all into my suggestion or examples I wrote.
If you answer them all, please re-request review.

scripts/extract-release-changelog.py
soheeeeP: I just thought that using a dictionary would be convenient to handle multiple urls. (in the case of adding new URLs related to logs).

fregataa (approved):
LGTM.
And you need to re-request review after you update.

kyujin-cho (approved):
LGTM!

PR #723, opened 2022-09-12
fix: allow run command without files for exec and vfolder option

### LINKED ISSUE #705: Allow to use run command without files for exec option
**Describe the bug**
When we use run command like this `backend.ai run --exec "python /home/work/test/hello-world.py" -v test python ./blank`, we have to upload at least one file even we don't need it.

**To Reproduce**
Execute run command like this. `backend.ai run --exec "python /home/work/test/hello-world.py" -v test python`. Assume you have vfolder named "test" including "hello-world.py" file

**Expected behavior**
Let user make a choice whether upload file or not. Above case, user can use run command without uploading file. `backend.ai run --exec "python /home/work/test/hello-world.py" -v test` python`

rjwharry:
This PR Fixes #705
- allow no files when exec and vfolder option is supplied
- run exec_loop function whether files are offered or not


src/ai/backend/client/cli/run.py
```diff
@@ -528,7 +528,7 @@ def run(
             "You can run only either source files or command-line " "code snippet.", file=sys.stderr
         )
         sys.exit(ExitCode.INVALID_ARGUMENT)
-    if not files and not code:
+    if not files and not code and not exec and not mount:
```
fregataa, line 531: why is `mount` used here??

src/ai/backend/client/cli/run.py
```diff
@@ -786,24 +786,24 @@ def indexed_vprint_done(msg):
...
-                        stderr,
-                        compute_session,
-                        "batch",
-                        "",
-                        opts=opts,
-                        vprint_done=indexed_vprint_done,
-                        is_multi=is_multi,
-                    )
+            if not is_multi:
+                vprint_done("[{0}] Uploading done.".format(idx))
```
fregataa, line 790: we can leave this code because this just print out the result of file uploading, which needs files.
am I correct?

src/ai/backend/client/cli/run.py
```diff
@@ -786,24 +786,24 @@ def indexed_vprint_done(msg):
...
+                await exec_loop(
+                    stdout,
+                    stderr,
+                    compute_session,
+                    "batch",
+                    "",
+                    opts=opts,
+                    vprint_done=indexed_vprint_done,
+                    is_multi=is_multi,
+                )
```
fregataa, line 806: line no 796 and 807 are exactly opposite condition. we can shorten this
```
if not terminal:
    ...
else:
    ...
```

fregataa (commented):
thank you for your contribute.
But the purpose of `--exec` option is to execute the files, so I think we need to reconsider about the issue.

src/ai/backend/client/cli/run.py
rjwharry: I considered that when use `exec option`, files are offered with two ways, **by uploading files** directly and **by mounting with vfolder**

src/ai/backend/client/cli/run.py
rjwharry: Yes, you're right!

src/ai/backend/client/cli/run.py
rjwharry: I'll apply you're comment. Thank you

rjwharry:
> thank you for your contribute. But the purpose of `--exec` option is to execute the files, so I think we need to reconsider about the issue.

If the files are mounted with vfolder, then we don't have to upload explicitly. That's why I wanted to allow to use run command without uploading files when `--exec` and `--mount` options are enabled simultaneously

achimnol:
This PR is not in urgency and has no activities for more than half a year.
Feel free to post a new, improved PR.
Closing this for now.

PR #725, opened 2022-09-13
fix: unable to purge user when there is no related keyapir

adrysn:
Set concurrency used in purging a user only when there is a related keypair.


kyujin-cho (approved):
LGTM!

PR #727, opened 2022-09-14
setup: Allow installation of package wheels on compatible Python versions

achimnol:
When the main interpreter constraint in `pants.toml` says "==3.10.7", this PR will generate `.tar.gz`, `.whl` distributions with ">=3.10,<3.11" to allow installation on compatible Python versions, while keeping the exact Python versions in CI scripts and the build toolchain.


kyujin-cho (approved):
LGTM!

PR #728, opened 2022-09-14

### LINKED ISSUE #678: Let's group string-literal `KernelEvent.reason` values into a manageable format

qkoo0833:
1. Add class `KernelLifecycleEventReason` in common/events.py to manage `kernelEvent.reason`

close #678


src/ai/backend/agent/agent.py
```diff
@@ -835,7 +836,9 @@ async def _handle_destroy_event(self, ev: ContainerLifecycleEvent) -> None:
                         await self.rescan_resource_usage()
                         if not ev.suppress_events:
                             await self.produce_event(
-                                KernelTerminatedEvent(ev.kernel_id, "already-terminated"),
+                                KernelTerminatedEvent(
```
kyujin-cho, line 841: Please update type definition of `KernelTerminationEventArgs.reason` variable accordingly. `KernelTerminationEventArgs` is a superclass of `KernelTerminatingEvent` and `KernelTerminatedEvent` event class.

src/ai/backend/manager/registry.py
```diff
@@ -1838,7 +1839,11 @@ async def destroy_session(
         async with self.db.begin_readonly() as conn:
             session = await session_getter(db_connection=conn)
         if not reason:
-            reason = "force-terminated" if forced else "user-requested"
+            reason = (
+                KernelLifecycleEventReason.FORCE_TERMINATED
+                if forced
+                else KernelLifecycleEventReason.USER_REQUESTED
+            )
```
kyujin-cho, line 1846: This `reason` variable is used below to populate `SessionCancelledEvent` and `KernelCancelledEvent`, so please update type definition of `SessionCreationEventArgs` and `KernelCreationEventArgs`'s `reason` variable accordingly, just like I mentioned on review above.

src/ai/backend/manager/registry.py
```diff
@@ -2033,7 +2038,9 @@ async def _update() -> None:
                         )
 
                     if kernel["agent_addr"] is None:
-                        await self.mark_kernel_terminated(kernel["id"], "missing-agent-allocation")
+                        await self.mark_kernel_terminated(
```
kyujin-cho: Same here! Please update type definition of `mark_kernel_terminated()` function's `reason` argument.

src/ai/backend/manager/registry.py
kyujin-cho: Also note that you'll need to change all other codes populating `Session*Event` or `Kernel*Event` event with `reason` variable.

kyujin-cho: Also note that you'll need to change all other codes populating `Session*Event` or `Kernel*Event` event with `reason` variable.

kyujin-cho (changes requested):
Thank you for your contribution. Some opinions about your codes:

src/ai/backend/common/events.py
```diff
@@ -194,7 +195,7 @@ def deserialize(cls, value: tuple):
 class KernelCreationEventArgs:
     kernel_id: KernelId = attr.ib()
     creation_id: str = attr.ib()
-    reason: str = attr.ib(default="")
+    reason: KernelLifecycleEventReason | str = attr.ib(default="")
```
kyujin-cho: It'll be better to just set `KernelLifecycleEventReason` as the only allowed type of `reason`, add `UNKNOWN` value to the Enum, and set that value as default of `reason` since it seems better not to allow any arbitrary reasons. No need to say this change should be applied to all other declaration os `reason` value in `*EventArgs` series! :)
```suggestion
    reason: KernelLifecycleEventReason = attr.ib(default=KernelLifecycleEventReason.UNKNOWN)
```

src/ai/backend/common/events.py
```diff
@@ -529,6 +530,21 @@ def __call__(
...
+    KILLED_BY_EVENT = "killed-by-event"
+    PENDING_TIMEOUT = "pending-timeout"
+    SELF_TERMINATED = "self-terminated"
+    TASK_DONE = "task-done"
+    TASK_FAILED = "task-failed"
+    TASK_TIMEOUT = "task-timeout"
+    TASK_CANCELLED = "task-cancelled"
+    TASK_FINISHED = "task-finished"
+    USER_REQUESTED = "user-requested"
+
```
kyujin-cho: Let's add new `UNKNOWN` value to be used as default value of `reason`.
```suggestion
    UNKNOWN = "unknown"
```

kyujin-cho (changes requested):

s4shantanu (approved):

qkoo0833:
close this PR since #855 replaces this.

PR #731, opened 2022-09-15
feat: always mount infiniband device if exists

kyujin-cho:
This PR alters infiniband device mount mechanism to always mount infiniband devices if it exists regardless of if it's multicontainer session or not.


achimnol (approved):

adrysn (approved):
LGTM.

PR #732, opened 2022-09-20
add: support webserver option for session enqueue

inureyes:
This change adds webserver option to support session enqueue feature, introduced from https://github.com/lablup/backend.ai-webui/pull/1162 

A user cannot create a compute session when there is no available resource in the Agent hosts. In the absence of the resources, the session launcher dialog disables the CREATE button.

Some users / groups find this inconvenient since they have to wait until other users destroy their compute sessions to create their own. They rather wanted to create a PENDING session immediately and then get the RUNNING session automatically just after the resources are available. Therefore, queueing the session creation request feature added in backend.ai-webui, and it is configurable via `config.toml` as a global service option. 

This PR adds the `alwaysEnqueueComputeSession` option under `general` section of webserver-side `config.toml` too.

### DIFF
--- a/changes/732.feature.md
+++ b/changes/732.feature.md
@@ -1,0 +1,1 @@
+Add webserver option to support session enqueue feature, introduced from lablup/backend.ai-webui#1162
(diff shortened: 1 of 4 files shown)


kyujin-cho (approved):
Looks good to me.

PR #737, opened 2022-09-22
feat: impl abusing report and agent local_config API

fregataa:
Implement kernel's abusing report API.
Add CLI command and add ComputeContainer graph field `abusing_report`.
example image)
abusing report is saved as following.
image

The report shows as following.
image

---
Since there are many unserializable items in agent's local_config that RPC function cannot parse, let agents return only a few items of local_config

image

### DIFF
--- a/changes/737.feature.md
+++ b/changes/737.feature.md
@@ -1,1 +1,1 @@
-impl graphql resolver querying agent local_config
+impl abusing report and agent local_config API
--- a/changes/737.feature.md
+++ b/changes/737.feature.md
@@ -1,0 +1,1 @@
+impl graphql resolver querying agent local_config
--- a/changes/737.feature.md
+++ b/changes/737.feature.md
@@ -1,1 +1,1 @@
-impl abusing report and agent local_config API
+Introduce abusing container status and agent local_config API
--- a/src/ai/backend/agent/server.py
+++ b/src/ai/backend/agent/server.py
@@ -560,9 +560,11 @@ class AgentRPCServer(aobject):
     @rpc_function
     @collect_error
     async def get_local_config(self) -> Mapping[str, Any]:
+        agent_config: Mapping[str, Any] = self.local_config["agent"]
+        report_path: Path | None = agent_config.get("abuse-report-path")
         return {
             "agent": {
-                "abuse-report-path": str(self.local_config["agent"].get("abuse-report-path", "")),
+                "abuse-report-path": str(report_path) if report_path is not None else "",
             },
             "watcher": self.local_config["watcher"],
         }
--- a/src/ai/backend/agent/server.py
+++ b/src/ai/backend/agent/server.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import functools
 import importlib
+import json
 import logging
 import logging.config
 import os
@@ -575,6 +576,19 @@ class AgentRPCServer(aobject):
             },
             "watcher": self.local_config["watcher"],
         }
+
+    @rpc_function
+    @collect_error
+    async def get_abusing_report(
+        self,
+        kernel_id,  # type: str
+    ):
+        if (abuse_path := self.local_config["agent"].get("abuse-report-path")) is not None:
+            report_path = Path(abuse_path, f"report.{kernel_id}.json")
+            if report_path.is_file():
+                with open(report_path) as file:
+                    return json.load(file)
+        return
 
     @rpc_function
     @collect_error
--- a/src/ai/backend/agent/server.py
+++ b/src/ai/backend/agent/server.py
@@ -559,20 +559,10 @@ class AgentRPCServer(aobject):
 
     @rpc_function
     @collect_error
-    async def get_local_config(self):
-        def get_or_empty(val):
-            if val is None:
-                return ""
-            return str(val)
-
+    async def get_local_config(self) -> Mapping[str, Any]:
         return {
             "agent": {
-                "abuse-report-path": get_or_empty(
-                    self.local_config["agent"].get("abuse-report-path")
-                ),
-                "image-commit-path": get_or_empty(
-                    self.local_config["agent"].get("image-commit-path")
-                ),
+                "abuse-report-path": str(self.local_config["agent"].get("abuse-report-path", "")),
             },
             "watcher": self.local_config["watcher"],
         }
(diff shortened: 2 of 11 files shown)


src/ai/backend/agent/server.py
```diff
@@ -556,6 +556,28 @@ async def commit(
             "path": str(Path(subdir, filename)),
         }
 
+    @rpc_function
+    @collect_error
+    async def get_local_config(self):
+        log.debug("rpc::get_local_config()")
```
agatha197: Why don't you erase it because it comes out too often? Because of this, it is difficult to check other logs.

Sujin-Kim1 (changes requested):

src/ai/backend/agent/server.py
fregataa: okay. seems this log is not useful that much

src/ai/backend/agent/server.py
```diff
@@ -556,6 +557,39 @@ async def commit(
...
+        def get_or_empty(val):
+            if val is None:
+                return ""
+            return str(val)
+
+        return {
+            "agent": {
+                "abuse-report-path": get_or_empty(
+                    self.local_config["agent"].get("abuse-report-path")
+                ),
```
kyujin-cho: Instead of defining custom function, adding default value to `get()` function will be sufficient.
```suggestion
                "abuse-report-path": self.local_config["agent"].get("abuse-report-path", ""),
```

src/ai/backend/agent/server.py
```diff
@@ -556,6 +557,39 @@ async def commit(
...
+                "image-commit-path": get_or_empty(
+                    self.local_config["agent"].get("image-commit-path")
+                ),
```
kyujin-cho: Same here.

src/ai/backend/agent/server.py
```diff
@@ -556,6 +557,39 @@ async def commit(
...
+    @collect_error
+    async def get_abusing_report(
+        self,
+        kernel_id,  # type: str
+    ):
+        if (abuse_path := self.local_config["agent"].get("abuse-report-path")) is not None:
+            report_path = Path(abuse_path, f"report.{kernel_id}.json")
+            if report_path.is_file():
+                with open(report_path, "r") as file:
+                    return json.load(file)
```
kyujin-cho: Please wrap blocking operation on a sync function and run it on executor.
```suggestion
                def _sync():
                    with open(report_path, "r") as file:
                        return json.load(file)
                return await asyncio.get_running_loop().run_in_executor(None, _sync)
```

src/ai/backend/agent/server.py
```diff
@@ -556,6 +557,39 @@ async def commit(
...
+        return
```
kyujin-cho: `return` here seems meaningless. If you were to return `None`, please write it down explicitly.
```suggestion
        return None
```

src/ai/backend/manager/api/session.py
```diff
@@ -1345,6 +1345,37 @@ async def get_commit_status(request: web.Request, params: Mapping[str, Any]) ->
...
+    ),
+    loads=_json_loads,
+)
+async def get_abusing_report(request: web.Request, params: Mapping[str, Any]) -> web.Response:
+    root_ctx: RootContext = request.app["_root.context"]
+    session_name: str = request.match_info["session_name"]
+    requester_access_key, owner_access_key = await get_access_key_scopes(request)
+
+    myself = asyncio.current_task()
+    assert myself is not None
```
kyujin-cho: `myself` seems not used.
```suggestion
```

src/ai/backend/agent/server.py
```diff
@@ -556,6 +557,39 @@ async def commit(
...
+            },
+            "watcher": self.local_config["watcher"],
+        }
+
+    @rpc_function
+    @collect_error
+    async def get_abusing_report(
+        self,
+        kernel_id,  # type: str
+    ):
```
kyujin-cho: Same here.

src/ai/backend/agent/server.py
```diff
@@ -556,6 +557,39 @@ async def commit(
             "path": str(Path(subdir, filename)),
         }
 
+    @rpc_function
+    @collect_error
+    async def get_local_config(self):
```
kyujin-cho: Please add return type of this function. mypy won't check code inside function without return type.

kyujin-cho (changes requested):

src/ai/backend/manager/registry.py
```diff
@@ -3046,6 +3046,38 @@ async def commit_session(
...
+        ) as rpc:
+            return await rpc.call.get_local_config()
+
+    async def get_abusing_report(
+        self,
+        session_name_or_id: Union[str, SessionId],
+        access_key: AccessKey | None = None,
+    ) -> Optional[Mapping[str, str]]:
+        if access_key is not None:
+            kernel = await self.get_session(session_name_or_id, access_key)
```
kyujin-cho: There's no need to get access key information if we already know a reference to session (either name or access key). If you are intending to add user validation logic so that only session owner can retrieve abusing report info, how about setting `access_key` as required variable, removing below logic which resolves session  information without access key, and updating `ComputeContainer.resolve_abusing_report()` to supply access key?

src/ai/backend/manager/models/kernel.py
```diff
@@ -968,6 +976,16 @@ async def resolve_commit_status(self, info: graphene.ResolveInfo) -> Optional[st
         commit_status = await graph_ctx.registry.get_commit_status(self.id, self.access_key)
         return commit_status["status"]
 
+    async def resolve_abusing_reports(
+        self, info: graphene.ResolveInfo
+    ) -> Iterable[Mapping[str, Any]]:
```
kyujin-cho: abusing report can be `None`.
```suggestion
    ) -> Iterable[Optional[Mapping[str, Any]]]:
```

kyujin-cho (changes requested):

src/ai/backend/manager/registry.py
fregataa: ok, I change `access_key` to required parameter and pass the value to the resolver.

changes/737.feature.md
```diff
@@ -0,0 +1 @@
+impl abusing report and agent local_config API
```
kyujin-cho:
```suggestion
Introduce abusing container status and agent local_config API
```

kyujin-cho (changes requested):

PR #738, opened 2022-09-23
fix: Fix wrong type matching in scaling group query method

fregataa:


adrysn (approved):
LGTM. Tested and it works.

src/ai/backend/manager/models/scaling_group.py
```diff
@@ -216,13 +216,15 @@ async def query_allowed_sgroups(
...
-            group_ids = []
+    match group:
+        case uuid.UUID() | str():
+            if group_id := await resolve_group_name_or_id(db_conn, domain_name, group):
+                group_ids = [group_id]
+            else:
+                group_ids = []
+        case list() | tuple() | set():
+            grps = await resolve_groups(db_conn, domain_name, cast(Iterable, group))
+            group_ids = [*grps]
```
kyujin-cho: Is there a specific reason unpacking and repacking `list` instead of using as is?

kyujin-cho (changes requested):

kyujin-cho (approved):

PR #744, opened 2022-09-25
feat: add support for IBM Spectrum Scale storage

kyujin-cho:
This resolves lablup/backend.ai-internal#15.


adrysn (approved):
I tested the feature and works well including per-vfolder quota.

PR #747, opened 2022-09-26
fix: alembic migration not working

kyujin-cho:
This PR fixes some old alembic migration script failing to execute.


achimnol (approved):

kyujin-cho:
Closing #785

PR #755, opened 2022-09-29
ci: Integration test automation workflow

### LINKED ISSUE #741: Add GitHub Actions workflow to run the integration test suite
[Refer the `ai.backend.test` source directory.](https://github.com/lablup/backend.ai/tree/main/src/ai/backend/test)

Let's run it [periodically](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) and [manually requested](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch).

Yaminyam:
close #741 
Add `script -q -e -c "bash {0}"` shell setting for tty use by `-i` option of `docker exec` according to the environment in github actions

Periodic execution using crontab
Executed by manual trigger using workflow_dispatch.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
   
 jobs:
-  test:
+  integration:
     if: ${{ !contains(github.event.pull_request.labels.*.name, 'skip:ci') }}
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -54,8 +54,8 @@ jobs:
       shell: 'script -q -e -c "bash {0}"'
       run: |
         ./scripts/install-dev.sh
-        source env-tester.sh
-        backend.ai test run-cli
+        source env-tester-admin.sh
+        ./backend.ai test run-cli
     - name: Upload pants log
       uses: actions/upload-artifact@v3
       with:
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -55,7 +55,9 @@ jobs:
       run: |
         ./scripts/install-dev.sh
         source env-tester-admin.sh
-        ./backend.ai test run-cli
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yml
```diff
@@ -0,0 +1,64 @@
+name: integration-test
+
+on:
+  schedule:
+    - cron: '0 0 * * 6'
```
fregataa, line 5: this test runs every saturday 00:00 ?

.github/workflows/integration.yml
```diff
@@ -0,0 +1,64 @@
...
+    - name: Apply cache for Pants
+      uses: actions/cache@v3
+      id: cache
+      with:
+        # pants-specific cache
+        path: |
+          ~/.cache/pants/setup
+          ~/.cache/pants/lmdb_store
+          ~/.cache/pants/named_caches
+        key: ${{ runner.os }}-${{ hashFiles('pants*.toml', '**/*.lock') }}-
```
fregataa, line 27: Every test reuses this cache?
And how long this cache remain?

fregataa (commented):
I left some comments

.github/workflows/integration.yml
Yaminyam: That's right. This test runs every Saturday on the hour.
It seems to run at 10:00 Korean time.

.github/workflows/integration.yml
Yaminyam: https://github.com/lablup/backend.ai/actions/caches
Recently updated github to check caches in .

I don't know exactly how much cache is maintained, but I checked until the cache was over a month old.

Yaminyam: https://github.com/actions/cache
Caches that have not been accessed within the past week are deleted, and if the 10GB capacity is full, the oldest cache is deleted.

achimnol (approved):
LGTM!
Let's run it first and fix any issues afterwards.

PR #756, opened 2022-09-29
fix: filter `None` in GraphQL response value

fregataa:
convert None to empty string or filter out in GraphQL response, especially list or JSON value.

response values as following seems to occur [an error](https://github.com/lablup/backend.ai-control-panel/pull/371#pullrequestreview-1119923137)
```
{
  "key": None
}

[None, 1, 2, 3]
```


src/ai/backend/manager/models/kernel.py
```diff
@@ -982,13 +1001,17 @@ async def resolve_commit_status(self, info: graphene.ResolveInfo) -> Optional[st
...
         if containers is None:
             containers = await self.resolve_containers(info)
         if containers is None:
             return []
-        return [(await con.resolve_abusing_report(info, self.access_key)) for con in containers]
+        return [
+            result
+            for con in containers
+            if (result := await con.resolve_abusing_report(info, self.access_key)) is not None
+        ]
```
achimnol: Could we rewrite this part using `aiotools.TaskGroup`?

src/ai/backend/manager/models/agent.py
```diff
@@ -197,7 +197,25 @@ async def resolve_hardware_metadata(
...
+            converted: MutableMapping[str, Any] = {}
+            for k, v in data.items():
+                if v is None:
+                    converted[k] = ""
+                    continue
+                if isinstance(v, dict):
+                    converted[k] = _convert_none_to_str(v)
+                    continue
+                converted[k] = v
+            return converted
```
achimnol: * Let's refactor out this function as it is also duplicate in `src/ai/backend/manager/models/kernel.py`. Probably `ai.backend.common.utils` would be a good place to put this kind of small utility functions, though `common.utils` should be always considered as "to-be-refactored".
* I'd suggest a more concise naming: `coerce_none`.
* Since the function is coercing None to the empty string always, I believe it's assuming that `data` is `Mapping[str, str | None]`. Am I correct?
* Let's rewrite the body using `if-elif-else` blocks instead of inserting `continue` in every `if` branch.

src/ai/backend/manager/models/kernel.py
```diff
@@ -669,9 +670,27 @@ async def resolve_abusing_report(
         access_key: AccessKey | None,
     ) -> Optional[Mapping[str, Any]]:
         graph_ctx: GraphQueryContext = info.context
+
+        def _convert_none_to_str(data: Mapping[str, Any]) -> Mapping[str, Any]:
```
achimnol: Duplicate, so let's refactor.

achimnol (changes requested):

src/ai/backend/manager/models/kernel.py
fregataa: I use TaskGroup here. thanks!

src/ai/backend/manager/models/agent.py
fregataa: reflexed. but I am not sure whether the 3rd question is correct or not. cause I guess the Value of the map can be Mapping, None or anything etc

kyujin-cho:
Closing as originated issue is already resolved by another solution.

Issue #399, opened 2022-03-22
Adopt the plugin architecture for agent kernel mounts


achimnol:
Let's split out commonly used kernel runner binaries (e.g., ttyd, vscode, ...) to be separately updatable and installable by making them provided by plugins.

Things to consider:
* backend-plugin compatibility check (docker, k8s, ...)
* image-plugin compatibility check (some plugins may cover all kernel images while others may cover only a specific set of images) - maybe using wildcard pattern against image name & tags / a functional interface to match the image metadata manually (with architecture, labels, name, tags)
* declaration of additional service ports & injection of service definition configs by plugins
* generic mount lists

fregataa:
where should `metadata` or service definition configs be located?

PR #757, opened 2022-09-29
feat: add code server bundle and impl plugin for agent

fregataa:
add code-server standalone tarball zip file and agent handle it like a plugin.

Mount [the standalone released tarball file](https://coder.com/docs/code-server/latest/install#standalone-releases) and unzip it in kernel initializing process

refs #399


fregataa:
close this PR since #765 replaces this

Issue #237, opened 2020-12-14
CephFS quota support


achimnol:
* Reference: https://docs.ceph.com/en/latest/cephfs/quota/
 * Let's extend the `vfs` backend in [our storage proxy](https://github.com/lablup/backend.ai-storage-proxy) to the `cephfs` backend, by overriding quota-related methods and directoy size queries to use the CephFS-specific attributes.
 * This is a part of #50.

achimnol:
Now this is reincarnated as VFolder v3 with quota scopes.

Issue #50, opened 2019-08-20
vfolder (storage) size limits


achimnol:
We need to implement size limits for vfolders.
The reasons why we haven't implement this so far:
 * ext3/ext4 filesystems does not support per-directory quota but they are most prevalent.
 * Docker's overlay2 filesystem also does not support per-directory quota unless the underlying filesystem is xfs.
 * NFS does not support per-directory quota by itself and rely on the underlying filesystem's user quota.
 * We cannot use per-user quota since all containers are executed as the same user/group IDs.
 * It is impractical to create filesystems per vfolder due to scale limits (e.g., AWS EFS allows up to 1,000 filesystems per region per account and 400 target mounts per VPC).
 * It is not scalable to use size-limited loopback files for vfolders because the vfolder size limits must be changeable (e.g., billing plan upgrades)

Since vfolders may be set up with different backends, we need to define the common abstract interface for vfolders and make specialized implementations.
(related: lablup/backend.ai-agent#70)

* [ ] A storage agent with simple HTTP REST APIs
  - It manages the mapping of (user|group, vfolder) -> (physical-directory)
  - It applies the quota in the user or group level.
  - [x] "create-dir-with-quota" API
  - [x] "change-dir-quota" API
  - [x] "delete-dir-with-quota" API
  - [ ] "get-dir-from-key" API to get the physical path
  - [x] "measure-dir-size" API
  - [ ] "stream-events" API (using HTML5 server-side push) to publish quota-exceeded or quota-recovered events for periodic checks
  - [x] A minimal authentication like agent watchers, using a shared secret key
  - [x] Let the manager use the storage agent API when creating a new vfolder, modifying an existing one, or deleting it.
  - [x] Configuration support for different vfolder drivers below
  - We keep the same assumption: all manager & agent nodes have the vfroot mounts at the same path.
  - vfolder creation/modification/deletion will go through the storage agent, while other operations such as mounting them into the container and up/downloading files works like now.
  - Deployment for distributed filesystems and opaque filesystems
    - The storage agent runs along with the manager or on a separate node.
    - The storage agent should have access to the filesystem management tools & APIs.
  - Deplyoment for centralized NAS
    - The storage agent runs on the NAS host and the manager accesses it via the network.
    - The storage agent should have direct access with proper permissions to the NAS' filesystem utilities.
    - This requires the NAS to be formatted using a filesystem capable of per-directory quota limits such as xfs.
* [ ] cephfs vfolder driver (#237)
[...]

PR #760, opened 2022-09-30
feature: Add CephFs backend to storage-proxy

leksikov:
Migration to Mono Repository.
Feature implementation and solution for issue _CephFS quota support_ [#237](https://github.com/lablup/backend.ai/issues/237) and part of [#50](https://github.com/lablup/backend.ai/issues/50).

After the CephFS is installed and vfolder directory is set and mounted. Storage proxy can create folders in designated directory and set the quotas limits.

The guide and script about testing Ceph is at [manual](https://github.com/lablup/backend.ai-storage-proxy/tree/feature/ceph-implementation/scripts/spawn-ceph-clusters)


leksikov:
I have tested the CephFs cluster on Mac-Mini-03 machine.
And linter checks passed and ready for integration.
Cluster of 5 nodes were created using Vagrant VM.
The pull request in separate Storage-Proxy repository can be removed.
https://github.com/lablup/backend.ai-storage-proxy/tree/feature/filebrowser-in-storage-proxy

src/ai/backend/storage/cephfs/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+from ..types import FSUsage, VFolderCreationOptions
+from ..vfs import BaseVolume
+
+
+async def read_file(loop: asyncio.AbstractEventLoop, filename: str) -> str:
+    def _read():
+        with open(filename, "r") as fr:
+            return fr.read()
+
+    return await loop.run_in_executor(None, lambda: _read())
```
kyujin-cho: You can just throw `_read` itself as a function, without wrapping it with lambda. 
```suggestion
    return await loop.run_in_executor(None, _read)
```

src/ai/backend/storage/cephfs/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+async def write_file(loop: asyncio.AbstractEventLoop, filename: str, contents: str, perm="w"):
+    def _write():
+        with open(filename, perm) as fw:
+            fw.write(contents)
+
+    await loop.run_in_executor(None, lambda: _write())
```
kyujin-cho: Same here.
```suggestion
    await loop.run_in_executor(None, _write)
```

src/ai/backend/storage/cephfs/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+                stderr=asyncio.subprocess.STDOUT,
+            )
+        except FileNotFoundError:
+            available = False
+
+        if not available:
+            raise RuntimeError("Ceph is not installed. ")
+
+    # ----- volume opeartions -----
+    async def create_vfolder(self, vfid: UUID, options: VFolderCreationOptions = None) -> None:
```
kyujin-cho: `options` should be `Optional`.
```suggestion
    async def create_vfolder(self, vfid: UUID, options: Optional[VFolderCreationOptions] = None) -> None:
```

src/ai/backend/storage/cephfs/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+        vfpath = self.mangle_vfpath(vfid)
+        await run(f"mkdir -p {vfpath}")
```
kyujin-cho: `BaseVolume.create_vfolder()`, overridden method of this method, does exactly same thing with these lines. I think it's be better to let them do its job.
```suggestion
        await super().create_vfolder(vfid, options, exist_ok=exist_ok)
```

src/ai/backend/storage/cephfs/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+        if len(stat.split()) == 6:
+            raise ExecutionError("'df -h' stdout is in an unexpected format")
+        if str(self.mount_path) != path:
+            raise ExecutionError("'df -h' stdout is in an unexpected format")
+        return FSUsage(
+            capacity_bytes=BinarySize.finite_from_str(capacity),
+            used_bytes=BinarySize.finite_from_str(used),
+        )
+
+    async def get_quota(self, vfpath):
```
kyujin-cho: Please add return type definition.
```suggestion
    async def get_quota(self, vfpath) -> int:
```

src/ai/backend/storage/cephfs/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+        report = await run(f"getfattr -n ceph.quota.max_bytes {vfpath}")
```
kyujin-cho: `getattr` and `setattr` utilizes `getxattr` and `setxattr` syscall, respectively, to do their job (ref: https://git.savannah.nongnu.org/cgit/attr.git/tree/tools/setfattr.c#n93). Python's `os` standard library is also [capable of](https://docs.python.org/3/library/os.html#os.getxattr) executing those syscalls. It'll be better to replace these subprocess calls with `os.getxattr()` and `os.setxattr().

src/ai/backend/storage/cephfs/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+        if len(report.split()) != 6:
+            raise ExecutionError("ceph quota report output is in unexpected format")
+        _, quota = report.split("=")
+        quota = quota.replace('"', "")
+        return int(quota)
+
+    async def set_quota(self, vfpath, size_bytes) -> None:
+        await run(f"setfattr -n ceph.quota.max_bytes -v {int(size_bytes)} {vfpath}")
```
kyujin-cho: Same here.

src/ai/backend/storage/spawn-ceph-clusters/README.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+# Mounting Ceph-Fuse
+On ceph-client machine it is already mounted on install.
+```
+sudo ceph-fuse -n client.admin --keyring=/etc/ceph/ceph.client.admin.keyring  -m ceph-server-1 /mnt/vfroot/ceph-fuse/
+```
+
+# Setting Quotas setup
+<<<<<<< HEAD
```
kyujin-cho: Hmm... These seem like signs of merge conflict. Please resolve it manually.

src/ai/backend/storage/spawn-ceph-clusters/README.md
```diff
@@ -0,0 +1,94 @@
+# Readme
```
kyujin-cho: The scripts inside `spawn-ceph-clusters` seems very useful to those who want to bootstrap ceph cluster, but I don't think it's not that related to the source code itself. Perhaps you can move this directory onto `scripts/stroage-proxy` directory of the repository.

src/ai/backend/storage/spawn-ceph-clusters/README.md
kyujin-cho: Hmm... These seem like signs of a merge conflict. Please resolve it.

kyujin-cho (changes requested):
Some changes requested!    

Also, please write down sample configuration about this plugin on `configs/storage-proxy/sample.toml`. Don't forget to also update `scripts/install-dev.sh` so that ceph configurations won't be added to installed configuration.

src/ai/backend/storage/cephfs/__init__.py
leksikov: `await super().create_vfolder(vfid, options, exist_ok=exist_ok)` It didn't work. It doesn't recognize super class.
Also, there are many new mypy errors related to Optional across Storage-Proxy files which I don't know how to resolve.

src/ai/backend/storage/cephfs/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+            raise ExecutionError("ceph quota report output is in unexpected format")
+        _, quota = report.split("=")
+        quota = quota.replace('"', "")
+        return int(quota)
+
+    async def set_quota(self, vfpath, size_bytes: BinarySize) -> None:
+        loop = asyncio.get_running_loop()
+        await loop.run_in_executor(
+            None,
+            lambda: os.setxattr(vfpath, "ceph.quota.max_bytes", str(int(size_bytes)).encode()),
```
leksikov: I just want simply convert BinarySize to byte string. There is no simple way to do it within BinarySize?

scripts/storage-proxy-spawn-ceph-clusters/install_storage_proxy.sh
```diff
@@ -0,0 +1,51 @@
...
+echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.profile
+echo 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.profile
+echo 'eval "$(pyenv init -)"' >> ~/.profile
+
+echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.bashrc
+echo 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.bashrc
+echo 'eval "$(pyenv init -)"' >> ~/.bashrc
+
+echo 'eval "$(pyenv virtualenv-init -)"' >> ~/.profile
+echo 'eval "$(pyenv virtualenv-init -)"' >> ~/.bashrc
```
kyujin-cho, line 35: Instead of doing implicitly, how do you think about letting user to install pyenv by themselves?

scripts/storage-proxy-spawn-ceph-clusters/install_storage_proxy.sh
```diff
@@ -0,0 +1,51 @@
...
+
+pyenv install 3.10.7
+pyenv global 3.10.7
+pyenv local 3.10.7
+
+git clone https://github.com/lablup/backend.ai bai-dev
+cd bai-dev
+sudo ./scripts/install-dev.sh
```
kyujin-cho, line 43: I think we can assume that user has completed running `install-dev.sh`. Let's skip this line.

scripts/storage-proxy-spawn-ceph-clusters/install_storage_proxy.sh
```diff
@@ -0,0 +1,51 @@
...
+sudo chmod 755 /
+cd $HOME/
+sudo apt update
+sudo apt-get install -y build-essential git-core libreadline-dev libsqlite3-dev libssl-dev libbz2-dev tk-dev libzmq3-dev libsnappy-dev libffi-dev zlib1g-dev wget curl llvm libncursesw5-dev xz-utils  libxml2-dev libxmlsec1-dev  liblzma-dev
+
+curl -fsSL https://get.docker.com -o get-docker.sh
+sudo sh get-docker.sh
+sudo apt-get install docker-compose
+sudo addgroup docker
+sudo adduser vagrant docker
```
kyujin-cho, line 13: Commands in this part are already covered by `install-dev.sh`, so we can skip this lines.

scripts/storage-proxy-spawn-ceph-clusters/install_storage_proxy.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/sh
+
+
+sudo chmod 755 /
```
kyujin-cho, line 4: It's not a really good idea to update the permission of root folder. Let's just remove this line.

src/ai/backend/storage/cephfs/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+async def write_file(loop: asyncio.AbstractEventLoop, filename: str, contents: str, perm="w"):
+    def _write():
+        with open(filename, perm) as fw:
+            fw.write(contents)
+
+    await loop.run_in_executor(None, _write())
+
+
```
kyujin-cho: These functions are not used anymore, so how about removing these?

src/ai/backend/storage/cephfs/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+        await loop.run_in_executor(
+            None,
+            lambda: vfpath.mkdir(0o755, parents=True, exist_ok=exist_ok),
+        )
+        if not (options is None or options.quota is None or options.quota == 0):
+            quota = options.quota
+            await self.set_quota(vfpath, quota)
+
+    async def get_fs_usage(self) -> FSUsage:
+        stat = await run(f"df -h {self.mount_path} | grep {self.mount_path}")
```
kyujin-cho: Parsing output of command execution should always be considered as last resort since it depends on lots of other things ($PATH setup, output format, ...). I think we can replace `df -h` with `os.statvfs()` API, which is provided by Python's built-in `os` library.

src/ai/backend/storage/cephfs/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+
+    async def init(self) -> None:
+        available = True
+        try:
+            await asyncio.create_subprocess_exec(
+                b"ceph",
+                b"--version",
+                stdout=asyncio.subprocess.PIPE,
+                stderr=asyncio.subprocess.STDOUT,
+            )
```
kyujin-cho, line 37: Likewise, Ceph provides python binding of LibCephFS API. We can utilize this instead of parsing command output.

src/ai/backend/storage/cephfs/__init__.py
kyujin-cho: Well... How about wrapping BinarySize object with `int()`, then converting again to string by `str()`? Converting `BinarySize` to `int()` will result number of bytes it represents, so that will work.

changes/760.feature.md
```diff
@@ -0,0 +1,2 @@
+Add support for Ceph file system in Storage Proxy.
+Migrate to mono-repository.
```
kyujin-cho:
```suggestion
```
We don't have to tell users that we've migrated this PR to mono repo. Let's remove this line.

kyujin-cho (changes requested):

scripts/storage-proxy-spawn-ceph-clusters/install_storage_proxy.sh
leksikov: This is for installing into vagrant VM for testing purposes of Ceph. I remember I got some permission issue in vagrant if it wasn't set like this

kyujin-cho: Ah, I was completely misunderstanding context. If this is for installing storage-proxy implementations under Vagrant VM, then all these steps will be understandable. Please ignore my previous reviews on this script.

scripts/storage-proxy-spawn-ceph-clusters/install_storage_proxy.sh
leksikov: Yes. Users can install by themselves. This is done for the convenience of a developer who wants to spawn ceph for testing purposes only via Vagrant VM to simulate multiple nodes.  
However, I don't think the installation scripts spawn-ceph-cluster should be included in a final pull request.

src/ai/backend/storage/cephfs/__init__.py
leksikov: There is a need in better  Python command to display more clear file system usage.
df -h /mnt/vfroot/ceph-fuse command gives following output:
```
Filesystem      Size  Used Avail Use% Mounted on
ceph-fuse       5.7G     0  5.7G   0% /mnt/vfroot/ceph-fuse
```
The os.statvfs gives following output:
```
os.statvfs("/mnt/vfroot/ceph-fuse")
os.statvfs_result(f_bsize=4194304, f_frsize=4194304, f_blocks=1450, f_bfree=1450, f_bavail=1450, f_files=79, f_ffree=0, f_favail=0, f_flag=4102, f_namemax=255)
```

I think LibCephFS library has stat and can be better to use it to display.

src/ai/backend/storage/cephfs/__init__.py
leksikov: LibCephFS uses API. So, there should a running Ceph REST Server.
I haven't  seen API  functionality and if it is supported by default by Ceph cluster deployment script 'ceph-deploy' tool. 
By ```ceph --version``` I am checking if ceph client program is installed on Client node which can be utilized by Storage Proxy.

PR #761, opened 2022-10-01
fix: update ssh command which spawns outdated `backend.ai` commands

fregataa:
Update cli commands use outdated `backend.ai` commands


fregataa:
leave issue #763 and close this PR

Issue #742, opened 2022-09-25
Auto-generate and auto-commit news fragment from PR title


achimnol:
It seems better to describe that you need to `git pull` after creating the PR instead of to teach how to write news fragments for new developers and external cotnributors.

Let's do the followings:

* [ ] Auto-generate and auto-commit the news fragments when the PR title is changed using GitHub Actions.
  - Take the extension of news fragment files from the matching labels.
  - The workflow should report "failure" if there is no matching label.
  - We may need to reorganize the labels in the repo configuration.
  - When changed, update only the news fragment file that has the same content of previous PR title and the label-matching extension.
    - We may still need to manually insert additional news fragments (e.g., when there are breaking changes).
    - We may still need to add more details to the news fragment, while this action will provide a good initial value.

Let's change our rule as well:

* **Don't** add conventional-commit-style prefixes to the PR titles (e.g., `fix:`, `feat:`, etc.) and just use explicit PR labels.

Yaminyam:
I wonder if it is generating a news fragment based on the label included in the PR title or referring to the label of the labels function of github.

Yaminyam:
Replaced by #989

PR #762, opened 2022-10-02
feature: Auto-generate and auto-commit news fragment from PR title

soheeeeP:
issue #742 

#### Description
- Add the python script `scripts/update-news-fragment.py` to handle the news fragments.
  - Check existence of the news fragment from the matching labels 
  - Create the news fragment by first assigned label when it does not exist.
  - Updates the news fragment by using the changed PR title
- Update `timeline-check.yml` action workflow to generate auto-commit, auto-push the news fragments

#### Question
- I didn't find a clear approach to retrieve the content of the previous PR title, so updated the news fragment for all cases when the PR title changes.

### DIFF
--- a/.github/workflows/label-matcher.json
+++ b/.github/workflows/label-matcher.json
@@ -1,0 +1,272 @@
+[
+  {
+    "id": 1797645205,
+    "node_id": "MDU6TGFiZWwxNzk3NjQ1MjA1",
+    "url": "https://api.github.com/repos/lablup/backend.ai/labels/agent",
+    "name": "agent",
+    "color": "7c61cf",
+    "default": false,
+    "description": ""
+  },
+  {
+    "id": 4329308633,
+    "node_id": "LA_kwDOBC4ECs8AAAABAgwB2Q",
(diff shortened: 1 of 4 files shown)


Yaminyam:
If the ci works normally, shouldn't the news fragment of the pr be created automatically?

soheeeeP:
@Yaminyam I've designed the ci workflow to be activated when more than one label is assigned to the PR.

Yaminyam:
I thought that the label we were talking about in this issue was not the PR label, but the (eg ci:) label in the PR title.
You might want to check out this part!

soheeeeP:
I think PR from forked repository does not have permission to read the secrets of upstream repository. (like what happened in #687. Is there any possible ways to access secret?
image

Yaminyam:
> @Yaminyam I think PR from forked repository does not have permission to read the secrets of upstream repository. (like what happened in #687. Is there any possible ways to access secret? image

I think it's probably not.
In the case of me, if a secret is needed, I made a repo for testing and confirmed that it works normally and submitted the pr.

achimnol:
In favor of #989, I'd like to keep manual writing of news fragments because they can provide more details compared to PR titles.

PR #764, opened 2022-10-02
feat: new Agent option to (or not to) force terminate abusing kernels

adrysn:
To enable monitoring the abusing containers, reported by Watcher service, we need to set the `abuse-report-path` in `agent.toml`. However, setting this config implicitly allows Agent to automatically terminate abusing containers without admin's decision.

However, there sometimes is a need for admins to watch what the containers actually do before terminating them. To support this use case, this PR introduces a new Agent option, `force-terminate-abusing-containers` (default: `false`). By combining with `abuse-report-path` option, the following scenarios can be supported in the future:

| `abuse-report-path` | `force-terminate-abusing-containers` | Effect |
|---|---|---|
| Not set | `false` | Do not monitor abusing containers. |
| Not set | `true` | Do not monitor abusing containers (same as the above). |
| Set | `false` | Monitor abusing containers, but **do not automatically terminates them**. Admins may check the abusing containers, for example, by lablup/backend.ai-control-panel#371. This mode of operation was not possible before. |
| Set | `true` | Monitor abusing containers, and automatically terminates them as soon as detected. |


fregataa (approved):
Good.
How do you think of append  `_cleanup_reported_kernels` timer whether `force-terminate-abusing-containers` is true or false, and check the config in `_cleanup_reported_kernels` method.
Maybe this is good for debugging, cause we can see the logs say "I am inspecting abusing kernels but I am not going to remove them."

configs/agent/sample.toml
```diff
@@ -66,8 +66,10 @@ scaling-group = "default"
 # This directory is used to store the last registry file.
 # var-base-path = "/var/lib/backend.ai"
 
-# Optional, directory to look up abusing kernel reports generated by watchers
-abuse-report-path = "/tmp/backend.ai/abuse-reports"
+# The directory to look up abusing kernel reports generated by watchers
```
kyujin-cho, line 69: Can you write down more about those flags? Explaining about the matrix on PR description will be enough.

kyujin-cho (changes requested):

configs/agent/sample.toml
adrysn: Updated the docs.

kyujin-cho (approved):

PR #770, opened 2022-10-04
hotfix: vfolder API crashing

kyujin-cho:
Fix `vfolder_filter_by_status_required()` decorator appilied on inappropriate API handlers


src/ai/backend/manager/api/vfolder.py
```diff
@@ -102,8 +104,20 @@ async def _wrapped(request: web.Request, *args, **kwargs) -> web.Response:
...
+                folder_id = request.match_info["id"]
+                vf_name_conds = vfolders.c.id == folder_id
+            elif "name" in request.match_info:
+                folder_name = request.match_info["name"]
+                vf_name_conds = vfolders.c.name == folder_name
+            elif isinstance(args[1], dict) and "id" in args[1]:
+                folder_id = args[1]["id"]
+                vf_name_conds = vfolders.c.id == folder_id
+            elif isinstance(args[1], dict) and "name" in args[1]:
+                folder_name = args[1]["name"]
```
fregataa, line 117: ~~`args[1]` occurs IndexError when `invitations()`, `list_sent_invitations()` or `list_mounts()` is called.~~
~~Need to handle such cases.~~

My bad, we do not need to handle it cause you removed the wrapper on those methods.

fregataa (changes requested):
a alembic revision file is added and `server_default` argument of vfolder.status field is removed here.

fregataa (approved):
Check vfolder functions. and works well.

src/ai/backend/manager/api/vfolder.py
lizable: FWIW, It seems working properly, but It would be better to categorize the argument received from requests in a structural way. Let's leave it for now and make another issue for that and solve it later.

lizable (approved):
LGTM.

PR #771, opened 2022-10-04
fix: unhandled exception in `read_stream` when the `reply` is an empty list

adrysn:
In some cases, `read_stream` raises an unhandled exception like the below:
```python
  File "/home/bai/.pyenv/.../ai/backend/common/redis_helper.py", line 379, in read_stream
    for msg_id, msg_data in reply[0][1]:
IndexError:  list index out of range
```

In this case, the value of `reply` is an empty list, a.k.a, `[]`. I don't know the reproduction condition, but this exception leads to the silent exit of the `_subscribe_loop` task, which always stucks a session in the `PREPARING` status.

So, this PR makes the `read_stream` continue to read even when the `reply` is an empty list.

### DIFF
--- a/changes/771.fix
+++ b/changes/771.fix
@@ -1,0 +1,1 @@
+Prevent unhandled `read_stream` exception which leads to the stuck of newly created sessions in the `PREPARING` status.


kyujin-cho (approved):
LGTM.

PR #775, opened 2022-10-05
fix: skip resolving fields from dead agent

fregataa:
Managers try to resolve Agent's fields, whose status is not `ALIVE`.
Return `None` value before managers fetch RPC call to dead agents.

### DIFF
--- a/src/ai/backend/manager/models/agent.py
+++ b/src/ai/backend/manager/models/agent.py
@@ -597,7 +597,8 @@ class ModifyAgent(graphene.Mutation):
         set_if_set(props, data, "schedulable")
         set_if_set(props, data, "scaling_group")
         # TODO: Need to skip the following RPC call if the agent is not alive, or timeout.
-        await graph_ctx.registry.update_scaling_group(id, data["scaling_group"])
+        if (scaling_group := data.get("scaling_group")) is not None:
+            await graph_ctx.registry.update_scaling_group(id, scaling_group)
 
         update_query = sa.update(agents).values(data).where(agents.c.id == id)
         return await simple_db_mutate(cls, graph_ctx, update_query)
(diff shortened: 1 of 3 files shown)


adrysn (approved):
Tested and it works.

src/ai/backend/manager/models/agent.py
```diff
@@ -191,11 +191,15 @@ async def resolve_mem_cur_bytes(self, info: graphene.ResolveInfo) -> Any:
     async def resolve_hardware_metadata(
         self,
         info: graphene.ResolveInfo,
-    ) -> Mapping[str, HardwareMetadata]:
+    ) -> Mapping[str, HardwareMetadata] | None:
```
kyujin-cho: Just to follow our convention:
```suggestion
    ) -> Optional[Mapping[str, HardwareMetadata]]
```

src/ai/backend/manager/models/agent.py
```diff
@@ -191,11 +191,15 @@ async def resolve_mem_cur_bytes(self, info: graphene.ResolveInfo) -> Any:
...
+        if self.status != AgentStatus.ALIVE:
+            return None
         graph_ctx: GraphQueryContext = info.context
         return await graph_ctx.registry.gather_agent_hwinfo(self.id)
 
-    async def resolve_local_config(self, info: graphene.ResolveInfo) -> Mapping[str, Any]:
+    async def resolve_local_config(self, info: graphene.ResolveInfo) -> Mapping[str, Any] | None:
```
kyujin-cho: Same here. 
```suggestion
    async def resolve_local_config(self, info: graphene.ResolveInfo) -> Optional[Mapping[str, Any]]:
```

kyujin-cho (approved):

Issue #774, opened 2022-10-05
When single file downloaded, the file is tar archived and attached tar header.


studioego:
### What Operating System(s) are you seeing this problem on?

macOS (Apple Silicon)

### Backend.AI version

main

### Describe the bug

Using Backend.AI's download_single method, I tried to download a plain text file.

But, the single file includes tar headers.

Original file content is here

```
n513SRYAtMdGv4cNHcbGYHHADf9meBci
```
However, downloading a single file content, a tar header is attached

Here is a tar header attached to single file content.

```
.password0100644000000000000000000000004114314764573011121 0ustar0000000000000000n513SRYAtMdGv4cNHcbGYHHADf9meBci
```

![image](https://user-images.githubusercontent.com/20002/194014743-7fc9e07b-8b2c-4ec8-bbf0-0189dfcc09bb.png)

It doesn't need to add a tar header to download a single file from Backend.AI.

### To Reproduce

1. turn on Backend.AI app(Electron)
2. run the ssh app
3. download ssh key 
4. It found the tar header

### Expected Behavior

1. turn on Backend.AI app(Electron)
2. run the ssh app
3. download ssh key 
4. don't found the tar header

### Anything else?

If you want to download single file (such as plain text or key file) on Backend.AI agent part,
you need to remove tar header.

PR #778, opened 2022-10-06
Fix missing tar-unpack before downloading single file from container

studioego:
Unpack tar file before downloading the Docker session's selected file If you want to select Docker session's file, It'll be downloaded after packing to tar archive.

So, I unpacked tar archived the selected Docker session file, and then processed to download it.

close https://github.com/lablup/backend.ai/issues/774

### DIFF
--- a/src/ai/backend/agent/docker/kernel.py
+++ b/src/ai/backend/agent/docker/kernel.py
@@ -7,6 +7,7 @@ import os
 import re
 import shutil
 import subprocess
+import tarfile
 import textwrap
 from pathlib import Path, PurePosixPath
 from typing import Any, Dict, Final, FrozenSet, Mapping, Optional, Sequence, Tuple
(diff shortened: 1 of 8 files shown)


studioego:
Oh my see, I'll change the code structure.

src/ai/backend/agent/docker/kernel.py
```diff
@@ -245,7 +246,14 @@ async def download_file(self, filepath: str):
                     fsize = tarobj.fileobj.tell()
                     if fsize > 1048576:
                         raise ValueError("too large file")
-                    tarbytes = tarobj.fileobj.getvalue()
+                    tarobj.fileobj.seek(0)
+                    tar = tarfile.open(mode="r", fileobj=tarobj.fileobj)
+                    extarobj = tar.extractfile(tar.getnames()[0])
+                    if extarobj:
+                        tarbytes = extarobj.read()
```
achimnol: Again, we don't need to recreate the tarfile object since we already have it in the memory.
```suggestion
                    inner_file = tarobj.extractfile(tarobj.getnames()[0])
                    if inner_file:
                        tarbytes = inner_file.read()
```

achimnol (changes requested):

src/ai/backend/agent/docker/kernel.py
studioego: Resolved it.

achimnol (changes requested):
And another question: It seems to have rewritten the existing `download_file()`, which is also used for downloading multiple files from the container.  I don't see any extra options or separate interfaces to the download-file implementation.  Would there be any possibility to break the existing multi-file downloads?  Please test with the client SDK.

src/ai/backend/agent/kubernetes/kernel.py
```diff
@@ -260,6 +260,11 @@ async def download_file(self, filepath: str):
 
         return None
 
+    async def download_single(self, filepath: str):
+        # TODO: Implement download single file operations with pure Kubernetes API
+        log.error("Committing in Kubernetes is not supported yet.")
```
achimnol:
```suggestion
        log.error("download_single() in the k8s backend is not supported yet.")
```

studioego: Apply it

achimnol (approved):

Issue #759, opened 2022-09-30
Refactor agent's kernel creation routines using a state machine


achimnol:
It is hard to debug with `print()` statements when there are many external resourecs invoked with `await` statements chained as a long function body.

If we could model the steps to create a kernel using a state machine and implement debugging features on the base state machine, it will make our life much easier.  For example, just dumping out all transitions as logs will allow pinpointing where the process got stuck.

So, let's write a simple `ai.backend.common.automata.AsyncStateMachine` (or maybe a separate `aiostate` package) which executes per-state asyncio tasks by the state transition.  The state transition may be triggered upon the result of tasks like success, exception, cancelled, timeout, and max-retry-limit-exceeded.  Each state may have optional timeout or retry conditions.  State transition callbacks may include production of `KernelStartedEvent` or `KernelCancelledEvent` depending on the target state.  The state machine itself could serve as a supervisor-scope (currently known as `PersistentTaskGroup`) with a planned feature in `aiotools` that implements a context-scoped `asyncio.create_task()` override so that termination of the state machine guarantees termination of all involved, nested asyncio tasks.

Some idea sketch:

```python
from tenacity import AsyncRetrying, ...

state_machine = AsyncStateMachine(
  ctx={"event_dispatcher": event_dispatcher, ...},  # state_machine is injected when invoking states/transition callbacks
  states=[
    State("check-image", coro_factory=lambda ctx: check_image(ctx), retry=None),
    State("pull-image", coro_factory=lambda ctx: pull_image(ctx),
          retry=AsyncRetrying(stop=stop_after_attempts(2))),
    State("prepare-network", coro_factory=lambda ctx: prepare_network(ctx),
          retry=AsyncRetrying(stop=stop_after_timeout(30.0))),
    State("create-container", coro_factory=lambda ctx: create_container(ctx), retry=None),
    State("start-container", coro_factory=lambda ctx: start_container(ctx), retry=None),
    State("wait-kernel-runner", coro_factory=lambda ctx: wait_check_status(ctx),
          retry=AsyncRetrying(stop=stop_after_attempts(10), wait=wait_exponential(...)),
    State("cleanup", coro_factory=lambda ctx: remove_container_if_exists(ctx), retry=None),
    State("done", coro_factory=lambda ctx: None),  # may be declared as an intrinsic state
    State("failure", coro_factory=lambda ctx: None),  # may be declared as an intrinsic state
  ],
  transitions=[
[...]

PR #783, opened 2022-10-11
fix: Prevent session statuses from being rolled backed to PREPARING after successful creation or timeout

fregataa:
partially related: #759

1. apply `KERNEL_STATUS_TRANSITION_MAP` to prevent statuses from being rolled backed
2. handle `asyncio.CancelledError` and `asyncio.TimeoutError` in `create_kernels()`, by cancelling `kernel_creation_tracker`


src/ai/backend/manager/models/kernel.py
```diff
@@ -154,6 +155,77 @@ def default_hostname(context) -> str:
     return f"{params['cluster_role']}{params['cluster_idx']}"
 
 
+def determine_available_next_status(status: KernelStatus) -> List[KernelStatus]:
```
achimnol: Let's refactor this function as a constant table definition (a simple dict of set would be sufficient) instead of writing as code branches.

src/ai/backend/manager/registry.py
```diff
@@ -1542,6 +1549,12 @@ async def _create_kernels_in_one_agent(
                     kernel_id = KernelId(uuid.UUID(created_info["id"]))
                     self._post_kernel_creation_infos[kernel_id].set_result(created_info)
                 await asyncio.gather(*post_tasks, return_exceptions=True)
+            except (asyncio.TimeoutError, asyncio.CancelledError) as e:
+                for binding in items:
+                    kernel_id = binding.kernel.kernel_id
+                    self.kernel_creation_tracker[kernel_id].cancel()
```
achimnol, line 1572: 👍🏼

src/ai/backend/manager/models/kernel.py
```diff
@@ -154,6 +155,77 @@ def default_hostname(context) -> str:
...
+                KernelStatus.BUILDING,
+                KernelStatus.PULLING,
+            ]
+        case KernelStatus.RUNNING:
+            blocking_statuses = [
+                KernelStatus.PENDING,
+                KernelStatus.SCHEDULED,
+                KernelStatus.BUILDING,
+                KernelStatus.PULLING,
+            ]
```
achimnol: So here is the key. It ignores status update toward PREPARING if it's already RUNNING by checking this state transition table from `set_kernel_status()`. 👍🏼

src/ai/backend/manager/registry.py
```diff
@@ -1391,51 +1435,14 @@ async def _post_create_kernel(
             log.exception("post_create_kernel(k:{}) unexpected error", kernel_id)
             return
         else:
+            await self.finalize_running(
```
achimnol, line 1453: Here it should also use `execute_with_retry()`.

src/ai/backend/manager/api/session.py
```diff
@@ -1454,6 +1454,7 @@ async def handle_kernel_creation_lifecycle(
         # post_create_kernel() coroutines are waiting for the creation tracker events to be set.
         if (tracker := root_ctx.registry.kernel_creation_tracker.get(ck_id)) and not tracker.done():
             tracker.set_result(None)
+            await root_ctx.registry.finalize_running(event.creation_info)
```
achimnol: Wrap with `execute_with_retry()`.

src/ai/backend/manager/registry.py
```diff
@@ -1550,6 +1563,43 @@ async def _create_kernels_in_one_agent(
                     self.kernel_creation_tracker[kernel_id].cancel()
                     self._post_kernel_creation_infos[kernel_id].set_exception(e)
                 await asyncio.gather(*post_tasks, return_exceptions=True)
+                async with self.db.begin() as conn:
```
achimnol: All writing db transactions must be wrapped as a function and `execute_with_retry()`.  Could we reuse `set_kernel_status()` here?

achimnol (changes requested):
I think the overall logic looks good, but requires some attention to `execute_with_retry()` for write-involved database transactions and minor refactoring.

src/ai/backend/manager/registry.py
fregataa: `finalize_running` method use `execute_with_retry()` internally, so I do not wrap this.

adrysn (approved):
LGTM. I tested to create and destroy >400 sessions and found no PREPARING session. So, I think this PR provides at least equal stability with the main branch.

achimnol (approved):

PR #787, opened 2022-10-13
fix: return before rpc call if kernels are not allocated

fregataa:


adrysn (commented):
This is just a suggestion. Would it be plausible to check the parameters from the [`RPCContext`](https://github.com/lablup/backend.ai/blob/7657f2599a42eb032a496a0bf86c10e042ed0764/src/ai/backend/manager/registry.py#L195-L197)  itself. I think it is better for `RPCContext` not to send the request (or raise an exception) if one of the essential parameters are missing?

fregataa:
> This is just a suggestion. Would it be plausible to check the parameters from the [`RPCContext`](https://github.com/lablup/backend.ai/blob/7657f2599a42eb032a496a0bf86c10e042ed0764/src/ai/backend/manager/registry.py#L195-L197) itself.

It is plausible. We can raise a customized or built-in error in `RPCContext` if any of essential parameters are missing.
And just returning `None` rather raising error in `RPCContext` seems not plausible.

src/ai/backend/manager/registry.py
```diff
@@ -3023,7 +3027,11 @@ async def get_commit_status(
         session_name_or_id: Union[str, SessionId],
         access_key: AccessKey,
     ) -> Mapping[str, str]:
-        kernel = await self.get_session(session_name_or_id, access_key)
+        kernel = await self.get_session(session_name_or_id, access_key, allow_stale=True)
+        if kernel["status"] != KernelStatus.RUNNING:
+            raise InvalidAPIParameters(
+                f'Unable to get commit status since kernel(id: {kernel["id"]}) is currently not RUNNING.'
+            )
```
kyujin-cho: Please use double quotation mark! 
```suggestion
                f"Unable to get commit status since kernel(id: {kernel["id"]}) is currently not RUNNING."
            )
```

src/ai/backend/manager/registry.py
```diff
@@ -3045,7 +3053,11 @@ async def commit_session(
         Commit a main kernel's container of the given session.
         """
 
-        kernel = await self.get_session(session_name_or_id, access_key)
+        kernel = await self.get_session(session_name_or_id, access_key, allow_stale=True)
+        if kernel["status"] != KernelStatus.RUNNING:
+            raise InvalidAPIParameters(
+                f'Unable to commit since kernel(id: {kernel["id"]}) is currently not RUNNING.'
+            )
```
kyujin-cho: Same here.
```suggestion
            raise InvalidAPIParameters(
                f"Unable to commit since kernel(id: {kernel["id"]}) is currently not RUNNING."
            )
```

kyujin-cho (changes requested):

PR #788, opened 2022-10-13
fix: correct agent status compare in resolver

fregataa:


adrysn (commented):
I think `self.status` is better to be an `AgentStatus` type instead of `str` since it is an internal object. However, it is assigned as a string from [here](https://github.com/lablup/backend.ai/blob/5601749fe7c8c3130ca94a7c4435447d7f074644/src/ai/backend/manager/models/agent.py#L135). I don't know this is due to some historical reasons or other, but it might be better to keep consistency in handling objects internally.

kyujin-cho:
> I think `self.status` is better to be an `AgentStatus` type instead of `str` since it is an internal object. However, it is assigned as a string from [here](https://github.com/lablup/backend.ai/blob/5601749fe7c8c3130ca94a7c4435447d7f074644/src/ai/backend/manager/models/agent.py#L135). I don't know this is due to some historical reasons or other, but it might be better to keep consistency in handling objects internally.

Agree; We need to change `self.status` being assigned by Enum's name, not changing these lines.

fregataa:
> I think `self.status` is better to be an `AgentStatus` type instead of `str` since it is an internal object. However, it is assigned as a string from [here](https://github.com/lablup/backend.ai/blob/5601749fe7c8c3130ca94a7c4435447d7f074644/src/ai/backend/manager/models/agent.py#L135). I don't know this is due to some historical reasons or other, but it might be better to keep consistency in handling objects internally.

This is the difference. The following is using `status` as it.
image

And this is when I change it into string. (`status.name`)
image

I guess it is reasonable to parse it to string. How do you think?
@kyujin-cho @adrysn

kyujin-cho:
Note: `self.status` is a `Graphene.String()` type, which is used to represent GraphQL column.

kyujin-cho (approved):

Issue #753, opened 2022-09-28
Implement vfolder-host access control


achimnol:
There are some user scenarios that admins want to allow a specific project X members to mount read-only vfolders from the vfolder host A while they can freely create/use vfolders in the vfolder host B—and they want to **prohibit** the project X member's _vfolder creation on the vfolder host A_ at the same time.

Currently, we only have the per-vfolder access control table (`vfolder_permissions`) and the access permission to vfolder hosts as a simple boolean set (merged from `domains`, `groups`, and `keypair_resource_policies`). As such, it is impossible to define a "read-only" permission to a specific vfolder host like that users may mount vfolders in the host but may not create new vfolders there.

The problem:
Since we have `allowed_vfolder_hosts` defined in 3 different objects (`domains`, `groups`, and `keypair_resource_policies`), we need to first define how to merge them when each object has different properties like below:

```
domains("domainA").vfolder_host_permissions = {"storge1": {"access": "rw"}}
groups("groupX").vfolder_host_permissions = {"storage1": {"access": "ro"}}
# or
domains("domainA").vfolder_host_permissions = {"storage1": {"access": "ro"}}
groups("groupX").vfolder_host_permissions = {"storage1": {"access": "rw"}}
# or, when the user belongs to both groupX and groupY
groups("groupX").vfolder_host_permissions = {"storage1": {"access": "ro"}}
groups("groupY").vfolder_host_permissions = {"storage1": {"access": "rw"}}
```

achimnol:
One idea by @fregataa:
- Let's add a property to specify whether "overriding by more specific objects" is allowed or not. (e.g., `groups("groupY").vfolder_host_permissions = {"storage1": {"access": "rw", "allow_override": True}}`)

achimnol:
Another way is to split the permission level ("ro" or "rw") into fine-grained actions like generic RBAC:
```
domains("domainA").vfolder_host_permissions = {"storge1": ["mount"]}
groups("groupX").vfolder_host_permissions = {"storage1": ["mount", "create-vfolder"]}
keypair_resource_policies("researchers").vfolder_host_permissions = {"storage1": ["mount", "create-vfolder", "delete-vfolder"]}
-> merged result: {"storage1": ["mount", "create-vfolder", "delete-vfolder"]}
```

achimnol:
A good discussion about RBAC and ReBAC, with role-explosion issue, presented in PyCon KR (Korean): https://youtu.be/DXf_rQtpLZQ

PR #790, opened 2022-10-13
feat: implement atomic vfolder host permission

fregataa:
ref: https://github.com/lablup/backend.ai/issues/753#issuecomment-1260603825

1. Migrate `domains.allowed_vfolder_hosts`, `groups.allowed_vfolder_hosts`, `keypair_resource_policies.allowed_vfolder_hosts` from ARRAY of string to JSONB

etc) 
```
# Before
domains.allowed_vfolder_hosts = ["local:volume1"]
domains.allowed_vfolder_hosts = ["local:volume2"]

# After
domains.allowed_vfolder_hosts = {"local:volume1": ["mount", "create-vfolder"]}
domains.allowed_vfolder_hosts = {"local:volume2": ["read-vfolder", "create-vfolder"]}
```

2. Impl `PermissionListColumn` which can be used by other **permission** spec columns.


kyujin-cho:
Please set PR status to "Ready for review" when you've done the final touches.

fregataa:
@yomybaby @lizable @Sujin-Kim1 
There should be update in frontend since `allowed_vfolder_hosts` field of `Domain`, `Group`,`KeyPairResourcePolicy` graphene and db schema will change in this PR.
There is an alternative, just letting `allowed_vfolder_hosts` as a legacy field and adding `allowed_vfolder_host_map` JSON field.

Please give me your opinion.

achimnol (commented):
We need to add an API to retrieve the list of possible permission names for the web UI and control panel, to avoid duplicated constant definitions.

src/ai/backend/manager/models/acl.py
```diff
@@ -0,0 +1,36 @@
...
+)
+
+
+def get_all_permission() -> Mapping[str, Any]:
+    return {
+        "vfolder_permission_list": [perm.value for perm in VFolderHostPermission],
+    }
+
+
+class Permission(graphene.ObjectType):
```
kyujin-cho: `Permission` only represents capability of vFolder host feature. Please rename it relevantly to a new name which represents only permission of vFolder hosts.

src/ai/backend/manager/models/acl.py
```diff
@@ -0,0 +1,36 @@
...
+    from .gql import GraphQueryContext
+
+
+__all__: Sequence[str] = (
+    "Permission",
+    "get_all_permission",
+)
+
+
+def get_all_permission() -> Mapping[str, Any]:
```
kyujin-cho: Plural suggestion!
```suggestion
def get_all_permissions() -> Mapping[str, Any]:
```

src/ai/backend/manager/models/acl.py
```diff
@@ -0,0 +1,36 @@
...
+    return {
+        "vfolder_permission_list": [perm.value for perm in VFolderHostPermission],
```
kyujin-cho: This should be vFolder "host", not just vFolder.
```suggestion
        "vfolder_host_permission_list": [perm.value for perm in VFolderHostPermission],
```

kyujin-cho (changes requested):

src/ai/backend/manager/models/acl.py
fregataa: Ok. I am going to rename it. I intended to add more permissions in this **general** graphene object when we implement ACL in the future.

src/ai/backend/client/cli/admin/acl.py
```diff
@@ -0,0 +1,32 @@
...
+    Permission administration commands.
+    """
+
+
+@permission.command()
+@pass_ctx_obj
+def list(ctx: CLIContext) -> None:
+    """
+    List atomic permissions.
+    (admin privilege required)
```
kyujin-cho: As this command is already a subset of `admin` group I think it's okay to remove this message.
```suggestion
```

src/ai/backend/client/cli/admin/domain.py
```diff
@@ -62,7 +62,10 @@ def list(ctx: CLIContext) -> None:
 @click.option("-i", "--inactive", is_flag=True, help="New domain will be inactive.")
 @click.option("--total-resource-slots", type=str, default="{}", help="Set total resource slots.")
 @click.option(
-    "--allowed-vfolder-hosts", type=str, multiple=True, help="Allowed virtual folder hosts."
+    "--allowed-vfolder-hosts",
```
kyujin-cho, line 65: We need to tell user how to fill out this parameter. After reading actual implementation it looks like the code expects string formatted as JSON syntax to be filled here, so let's tell user that way.

src/ai/backend/client/cli/admin/domain.py
```diff
@@ -118,9 +121,7 @@ def add(
 @click.option("--description", type=str, help="Description of the domain")
 @click.option("--is-active", type=bool, help="Set domain inactive.")
 @click.option("--total-resource-slots", type=str, help="Update total resource slots.")
-@click.option(
-    "--allowed-vfolder-hosts", type=str, multiple=True, help="Allowed virtual folder hosts."
-)
+@click.option("--allowed-vfolder-hosts", type=str, help="Allowed virtual folder hosts.")
```
kyujin-cho: Same here.

src/ai/backend/client/cli/admin/group.py
```diff
@@ -88,7 +88,10 @@ def list(ctx: CLIContext, domain_name) -> None:
 @click.option("-i", "--inactive", is_flag=True, help="New group will be inactive.")
 @click.option("--total-resource-slots", type=str, default="{}", help="Set total resource slots.")
 @click.option(
-    "--allowed-vfolder-hosts", type=str, multiple=True, help="Allowed virtual folder hosts."
+    "--allowed-vfolder-hosts",
```
kyujin-cho, line 91: Also here.

src/ai/backend/client/cli/admin/group.py
```diff
@@ -143,9 +146,7 @@ def add(
 @click.option("-d", "--description", type=str, help="Description of the group")
 @click.option("--is-active", type=bool, help="Set group inactive.")
 @click.option("--total-resource-slots", type=str, help="Update total resource slots.")
-@click.option(
-    "--allowed-vfolder-hosts", type=str, multiple=True, help="Allowed virtual folder hosts."
-)
+@click.option("--allowed-vfolder-hosts", type=str, help="Allowed virtual folder hosts.")
```
kyujin-cho: Same here.

src/ai/backend/client/cli/admin/resource_policy.py
```diff
@@ -93,10 +94,25 @@ def list(ctx):
...
     help="The maximum period of time allowed for kernels to wait " "further requests.",
 )
-# @click.option('--allowed-vfolder-hosts', type=click.Tuple(str), default=['local'],
-#               help='Locations to create virtual folders.')
 @click.option(
-    "--allowed-vfolder-hosts", default=["local"], help="Locations to create virtual folders."
+    "--allowed-vfolder-hosts",
+    default=json.dumps(
+        {
+            "local": [
```
kyujin-cho: Hmm, shouldn't it be `volume1:local`?

kyujin-cho (changes requested):

src/ai/backend/client/cli/admin/resource_policy.py
fregataa: but the original default value was `["local"]`. Is it good to change?

kyujin-cho: Yes. The value added on database has format of `host:volume`, so the example should follow that rule also.

src/ai/backend/client/func/acl.py
```diff
@@ -0,0 +1,36 @@
...
+    ) -> Sequence[str]:
+        """
+        Fetches the list of atomic permissions.
+
+        :param fields: Additional permission query fields to fetch.
+        """
+        query = textwrap.dedent(
+            """\
+            query {
+                permissions {$fields}
```
lizable: According to the query schema defined [here](https://github.com/lablup/backend.ai/pull/790/files#diff-56a2d9f1fc07827156313028f4414a518eb3f14bc56739d16d6bbb3e32956019R490-R493), the query name should be changed to "vfolder_host_permission", not "permissions".

src/ai/backend/client/func/acl.py
```diff
@@ -0,0 +1,36 @@
...
+            }
+        """
+        )
+        query = query.replace("$fields", " ".join(f.field_ref for f in fields))
+        data = await api_session.get().Admin._query(query)
+        return data["permissions"]
```
lizable: Of course, the return type should be changed, too.
`permissions` to `vfolder_host_permissions`.

lizable (changes requested):
I found that there are a few updates on vfolder host permissions.
It's such a small update, but crucial for those who need to use that API.

src/ai/backend/client/func/keypair_resource_policy.py
```diff
@@ -54,7 +54,7 @@ async def create(
         max_vfolder_count: int,
         max_vfolder_size: int,
         idle_timeout: int,
-        allowed_vfolder_hosts: Sequence[str],
+        allowed_vfolder_hosts: str = None,
         fields: Iterable[FieldSpec | str] = None,
```
lizable, line 58: Btw, as long as I know,  both`total_resource_slots` and `allowed_vfolder_hosts` have the same type (stringified JSON). Therefore, the type definition for `total_resource_slots` needs to be changed likewise.

src/ai/backend/client/func/keypair_resource_policy.py
```diff
@@ -102,7 +102,7 @@ async def update(
         max_vfolder_count: int,
         max_vfolder_size: int,
         idle_timeout: int,
-        allowed_vfolder_hosts: Sequence[str],
+        allowed_vfolder_hosts: str = None,
     ) -> dict:
```
lizable, line 106: Here as well. The type definition of `total_resource_slots` is `str`, not `int`.

lizable (changes requested):
IMO, there are a few requests written in the comments and also one more thing to be fixed before merging this PR.
TL;DR: We need to change the code in here(`src/ai/backend/manager/model/gql.py#1428`)
https://github.com/lablup/backend.ai/blob/40c04807d9b03040218b39ead81262ff483e79f2/src/ai/backend/manager/models/gql.py#L1415
to this
`mutation_cls = getattr(Mutations, info.field_name).type`

Considering the overall key `allowed_vfolder_hosts` are located in several policies (such as `domains`, `groups(project groups`, and `keypair_resource_policies`), I think It's quite crucial to show the overview of access in those policies easily.
In light of this, It's not surprising that users may want features for updating those access controls at once. Therefore, I decided to chain mutations so that It reduces(not completely, but some) repetitive requests to manager server. To make chained mutations, it needs [aliasing](https://atheros.ai/blog/how-to-use-graphql-aliases). Unfortunately, In current code, It regards aliasing string as one of the variables in Mutations class, which would definitely throw an error. I checked many graphql requests and It doesn't affect any mutations at all.
It would be appreciated you to apply the new code described above.

lizable:
> IMO, there are a few requests written in the comments and also one more thing to be fixed before merging this PR. TL;DR: We need to change the code in here(`src/ai/backend/manager/model/gql.py#1428`)
> 
> https://github.com/lablup/backend.ai/blob/40c04807d9b03040218b39ead81262ff483e79f2/src/ai/backend/manager/models/gql.py#L1415
> 
> 
> to this
> `mutation_cls = getattr(Mutations, info.field_name).type`
> Considering the overall key `allowed_vfolder_hosts` are located in several policies (such as `domains`, `groups(project groups`, and `keypair_resource_policies`), I think It's quite crucial to show the overview of access in those policies easily. In light of this, It's not surprising that users may want features for updating those access controls at once. Therefore, I decided to chain mutations so that It reduces(not completely, but some) repetitive requests to manager server. To make chained mutations, it needs [aliasing](https://atheros.ai/blog/how-to-use-graphql-aliases). Unfortunately, In current code, It regards aliasing string as one of the variables in Mutations class, which would definitely throw an error. I checked many graphql requests and It doesn't affect any mutations at all. It would be appreciated you to apply the new code described above.

It's better to handle this in as a separate issue(#879), which is already opened by @yomybaby .

src/ai/backend/client/func/keypair_resource_policy.py
lizable: @fregataa FWIW, shouldn't be `total_resource_slots` and `allowed_vfolder_hosts` to be the same type?

lizable: You already applied it [here](https://github.com/lablup/backend.ai/pull/790/files#diff-4f594e20f517e33a816d296b20b711339b333fdfbbaf42dfa95c648682cac272R104).

lizable (changes requested):
@fregataa 
There's a minor request about correcting type definition on the parameter.
Except that, It works fine.

src/ai/backend/client/func/keypair_resource_policy.py
fregataa: I read how keypair_resource_policy mutation works and it seems `total_resource_slots` should not be `None` because the field determine the policy. but I guess `total_resource_slots` can be `None` so I leave it Optional.

Plus, group and domain are different from keypair_resource_policy. they can have `total_resource_slots = None`

lizable: I get it now. It should be left as is.

lizable (approved):
LGTM.
Btw I dunno why a couple of irrelevant test codes failed, but I guess we can ignore this.
And also, I guess we need one more approval from reviewers (probably @kyujin-cho or @achimnol  or @adrysn ) since this change is epic.

kyujin-cho (approved):

PR #791, opened 2022-10-13
fix: `/folders/_/usage` API not working on privileged call

kyujin-cho:
This PR fixes vFolder API crashing when vFolder reference is passed via query params, and lets admin allow access to all vFolders regardless of ownership when using `query_accessible_vfolders()`.

### DIFF
--- a/src/ai/backend/manager/api/vfolder.py
+++ b/src/ai/backend/manager/api/vfolder.py
@@ -110,12 +110,19 @@ def vfolder_filter_by_status_required(perm: VFolderAccessStatus):
             elif "name" in request.match_info:
                 folder_name = request.match_info["name"]
                 vf_name_conds = vfolders.c.name == folder_name
-            elif isinstance(args[1], dict) and "id" in args[1]:
-                folder_id = args[1]["id"]
+            elif "id" in request.query:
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/api/vfolder.py
```diff
@@ -105,17 +105,24 @@ async def _wrapped(request: web.Request, *args, **kwargs) -> web.Response:
...
+            elif "name" in request.query:
+                query_folder_name = request.query["name"]
+                vf_name_conds = vfolders.c.name == query_folder_name
+            elif len(args) > 1 and isinstance(args[1], dict):
+                if body_folder_id := args[1].get("id"):
+                    vf_name_conds = vfolders.c.id == body_folder_id
+                elif body_folder_name := args[1].get("name"):
+                    vf_name_conds = vfolders.c.name == body_folder_name
+                else:
+                    raise VFolderFilterStatusFailed("either vFolder id nor name not supplied")
```
adrysn: When the `id` or `name` is delivered via `POST` body parameter, this logic raises the `VFolderFilterStatusFailed`. We also need to take into account for the body parameter.

adrysn (changes requested):

PR #804, opened 2022-10-20
feat: Improve cpuset detection to use Docker's cgroup cpuset

### LINKED ISSUE #803: Auto-detect and exclude exclusively reserved CPU cores
In some cases, Linux cgroups may have "exclusively assigned" CPU cores to a specific cgroup (usually an I/O-intensive filesystem agents), causing Docker to crash when trying to create a container with overlapping CPU cores, like:

```
runc create failed: unable to start container process:
  unable to apply cgroup configuration:
  failed to write "0,1,2,...":
  write /sys/fs/cgroup/cpuset/docker/xxxxx/cpuset.cpus:
  permission denied: unkown
```

Reference: moby/moby#29496

We need to use the explicitly cgroup cpuset cores assigned to the Docker service, [instead of taking the `NCPU` field from the local docker info API](https://github.com/lablup/backend.ai/blob/main/src/ai/backend/agent/docker/intrinsic.py#L116).

Example:
```console
# cat /sys/fs/cgroup/cpuset/docker/cpuset.cpus
0,3-15,18-31,34-47,50-64,67-79,82-95,98-111,114-127
```

achimnol:
resolves #803

- refactor: Remove unnecessary dependency to `local_config` pytest fixture which
  causes popuplation of temporary etcd/redis containers


achimnol:
I'm now thinking about how to improve the abstraction to embrace both internal CPU plugin overrides from the license plugin while keeping consistent CPU set retrieval interface for both Docker and K8s backends.

achimnol:
I will make a separate issue for the refactoring.

kyujin-cho (approved):

PR #820, opened 2022-10-22
fix: return empty commit status when the session is not running

adrysn:


fregataa (approved):

kyujin-cho:
This PR will be a partial rollback from #787.

kyujin-cho (approved):

PR #824, opened 2022-10-22
docs: Readme.md for update Badges

### LINKED ISSUE #793: Update outdated `README.md` badges
### Main idea

Since our README‘s badges related to PyPI package points `backend.ai` package, which is abandoned after we’ve migrated to mono-repo, is referencing outdated informations.
IMG_0070
I think we can resolve it by just change it to point `backend.ai-manager`.

### Alternative ideas

_No response_

### Anything else?

_No response_

hekang42:
close #793 
@lizable
@kyujin-cho

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 Backend.AI
 ==========
 
-[![PyPI release version](https://badge.fury.io/py/backend.ai.svg)](https://pypi.org/project/backend.ai/)
-![Supported Python versions](https://img.shields.io/pypi/pyversions/backend.ai.svg)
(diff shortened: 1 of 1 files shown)


lizable (approved):
Thx for updating our README!
It seems like there's an unexpected error for versioning python we need to update our metadata. Anyway, we'll fix it.

kyujin-cho:
Merged. We still need to figure out way to set supported Python version correctly (not just a ambiguous `3`) when generating setup.cfg. This will be done on a separate PR.

PR #827, opened 2022-10-22
feat: Support `.sessionenv` to auto-populate environment variables in client CLI 

### LINKED ISSUE #799: Support `.sessionenv` to auto-populate environment variables in client CLI
Let's use `dotenv` to automatically read the current working directory's `.sessionenv` file (or the one designated by a CLI option) to pass multiple environment variables to a new compute session with ease.

upa-r-upa:
Resolve #799 

Now, users can inject options through `.sessionenv` without entering any command options when using the `start` and `run` commands.

### changes

- Add `auto_envvar_prefix` option in `./backend-ai start`, `./backend-ai run` command.
  - `.env` config file name is `.sessionenv` and prefix is `BACKEND_CLI`.


kyujin-cho:
Please convert PR to "Ready for review" if you're ready to do so.

upa-r-upa:
> Please convert PR to "Ready for review" if you're ready to do so.

Hello!
This PR does not work properly yet to read dotenv.
I'm going to solve it quickly.

upa-r-upa:
@kyujin-cho Please review!

achimnol (changes requested):
It's a wrong implementation and I think you have misunderstood the issue.  We should not load `.sessionenv` into the client process's environment variables.

Instead, the PR should just read `.sessionenv` in the client (so in this case we don't have to use `dotenv` because we are not going to inject it into `os.environ`), pass it to the manager via the session creation API's arguments, and update the environment variable list passed to the kernel creation RPC call.  The goal is to ensure the container processes have the env-vars set from the client-side `.sessionenv`.

The environment variables are passed as-is and do not interfere with the client-side environment variables, so we *don't need* any special prefix like you did (`CLI_ENV_PREFIX`).

achimnol:
Closing as stale. We will start a new PR.

PR #830, opened 2022-10-22
Added installation guide for wsl2

ArnavScripts:


lizable (changes requested):
Hmm, I think you just misunderstood sth. It doesn't mean that we need to install WSL2 here.
All we need is a complete installation of Backend.AI Cluster in WSL2, not install WSL2 itself.
Please refer to the document and apply it with screenshots if needed.

achimnol:
Replaced with #1722.

PR #840, opened 2022-10-26
fix: logical errors in calculating monthly stats

### LINKED ISSUE #839: The monthly stat value is not accurate.

macOS (Apple Silicon)

### Backend.AI version

22.09.4

### Describe the bug

[`get_time_binned_monthly_stats()`](https://github.com/lablup/backend.ai/blob/main/src/ai/backend/manager/api/resource.py#L550) has a logical error, such as `num_sessions` not being counted properly even though the session is being used at that time. For example, if a user has used a session for the time period from 10:00 to 17:00, the resource quota should remain floating during that time. However, it is currently counted only at the session start time. It needs to modify it to calculate the proper stat.

### To Reproduce

_No response_

### Expected Behavior

Both the number and resources of the session must be properly combined during the session run time.

### Anything else?

_No response_

Sujin-Kim1:
this resolves #839 
Modify calculating monthly stats logic so that number of sessions and resource usage is calculated appropriately during all session execution times.

### DIFF
--- a/src/ai/backend/manager/api/resource.py
+++ b/src/ai/backend/manager/api/resource.py
@@ -585,11 +585,9 @@ async def get_time_binned_monthly_stats(request: web.Request, user_uuid=None):
         rows = result.fetchall()
 
     # Build time-series of time-binned stats.
-    rowcount = len(rows)
     now_ts = now.timestamp()
     start_date_ts = start_date.timestamp()
     ts = start_date_ts
-    idx = 0
     tseries = []
     # Iterate over each time window.
     while ts < now_ts:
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/api/resource.py
```diff
@@ -601,29 +601,27 @@ async def get_time_binned_monthly_stats(request: web.Request, user_uuid=None):
...
+                if "cuda.shares" in row.occupied_slots:
+                    gpu_allocated += Decimal(row.occupied_slots["cuda.shares"])
+                raw_stat = await redis_helper.execute(
+                    root_ctx.redis_stat, lambda r: r.get(str(row["id"]))
+                )
+                if raw_stat:
+                    last_stat = msgpack.unpackb(raw_stat)
+                    io_read_bytes += int(nmget(last_stat, "io_read.current", 0))
+                    io_write_bytes += int(nmget(last_stat, "io_write.current", 0))
+                    disk_used += int(nmget(last_stat, "io_scratch_size/stats.max", 0, "/"))
```
fregataa:
```suggestion
        for row in rows:
            if ts + time_window <= row.created_at.timestamp() or ts >= row.terminated_at.timestamp():
                continue
            num_sessions += 1
            cpu_allocated += int(row.occupied_slots.get("cpu", 0))
            mem_allocated += int(row.occupied_slots.get("mem", 0))
            if "cuda.devices" in row.occupied_slots:
                gpu_allocated += int(row.occupied_slots["cuda.devices"])
            if "cuda.shares" in row.occupied_slots:
                gpu_allocated += Decimal(row.occupied_slots["cuda.shares"])
            raw_stat = await redis_helper.execute(
                root_ctx.redis_stat, lambda r: r.get(str(row["id"]))
            )
            if raw_stat:
                last_stat = msgpack.unpackb(raw_stat)
                io_read_bytes += int(nmget(last_stat, "io_read.current", 0))
                io_write_bytes += int(nmget(last_stat, "io_write.current", 0))
                disk_used += int(nmget(last_stat, "io_scratch_size/stats.max", 0, "/"))
```
We can simply iterate the rows.

fregataa (changes requested):
Let's replace `while` to `for ... in` statement

src/ai/backend/manager/api/resource.py
agatha197: reflected! thanks :)

fregataa (approved):
lgtm

Issue #496, opened 2022-06-27
Running multiple agents on a single node


achimnol:
To run two or more Backend.AI agents on a single GPU node, we need to do the followings:

* [x] Have a config option to use separate IPC socket directories. (lablup/backend.ai-agent#347)
* [ ] Have localized (i.e., in the `agent.toml` file) device mask config options for all compute plugins
* [ ] Distinguish which agent instance owns a running kernel container when rescanning them and when processing the Docker event (lifecycle events) stream
  - We need to add a new label when creating the container to tag which agent owns that container.
* [ ] Test and improve our CUDA plugins to handle statistics from non-owned containers
* [ ] Q: Should we have separate watchers for each agent instance or a unified one?

achimnol:
refs #208

fregataa:
1. https://github.com/lablup/backend.ai/pull/624 
Is this PR able to resolve second item of this issue - have localized device mask config options?

2. https://github.com/lablup/backend.ai/pull/712
And this PR can resolve third item of this issue - distinguish which agent instance owns a running kernel container?
I just let agents filter out the kernels which is not registered in agent's kernel registry because I think that one kernel cannot have multiple kernels which spawn by different agents.

achimnol:
> 1. [feat: allow passing blocklist to compute plugin ctx read from local config #624](https://github.com/lablup/backend.ai/pull/624)
>    Is this PR able to resolve second item of this issue - have localized device mask config options?

I think we should have "allowlist" approach instead of "blocklist" for more human-friendly configuration for multi-agent on single-node.

> 2. [feat: skip containers owned by other agents in the same host #712](https://github.com/lablup/backend.ai/pull/712)
>    And this PR can resolve third item of this issue - distinguish which agent instance owns a running kernel container?
>    I just let agents filter out the kernels which is not registered in agent's kernel registry because I think that one kernel cannot have multiple kernels which spawn by different agents.

#712 is not what I want. We should attach explicit labels to conatiners when creating them to distinguish the owner agent. Ultimately I'm going to remvoe `kernel_registry` at all and migrate to the reconcilation loop design.

PR #841, opened 2022-10-27
setup: Add `--ipc-base-path` option to install-dev.sh

achimnol:
When installing two or more agents/dev-setups in a single node, we need to be able to customize `ipc-base-path` of the manager and agent.

This PR also changes the manager's default `ipc-base-path` from `/tmp/backend.ai/manager/ipc` to `/tmp/backend.ai/ipc` for consistency.  As long as the manager instances are gracefully restarted upon updates, there should be no problems as the only content in the manager's IPC directory is the unix socket for relaying log records.

refs #496 -- Note that this PR does not implement the full requirements of #496 but only 
covers the update of `install-dev.sh` while assuming that the Docker daemon is separated.

It's a follow-up to lablup/backend.ai-agent#347.

### DIFF
--- a/changes/841.misc.md
+++ b/changes/841.misc.md
@@ -1,0 +1,1 @@
+Add `--ipc-base-path` option to `install-dev.sh` to ease setting up multiple development environments in a single node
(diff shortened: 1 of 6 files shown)


inureyes (approved):
LGTM

PR #845, opened 2022-11-01
fix: Change `ExitCode` into `IntEnum` to remove unintended console output upon errorneous exits

### LINKED ISSUE #844: `ExitCode` is printed on CLI terminal

macOS (Apple Silicon)

### Backend.AI version

main

### Describe the bug

image

`ExitCode` is printed on terminal unintentionally.

### To Reproduce

You can simply raise an error in any CLI command method.
e.g) Raise an `KeyError` in vfolder list cmd like below.
image

### Expected Behavior

Do not print `ExitCode`

### Anything else?

_No response_

Yaminyam:
close #844 
When sys.exit() receives a non-Int value as an argument, it is judged as a message and printed.
So, I modified the argument to receive enum.IntEnum.
refs. https://docs.python.org/3/library/sys.html#sys.exit

### DIFF
--- a/changes/845.fix.md
+++ b/changes/845.fix.md
@@ -1,0 +1,1 @@
+Modify to receive `enum.IntEnum` as an argument so that `sys.exit()` does not output a message.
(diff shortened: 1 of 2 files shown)


rapsealk:
The code looks good. However, I humbly recommend rewriting the changelog to clarify **WHAT** you modified. It might be `ExitCode` in this case. Each changelog should be understandable by itself without reading any changes in the source code.

achimnol (approved):

PR #849, opened 2022-11-02
fix: Serialization failures when returning kernel creation results

achimnol:
Taken from https://github.com/lablup/backend.ai/pull/491/commits/39ab5756717af673570ba81002a41df2b41948ce (part of #491)

This PR improves it further by introducing [msgpack extended types](https://github.com/msgpack/msgpack-python#extended-types), allowing seamless serialization of UUIDs, enums, datetimes, Path, etc.

### DIFF
--- a/changes/849.fix.md
+++ b/changes/849.fix.md
@@ -1,0 +1,1 @@
+Fix serialization failures when returning kernel creation results with additional accelerator device information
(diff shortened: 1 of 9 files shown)


kyujin-cho:
@achimnol test is failing, please fix it

achimnol:
According to [the temporenc docs](https://temporenc.readthedocs.io/en/latest/#temporenc.Moment.datetime), temporenc is timezone-aware and will return UTC values by default.

kyujin-cho (approved):

PR #851, opened 2022-11-03
fix:handle kernel creation failure and destroying fail-to-create kernels

fregataa:


fregataa:
close this PR since it was made from wrong branch

PR #852, opened 2022-11-03
fix: Handle kernel creation failure and destroy failed kernel

fregataa:
fix: #555


src/ai/backend/agent/agent.py
```diff
@@ -865,6 +854,17 @@ async def _handle_destroy_event(self, ev: ContainerLifecycleEvent) -> None:
                     if ev.done_future is not None:
                         ev.done_future.set_exception(e)
                     raise
```
achimnol, line 858: What happens if the control goes here?
I think we should put the moved block inside a `finally` block following the above `try` block.

src/ai/backend/agent/docker/agent.py
```diff
@@ -762,6 +762,7 @@ async def start_container(
                     config=container_config, name=kernel_name
                 )
                 cid = container._id
+                raise RuntimeError
```
achimnol: This should be removed! 😉

src/ai/backend/agent/docker/agent.py
```diff
@@ -778,9 +779,13 @@ async def start_container(
                             await writer.write(f"{k}={v}\n")
 
                 await container.start()
-            except asyncio.CancelledError:
-                raise
-            except Exception:
+            except asyncio.CancelledError as e:
+                try:
+                    cid = container._id
+                except (UnboundLocalError, AttributeError):
```
achimnol: Could we remove this new inner `try` block?
I think this is the issue of control flow, so we could initialize `container` as `None` before entering the topmost `try` block and explicitly check if it's `None` here.

src/ai/backend/agent/docker/agent.py
```diff
@@ -794,7 +799,11 @@ async def start_container(
                 async with self.resource_lock:
                     for dev_name, device_alloc in resource_spec.allocations.items():
                         self.computers[dev_name].alloc_map.free(device_alloc)
-                raise
+                try:
+                    cid = container._id
+                except (UnboundLocalError, AttributeError):
```
achimnol: Same here.

achimnol (changes requested):

src/ai/backend/agent/agent.py
fregataa: I see.
I did not know that this flow will not put the `CLEAN` event in the queue.

src/ai/backend/agent/agent.py
```diff
@@ -112,7 +112,7 @@
 
 from . import __version__ as VERSION
 from . import resources as resources_mod
-from .exception import AgentError, ResourceError
+from .exception import AgentError, DockerContainerCreationError, ResourceError
```
kyujin-cho: `DockerContainerCreationError`, name of the error, is too specific for docker backend. How about changing it to `ContainerCreationError`, without Docker?
```suggestion
from .exception import AgentError, ContainerCreationError, ResourceError
```

kyujin-cho (changes requested):

src/ai/backend/agent/agent.py
fregataa: I renamed it. Thanks!

src/ai/backend/agent/agent.py
```diff
@@ -1590,12 +1592,28 @@ async def create_kernel(
...
+                preopen_ports,
+            )
+        except ContainerCreationError as e:
+            log.warning(
+                "Kernel failed to create container (k:{}). Kernel is going to be destroyed.",
+                ctx.kernel_id,
+            )
+            cid = e.container_id
+            async with self.registry_lock:
+                self.kernel_registry[ctx.kernel_id]["container_id"] = cid
```
kyujin-cho, line 1644: Please add `container_id` to the `AbstractKernel` definition.

src/ai/backend/agent/exception.py
```diff
@@ -34,6 +34,12 @@ class UnsupportedBaseDistroError(RuntimeError):
     pass
 
 
+class ContainerCreationError(Exception):
+    def __init__(self, *args, container_id, **kwargs):
```
kyujin-cho: `*args` should appear before positional argument `container_id`.
```suggestion
    def __init__(self, container_id, *args, **kwargs):
```

fregataa: I locate `container_id` after positional argument cause I intended to set `container_id` only by keyword argument. is it okay?

fregataa: I updated it to `def __init__(self, container_id, *args, **kwargs):`

PR #855, opened 2022-11-04

### LINKED ISSUE #678: Let's group string-literal `KernelEvent.reason` values into a manageable format

qkoo0833:
1. add class KernelLifecycleEventReason in common/events.py to manage kernelEvent.reason

3. several cases are managed by method

close #678


rapsealk:
@qkoo0833 Please write a changelog for this PR! This [changes/README.md](https://github.com/lablup/backend.ai/blob/main/changes/README.md) will help you.

src/ai/backend/agent/agent.py
```diff
@@ -1198,7 +1201,7 @@ def _rm(path: Path) -> None:
                             KernelId(UUID(body["ID"])),
                             ContainerId(body["CID"]),
                             LifecycleEvent.DESTROY,
-                            body.get("reason", "anomaly-detected"),
+                            KernelLifecycleEventReason.get_or_anomaly_detected(body.get("reason")),
```
kyujin-cho: `get_or_anomaly_detected()` is too specific for this use case only. I think it's better to maintain original structure.
```suggestion
                            body.get("reason", KernelLifecycleEventReason.ANOMALY_DETECTED),
```

src/ai/backend/manager/idle.py
```diff
@@ -188,7 +189,10 @@ async def _do_idle_check(
                             session["id"],
                         )
                         await self._event_producer.produce_event(
-                            DoTerminateSessionEvent(session["id"], f"idle-{checker.name}"),
+                            DoTerminateSessionEvent(
+                                session["id"],
+                                KernelLifecycleEventReason.idle_checker_name(checker.name),
```
kyujin-cho: Same here.
```suggestion
                                f"idle-{checker.name}",
```

kyujin-cho (changes requested):

src/ai/backend/agent/agent.py
qkoo0833: @kyujin-cho Since I changed the attr types of some related classes from str to just KernelLifecycleReasonEvent, this is where the problem occurs during type checking. I think that some class's type hint would be changed or additional handling would be needed.  what could be better?

rapsealk: You can wrap the value `body.get("reason")` with enum class. How about this? At least this will help you keep type consistency.

```python
class KernelLifecycleEventReason(str, enum.Enum):  # enum.StrEnum is coming in version 3.11.
  @classmethod
  def from_value(cls, value: str) -> Optional["KernelLifecycleEventReason"]:
    try:
      return cls(value)
    except ValueError:
      pass
    return None

...

KernelLifecycleEventReason.from_value(body.get("reason")) or KernelLifecycleEventReason.ANOMALY_DETECTED
```

kyujin-cho (approved):

Issue #749, opened 2022-09-27
Scalability consideration for abusing reports


kyujin-cho:
### Main idea

This is a (potential) follow-up issue for #737.     
The downside of current implementation is that manager need to perform RPC call for each container. When number of container rises, this can cause a performance issue to both manager and agent. We need to fix this.   
Some possible options are:
- make agent scan abusing report path periodically, and update redis (or any other state manager) if new container is found
- same, but use heartbeat as a report channel

Any future ideas are welcome.

fregataa:
how about adding abusing report info in any RPC call response from agent?

PR #858, opened 2022-11-07
refactor: store abusing reports in redis

fregataa:
resolves: #749 
- Save `abusing_report_path` and `abusing_container_auto_terminate` config values in manager's DB
- store abusing reports in redis so that manager don't have to call RPC for every kernels


src/ai/backend/agent/agent.py
```diff
@@ -726,6 +772,10 @@ async def heartbeat(self, interval: float):
                     msgpack.packb([(repo_tag, digest) for repo_tag, digest in self.images.items()])
                 ),
                 "architecture": get_arch_name(),
+                "abuse_report_path": self.local_config["agent"].get("abuse-report-path"),
```
kyujin-cho: Hmm, is it safe to let msgpack serialize Path object by itself? I think the library isn't capable of handling it...

src/ai/backend/manager/models/agent.py
```diff
@@ -83,6 +83,8 @@ class AgentStatus(enum.Enum):
     sa.Column("version", sa.String(length=64), nullable=False),
     sa.Column("architecture", sa.String(length=32), nullable=False),
     sa.Column("compute_plugins", pgsql.JSONB(), nullable=False, default={}),
+    sa.Column("abuse_report_path", sa.String(length=64), nullable=True),
```
kyujin-cho: Maximum length of 64 can be a little bit short to represent full path. How about setting it to 128?

src/ai/backend/agent/agent.py
```diff
@@ -696,6 +702,46 @@ async def produce_error_event(
...
+                str(kernel_id2): CommitStatus.ONGOING.value,
+            },
+            str(subdir2): ...,
+        }
+        """
+        base_commit_path: Path = self.local_config["agent"]["image-commit-path"]
+        status_map = {}
+        for subdir in base_commit_path.iterdir():
+            ongoing_list = [commit_path.name for commit_path in subdir.glob("./**/lock/*")]
+            status_map[str(subdir)] = {kern: CommitStatus.ONGOING.value for kern in ongoing_list}
```
kyujin-cho: I don't see any counterpart logic of this line on this PR. We need to set back this value to `READY` when container commit is completed, otherwise the API will wait infinitely even if actual commit process is done.

kyujin-cho (changes requested):

src/ai/backend/agent/agent.py
fregataa: I updated the counterpart, by checking remaining kernel status map and update them to `READY` if they do not have lock file

src/ai/backend/manager/registry.py
```diff
@@ -2691,6 +2696,8 @@ async def _update() -> None:
                                 "version": agent_info["version"],
                                 "compute_plugins": agent_info["compute_plugins"],
                                 "architecture": agent_info.get("architecture", "x86_64"),
+                                "abuse_report_path": abuse_report_path,
+                                "auto_terminate": auto_terminate,
```
kyujin-cho: Is there a specific reason to store these config values to database instead of retriving on-demand with RPC?

src/ai/backend/agent/agent.py
```diff
@@ -696,6 +702,51 @@ async def produce_error_event(
...
+                str(kernel_id1): CommitStatus.ONGOING.value,
+                str(kernel_id2): CommitStatus.ONGOING.value,
+            },
+            str(subdir2): ...,
+        }
+        """
+        base_commit_path: Path = self.local_config["agent"]["image-commit-path"]
+        status_map = {}
+        for subdir in base_commit_path.iterdir():
+            ongoing_list = [commit_path.name for commit_path in subdir.glob("./**/lock/*")]
```
kyujin-cho: I know it is really easy to forget, but please note that `Path.glob` also is a blocking function. Depending on total count of committed containers and disk performance this can cause a potential performance issue, so let's wrap `subdir.glob` with `loop.run_in_executor()`.

kyujin-cho (changes requested):

src/ai/backend/manager/registry.py
fregataa: Storing these values in db seems better than retriving them with RPC cause the values do not change that much. They only change when the agent's status changes. But using RPC needs to retrieve them with RPC for every request.
If this idea is bad, I will change this code. How do you think?

src/ai/backend/agent/agent.py
fregataa: update this part. thank you!

kyujin-cho:
Some of the changes updated at this PR will now be discussed at #1015.

src/ai/backend/manager/models/agent.py
```diff
@@ -88,6 +88,7 @@ class AgentStatus(enum.Enum):
     sa.Column("version", sa.String(length=64), nullable=False),
     sa.Column("architecture", sa.String(length=32), nullable=False),
     sa.Column("compute_plugins", pgsql.JSONB(), nullable=False, default={}),
+    sa.Column("auto_terminate", sa.Boolean(), nullable=False, server_default=true(), default=False),
```
achimnol: I think the column name is too vague. How is "auto_terminate" different from "batch-type session"?
The naming should include hints for abusing.

achimnol (changes requested):

kyujin-cho (approved):

PR #866, opened 2022-11-10
feat: allow manager to load selected plugin.

### LINKED ISSUE #716: Let manager choose which plugins to load
### Main idea

This idea was first introduced in our [agent implementation](https://github.com/lablup/backend.ai/blob/main/configs/agent/sample.toml#L63). After filling out this value on `agent.toml` config, agent will load plugins only described under this config. We need same feature for manager also.

### Alternative ideas

_No response_

### Anything else?

_No response_

qkoo0833:
This PR add support for loading just selected plugins with manager.

resolve #716

### DIFF
--- a/changes/866.feature.md
+++ b/changes/866.feature.md
@@ -1,0 +1,1 @@
+Load plugins by reading allowlists from manager local configuration.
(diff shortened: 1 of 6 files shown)


kyujin-cho (changes requested):
Adding whitelist to manager plugin loading requires more than just adding configurable option to manager's local config. Please refer [agent side implementation](https://github.com/lablup/backend.ai/blob/e8329478bd2ec499db859cb87826d2094f8fa9f0/src/ai/backend/agent/docker/resources.py#L50) to see how agent accepts whitelist option and applies to plugin loading structure.

kyujin-cho (approved):
LGTM.

Issue #809, opened 2022-10-21
Unable to see invited and project virtual (data) folders with CLI


adrysn:
### What Operating System(s) are you seeing this problem on?

Linux (aarch64)

### Backend.AI version

22.9.2

### Describe the bug

With `backend.ai-client` package, we can see the available virtual folder list by `backend.ai vfolder list`. However, it only displays the virtual folders that the user created, not the invited and project folders.

### To Reproduce

* Assume there is a user account `user@backend.ai`.
* With another account, create a user folder and share (invite) the folder to the `user@backend.ai`.
* With an admin account, create a project folder.
* With `user@backend.ai` account, execute the following command:
  ```shell
  backend.ai vfolder list
  ```
* It will display the user's own folders, not invited, and project folders.

For example, from Web-UI, I was able to see invited, and project folders as well as my own folders:
<img width="732" alt="image" src="https://user-images.githubusercontent.com/7539358/197223469-a0fe2375-204f-4309-a22b-593198796eed.png">

However, in CLI, I only saw the virtual folders that I have created:
<img width="1257" alt="image" src="https://user-images.githubusercontent.com/7539358/197223513-5b97c031-74e5-4ebe-a681-696b0522ad51.png">

### Expected Behavior

With the CLI command, we need to see invited and project folders as well.

achimnol:
This is superseded by #1955.

PR #868, opened 2022-11-10
fix: Let vfolder resolver fetch permission overriden vfolders and fetch vfolders of groups the user belongs to.

fregataa:
resolves: #809


src/ai/backend/manager/models/group.py
```diff
@@ -193,6 +196,48 @@ async def resolve_groups(
...
+
+
+async def query_user_accessible_groups(
+    db_conn: SAConnection,
+    user_uuid: uuid.UUID,
+) -> Iterable[uuid.UUID]:
+    """
+    Query all groups the user belongs to.
+    """
+    from .user import users
```
kyujin-cho: Importing package every time when executing function will create an overhead. Let's move this line to outside of this function.

src/ai/backend/manager/models/group.py
```diff
@@ -193,6 +196,48 @@ async def resolve_groups(
...
+    db_conn: SAConnection,
+) -> None:
+    """
+    Query all groups.
+    This function returns `None` since super admin can access all groups
+    and fetching all groups from DB is a big overhead.
+    """
+    return None
+
+
```
kyujin-cho, line 214: Is there a reason to leave non-functioning function in code?

src/ai/backend/manager/models/vfolder.py
```diff
@@ -784,22 +786,39 @@ async def load_count(
...
+        j = (
+            sa.join(
+                vfolders,
+                vfolder_permissions,
+                vfolders.c.id == vfolder_permissions.c.vfolder,
+                isouter=True,
+            )
+            .join(users, vfolders.c.user == users.c.uuid, isouter=True)
+            .join(groups, vfolders.c.group == groups.c.id, isouter=True)
+        )
```
kyujin-cho: JOINing triple tables in a single query is a really, really bad idea. High possibilities of slow query. Two queries with no or single join will definitely be much faster when table grows much. Please refactor this query.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -814,24 +833,46 @@ async def load_slice(
...
-            groups, vfolders.c.group == groups.c.id, isouter=True
+        j = (
+            sa.join(
+                vfolders,
+                vfolder_permissions,
+                vfolders.c.id == vfolder_permissions.c.vfolder,
+                isouter=True,
+            )
+            .join(users, vfolders.c.user == users.c.uuid, isouter=True)
+            .join(groups, vfolders.c.group == groups.c.id, isouter=True)
```
kyujin-cho, line 843: Same here.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -843,8 +884,8 @@ async def load_slice(
         async with graph_ctx.db.begin_readonly() as conn:
             return [
                 obj
-                async for r in (await conn.stream(query))
-                if (obj := cls.from_row(graph_ctx, r)) is not None
+                for row in [r async for r in (await conn.stream(query))]
+                if (obj := cls.from_row(graph_ctx, row)) is not None
             ]
```
kyujin-cho: These lines are not that readable. Can you flat out this block, avoiding inline list constructions so that everyone can easily read without any burden?

kyujin-cho (changes requested):

src/ai/backend/manager/models/group.py
fregataa: I want `query_accessible_groups()` method for all User roles. This function does not do anything but it seems normalized and readable code for me. How do you think? I will remove it if it does not make sense.

src/ai/backend/manager/models/vfolder.py
fregataa: I remove joins on this code since they are not necessary. thanks!

src/ai/backend/manager/models/vfolder.py
fregataa: The original code make 3 joins and I did not add or remove any join here because 3 joins does not seem to have bad performance.
How about adding index on email column of users, name column of groups and permission column of vfolder_permissions for performance on this query?

src/ai/backend/manager/models/group.py
kyujin-cho: Adding comment that indicates superadmin can access all groups in place of this empty function seems enough.

fregataa:
close this PR since we will use graphQL more and https://github.com/lablup/backend.ai/pull/1437 will cover the need of graphQL.

PR #873, opened 2022-11-13
feat: Upgrade krunner packages

achimnol:
Let's upgrade its base Python version and dependency packages.


adrysn (approved):
LGTM. Tested and works well with backend.ai-krunner-static-gnu==3.0.0.

Issue #879, opened 2022-11-14
GraphQL mutation request with alias doesn't work.


yomybaby:
### What Operating System(s) are you seeing this problem on?

macOS (Apple Silicon)

### Backend.AI version

22.09.04

### Describe the bug

GraphQL mutation request with alias doesn't work.

### To Reproduce

```graphql
mutation {
  a: modify_group(gid: "1de2b969-0000-0000-0000-0bc8adb3c831" props: {
     name: "default"
   }) {
     ok
      group {
        id
        name
        description
      }
   }
   b: modify_group(gid: "1de2b969-1111-0000-0000-0bc8adb3c831" props: {
     name: "default"
   }) {
     ok
      group {
        id
        name
        description
      }
   }
}
```
Actual result

```
{
  "type": "https://api.backend.ai/probs/graphql-error",
  "title": "GraphQL-generated error.",
  "data": [
    {
      "message": "type object 'Mutations' has no attribute 'a'",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "a"
      ]
    },
    {
      "message": "type object 'Mutations' has no attribute 'b'",
      "locations": [
        {
          "line": 12,
          "column": 3
        }
      ],
      "path": [
        "b"
      ]
    }
  ]
}
```

### Expected Behavior

result format:

```json
{
  "a": {
    "ok": false,
    "group": null
  },
  "b": {
    "ok": false,
    "group": null
  }
}
```

PR #882, opened 2022-11-15
fix: Get field_name from ResolveInfo in gql middleware

fregataa:
resolves: #879 

Get mutation name from `ResolveInfo.field_name` rather from `ResolveInfo.path[0]`, and this enable to use alias.
https://github.com/graphql-python/graphql-core/blob/main/src/graphql/type/definition.py#L613-L635
`ResolveInfo` has `field_name.

### DIFF
--- a/changes/882.fix.md
+++ b/changes/882.fix.md
@@ -1,0 +1,1 @@
+Get mutation name from ResolveInfo.field_name in gql mutation middleware.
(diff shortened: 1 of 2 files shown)


yomybaby (approved):
LGTM!

lizable (approved):
LGTM.

PR #885, opened 2022-11-16
feat: create human-friendly random name for session name

### LINKED ISSUE #734: Let's adopt human-friendly random name generator for session name
### Main idea

Backend.AI client generates a random session name when `session_name` is not given.
It generates random names with prefix "pysdk-".
Let's adopt a human friendly random string generator like how Docker generate names of containers.

### Alternative ideas

We can create or adopt a random generator which combines or refers dinosaur names.

### Anything else?

_No response_

qkoo0833:
this PR support for generating docker-style name randomly when session_name is not given.

resolve #734


src/ai/backend/client/namegenerator.py
```diff
@@ -0,0 +1,360 @@
...
+    "zhukovsky",
+]
+
+
+def get_random_name() -> str:
+    return f"{random.choice(left)}_{random.choice(right)}"
+
+
+def duplicated_name(name: str) -> str:
+    return f"{name}{random.randrange(0, 10)}"
```
fregataa: the name of the function is ambiguous because what this function does is appending random integer for duplicated Session names.
Let's rename it into a more intuitive name like `append_random_int` etc

src/ai/backend/client/cli/session.py
```diff
@@ -270,6 +267,13 @@ def create(
         assigned_agent_list = assign_agent
         with Session() as session:
             try:
+                if name is None:
+                    result = session.ComputeSession.paginated_list(fields=[session_fields["name"]])
+                    names = [item["name"] for item in result.items]
+                    if (name := get_random_name()) in names:
+                        name = duplicated_name(name)
```
fregataa: It is a good idea to duplication check here, but we do not need to stay session name as unique.
And this loop iterates all the existing session names.

src/ai/backend/client/func/session.py
```diff
@@ -252,7 +253,10 @@ async def get_or_create(
         if name is not None:
             assert 4 <= len(name) <= 64, "Client session token should be 4 to 64 characters long."
         else:
-            name = f"pysdk-{secrets.token_hex(5)}"
+            result = await ComputeSession.paginated_list(fields=[session_fields["name"]])
+            names = [item["name"] for item in result.items]
+            if (name := get_random_name()) in names:
+                name = duplicated_name(name)
```
fregataa: same here

fregataa (changes requested):
Please assign a reviewer when you are ready for review.
Or you can change this PR into a draft.
And how about using [random-word generator](https://github.com/vaibhavsingh97/random-word)? I need more opinions and reviews on this suggestion using 3rd party module.

qkoo0833:
> Please assign a reviewer when you are ready for review. Or you can change this PR into a draft. And how about using [random-word generator](https://github.com/vaibhavsingh97/random-word)? I need more opinions and reviews on this suggestion using 3rd party module.

@fregataa I'll assign reviewer, thx.
I think that the 3rd party module has pros&cons.
SmartSelectImage_2022-11-17-11-57-25
That module include file I/O and coverting from raw data to list. so, little slow.
But, It will generate various name because it has much more amount of words than docker-style.
Or, I can look for other 3rd party modules.
Which would be more suitable? Any opinion is welcome.

src/ai/backend/client/cli/session.py
```diff
@@ -256,7 +257,8 @@ def create(
                runtime or programming language.
         """
         if name is None:
-            name = f"pysdk-{secrets.token_hex(5)}"
+            fake = Faker()
+            name = f"{fake.first_name()}_{fake.last_name()}".lower()
```
kyujin-cho: How about keeping `pysdk-` prefix to match policy with webui?

kyujin-cho (changes requested):

src/ai/backend/client/cli/session.py
```diff
@@ -256,7 +257,8 @@ def create(
                runtime or programming language.
         """
         if name is None:
-            name = f"pysdk-{secrets.token_hex(5)}"
+            fake = Faker()
```
achimnol: Rename this to `faker` as it is a subjective.

src/ai/backend/client/func/session.py
```diff
@@ -252,7 +253,8 @@ async def get_or_create(
         if name is not None:
             assert 4 <= len(name) <= 64, "Client session token should be 4 to 64 characters long."
         else:
-            name = f"pysdk-{secrets.token_hex(5)}"
+            fake = Faker()
```
achimnol: Same here.

achimnol (changes requested):
The code looks okay, but I'm afraid that just using `faker.last_name()` may introduce name duplication. Could we use the full name with hyphens as the separator?  It may be even better to use `faker.user_name()`.

achimnol (approved):

Issue #396, opened 2022-03-21
Add `session show-graph` command to visualize session dependencies


achimnol:
It would be useful to have a visualization of related rows from the `session_dependencies` table in CLI.

kin4496:
@achimnol please give me more information about table?

jopemachine:
I think this issue should be closed.

PR #886, opened 2022-11-17
feat: Add `session show-graph` command to visualize session dependencies

jopemachine:
This PR implements #396's `session show-graph` command.

## Use case

```
$./backend.ai session show-graph <SESSID>
```

`<SESSID>` could be `session_id` or `session_name`.

## Test case

I tested the command through the below graph with t5 session id.

스크린샷 2022-11-18 오전 10 04 49

## Result

```
t5
├── t4_1
│   └── t3_1
│       └── t2_1
│           └── t1_1
├── t4_2
│   └── t3_1
│       └── t2_1
│           └── t1_1
└── t4_3
    └── t3_2
        └── t2_2
            └── t1_1
```


jopemachine:
## Result 1

```
$  ./backend.ai session show-graph <SESSID>
```



## Result 2

```
$  ./backend.ai session show-graph --table <SESSID>
```

src/ai/backend/manager/api/session.py
```diff
@@ -2178,6 +2180,111 @@ async def shutdown_service(request: web.Request, params: Any) -> web.Response:
     return web.Response(status=204)
 
 
+async def find_dependency_sessions(
```
kyujin-cho: To make sense in english:
```suggestion
async def find_dependent_sessions(
```

src/ai/backend/manager/api/session.py
```diff
@@ -2178,6 +2180,111 @@ async def shutdown_service(request: web.Request, params: Any) -> web.Response:
...
+    root_session_name_or_id: Union[uuid.UUID, str],
+    db_connection: SAConnection,
+    access_key: AccessKey,
+):
+    dependency_session_cache: Dict[str, List] = {}
+
+    async def _find_dependency_sessions(session_name_or_id: Union[uuid.UUID, str]):
```
kyujin-cho: Is there a specific reason to create inner function with every logic included?

src/ai/backend/manager/api/session.py
```diff
@@ -2178,6 +2180,111 @@ async def shutdown_service(request: web.Request, params: Any) -> web.Response:
...
+@server_status_required(READ_ALLOWED)
+@auth_required
+async def get_dependency_graph(request: web.Request) -> web.Response:
+    root_ctx: RootContext = request.app["_root.context"]
+    root_session_name = request.match_info["session_name"]
+
+    requester_access_key, owner_access_key = await get_access_key_scopes(request)
+
+    log.info(
+        "SHOW_DEPENDENCY_GRAPH (ak:{0}/{1}, s:{2})",
```
kyujin-cho: It'll be better to match phrase printed on log with actual method name.
```suggestion
        "GET_DEPENDENCY_GRAPH (ak:{0}/{1}, s:{2})",
```

src/ai/backend/manager/api/session.py
```diff
@@ -2178,6 +2180,111 @@ async def shutdown_service(request: web.Request, params: Any) -> web.Response:
...
+            map(
+                lambda result: result.mappings().all(),
+                await asyncio.gather(
+                    db_connection.execute(session_dependency_query),
+                    db_connection.execute(kernel_query),
+                ),
+            )
+        )
+
+        dependency_session_ids = list(
```
kyujin-cho:
```suggestion
        dependent_session_ids = list(
```

src/ai/backend/manager/api/session.py
```diff
@@ -2178,6 +2180,111 @@ async def shutdown_service(request: web.Request, params: Any) -> web.Response:
...
+            map(lambda x: str(x.get("depends_on")), session_dependency_query_result)
+        )
```
kyujin-cho: This will be much more neat.

```suggestion
        dependency_session_ids = [
            str(x.get("depends_on"))
            for x in session_dependency_query_result
        ]
```

src/ai/backend/manager/api/session.py
```diff
@@ -2178,6 +2180,111 @@ async def shutdown_service(request: web.Request, params: Any) -> web.Response:
...
+            return session_info
+
+        session_info["depends_on"] = await asyncio.gather(
+            *map(
+                lambda dependency_session_id: asyncio.create_task(
+                    _find_dependency_sessions(dependency_session_id)
+                ),
+                dependency_session_ids,
+            )
+        )
```
kyujin-cho: Is there an advantage to creating and gathering multiple tasks? It'll make exception handling trickier. Also, if you have specific reason to do so, please add ways to handle potential exceptions which can be raised on `_find_dependency_sessions()`.

src/ai/backend/manager/api/session.py
```diff
@@ -2178,6 +2180,111 @@ async def shutdown_service(request: web.Request, params: Any) -> web.Response:
     return web.Response(status=204)
 
 
+async def find_dependency_sessions(
+    root_session_name_or_id: Union[uuid.UUID, str],
+    db_connection: SAConnection,
+    access_key: AccessKey,
+):
+    dependency_session_cache: Dict[str, List] = {}
```
kyujin-cho: How about using ready-to-use methods, like [aiotools.lru_cache()](https://aiotools.readthedocs.io/en/latest/aiotools.func.html#aiotools.func.lru_cache) instead implementing our own caching method?

kyujin-cho (changes requested):

src/ai/backend/manager/api/session.py
jopemachine: In the test case of this issue, this function with t5 session id at first finds t4_1, t4_2, t4_3, since t4_1, t4_2, t4_3 are required to execute t5.
So, I think maybe t4_1, t4_2, and t4_3 are dependency of t5 session,
and t5 is a dependent session of t4_1, t4_2, and t4_3.

src/ai/backend/manager/api/session.py
jopemachine: 1. I wanted to exclude `db_connection`, `access_key` from the arguments of the recursive function explicitly because they should be the same object as the initial argument in all recursive functions.

2. `dependency_session_cache` could be assigned when the `find_dependency_sessions` is called and could be accessed in all recursive calls without increasing argument counts. and it also removes passing an empty dictionary from the callsite.

src/ai/backend/manager/api/session.py
jopemachine: I thought `db_connection.execute` is a relatively high-cost operation and should be executed recursively in this function, so I wanted them not to block the Backend.ai manager's other operations while running a bunch of `db_connection.execute`.

But actually, I'm a little confused about python asyncio model yet. For that purpose, is forking a thread and running the operations from that thread separately better? Is this code cannot ensure concurrency with the Backend.ai manager code?

jopemachine:
@kyujin-cho
 
I think the below code might be fragile.

I assumed the session_name is always the form of the below string, but I'm not sure about it.

`(pipline_name)-(job_name)-l0Z1g1dd-ek`

https://github.com/lablup/backend.ai/pull/886/files#diff-617867996601b356e12de71696162ef8390144f3ba6aaba12da78cb7efabf4f7R947

Splitting with `pipeline_name` would be better if I could get the `pipeline_name`, but I'm not sure it is possible.

src/ai/backend/manager/api/session.py
```diff
@@ -2179,6 +2181,92 @@ async def shutdown_service(request: web.Request, params: Any) -> web.Response:
     return web.Response(status=204)
 
 
+async def find_dependency_sessions(
+    root_session_name_or_id: Union[uuid.UUID, str],
+    db_connection: SAConnection,
+    access_key: AccessKey,
+):
+    @aiotools.lru_cache(maxsize=100)
+    async def _find_dependency_sessions(session_name_or_id: Union[uuid.UUID, str]):
```
kyujin-cho: Adding `lru_cache()` decorator to inner function will be meaningless since this function will be redeclared every time outer function is called. Please declare it as a separate outer function.

kyujin-cho (changes requested):

jopemachine:
I closed this PR by mistake while I'm working on it, so I made new PR. https://github.com/lablup/backend.ai/pull/1532

PR #892, opened 2022-11-22
fix: Typo in runner BUILD file for Apple intel mac

qkoo0833:
This PR support for modifying typo in runner.BUILD dependency map table.


achimnol (approved):
Good catch!

PR #896, opened 2022-11-23
fix: Set correct vfolder host value in install script

fregataa:
PR #790 made error in install script.
This pr prevents from JSON parsing error when setting `allowed_vfolder_host` values in `domains`, `groups` and `users` tables


rapsealk (approved):
LGTM!

PR #900, opened 2022-11-27
feat: remove usage of attr decorator on AbstractComputeDevice

kyujin-cho:


kyujin-cho:
Merging this will require accelerator plugins to have v21 entrypoint.

PR #905, opened 2022-11-30
fix: Invalid default value of the agent's affinity-policy config option

achimnol:
Follow-up of #491


achimnol:
The CI failures are due to too fast merge and the missing merge base commit when Pants ran. Safe to ignore.

PR #909, opened 2022-12-01
hotfix: Rewrite NUMA-aware allocation to support 3 or more NUMA nodes

achimnol:
When allocating n-th step compute devices using the allocation result
of the (n-1)-th step compute devices, we need to include the n-th step
compute devices when filtering them with the zero distance to (n-1)-th
step compute devices allocated in the prior step.

This is a follow-up fix to #491.


adrysn:
Tested this with every possible combination of CPU cores and GPUs in a 128 cores / 8 GPU system, and found no problem.

PR #914, opened 2022-12-05
fix: Make nullability of `scaling_group.use_host_network` column consistent with actual schema

fregataa:
alter scaling_group's `use_host_network` to not nullable, because schema and real model has different nullability of `use_host_network` column


achimnol (commented):
Don't we need explicit migration to alter the database schema?

fregataa:
I added the context of this pr

achimnol (approved):

PR #915, opened 2022-12-06
fix: Add missing `await` for a jupyter-client API when shutting down the kernel runner

rapsealk:
I faced a `RuntimeWarning` message while I was running a python-kernel,
```sh
python-kernel: Service ttyd (pid: 62) has terminated with exit code: 0
/opt/backend.ai/lib/python3.10/site-packages/ai/backend/kernel/base.py:262: RuntimeWarning: coroutine 'KernelManager._async_is_alive' was never awaited
  if self.kernel_mgr and self.kernel_mgr.is_alive():
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
python-kernel: shutting down python kernel...
```
and found there is a missing `await` in our code.

You may refer the links below:
https://github.com/jupyter/jupyter_client/blob/641d60a9efb199cd8263cd45dc4a82016c21945c/jupyter_client/manager.py#L673
https://github.com/jupyter/jupyter_client/blob/641d60a9efb199cd8263cd45dc4a82016c21945c/jupyter_client/manager.py#L624

This is a follow-up fix to #873.


fregataa (approved):

achimnol (approved):

achimnol:
I think in this case it's better to have an explicit changelog as it was an explicit bug.

PR #916, opened 2022-12-06
feat: agent gather process stats in container

### LINKED ISSUE #696: Container process monitoring
Let's add a monitoring feature of container processes.
The superadmin and the user should be able to inspect to what all/their individual containers are doing.

* [ ] **agent**: Add a periodic timer to retrieve active kernel container's process list with their basic resource usage such as CPU, memory, IO bytes/sec, etc. Store them in an auto-expiring Redis key.
  - The timer may be combined with the existing statistics infra, but the process list should be a live snapshot instead of a time-series.
  - We need to exclude some system and krunner processes for brevity.
  - Could we combine the GPU stats as well?
* [ ] **manager**: Add a GraphQL field to `ComputeContainer` to fetch this data from Redis.
* [ ] **webui**: Add a UI to the session container list to inspect running processes.
* [ ] **control-panel**: Same to webui.

qkoo0833:
This PR support that agent gather process list and process resource usage.
1. Collect process's cpu, mem, I/O usage stats using proc filesystem or 3rd party module.
2. Timer retrieve stats.
3. exclude to collect process stats from root user.
resolve #696

### DIFF
--- a/src/ai/backend/agent/docker/intrinsic.py
+++ b/src/ai/backend/agent/docker/intrinsic.py
@@ -256,18 +256,6 @@ class CPUPlugin(AbstractComputePlugin):
     async def gather_process_measures(
         self, ctx: StatContext, pids: Sequence[int]
     ) -> Sequence[ProcessMeasurement]:
-        async def proc_impl(pid: int):
-            try:
-                proc_path = Path(f"/proc/{pid}/stat")
-                cpu_user, cpu_kernel = map(int, proc_path.read_text().split(" ")[13:15])
-            except FileNotFoundError:
-                log.warning("cannot read stats: file not found for Process {0}", pid)
-            else:
-                clk_per_sec = os.sysconf("SC_CLK_TCK")
-                cpu_used = Decimal(cpu_user + cpu_kernel) / clk_per_sec * 1000
-                return cpu_used
-            return None
-
         async def psutil_impl(pid: int):
             try:
                 p = psutil.Process(pid)
@@ -279,16 +267,12 @@ class CPUPlugin(AbstractComputePlugin):
                 return cpu_used
             return None
 
-        if ctx.mode == StatModes.CGROUP:
-            impl = proc_impl
-        else:
-            impl = psutil_impl
         per_process_cpu_util = {}
         per_process_cpu_used = {}
         tasks = []
         q = Decimal("0.000")
         for pid in pids:
-            tasks.append(asyncio.ensure_future(impl(pid)))
+            tasks.append(asyncio.ensure_future(psutil_impl(pid)))
         results = await asyncio.gather(*tasks)
         for pid, cpu_used in zip(pids, results):
             if cpu_used is None:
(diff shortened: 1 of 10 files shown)


src/ai/backend/agent/docker/intrinsic.py
```diff
@@ -240,6 +241,51 @@ async def api_impl(container_id):
...
+        async def psutil_impl(pid: int):
+            try:
+                p = psutil.Process(pid)
+            except psutil.NoSuchProcess:
+                log.warning("psutil cannot found process {0}", pid)
+            else:
+                cpu_times = p.cpu_times()
+                cpu_used = Decimal(cpu_times.user + cpu_times.system) * 1000
+                return cpu_used
+            return None
```
kyujin-cho: Using this kind of approach on macOS won't work. Containers will run on isolated VM aside from agent host.

src/ai/backend/agent/docker/intrinsic.py
```diff
@@ -568,6 +614,59 @@ async def api_impl(container_id):
...
+                p = psutil.Process(pid)
+            except psutil.NoSuchProcess:
+                log.warning("psutil cannot found process {0}", pid)
+            else:
+                stats = p.as_dict(attrs=["memory_info", "io_counters"])
+                mem_cur_bytes = stats["memory_info"].rss
+                io_read_bytes = stats["io_counters"].read_bytes
+                io_write_bytes = stats["io_counters"].write_bytes
+                return mem_cur_bytes, io_read_bytes, io_write_bytes
+            return None
```
kyujin-cho: Same here.

src/ai/backend/agent/stats.py
```diff
@@ -454,3 +475,105 @@ async def _pipe_builder(r: Redis) -> Pipeline:
...
+                    )
+                else:
+                    for unused_pid in unused_pids:
+                        log.debug("removing pid_metric for {}: {}", cid, unused_pid)
+                        self.process_metrics[cid].pop(unused_pid, None)
+                finally:
+                    await docker.close()
+                for pid in pids:
+                    pid_map[pid] = cid
+                total_pids.extend(pids)
```
kyujin-cho: What happens if original process represented by the PID is killed and the PID is reused by another process?

src/ai/backend/agent/stats.py
```diff
@@ -454,3 +475,105 @@ async def _pipe_builder(r: Redis) -> Pipeline:
...
+                    result = await docker._query_json(f"containers/{cid}/top", method="GET")
+                    procs = result["Processes"]
+                    pids = [PID(int(proc[1])) for proc in procs if proc[0] != "root"]
+                    if os.getuid() == 0:
+                        pids.extend([PID(int(proc[1])) for proc in procs if proc[0] == "root"])
+                    unused_pids = set(self.process_metrics[cid].keys()) - set(pids)
+                except (KeyError, aiodocker.exceptions.DockerError):
+                    log.warning(
+                        "collect_per_container_process_stat(): cannot found container {}", cid
+                    )
```
kyujin-cho: Not finding container with preserved ID is nothing to warn about, since the container always has chance to gracefully exit when user terminates kernel. Let's change log level to "debug" instead of "warn" not to spam user with the logging.

kyujin-cho (changes requested):

PR #918, opened 2022-12-07
ci: add Actions for Windows distribution client SDK

### LINKED ISSUE #740: Automate Windows distribution of Client SDK
When we distribute our client SDK and the official CLI to the Windows platforms, we had internally used [`conda pack`](https://conda.github.io/conda-pack/).

Let's update our deployment workflow (in `.github/workflows/default.yml`) to automatically build it from this repository's releases using Windows x64 runner hosts.

NOTE: Currently `pants` and `pex` only run with WSL on Windows, which is mostly yet another Linux.  The goal of this issue is to make native **client SDK** distributions for Windows (server components such as manager and agent are not included!).

We may need to run the Windows client distribution build *after* making release wheels instead of in parallel with the `pants package` step.

Yaminyam:
close #740


.github/workflows/default.yml
```diff
@@ -279,3 +279,71 @@ jobs:
         name: pants.deploy.log
         path: .pants.d/pants.log
       if: always()  # We want the log even on failures.
+      
+  windows-conda:
+    needs: [lint, typecheck, test]
```
achimnol: I think `    if: github.event_name == 'push' && contains(github.ref, 'refs/tags/')` is missing here!

achimnol (changes requested):

achimnol (approved):

achimnol:
Currently I cannot merge `python.lock` by regenerating the lockfile due to the following error:
image

PR #919, opened 2022-12-07
feature: add DELETABLE status for VFolderAccessStatus

fregataa:
About `VFolderAccessStatus`, `UPDATABLE` status has been used in any of update methods including delete vfolder methods.
delete vfolder methods should be distinguished from `UPDATABLE` status. e.g, mounted vfolders can be updated but should not be deleted.


achimnol (commented):
Let's clarify the error message like "filtering has failed" to clearer ones like "This folder is being used by one or more container.".

src/ai/backend/manager/api/vfolder.py
```diff
@@ -113,35 +113,59 @@ async def ensure_vfolder_status(
...
             allow_privileged_access=True,
         )
         if len(entries) == 0:
-            raise VFolderFilterStatusFailed()
+            raise VFolderFilterStatusFailed(
+                f"Cannot find any folder with the given identity ({folder_id = }, {folder_name = })"
+            )
+        for entry in entries:
+            if entry["status"] in available_vf_statuses:
+                break
```
kyujin-cho: Hmm, I'm a little bit unsure about this logic: Is it okay to just say okay if any of the folder has desired status? Not to all the vFolders?

src/ai/backend/manager/models/vfolder.py
```diff
@@ -130,6 +130,7 @@ class VFolderAccessStatus(str, enum.Enum):
 
     READABLE = "readable"
     UPDATABLE = "updatable"
+    DELETABLE = "deletable"
```
kyujin-cho, line 136: Please check whether we need additional database migration script or not.

kyujin-cho (changes requested):

src/ai/backend/manager/models/vfolder.py
fregataa: we do not need db migration because `VFolderAccessStatus` is not used by any db column

src/ai/backend/manager/api/vfolder.py
fregataa: I changed it to check all vfolders. thanks!

src/ai/backend/manager/api/vfolder.py
```diff
@@ -113,35 +113,49 @@ async def ensure_vfolder_status(
...
+            # if UPDATABLE access status is requested, READY and MOUNTED operation statuses are accepted.
+            available_vf_statuses = {
+                VFolderOperationStatus.READY,
+                VFolderOperationStatus.MOUNTED,
+            }
+        case VFolderAccessStatus.DELETABLE:
+            # if DELETABLE access status is requested, only READY operation status is accepted.
+            available_vf_statuses = {
+                VFolderOperationStatus.READY,
+                VFolderOperationStatus.DELETING,
```
kyujin-cho: Are vFolders on `DELETING` status really suits for `DELETE` operation?

kyujin-cho (changes requested):

src/ai/backend/manager/api/vfolder.py
fregataa: The intention is to enable multiple tries or something like `forced` option.
but it seems not reasonable for now. thanks!

PR #928, opened 2022-12-12
fix: Adjust aiomonitor default port ranges and its startup error handling

achimnol:
* aiomonitor is not a strictly required service for our operation
* fix(agent): Move aiomonitor initialization upfront of other init codes
* fix: Move the default aiomonitor port range out from the ephemeral range
  - 49152 to 65535 is reserved for dynamic use by the operating system,
    and there are chances to a random TCP client process in the host may
    take over a port number first.


kyujin-cho (approved):

kyujin-cho:
Approved. Please merge main branch before merging to main. Also in current state the milestone is set to 22.09. Are you planning to backport changes to 22.09 branch?

achimnol:
Yep, this will be backported to 22.09.

PR #935, opened 2022-12-16
ci: Use Pants-official init action to init and cache Pants

Yaminyam:


achimnol:
Let's use these:
* https://github.com/pantsbuild/actions/tree/main/init-pants/
* https://github.com/pantsbuild/example-python/blob/main/.github/workflows/pants.yaml#L27-L47

PR #940, opened 2022-12-20
ci: uniform cache name

### LINKED ISSUE #926: Is there any reason to use a different cache for `deploy-to-pypi`?

Linux (x86-64)

### Backend.AI version

22.09.6 (main)

### Describe the bug

It seems that a cache for pants is used in many places to reduce time. However, I found that one action named `deploy-to-pypi` is using a cache with a different key.

Most of them are using keys with the prefix `pants-`.
https://github.com/lablup/backend.ai/blob/cc272fb77843f77c023b46d6aed68d1c828a9ba5/.github/workflows/default.yml#L16-L24
https://github.com/lablup/backend.ai/blob/cc272fb77843f77c023b46d6aed68d1c828a9ba5/.github/workflows/default.yml#L74-L82

However, `deploy-to-pypi` is using a key without the prefix `pants-`.
https://github.com/lablup/backend.ai/blob/cc272fb77843f77c023b46d6aed68d1c828a9ba5/.github/workflows/default.yml#L193-L202

### To Reproduce

_No response_

### Expected Behavior

Unless there is a reason to use a different cache, I think they have to be unified.

### Anything else?

_No response_

Yaminyam:
close #926

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -199,7 +199,7 @@ jobs:
           ~/.cache/pants/setup
           ~/.cache/pants/lmdb_store
           ~/.cache/pants/named_caches
-        key: ${{ runner.os }}-${{ hashFiles('pants*.toml', '**/*.lock') }}-
+        key: pants-${{ runner.os }}-${{ hashFiles('pants*.toml', '**/*.lock') }}-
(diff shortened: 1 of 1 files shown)


achimnol (approved):
Thanks for catching this!

PR #944, opened 2022-12-23
fix: Increase session concurrency after successful session creation

fregataa:
increase session concurrency after successful session creation rather in predicate check.

### DIFF
--- a/changes/944.fix.md
+++ b/changes/944.fix.md
@@ -1,0 +1,1 @@
+Increase session concurrency after successful session creation.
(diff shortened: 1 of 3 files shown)


achimnol (commented):
I remember that this was by design, as it may bypass the session concurrency limit by posting multiple session creation requests at once when there are enough remaining resources.
Could you check this scenario and if the limit still applies well?

fregataa:
I will reopen this pr after I take a session creation test.

achimnol:
Resolved by #1004!

PR #946, opened 2022-12-25
fix: Apply vfolder_host_permission to vfolder api

fregataa:
apply the feature to all vfolder APIs.
Since some of vfolder APIs have not checked `vfolder_host_permission`, this PR inserts permission checking process into all vfolder APIs.


lizable:
@fregataa Could you explain what's the meaning of "partially applied"?

fregataa:
> @fregataa Could you explain what's the meaning of "partially applied"?

I made a change of the description. sorry for my bad explanation.

PR #948, opened 2022-12-27
feat: support systemd docker cgroup driver

### LINKED ISSUE #363: Generalize docker cgroup driver support
Currently the agent adds our own cgroup control routines to retrieve container statistics of containers after their termination on top of Docker's cgroupfs lifecycle.

This works well with the default cgroup driver, but there are reports that it conflicts with the "systemd" cgroup driver.  The relevant configuration option name is `native.cgroupdriver`.

There seems no big technical difference between systemd-managed cgroups and cgroupfs-based cgroups, but they have different directory namings under the sysfs.  It is also notable that Kubernetes recommends to use the "systemd" cgroup driver and keep the configurations for kubelets and Docker same.

* https://stackoverflow.com/questions/43794169/docker-change-cgroup-driver-to-systemd/65870152
* https://kubernetes.io/docs/setup/production-environment/container-runtimes/
* https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver/

Let's support both types of cgroup drivers to prevent unexpected conflicts on various systems, by adding auto-detection of the current cgroup driver and using different directory names that match with each driver when collecting stats.

qkoo0833:
1. Automatically detect docker's current cgroup driver

2. Match different path prefix with each driver

resolve #363


achimnol (changes requested):
Let's use [the new pattern matching syntax](https://peps.python.org/pep-0636/)!
This will allow type checkers to validate whether all enum branches are explicitly taken or not.

src/ai/backend/agent/stats.py
```diff
@@ -266,7 +281,7 @@ def __init__(
         self, agent: "AbstractAgent", mode: StatModes = None, *, cache_lifespan: int = 120
     ) -> None:
         self.agent = agent
-        self.mode = mode if mode is not None else StatModes.get_preferred_mode()
+        self.mode = mode if mode == StatModes.DOCKER else StatModes.get_preferred_mode()
```
achimnol: The argument is optional and `None` means that the caller is leaving the choice of mode to the implementation. So, this should not be changed.

src/ai/backend/agent/stats.py
```diff
@@ -66,7 +68,9 @@ def check_cgroup_available():
 
 class StatModes(enum.Enum):
     CGROUP = "cgroup"
+    CGROUPFS = "cgroupfs"
     DOCKER = "docker"
+    SYSTEMD = "systemd"
 
     @staticmethod
     def get_preferred_mode():
```
achimnol: This function should be updated to auto-detect the cgroupfs/systemd modes as well.

achimnol (changes requested):

src/ai/backend/agent/stats.py
```diff
@@ -65,18 +68,39 @@ def check_cgroup_available():
 
 
 class StatModes(enum.Enum):
+    """
+    CGROUP = Tries to automatically detect docker's current cgroup driver
+    CGROUPFS = Manually set cgroup driver to cgroupfs
+    SYSTEMD = Manually set cgroup driver to systemd
```
achimnol, line 75: Please give a small favor to `DOCKER` as well. :wink:

src/ai/backend/agent/stats.py
```diff
@@ -65,18 +68,39 @@ def check_cgroup_available():
...
-            return StatModes.CGROUP
+            docker_cgroup_driver = await StatModes.check_docker_cgroup_driver()
+            if docker_cgroup_driver is not None:
+                return docker_cgroup_driver
+            else:  # Fallback to default (cgroupfs)
+                return StatModes.CGROUPFS
         return StatModes.DOCKER
 
+    @classmethod
+    async def check_docker_cgroup_driver(cls):
```
achimnol: Could you specify the return type explicitly?
If required, you may add `from __future__ import annotations`.

achimnol (changes requested):

src/ai/backend/agent/stats.py
```diff
@@ -93,7 +95,7 @@ async def get_preferred_mode():
         return StatModes.DOCKER
 
     @classmethod
-    async def check_docker_cgroup_driver(cls):
+    async def check_docker_cgroup_driver(cls) -> StatModes:
```
achimnol: I think this should be `Optional` as the method may return `None`.

kmkwon94: Hmm? I alreday applied Optional type in this commit 44bb4cf

achimnol: Ah, my mistake. I just checked the first commit only; Thanks!

src/ai/backend/agent/stats.py
```diff
@@ -65,18 +69,40 @@ def check_cgroup_available():
...
+                return docker_cgroup_driver
+            else:  # Fallback to default (cgroupfs)
+                return StatModes.CGROUPFS
         return StatModes.DOCKER
 
+    @classmethod
+    async def check_docker_cgroup_driver(cls) -> Optional[StatModes]:
+        async with Docker() as docker:
+            try:
+                result = await docker._query_json("info", method="GET")
```
sanxiyn, line 101: This should be `docker.system.info()`.

achimnol:
This is replaced by #1000 with better handling of 4 combinations of (systemd, cgroup) x (cgroup v1, cgroup v2).

PR #953, opened 2023-01-03
feature: Add project admin role and define role based actions

fregataa:
blocked by #963


fregataa:
This pr is replaced by #1329

PR #954, opened 2023-01-04
fix: set `resp` after setting `user_uuid` to avoid AttributeError

Sujin-Kim1:
In order to avoid the following attribute error during converting from string to hex, the `resp` setting position was changed.
image

### DIFF
--- a/src/ai/backend/manager/api/session_template.py
+++ b/src/ai/backend/manager/api/session_template.py
@@ -63,10 +63,6 @@ async def create(request: web.Request, params: Any) -> web.Response:
         for st in body["session_templates"]:
             template_data = check_task_template(st["template"])
             template_id = uuid.uuid4().hex
-            resp = {
-                "id": template_id,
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/api/session_template.py
```diff
@@ -85,6 +81,10 @@ async def create(request: web.Request, params: Any) -> web.Response:
                 }
             )
             result = await conn.execute(query)
+            resp = {
+                "id": template_id,
+                "user": user_uuid,
```
kyujin-cho: When `UUID` object is used as `user_uuid`, Python's JSON library will fail to render this dict as JSON string. Let's add a fallback method.
```suggestion
                "user": user_uuid if isinstance(user_uuid, str) else user_uuid.hex,
```

kyujin-cho (changes requested):

kyujin-cho (approved):
LGTM.

PR #962, opened 2023-01-09
feat: session based usage stats for given period

### LINKED ISSUE #961: Add an API that calculates session-based usage stats used in a given time period
### Main idea

Let's add an API to measure usage logs on a session-based basis for a given period to provide usage stats for node:container pair.
The original API `get_container_stats_for_period` returns container-based usage stats. Additional efforts are required to bundle containers with the same session ID to give usage information. So instead of refactoring that API, let's add a new API to give session-based basis usage stats.

The structure of the new API is roughly as follows.
```
[
  {
    'domain_name': '...'
    'g_infos': [
       {
          ...,
          's_infos': [
             {
                ...,
               'c_infos': [{...}, ...]
            },
          ...
          ]
       },
       ...
    ]
  }, 
  ...
]
```

### Alternative ideas

_No response_

### Anything else?

This API needs to be modified after merging #576

agatha197:
resolves #961 
related PR: https://github.com/lablup/backend.ai-control-panel/pull/473

- `g_name` -> `g_group_name`  for clarity.
- `get_stats_for_period`
  - return usage statistics of sessions used within a given time period, group_ids, aggregated by group id.
  - [{domain_name: ..., g_infos: [s_infos: <result of `get_session_stats_for_period`>, ...]}, {...}]

- `get_session_stats_for_period`
  - return usage statistics of sessions used within a given time period, group_ids. 
  - [{id: '...', session_id: '...', c_infos: <result of `get_container_stats_by_session_id`>}, {...}]
  
- `get_container_stats_by_session_id`
  - return usage statistics for containers within a specific session.
  - [{id: '...', container_id: '...', ...}, {...}]


src/ai/backend/manager/api/resource.py
```diff
@@ -295,6 +295,322 @@ async def recalculate_usage(request: web.Request) -> web.Response:
     return web.json_response({}, status=200)
 
 
+async def get_stats_for_period(request: web.Request, start_date, end_date, group_ids=None):
```
fregataa: It would be good to relocate this function to session table of model layer.
At least, we can locate them in kernel table.

fregataa (commented):
This work seems related to #963, which change name "group" to "project" in manager and
strongly related to #576, which divide kernel table and session table.
Plus, how about implementing a dataclass or attrs class for stats?
I would do this job if you do not mind.

src/ai/backend/manager/models/kernel.py
```diff
@@ -22,6 +22,7 @@
 import sqlalchemy as sa
 from dateutil.parser import parse as dtparse
 from dateutil.tz import tzutc
+from dateutil.tz.tz import tzfile
```
achimnol: This could be merged to the above line.  Maybe your editor got confused about nested imports. :wink:
`from dateutil.tz import tzfile, tzutc`

achimnol (changes requested):
If you replace existing codes, please remove them.
Or, apply proper refactor/fix to existing codes.

kyujin-cho (approved):

PR #963, opened 2023-01-09
refactor: Rename group to project

fregataa:
For a consistent glossary, rename `groups` table to `projects`.
All related names of variables, files and modules will change.
# MUST
1. run `alembic stamp <HEAD>` command after upgrade head.
2. update plugins that uses `groups` table.


fregataa:
Close this PR since it is outdated

Issue #299, opened 2021-10-09
Add command line logging option 


mindgitrwx:
**Is your feature request related to a problem? Please describe.**
Current backend.ai modules can be executed with the 'debug' option. 
For example, `python -m ai.backend.manager.server --debug`
But other logging options like 'notset', 'info', 'warning', 'error', 'critical' options are must be setted on .toml files.

I want to use all logging options directly on the command line.

**Describe the solution you'd like**

These pictures are brief implementations after modified code.
 
`python -m ai.backend.manager.server --debug`

![Selection_259](https://user-images.githubusercontent.com/37054916/136644146-b9f38e1f-c9e7-4592-ba57-33fcd013123f.png)

`python -m ai.backend.manager.server --info`

![Selection_260](https://user-images.githubusercontent.com/37054916/136644144-593a0831-905e-4aed-8eb3-ecf446e04ed1.png)

`python -m ai.backend.manager.server --warning`

![Selection_261](https://user-images.githubusercontent.com/37054916/136644134-a028ffce-db9a-4b82-bdb9-e254e728cf8e.png)

achimnol:
Maybe it would be better to have a general scheme to map the CLI options with TOML options for CLI-side override.

PR #967, opened 2023-01-10
fix: add command line logging option

leejiwon1125:
"mindgitrwx" wanted to use various logging options directly on command line. 

But we can only use 'debug' option. (In fact, we can use 'info' option by simply removing  "--debug" in `./backend.ai mgr start-server --debug` command)

So, I accepted the demand for "mindgitrwx". 

From now on, we can use options such as debug, info, warning, error and critical.

To run manager server, type in terminal
 `./backend.ai mgr start-server --log-level=~`   (~ is one of the logging options: debug, info, warning, error, critical)
instead of `./backend.ai mgr start-server --debug`

Plus, according to changed code, the default logging option is info so if you just type `./backend.ai mgr start-server` in terminal, then logging option is info.

### DIFF
--- a/src/ai/backend/manager/cli/__main__.py
+++ b/src/ai/backend/manager/cli/__main__.py
@@ -23,7 +23,6 @@ from .context import CLIContext, init_logger, redis_ctx
 
 log = BraceStyleAdapter(logging.getLogger("ai.backend.manager.cli"))
 
-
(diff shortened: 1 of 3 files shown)


rapsealk:
FYI: This PR resolves #299

leejiwon1125:
resolves #299

leejiwon1125:
continued by #971

Issue #969, opened 2023-01-11
CLI manager command with no args doesn't show up any description


ChangHoon-Sung:
### What Operating System(s) are you seeing this problem on?

Linux (x86-64)

### Backend.AI version

23.03.0dev0

### Describe the bug

Command 'mgr' doesn't show up any description or help when executing the command without any argument.

<img width="693" alt="ScreenShot 2023-01-11 at 10 46 27@2x" src="https://user-images.githubusercontent.com/37946887/211698898-cf1332bd-db7c-4ac4-acf4-24f816af76ab.png">

`-h` or `--help` argument is still working.

### To Reproduce

run `./backend.ai mgr`

### Expected Behavior

It should show up the help information for convenience like other commands. The figure below is 'ag(agent)' case.

<img width="638" alt="ScreenShot 2023-01-11 at 10 52 46@2x" src="https://user-images.githubusercontent.com/37946887/211699635-2ff853f2-1001-4605-a4de-b754a52c1cc2.png">

PR #970, opened 2023-01-11
fix: Show up the manager cli help when called without sub-command

ChangHoon-Sung:
A manager command seems not used alone (always used with arguments).
This PR prevents the manager command invocation without the arguments and makes the command print the default help message.

resolves: #969


kyujin-cho (approved):
LGTM!

PR #981, opened 2023-01-17
feature: Advanced type hints for nested dictionary

### LINKED ISSUE #980: Support advanced type hints for nested dictionary
### Main idea

From the code below, the type hint for `get_prefix()` fails to represent cases of returning a nested dictionary.
https://github.com/lablup/backend.ai/blob/772b6a821a9747c35f5fdc6bc17b935f26ecc949/src/ai/backend/common/etcd.py#L300-L312

It needs to be updated to support dictionary (or Mapping generally), like `Mapping[str, Union[Optional[str], Mapping]]` for instance. However, it is not enough yet. We need to define a recursive type (`RecursiveDict = Mapping[str, RecursiveDict | str]`) to represent the whole cases perfectly.

### Alternative ideas

_No response_

### Anything else?

_No response_

rapsealk:
This PR resolved #980


src/ai/backend/common/etcd.py
```diff
@@ -98,6 +98,8 @@ def _slash(v: str):
 P = ParamSpec("P")
 R = TypeVar("R")
 
+RecursiveMapping = Mapping[str, Union["RecursiveMapping", Optional[str]]]
```
kyujin-cho: Let's rename this type name to somewhat related to `get_prefix()`. Vague names can create confusion to those who read the code.
```suggestion
GetPrefixValue = Mapping[str, Union["GetPrefixValue", Optional[str]]]
```

kyujin-cho (changes requested):

src/ai/backend/common/etcd.py
rapsealk: Yes, it was too general.

PR #983, opened 2023-01-18
fix: truncated paginated list piped output

### LINKED ISSUE #677: Piping the output of agent list command only returns up to 20 entries
**Describe the bug**
I tried to get every Agent ID with the CLI command. There were over 30 agents in the server farm and `backend.ai admin agent list` command just worked well by returning every Agent list with the pagination. However, when I pipe the result with commands like `awk` or `wc -l`, I only got 20 entries.

**To Reproduce**
1. First, since a test cluster usually does not have >20 Agents, I created 50 fake `TERMINTED` agents.
    ```shell
    for i in {1..50}; do
        slots='{"cpu": "4", "mem": "34359738368", "cuda.shares": "0"}'
        fields='"id", status, region, scaling_group, available_slots, occupied_slots, addr, version, compute_plugins, architecture'
        values="'i-test-${i}', 'TERMINATED', 'local', 'default', '${slots}'::jsonb, '${slots}'::jsonb, 'tcp://1.2.3.4:6001', '22.09.0b4', '{}'::jsonb, 'x86_64'"
        insert_query="INSERT INTO agents ($fields) VALUES ($values);"
        echo $insert_query
        # Use your own DB's container name or ID.
        docker exec -it main_backendai-half-db_1 psql -U postgres -d backend -c "$insert_query"
    done
    ```
2. Confirm that `backend.ai admin agent list -s TERMINATED` command returns >=50 Agents (including the real ones). 
    image
3. Now, pipe the result with any command. For example, I used `awk` to get the Agent's IDs only: `backend.ai admin agent list -s TERMINATED | awk '{print $1}'`. Then, I got just 20 Agent IDs:
    image
4. `wc -l` command confirms again (2 for header rows):
    image

**Expected behavior**
Piping the output of the Agent list command should return every item.

**Client:**
 - OS: CentOS 7.9
 - Python version: 3.10.5

**Server:**
 - OS: CentOS 7.9
   - Output of `uname -snrmpio`: `Linux asto1 3.10.0-1160.21.1.el7.x86_64 x86_64 x86_64 x86_64 GNU/Linux`
 - Backend.AI version: 22.09.0b4
 - Python version: 3.10.5
 - Installation method: on-premise

**Additional context**
Other commands like `session list` also shows the same issue:
image

If you want to delete the test Agents, just execute the script:
```shell
for i in {1..50}; do
    delete_query="DELETE FROM agents WHERE \"id\" = 'i-test-${i}';"
    echo $delete_query
    # Use your own DB's container name or ID.
    docker exec -it main_backendai-half-db_1 psql -U postgres -d backend -c "$delete_query"
done
```

ChangHoon-Sung:
resolves #677

### DIFF
--- a/src/ai/backend/client/output/console.py
+++ b/src/ai/backend/client/output/console.py
@@ -18,7 +18,7 @@ class NoItems(Exception):
 class ConsoleOutputHandler(BaseOutputHandler):
     def print_item(
         self,
-        item: Mapping[str, Any] | None,
+        item: Optional[Mapping[str, Any]],
         fields: Sequence[FieldSpec],
     ) -> None:
         if item is None:
@@ -115,7 +115,7 @@ class ConsoleOutputHandler(BaseOutputHandler):
         self,
         fetch_func: Callable[[int, int], PaginatedResult],
         initial_page_offset: int,
-        page_size: int = None,
+        page_size: Optional[int] = None,
     ) -> None:
         fields: Sequence[FieldSpec] = []
 
--- a/src/ai/backend/client/output/console.py
+++ b/src/ai/backend/client/output/console.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import sys
-from typing import Any, Callable, List, Mapping, Optional, Sequence, Iterator, MutableMapping
+from typing import Any, Callable, Iterator, List, Mapping, MutableMapping, Optional, Sequence
 
 from tabulate import tabulate
 
@@ -9,7 +9,6 @@ from ai.backend.client.cli.pagination import echo_via_pager, get_preferred_page_
 from ai.backend.client.cli.pretty import print_error, print_fail
 
 from .types import BaseOutputHandler, FieldSpec, PaginatedResult
-
 
 _Item = MutableMapping[str, Any]
 
@@ -137,18 +136,19 @@ class ConsoleOutputHandler(BaseOutputHandler):
                     break
 
         if sys.stdout.isatty() and page_size is None:
-            page_size = get_preferred_page_size()
+            preferred_page_size = get_preferred_page_size()
             try:
                 echo_via_pager(
                     tabulate_items(
-                        infinite_fetch(page_size),
+                        infinite_fetch(preferred_page_size),
                         fields,
                     ),
                 )
             except NoItems:
                 print("No matching items.")
         else:
-            page_size = page_size or 20
+            if page_size is None:
+                page_size = 20
             for line in tabulate_items(
                 infinite_fetch(page_size),
                 fields,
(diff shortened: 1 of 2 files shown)


fregataa (approved):
LGTM.
one thing I'd like to comment is about the name of the variable `page_size` and type hint of it.
How about adding `Optional` type on `page_size` parameter and rename the local variables `page_size` in it? Because naming the same with a local variable and a parameter is not good.

ChangHoon-Sung:
I rebased this branch for matching the python version with the main branch.

src/ai/backend/client/output/console.py
```diff
@@ -117,39 +117,38 @@ def print_paginated_list(
         initial_page_offset: int,
         page_size: int = None,
     ) -> None:
+        fields: Sequence[FieldSpec] = []
+
+        def infinite_fetch(page_size):
```
kyujin-cho: The `page_size` parameter is shadowing definition of already defined `page_size` variable. Let's rename it.
```suggestion
        def infinite_fetch(_page_size):
```

kyujin-cho (changes requested):

src/ai/backend/client/output/console.py
```diff
@@ -115,41 +117,41 @@ def print_paginated_list(
...
-                        raise NoItems
-                    current_offset += len(result.items)
-                    if not fields:
-                        fields.extend(result.fields)
-                    yield from result.items
-                    if current_offset >= result.total_count:
-                        break
+        fields: List[FieldSpec] = []
+
+        def infinite_fetch(_page_size) -> Iterator[_Item]:
```
kyujin-cho: Please add type annotation to `_page_size` parameter also.
```suggestion
        def infinite_fetch(_page_size: int) -> Iterator[_Item]:
```

src/ai/backend/client/output/console.py
```diff
@@ -18,7 +20,7 @@ class NoItems(Exception):
 class ConsoleOutputHandler(BaseOutputHandler):
     def print_item(
         self,
-        item: Mapping[str, Any] | None,
+        item: Optional[Mapping[str, Any]],
```
kyujin-cho: We can also utilize `_Item` type alias here.
```suggestion
        item: Optional[_Item],
```

kyujin-cho (changes requested):

src/ai/backend/client/output/console.py
changhoon-sung: Oh, I missed it. Thank you.

src/ai/backend/client/output/console.py
changhoon-sung: I found there are `Mapping` and `MutableMapping`.
In our context, I think `Mapping` is more suitable for the item. An abstract class `BaseOutputHandler` which declared the item as `Mapping`. A consuming function `tabulate_items` declared the item as `MutableMapping` but there's no mutation and it just iterates the items.

changhoon-sung: Please review the following commit.

kyujin-cho: Fair enough.

kyujin-cho (approved):

PR #987, opened 2023-01-20
feat: Add fish shell support to `install-dev.sh` script

jopemachine:
For example, below command is not working with fish shell because [`unset` command is not supported in fish](https://stackoverflow.com/questions/30703860/how-do-i-unset-a-variable-in-the-fish-shell).

```
$ source ./env-local-admin-session.sh
```

This PR fixes this by using `set -e` command instead of `unset` when user works in fish.

### DIFF
--- a/changes/987.feature.md
+++ b/changes/987.feature.md
@@ -1,0 +1,1 @@
+Support fish shell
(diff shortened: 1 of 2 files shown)


changes/987.feature.md
```diff
@@ -0,0 +1 @@
+Support fish shell
```
kyujin-cho: Please write news fragment more concisely so that viewers can understand that this patch is to support `install-dev.sh` script working on fish shell.

kyujin-cho (changes requested):

PR #989, opened 2023-01-20
ci: automate making news fragment process

leejiwon1125:
* This PR is continued by PR #977 which I closed by my mistake.

 I was intended to implement 'fully automate news fragment making workflow.'
But due to some functional limits and considering user's convenience, I only made half(and passive) automate news fragment making workflow. If you think if there is more graceful implement, please let me know and I will reflect it happily.

 If you use this workflow, you **don't need to check PR number manually** anymore. For using this workflow, what you have to do is only making 'feature' named file in changes folder(of course you have to write your news fragment contents in this file). If your PR's towncrier type is doc, then make **doc** named file in changes folder. There is no filename extension. Just put your towncrier type in file name. If your PR's number is 956, then this workflow change your file name: doc to **956.doc.md**. Note that towncrier type is one of this value: "breaking", "feature", "enhance", "deprecation", "fix", "doc", "deps", "misc". So you should choose one that most fit with your PR. 

 But, you don't have to necessarily use this workflow. This workflow doesn't bother you for the reason that you don't use this workflow in the process of making news fragment. This workflow only helps you passively. In upper case I suggested, if you make **doc** file in changes folder, then this workflow glady change its name to **956.doc.md**. If you don't make **doc** file, it's okay because you can make it yourself. To sum up, this workflow is not mandatory but optional choice for your convenience that only takes care of your news fragment 'file name format' with PR number.

 Plus, if this workflow made changes of your file name, it commit on your PR like underneath picture. According to this function, there is no change and no commit message when there already exists 'correct file name with PR number'(ex. **956.doc.md**) or 'no reference file(named with towncrier type file: ex. **doc**) to make correct file name'.

image

more advice.
-> If this workflow changes file, you have to pull this remote branch to your local branch to push again.


rapsealk:
Nice work!

leejiwon1125:
> Nice work!

Thanks! Also, thank you for your kind advice. It was very helpful :)

.github/workflows/auto-recode.yml
```diff
@@ -0,0 +1,60 @@
+name: auto-recoder
```
rapsealk: Does it mean re-code? I think this name does not represent what this workflow does.

leejiwon1125: I changed name auto-recoder to pr-number-assign!

.github/workflows/auto-recode.yml
```diff
@@ -0,0 +1,60 @@
...
+        files = os.listdir(path)
+        print(str(files) + " are in changes floder.")
+
+        for file in files :
+            if str(file[0:file.find(".")]) == pr_number :
+                print("Change log file already exist for this PR.")
+                exit(0)
+
+        for t_type in t_types :
+            if t_type in files :
```
Yaminyam: It doesn't seem to handle the case of writing anything but a pr number, like `doc.md`.

.github/workflows/auto-recode.yml
```diff
@@ -0,0 +1,60 @@
...
+      with:
+        ref: ${{ github.head_ref }}
+    
+    - name: do work
+      run: |
+        script=$(cat <<EOF
+        import os
+
+        pr_number = "${{github.event.pull_request.number}}"
+        path = "/home/runner/work/backend.ai/backend.ai/changes"
```
Yaminyam: I know that the current path is the best path to repository, but if you specify only the changes folder as the relative path instead of the absolute path, I think it will be a script that can be used universally in other repositories.

.github/workflows/auto-recode.yml
leejiwon1125: I wanted to make my workflow as 'optional' thing for developers for the reason that they can check pr number manually as they did. I don't want to bothering them to use this workflow. Under this opinion, I made this workflow only change file name 'feat' to 'pr_number.feat.md'.

As you can see, there is not error case in my workflow because there is timeline-check workflow that check if there is pr number in file name. So 'doc.md' case will be treated as last case and it will print "There is not change log file for this PR in changes folder." with no error in my workflow and make error in timeline check workflow. This means my workflow don't need to consider such a case like 'doc.md'.

.github/workflows/auto-recode.yml
leejiwon1125: I think using relative path is more nice implement than current code. I'll reflect your comment. Thanks!!

.github/workflows/auto-recode.yml
```diff
@@ -0,0 +1,60 @@
...
+                exit(0)
+
+        if target_file == "":
+            print("There is not change log file for this PR in changes folder.")
+            print("Also, there is not feature named file in changes folder.")
+            exit(0)
+        
+        EOF
+        )
+        echo "$script" | python
```
achimnol: I think you could instead set the `shell` directive `"/usr/bin/env python"` to make it as a Python script.

Also, please keep the PEP-8/Black compliant codestyle even for this embedded Python script. If it becomes too long, I'd recommend to put it as a standalone script file under `scripts/` and call it from the workflow YAML, for easier testing and better editor supports.

.github/workflows/auto-recode.yml
```diff
@@ -0,0 +1,60 @@
...
+        ref: ${{ github.head_ref }}
+    
+    - name: do work
+      run: |
+        script=$(cat <<EOF
+        import os
+
+        pr_number = "${{github.event.pull_request.number}}"
+        path = "/home/runner/work/backend.ai/backend.ai/changes"
+        t_types = ["breaking","feature","enhance","deprecation","fix","doc","deps","misc"]
```
achimnol: Could you let it source this list from `pyproject.toml`?

.github/workflows/auto-recode.yml
achimnol: It should support both ".{category}" and ".{category}.md" filename patterns.

achimnol (changes requested):

.github/workflows/auto-recode.yml
achimnol: I think it may be better to use regex, such as:
```python
if match := re.search(r"\." + re.escape(t_type) + r"(\.md)?$", filename):
    ...
```

achimnol: Also I'd recommend to use [`pathlib`](https://docs.python.org/3/library/pathlib.html) instead of `os` and `os.path` APIs, to get used with modern Python path manipulation APIs.

.github/workflows/auto-recode.yml
Yaminyam: The file name remains the same, but I think it would be good to change the file name.

leejiwon1125:
- I reflected all the comments above except Joongi's one recommendation that suggests **moving python script in 'pr-number-assign.yml' under 'scripts/' directory** for the reason that can check in PR #977.

- Now, It works not only "{category}" named file, but also "{category}.", "{category}.md", ".{category}", ".{category}.", and ".{category}.md".

- If the **timeline-check** workflow occurs earlier than **pr-number-assign** workflow, there is error in **timeline-check**  because change log file name changes correctly after the **timeline-check**. So, I made dependency: **timeline-check** works only after **pr-number-assign** workes.

- Kyujin worried that github-acitons will be co-commiter for all the PR if we use this workflow. So, I'm now trying to fix this issue.

If you have any question or suggestion for this PR, please let me know.

rapsealk:
> * I reflected all the comments above except Joongi's one recommendation that suggests **moving python script in 'pr-number-assign.yml' under 'scripts/' directory** for the reason that can check in PR [ci: add changelog auto making workflow #977](https://github.com/lablup/backend.ai/pull/977).

What if the python script file is added from another PR?

leejiwon1125:
I think this isn't nice way but this may be the best way to ignore 'github actions' in commit process.

.github/workflows/pr-number-assign.yml
```diff
@@ -0,0 +1,80 @@
...
+            exit(0)
+
+        EOF
+        )
+        echo "$scripts" | python
+
+    - name: Make commit message for changing change log file
+      uses: stefanzweifel/git-auto-commit-action@v4
+      with:
+        commit_user_name: ${{ github.actor }}
```
leejiwon1125, line 78: I think this isn't nice way but this may be the best way to ignore 'github actions' in commit process: blank commit_user_email part so that default value "41898282+github-actions[bot]@users.noreply.github.com" couldn't set.

.github/workflows/pr-number-assign.yml
```diff
@@ -0,0 +1,80 @@
...
+        commit_user_email:
```
leejiwon1125, line 79: I think this isn't nice way but this may be the best way to ignore 'github actions' in commit process: blank commit_user_email part so that default value "41898282+github-actions[bot]@users.noreply.github.com" couldn't set.

leejiwon1125: I think this isn't nice way but this may be the best way to ignore 'github actions' in commit process: blank commit_user_email part so that default value "41898282+github-actions[bot]@users.noreply.github.com" couldn't set.

Yaminyam: I think it would be good to parse the information of the author of the pr and hand it over to the factor, but why do you think the method is bad?

.github/workflows/pr-number-assign.yml
```diff
@@ -0,0 +1,94 @@
...
+            exit(0)
+
+        EOF
+        )
+        echo "$scripts" | python
+
+    - name: Get user info
+      id: get_user_info
+      run: | 
+        echo "email=$(curl -Ls https://api.github.com/users/$GITHUB_ACTOR/events/public | jq -r '.[0].payload.commits[-1].author.email')" >> $GITHUB_OUTPUT
```
leejiwon1125: I didn't knew how exactly access to commit making person's email.
But using github api, I  can get email address of last committer now. @Yaminyam

.github/workflows/timeline-check.yml
```diff
@@ -5,7 +5,11 @@ on:
     types: [labeled, unlabeled, opened, synchronize, reopened]
 
 jobs:
+  pr-number-assign:
+    uses: ./.github/workflows/pr-number-assign.yml
+
   towncrier:
+    needs: [pr-number-assign]
     if: ${{ !contains(github.event.pull_request.labels.*.name, 'skip:changelog') }}
```
Yaminyam:
```suggestion
  pr-number-assign:
    if: ${{ !contains(github.event.pull_request.labels.*.name, 'skip:changelog') }}
    uses: ./.github/workflows/pr-number-assign.yml

  towncrier:
    needs: [pr-number-assign]
```

If the label 'skip: changelog' exists, I don't think the job needs to be executed.
And since the towncrier relies on pr-number-assign, I don't think you need the existing if syntax.

leejiwon1125: comment reflected!

.github/workflows/pr-number-assign.yml
```diff
@@ -0,0 +1,94 @@
...
+                exit(0)
+
+        if result:
+            print("There is not change log file for this PR in changes folder.")
+            print("Also, there is not feature named file in changes folder.")
+            exit(0)
+
+        EOF
+        )
+        echo "$scripts" | python
```
kyujin-cho: Any kind of script this length long needs to be pulled out as a separate file.

.github/workflows/pr-number-assign.yml
```diff
@@ -0,0 +1,94 @@
...
+        echo "email=$(curl -Ls https://api.github.com/users/$GITHUB_ACTOR/events/public | jq -r '.[0].payload.commits[-1].author.email')" >> $GITHUB_OUTPUT
+        echo "id=$(curl -Ls https://api.github.com/users/$GITHUB_ACTOR/events/public | jq -r '.[0].id')" >> $GITHUB_OUTPUT
+        echo "message=$(curl -Ls https://api.github.com/users/$GITHUB_ACTOR/events/public | jq -r '.[0].payload.commits[-1].message')" >> $GITHUB_OUTPUT
+
+    - name: Check user info
+      run: |
+        echo "Actor: $GITHUB_ACTOR"
+        echo "User email: ${{ steps.get_user_info.outputs.email }}"
+        echo "Recent id in github api(https://api.github.com/users/${{github.actor}}/events/public): ${{ steps.get_user_info.outputs.id }}"
+        echo "Workflow-Triggered commit message: ${{ steps.get_user_info.outputs.message }}"
```
kyujin-cho, line 38: Are these for debugging purpose only?

kyujin-cho (changes requested):

Yaminyam:
image
Is there a reason to make a new file and call the work that I add in this pr?
I think it is less readable to call other workflows with us as shown in the picture above.

And is there a reason for separating the pr that separates the python file?

leejiwon1125:
image
Due to low readability caused by dependency of pr-number-assign.yml and timeline-check.yml, I'll put pr-number-assign function into timeline-check.yml so that we can read nicer than now. @Yaminyam

.github/workflows/pr-number-assign.yml
leejiwon1125: I think those informations are needed for users. Should I delete this part?

achimnol (approved):

PR #991, opened 2023-01-25
refactor: Bgtask event client data structure

ChangHoon-Sung:
- current: produces the unnecessary event for every client
- this PR: clients are grouped by task id and produce an event to the client set with task id as a key.

ScreenShot 2023-01-25 at 13 46 37@2x


fregataa (commented):
As I know, it is our intention to `subscribe` all related events by clients.
But your work seems reasonable and nice because clients which do not have the task do not need to process anything about the task. 👍
I'd like to know other's opinion.

src/ai/backend/common/bgtask.py
```diff
@@ -92,12 +104,12 @@ async def _pipe_builder(r: Redis) -> Pipeline:
...
     event_producer: EventProducer
     ongoing_tasks: weakref.WeakSet[asyncio.Task]
-    task_update_queues: Set[asyncio.Queue[Sentinel | BgtaskEvents]]
+    task_update_queues: DefaultDict[uuid.UUID, Set[asyncio.Queue[Sentinel | BgtaskEvents]]]
 
     def __init__(self, event_producer: EventProducer) -> None:
         self.event_producer = event_producer
         self.ongoing_tasks = weakref.WeakSet()
-        self.task_update_queues = set()
+        self.task_update_queues = defaultdict(set)
```
kyujin-cho, line 113: I can't see the expression to create fresh set at desired task_id key.

src/ai/backend/common/bgtask.py
```diff
@@ -188,7 +201,9 @@ async def push_bgtask_events(
                 finally:
                     return resp
         finally:
-            self.task_update_queues.remove(my_queue)
+            self.task_update_queues[task_id].remove(my_queue)
+            if len(self.task_update_queues[task_id]) == 0:
+                del self.task_update_queues[task_id]
```
kyujin-cho: Is there a consideration for concurrency (e.g. handling multiple `push_bgtask_events()` calls to same task_id)?

kyujin-cho (changes requested):

src/ai/backend/common/bgtask.py
changhoon-sung: defaultdict will create the fresh set for the given task_id key if given key doesn't exist in the dictionary

kyujin-cho: Oh that's awesome.

ChangHoon-Sung:
Oops, it was just rebase.

src/ai/backend/common/bgtask.py
changhoon-sung: I missed it. I added the critical section on the creation and deletion of the set with asyncio.Lock.
Please check the latest commit. (The rebase messed up the commit history;)

kyujin-cho (approved):

Issue #865, opened 2022-11-09
Support cgroups v2


achimnol:
Ubuntu 22.04 now uses cgroups v2 by default.
Our agent should be seamlessly detect it and work with it.

Refs:
* https://medium.com/some-tldrs/tldr-understanding-the-new-control-groups-api-by-rami-rosen-980df476f633
* https://thenewstack.io/linux-cgroups-v2-brings-rootless-containers-superior-memory-management/
* https://kubernetes.io/docs/concepts/architecture/cgroups/

achimnol:
Resolved via the combination of:
- #992
- #1000
- #1005

PR #992, opened 2023-01-25
feat: Support cgroup v2 on get_available_cores()

ChangHoon-Sung:
This PR is part of #865 

The source of the number of available cores for the docker daemon now changed.

**Before**
- In cgroup v1, `/sys/fs/cgroup/cpuset/docker/cpuset.cpus`

**After**
- In cgroup v2, `/sys/fs/cgroup/system.slice/docker.service/cpuset.cpus.effective` 
- In cgroup v1, `/sys/fs/cgroup/cpuset/docker/cpuset.effective_cpus`

**Reason**
- In cgroup v2, `cpuset.cpus` represents the number of requested cores, not the number of available cores. However, it may be ignored by the parent cgroup and the actual number of available cores may be less than `cpuset.cpus`.
- But `cpuset.effective_cpus` and `cpuset.cpus.effective` represents the actual number of available cores, considering these constraints, so it is reasonable to use them.
- In cgroup v1, `cpuset.cpus` represents the number of available cores, same as `cpuset.effective_cpus`, but I just synchronized with cgroup v2 for better understand.

### DIFF
--- a/changes/992.feature.md
+++ b/changes/992.feature.md
@@ -1,0 +1,1 @@
+Support cgroup v2 on get_available_cores
--- a/src/ai/backend/agent/vendor/linux.py
+++ b/src/ai/backend/agent/vendor/linux.py
@@ -3,7 +3,6 @@ import ctypes.util
 import logging
 import os
 import sys
-from pathlib import Path
 from typing import Iterator
 
 import aiohttp
--- a/src/ai/backend/agent/vendor/linux.py
+++ b/src/ai/backend/agent/vendor/linux.py
@@ -82,18 +82,17 @@ class libnuma:
 
                             match driver:
                                 case "cgroupfs":
-                                    cgroup = "docker"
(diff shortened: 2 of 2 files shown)


ChangHoon-Sung:
The location of cgroup resources depends on not only cgroup version, but also cgroup driver of docker daemon (See #1000)
The PR above only covers cgroup version so I'm working on covering cgroup driver.

ChangHoon-Sung:
There was a misunderstanding about docker.service in cgroup paths. At first, docker.service or docker daemon cgroup path was considered as all container's cgroup parent but it is not true. Docker provides `CgroupParent` option. See [Docker docs](https://docs.docker.com/engine/reference/commandline/dockerd/#:~:text=log%202%3E%261-,Default%20cgroup%20parent,-The%20%2D%2Dcgroup%2Dparent).
Following commits fix the cgroup parent path.

sanxiyn (approved):

PR #993, opened 2023-01-25
bugfix: container does not start on macos after #986

inureyes:
This PR resolves bind mount error when container starts on Docker for mac + virtioFS (#986)


achimnol (approved):
Ah, it was missing in #986! Thanks!

PR #995, opened 2023-01-26
fix: `mpirun` not working occassionally on muti-node session

kyujin-cho:
This PR fixes issue `mpirun` failing to work when multiple host is provided on multi-node session.


src/ai/backend/agent/docker/agent.py
```diff
@@ -957,6 +967,13 @@ async def __ainit__(self) -> None:
                 )
         await super().__ainit__()
         await self.check_swarm_status()
+        try:
+            async with Docker() as docker:
+                gwbridge = await docker.networks.get("docker_gwbridge")
+                gwbridge_info = await gwbridge.show()
+                self.gwbridge_subnet = gwbridge_info["IPAM"]["Config"][0]["Subnet"]
+        except DockerError:
```
achimnol: I think it would be safer to catch `KeyError` and `IndexError` as well here, for when `gwbridge_info` has an unexpected structure.

achimnol (changes requested):

achimnol (approved):

PR #998, opened 2023-01-26
setup: Upgrade Pants to 2.16.0.dev5

achimnol:
Pants 2.16 introduces the official **linux-aarch64** binary wheels, finally!

Also it includes some notable new features like:
- The `--resolve` arguments to `./pants export` to recreate virtualenvs for specific resolves/lockfiles only.
- The intrinsic `env()` function in `BUILD` files to interpolate environment variables with defaults
- Faster pytest execution with optional `batch_compatibility_tag` to group multiple pytest targets into a single pytest process and run expensive high-level fixtures only once

## What's included

- Update of `./pants` to the latest one
- Update of the Pants plugin API in our own plugin codes
- Removal of custom-build / source-build related codes in `install-dev.sh` and `./pants`
- Update of all occurrences in the documentation mentioning `./pants export ::` to say `./pants export`
- Minor fix from #999 
- Regenerated lockfiles

## Tests

- Confirmed working of daily workflows (e.g., `./pants lint ::`, ...) in both linux-aarch64 and macos-arm64 platforms
- Confirmed correct build of wheel packages with platform-specific binaries (e.g., `./pants --tag="wheel" package src/ai/backend/runner::`) in both linux-aarch64 and macos-arm64 platforms
- Confirmed all CI workflows working fine

## Migration Guide

```console
$ rm -rf ~/.cache/pants/* .tmp/immutable*
$ ./pants version
$ ./pants export
```

NOTE: You should _not_ use `::` when running `./pants export` any more, because the `export` command has become more consistent with the `generate-lockfiles` command and now prefers explicit `--resolve` arguments.

PR #999, opened 2023-01-26
refactor: mark with "# pants: no-infer-dep" comment

studioego:
Mark with "# pants: no-infer-dep" comment
The python 1 line code is indirect dependency of other explicit dependencies


achimnol:
This is already included in #998! :wink:

achimnol (approved):

PR #1000, opened 2023-01-26
feat: Generate cgroup path from driver and version

### LINKED ISSUE #363: Generalize docker cgroup driver support

sanxiyn:
This PR automatically detects Docker's cgroup driver(can be "cgroupfs" or "systemd") and cgroup version(can be "1" or "2"), and generates cgroup path accordingly.

This PR does not address cgroup v2 support(#865) yet. While correct directory path is generated, cgroup v1 and v2 also differ in file name and format. For example, `cpu.stat` should be read instead of `cpuacct.usage`. This will be addressed in the following PR.

This PR was tested on Ubuntu 20.04 (cgroup v1 system), with both cgroup drivers. Docker uses cgroupfs driver by default on cgroup v1 system. Create `/etc/docker/daemon.json` as follows to let Docker use systemd driver.

```json
{"exec-opts": ["native.cgroupdriver=systemd"]}
```

Following Backend.AI configuration is needed to test:

```toml
# agent.toml

[container]
stats-type = "cgroup"

[debug]
log-stats = true
```

This is an alternative to #948.

Fix #363.


achimnol (approved):
Discussed in person, and LGTM!

PR #1003, opened 2023-01-27
fix: `--limit` options type

### LINKED ISSUE #1001: CLI `--limit` options have a wrong type

Linux (aarch64)

### Backend.AI version

main

### Describe the bug

image

Many CLI commands which have `--limit` option occurs wrong-type error.
We should set the options as `int` type

### To Reproduce

e.g command
```
./backend.ai admin user list --limit 2
```

### Expected Behavior

_No response_

### Anything else?

_No response_

Yaminyam:
close #1001 

Changed the value of --limit option to recognize as int type that was being treated as str.


fregataa (changes requested):
the list cmd of agent, keypair, storage, session, user, vfolder have `--limit` option and only session's list have `int` type on the option.
please check all of those commands, not only user's cmd.

changes/1003.fix.md
```diff
@@ -0,0 +1 @@
+Changed the value of `--limit` option to recognize as int type that was being treated as str.
```
kyujin-cho: We're changing "type" of `--limit` option, not "value". 😉
```suggestion
Fix `--limit` option being treated as `str` instead of `int`
```

kyujin-cho (changes requested):

kyujin-cho (approved):

PR #1005, opened 2023-01-27
feat: Stop using pids cgroup controller

sanxiyn:
Instead of using pids cgroup controller to list all PIDs in the given container, get the container ID of the given PID directly from `/proc/PID/cgroup`. 

cc #865


ChangHoon-Sung:
LGTM!

PR #1006, opened 2023-01-27
feat: Support cgroup v2 for container stat measures

ChangHoon-Sung:
This is the following PR of #1000
This PR makes measuring container stats supports cgroup v2.

Like #1000, Following Backend.AI configuration is needed to test:

```toml
# agent.toml

[container]
stats-type = "cgroup"

[debug]
log-stats = true
```


sanxiyn (approved):

sanxiyn:
cc #865

PR #1008, opened 2023-01-28
feat: apply `VERIFY_PASSWORD_FORMAT` hook for signup handler

adrysn:
Major changes of this PR:
- Manager
  - Change the parameters for `VERIFY_PASSWORD_FORMAT`
    - (`old_password`, `new_password1`) -> (`request`, `params`)
    - The hook requires some information in the `request` object, such as user's email, to check the password similarity with user's information.
  - Apply the `VERIFY_PASSWORD_FORMAT` hook to the signup handler to enforce the password policy for new users.
  - Display `email` in vfolder API handler logging. What site admins want to know is the requester's name or email, not the access key.
- Webserver
  - Log client IP in login handler. There are requests for admins to let them know the source IP when there are many failed login attempts.
    ```shell
    LOGIN_HANDLER: Too many consecutive login fails (email:admin@a.bc, count:10, ip:10.8.0.14)
    ```
  - DRY: Create a separate function to add `X-Forwarded-For` header in an API session. In each API session, the function should be manually called, if required.
    - Would there be a good way to automatically add the header without manual calling?


src/ai/backend/manager/api/auth.py
```diff
@@ -922,19 +935,19 @@ async def update_password(request: web.Request, params: Any) -> web.Response:
...
         "VERIFY_PASSWORD_FORMAT",
-        (params["old_password"], params["new_password"]),
+        (request, params),
         return_when=ALL_COMPLETED,
     )
     if hook_result.status != PASSED:
         hook_result.reason = hook_result.reason or "invalid password format"
         raise RejectedByHook.from_hook_result(hook_result)
 
+    raise AuthorizationFailed("Test Error")
```
fregataa: Please remove test err!

fregataa (changes requested):

src/ai/backend/manager/api/auth.py
adrysn: Oops... I removed the code. Thanks.

src/ai/backend/web/auth.py
```diff
@@ -81,3 +81,21 @@ async def get_anonymous_session(
...
+    return client_ip
+
+
+def fill_x_forwarded_for_header_to_api_session(
+    request: web.Request, api_session: APISession
+) -> None:
+    client_ip = get_client_ip(request)
+    if client_ip:
+        _headers = {"X-Forwarded-For": client_ip}
+        api_session.aiohttp_session.headers.update(_headers)
```
kyujin-cho, line 101: Is there a specific reason for removing fallback logic of value of "X-Forwarded-For"?

kyujin-cho (changes requested):

src/ai/backend/web/auth.py
kyujin-cho: My mistake. The fallback logic is covered in `get_client_ip()`.

kyujin-cho (approved):

PR #1009, opened 2023-01-29
feat: Add db dump scripts

fregataa:


fregataa:
close this PR since it is outdated and no more work on it

PR #1010, opened 2023-01-30
deps: replace snappy with zlib

kyujin-cho:
This PR removes dependency to external module `python-snappy` and refactors compression/decompression steps to use zlib.


achimnol (approved):
LGTM!

PR #1015, opened 2023-02-03
feat: sync container commit status using redis

fregataa:
Fetch commit status through redis rather direct RPC call


src/ai/backend/agent/agent.py
```diff
@@ -703,6 +709,60 @@ async def produce_error_event(
...
+                all_kern_statuses = await pipe.hgetall(hash_name)
+                deleting_statuses: list[str] = []
+                for kern_id in self.kernel_registry:
+                    str_kern_id = str(kern_id)
+                    if str_kern_id in all_kern_statuses and str_kern_id not in ongoing_kernel_map:
+                        deleting_statuses.append(str_kern_id)
+
+                await pipe.hdel(hash_name, *deleting_statuses)
+                await pipe.hset(hash_name, mapping=cast(Mapping, ongoing_kernel_map))
+            return pipe
```
fregataa: This part should be replaced to Lua script.
Here is an example
https://github.com/lablup/backend.ai/blob/80f10b08ab91a0386b1a39b99662f955a26f7cca/src/ai/backend/manager/api/stream.py#L507-L533

PR #1017, opened 2023-02-06
feature: Add vfolder status in api response and update cli to follow

fregataa:
- Add vfolder status in api response
- Update vfolder cli `list`, `info` commands to show vfolder's status


Sujin-Kim1 (approved):
LGTM

PR #1018, opened 2023-02-06
feat: delete vfolders asynchronously

fregataa:
delete vfolder under `PersistentTaskGroup`

### DIFF
--- a/src/ai/backend/manager/models/vfolder.py
+++ b/src/ai/backend/manager/models/vfolder.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import enum
+import logging
 import os.path
 import uuid
 from pathlib import PurePosixPath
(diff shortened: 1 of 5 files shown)


src/ai/backend/manager/models/vfolder.py
```diff
@@ -734,6 +739,53 @@ async def filter_host_allowed_permission(
     return allowed_hosts
 
 
+async def delete_vfolder_by_ids(
+    db_session: SASession,
+    storage_manager: StorageSessionManager,
+    storage_ptask_group: aiotools.PersistentTaskGroup,
+    *,
+    vfolder_infos: Sequence[tuple[uuid.UUID, str]],  # id, host
```
kyujin-cho: Using `tuple` to represent complex data type like this can make code reader hard to understand the context. Please rewrite this to use `TypedDict` or `NamedTuple`.

kyujin-cho (changes requested):

src/ai/backend/manager/models/vfolder.py
fregataa: Updated. thank you!

PR #1019, opened 2023-02-06
fix: hotfix session migration

fregataa:
- none check on comparing row value


kyujin-cho:
Follow-up PR of #576.

PR #1022, opened 2023-02-07
feat: Let the queryfilter-minilang engine to fetch multiple rows from joined tables

fregataa:
- enable query multiple rows from joined tables
- In keypair graphene object, `project` is added

Test CLI cmd
```
./backend.ai admin keypair list --order "projects" --filter 'projects contains "default"'
```


src/ai/backend/manager/models/minilang/ordering.py
```diff
@@ -22,23 +23,31 @@
...
-                col = self._sa_table.c[self._column_map[col_name]]
+                col_value = self._column_map[col_name]
+                if isinstance(col_value, tuple) or isinstance(col_value, list):
+                    if len(col_value) == 1:
+                        col = col_value[0]
+                    else:
+                        func = col_value[1]
+                        col = func(col_value[0])
+                else:
+                    col = self._sa_table.c[self._column_map[col_name]]
```
fregataa: This branch seems not good.
Should I change all `_queryorder_colmap` from `Mapping[str, str]` into `Mapping[str, tuple]`?

achimnol (commented):
Could we use `array_agg` instead of `string_agg` so that we could return the multiple belonging projects as a structured list?

fregataa:
> Could we use `array_agg` instead of `string_agg` so that we could return the multiple belonging projects as a structured list?

I add `array_agg` function. thank you for your review!

achimnol (changes requested):
Please update the test case!

achimnol (approved):

Issue #1024, opened 2023-02-08
Upgrade hiredis to 2.2.1


studioego:
### Main idea

Hiredis 2.2.1 was released on Feb 2, 2023.

https://github.com/redis/hiredis-py/pull/152

A few months ago, I found hiredis 2.0.0 bug on Backend.AI 
(Below link is my submitted hireis 2.0.0 bug issue link https://github.com/lablup/backend.ai/issues/609 )

As a result, I rolled back the hiredis 2.0.0 to 1.1.0  https://github.com/lablup/backend.ai/pull/636 

Also, I submitted hiredis-py PR https://github.com/redis/hiredis-py/pull/129 .

But, Hiredis version upgrade to 2.2.1. 
I posted the issue [change git submodule url scheme & .travis.yml: make use of CFLAG=-std=c99](https://github.com/redis/hiredis-py/pull/129) on github that there are no wheels to support old gcc versions (manylinux1, centos7 with gcc 4.8.5). 
Now, hiredis 2.2.1 supports old gcc versions (manylinux1, centos7 with gcc 4.8.5) and  multi-platforms wheels.

Let's upgrade hiredis to 2.2.1 on Baceknd.AI.

PR #1025, opened 2023-02-08
setup: Upgrade hiredis to 2.2.1

studioego:
setup: Upgrade hiredis to 2.2.1
 - support some wheels for old gcc versions and multi-platform wheels

resolves https://github.com/lablup/backend.ai/issues/1024

See details 
* https://github.com/redis/hiredis-py/pull/152
* https://pypi.org/project/hiredis/2.2.1/


studioego:
I installed new CentOS7 image and Backend.AI on new CentOS7 image.

image
image

It works correctly on both CentOS7 and CentOS8(Also, I checked MacOS M1(arm64) and Fedora core 37(x86_64)

kyujin-cho (changes requested):
Have you checked about the changes made to hiredis when they upgrade major version to 2? I'm frustrated that this might potentially break our code. Please let me know.

achimnol:
> Have you checked about the changes made to hiredis when they upgrade major version to 2? I'm frustrated that this might potentially break our code. Please let me know.

hiredis is used by redis-py as an internal parser replacement to accelerate its base implementation written in Python. It should have no impacts to our codes.

PR #1027, opened 2023-02-08
chore: update graphene to v3

fregataa:
resolve: #1028


fregataa:
close this PR since we have not decided to upgrade graphene to v3

PR #1029, opened 2023-02-09
fix: update concurrency by number of session rather kernel

fregataa:
Session concurrency has been calculated depending on the number of kernels, not sessions.
Let's calculate concurrency by the number of sessions.

### DIFF
--- a/changes/1029.fix.md
+++ b/changes/1029.fix.md
@@ -1,1 +1,1 @@
-Update concurrency by number of session rather kernel.
+Update `concurrency_used` per access key by number of sessions rather kernels.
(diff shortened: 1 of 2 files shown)


lizable (approved):
LGTM.

changes/1029.fix.md
```diff
@@ -0,0 +1 @@
+Update concurrency by number of session rather kernel.
```
kyujin-cho:
```suggestion
Update `concurrency_used` per access key by number of sessions rather kernels.
```

kyujin-cho (changes requested):

Issue #898, opened 2022-11-24
Improve handling of multiple concurrent docker image pulls in agents


achimnol:
Docker's image pull API works like a singletone queue inside the docker daemon, which means that:
- Regardless how many docker API clients request the pull of a single image, the pull operation occurs only once and all clients are notified with the same progress.
- If there are multiple pull requests for different images, they are queued and processed serially by the docker daemon.

It seems that our agent (and maybe the manager) gets stuck when there are multiple concurrent image pulls.  Let's investigate what's happening and improve the handling and progress reports.

Issue #306, opened 2021-11-01
Refactor the bgtask framework as a common module


achimnol:
Currently the background task framework is only being used by several internals of the manager:
* image rescan progress tracking
* kernel pull progress tracking

We have plans to extend the storage-proxy to provide long-running tasks such as cloning vfolders,
and this requires that storage-proxy should also be able to make progress reports.

Let's move the background task framework implementation from `manager.background` to `ai.backend.common.bgtask`, and reuse across manager, storage-proxy, and (possibly) agent as well.

PR #1033, opened 2023-02-10
feat: Spinner on client

ChangHoon-Sung:
Spinners are intended to be used in long background task which doesn't provide progress.
(If the background task can provide the progress, we can show the progress more visually by using like tqdm)

ScreenShot 2023-02-10 at 15 11 05

The use case is `vfolder clone`
 - Reporting the progress from storage-proxy is not supported yet.
 - Storage-proxy has various backends (VFS, PureStorage, CephFS, ..), and most of the storage backend operation doesn't provide the progress feature.
 - Making our own operation on each backend for reporting the progress is not easy.
 - Currently, if there is a large amount of files that need to be cloned, it will seem as if they are stuck on the client.
 
Tested on VFS.

refs: #898

### DIFF
--- a/src/ai/backend/client/cli/vfolder.py
+++ b/src/ai/backend/client/cli/vfolder.py
@@ -6,7 +6,6 @@ from pathlib import Path
 import click
 import humanize
 from tabulate import tabulate
-from tqdm import tqdm
 
 from ai.backend.cli.interaction import ask_yn
 from ai.backend.cli.main import main
@@ -17,7 +16,7 @@ from ai.backend.client.session import Session
 from ..compat import asyncio_run
 from ..session import AsyncSession
 from .params import ByteSizeParamCheckType, ByteSizeParamType, CommaSeparatedKVListParamType
-from .pretty import print_done, print_error, print_fail, print_info, print_wait, print_warn
+from .pretty import Spinner, print_done, print_error, print_fail, print_info, print_wait, print_warn
 
 
 @main.group()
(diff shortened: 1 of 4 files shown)


src/ai/backend/client/cli/vfolder.py
```diff
@@ -711,24 +710,21 @@ def clone(name, target_name, target_host, usage_mode, permission):
...
             try:
                 bgtask = session.BackgroundTask(bgtask_id)
                 completion_msg_func = lambda: print_done("Cloning the vfolder is complete.")
                 async with bgtask.listen_events() as response:
-                    # TODO: get the unit of progress from response
-                    with tqdm(unit="bytes", disable=True) as pbar:
+                    async with Spinner(
+                        "Cloning the vfolder... "
+                        "(This may take a while depending on its size and number of files!)",
+                    ):
```
fregataa, line 723:
```
async with (
  bgtask.listen_events() as response,
  Spinner(
    "Cloning the vfolder... "
    "(This may take a while depending on its size and number of files!)",
  ):
  ...
```
we can use [parenthesized context manager](https://docs.python.org/3.10/whatsnew/3.10.html#parenthesized-context-managers)

fregataa (changes requested):
How many commands would you apply Spinner?

ChangHoon-Sung:
> How many commands would you apply Spinner?

Spinner can be applied in backgroud task which is refactored by #306,  but only `vfolder clone` and `image rescan` support the refactored backgroud task.
Rescanning the image can get the progress from the docker API, it doesn't need a spinner, so the `vfolder clone` is the only one that applied a spinner.

There're several commands that can be refactored to background task on storage-proxy command like delete, cp, etc.
I think, in a long term, all asynchronous commands may be integrated with background task.

src/ai/backend/client/cli/vfolder.py
changhoon-sung: Oh that's good!

changhoon-sung: The discussion is ongoing about how to use the spinner and progress, so I'll leave it for now and fix it after the discussion.

kyujin-cho (approved):

PR #1036, opened 2023-02-10
feature: get enum type `queryfilter` field according to DB column type

fregataa:
CLI commands only allow capital letters for filtering fields like below, although the value is saved as capital letters or small letters.
```
# users status field is saved as small letters.

# works well
./backend.ai admin user list --filter 'status == "ACTIVE"'

# also works well
./backend.ai admin user list --filter 'status == "active"'
```

This PR matches `queryfilter` to column type.
```
# works well
./backend.ai admin user list --filter 'status == "active"'
```

and add user friendly message to correct enum type fields
image


ChangHoon-Sung (commented):
LGTM!

achimnol (commented):
How about supporting case-insensitive match for more human-friendliness?

fregataa:
> How about supporting case-insensitive match for more human-friendliness?

at first, I implemented the human friendly case-insensitive match. but later, I and @kyujin-cho  discussed about it and we decided not to support case-insensitive match because we guessed that we made `EnumType` and `EnumValueType` individually to differentiate those values and do not want users get confused of it. but I am not sure what should I do.

fregataa:
> at first, I implemented the human friendly case-insensitive match. but later, I and @kyujin-cho discussed about it and we decided not to support case-insensitive match because we guessed that we made `EnumType` and `EnumValueType` individually to differentiate those values and do not want users get confused of it. but I am not sure what should I do.

We decided to implement case-insensitive match.

src/ai/backend/manager/models/minilang/queryfilter.py
```diff
@@ -50,6 +51,20 @@
 FieldSpecType: TypeAlias = Mapping[str, FieldSpecItem] | None
 
 
+def enum_field_getter(enum_cls: Type[Enum]) -> Callable[[str], Enum]:
```
achimnol: Could we use generic types here? Or, is it sufficient to use just `enum.Enum`?

src/ai/backend/manager/models/minilang/queryfilter.py
```diff
@@ -50,6 +51,20 @@
...
+    def get_enum(value: str) -> Enum:
+        try:
+            return enum_cls[value]
+        except KeyError:
+            try:
+                return enum_cls(value)
+            except ValueError:
+                enum_names = ", ".join([e.value for e in enum_cls])
+                raise ValueError(f"expected one of `{enum_names}`, got `{value}`")
```
achimnol: A very minor suggestion to disambiguate the commas:
```suggestion
                raise ValueError(f"expected one of `{enum_names}`; got `{value}`")
```

changes/1036.feature.md
```diff
@@ -0,0 +1 @@
+Enable small letter string for query filtering item.
```
achimnol: The implementation is about allowing both raw values and names of the enum constants, not allowing just case-insensitive matching.

If this message is the original intention, the implementation should be *rewritten* to do a case-insensitive match instead of checking both values and names of enum items.

Let's be clear.

achimnol (changes requested):

changes/1036.feature.md
achimnol: Note: enum values are usually implementation details, i.e., hidden under the abstraction. I think it would be better to stick on the names only with relaxation of the case sensitivity when matching them.

fregataa: I agree with you. I updated the code to enable only enum names, with small letter string.

src/ai/backend/manager/models/minilang/queryfilter.py
fregataa: I tried but `enum.Enum` does not pass type-checking and I have no idea how to use generic types here.
this is what failure I got when I run type-checking with `enum.Enum`.
image

achimnol: I've updated this part by myself.

achimnol (approved):

PR #1045, opened 2023-02-13
fix: correct query filter and order spec of session

fregataa:
correct query filter and query order spec of session resolver.
this is a side effect of #576 , session/kernel table division.


src/ai/backend/manager/models/session.py
```diff
@@ -1216,49 +1216,48 @@ async def resolve_abusing_reports(
...
 
     _queryfilter_fieldspec = {
-        "session_type": ("session_type", lambda s: SessionTypes[s]),
-        "name": ("name", None),
-        "image": ("image", None),
-        "architecture": ("architecture", None),
-        "domain_name": ("domain_name", None),
-        "group_name": ("groups_group_name", None),
+        "id": ("sessions_id", None),
+        "session_type": ("sessions_session_type", lambda s: SessionTypes[s]),
```
rapsealk: I wonder if the prefix "session_" is necessary since it is not aligned with the others.
```suggestion
        "type": ("sessions_session_type", lambda s: SessionTypes[s]),
```

src/ai/backend/manager/models/session.py
```diff
@@ -1216,49 +1216,48 @@ async def resolve_abusing_reports(
...
     _queryorder_colmap = {
-        "id": "id",
-        "session_type": "session_type",
-        "name": "name",
-        "image": "image",
-        "architecture": "architecture",
-        "domain_name": "domain_name",
-        "group_name": "group_name",
+        "id": "sessions_id",
+        "session_type": "sessions_session_type",
```
rapsealk: Same here.
```suggestion
        "type": "sessions_session_type",
```

rapsealk (approved):
Approved with some comments!

src/ai/backend/manager/models/session.py
fregataa: thanks! I updated the codes

kyujin-cho (approved):

PR #1047, opened 2023-02-13
fix: update query statement to check occupying resource from session related kernels

fregataa:
Update query statement for checking resource occupancy of kernel Since kernel table's `scaling_group` column is empty and has been migrated to session table.


lizable (approved):
LGTM.

PR #1057, opened 2023-02-15
feature: Phase 1 of Model Serving

rapsealk:
This is a phase-1 implementation to support user-defined model inference APIs.

- [x] New session type: `"INFERENCE"` (@rapsealk)
- [ ] Add a new API to start a service (instead of a compute session) with options to mount a specific model vfolder
- [x] Add a new conceptual entity `endpoint` to abstract the URL endpoint where one or more specific models/versions are bound (@rapsealk)
- [x] Add a new mapping, i.e., `routing` table, to store the mappings between endpoints and inference sessions with additional properties like traffic ratio.
- [ ] Add a new filtered view of service sessions in the webui and control-panel
- [x] Update the image label spec (lablup/backend.ai-internal#75)
- [x] Add `backend.ai serve` command group (@fregataa)
  - CRUD of endpoints
  - Should respond with 404 (no such session) if called against non-inference sessions
- [x] Add `backend.ai model` command group (@fregataa)
  - CRUD of model vfolders


achimnol (approved):
This is a preliminary prototype version. We will have more follow-up PRs to concretize the implementation and APIs.

Issue #930, opened 2022-12-15
'ImageRow' object has no attribute 'installed'


rapsealk:
### What Operating System(s) are you seeing this problem on?

macOS (Apple Silicon)

### Backend.AI version

22.09.6

### Describe the bug

Trying to list installed images, I faced an error:
```bash
2022-12-15 12:24:12.090 ERROR ai.backend.manager.cli.image_impl [20709] An error occurred.
Traceback (most recent call last):
  File "/Users/rapsealk/Desktop/git/backend.ai-dev/src/ai/backend/manager/cli/image_impl.py", line 54, in list_images
    if installed and not item.installed:
AttributeError: 'ImageRow' object has no attribute 'installed'
```

It tries to access the `installed` field of an `ImageRow` object,
https://github.com/lablup/backend.ai/blob/174b844e5c8bb321157027d9efbff45d6a329642/src/ai/backend/manager/cli/image_impl.py#L50-L54

but `ImageRow` has no attribute/property named 'installed'.
https://github.com/lablup/backend.ai/blob/174b844e5c8bb321157027d9efbff45d6a329642/src/ai/backend/manager/models/image.py#L157-L193

### To Reproduce

```bash
% ./backend.ai mgr image list --installed

2022-12-15 12:24:12.090 ERROR ai.backend.manager.cli.image_impl [20709] An error occurred.
Traceback (most recent call last):
  File "/Users/rapsealk/Desktop/git/backend.ai-dev/src/ai/backend/manager/cli/image_impl.py", line 54, in list_images
    if installed and not item.installed:
AttributeError: 'ImageRow' object has no attribute 'installed'
```

rapsealk:
We may just prevent error by replacing
https://github.com/lablup/backend.ai/blob/174b844e5c8bb321157027d9efbff45d6a329642/src/ai/backend/manager/cli/image_impl.py#L54-L55
with
```python
        if installed and not (hasattr(item, "installed") and item.installed):
            log.warning("NOTE: installed/installed_agents fields are no longer provided in CLI, until we finish the epic refactoring of image metadata db.")
            return
```

PR #1063, opened 2023-02-16
fix: Replace accessing non-existent `ImageRow.installed` with Redis queries in the manager CLI

san-kim0106:
As specificed in #930, the `$ ./backend.ai mgr image list --installed` command runs into an errer as the program tries to access an undefined class attribute `ImageRow().installed`. The below code fixes this error by querying the `redis` database that stores image metadata stored on the each agent.
Below is a short illustration of the command, and one can now check that `python3.9` image is installed on agent `SanKim`.

https://user-images.githubusercontent.com/86462416/219317232-53cadd4a-415c-40fb-8b27-3437b5a5f0ba.mov


achimnol (approved):

achimnol:
I've rewritten the implementation to provide cleaner output using `pformat()` appropriately when interpolating objects inside a formatted string and apply batching to reduce the number of round-trips to the redis server.

PR #1073, opened 2023-02-20
feat: shorten response time of `image_list` GQL query

kyujin-cho:

### DIFF
--- a/src/ai/backend/manager/models/image.py
+++ b/src/ai/backend/manager/models/image.py
@@ -8,6 +8,7 @@ from pathlib import Path
 from typing import (
     TYPE_CHECKING,
     Any,
+    AsyncIterator,
     List,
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/models/image.py
```diff
@@ -597,11 +630,8 @@ async def load_all(
     ) -> Sequence[Image]:
         async with ctx.db.begin_readonly_session() as session:
             rows = await ImageRow.list(session, load_aliases=True)
-        items: List[Image] = []
+        items = [gen async for gen in cls.bulk_load(ctx, rows)]
```
achimnol: `gen` is not a generator but it's a single yielded item.
How about renaming it to `item`?

achimnol (changes requested):

PR #1083, opened 2023-02-21
fix: Update session-related events, introducing `SessionTerminating` and using proper session IDs

rapsealk:
I found that several `Session`-related events are using `kernel_id` for its own field `session_id`, whilst any `session.id` and `kernel.id` does not correspond to each other any more.
In this PR, `SessionSuccessEvent` and `SessionFailureEvent` are changed to use the right value(`session_id`).

https://github.com/lablup/backend.ai/blob/29f4a12ecc73a12ab414f72726c08c4568990756/src/ai/backend/agent/agent.py#L1375-L1397


src/ai/backend/agent/server.py
```diff
@@ -494,12 +496,14 @@ async def execute(
     @collect_error
     async def execute_batch(
         self,
+        session_id: str,
```
sanxiyn: This RPC function is unused and can be removed.

rapsealk: Thanks for the feedback! This function has been removed by commit d9bb407.

achimnol (approved):

PR #1084, opened 2023-02-21
ci: more stable container creation for halfstack services

kyujin-cho:
This PR updates halfstack container fixtures to wrap container creation with filelock so that duplicate ports won't be assigned.


achimnol (approved):
LGTM!

PR #1090, opened 2023-02-22
feat: Set process title for kernel runner

sanxiyn:
This prevents kernel runner from being killed by killall python.

### DIFF
--- a/src/ai/backend/kernel/__main__.py
+++ b/src/ai/backend/kernel/__main__.py
@@ -19,8 +19,10 @@ def setproctitle(title):
     # setproctitle package doesn't work for unknown reasons
     # We don't need a portable implementation, so here is a Linux-only version
     import ctypes
+    import ctypes.util
 
-    libc = ctypes.CDLL("libc.so.6")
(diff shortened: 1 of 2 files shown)


src/ai/backend/kernel/__main__.py
```diff
@@ -15,6 +15,17 @@
 from .compat import asyncio_run_forever
 
 
+def setproctitle(title):
+    # setproctitle package doesn't work for unknown reasons
+    # We don't need a portable implementation, so here is a Linux-only version
+    import ctypes
+
+    libc = ctypes.CDLL("libc.so.6")
```
achimnol: How about using [`ctypes.util.find_library("c")`](https://docs.python.org/3/library/ctypes.html#finding-shared-libraries) to avoid hardcoding the libc version?

achimnol (changes requested):

src/ai/backend/kernel/__main__.py
sanxiyn: Sure.

achimnol (approved):

PR #1094, opened 2023-02-23
feat: introduce event stream plugin

rapsealk:
This PR utilizes a new plugin, [backend.ai-event-stream-plugin](https://github.com/lablup/backend.ai-event-stream-plugin), to write an event stream for the external services(e.g. Backend.AI FastTrack).

Reference: [lablup/backend.ai-event-stream-plugin](https://github.com/lablup/backend.ai-event-stream-plugin)


rapsealk:
ce0f117 has to be reverted after #1083 is merged.

src/ai/backend/manager/defs.py
```diff
@@ -59,6 +59,7 @@
 REDIS_IMAGE_DB: Final = 3
 REDIS_STREAM_DB: Final = 4
 REDIS_STREAM_LOCK: Final = 5
+REDIS_EVENT_STREAM_DB: Final = 1
```
kyujin-cho: The name of this constant creates confusion when combined with values defined above. If this value is only for a separate redis cluster, then please rename it.

kyujin-cho (changes requested):

PR #1096, opened 2023-02-23
fix: Preserve image-provided path env-vars

### LINKED ISSUE #1095: Preserve image-defined path environment variables
Currently the krunner is overriding `PYTHONPATH`. Let's make it to preserve the image-defined value and prepend necessary stuffs.

achimnol:
resolves #1095


adrysn (approved):
LGTM!

PR #1097, opened 2023-02-23
feat: Support scanning the local Docker daemon's image list

### LINKED ISSUE #1089: Add `backend.ai mgr image rescan --local` option
Let's add an option `--local` to the `mgr image rescan` command which allows scanning the image metadata directly from the local Docker daemon, instead of pushing to a remote registry, for ease of debugging and development of kernel images.

achimnol:
resolves #1089

Note that I don't add `--local` option to the client CLI's `admin image rescan` command because this is intended for local all-in-one development setups where you can directly access the manager CLI command `mgr image rescan`.

### Known restriction
When starting a new session via CLI, you must prepend "index.docker.io/library/" to your locally built image names if they don't have hostname/path prefixes.

Examples:
- `bai-python:3.11` -> `index.docker.io/library/bai-python:3.11`
- `localdev/bai-python:3.11` -> `index.docker.io/localdev/bai-python:3.11`

You also need to add `index.docker.io` to the allowed registry list of the `default` domain.


achimnol:
Remaining TODO: Update the existing test suite

achimnol:
TODO: prevent duplication of pulled images?

PR #836, opened 2022-10-24
Fix session ssh command

chromato99:
resolves #763 
Hello. I upload a draft PR what I worked on by participating in the Hectorfest Seoul event.
Based on the work done so far, an unauthorized error occurs as follows. 

Screenshot from 2022-10-24 22-03-35
Therefore, it would be good if others continue to work on it or give feedback.

PR #1101, opened 2023-02-24
fix: `ssh` not working bug when `backend.ai` executable doesn't exist.

jopemachine:
## Before

```
$ ./backend.ai ssh mysession
✘ FileNotFoundError: 2
  No such file or directory*** Traceback ***
  File "/Users/jopemachine/Desktop/backend.ai/src/ai/backend/client/cli/session.py", line 1015, in ssh
      with container_ssh_ctx(session_ref, port) as key_path:
    File "/Users/jopemachine/.pyenv/versions/3.11.2/lib/python3.11/contextlib.py", line 137, in __enter__
      return next(self.gen)
             ^^^^^^^^^^^^^^
    File "/Users/jopemachine/Desktop/backend.ai/src/ai/backend/client/cli/ssh.py", line 21, in container_ssh_ctx
      subprocess.run(
    File "/Users/jopemachine/.pyenv/versions/3.11.2/lib/python3.11/subprocess.py", line 548, in run
      with Popen(*popenargs, **kwargs) as process:
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/Users/jopemachine/.pyenv/versions/3.11.2/lib/python3.11/subprocess.py", line 1024, in __init__
      self._execute_child(args, executable, preexec_fn, close_fds,
    File "/Users/jopemachine/.pyenv/versions/3.11.2/lib/python3.11/subprocess.py", line 1901, in _execute_child
      raise child_exception_type(errno_num, err_msg, err_filename)
```

The above error occurs because `backend.ai` alias does not exist in the development environment,

I think maybe replacing it with `./backend.ai` in the dev environment can reduce confusion about it when the `./backend.ai` file exist.

## After

```
$ ./backend.ai ssh mysession
∙ running a temporary sshd proxy at localhost:9922 ...
work@main1[mysession]:~$
```


src/ai/backend/client/cli/ssh.py
```diff
@@ -12,14 +12,20 @@
 from .pretty import print_fail, print_info
 
 
+def find_backendai_executable():
+    if __debug__ and os.path.isfile("./backend.ai"):
+        return "./backend.ai"
+    return "backend.ai"
```
achimnol: I think it would be better to use:
```suggestion
    return [sys.executable, "-m", "ai.backend.cli"]
```

achimnol (changes requested):

src/ai/backend/client/cli/ssh.py
```diff
@@ -12,14 +12,20 @@
...
+
 @contextlib.contextmanager
 def container_ssh_ctx(session_ref: str, port: int) -> Iterator[Path]:
     random_id = secrets.token_hex(16)
     key_filename = "id_container"
     key_path = Path(f"~/.ssh/id_{random_id}").expanduser()
     try:
         subprocess.run(
-            ["backend.ai", "session", "download", session_ref, key_filename],
+            [find_backendai_executable(), "session", "download", session_ref, key_filename],
```
rapsealk:
```suggestion
            [*find_backendai_executable(), "session", "download", session_ref, key_filename],
```
Since `find_backendai_executable()` returns a list!

achimnol:
Since this is a duplicate of #836, but I'd prefer this PR's implementationt.
Let's just credit the first author using git trailer: `Co-authored-by: Beomgi Kim <chromato99@gmail.com>`.

PR #1124, opened 2023-03-02
fix: Remove script snippets relate to `.pants.env`

rapsealk:
Support for `.pants.env` has been stopped after 24e5a1f.
Therefore, this PR removes several scripts relate to `.pants.env`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,7 +100,6 @@ ENV/
 /storage-proxy.toml
 /webserver.conf
 /cuda-mock.toml
-/.pants.env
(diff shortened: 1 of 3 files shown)


achimnol (approved):

.gitignore
```diff
@@ -100,7 +100,6 @@ ENV/
 /storage-proxy.toml
 /webserver.conf
 /cuda-mock.toml
-/.pants.env
```
achimnol: I think it's safer to keep this in gitignore, though.

PR #1126, opened 2023-03-04
feat: add option to enable/disable 2FA (`enable_2FA`)

adrysn:
- feat: add `hide_2FA` option to compose the corresponding Web UI config
- Add news fragment


achimnol (commented):
The code looks fine, but I'd like to rename `hide_2FA` to `enable_2FA`, to reduce the negative indirection.

PR #1136, opened 2023-03-07
feature: add git service token management

studioego:
feature: add git service token management

- [ ] Initializing git_tokens table
- [ ] CRUD feature with git_tokens table
- [ ] source code refactoring - GitHub and GitLab importing feature

### DIFF
--- a/src/ai/backend/manager/models/git_token.py
+++ b/src/ai/backend/manager/models/git_token.py
@@ -1,0 +1,41 @@
+from __future__ import annotations
+
+import graphene
+import sqlalchemy as sa
+from graphene.types.datetime import DateTime as GQLDateTime
+
+from .base import GUID, ForeignKeyIDColumn, metadata
+
+__all__ = [
+    "git_tokens",
+]
+
+git_tokens = sa.Table(
+    "git_tokens",
+    metadata,
+    sa.Column(
+        "user_id",
+        GUID,
+        sa.ForeignKey("users.uuid", onupdate="CASCADE", ondelete="CASCADE"),
+        nullable=False,
+        primary_key=True,
+    ),
+    sa.Column("domain", sa.String(length=200, unique=True, primary_key=True, nullable=False)),
+    sa.Column("token", sa.String(length=200)),
+    sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.func.now()),
+    sa.Column(
+        "modified_at",
+        sa.DateTime(timezone=True),
+        server_default=sa.func.now(),
+        onupdate=sa.func.current_timestamp(),
+    ),
+    ForeignKeyIDColumn("user", "users.uuid", nullable=False),
+)
+
+
+class GitToken(graphene.ObjectType):
+    user_id = graphene.UUID()
+    domain = graphene.String()
+    token = graphene.String()
+    created_at = GQLDateTime()
+    modified_at = GQLDateTime()
--- a/src/ai/backend/manager/models/alembic/versions/6da5c6b22bd5_add_git_tokens_table.py
+++ b/src/ai/backend/manager/models/alembic/versions/6da5c6b22bd5_add_git_tokens_table.py
@@ -1,0 +1,46 @@
+"""add git_tokens table
+
+Revision ID: 6da5c6b22bd5
+Revises: 10c58e701d87
+Create Date: 2023-03-06 17:42:53.712150
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+from ai.backend.manager.models.base import GUID
+
+# revision identifiers, used by Alembic.
+revision = "6da5c6b22bd5"
+down_revision = "10c58e701d87"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "git_tokens",
+        sa.Column("user_id", GUID(), nullable=False, index=True),
+        sa.Column("domain", sa.String(length=200), nullable=False, index=True),
+        sa.Column("token", sa.String(length=200)),
+        sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.func.now()),
+        sa.Column(
+            "modified_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=True,
+        ),
+        sa.ForeignKeyConstraint(
+            ["user_id"],
+            ["users.uuid"],
+            name=op.f("fk_git_tokens_user_id_users"),
+            onupdate="CASCADE",
+            ondelete="CASCADE",
+        ),
+        sa.UniqueConstraint("user_id", "domain", name="uq_git_tokens_user_id_domain"),
+    )
+    # Create a default github and gitlab tokens
+
+
+def downgrade():
+    op.drop_table("git_tokens")


src/ai/backend/manager/models/git_token.py
```diff
@@ -0,0 +1,93 @@
...
+    sa.Column("domain", sa.String(length=200), unique=True, primary_key=True, nullable=False),
+    sa.Column("token", sa.String(length=200)),
+    sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.func.now()),
+    sa.Column(
+        "modified_at",
+        sa.DateTime(timezone=True),
+        server_default=sa.func.now(),
+        onupdate=sa.func.current_timestamp(),
+    ),
+)
```
fregataa, line 50: Since we decided to use ORM table, please define the table in ORM
This is [an example code of our Session table.](https://github.com/lablup/backend.ai/blob/main/src/ai/backend/manager/models/session.py#L528)

src/ai/backend/manager/models/git_token.py
```diff
@@ -0,0 +1,93 @@
...
+class GitToken(graphene.ObjectType):
+    user_id = graphene.UUID()
+    domain = graphene.String()
+    token = graphene.String()
+    created_at = GQLDateTime()
+    modified_at = GQLDateTime()
+
+
+class GitTokenRow(Base):
+    __table__ = git_tokens
```
fregataa, line 93: If we define the table in ORM, we can remove this overriding part.

src/ai/backend/manager/api/userconfig.py
```diff
@@ -241,6 +245,47 @@ async def get_bootstrap_script(request: web.Request) -> web.Response:
...
+        },
+    )
+)
+async def update_git_tokens(request: web.Request, params: Any) -> web.Response:
+    resp = []
+    root_ctx: RootContext = request.app["_root.context"]
+    user_uuid = request["user"]["uuid"]
+
+    # get token list then
+    token_list = params["params"]
```
fregataa, line 272:
```suggestion
            t.Key("tokens"): t.Mapping(t.String, t.String),
        },
    )
)
async def update_git_tokens(request: web.Request, params: Any) -> web.Response:
    resp = []
    root_ctx: RootContext = request.app["_root.context"]
    user_uuid = request["user"]["uuid"]

    # get token dictionary then
    tokens: dict[str, str] = params["tokens"]
```

First, the name of parameter is too vague. Let's change it.
If we successfully change the parameter type, then let's update `git_token.insert_update_git_tokens()`!

src/ai/backend/manager/api/userconfig.py
```diff
@@ -241,6 +245,47 @@ async def get_bootstrap_script(request: web.Request) -> web.Response:
...
+@auth_required
+@server_status_required(READ_ALLOWED)
+async def get_git_tokens(request: web.Request) -> web.Response:
+    root_ctx: RootContext = request.app["_root.context"]
+    user_uuid = request["user"]["uuid"]
+
+    async with root_ctx.db.begin_readonly() as conn:
+        query = sa.select([git_tokens.c.domain, git_tokens.c.token]).where(
+            git_tokens.c.user_id == user_uuid
+        )
```
fregataa, line 251: Is there any use case of getting other user's git tokens?
For example, admins would like to get other user's tokens.

src/ai/backend/manager/api/userconfig.py
```diff
@@ -241,6 +245,47 @@ async def get_bootstrap_script(request: web.Request) -> web.Response:
...
+    t.Dict(
+        {
+            t.Key("params"): t.String,
+        },
+    )
+)
+async def update_git_tokens(request: web.Request, params: Any) -> web.Response:
+    resp = []
+    root_ctx: RootContext = request.app["_root.context"]
+    user_uuid = request["user"]["uuid"]
```
fregataa, line 269: Same here.
Please let me know if there is any use case of getting or updating other user's git token,

src/ai/backend/manager/api/userconfig.py
```diff
@@ -11,6 +13,8 @@
 
 from ..models import (
     MAXIMUM_DOTFILE_SIZE,
+    git_token,
+    git_tokens,
```
fregataa, line 17:
```suggestion
    insert_update_git_tokens,
    git_tokens,
```

We'd like to import classes or functions, not modules.

src/ai/backend/manager/models/git_token.py
```diff
@@ -0,0 +1,93 @@
...
+    "git_tokens",
+    metadata,
+    sa.Column(
+        "user_id",
+        GUID,
+        sa.ForeignKey("users.uuid", onupdate="CASCADE", ondelete="CASCADE"),
+        index=True,
+        nullable=False,
+        primary_key=True,
+    ),
```
fregataa, line 40: It would be nice to have a unique `id` field other than the `user_id` field.

src/ai/backend/manager/models/git_token.py
```diff
@@ -0,0 +1,93 @@
...
+    sa.Column("domain", sa.String(length=200), unique=True, primary_key=True, nullable=False),
```
fregataa, line 41: Is git token unique within each domain?

src/ai/backend/manager/models/git_token.py
```diff
@@ -0,0 +1,93 @@
...
+    sa.Column("token", sa.String(length=200)),
```
fregataa, line 42: Just a question!
How long git token usually is??

src/ai/backend/manager/models/git_token.py
```diff
@@ -0,0 +1,93 @@
...
+    sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.func.now()),
+    sa.Column(
+        "modified_at",
+        sa.DateTime(timezone=True),
+        server_default=sa.func.now(),
```
fregataa, line 47: I think that this field should be nullable and default value should be null if we think of a token row without any modification

src/ai/backend/manager/models/git_token.py
```diff
@@ -0,0 +1,93 @@
...
+        sa.and_(git_tokens.c.user_id == user_uuid), sa.not_(git_tokens.c.domain.in_(domain_list))
+    )
+    await conn.execute(delete_stmt)
+
+    for key, value in token_list.items():
+        data = {
+            "user_id": user_uuid,
+            "domain": key,
+            "token": value,
+            "modified_at": datetime.datetime.now(),
```
fregataa, line 72: I think that we do not need to fill this part because of the field's `onupdate` function.

src/ai/backend/manager/models/git_token.py
```diff
@@ -0,0 +1,93 @@
...
+        onupdate=sa.func.current_timestamp(),
+    ),
+)
+
+
+async def insert_update_git_tokens(
+    conn: SAConnection, user_uuid: uuid.UUID, token_list_str: str
+) -> None:
+    token_list = json.loads(token_list_str)
+    domain_list = []
```
fregataa, line 57: This variable is actually of type `dict`, but it has `list` in its name, which confuses me into thinking it is of type `list`.
let's rename them and add type hint!
```suggestion
    tokens: dict[str, Any] = json.loads(token_list_str)
    domain_list: list[str] = []
```

src/ai/backend/manager/models/git_token.py
fregataa, line 40: Is git token unique for every user?

fregataa (changes requested):
I left some comment!

Plus, it will be good to make a new alembic migration file rather than merge them because we had lots of work to resolve migration sequence between 22.09 and 23.03. I am afraid that this merged file will cause problems.

src/ai/backend/manager/models/git_token.py
```diff
@@ -0,0 +1,93 @@
...
+    # check domain_list
+    for key, value in token_list.items():
```
fregataa, line 59:
```suggestion
    for domain_name, token in token_list.items():
```
Let's use more clear name!

src/ai/backend/manager/models/git_token.py
```diff
@@ -0,0 +1,93 @@
...
+        domain_list.append(key)
+    # delete some non-exist key on domain list
+    delete_stmt = sa.delete(git_tokens).where(
+        sa.and_(git_tokens.c.user_id == user_uuid), sa.not_(git_tokens.c.domain.in_(domain_list))
+    )
+    await conn.execute(delete_stmt)
+
+    for key, value in token_list.items():
```
fregataa, line 67:
```suggestion
    for domain_name, token in token_list.items():
```
Same here

src/ai/backend/manager/models/git_token.py
fregataa, line 41: And `domains` table has `name` field which is a primary key.
How about defining this field as a `ForeignKey`?

studioego:
> I left some comment!
> 
> Plus, it will be good to make a new alembic migration file rather than merge them because we had lots of work to resolve migration sequence between 22.09 and 23.03. I am afraid that this merged file will cause problems.

According to your opinion, I make a new alembic version. 
Because, merged file can easily cause problems.

HyeockJinKim:
It seems like there are no current usage plans, so I will close it.

PR #1141, opened 2023-03-08
fix: `ssh-add` not working due to missing env-var inheritance and wrong process ownership

jopemachine:
Fixes https://github.com/lablup/giftbox/issues/216.

When `$USER_ID` is not 0 (root account), `ssh-agent` must run through the user account (`/home/work`).

This PR accomplish this by using `su-exec`.

## Test

Tested manually.

테스트

### DIFF
--- a/src/ai/backend/runner/entrypoint.sh
+++ b/src/ai/backend/runner/entrypoint.sh
@@ -102,12 +102,12 @@ else
 
   # Start ssh-agent if it is available
   if command -v ssh-agent > /dev/null; then
-    AGENT_ENVIRONMENT_PATH="/home/work/ssh-agent.env"
+    SSH_ENV_PATH="/home/work/ssh-agent.env"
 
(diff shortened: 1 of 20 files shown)


src/ai/backend/runner/entrypoint.sh
```diff
@@ -102,7 +102,13 @@ else
...
   # Start ssh-agent if it is available
   if command -v ssh-agent > /dev/null; then
-    eval "$(ssh-agent -s)"
+    SSH_ENV_PATH="/home/work/ssh-agent.env"
+
+    /opt/kernel/su-exec $USER_ID:$GROUP_ID ssh-agent > $SSH_ENV_PATH
+    chmod +x $SSH_ENV_PATH
+    . $SSH_ENV_PATH
+    rm $SSH_ENV_PATH
+
```
achimnol: How about reducing this to:
```suggestion
    eval "$(/opt/kernel/su-exec $USER_ID:$GROUP_ID ssh-agent)"
```

Using `eval "$(...)"` is orthogonal to using `su-exec`.

achimnol (changes requested):

PR #1142, opened 2023-03-08
feat: account for `totp_activated` field in updating user

adrysn:


kyujin-cho (changes requested):
Please also add default value definition to schema definition of `users` table (src/ai/backend/manager/models/users.py).

PR #1145, opened 2023-03-08
fix: resolve vfolder by id rather than by name when cloning

fregataa:
vfolder clone API has updated the status of cloned vfolder by name not id.
This PR fix the bug and uses `execute_with_retry` function to insert and update vfolder table.


src/ai/backend/manager/api/vfolder.py
```diff
@@ -2345,95 +2347,34 @@ async def clone(request: web.Request, params: Any, row: VFolderRow) -> web.Respo
...
-            result = await conn.execute(insert_query)
-        except sa.exc.DataError:
-            # TODO: pass exception info
-            raise InvalidAPIParameters
-
-    # Start the clone operation as a background task.
-    # NOTE: Tracking the progress from the storage-proxy is not supported yet. (See #1033)
-    async def _clone_bgtask(reporter: ProgressReporter) -> None:
-        async with root_ctx.storage_manager.request(
+    task_id, trg_folder_id = await initiate_vfolder_clone(
```
kyujin-cho: Let's write variable name as `target` instead of `trg` to enhance readability. 
```suggestion
    task_id, target_folder_id = await initiate_vfolder_clone(
```

src/ai/backend/manager/models/vfolder.py
```diff
@@ -150,6 +154,20 @@ class VFolderDeletionInfo(NamedTuple):
     host: str
 
 
+class VFolderCloneInfo(NamedTuple):
+    src_vfolder_id: uuid.UUID
+    src_host: str
+
+    # Target Vfolder infos
+    trg_vfolder_name: str
+    trg_host: str
```
kyujin-cho: Same here. 
```suggestion
    target_vfolder_name: str
    target_host: str
```

src/ai/backend/manager/models/vfolder.py
```diff
@@ -751,8 +769,108 @@ async def filter_host_allowed_permission(
...
+
+    trg_proxy, trg_volume = storage_manager.split_host(vfolder_info.trg_host)
+    src_proxy, src_volume = storage_manager.split_host(vfolder_info.src_host)
+
+    # Generate the ID of the destination vfolder.
+    # TODO: If we refactor to use ORM, the folder ID will be created from the database by inserting
+    #       the actual object (with RETURNING clause).  In that case, we need to temporarily
+    #       mark the object to be "unusable-yet" until the storage proxy craetes the destination
+    #       vfolder.  After done, we need to make another transaction to clear the unusable state.
+    trg_folder_id = uuid.uuid4()
```
kyujin-cho: Also here. 
```suggestion
    target_proxy, target_volume = storage_manager.split_host(vfolder_info.trg_host)
    src_proxy, src_volume = storage_manager.split_host(vfolder_info.src_host)

    # Generate the ID of the destination vfolder.
    # TODO: If we refactor to use ORM, the folder ID will be created from the database by inserting
    #       the actual object (with RETURNING clause).  In that case, we need to temporarily
    #       mark the object to be "unusable-yet" until the storage proxy craetes the destination
    #       vfolder.  After done, we need to make another transaction to clear the unusable state.
    target_folder_id = uuid.uuid4()
```

kyujin-cho (changes requested):

achimnol:
Also please add the news fragment!

kyujin-cho (approved):

PR #1153, opened 2023-03-10
feat: add new API to advertise accelerator plugin's metadata

### LINKED ISSUE #1085: Let accelerator plugins provide richer resource slot metadata
Currently `/config/resource-slots` API only provides the resource slot names and their types (`COUNT`, `BYTES`, `UNIQUE`).

Let's add more information for extensibility of webui/control-panel:

- Human-friendly resource slot name (to be shown in the UI)
- The desired number format (fractions, thousand delimeters, etc.)
- The desired UI resource icon type
- Description

For backward compatibility of existing accelerator plugins, let's fill up these information in the manager for known resource slots and use the information if the (future) plugin provides.

kyujin-cho:
This resolves #1085.


src/ai/backend/manager/api/etcd.py
```diff
@@ -21,13 +24,57 @@
...
+    log.info("ETCD.GET_RESOURCE_METADATA ()")
+    root_ctx: RootContext = request.app["_root.context"]
+    available_slot_metadata: Dict[str, Any] = {}
+
+    async with root_ctx.db.begin_readonly() as conn:
+        query = sa.select([agents.c.compute_plugins]).select_from(agents).where()
+        result = await conn.execute(query)
+        for row in result:
+            for key, value in row["compute_plugins"].items():
+                if _data := value.get("metadata", KNOWN_SLOT_METADATA.get(key)):
```
achimnol, line 91: What happens if there is no metadata in both the value and `KNOWN_SLOT_METADATA`?
Let's add a little more friendly error description.

src/ai/backend/manager/api/etcd.py
```diff
@@ -21,13 +24,57 @@
 log = BraceStyleAdapter(logging.getLogger(__spec__.name))  # type: ignore[name-defined]
 
 
+KNOWN_SLOT_METADATA: Mapping[str, AcceleratorMetadata] = {
```
achimnol, line 27: Let's add TPU and ROCM defaults.

achimnol (changes requested):

src/ai/backend/manager/api/etcd.py
kyujin-cho: I've ignored raising error for slot types without metadata intentionally. Added a brief description about these as a block of comment.

achimnol: Um, `dict.get()`'s default second argument is `None`, not `{}`.

PR #1157, opened 2023-03-10
feat: store accelerator metadata on redis

### LINKED ISSUE #1156: Improve #1153 by making the resource slot metadata query to be O(1)
Currently #1153 scans the whole `agents` table. When there are many agents, this will become very slow.

Let's store/update the resource slot metadata upon heartbeat processing using etcd or redis, so that we can read it at O(1) time.  It would be better to avoid using postgresql because heartbeats may also increase the write load of the database.  (In the future, we also need to migrate the agent status outside the SQL database!)

refs #1085

kyujin-cho:
This resolves #1156. Ref #1153


achimnol:
The number format fields should be updated following #1158.

achimnol (approved):

PR #1160, opened 2023-03-13
feat: make idle checker report leftover time and utilization

fregataa:
idle checkers report their idle remaining time to termination and utilization through Redis.
Graphene object will fetch the report.

return value is a map shaped like below
```
{
  "timeout": 10.0,
  "session_lifetime": 20.0,
  "utilization": {"cpu": 0.5, "mem": 1.0}
}
```

---
Test config
1. save the config below to file and set the config to etcd by running
```
./backend.ai mgr etcd put-json config/idle idle_test.txt
```
```
# idle_test.txt
{ 
 "app-streaming-packet-timeout": "", 
 "checkers": { 
 "timeout": { 
 "threshold": "30"
 }, 
 "utilization": { 
 "initial-grace-period": "30", 
 "resource-thresholds": { 
 "cpu_util": { 
 "average": "10" 
 }, 
 "cuda_mem": { 
 "average": "10" 
 }, 
 "cuda_util": { 
 "average": "10" 
 }, 
 "mem": { 
 "average": "10" 
 } 
 }, 
 "thresholds-check-operator": "or", 
 "time-window": "15" 
 } 
 }, 
 "enabled": "utilization" # if you want to add timeout idle checker, use this value "utilization, timeout"
} 
```

3. set idle timeout like below
```
./backend.ai admin keypair-resource-policy update --idle-timeout 15 default
```


src/ai/backend/manager/idle.py
```diff
@@ -63,6 +69,28 @@
...
+    """
+
+
+@dataclass(kw_only=True)
+class IdleCheckReport:
+    timeout: Optional[float] = None
+    session_lifetime: Optional[float] = None
+    avg_utilization: Optional[Mapping[str, float]] = None
+
+    def to_json(self) -> Mapping[str, Any]:
```
kyujin-cho: This function is returning `Dict`, not "JSON".
```suggestion
    def to_dict(self) -> Mapping[str, Any]:
```

src/ai/backend/manager/idle.py
```diff
@@ -344,66 +419,119 @@ async def _execution_exited_cb(
...
+            # setting idle_timeout:
+            # - zero/inf means "infinite"
+            # - negative means "undefined"
+            if policy["idle_timeout"] > 0:
+                idle_timeout = float(policy["idle_timeout"])
+            if (idle_timeout <= 0) or (math.isinf(idle_timeout) and idle_timeout > 0):
+                return None
+            idle_time: float = t - last_access
+            remaining: float = idle_timeout - idle_time
+            return remaining
```
kyujin-cho: Is there any specific reason to wrap this logic as a inner function?

src/ai/backend/manager/idle.py
```diff
@@ -344,66 +419,119 @@ async def _execution_exited_cb(
...
+        if result is not None:
+            await redis_helper.execute(
+                redis_obj,
+                lambda r: r.set(
+                    self.get_report_key(session_id),
+                    msgpack.packb(result),
+                    ex=int(DEFAULT_CHECK_INTERVAL) * 10,
+                ),
+            )
+        return True if result is not None and result > 0 else False
```
kyujin-cho: We can simplify this as:
```suggestion
        return result is not None and result > 0
```

src/ai/backend/manager/idle.py
```diff
@@ -231,14 +294,25 @@ async def update_app_streaming_status(
...
     @abstractmethod
-    async def check_session(self, session: Row, dbconn: SAConnection, policy: Row) -> bool:
-        """
-        Return True if the session should be kept alive or
-        return False if the session should be terminated.
-        """
+    async def check_kernel(
+        self, kernel: Row, dbconn: SAConnection, policy: Row, redis_obj: RedisConnectionInfo
+    ) -> bool:
         return True
```
kyujin-cho: Please add a description so that future workers can understand what this interface does.

src/ai/backend/manager/idle.py
```diff
@@ -231,14 +294,25 @@ async def update_app_streaming_status(
...
 
+    @abstractmethod
+    async def assign_check_result(
+        self,
+        redis_obj: RedisConnectionInfo,
+        report: IdleCheckReport,
+        session_id: SessionId,
+    ) -> None:
+        pass
+
```
kyujin-cho: Documentation here also.

src/ai/backend/manager/idle.py
```diff
@@ -344,66 +419,119 @@ async def _execution_exited_cb(
...
+        if result is not None:
+            await redis_helper.execute(
+                redis_obj,
+                lambda r: r.set(
+                    self.get_report_key(kernel["session_id"]),
+                    msgpack.packb(result),
+                    ex=int(DEFAULT_CHECK_INTERVAL) * 10,
+                ),
+            )
+        return True if result is not None and result > 0 else False
```
kyujin-cho:
```suggestion
        return result is not None and result > 0
```

src/ai/backend/manager/idle.py
```diff
@@ -453,128 +591,149 @@ async def populate_config(self, raw_config: Mapping[str, Any]) -> None:
...
-        interval = IdleCheckerHost.check_interval
-        window_size = int(self.time_window.total_seconds() / interval)
-        occupied_slots = session["occupied_slots"]
-        unavailable_resources: Set[str] = set()
+    async def check_kernel(
+        self, kernel: Row, dbconn: SAConnection, policy: Row, redis_obj: RedisConnectionInfo
+    ) -> bool:
+        session_id = kernel["session_id"]
+
+        async def _check_kernel() -> tuple[bool, Optional[dict[str, float]]]:
```
kyujin-cho: Also here.

src/ai/backend/manager/idle.py
```diff
@@ -453,128 +591,149 @@ async def populate_config(self, raw_config: Mapping[str, Any]) -> None:
...
-        if not check_result:
+            if len(sufficiently_utilized) < 1:
+                check_result = True
+            elif self.thresholds_check_operator == ThresholdOperator.OR:
+                check_result = all(sufficiently_utilized.values())
+            else:  # "and" operation is the default
+                check_result = any(sufficiently_utilized.values())
+            return check_result, avg_utils
+
+        is_alive, avg_utils = await _check_kernel()
```
kyujin-cho: `should_alive` rather than `is_alive` makes more sense.

kyujin-cho (changes requested):

src/ai/backend/manager/idle.py
```diff
@@ -368,42 +457,92 @@ async def check_session(self, session: Row, dbconn: SAConnection, policy: Row) -
             return True
         last_access = float(raw_last_access)
         # serves as the default fallback if keypair resource policy's idle_timeout is "undefined"
-        idle_timeout = self.idle_timeout.total_seconds()
+        idle_timeout: float = self.idle_timeout.total_seconds()
         # setting idle_timeout:
         # - zero/inf means "infinite"
         # - negative means "undefined"
-        if policy["idle_timeout"] >= 0:
+        if policy["idle_timeout"] > 0:
```
kyujin-cho: Is there a reason that comparator has changed to >?

src/ai/backend/manager/idle.py
```diff
@@ -231,14 +294,32 @@ async def update_app_streaming_status(
...
 
+    @classmethod
+    def get_report_key(cls, session_id: SessionId) -> str:
+        return f"session.{session_id}.{cls.name}.report"
+
     @abstractmethod
-    async def check_session(self, session: Row, dbconn: SAConnection, policy: Row) -> bool:
+    async def report_idle_check(
+        self, kernel: Row, dbconn: SAConnection, policy: Row, redis_obj: RedisConnectionInfo
+    ) -> bool:
```
kyujin-cho: Main characteristic of this function is to determine if the kernel violates current idle checker policy. Please update the interface name so that users can understand express the characteristic at first glance.

src/ai/backend/manager/idle.py
```diff
@@ -368,42 +457,92 @@ async def check_session(self, session: Row, dbconn: SAConnection, policy: Row) -
...
+
+        result: Optional[float] = None
+        if (max_session_lifetime := policy["max_session_lifetime"]) > 0:
             # TODO: once per-status time tracking is implemented, let's change created_at
             #       to the timestamp when the session entered PREPARING status.
-            if now - session["created_at"] >= timedelta(seconds=policy["max_session_lifetime"]):
-                return False
-        return True
+            idle_timeout = timedelta(seconds=max_session_lifetime)
+            now = await dbconn.scalar(sa.select(sa.func.now()))
```
kyujin-cho, line 502: How about using redis' `TIME` command? Most of the components in Backend.AI which requires precise time sync uses that one. Also, in general, using `TIME` command is way faster than querying result of `NOW()` function in postgresql.

kyujin-cho (changes requested):

kyujin-cho (approved):

PR #1163, opened 2023-03-14
ci: add action to deploy backend.ai-client docker image

Yaminyam:
To use backend.ai-client regardless of the environment, build it as a docker image and share it to docker hub.
To use this action, DOCKERHUB_USERNAME and DOCKERHUB_PASSWORD must be added as secret keys.


client.Dockerfile
kyujin-cho, line 1: Let's update file name concisely so that newcomers can understand what this Dockerfile do.

client.Dockerfile
```diff
@@ -0,0 +1,10 @@
+ARG PYTHON_VERSION
+FROM python:${PYTHON_VERSION} AS builder
+ARG PKGVER
+RUN pip wheel --wheel-dir=/wheels --no-cache-dir backend.ai-client==${PKGVER}
+
+FROM python:${PYTHON_VERSION}
+COPY --from=builder /wheels /wheels
+RUN mkdir -p /root/.ssh
```
kyujin-cho: Can you describe about what this line is for?

client.Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+RUN apt-get update && apt-get install -y vim
```
kyujin-cho: It is recommended to clear apt caches after installing apt packages. Ref: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

src/ai/backend/runner/jail.ubuntu16.04.x86_64.bin
kyujin-cho, line 1: What is this for?

kyujin-cho (changes requested):

client.Dockerfile
Yaminyam: This is to solve the problem that the ssh key is not created because the folder does not exist when using the ssh function of backend.ai.

src/ai/backend/runner/jail.ubuntu16.04.x86_64.bin
Yaminyam: This is my mistake.

Yaminyam: How can I undo binary file changes?

client.Dockerfile
Yaminyam: It is a dockerfile that creates a backend.ai-client image. Which one do you think is more suitable?

redis-sentinel.Dockerfile
kyujin-cho, line 1: This file is already being referenced as `redis-sentinel.dockerfile` on other files.

client.Dockerfile
kyujin-cho: `backend.ai-client.Dockerfile` then. Straightforward!

kyujin-cho (changes requested):

redis-sentinel.Dockerfile
Yaminyam: The dockerfile convention has been decided to be xxx.Dockerfile.
We will change other code referencing that file.

kyujin-cho (approved):

PR #1168, opened 2023-03-15
feat: serve metrics related to inference API

kyujin-cho:
This PR adds new `inference_metrics` column to `compute_session` GQL query, which tells about inference endpoints served on inference sessions. Currently supports:
- `requests` : total number of requests called at the API
- `last_response_ms`: time taken to respond to the last API call
Metrics are collected by wsproxy components and are stored at redis.


achimnol (approved):
LGTM!

PR #1169, opened 2023-03-16
ci: install local wheels when building conda-pack

Yaminyam:
After uploading to pypi, conda-pack is being created, so it was judged that there would be no problem in the order, but it was judged that it took time to find the package in pypi.
So, modified to install the newly built wheels files by passing them as artifacts.

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -326,6 +326,7 @@ jobs:
         name: wheels
         path: dist
     - name: Create conda environment
+      # FIXME: Let's think about resolving dependency of backend.ai-client package programmatically, instead of hardcoding it.
       run: |
         pip install conda-pack
         conda create -n backend.ai-client python=${{ env.PROJECT_PYTHON_VERSION }}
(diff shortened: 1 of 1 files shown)


.github/workflows/default.yml
```diff
@@ -305,16 +312,25 @@ jobs:
...
+      with:
+        name: wheels
+        path: dist
     - name: Create conda environment
       run: |
         pip install conda-pack
         conda create -n backend.ai-client python=${{ env.PROJECT_PYTHON_VERSION }}
         conda activate backend.ai-client
-        pip install backend.ai-client==${{ github.ref_name }}
+        pip install dist/backend.ai_client-${{ env.PKGVER }}-py3-none-any.whl dist/backend.ai_cli-${{ env.PKGVER }}-py3-none-any.whl dist/backend.ai_common-${{ env.PKGVER }}-py3-none-any.whl dist/backend.ai_plugin-${{ env.PKGVER }}-py3-none-any.whl
```
kyujin-cho, line 334: Let's think about resolving dependency of backend.ai-client package programmatically, instead of hardcoding it.

kyujin-cho (changes requested):

kyujin-cho (approved):

PR #1170, opened 2023-03-16
feat: Provide status information about usage, capacity of storage volume to `list-hosts` API

### LINKED ISSUE #1164: Let's provide an status information for current allocatable capacity of storage volume
### Main idea

For now, when we try to create a virtual folder inside the storage volume with no space left, It throws an error; which degrades the overall user experience.
Let's provide at least current usage as an abstract tag (such as percentage, not the whole capacity) so that the user can know the current status of storage volume.

[teams discussion thread](https://teams.microsoft.com/l/message/19:3c0acc0825024daeb2211f55c4e7f4aa@thread.skype/1678937127668?tenantId=13c6a44d-9b52-4b9e-aa34-0513ee7131f2&groupId=74ae2c4d-ec4d-4fdf-b2c2-f5041d1e8631&parentMessageId=1678932275317&teamName=devops&channelName=Backend.AI%20Talks&createdTime=1678937127668&allowXTenantAccess=false)

### Alternative ideas

_No response_

### Anything else?

_No response_

jopemachine:
Fixes #1164.

* Add `exposed_volume_info` to StorageSessionManager
* Add usage and capacity information to `list_hosts` API

## Test

Tested on manual through the below command.

```
$ ./backend.ai vfolder list-hosts
```

Since above command only prints `resp["default"]`, `resp["allowed"]`, I tested it by printing [`resp` itself](https://github.com/lablup/backend.ai/blob/main/src/ai/backend/client/cli/vfolder.py#L40).


src/ai/backend/storage/api/manager.py
```diff
@@ -338,11 +338,14 @@ async def get_vfolder_fs_usage(request: web.Request) -> web.Response:
...
         async with ctx.get_volume(params["volume"]) as volume:
             fs_usage = await volume.get_fs_usage()
+            hide_vfolder_usage = ctx.local_config["storage-proxy"]["hide_vfolder_usage"]
+
             return web.json_response(
                 {
-                    "capacity_bytes": fs_usage.capacity_bytes,
-                    "used_bytes": fs_usage.used_bytes,
-                },
+                    "percentage": fs_usage.used_bytes / fs_usage.capacity_bytes,
```
yomybaby: percentage means "part / total * 100". 
`* 100` is missing.

yomybaby (changes requested):

src/ai/backend/storage/config.py
```diff
@@ -30,6 +30,7 @@
                     t.Key("max-upload-size", default="100g"): tx.BinarySize,
                     t.Key("secret"): t.String,  # used to generate JWT tokens
                     t.Key("session-expire"): tx.TimeDuration,
+                    t.Key("hide_vfolder_usage", default=True): t.ToBool,
```
yomybaby: 1. the name of other options use `-` instead of `_`.

2. To meet the requirement, depending on the settings it will determine which information to expose: 
- return `percentage` only
- return `percentage`, `usage` and `total`.
- return nothing.

@adrysn , right?

yomybaby (changes requested):

src/ai/backend/manager/config.py
```diff
@@ -319,6 +319,11 @@
     "watcher": {
         "token": None,
     },
+    "expose-host-volume": {
+        "percentage": True,
+        "used-bytes": False,
+        "capacity-bytes": False,
+    },
```
adrysn: I think this option should be placed under `volumes` key.
https://github.com/lablup/backend.ai/blob/c8cf41264f4e4c0edb119c2efcecd7d3a9455dc3/src/ai/backend/manager/config.py#L124-L141

configs/manager/sample.etcd.config.json
```diff
@@ -46,5 +46,10 @@
         "num_retries_to_skip": 3
       }
     }
+  },
+  "expose-host-volume": {
+    "percentage": true,
+    "used-bytes": false,
+    "capacity-bytes": false
   }
```
adrysn: How about the following option structure?
```
"allowed_volume_info": "percentage"  # allow percentage only for users
"allowed_volume_info": "percentage,used-bytes,capacity-bytes"  # allow all stats for users
```

adrysn (changes requested):

configs/manager/sample.etcd.volumes.json
```diff
@@ -11,5 +11,6 @@
       "secret": "some-secret-shared-with-storage-proxy",
       "ssl_verify": "false"
     }
-  }
+  },
+  "allowed_volume_info": "percentage"
```
kyujin-cho: The name `allowed_volume_info` seems a bit ambiguous to me. Who are we allowing to access the volume info? Who is exposing the information?

src/ai/backend/manager/api/vfolder.py
```diff
@@ -612,6 +618,60 @@ async def delete_by_id(request: web.Request, params: Any) -> web.Response:
...
+    storage_manager: StorageSessionManager,
+    redis_connection: RedisConnectionInfo,
+    proxy_name: str,
+    volume_name: str,
+):
+    volume_usage = {}
+
+    show_percentage = "percentage" in storage_manager._allowed_volume_info
+    show_used = "used_bytes" in storage_manager._allowed_volume_info
+    show_total = "capacity_bytes" in storage_manager._allowed_volume_info
```
kyujin-cho: Let's wrap these string literals under a single Enum.

src/ai/backend/manager/api/vfolder.py
```diff
@@ -612,6 +618,60 @@ async def delete_by_id(request: web.Request, params: Any) -> web.Response:
     return web.Response(status=204)
 
 
+async def fetch_allowed_volume_usage(
+    storage_manager: StorageSessionManager,
+    redis_connection: RedisConnectionInfo,
+    proxy_name: str,
+    volume_name: str,
+):
```
kyujin-cho: Please declare return type.

kyujin-cho (changes requested):

jopemachine:
Also tested manually on WebUI through [this branch](https://github.com/lablup/backend.ai-webui/pull/1640).

스크린샷 2023-03-21 오후 6 44 30

kyujin-cho (approved):

PR #1174, opened 2023-03-18
feat: make agent list query accept multiple statuses

adrysn:


src/ai/backend/manager/models/agent.py
```diff
@@ -264,11 +264,15 @@ async def load_count(
         raw_status: str = None,
```
kyujin-cho: If this function is capable of accepting `raw_status` value with `AgentStatus` type then we also need to update the type definition to reflect it.
```suggestion
        raw_status: Optional[str | AgentStatus] = None,
```

kyujin-cho (changes requested):

kyujin-cho (approved):
LGTM.

PR #1175, opened 2023-03-19
fix: Rewrite alembic migration history to upgrade from 22.09

adrysn:
Migration from 22.09 (22.09.17 for my testing) to 23.03 raises the exception below:
```
    return await executor (protocol)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
asyncpa. exceptions .UndefinedTableError: relation "sessions" does not exist

The above exception was the direct cause of the following exception:
```

This is partly due to the divergence of the alembic migration path between 22.09 (light green boxes) and 23.03 (white boxes):
image

This PR rewrites the alembic history for smoother migration from 22.09 like the below image. The red boxes are changed ones.
image

PR #1177, opened 2023-03-20
fix: Migrate session dependencies table

fregataa:
#576 did not migrate data of session dependencies table.
Since single kernel sessions had the same id with kernels and sessions, the session_id foreign key should be updated.


fregataa:
close and update this to #1175

PR #1179, opened 2023-03-20
feat: Gather per-container network statistics

sanxiyn:


src/ai/backend/agent/docker/intrinsic.py
```diff
@@ -597,9 +598,29 @@ async def sysfs_impl(container_id):
                     e,
                 )
                 return None
+            async with closing_async(Docker()) as docker:
+                container = DockerContainer(docker, id=container_id)
+                data = await container.show()
+                sandbox_key = data["NetworkSettings"]["SandboxKey"]
```
achimnol, line 625: How about storing this sandbox key in `kernel_registry`?

src/ai/backend/agent/docker/intrinsic.py
```diff
@@ -597,9 +598,29 @@ async def sysfs_impl(container_id):
...
+            net_rx_bytes = 0
+            net_tx_bytes = 0
+            with nsenter(sandbox_key):
+                nstat = psutil.net_io_counters(pernic=True)
```
achimnol: Would this affect existing socket connections though it should take a very short amount of time?

sanxiyn: Oops, you are right. This should be executed in the new thread.

src/ai/backend/agent/docker/intrinsic.py
sanxiyn: Sure.

sanxiyn: I created a new issue for this: #1196.

src/ai/backend/agent/docker/intrinsic.py
sanxiyn: Rather, this should be executed in the new process... See my new commit and comment there for details.

achimnol (approved):

PR #1182, opened 2023-03-22
feat: add API for changing vfolder ownership

### LINKED ISSUE #408: vfolder migration between users and projects
In the current architecture, we could just change the owner field of the vfolders table as we use a flat UUID namespace for all vfolders.

Still, we need to take extra care when vfolders should be migrated to different storage proxy volumes as it requires explicit data copy and quota checks.

When we introduce a new planned vfolder architecture to better support per-user/per-group quota, we also need to let the storage-proxy to migrate vfolders between different quota namespaces and _depending on the storage vendors_ this may or may not require explicit data copy.

Let's abstract the vfolder migration operation in the storage-proxy as a long-running background process and make the manager API to proxy the progress with necessary database updates.

Q. Should we introduce a "locked" status of vfolders which prevent mounting of vfolders during such long-running operation?  When locking a vfolder, we also need to check existing mounts, but currently there is no database index for vfolder mounts.
Q. During the migration, the storage space will be taken twice of the vfolder size if we copy the full content first and then delete the original vfolder.  Should we incrementally migrate individual files to limit additional space usage?  If so, how to handle failure in the middle?

lizable:
This PR partially resolves #408.

FYI: 
- Support changing the `owner` of vfolder, not `creator`.
- Support only users who can access the same storage host that vfolder is located. (It will be expanded in phase 2)
- Support changing `user` type of vfolder, not `group(project)` type. (This also will be implemented in phase 2)

### DIFF
--- a/src/ai/backend/client/func/vfolder.py
+++ b/src/ai/backend/client/func/vfolder.py
@@ -625,7 +625,7 @@ class VFolder(BaseFunction):
     @api_function
     @classmethod
     async def change_vfolder_ownership(cls, vfolder: str, user: str):
-        rqst = Request("POST", "/folders/_/change_ownership")
+        rqst = Request("POST", "/folders/_/change-ownership")
         rqst.set_json(
             {
                 "vfolder": vfolder,
--- a/src/ai/backend/client/func/vfolder.py
+++ b/src/ai/backend/client/func/vfolder.py
@@ -633,4 +633,4 @@ class VFolder(BaseFunction):
             }
         )
         async with rqst.fetch() as resp:
-            return await resp.json()
+            return await resp.text()
--- a/src/ai/backend/client/func/vfolder.py
+++ b/src/ai/backend/client/func/vfolder.py
@@ -633,4 +633,4 @@ class VFolder(BaseFunction):
             }
         )
         async with rqst.fetch() as resp:
-            return await resp.text()
+            return await resp.json()
--- a/src/ai/backend/client/func/vfolder.py
+++ b/src/ai/backend/client/func/vfolder.py
@@ -621,3 +621,16 @@ class VFolder(BaseFunction):
         )
         async with rqst.fetch() as resp:
             return await resp.json()
+
+    @api_function
+    @classmethod
+    async def change_vfolder_ownership(cls, vfolder: str, user: str):
+        rqst = Request("POST", "/folders/_/change_ownership")
+        rqst.set_json(
+            {
+                "vfolder": vfolder,
+                "user": user,
+            }
+        )
+        async with rqst.fetch() as resp:
+            return await resp.json()
--- a/src/ai/backend/manager/api/vfolder.py
+++ b/src/ai/backend/manager/api/vfolder.py
@@ -3008,7 +3008,7 @@ async def change_vfolder_ownership(request: web.Request, params: Any) -> web.Res
 
     # TODO: Implement Change ownership using DB transaction
     log.info(
-        "VFOLDER.CHANGE_VFOLDER_OWNERSHIP(emial:{}, ak:{}, vfid:{}, uid:{})",
+        "VFOLDER.CHANGE_VFOLDER_OWNERSHIP(email:{}, ak:{}, vfid:{}, uid:{})",
         request["user"]["email"],
         access_key,
         vfolder_id,
--- a/src/ai/backend/manager/api/vfolder.py
+++ b/src/ai/backend/manager/api/vfolder.py
@@ -48,6 +48,7 @@ from ..models import (
     AgentStatus,
     KernelStatus,
     UserRole,
+    UserStatus,
     VFolderAccessStatus,
     VFolderCloneInfo,
     VFolderDeletionInfo,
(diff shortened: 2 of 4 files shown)


src/ai/backend/manager/api/vfolder.py
```diff
@@ -3055,6 +3127,7 @@ def create_app(default_cors_options):
     cors.add(add_route("GET", r"/_/mounts", list_mounts))
     cors.add(add_route("POST", r"/_/mounts", mount_host))
     cors.add(add_route("DELETE", r"/_/mounts", umount_host))
+    cors.add(add_route("POST", r"/_/change_ownership", change_vfolder_ownership))
```
kyujin-cho: Using underbar to represent whitespace is now considered as legacy naming convention. Please use hyphen.
```suggestion
    cors.add(add_route("POST", r"/_/change-ownership", change_vfolder_ownership))
```

src/ai/backend/client/func/vfolder.py
```diff
@@ -621,3 +621,16 @@ async def update_shared_vfolder(cls, vfolder: str, user: str, perm: str = None):
         )
         async with rqst.fetch() as resp:
             return await resp.json()
+
+    @api_function
+    @classmethod
+    async def change_vfolder_ownership(cls, vfolder: str, user: str):
+        rqst = Request("POST", "/folders/_/change_ownership")
```
kyujin-cho: Also here. 
```suggestion
        rqst = Request("POST", "/folders/_/change-ownership")
```

src/ai/backend/manager/api/vfolder.py
```diff
@@ -2985,6 +2986,77 @@ async def storage_task_exception_handler(
...
+        #       for now we only support migration btw user folder only
+        #
+        query = (
+            sa.update(vfolders)
+            .values(user=user_uuid)
+            .where(vfolders.c.id == vfolder_id)
+            .where(vfolders.c.ownership_type == VFolderOwnershipType.USER)
+        )
+        await conn.execute(query)
+    return web.json_response({}, status=200)
```
fregataa, line 3073:
```suggestion
    return web.Response(status=200)
```
we can just return status with `web.Response`

kyujin-cho (changes requested):

src/ai/backend/manager/api/vfolder.py
```diff
@@ -2985,6 +2986,77 @@ async def storage_task_exception_handler(
...
+        user_uuid,
+    )
+    allowed_hosts_by_user = VFolderHostPermissionMap()
+    async with root_ctx.db.begin_readonly() as conn:
+        j = sa.join(users, keypairs, users.c.uuid == keypairs.c.user)
+        query = (
+            sa.select([users.c.email, users.c.domain_name, keypairs.c.resource_policy])
+            .select_from(j)
+            .where(users.c.uuid == user_uuid)
+            .where(users.c.status == users.UserStatus.ACTIVE)
```
```suggestion
            .where(users.c.uuid == user_uuid & users.c.status == users.UserStatus.ACTIVE)
```

src/ai/backend/manager/api/vfolder.py
```diff
@@ -2985,6 +2986,77 @@ async def storage_task_exception_handler(
...
+    For now, we only provide changing the ownership of user-folder
+    """
+    access_key = request["keypair"]["access_key"]
+    vfolder_id = params["vfolder"]
+    user_uuid = params["user"]
+    root_ctx: RootContext = request.app["_root.context"]
+
+    # TODO: Implement Change ownership using DB transaction
+    log.info(
+        "VFOLDER.CHANGE_VFOLDER_OWNERSHIP(emial:{}, ak:{}, vfid:{}, uid:{})",
```
```suggestion
        "VFOLDER.CHANGE_VFOLDER_OWNERSHIP(email:{}, ak:{}, vfid:{}, uid:{})",
```
mistypo on `email`

src/ai/backend/manager/api/vfolder.py
```diff
@@ -2985,6 +2986,77 @@ async def storage_task_exception_handler(
...
+        # TODO: we need to implement migration from project to other project
+        #       for now we only support migration btw user folder only
+        #
+        query = (
+            sa.update(vfolders)
+            .values(user=user_uuid)
+            .where(vfolders.c.id == vfolder_id)
+            .where(vfolders.c.ownership_type == VFolderOwnershipType.USER)
+        )
+        await conn.execute(query)
```
```suggestion
    async def _update() -> None:
        async with root_ctx.db.begin() as conn:
        # TODO: we need to implement migration from project to other project
        #       for now we only support migration btw user folder only
        #
            query = (
                sa.update(vfolders)
                .values(user=user_uuid)
                .where(vfolders.c.id == vfolder_id & vfolders.c.ownership_type == VFolderOwnershipType.USER)
            )
            await conn.execute(query)
    await execute_with_retry(_update)
```
Any insert or update queries should be executed in `execute_with_retry` function.

kyujin-cho (approved):

fregataa (changes requested):
Thank you for your work on it. I left some comments

src/ai/backend/manager/api/vfolder.py
lizable: Bullseye! 👀

src/ai/backend/manager/api/vfolder.py
lizable: Noted!

src/ai/backend/manager/api/vfolder.py
lizable: I agree. I'll apply it right away.

src/ai/backend/manager/api/vfolder.py
lizable: Noted!

kyujin-cho (approved):

PR #1185, opened 2023-03-22
feat: update utilization based idle checker policy

fregataa:
follow up #1160 

Update utilization idle checker's policy

## Utilization expire time policy
- If the time window is infinite or zero, expire time is always `None`
- if the kernel's running time is shorter than initial grace period, expire time is  (initial grace period + time window size - (now - kernel creation time))
```
if now - kernel["created_at"] <= self.initial_grace_period + window_size)
-> expire_time = self.initial_grace_period + window_size - (now - kernel["created_at"])
```
- else, expire time is `None`.

Plus, the `utilization_extra` key of response has data like below
```
{
  <RESOURCE_NAME_1> : [average_utilization, threshold],
  <RESOURCE_NAME_2> : [average_utilization, threshold],
  ...
}
```

### DIFF
--- a/src/ai/backend/manager/idle.py
+++ b/src/ai/backend/manager/idle.py
@@ -78,16 +78,10 @@ class IdleCheckerError(TaskGroupError):
     """
 
 
-RESOURCE_UNIT_MAP = {
-    "mem": lambda v: BinarySize(int(v)),
-    "cuda_mem": lambda v: BinarySize(int(v)),
-}
-
-
 def parse_unit(resource_name: str, value: float | int) -> float | int:
-    if (func := RESOURCE_UNIT_MAP.get(resource_name)) is not None:
-        return func(value)
-    return value
+    if resource_name.find("mem") == -1:
+        return value
+    return BinarySize(int(value))
 
 
 class UtilizationExtraInfo(NamedTuple):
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/idle.py
```diff
@@ -706,22 +751,50 @@ def _avg(util_list: list[float]) -> float:
...
+
+        # util_avg_thresholds's key is resource name
+        # and value is `(avg_util, threshold)`
+        util_avg_thresholds: dict[str, UtilizationExtraInfo] = {
+            k: UtilizationExtraInfo(float(avg_utils[k]), float(threshold))
             for k, threshold in self.resource_thresholds.items()
             if (threshold is not None) and (k not in unavailable_resources)
         }
+        for k in util_avg_thresholds:
+            if k in ("mem", "cuda_mem"):
```
kyujin-cho: Is there a way to generalize detection of memory resources, so that we can support all other accelerators (TPU, ROCm, ...) also on idle checker?

kyujin-cho (changes requested):

src/ai/backend/manager/idle.py
fregataa: I still have no idea.. I have been thinking of applying such unit type.

kyujin-cho (approved):

PR #1188, opened 2023-03-24
refactor: change user_id input to user email

lizable:
## Description
Related PR: #1186.
For providing a better user interface on Backend.AI CLI, I slightly changed API argument and corresponding functions.
(user id -> user email)

### Screenshot(s)
Screenshot 2023-03-24 at 3 20 33 PM

### DIFF
--- a/src/ai/backend/manager/api/vfolder.py
+++ b/src/ai/backend/manager/api/vfolder.py
@@ -3008,12 +3008,6 @@ async def change_vfolder_ownership(request: web.Request, params: Any) -> web.Res
     user_email = params["user_email"]
     root_ctx: RootContext = request.app["_root.context"]
 
-    # TODO: Implement Change ownership using DB transaction
-    log.info(
-        "VFOLDER.CHANGE_VFOLDER_OWNERSHIP(new owner email:{}, vfid:{})",
-        user_email,
-        vfolder_id,
-    )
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/api/vfolder.py
```diff
@@ -3004,26 +3004,23 @@ async def change_vfolder_ownership(request: web.Request, params: Any) -> web.Res
...
+    user_email = params["user_email"]
     root_ctx: RootContext = request.app["_root.context"]
 
     # TODO: Implement Change ownership using DB transaction
     log.info(
-        "VFOLDER.CHANGE_VFOLDER_OWNERSHIP(email:{}, ak:{}, vfid:{}, uid:{})",
-        request["user"]["email"],
-        access_key,
+        "VFOLDER.CHANGE_VFOLDER_OWNERSHIP(new owner email:{}, vfid:{})",
+        user_email,
```
fregataa: Let's remain the logging of requester's info! we can add new logging of "new owner"

lizable: ~~Oh, It doesn't reflect the original "owner", It's the user(actually, super admin) who requested it.~~
I agree. I just changed log msg according to your suggestion. Thx!

fregataa (approved):

PR #1197, opened 2023-03-26
docs: installation from multipass

kimmokalover:


achimnol (changes requested):
Thanks for the contribution, but please first sign the CLA.
I think it would be better to place multipass-specific instruction inside `dev/development-setup.rst` as a special section, rather than a separate document at the same hierarchical level of other documents describing each target deployment environments.

achimnol:
This PR is marked as stale.
Please feel free to reopen once you are ready to sign the CLA and follow up the review process.

PR #1198, opened 2023-03-26
fix: delete utilization expire time after initial grace period

fregataa:
follow up: #1185 

Delete utilization expire time only after initial grace period

### DIFF
--- a/src/ai/backend/manager/idle.py
+++ b/src/ai/backend/manager/idle.py
@@ -662,10 +662,14 @@ class UtilizationIdleChecker(BaseIdleChecker):
         )
         return msgpack.unpackb(data) if data is not None else None
 
-    async def delete_expire_time_report(self, session_id: SessionId) -> None:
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/idle.py
```diff
@@ -888,7 +886,11 @@ async def get_checker_result(
     ) -> Optional[float]:
         key = self.get_report_key(session_id)
         data = await redis_helper.execute(redis_obj, lambda r: r.get(key))
-        return msgpack.unpackb(data) if data is not None else None
+        if data is not None:
+            if (unpacked := msgpack.unpackb(data)) > 0:
+                return unpacked
+            return None
+        return None
```
kyujin-cho: We can simplify this like:
```suggestion
        if data is not None and (unpacked := msgpack.unpackb(data)) > 0:
            return unpacked
        return None
```

kyujin-cho (changes requested):

PR #1205, opened 2023-04-01
fix: calculate idle checker's expiration time correctly

fregataa:
Calculate idle checker's remaining time to expire correctly.
Previously, `user_initial_grace_period` (=new user grace period) is added to all of checker's remaining time to expire, which is wrong.
Correct flow of calculating remaining time is below.
1. if `user_initial_grace_period` is set
 - the baseline of idle time is one of `user_initial_grace_period` and `kernel_created_at`(=kernel creation time), whichever is later.
 baseline: datetime type.
 now: datetime type. the current time
 timeout: timedelta type. the period of timeout. ex) 100 seconds
 - if baseline is later than now, remaining time is ((baseline - now) + timeout)
 - else, remaining time is (timeout - (now - baseline))
2. if `user_initial_grace_period` is not set, then just calculate the remaining time as we have done.

---

### test config
```
{
    "app-streaming-packet-timeout": "",
    "checkers": {
        "network_timeout": {
            "threshold": "500"
        },
        "timeout": {
            "threshold": "500"
        },
        "user_grace_period": {
            "user_initial_grace_period": "500"
        },
        "utilization": {
            "initial-grace-period": "60",
            "resource-thresholds": {
                "cpu_util": {
                    "average": "10"
                },
                "cuda_mem": {
                    "average": "10"
                },
                "cuda_util": {
                    "average": "10"
                },
                "mem": {
                    "average": "10"
                }
            },
            "thresholds-check-operator": "or",
            "time-window": "60"
        }
    },
    "enabled": "network_timeout,utilization"
}
```

### DIFF
--- a/src/ai/backend/manager/idle.py
+++ b/src/ai/backend/manager/idle.py
@@ -244,10 +244,8 @@ class IdleCheckerHost:
             )
             result = await conn.execute(query)
             rows = result.fetchall()
+            grace_period = self._grace_period_checker.grace_period_const
             for kernel in rows:
-                grace_period = await self._grace_period_checker.get_grace_period(
-                    kernel, conn, self._redis_live
-                )
                 policy = policy_cache.get(kernel["access_key"], None)
                 if policy is None:
                     query = (
@@ -440,7 +438,9 @@ class NewUserGracePeriodChecker(AbstractIdleCheckReporter):
         )
 
     async def get_grace_period(
-        self, kernel: Row, dbconn: SAConnection, redis_obj: RedisConnectionInfo
+        self,
+        kernel: Row,
+        dbconn: SAConnection,
     ) -> float:
         """
         Calculate the user's initial grace period for idle checkers.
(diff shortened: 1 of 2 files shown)


adrysn (approved):
LGTM.

src/ai/backend/manager/idle.py
```diff
@@ -655,7 +669,13 @@ async def get_extra_info(self, session_id: SessionId) -> Optional[dict[str, Any]
...
 
     async def check_idleness(
-        self, kernel: Row, dbconn: SAConnection, policy: Row, redis_obj: RedisConnectionInfo
+        self,
+        kernel: Row,
+        dbconn: SAConnection,
+        policy: Row,
+        redis_obj: RedisConnectionInfo,
+        *,
+        grace_period_end: Optional[datetime],
```
kyujin-cho: Did it meant to be a keyword parameter? If then please add default value. 
```suggestion
        grace_period_end: Optional[datetime] = None,
```

src/ai/backend/manager/idle.py
```diff
@@ -667,11 +687,22 @@ async def check_idleness(
...
+                idle_time_start = max(kernel_created_at, grace_period_end)
+                if idle_time_start > now:
+                    remaining = idle_time_start - now + idle_timeout
+                else:
+                    idle_time = now - idle_time_start
+                    remaining = idle_timeout - idle_time
             result = remaining.total_seconds()
-            await self.set_remaining_time_report(redis_obj, session_id, result)
+            await self.set_remaining_time_report(
+                redis_obj, session_id, result if result > 0 else -1
```
kyujin-cho: What is the meaning of `-1`?

kyujin-cho (changes requested):

kyujin-cho (approved):

PR #1209, opened 2023-04-04
ci: add test cases for idle checker

fregataa:
Add test codes for idle checkers since it is hard to understand how they work.


kyujin-cho (changes requested):
Let's think about shortening test names.

PR #1212, opened 2023-04-04
fix: Remove pipeline-related configuration from manager

rapsealk:
In the previous PR #1094 , the field `pipeline.event-queue` has been added to the manager's configuration. However, it is uncomfortable to use since it forces to have the field, even if it is unnecessary.

https://github.com/lablup/backend.ai/blob/5a0ceccffb3c687c7f87397a5d1262e799c2ca71/configs/manager/halfstack.toml#L34-L35

Therefore, this PR removes pipeline-related configuration from manager and moves it to the plugin-side along with the PR below:
- lablup/backend.ai-event-stream-plugin#4


rapsealk:
Now I realized that `invoke_session_callback()` can be broadly used.
In conclusion, I think it would be better to keep `_make_session_callback()` and callback-related components inside the manager, not a plugin.

achimnol (approved):

PR #1214, opened 2023-04-07
fix: Optimize monthly stat API

fregataa:
So far, monthly stat API do iteration for (2880 * num_kernels) times constantly and fetch Kernel stats from Redis for every iteration.
This PR reduce the iteration and the number of Redis fetch.

# Before
Waiting for server response 4.77 s
* Not "milisecond", it is "second"
image

# After
Waiting for server response 140.08 ms
image

### DIFF
--- a/src/ai/backend/manager/api/resource.py
+++ b/src/ai/backend/manager/api/resource.py
@@ -9,7 +9,7 @@ import logging
 import re
 from datetime import datetime, timedelta
 from decimal import Decimal
-from typing import TYPE_CHECKING, Any, Iterable, MutableMapping, Tuple
+from typing import TYPE_CHECKING, Any, Iterable, Mapping, MutableMapping, Tuple
 
 import aiohttp
 import aiohttp_cors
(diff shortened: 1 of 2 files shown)


achimnol (approved):
LGTM for me, but need to confirm that the result is same to the previous implementation. @adrysn

src/ai/backend/manager/api/resource.py
```diff
@@ -605,74 +613,86 @@ async def get_time_binned_monthly_stats(request: web.Request, user_uuid=None):
...
-            if raw_stat:
-                last_stat = msgpack.unpackb(raw_stat)
-                io_read_bytes += int(nmget(last_stat, "io_read.current", 0))
-                io_write_bytes += int(nmget(last_stat, "io_write.current", 0))
-                disk_used += int(nmget(last_stat, "io_scratch_size/stats.max", 0, "/"))
-        stat = {
-            "date": ts,
+    time_series_list: list[dict[str, Any]] = [
+        {
+            "date": idx * time_window + now_ts,
```
adrysn: `date` calculation is incorrect. The `date` is starting from `now_ts` in this logic, but this should start from one month ago. Please refer to the previous logic, where `ts` starts from 30 days ago: https://github.com/lablup/backend.ai/pull/1214/files?w=1#diff-51adf4e1db91582bd045aea8a34e9565383e2c437c8cf7aa9eab64aebbdf9b3cL643

adrysn (changes requested):
The value of time in the time series data is incorrect.

src/ai/backend/manager/api/resource.py
fregataa: fix this part. thank you!

adrysn (approved):
LGTM.

PR #1219, opened 2023-04-10
ci: Change labeler settings to match the newly reorganized label

Yaminyam:
Labels have been reorganized.
Therefore, the setting of the labeler, which recognizes diff and automatically manages labels, also needs to be changed.


inureyes (approved):
LGTM

PR #1224, opened 2023-04-13
fix: remove kernel creation tracker

fregataa:
Initial design of session creation event handling logic was to let all manager processes "subscribe", instead of "comsume" events and have its own `kernel_creation_tracker` dictionary which keeps track of kernels spawned by each manager process. Having `kernel_creation_tracker` seemed to release burden for each process to query database the record of target kernel. But during the investigation (to enhance DB pressure) we found that `kernel_creation_tracker` method was not applied to some of event listeners, creating unwanted persistent DB queries. Moreover, after discussing in person, we made a conclusion that we no longer need to "subscribe" the event and have an additional `kernel_creation_tracker`. Therefore this PR removes usage of `kernel_creation_tracker` and makes manager process "consume" every kernel related events.


src/ai/backend/manager/models/kernel.py
```diff
@@ -589,19 +589,24 @@ async def get_kernel(
...
+    @classmethod
     async def set_kernel_status(
+        cls,
         db: ExtendedAsyncSAEngine,
         kernel_id: KernelId,
         status: KernelStatus,
         *,
         status_data: Optional[Mapping[str, Any]] = None,
         reason: Optional[str] = None,
+        current_time: Optional[datetime] = None,
```
kyujin-cho: How about changing parameter name to `status_changed_at`? Updating "current time" does not make sense literally.
```suggestion
        status_changed_at: Optional[datetime] = None,
```

src/ai/backend/manager/models/kernel.py
```diff
@@ -620,15 +625,63 @@ async def set_kernel_status(
...
+                    .options(
+                        noload("*"),
+                        load_only(KernelRow.status, KernelRow.session_id),
+                    )
+                )
+                kernel_row = (await db_session.scalars(kernel_query)).first()
+
+                if new_status not in KERNEL_STATUS_TRANSITION_MAP[kernel_row.status]:
+                    # TODO: log or raise error
+                    return False
```
kyujin-cho, line 660: Evaluating validity of `new_status` before starting DB transaction will prevent creating useless transaction.
```suggestion
        if new_status not in KERNEL_STATUS_TRANSITION_MAP[kernel_row.status]:
            # TODO: log or raise error
            return False

        async def _update() -> bool:
            async with db.begin_session() as db_session:
                kernel_query = (
                    sa.select(KernelRow)
                    .where(KernelRow.id == kernel_id)
                    .with_for_update()
                    .options(
                        noload("*"),
                        load_only(KernelRow.status, KernelRow.session_id),
                    )
                )
                kernel_row = (await db_session.scalars(kernel_query)).first()
```

src/ai/backend/manager/registry.py
```diff
@@ -1033,128 +1009,64 @@ def convert_resource_spec_to_resource_slot(
...
+                "status_history": sql_json_merge(
+                    kernels.c.status_history,
+                    (),
+                    {
+                        new_status.name: datetime.now(tzutc()).isoformat(),
+                    },
+                ),
+            }
+            self._kernel_actual_allocated_resources[kernel_id] = actual_allocs
+            is_updated = await KernelRow.update_kernel(
```
kyujin-cho: To make sense: 
```suggestion
            kernel_did_update = await KernelRow.update_kernel(
```

src/ai/backend/manager/models/kernel.py
fregataa: we cannot validate the kernel status because the transaction fetches `kernel_row` by `kernel_id`.

kyujin-cho: oops

kyujin-cho (changes requested):

src/ai/backend/agent/agent.py
```diff
@@ -1301,6 +1321,7 @@ def _rm(path: Path) -> None:
                         log.debug("cleanup requested: {} ({})", body["ID"], body.get("reason"))
                         terminated_kernels[body["ID"]] = ContainerLifecycleEvent(
                             KernelId(UUID(body["ID"])),
+                            SessionId(UUID(body["SID"])),
                             ContainerId(body["CID"]),
                             LifecycleEvent.DESTROY,
                             KernelLifecycleEventReason.from_value(body.get("reason"))
```
kyujin-cho: I think we can utilize `kernel_registry` and if matching kernel information does not exist we can also read `session-id` container label instead of updating Watcher's file format.

kyujin-cho (changes requested):

src/ai/backend/agent/agent.py
fregataa: I updated this part to utilize `kernel_registry`

kyujin-cho:
We can also think of removing `session_creation_tracker`.

kyujin-cho (approved):

PR #1226, opened 2023-04-14
feature: add pending session policy and impl predicate

### LINKED ISSUE #1120: Restriction of session creation with pending session count & pending requested resource slots limits
Let's implement new resource policy options to limit:
- the maximum number of pending sessions to create
- the maximum amount of resource slots allowed to stay as pending

This policy could be applied to both individual keypairs or projects.

If configured, these limits can be used to filter resource presets and set the maximum values of the custom resource sliders in the session launcher.

fregataa:
resolves #1120


src/ai/backend/manager/models/resource_policy.py
```diff
@@ -57,6 +57,8 @@
     sa.Column("total_resource_slots", ResourceSlotColumn(), nullable=False),
     sa.Column("max_session_lifetime", sa.Integer(), nullable=False, server_default=sa.text("0")),
     sa.Column("max_concurrent_sessions", sa.Integer(), nullable=False),
+    sa.Column("max_pernding_sessions", sa.Integer(), nullable=True),
```
inureyes: Typo here

achimnol: Typo! Also I'd like to rename this to `max_pending_session_count` to make it consistent and easier to distinguish with the followed `max_pending_session_resource_slots`.
```suggestion
    sa.Column("max_pending_session_count", sa.Integer(), nullable=True),
```

src/ai/backend/manager/scheduler/predicates.py
```diff
@@ -237,3 +243,78 @@ async def check_domain_resource_limit(
...
+    max_pending_session: Optional[int] = pending_session_policy["max_pernding_sessions"]
+    if max_pending_session is not None and max_pending_session > 0:
+        if len(pending_kernels) >= max_pending_session:
+            result = False
+            failure_msgs.append(
+                f"You cannot create more than {max_pending_session} pending session."
+            )
+
+    max_pending_session_resource: Optional[ResourceSlot] = pending_session_policy[
+        "max_pernding_sessions"
```
achimnol: Shouldn't this be `max_pending_session_resource_slots`?

src/ai/backend/manager/scheduler/predicates.py
```diff
@@ -237,3 +243,78 @@ async def check_domain_resource_limit(
...
+                keypair_resource_policies.c.max_pernding_sessions,
+                keypair_resource_policies.c.max_pending_session_resource_slots,
+            ]
+        )
+        .select_from(j)
+        .where(keypairs.c.access_key == sess_ctx.access_key)
+    )
+    pending_session_policy: dict[str, Any] = (await db_sess.execute(policy_query)).first()
+
+    max_pending_session: Optional[int] = pending_session_policy["max_pernding_sessions"]
```
achimnol: Rename the variables:

* `pending_session_policy` -> `policy`
* `max_pending_session` -> `max_pending_session_count`
  (in accordance with `max_pending_session_resource` below)
  - Forthermore, we can do for more clarity and readability:
    - `max_pending_session_count` -> `pending_count_limit`
    - `max_pending_session_resources` -> `pending_resource_limit`

src/ai/backend/manager/scheduler/predicates.py
```diff
@@ -237,3 +243,78 @@ async def check_domain_resource_limit(
...
+            msg = "Your pending session quota is exceeded. ({})".format(
+                " ".join(
+                    f"{k}={v}"
+                    for k, v in current_pending_session_slots.to_humanized(
+                        sched_ctx.known_slot_types
+                    ).items()
+                )
+            )
+            failure_msgs.append(msg)
+
```
achimnol, line 311: Where is the check for the number of pending sessions (`max_pernding_sessions`)?

achimnol (changes requested):

src/ai/backend/manager/scheduler/predicates.py
```diff
@@ -237,3 +243,78 @@ async def check_domain_resource_limit(
...
+            result = False
+            failure_msgs.append(
+                f"You cannot create more than {max_pending_session} pending session."
+            )
+
+    max_pending_session_resource: Optional[ResourceSlot] = pending_session_policy[
+        "max_pernding_sessions"
+    ]
+
+    if max_pending_session_resource:
```
achimnol: What if all resource slot values are 0?
Isn't it different from `null` value?
If so, it should do the explicit `None` check like the count check above.

achimnol (changes requested):

inureyes (changes requested):
Review left.

src/ai/backend/manager/scheduler/predicates.py
achimnol: Principle: If the current context and scope make the meaning unique, we can omit the prefixes in local variables.

src/ai/backend/manager/scheduler/predicates.py
```diff
@@ -237,3 +243,78 @@ async def check_domain_resource_limit(
...
+    kernel_query = (
+        sa.select([kernels.c.occupied_slots])
+        .select_from(kernels)
+        .where(
+            (kernels.c.status == KernelStatus.PENDING)
+            & (kernels.c.cluster_role == DEFAULT_ROLE)
+            & (kernels.c.access_key == sess_ctx.access_key)
+        )
+    )
+    pending_kernels: list[dict[str, Any]] = (await db_sess.execute(kernel_query)).fetchall()
```
achimnol: Doesn't the session rows have the pre-calculated sum of the resource slots of all its own kernels??

src/ai/backend/manager/scheduler/predicates.py
```diff
@@ -272,32 +272,30 @@ async def check_pending_session_limit(
...
-    max_pending_session: Optional[int] = pending_session_policy["max_pernding_sessions"]
-    if max_pending_session is not None and max_pending_session > 0:
-        if len(pending_kernels) >= max_pending_session:
+    pending_count_limit: Optional[int] = policy["max_pending_session_count"]
+    if pending_count_limit is not None and pending_count_limit > 0:
+        if len(pending_kernels) >= pending_count_limit:
             result = False
             failure_msgs.append(
-                f"You cannot create more than {max_pending_session} pending session."
+                f"You cannot create more than {pending_count_limit} pending session."
```
achimnol: A very minor fix:
```suggestion
                f"You cannot create more than {pending_count_limit} pending session(s)."
```

src/ai/backend/manager/scheduler/predicates.py
```diff
@@ -253,47 +247,23 @@ async def check_pending_session_limit(
...
-            ]
-        )
-        .select_from(j)
-        .where(keypairs.c.access_key == sess_ctx.access_key)
-    )
-    policy: dict[str, Any] = (await db_sess.execute(policy_query)).first()
-
-    pending_count_limit: Optional[int] = policy["max_pending_session_count"]
+    pending_count_limit: int | None = policy.max_pending_session_count
     if pending_count_limit is not None and pending_count_limit > 0:
```
achimnol: Hm.. What if the pending count limit is zero but not null?
The current implementation treats "zero" as "infinity".
But this may be confusing because there is no other way than reading the code to notice if "zero" means "forbidden" or "unlimited". (See lablup/backend.ai-webui#1328 and [@lizable's comment](https://github.com/lablup/backend.ai-webui/pull/1328#issuecomment-1193135636))

Since it is a nullable field, I think it would be clearer to treat "null" as "unlimited/unspecified".
Also, we need to avoid making two or more distinct values have the same meaning in general, when designing a field like this.
If you want to keep "zero" as "unlimited", then I'd suggest making the field not nullable.

How about your thoughts?

achimnol (changes requested):
LGTM mostly, but please check out the comment about the meaning of "zero".

src/ai/backend/manager/scheduler/predicates.py
fregataa: I agree with you that "zero" should be different with "null".
users want to set the value of limitation as "zero".

src/ai/backend/manager/models/session.py
achimnol, line 1: I think this is too arbitrary extraction of a method.
Could we just perform this query inside the predicate?

src/ai/backend/manager/scheduler/predicates.py
```diff
@@ -245,3 +247,76 @@ async def check_domain_resource_limit(
             ),
         )
     return PredicateResult(True)
+
+
+async def check_pending_session_limit(
```
achimnol: How about splitting the pending sesison count limit and the pending resource limit for clarity?

src/ai/backend/manager/models/alembic/versions/c422e982597d_add_max_pending_session_count.py
```diff
@@ -0,0 +1,40 @@
+"""add_max_pending_session_count
+
+Revision ID: c422e982597d
+Revises: d04592473df7
```
achimnol: Please "rebase" this migration onto the latest main.

achimnol (changes requested):

src/ai/backend/manager/models/session.py
fregataa: this method is not used. I removed it!

achimnol:
This PR should be revisited after migrating keypair resource policies to user resource polices.

PR #1229, opened 2023-04-16
setup: Upgrade aiomonitor-ng to 0.7.1

achimnol:
aiomonitor-ng 0.7.1 fixes transparent passing of `context` kwarg when using `asyncio.create_task()` and the task factory hook, which is a necessary support requirement for Python 3.11.

refs)
- https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task
  image
- achimnol/aiomonitor-ng#10


fregataa:
`asyncio.create_task` has `context` parameter but `asyncio.Task` does not have `context` parameter
https://docs.python.org/ko/3/library/asyncio-task.html#asyncio.Task
we should not pass the `context` parameter when init `TracedTask` object

PR #1233, opened 2023-04-17
feature: Enable folder upload

fregataa:
Enable vfolder upload function to allow you to upload folders by making parent directory of a target path.


src/ai/backend/storage/api/client.py
```diff
@@ -277,7 +277,8 @@ async def tus_upload_part(request: web.Request) -> web.Response:
                     parent_dir = vfpath / dst_dir
                 if not parent_dir.exists():
                     parent_dir.mkdir(parents=True, exist_ok=True)
-                target_path = parent_dir / token_data["relpath"]
+                target_path: Path = parent_dir / token_data["relpath"]
+                target_path.parent.mkdir(parents=True, exist_ok=True)
```
agatha197: I think 278~279 lines can be removed because the 281 line can cover it.

Sujin-Kim1 (changes requested):
I commented minor one. But it works!

src/ai/backend/storage/api/client.py
fregataa: thanks! I updated it.

Sujin-Kim1 (approved):
LGTM

lizable (changes requested):
folder upload fails when the directory is empty. @fregataa Could you check it and fix it?
Here's the directory tree that I tested.
Screenshot 2023-04-19 at 6 01 31 PM

achimnol (approved):

PR #1235, opened 2023-04-21
feature: Impl recursive vfolder upload

fregataa:
Add `recursive` option to vfolder upload function and command.


src/ai/backend/client/func/vfolder.py
```diff
@@ -374,15 +371,57 @@ async def upload(
...
+        show_progress: bool = False,
+    ) -> None:
+        if basedir:
+            src_paths = [basedir / Path(src) for src in sources]
+        else:
+            src_paths = [Path(src).resolve() for src in sources]
+        if recursive:
+            await self._upload_recursively(src_paths, basedir, dst_dir, chunk_size, address_map)
+            return
+        await self._upload_files(src_paths, basedir, dst_dir, chunk_size, address_map)
```
achimnol: I think the following pattern is more readable, as this method does not have any return value.
```suggestion
        else:
            await self._upload_files(src_paths, basedir, dst_dir, chunk_size, address_map)
```

src/ai/backend/client/func/vfolder.py
```diff
@@ -394,6 +433,15 @@ async def mkdir(
         async with rqst.fetch() as resp:
             return await resp.text()
 
+    @api_function
+    async def mkdir(
+        self,
+        path: Union[str, Path],
+        parents: Optional[bool] = False,
+        exist_ok: Optional[bool] = False,
+    ):
```
achimnol: Following the type annotation you've added:
```suggestion
    ) -> str:
```

achimnol (changes requested):
Some minor clean-up please.
Otherwise, LGTM.

achimnol (approved):

achimnol:
I'm going to propose two additional enhancements, after first merging this PR:

- Rewrite `_upload_recursively()` using an iteration loop without using recursion
- Wrap file uploads as subtasks of a `aiotools.Supervisor` and use `asyncio.Semaphore` to limit the concurrency (configurable via CLI option)

PR #1237, opened 2023-04-24
repo: Replace ./pants with scie-pants

achimnol:
Pants recently [has replaced](https://blog.pantsbuild.org/the-pants-launcher-binary-a-much-simpler-way-to-install-and-run-pants/) their bootstrapping process from a shell script `./pants` to a custom Rust-based binary [`scie-pants`](https://github.com/pantsbuild/scie-pants/).
This makes it faster to launch pants commands and also simplifies the installation process as it now automatically downloads and installs the Python interpreter required by Pannts!

### Migration Guide

Just follow the new official installation instructions:
https://www.pantsbuild.org/docs/installation

- Linux: `curl --proto '=https' --tlsv1.2 -fsSL https://static.pantsbuild.org/setup/get-pants.sh`
- macOS: `brew install pantsbuild/tap/pants`

You may remove any locally built pants, if any, including the `pants-local` script and `tools/pants-src` directory.
Also you may remove additional pants configurations generated by either yourself or the previous install-dev script, such as `.pants.bootstrap` and `.pants.env`.

Now all pants commands will work with `pants`, instead of `./pants`.

### Changes to the install-dev script

It will now first look for the availability of `pants` command.
If not exists, it will try the installation by using the official script on Linux and the brew command on macOS.


achimnol:
Confirmed working of install-dev on a new setup.

PR #1243, opened 2023-04-24
feat: don't count `SYSTEM` kernels when calculating occupancy

kyujin-cho:
This PR exempts `SYSTEM` role kernels when calculating resource occupancy (per-domain, keypair, project).


inureyes (approved):
LGTM

inureyes (changes requested):
Session count is still increasing when running system container.

inureyes (approved):

PR #1251, opened 2023-04-26
fix: Change how to assign scaling group when enqueue session

fregataa:
Public scaling group is given priority only when enqueuing session with no scaling group specified


inureyes (approved):
LGTM

inureyes (changes requested):
The code is hard to understand.

kyujin-cho:
Closing as #1253 is merged.

PR #1264, opened 2023-05-05
setup: standalone python build on `install-dev.sh`

kyujin-cho:
This PR updates `scripts/install-dev.sh` script to use [indygreg's standalone python build](https://github.com/indygreg/python-build-standalone/releases) as python runtime for bootstraping. Doing so will reduce burden to take care of multiple python versions.


scripts/install-dev.sh
```diff
@@ -146,13 +146,23 @@ show_important_note() {
...
   else
     echo 1
   fi
 }
 
+install_indygreg_python() {
+  url="https://github.com/indygreg/python-build-standalone/releases/download/20230116/cpython-3.11.1+20230116-$INDYGREG_ARCH-$INDYGREG_PLATFORM-install_only.tar.gz"
+  cwd=$(pwd)
+  mkdir -p python-runtime && cd python-runtime
+  show_info "Downloading and installing standalone Python 3.11.1..."
```
Yaminyam: Is there a reason you specified the python version directly without using a variable?

scripts/install-dev.sh
```diff
@@ -146,13 +146,23 @@ show_important_note() {
...
     echo 0
   else
     echo 1
   fi
 }
 
+install_indygreg_python() {
+  url="https://github.com/indygreg/python-build-standalone/releases/download/20230116/cpython-3.11.1+20230116-$INDYGREG_ARCH-$INDYGREG_PLATFORM-install_only.tar.gz"
+  cwd=$(pwd)
+  mkdir -p python-runtime && cd python-runtime
```
achimnol: How about using a hidden directory, like `.python-runtime`?

scripts/install-dev.sh
achimnol: I think it's just the latest version available at the time of writing. The version requirement is from what we do in `install-dev.sh`, and this is separate from what Backend.AI itself needs.

scripts/install-dev.sh
```diff
@@ -146,13 +146,23 @@ show_important_note() {
...
   "$1" -c '' >/dev/null 2>&1
-  if [ "$?" -eq 127 ]; then
+  if [ "$?" -ne 0 ]; then
     echo 0
   else
     echo 1
   fi
 }
 
+install_indygreg_python() {
```
achimnol: We appreciate the hard work by indygreg, but it's not necessary to name our functions and variables following him. :wink:
Let's just use the purpose-oriented naming like "boostrapping", "relocatable", "portable", "static", etc.

achimnol (changes requested):

scripts/install-dev.sh
achimnol: Also we should not make an illusion that this runtime is also used for Backend.AI.
Maybe something like `.bootstrap-python`?

scripts/install-dev.sh
kyujin-cho: You've got it right. Additionally most of the jobs `$bpython` is to execute standalone python scripts (which does not require state-of-the-art python versions) and thus I think it is okay to just leave the version hard-coded rather than being customizable (either by user or maintainer).

achimnol (approved):
Fixed some mistakes and now it is confirmed working in a fresh new Ubuntu VM.

PR #1266, opened 2023-05-06
Add `agent_id` label to containers

### LINKED ISSUE #875: Distinguish which agent instance owns a running kernel container
When rescanning the running containers and processing the Docker event (lifecycle events) stream, we need to identify which agent instance is the owner of the related container.

To prevent dangling containers when the agent restarts with removal of pickled `kernel_registry` during the recovery process, we need to explicitly attach container labels such as `ai.backend.owner` set as the agent instance ID (from `[agent].id` of `agent.toml`).

Previously #712 has implemented another way that just skips any containers not present in `kernel_registry`, but we need to rewrite it to use explicit container labels.

fregataa:
resolves #875

### DIFF
--- a/changes/1266.feature.md
+++ b/changes/1266.feature.md
@@ -1,0 +1,1 @@
+Add `agent_id` label to containers as `ai.backend.owner`.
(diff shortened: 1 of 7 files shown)


achimnol (changes requested):
I think we also need to consider the container's `ai.backend.owner` labels when scanning contaienrs in the agents.

fregataa:
> I think we also need to consider the container's `ai.backend.owner` labels when scanning contaienrs in the agents.

Should we remove `kernel_registry` or impl a method to sync it by scanning containers?

fregataa:
> I think we also need to consider the container's `ai.backend.owner` labels when scanning contaienrs in the agents.

When scanning containers, should I sync `kernel_registry` by creating missing kernel object or just skip other's kernel object, or do both?

achimnol (approved):

PR #1267, opened 2023-05-08
feat: use `zsh` as the default shell if the binary is available in the kernel image

hydroxyde:
This PR resolves https://github.com/lablup/giftbox/issues/284

Zsh will be used as a default shell if available in kernel image. It also adds a .zshrc for a better shell experience which includes auto-completion on tabulation and more text coloration.


src/ai/backend/kernel/intrinsic.py
```diff
@@ -148,9 +148,16 @@ async def prepare_sshd_service(service_info):
 
 async def prepare_ttyd_service(service_info):
     shell = "sh"
-    if Path("/bin/bash").exists():
+    if Path("/bin/zsh"):
+        shell = "zsh"
+        zshrc_path = Path("~/.zshrc").expanduser()
+        with zshrc_path.open("w") as f:
+            f.write("export PATH=$PATH:/opt/backend.ai/bin\n")
```
achimnol: Is this necessary? `/opt/backend.ai/bin` is intended to be private for our kernel runner. For instance, if the user executes "python" in their terminal sessions, it should launch the image's default Python instead of our own prebuilt Python tailored for the kernel runner.

src/ai/backend/kernel/intrinsic.py
```diff
@@ -148,9 +148,16 @@ async def prepare_sshd_service(service_info):
...
+            f.write("export TERM=xterm-256color\n")
```
achimnol: I think we are already providing the `TERM` env-var with the value "xterm".
Is there any extra features in zsh that requires this?

src/ai/backend/kernel/intrinsic.py
```diff
@@ -148,9 +148,16 @@ async def prepare_sshd_service(service_info):
...
+            f.write("autoload -Uz compinit && compinit\n")
```
achimnol: I think this should be done automatically. Could you explain why this is required for zsh?

hydroxyde: I think this command is not always enabled by default in Zsh, that's why I added this line

src/ai/backend/kernel/intrinsic.py
hydroxyde: In that case we can remove this line. The main purpose was to add colors to the shell

src/ai/backend/kernel/intrinsic.py
hydroxyde: In that case you are right, this isn't necessary

achimnol:
@hydroxyde How is this going?

achimnol (approved):

PR #1268, opened 2023-05-08
health check on storage proxy

hydroxyde:
This PR resolves https://github.com/lablup/giftbox/issues/198

 An endpoint `status` is added to the manager-facing API and if storage-proxy is alive an OK 200 is returned. 
 
 To test this run: 
      `curl -k https://127.0.0.1:6022/status`


src/ai/backend/storage/server.py
```diff
@@ -31,6 +31,10 @@
 log = BraceStyleAdapter(logging.getLogger(__spec__.name))  # type: ignore[name-defined]
 
 
+async def health_check(request: web.Request) -> web.Response:
+    return web.json_response({"status": "OK"})
+
+
```
kyujin-cho: API handlers on storage-proxy are defined under `src/ai/backend/storage/api/client.py` `src/ai/backend/storage/api/manager.py`, depending on its characteristic. Looks like the purpose of this health check API is for the client side so let's move it to client.py.

kyujin-cho (changes requested):

src/ai/backend/storage/server.py
adrysn: This API is required from HAProxy to check the status of the Storage Proxy service, as stated in lablup/giftbox#198, so I think it is correct to implement  in `server.py`.

src/ai/backend/storage/api/client.py
```diff
@@ -346,6 +346,8 @@ async def prepare_tus_session_headers(
     headers["Upload-Length"] = str(token_data["size"])
     return headers
 
+async def health_check(request: web.Request) -> web.Response:
```
adrysn: This is totally my preference, but I think it would be enough to name the handler as `status`, instead of `health_check`, since it is shorter...

src/ai/backend/storage/api/client.py
```diff
@@ -365,4 +367,5 @@ async def init_client_app(ctx: Context) -> web.Application:
     r.add_route("OPTIONS", tus_options)
     r.add_route("HEAD", tus_check_session)
     r.add_route("PATCH", tus_upload_part)
+    cors.add(app.router.add_route("GET", r"/health_check", health_check))
```
adrysn: Sames goes for the endpoint name.

adrysn (changes requested):
Please fix the linting issues before requesting the review.

src/ai/backend/storage/server.py
adrysn: On second thought, to query the status of Storage Proxy from the Control Panel, it is also required to implement in client.py ... Sorry for my confusion.

adrysn: Hm... I'm confused. Seems like implementing the status API in `server.py` would be enough to query it from Control Panel. What is needed in deploy the service via HAProxy is to return `200 OK` for `http://<storage-proxy-url>/status`.

hydroxyde:
@adrysn About the name of the status function, there is already a get_status function. I think we could make it a little be more explicit to avoid confusion between both purposes. 

- [ ] get_storage_proxy_status()

- [ ] health()
What do you think?

adrysn:
> @adrysn About the name of the status() function, there is already a get_status() function. I think we could make it a little more explicit to avoid confusion between both purposes.
> 
> * [ ]  get_storage_proxy_status()
> * [ ]  health()

@hydroxyde  I was not able to find `get_status` function in storage-proxy. Can you leave a link for the handler?

hydroxyde:
> > @adrysn About the name of the status() function, there is already a get_status() function. I think we could make it a little more explicit to avoid confusion between both purposes.
> > 
> > * [ ]   get_storage_proxy_status()
> > * [ ]   health()
> 
> @hydroxyde I was not able to find `get_status` function in storage-proxy. Can you leave a link for the handler?

I was referring to this function: https://github.com/lablup/backend.ai/blob/09a9d51afb4490a21d5fc334feae985385c50329/src/ai/backend/storage/api/manager.py#L45

adrysn:
> > > @adrysn About the name of the status() function, there is already a get_status() function. I think we could make it a little more explicit to avoid confusion between both purposes.
> > > 
> > > * [ ]    get_storage_proxy_status()
> > > * [ ]    health()
> > 
> > 
> > @hydroxyde I was not able to find `get_status` function in storage-proxy. Can you leave a link for the handler?
> 
> I was referring to this function:
> 
> https://github.com/lablup/backend.ai/blob/09a9d51afb4490a21d5fc334feae985385c50329/src/ai/backend/storage/api/manager.py#L45

Good. I think the API needs to be used without secret header for proxy clients.

achimnol (approved):
LGTM!

achimnol (approved):
LGTM!

PR #1269, opened 2023-05-08
fix: Skip execution of `pr-number-assign.py` script in merge queues

kyujin-cho:
This PR updates `pr-number-assign.py` to prevent update-news-fragments action failing when PR is merged by merge queue. Looks like actions with trigger `workflow_call` is ran on push to `main` branch when PR is merged via merge queue Because of this, the `pr-number-assign.py` will be started with empty string (originated from empty "github.event.pull_request.number" action variable) as an argument, making script to crash. The updated script will now also gracefully terminate with a warning message complaining about missing Pull Request number, making action runner to convince that execution has done without any error.


achimnol:
It would be better to use GitHub Action's conditional workflow execution syntax.
refs)
https://docs.github.com/en/actions/learn-github-actions/expressions
https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution

Yaminyam (commented):
Does `timeline-check` need to be triggered with merge_group?

achimnol:
> Does `timeline-check` need to be triggered with merge_group?

Yes, it is a required check for all merging PRs, unless the PR is marked with "skip:changelog".
But if the PR number and labels are not available in the merge group action, it is fine to skip it.

scripts/pr-number-assign.py
```diff
@@ -61,5 +61,8 @@ def main(pr_number: str) -> None:
 
 
 if __name__ == "__main__":
+    if len(sys.argv) < 2:
+        print(f"Usage: python ${sys.argv[0]} <PR Number>")
+        exit(0)
```
achimnol: I think still this should be an explicit failure.

achimnol (changes requested):

achimnol (approved):

PR #1272, opened 2023-05-08
chore: Add the python version to the classifiers of package

Yaminyam:
https://pypi.org/project/backend.ai/
https://pypi.org/project/backend.ai-manager/
Packages prior to the existing monorepo included more information in the classifier.
In particular, the python version is indicated in required, but I thought it would be nice to have it in the package classification information, so I added it.


achimnol (approved):
👍🏼

PR #1273, opened 2023-05-09
ci: Do not run the workflow when it is not needed using paths-ignore

Yaminyam:
Use paths-ignore to prevent actions from being executed meaninglessly.
When actions are skipped using paths-ignore, required is not satisfied.
Since required is recognized as the name of the job, the job corresponding to the skipped workflow must be forcibly executed.
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/troubleshooting-required-status-checks#handling-skipped-but-required-checks

### DIFF
--- a/.github/workflows/default-skipped.yml
+++ b/.github/workflows/default-skipped.yml
@@ -1,0 +1,20 @@
+name: Skipped Require default
+on:
+  pull_request:
+    paths:
+      - 'docs/**'
+      - '.github/**'
+      - '!.github/workflows/default.yml'
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - run: 'echo "No default required"'
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - run: 'echo "No default required"'
+  typecheck:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/default-skipped.yml
```diff
@@ -0,0 +1,20 @@
+name: Skipped Require default
```
achimnol, line 1: Let's use consistent naming style for the action workflows.
Please rename it to `default-skipped` for brevity in the PR check list.

.github/workflows/timeline-check.yml
```diff
@@ -3,6 +3,9 @@ name: timeline-check
 on:
   pull_request:
     types: [labeled, unlabeled, opened, synchronize, reopened]
+    paths-ignore:
+      - 'docs/**'
+      - '.github/**'
```
achimnol, line 8: Like default.yml, shouldn't we also add:
`- '!.github/workflows/timeline-check.yml'`
for the cases when it is self-modified?

achimnol: And how about other workflows?

achimnol (changes requested):

PR #1277, opened 2023-05-14
feature: impl reconciliation loop

fregataa:
Impl basic reconciliation loop in common and apply it to manager and agent.


achimnol:
This will require #1283 for thorough testing.

fregataa:
Closed this PR because it is too old to maintain and we will implement a new loop concept `(Loop)Runner`.

PR #1280, opened 2023-05-15
setup: Preserve cached bootstrapping python and improve macOS compatibility

kyujin-cho:
Follow-up PR of #1264.

- Use `~/.cache/bai/bootstrap/cpython/3.11.3` as the cached runtime directory path
- Use `shasum -a 256` for BSD (macOS) compatibility instead of `sha256sum`
- Avoid confusion with `PYTHONPATH`


scripts/install-dev.sh
```diff
@@ -192,6 +192,7 @@ fi
 KNOWN_DISTRO="(Debian|Ubuntu|RedHat|CentOS|openSUSE|Amazon|Arista|SUSE)"
 DISTRO=$(lsb_release -d 2>/dev/null | grep -Eo $KNOWN_DISTRO  || grep -Eo $KNOWN_DISTRO /etc/issue 2>/dev/null || uname -s)
 
+STANDALONE_PYTHON_PATH="$HOME/.cache/python-runtime"
```
achimnol: I think for future expansion, it would be better to include the version number in the directory name, and prefix it with a namespace like other apps.

e.g., `$HOME/.cache/bai/bootstrap/cpython/3.11.3`

PR #1282, opened 2023-05-17
fix: Resolve session `resource_opts` from sibling kernels

fregataa:
resolve session's `resource_opts` from sibling kernels since `resource_opts` depends on kernel's images.
and remove `resource_opts` column of session table.


lizable (approved):
LGTM.

PR #1284, opened 2023-05-18
impl feature to force password updates periodically

fregataa:
related issue: https://github.com/lablup/giftbox/issues/343

impl feature to force password updates periodically.

---
## How to set up for test

1. upgrade migration by running `./py -m alembic upgrade head`
2. insert auth config to etcd
ex - save the json below to a <FILE_NAME> and put it by running
`backend.ai mgr etcd put-json config/auth <FILE_NAME>`
```
{
  "max_password_age": "1m",
  "shared_secret": "random",
  "exp": "5m"
}
```
"1m" means 1 minute! Available units to represent the time above are "w, d, h, m, s".

3. create user and check it!

---

## APIs
```
# Direct to manager
POST <MANAGER_URL>/auth/update-password-no-auth
{
  "domain" : "",
  "username" : "",
  "current_password": "",
  "new_password": ""
}

# To webserver
POST <WEBSERVER_URL>/server/update-password-no-auth
{
  "username" : "",
  "current_password": "",
  "new_password": ""
}

```

## Example request

image

```
curl --request POST \
  --url http://127.0.0.1:8090/server/update-password-no-auth \
  --header 'Content-Type: application/json' \
  --header 'Date: 2023-05-23T08:59:49.685263+00:00' \
  --header 'X-BackendAI-Domain: default' \
  --header 'X-BackendAI-Version: v6.20230315' \
  --data '{
	"username": "test@email.com",
	"current_password": "1234",
	"new_password": "1234"
}
'
```


src/ai/backend/manager/api/exceptions.py
```diff
@@ -173,6 +173,11 @@ class AuthorizationFailed(BackendError, web.HTTPUnauthorized):
     error_title = "Credential/signature mismatch."
 
 
+class PasswordExpired(BackendError, web.HTTPUnauthorized):
+    error_type = "https://api.backend.ai/probs/password-expired"
+    error_title = "Should update password."
```
yomybaby: It would be helpful if the error message for PasswordExpired is more specific, such as "Password has exceeded the maxage for change."

src/ai/backend/manager/api/auth.py
```diff
@@ -502,6 +533,8 @@ async def _pipe_builder(r: Redis) -> RedisPipeline:
             pass
```
yomybaby: It would be nice to change this message as well.

src/ai/backend/manager/api/auth.py
```diff
@@ -951,12 +985,75 @@ async def update_password(request: web.Request, params: Any) -> web.Response:
...
         await conn.execute(query)
     return web.json_response({}, status=200)
 
 
+@check_api_params(
+    t.Dict(
+        {
+            t.Key("domain"): t.String,
+            t.Key("email"): t.String,
+            t.Key("old_password"): t.String,
```
yomybaby: "current_password" or just "password" is better.

src/ai/backend/manager/api/auth.py
```diff
@@ -951,12 +985,75 @@ async def update_password(request: web.Request, params: Any) -> web.Response:
...
+            t.Key("new_password"): t.String,
+            t.Key("new_password2"): t.String,
```
yomybaby: Is it necessary to include `new_password2` in the API? I think that `new_password` is enough. Confirming the password is not a concern of the API.

yomybaby (changes requested):

src/ai/backend/manager/api/exceptions.py
fregataa: update the message! but extra information will be passed by `extra_msg` parameter.

src/ai/backend/manager/api/auth.py
fregataa: which part is this?

src/ai/backend/manager/api/auth.py
fregataa: I agree with you. I tried to match the parameters with existing `update_password` API but it is cumbersome. I updated it!

lizable:
@yomybaby @adrysn @Sujin-Kim1 It seems like for now we can only configure the `max_password_age` by inserting via CLI, but I think It would be great to have a corresponding UI (at least for administrator) for that. What do you think?

PR #1288, opened 2023-05-20
fix: Handle `KeyError` on `session.status_changed` GraphQL resolver

fregataa:
Handle error of Session `status_changed` if any status is not registered in `status_history`.


adrysn (approved):
LGTM for me.

PR #1291, opened 2023-05-23
fix: MemoryPlugin implementation bug

jopemachine:
The PR fixed a [bug](https://github.com/lablup/giftbox/issues/374) in the `gather_container_measures` function of the agent where memory was miscalculated when the StatModes was set to cgroup.

## Test

To be honest, I couldn't find a way to change `ctx.mode` to cgroup. I thought it would automatically be set to cgroup when the host OS is Linux, but it still seems to be detected as Docker. So I changed the below value to `sysfs_impl` when testing.

https://github.com/lablup/backend.ai/blob/main/src/ai/backend/agent/docker/intrinsic.py#L682

I was able to verify that it produces values that are almost identical to the output of the `docker container stats` command.

### cgroup v1

v1

`result 1` is 157,470,720 bytes, and docker says 157,529,192 bytes.

### cgroup v2

v2

`result 2` is 123,011,072 bytes, and docker says 123,036,067 bytes.

### Others

You can use cgroup v1 by set `GRUB_CMDLINE_LINUX="systemd.unified_cgroup_hierarchy=0"` in `/etc/default/grub`.


achimnol (approved):
LGTM!

PR #1292, opened 2023-05-23
feature: Sync manager db and agent kernel_registry

fregataa:
Enable hard-sync manager DB and agent's `kernel_registry`
1. If there is alive kernel row related to certain agent in DB but not found in `kernel_registry`, then produce `KernelTerminated` event.
2. If there is kernel data in `kernel_registry` but not found in alive kernel row of the agent in DB, then destroy the kernel.

## Updated manager API
image

```
curl --request POST \
  --url http://127.0.0.1:8090/func/session/_/sync-agent-registry \
  --header 'Content-Type: application/json' \
  --header 'Date: 2023-05-24T07:15:49.685263+00:00' \
  --header 'X-BackendAI-Domain: default' \
  --header 'X-BackendAI-Version: v6.20230315' \
  --cookie AIOHTTP_SESSION=<COOKIE> \
  --data '{
	"agent": "agent-id"
}
'
```


src/ai/backend/manager/api/session.py
```diff
@@ -2807,4 +2832,7 @@ def create_app(
     cors.add(app.router.add_route("POST", "/{session_name}/commit", commit_session))
     cors.add(app.router.add_route("GET", "/{session_name}/commit", get_commit_status))
     cors.add(app.router.add_route("GET", "/{session_name}/abusing-report", get_abusing_report))
+    cors.add(
+        app.router.add_route("POST", "/{session_name}/sync-agent-registry", sync_agent_registry)
```
achimnol: This should not take `session_name` parameter. Let's use `/_/sync-agent-registry`.

achimnol (changes requested):
Just a small fix please. Otherwise LGTM.

src/ai/backend/agent/server.py
```diff
@@ -371,8 +371,10 @@ async def sync_kernel_registry(
                         reason=KernelLifecycleEventReason.ALREADY_TERMINATED,
                     )
                 )
+
+        kernel_ids = {kern_id for kern_id, sess_id in kernel_session_ids}
```
achimnol, line 374: FYI: If a variable is not used, you may use `_` as a convention to indicate "unused" variable.

src/ai/backend/manager/api/session.py
```diff
@@ -1438,18 +1438,15 @@ async def get_abusing_report(request: web.Request, params: Mapping[str, Any]) ->
 )
 async def sync_agent_registry(request: web.Request, params: Any) -> web.StreamResponse:
     root_ctx: RootContext = request.app["_root.context"]
-    session_name: str = request.match_info["session_name"]
     requester_access_key, owner_access_key = await get_access_key_scopes(request)
 
-    log.info(
-        "SYNC_AGENT_REGISTRY (ak:{}/{}, s:{})", requester_access_key, owner_access_key, session_name
-    )
+    log.info("SYNC_AGENT_REGISTRY (ak:{}/{})", requester_access_key, owner_access_key)
```
achimnol: Let's include the agent ID here so that admins could later inspect which agents were force-synchronized.

src/ai/backend/manager/api/session.py
```diff
@@ -1438,18 +1438,15 @@ async def get_abusing_report(request: web.Request, params: Mapping[str, Any]) ->
...
     try:
-        await root_ctx.registry.sync_agent_kernel_registry(params["agent"])
+        await root_ctx.registry.sync_agent_kernel_registry(AgentId(params["agent"]))
     except BackendError:
         log.exception("SYNC_AGENT_REGISTRY: exception")
         raise
-    return web.json_response({}, status=200)
+    return web.json_response({}, status=201)
```
achimnol: In this case, we should keep "200 OK" instead of "201 Created", because we have not added anything.

src/ai/backend/agent/server.py
```diff
@@ -346,6 +351,39 @@ async def gather_hwinfo(self) -> Mapping[str, HardwareMetadata]:
...
+        for kid, kernel in self.agent.kernel_registry.items():
+            if kid not in kernel_ids:
+                # destroy kernel
+                await self.agent.inject_container_lifecycle_event(
+                    kid,
+                    kernel.session_id,
+                    LifecycleEvent.DESTROY,
+                    KernelLifecycleEventReason.NOT_FOUND_IN_MANAGER,
+                    suppress_events=True,
+                )
```
achimnol, line 384: How about saving `self.agent.kernel_registry` into the `last_registry` file after sync?

achimnol (changes requested):

src/ai/backend/agent/server.py
fregataa: I prefer to name the unpacked value over using `_` for better readability

src/ai/backend/manager/api/session.py
fregataa: updated!

src/ai/backend/agent/server.py
fregataa: good point. then the agent will not load the "broken" registry pickle

src/ai/backend/agent/server.py
achimnol: OK for this case. Note that the pattern matching syntax uses `_` as an explicit meaning for fallback.

achimnol (approved):

PR #1295, opened 2023-05-24
fix: Agent resource doubling with recalc txn retries

achimnol:
There are higher chances to get doubling of agent's occupied resources when the transaction in `recalc_resource_usage()` is retried, because `occupied_slots_per_agent` is not re-initialized when the txn function is retried.
This function is called when a session creation is failed or cancelled, and at that timing it has some chances to get retried if there are many session creation requests.

This PR moves the declaration of `occupied_slots_per_agent` into the inner txn function to make it reinitialized everytime.

Note that `concurrency_used_per_key` also needs to be reinitialized in principle, but it is referred after & outside of the txn function and it uses `set`, so it's safe to re-add the partial txn results there as set removes duplication automatically.


adrysn (approved):
LGTM.

fregataa (approved):

PR #1299, opened 2023-05-25
feat: concurrency limit for system sessions

kyujin-cho:
Follow-up of #1243. This PR introduces `keypair_resource_policies.max_concurrent_sftp_sessions` column. Sessions created from SFTP scaling groups will now follow this rule, overriding previous behavior, which does not limit maximum number of SFTP upload sessions. In contrast to normal sesions, `SYSTEM` sessions (also known as SFTP upload sessions) failing concurrency check will instantly transit to `CANCELLED` status rather than keeping its state at `SCHEDULING`. This is to feedback UI that user has reached its own concurrency limit.

### DIFF
--- a/changes/1299.feature.md
+++ b/changes/1299.feature.md
@@ -1,0 +1,1 @@
+Intoduce separate concurrency limit for SFTP upload sessions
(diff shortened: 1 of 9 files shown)


fregataa (commented):
looks good overall.
but is it possible to build any other `SYSTEM` role image which is not sftp in the future?

kyujin-cho:
> looks good overall. but is it possible to build any other `SYSTEM` role image which is not sftp in the future?

That can happen in future, but I think current structure is not that complicated for further expansion (and also generalizations).

PR #1301, opened 2023-05-25
ci: Fix wrong db migration file path

Yaminyam:
Modify that the models path is missing in the middle of the labeler path that sets the required:db-migration label.


achimnol (approved):
Ooops! :eyes:

PR #1304, opened 2023-05-26
fix: use session_id to destroy kernels instead of kernel.session_id

fregataa:
In `destroy_session()`, using `kernel.session_id` which is not eager-loaded occurs error.
This pr fix it by using `session_id` which is already loaded.


adrysn (approved):
LGTM

adrysn (approved):
LGTM

PR #1312, opened 2023-05-30
fix: Handle buggy orm field loading when destroy session

fregataa:
ORM objects must load its fields before access to them.
Else, we get [DetachedInstanceError](https://docs.sqlalchemy.org/en/20/orm/exceptions.html#sqlalchemy.orm.exc.DetachedInstanceError)

```
File "/home/bai-23.03-manager/lib/python3.11/site-packages/ai/backend/manager/registry.py", line 1632, in destroy_session
      kernel.session_creation_id,
      ^^^^^^^^^^^^^^^^^

```


adrysn (approved):
LGTM.

achimnol:
The current test failure looks like an issue from Pants: https://github.com/pantsbuild/pants/issues/18519

PR #1313, opened 2023-06-01
feat: Impl dummy agent

### LINKED ISSUE #1283: Add a "dummy" agent for scalability tests
Let's write a dummy agent backend and a script to spawn hundreds of agents on a single machine to test the manager's scalability regarding database transactions, reconcilation loop designs, and potential race condition bugs.

The dummy backend actually does not create any container but pretends as if it creates/destroys the containers.

* A separate configuration file defines:
  - the delay patterns of various container creation/termination steps
  - the image list that the agent is assumed to already have / need to pull / report as missing
* Most container creation/termination steps are no-op but just sleeps.
* The container list is preserved in the memory.
* All code execution requests will be silently ignored but replied with a default message in stdout.
  - No service ports are supported. (Maybe added in the future)

fregataa:
resolves #1283 

- dummy agents spawn dummy kernels.
- dummy kernels does not affect to agent's resource stat.
- dummy agent's kernels create one of two kind of code runner, `DummyCodeRunner` or `DummyFakeCodeRunner`. `DummyCodeRunner` initiates ZMQ sockets and implements only abstract methods of `AbstractCodeRunner`. 
`DummyFakeCodeRunner` does not initiate ZMQ sockets and it overrides all methods of `AbstractCodeRunner`.
- we can mimic dummy agent's time-consuming steps by setting configuration file `dummy-agent.toml`. we can set the delay of each step by a tuple of 2 numbers that will convert to a range of random float or just a float that will convert to a delay time(sec).
- dummy agents do not run `sync_container_lifecycles`.

## How to test
1. make a config file to `<BACKENDAI_ROOT>/dummy-agent-config.toml` like below and set each value. You can find it from `<BACKENDAI_ROOT>/configs/agent/sample-dummy-config.toml`
```
[agent.delay]
scan-image = 0.1
pull-image = 0.1
destroy-kernel = 0.1
clean-kernel = 0.1
create-network = 0.1
destroy-network = 0.1


[agent.image]
already-have = []
need-to-pull = []
missing = []


[agent.resource.cpu]
core-indexes = [0, 1, 2, 3, 4]


[agent.resource.memory]
size = 34359738368


[kernel-creation-ctx.delay]
prepare-scratch = 0.1
prepare-ssh = 0.1
spawn = 0.1
start-container = 2.0
mount-krunner = 0.1


[kernel]
use-fake-code-runner = true


[kernel.delay]
check-status = 0.1
get-completions = 0.1
get-logs = 0.1
interrupt-kernel = 0.1
start-service = 0.1
start-model-service = 0.1
shutdown-service = 0.1
commit = 10.0
get-service-apps = 0.1
accept-file = 0.1
download-file = 0.1
download-single = 0.1
list-files = 0.1
```

2. Set backend config at `agent.toml` to `dummy`
```
[agent]
mode = "dummy" # or backend = "dummy"
rpc-listen-addr = { host = "127.0.0.1", port = 6011 }
...
```


.gitignore
```diff
@@ -111,6 +111,7 @@ ENV/
 /dev.etcd.volumes.json
 /dev.etcd.installed.json
 /env-*.sh
+/dummy-agent-config.toml
```
achimnol: I'd like to rename this to have the namespace first: `agent.dummy.toml`, with dots as the delimeter to indicate the subcomponent.

scripts/install-dev.sh
```diff
@@ -751,6 +751,9 @@ configure_backendai() {
     sed_inplace "s/# allow-compute-plugins =.*/allow-compute-plugins = []/" ./agent.toml
   fi
 
+  # configure agent
+  cp configs/agent/sample-dummy-config.toml ./dummy-agent-config.toml
```
achimnol: Same here.

src/ai/backend/agent/dummy/utils.py
achimnol, line 1: How about refactor this as a trafaret (like `manager/config.py` or `agent/config.py`), which converts the config value into a callable that does what `get_delay_from_cfg()` does.

achimnol (changes requested):
Just some minor requested changes. The overall impl looks fine.

src/ai/backend/agent/dummy/utils.py
fregataa: that is nice. I made `Delay` trafaret validator in `agent/dummy/config.py`. thank you!

src/ai/backend/agent/dummy/config.py
```diff
@@ -0,0 +1,102 @@
...
+
+from ai.backend.common import validators as tx
+
+DEFAULT_CONFIG_PATH = Path.cwd() / "agent.dummy.toml"
+
+RandomRange = t.Tuple(t.ToFloat, t.ToFloat)
+core_idx = {0, 1, 2, 3, 4}
+
+
+class Delay(t.Trafaret):
```
achimnol: Let's move this to `ai.backend.common.validators` like other "trafaret extensions" (`tx`) classes.

achimnol (changes requested):
A minor refactoring please!

achimnol (approved):

PR #1314, opened 2023-06-01
feature: ensure vfolder status when preparing mount

fregataa:
related issue https://github.com/lablup/backend.ai/issues/1427

As we ensure vfolder statuses before any actions about vfolders such as `get_info()` or `delete()`, let's ensure the status when enqueue a session and prepare mounts for it.


fregataa:
Close this PR because we have a plan to implement a new VFolder API handler and deprecate legacy functions

PR #1317, opened 2023-06-05
setup: Move pants' local-exec-root outside of the build root

achimnol:
This PR reduces the chances of retrying the test suite runs in macOS due to an
overflow of the filesystem change notification queue when the Pant's
`local_execution_root` is inside the build root. (refs pantsbuild/pants#19170)

It also improves the support Snap and OrbStack with the updated `scripts/check-docker.py` implementation,
which now generates the host-specific preferred `local_execution_root` path value.
`scripts/install-dev.sh` now stores it inside the `.pants.rc` file.

### Migration of existing setups

This PR does not auto-migrate existing setups. It is recommended to update `.pants.rc` by, especially if you are on macOS or using Snap-based Docker:

```shell
pants_local_exec_root=$(~/.cache/bai/bootstrap/cpython/3.11.3/bin/python3 scripts/check-docker.py --get-preferred-pants-local-exec-root)
mkdir -p "$pants_local_exec_root"
./py scripts/tomltool.py -f .pants.rc set 'GLOBAL.local_execution_root_dir' "$pants_local_exec_root"
```


achimnol:
Confirmed working on a fresh new Ubuntu 22.04 (linux aarch64) VM with Snap-based Docker.

achimnol:
TODO: could we update `.pants.rc` using TOML-aware parser and writer?

achimnol:
> TODO: could we update `.pants.rc` using TOML-aware parser and writer?

Done!

PR #1320, opened 2023-06-07
fix: Use a more sensible value of the warning threshold for max concurrent transactions

achimnol:
This PR changes the fixed value (8) of the max-concurrent txns warning threshold to a calculated value from the actual pool-size and max-overflow config values.

It also adds explicit sample configurations to `configs/manager/sample.toml` about the connection pool params.


adrysn (approved):
LGTM.

PR #1324, opened 2023-06-09
ci: Detect packages that do not have binary files

### LINKED ISSUE #1303: Add an action workflow to check binary wheel availability
Since we are delivering to air-gapped offline sites, it is important to check if all our dependencies are available as precompiled binaries.

Let's add an action workflow to automatically check whenever there are changes in `requirements.txt` and the lockfiles.  We could use the lockfiles generated pants to list up the wheel files and match their target platforms.

Note that the lockfiles include comments, we should use [`json5`](https://pypi.org/project/json5/) to parse them.

Yaminyam:
resolve #1303


.github/workflows/check-requiredment.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  check-requirement:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Check if a precompiled binary file exists in dependencies
+        run: |
+            cat python.lock | grep -v '^//' | jq -r '.locked_resolves[].locked_requirements[]
+            | select(any(.artifacts[].url; endswith(".whl")) | not) | .project_name' >> no-exist-binary.txt
```
achimnol: * `no-exist-binary.txt` looks like too Konglish...
  Let's rename it to `packages-missing-precompiled-wheels`.

* What if the target package is a pure Python package that provides only the source distribution?

achimnol (changes requested):

.github/workflows/check-requiredment.yml
kyujin-cho, line 1: You've meant to represent `requirement`, right?

kyujin-cho (changes requested):

.github/workflows/check-requiredment.yml
Yaminyam: Oh yes my mistake.

PR #1325, opened 2023-06-11
fix: ensure type of `agent_id` to be consistent

fregataa:
Currently variable `agent` has an inconsistent type in `SchedulerDispatcher._schedule_multi_node_session()`.
It can be `AgentRow` type or `AgentId`(string) type depending on the condition.
Let's check the type strictly.


kyujin-cho (approved):
LGTM!

PR #1332, opened 2023-06-17
fix: heartbeat failing when public-host is set

kyujin-cho:
This PR fixes heartbeat failing when public-host is set.


kyujin-cho:
> How about returning `str()` parsed value from `_get_public_host()`?

Since it's an issue about heartbeat sender side only accepting narrow range of types I think it is better to stay with current implementation.

achimnol (approved):

PR #1334, opened 2023-06-19
feat: upgrade miniling to filter and order by JSON column

fregataa:
upgrade miniling to filter and order by JSON column and enable to query by `scheduled_at` column from session and kernel tables.


Sujin-Kim1 (approved):
LGTM

adrysn (changes requested):
It may not be optimal to include a duplicate `scheduled_at` field since it's already present in `session_history.scheduled_at`. If feasible, implementing a feature to query sub-fields universally could be a more beneficial solution.

achimnol:
Since #1022 is merged, this PR requires update!

PR #1337, opened 2023-06-19
fix: add `agent_ids` fields to filter and order by agent id when listing sessions

fregataa:
~~blocked by: #1022~~

Enable to list sessions by `agent_id` of sibling kernels


achimnol:
Now merged #1022!

src/ai/backend/client/func/session.py
```diff
@@ -126,7 +126,8 @@ async def paginated_list(
                 "filter": (filter, "String"),
                 "order": (order, "String"),
             },
-            fields,
+            # fields,
+            [session_fields["session_id"], session_fields["agent_ids"]],
```
achimnol: This removes all the default fields in the `backend.ai ps` command! Please include `agent_ids` inside `_default_list_fields`.

achimnol (changes requested):
Mostly looks good, please check out the following minor fix.

src/ai/backend/client/func/session.py
fregataa: thanks! my mistake..

achimnol (approved):

Issue #1328, opened 2023-06-13
Limit the number of kernels that can be created within agents


fregataa:
Docker network specification limits the number of docker containers that can be created on an the network.
For example, if docker network subnet has 172.10.0.0/26 CIDR range, then only 62 IP addresses are available. But users still can create sessions and those new sessions can be assigned under the **full** network. Then docker raises error like below
```
returned error: no available IPv4 addresses on this network's address pools: bridge (...)
```

Let's do **hard limit** the number of kernels in agents when scheduler assign a session to agents.
Please suggest any good idea to prevent such problems.

PR #1338, opened 2023-06-19
feature: limit the number of containers per agents

fregataa:
resolves: #1328 
`agent` table has `max_container_count` column and `container_count` per agent is saved at redis.
check the container count when schedule session -right before assign agent for session-.
increase the count after schedule the session to agent successfully.
decrease the count after clean the session successfully.


fregataa:
need to discuss
Should the **heartbeat** check the number of containers in agent?
or something else to check it periodically?

achimnol:
I remember that we've discussed the implementation to use etcd-based "scoped" configurations. (global default + per-node & per-scaling-group overrides)

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -583,6 +587,20 @@ async def _update_session_status_data() -> None:
         if num_scheduled > 0:
             await self.event_producer.produce_event(DoPrepareEvent())
 
+    async def _filter_agent_by_container_limit(self, cand_agents: list[AgentRow]) -> list[AgentRow]:
```
achimnol: `cand_agents` -> `candidate_agents`

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -615,6 +633,17 @@ async def _schedule_single_node_session(
                         f"arch: {requested_architecture})"
                     ),
                 )
+            available_candidate_agents = await self._filter_agent_by_container_limit(
+                compatible_candidate_agents
+            )
+            if not available_candidate_agents:
+                raise InstanceNotAvailable(
+                    extra_msg=(
+                        "No agents found to be available because all agents have reached the limit"
```
achimnol: "reached the limit" -> "reached the hard limit"

achimnol (changes requested):
* We need to update the existing `recalc_resource_usage()`.
* The number of containers are the implementation detail of agents, and that's why I thought we should put the scoped/namespaced configs in the etcd so that agents can read their own limits directly.
* It is a good idea to preliminarily exclude "already-full" agents upon scheduling, but I think the count tracking should be done in the agents to make it more robust against agent failures and restarts.

fregataa:
> * The number of containers are the implementation detail of agents, and that's why I thought we should put the scoped/namespaced configs in the etcd so that agents can read their own limits directly.
> * It is a good idea to preliminarily exclude "already-full" agents upon scheduling, but I think the count tracking should be done in the agents to make it more robust against agent failures and restarts.

I move the container limit config  from db agent table to etcd. The config has `SGROUP` scope now. should I move it into `NODE` scope or both scope should have the config?
And now agents check and set the number of container to redis.

achimnol:
> > * The number of containers are the implementation detail of agents, and that's why I thought we should put the scoped/namespaced configs in the etcd so that agents can read their own limits directly.
> > * It is a good idea to preliminarily exclude "already-full" agents upon scheduling, but I think the count tracking should be done in the agents to make it more robust against agent failures and restarts.
> 
> I move the container limit config from db agent table to etcd. The config has `SGROUP` scope now. should I move it into `NODE` scope or both scope should have the config? And now agents check and set the number of container to redis.

Which scope to use for actual configuration is up to the user (cluster admin). As long as we read the etcd config through our `ai.backend.common.etcd` interface, it will be handled automatically.

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -583,6 +587,29 @@ async def _update_session_status_data() -> None:
         if num_scheduled > 0:
             await self.event_producer.produce_event(DoPrepareEvent())
 
+    async def _filter_agent_by_container_limit(
+        self, candidate_agents: list[AgentRow]
+    ) -> list[AgentRow]:
+        raw_value = await self.shared_config.etcd.get(
+            "config/agent/max-container-count", scope=ConfigScopes.SGROUP
+        )
```
achimnol: We don't have to explicit specify the scope when _querying_ the configuration if we query the etcd from the agents, because `common.etcd` will automatically compose the values from multiple scopes based on the scopes initialized upon the agent startup.

So, to keep this preliminary filtering logic in the manager side, the agents should also store the hard-limit values in the Redis so that this function could fetch them at once to filter out.

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -600,7 +627,7 @@ async def _schedule_single_node_session(
         requested_architectures = set(k.architecture for k in sess_ctx.kernels)
         if len(requested_architectures) > 1:
             raise GenericBadRequest(
-                "Cannot assign multiple kernels with different architectureon single node session",
+                "Cannot assign multiple kernels with different architecture on single node session",
```
achimnol: 👍🏼

achimnol (changes requested):

src/ai/backend/manager/scheduler/dispatcher.py
fregataa: agents store their current container count to Redis.
Should I also store hard-limit value to Redis for this implementation?

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -623,6 +626,27 @@ async def _update_session_status_data() -> None:
...
+            pipe = r.pipeline()
+            for ag in candidate_agents:
+                await pipe.get(f"container_count.{ag.id}")
+            return pipe
+
+        raw_counts = await redis_helper.execute(self.registry.redis_stat, _pipe_builder)
+        return [
+            ag
+            for ag, count in zip(candidate_agents, raw_counts)
+            if max_container_count > int(count)
```
achimnol: It would be safer to check if `count is not None` for cases when there are missing results when querying `container_count.{agent_id}` keys in Redis.

achimnol (changes requested):
A final comment.

src/ai/backend/manager/scheduler/dispatcher.py
fregataa: thanks, I forgot that the result from redis could be `None`.
And I add `container_count` field to gql object.

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -634,7 +634,7 @@ async def _pipe_builder(r: Redis) -> RedisPipeline:
         return [
             ag
             for ag, count in zip(candidate_agents, raw_counts)
-            if max_container_count > int(count)
+            if count is not None and max_container_count > int(count)
```
achimnol: Hm, I think we should include the agent when we cannot determine the current container count of it, as a fallback. Let's treat `None` as `0`.

achimnol (changes requested):

src/ai/backend/manager/scheduler/dispatcher.py
fregataa: thanks, I updated the code!

achimnol (approved):

PR #1340, opened 2023-06-20
docs: Renew versioning scheme document

hyoinandout:
Hello,

The previous versioning scheme document was pointing to an invalid link.
I changed the link to be valid.


achimnol (approved):
Thanks!

PR #1343, opened 2023-06-22
fix: `ComputeSessionList` GraphQL query raising 400

kyujin-cho:
This PR updates both `sessions.vfolder_mounts` and `kernels.vfolder_mounts` to match new VFolderID format.


fregataa:
let's update news fragment. This pr update VFolderID to handle null `quota_scope_id` field, so it would be good to mention about it.

src/ai/backend/manager/models/alembic/versions/9e599b62f6f1_update_kernels_vfolder_mounts_vfid_.py
```diff
@@ -0,0 +1,188 @@
...
+"""
+from uuid import UUID
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+from sqlalchemy.sql.expression import bindparam
+
+from ai.backend.common.types import VFolderMount
+from ai.backend.manager.models import SessionRow, kernels, vfolders
```
achimnol: It is better to avoid directly import the model definitions in alembic migration but manually define the table schema with only relevant columns.
If the database schema evolves in the future, these imports may not represent the state when this migration script is written.

achimnol (changes requested):

kyujin-cho:
We have a quota scope ID syntax to come, which is being discussed on #1354. Therefore delaying merge of this branch until we're done with the PR.

Issue #1118, opened 2023-02-27
Cannot delete sessions that failed to create while trying to download images larger than the remaining capacity


Yaminyam:
### What Operating System(s) are you seeing this problem on?

Linux (x86-64)

### Backend.AI version

22.09

### Describe the bug

Temporarily created sessions remain in the preparing state if image download fails.If you try to force a termination of that session, it stops in terminating state and does not disappear.

I think an exception is needed to remove temporary sessions when image down fails.

### To Reproduce

1. Download images larger than the remaining capacity.
1. Force termination of the session in the preparing state in the session list.
   ### Expected Behavior

_No response_

JIRA Issue: BA-219

Yaminyam:
This problem can occur not only when the download fails due to insufficient capacity, but also when there is a mismatch between the actual session and the db.

achimnol:
To resolve this issue, we need to get the remaining capacity of the filesystem actually used by the Docker runtime.  The Docker API exposes the Docker's data root directory via the `/info` API (`docker info`), and we can use [`os.statvfs()`](https://docs.python.org/3/library/os.html#os.statvfs) against the data root path.

The problem is that the data root directory may _not be on the same filesystem namespace_ where the agent runs.  Typical examples are Docker Desktop and OrbStack, which runs virtual machines to run the Docker runtime in macOS.

We could detect such cases by comparing the inode number of the host root directory mounted inside the container and the inode number read by the agent:

In native Linux:
```console
$ stat -c '%i' /
2
$ docker run --rm -it -v /:/host alpine:latest stat -c '%i' /host
2
```

In macOS with OrbStack:
```console
$ stat -f '%i' /  # note that the option name differs from GNU utils
2
$ docker run --rm -it -v /:/host alpine:latest stat -c '%i' /host
73
```

If we detect the difference, then we need to run `os.statvfs()` **inside** a container created via the Docker API, to get the actual data root's filesystem capacity inside the VM:

```shell
docker info --format=json \
| jq -r .DockerRootDir \
| docker run --rm -i --pid=host --privileged python:3.11-slim \
python -c 'import os,sys;os.system("pip install -q nsenter");'\
'from nsenter import Namespace;n=Namespace(1, "mnt");'\
'n.__enter__();'\
's=os.statvfs(sys.stdin.read().strip());print(f"{s.f_bfree * s.f_bsize / (2**30):.2f} GiB");'\
'n.__exit__(None,None,None)'
```
```
[...]

PR #1346, opened 2023-06-23
feature: limit image disk and impl pre unload img

### LINKED ISSUE #39: Population and removal of images in agents
Let's add an admin-only ability to pre-populate and remove specific Docker images in agents.

Things to consider:
 * What if there are hundreds of agents? How many agents could be pulling the image at the same time without impacting the internal networking bandwidth?
 * We need some mechanism to keep track of such long-running API operations and retrieve the results asynchronously.
 * While downloading an image, what if a kernel creation request that wants to use the specific image comes in? Or, what if the designated image is being removed upon kernel creation requests? Should it be an error or a re-download trigger?

Related:
 * lablup/backend.ai-manager#84
 * lablup/backend.ai-manager#110

fregataa:
related to #1118 
partially-resolve #39 (Phase 2)

1. hard-limit on docker image disk. Check **free disk of root** of agent before pull and raise `AgentError` if **free disk of root** is less than 50 GiB
2. implement preload and unload image


achimnol:
Let's focus on the production setups where the Backend.AI agent and the Docker runtime runs in the same host or VM. We still need to detect if the Docker runtime runs in a different host or VM by checking the mount of `/` and the difference of inode numbers.

Since the VM usecase is mostly for development setups, it is not urgent to implement the hack to run `os.statvfs()` in the same "mount" namespace to the PID 1's that the Docker runtime sees.

achimnol (changes requested):
It seems that the documentation build is broken due to changes in the `PreloadImage()` GraphQL mutation. Please check it out.

fregataa:
> It seems that the documentation build is broken due to changes in the `PreloadImage()` GraphQL mutation. Please check it out.

it was my mistake. `default` value of arguments of Graphene objects should be passed by `default_value`, not `default`

src/ai/backend/manager/models/image.py
```diff
@@ -676,8 +700,40 @@ async def mutate(
...
+        force: bool,
     ) -> PreloadImage:
-        return PreloadImage(ok=False, msg="Not implemented.", task_id=None)
+        log.info(
+            "preloading images ({0}) to agents ({1}) by API request",
+            ", ".join([img for img in references]),
+            ", ".join(target_agents),
+        )
+        ctx: GraphQueryContext = info.context
+        async with ctx.db.begin_session() as db_sess:
```
achimnol: Shouldn't we use the read-only session here?

src/ai/backend/agent/agent.py
```diff
@@ -1158,6 +1160,33 @@ async def inject_container_lifecycle_event(
             ),
         )
 
+    async def get_free_image_disk(self) -> int | float:
+        """
+        Get free disk for images.
+
+        TODO: get free disk of backend's data root.
+        """
+        stat = os.statvfs("/")
```
achimnol: Let's update here to use the Docker daemon's data root directory, and falls back to "/" if that directory does not exist in the view of the agent.

achimnol (changes requested):

src/ai/backend/manager/models/image.py
```diff
@@ -692,7 +692,7 @@ class Arguments:
 
     ok = graphene.Boolean()
     msg = graphene.String()
-    task_id = graphene.String()
+    tasks = graphene.JSONString()
```
achimnol: I think we should avoid using an arbitrarily structured `JSONString` field here. If it contains just the list of task IDs, it should be something like `graphene.List(graphene.String())`.

src/ai/backend/manager/models/image.py
```diff
@@ -780,12 +784,12 @@ async def mutate(
                 for ref in references
             ]
 
-        async def _unload_task(reporter: ProgressReporter) -> None:
-            for agent in agents:
-                await ctx.registry.remove_image(agent.id, agent.addr, images)
+        tasks = {}
```
achimnol: Please specify the types or structure of the value.

src/ai/backend/manager/registry.py
```diff
@@ -1346,25 +1346,25 @@ async def pull_image(
         agent_addr: str,
         image_maps: Sequence[dict[str, Any]],
         force: bool = False,
-    ) -> None:
+    ) -> dict[str, Any]:
```
achimnol: It would be better to use a TypedDict or attrs class to make it structured.

achimnol (changes requested):

src/ai/backend/agent/config.py
```diff
@@ -22,6 +22,11 @@
             t.Key("agent"): t.Dict(
                 {
                     tx.AliasedKey(["backend", "mode"]): tx.Enum(AgentBackend),
+                    t.Key(
+                        "check-free-image-disk",
```
achimnol: Let's rename this to "check-free-image-space".

src/ai/backend/agent/config.py
```diff
@@ -22,6 +22,11 @@
...
+                        default=True,
+                    ): t.Bool(),
+                    t.Key("backend-data-root", default="/var/lib/docker"): t.String(),
```
achimnol: Let's let the agent read this value from the Docker's info API on startup.

achimnol (changes requested):

src/ai/backend/agent/config.py
achimnol: Let's also add the configuration for the minimum required space.

src/ai/backend/agent/agent.py
```diff
@@ -268,8 +268,9 @@ async def generate_accelerator_mounts(
     ) -> List[MountInfo]:
         raise NotImplementedError
 
+    @staticmethod
     @abstractmethod
-    def resolve_krunner_filepath(self, filename) -> Path:
+    def resolve_krunner_filepath(filename) -> Path:
```
achimnol: If there is no good reason, I think it's better to keep this an instance method, to allow future possibilites of accessing self's states in other subclass implementations.

src/ai/backend/agent/docker/agent.py
achimnol, line 1: Let's just use the data root dir provided by the docker info API on startup and falls back to "/" if the directory is not present in the agent's view.
It would be too much complication to write a completely working "docker-in-VM" detector for now.

achimnol (changes requested):

src/ai/backend/manager/models/image.py
fregataa: I defined a new type for it.

fregataa:
Closed this PR because it is too old to maintain and we have to review the idea of image preload & unload

PR #1348, opened 2023-06-27
ci: Update lablup/sparse-checkout action

achimnol:
..to avoid the segfault in latest Git with "--filter=blob:none" option
with sparse checkouts

refs lablup/sparse-checkout#2

### DIFF
--- a/.github/workflows/timeline-check.yml
+++ b/.github/workflows/timeline-check.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Sparse-checkout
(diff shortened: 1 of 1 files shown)


achimnol:
Intentionally temporarily removed `skip:changelog` lable to trigger the timeline check action and test the sparse-checkout action.

achimnol:
Confirmed the new sparse-checkout@v2 works.

achimnol:
This PR is ultimately removed in favor of `actions/checkout@v3`'s official support of sparse checkout.  It also handles `--filter=blob:none` cleanly with a fixed git version.

PR #1349, opened 2023-06-27
fix: Update heartbeat-timeout in halfstack.toml

sanxiyn:
In #1223, heartbeat-timeout was updated, but not in halfstack.toml. As a result, agents are repeatedly lost in development installation installed with install-dev.sh. Remove the key and use the default value instead.


configs/manager/halfstack.toml
```diff
@@ -22,7 +22,6 @@ ssl-enabled = false
 #ssl-cert = "/etc/backend.ai/ssl/apiserver-fullchain.pem"   # env: BACKNED_SSL_CERT
 #ssl-privkey = "/etc/backend.ai/ssl/apiserver-privkey.pem"  # env: BACKNED_SSL_KEY
 # ipc-base-path = "/tmp/backend.ai/ipc"
-heartbeat-timeout = 10.0
```
inureyes, line 25: To let developers know the existence of the parameters, let's keep the parameter with 40s. default for now. :D

inureyes (changes requested):

configs/manager/halfstack.toml
sanxiyn: Done.

achimnol (approved):

achimnol:
Let's refactor and merge `halfstack.toml` into `sample.toml` in a follow-up PR.

PR #1354, opened 2023-06-28
feat: add GQL queries for quota scope

### LINKED ISSUE #1353: The get/set API for quota configs per quota scope
Let's add a simple GraphQL query and mutation to get/set the quota configs and usage for individual quota scopes on a storage host, by invoking the storage-proxy's manager-facing APIs.

```[tasklist]
### Tasks
- [ ] `query QuotaConfig(storage_host_name, quota_scope_id) -> { quota_config { hard_limit_bytes } quota_usage { usage_bytes usage_count } }` (potential responses: configured quota config, quota scope not created yet)
- [ ] `mutation SetQuotaConfig(storage_host_name, quota_scope_id, { hard_limit_bytes })` (potential responses: ok, quota-not-supported, failure, invalid-input, etc.)
- [ ] Enhance the error handling when vfolder upload fails in the middle because `write()` fails (at least we should warn the user with "please check the remaining quota" or ask whether to "delete the incomplete file")
```

If the client is a non-superadmin, the query handler must accept only the quota scope ID from the requesting user's UUID or the project UUIDs that the user has access to.
The mutation should be allowed to superadmin or domain-admins only.

Q. Do we need to add `query QuotaScopeList(storage_host_name)` as a paginated list (for admin-oriented storage-centric page)?

kyujin-cho:
resolves #1353.  
This PR implements folderQuota query, updateFolderQuota/clearFolderQuota queries (both for GQL) and introduces new resource policies which will be used to constrain per-user and per-project resource limitations.


achimnol (approved):

achimnol:
Quota configurations in user/project resource policies are "templates" (i.e., default value when creating a new user or project). **The actual quota configurations are per user & project (and thus per quota scope).** There is no need to iterate over all users/projects affected by a specific policy when updating the quota config in a resource policy.

PR #1355, opened 2023-07-01
fix: prevent scanning every sub-directories in listing vfolder files when `recursive` is `False`

adrysn:
The File Listing API scans all sub-directories under a data folder in the current implementation, even when the 'recursive' option is off.

For example, for a data folder which contains `CI` directory only at the root,
image

the file listing API returns every file under the data folder by scanning all of the sub-directories.
image

This PR prevents this by not scanning every sub-directory when the `recursive` option is off. With this PR, only the `CI` directory is shown like image below.
image

### DIFF
--- a/src/ai/backend/storage/vfs/__init__.py
+++ b/src/ai/backend/storage/vfs/__init__.py
@@ -254,8 +254,10 @@ class BaseFSOpModel(AbstractFSOpModel):
                         if entry.is_dir() and not entry.is_symlink():
                             next_paths.append(Path(entry.path))
                         count += 1
-                        if recursive and limit > 0 and count == limit:
+                        if limit > 0 and count == limit:
                             break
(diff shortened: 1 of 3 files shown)


src/ai/backend/storage/vfs/__init__.py
```diff
@@ -254,8 +254,10 @@ def _scandir(target_path: Path, q: janus._SyncQueueProxy[Sentinel | DirEntry]) -
                         if entry.is_dir() and not entry.is_symlink():
                             next_paths.append(Path(entry.path))
```
achimnol, line 257: I think `break if not recursive` should come here, because the current implementation depends on the order that `os.scandir()` returns the directory entries.

achimnol (changes requested):

PR #1360, opened 2023-07-02
fix: `Keypair` graphene query mismatch between count and slice

fregataa:
If any user does not *associate* with any group, the value of `count` and length of `slice` from `Keypair` graphene are different.
let's apply `DISTINCT` to count query to filter out duplicated rows.

### DIFF
--- a/src/ai/backend/manager/models/keypair.py
+++ b/src/ai/backend/manager/models/keypair.py
@@ -214,7 +214,7 @@ class KeyPair(graphene.ObjectType):
             user=row["user"],
             ssh_public_key=row["ssh_public_key"],
             concurrency_limit=0,  # deprecated
-            projects=row["groups_name"],
+            projects=row["groups_name"] if "groups_name" in row.keys() else [],
         )
 
     async def resolve_num_queries(self, info: graphene.ResolveInfo) -> int:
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/models/keypair.py
```diff
@@ -325,11 +325,15 @@ async def load_count(
...
-            sa.join(keypairs, users, keypairs.c.user == users.c.uuid)
-            .join(association_groups_users, users.c.uuid == association_groups_users.c.user_id)
-            .join(groups, association_groups_users.c.group_id == groups.c.id)
+            sa.join(keypairs, users, keypairs.c.user == users.c.uuid, isouter=True)
+            .join(
+                association_groups_users,
+                users.c.uuid == association_groups_users.c.user_id,
+                isouter=True,
+            )
+            .join(groups, association_groups_users.c.group_id == groups.c.id, isouter=True)
```
kyujin-cho: We should never OUTER JOIN three tables in any case.

kyujin-cho (changes requested):

achimnol (changes requested):
image
In my setup, the PR seems not to work as expected...

kyujin-cho:
This should be resolve at 23.09.5.

fregataa:
@achimnol I resolved the issue that unable to fetch keypairs whose user is not associate with any project.
But I still cannot reproduce the login issue. it just work when such non-associate users exist

image

fregataa:

PR #1370, opened 2023-07-03
setup: Vendor some old source-only dependencies

achimnol:
- Vendor asyncudp from https://github.com/eerimoq/asyncudp @ main (62b859b)
- Vendor temporenc from https://github.com/temporenc/temporenc-python @ master (376ca06)
- setup: Use local requirements to refer the prebuilt vendored dependencies


achimnol:
Abandoning this approach in favor of #1374.

PR #1372, opened 2023-07-04
refactor: decouple model and domain layers

fregataa:
blocked by #1817 , because `DBContext` will contain DB connection rather than DB engine.

1. Implement **domain layer** using pydantic. the domain layer works as internal API
2. Replace session/kernel status management from raw SQLAlchemy query to internal API.


fregataa:
Closed this PR because we have a plan to apply service-repository layer architecture

PR #1191, opened 2023-03-24
feat: basic support for VFolder v3

achimnol:
refs lablup/backend.ai-internal#10

### <samp>🤖 Generated by Copilot at a14abe9</samp>

This pull request implements a new quota model for virtual folders and refactors the storage volume and API classes to support it. It also simplifies some configuration variables and renames and deletes some scripts related to the storage proxy and ceph clusters.

### <samp>🤖 Generated by Copilot at a14abe9</samp>

[...]

PR #1376, opened 2023-07-05
fix: storage proxy reporting wrong stat on some backends

kyujin-cho:
Follow-up PR of #1191. Fixes lablup/giftbox#439. Updates implementation of `get_fs_usage()` in GPFS and Weka backend to explicitly mention name of the attr parameters, since without it the order matters when initiating `CapacitiyUsage()` object.


achimnol (approved):
👍🏼

PR #1378, opened 2023-07-06
fix: add `images` column to sessions and enable to order or filter by image

fregataa:
Add `images` column to filter or order by image when list sessions.
`images` column is ARRAY type and the first component of `images` column is main kernel's image.
graphene will return `image` when we query sessions, which is the image of main kernel.


Sujin-Kim1 (changes requested):
image

@fregataa When I tested, I got this error.

fregataa:
> image
> @fregataa When I tested, I got this error.

updated. please re-review it!

adrysn (approved):
LGTM.
image

kyujin-cho (approved):

PR #1385, opened 2023-07-10
fix: Avoid NaN values in stat metrics

achimnol:
Sometimes when an agent is restarted, the first measured capacity value of the CPU utilization may become `NaN` because it requires an interval with two data points to calculate. Let's return zeros in such cases to avoid calculation errors.


adrysn (approved):
LGTM.

Issue #1384, opened 2023-07-10
Implement new kernel scheduling algorithm for model services


kyujin-cho:
### Main idea

Follow-up issue of #1278. As we now support servicing models on top of Backend.AI architecture, we also need to take care of the availability of each service API. Allowing multiple sessions to be spawned under single service (which is already implemented in #1278) is a small effort to achieve that objective. With this feature Backend.AI can now handle some random containers being killed for unknown reason.    
But this is not enough for system operators to claim API users for minimal downtime of the service, since there still is a chance of the physical node going down. In this case, when all sessions forming of a single service are created and running on a malfunctioning physical node, the whole service will go down.    
To eliminate this possibility we can think of adding new scheduling algorithm which spreads all sessions which belongs to a single service evenly to the whole cluster. By letting inference workloads use this strategy, even if some of the nodes go down, system operators still will be able to serve inference traffics.

PR #1394, opened 2023-07-13
feat: Add more agent selection scheduling strategies

jopemachine:
This PR relates to https://github.com/lablup/backend.ai/issues/1384#event-9770189002.

Discussion: https://teams.microsoft.com/l/message/19:3c0acc0825024daeb2211f55c4e7f4aa@thread.skype/1689215390841?tenantId=13c6a44d-9b52-4b9e-aa34-0513ee7131f2&groupId=74ae2c4d-ec4d-4fdf-b2c2-f5041d1e8631&parentMessageId=1689215390841&teamName=devops&channelName=Backend.AI%20Talks&createdTime=1689215390841

1. In the existing logic, the manager process should select the agent that has the most `remaining slots`, not `available_slots` which are fixed values.

2. `agent_selection_strategy` flag in `scheduler_opts` decides a strategy agent selection logic when pending sessions scheduling


src/ai/backend/common/types.py
```diff
@@ -1125,3 +1128,9 @@ class AcceleratorMetadata(TypedDict):
     display_unit: str
     number_format: AcceleratorNumberFormat
     display_icon: str
+
+
+class AgentSelectionStrategy(enum.StrEnum):
+    DISPERSED = "dispersed"
+    CONCENTRATED = "concentrated"
+    LEGACY = "legacy"
```
fregataa, line 1137: How about leaving any hint comment for this `LEGACY` strategy?

src/ai/backend/manager/models/scaling_group.py
```diff
@@ -89,6 +91,9 @@ def as_trafaret(cls) -> t.Trafaret:
                 t.Key("pending_timeout", default=0): tx.TimeDuration(allow_negative=False),
                 # Each scheduler impl refers an additional "config" key.
                 t.Key("config", default={}): t.Mapping(t.String, t.Any),
+                t.Key("agent_selection_strategy", default=AgentSelectionStrategy.DISPERSED): t.Enum(
+                    *[e.value for e in AgentSelectionStrategy]
+                ),
```
fregataa:
```suggestion
                t.Key("agent_selection_strategy", default=AgentSelectionStrategy.DISPERSED): tx.Enum(
                    AgentSelectionStrategy
                ),
```

we have [customized Enum trafaret type](https://github.com/lablup/backend.ai/blob/main/src/ai/backend/common/validators.py#L212). Let's use this!

src/ai/backend/manager/scheduler/drf.py
```diff
@@ -17,6 +23,22 @@
...
+    comparator = None
+    match agent_selection_strategy:
+        case AgentSelectionStrategy.LEGACY:
+            comparator = agent.available_slots - agent.occupied_slots
+        case AgentSelectionStrategy.CONCENTRATED:
+            comparator = -(agent.available_slots - agent.occupied_slots)
+        case AgentSelectionStrategy.DISPERSED | _:
+            comparator = agent.available_slots - agent.occupied_slots
+
+    return comparator
```
fregataa: 👍

fregataa (changes requested):
It is great in overall!
I left some comments.

src/ai/backend/common/types.py
achimnol: Note: `LEGACY` chooses the largest agent (the sort key is a tuple of resource slots).

Currently there is no explicit slot preference list to create a sort key but only 
moves "strict" less/greater items in the agent list.

Let's also add an explicit configuration like [the agent's `allocation-order` config](https://github.com/lablup/backend.ai/blob/6da46120/configs/agent/sample.toml#L261) and [the sorting mechanism](https://github.com/lablup/backend.ai/blob/6da46120/src/ai/backend/agent/agent.py#L1671-L1707).

src/ai/backend/manager/scheduler/drf.py
```diff
@@ -17,6 +24,47 @@
...
+        return agent_selection_order.index(slotname)
+    except ValueError:
+        return sys.maxsize
+
+
+def key_by_requested_slots(
+    agent: AgentRow,
+    agent_selection_strategy: AgentSelectionStrategy,
+    agent_selection_order: list[str],
+) -> list[int]:
```
fregataa: I guess the return value of this function is a list of resource remaining or available in total.
Then, the return value should be `list[Decimal]` type. Please let me know whether I guess it correctly!

jopemachine: Thanks for the correction! 
I think using Decimal would be more proper here.
and I checked the ResourcePolicy keys' types like `cpu`, `mem` are Decimal.
So, let's change these return types using `Decimal`

src/ai/backend/manager/scheduler/fifo.py
```diff
@@ -34,27 +34,34 @@ def key_by_requested_slots(
         if k in unused_slot_keys and v > Decimal(0):
             num_extras += 1
 
-    sorted_agent_selection_order = sorted(
-        requested_slots.data.keys(), key=lambda item: get_slot_index(item, agent_selection_order)
+    for requested_slot_key in sorted(requested_slots.data.keys(), reverse=True):
+        first = requested_slot_key.split(".")[0]
```
achimnol: Conventionally we have called this string as `device_name`.
Let's follow the same convention.
(ref: https://github.com/lablup/backend.ai/blob/08c81ca48/src/ai/backend/agent/alloc_map.py#L127)

src/ai/backend/manager/scheduler/drf.py
```diff
@@ -24,21 +24,32 @@
...
     agent_selection_strategy: AgentSelectionStrategy,
-    agent_selection_order: list[str],
-) -> list[int]:
-    sorted_agent_selection_order = sorted(
-        agent.available_slots.data.keys(),
-        key=lambda item: get_slot_index(item, agent_selection_order),
+    agent_selection_resource_priority: list[str],
+) -> list[Decimal]:
+    for requested_slot_key in sorted(requested_slots.data.keys(), reverse=True):
+        first = requested_slot_key.split(".")[0]
```
achimnol: Same here!

achimnol (changes requested):
Almost there!

achimnol (changes requested):
When I tried to run a hello-world with this PR, the manager crashes:
image

I see there is no dastabase migrations required.
Could we make a fallback for existing setups to continue working after applying this PR?

jopemachine:
> When I tried to run a hello-world with this PR, the manager crashes: image
> 
> I see there are no dastabase migrations required. Could we make a fallback for existing setups to continue working after applying this PR?

I'm sorry for forgetting to add `agent-selection-resource-priority` to the `config.py`.

I think the default value of the `agent-selection-resource-priority` specified in the `config.py` will handle in case of  `agent-selection-resource-priority` is missing (default fallback)

achimnol:
> I think the default value of the `agent-selection-resource-priority` specified in the `config.py` will handle in case of `agent-selection-resource-priority` is missing (default fallback)

Yes, that's correct!

achimnol (approved):

PR #1395, opened 2023-07-13
feat: add aliased key of `session_name`

pderer:
When request creating session API, response notes that "session_name" is required.

MicrosoftTeams-image (1)

> 

but when request parameter contain "session_name", response notes that "session_name" is not allowed key.

MicrosoftTeams-image (2)

To resolve this ambiguity, session_name is added to the aliased key of session_name.

So, request contained "session_name" is allowed.

### DIFF
--- a/src/ai/backend/manager/api/service.py
+++ b/src/ai/backend/manager/api/service.py
@@ -163,7 +163,7 @@ async def get_info(request: web.Request) -> web.Response:
 @check_api_params(
     t.Dict(
         {
-            tx.AliasedKey(["name", "clientSessionToken"])
+            tx.AliasedKey(["name", "session_name", "clientSessionToken"])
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/api/service.py
```diff
@@ -163,7 +163,7 @@ async def get_info(request: web.Request) -> web.Response:
 @check_api_params(
     t.Dict(
         {
-            tx.AliasedKey(["name", "clientSessionToken"])
+            tx.AliasedKey(["name", "session_name", "clientSessionToken"])
```
kyujin-cho: This should be `service_name`.
```suggestion
            tx.AliasedKey(["name", "service_name", "clientSessionToken"])
```

kyujin-cho (changes requested):

src/ai/backend/manager/api/service.py
pderer: Thanks to reviewing my mistake.

PR #1397, opened 2023-07-14
fix: avoid checking `max_vfolder_count`

### LINKED ISSUE #1391: Make it available to create a project folder regardless of `max_vfolder_count`.

macOS (Apple Silicon)

### Backend.AI version

23.03

### Describe the bug

If a number of created user vfolder is the same as `max_vfolder_count`, the admin cannot create a new project folder.
The number of project folder limit should be related to the `ProjectResourcePolicy` not `KeypairResourcePolicy`.

So let's make it available to create a project folder regardless of `max_vfolder_count`. 
If possible, let's add `max_vfolder_count` to the `ProjectResourcePolicy` and limit the number of creating project folders using that.

### To Reproduce

Create as many user folders as `max_vfolder_count` and create a project folder. A new project folder cannot be created, even though it must be created regardless of `max_vfolder_count`.

### Expected Behavior

Project folder should be created.

### Anything else?

_No response_

pderer:
This PR resolves #1391.


Sujin-Kim1 (changes requested):
`max_vfolder_count` should block only "PROJECT" folder.

pderer:
> The problem is keypair `max_vfolder_count` affect project folder creation.

I fix it in create vfolder function, but there is a bug in clone function too.
Decision is that clone function return `user` ownership type, and admin can clone `project` ownership type regardless of `max_vfolder_count`.
I may fix it in this PR.
But if I modify the code too much, I will create a new issue.

pderer:
I made a new PR because the added issue is focused on the clone function. (#1400)

achimnol:
It seems that I have accidentally closed this PR in favor of #1400 as it described that #1400 includes #1397, but the actual code diff does *not*.

PR #1400, opened 2023-07-17
fix: cloning vfolder bugs

pderer:
There are several bugs when I try to resolve #1397.

- There was code that creating vfolder request before cloning vfolder request.
    - So request cloning vfolder, makes redundant folder and makes 500 server error(FileExistedError)
    - Result is cloned vfolder is `READY` and cloning vfolder is `CLONING`forever.
    - But response is 201, so It is ambiguous.
- Parameter sequence problem when defining VfolderCloneInfo instance.
- From #1397, there is `max_vfolder_count`problem in cloning too.

This PR resolves these bugs.


changes/1400.fix.md
```diff
@@ -0,0 +1,4 @@
+Fixing cloning vfolder bugs
+* Removed creating vfolder request code in cloning vfolder function.
+* Changed sequence of parameter when defining VFolderCloneInfo class.
+* Avoided check `max_vfolder_count` when admin request cloning project type vfolder.
```
fregataa: We write news fragment file in one line without any "item" prefix. Please refer other merged PRs news fragment.

src/ai/backend/manager/api/vfolder.py
```diff
@@ -2424,6 +2424,15 @@ async def clone(request: web.Request, params: Any, row: VFolderRow) -> web.Respo
...
 
+    if row["group"] is not None:
+        ownership_type = "group"
+        if not request["is_admin"]:
+            raise GenericForbidden("no permission")
+    else:
+        ownership_type = "user"
+    if ownership_type not in allowed_vfolder_types:
+        raise InvalidAPIParameters(f"{ownership_type}-owned vfolder is not allowed in this cluster")
+
```
fregataa: We already handle `allowed_vfolder_types` in `query_accessible_vfolders()` method.
As you can see in `query_accessible_vfolders()`, we assume that the system has ownership type `user` as a default value.
So how about removing this part?

fregataa (changes requested):
Thanks!
I left some comments.

changes/1400.fix.md
pderer: Thanks! I will fix it.

src/ai/backend/manager/api/vfolder.py
pderer: Thanks! I will remove code about allowed_vfolder_types.
And I have a question about checking admin code.
When creating a vfolder, if user is not an admin, user cannot create a project vfolder, so there is no need to check whether user is an admin or not in the clone function, right?
Or is it correct to explicitly check whether or not user is an admin for any unknown risks?

fregataa: As long as it is unable to clone a vfolder as a project folder, I think we dont need to double check it.

pderer: Thanks to reviewing!
I will remove that code while keeping it working.

fregataa (approved):
LGTM!

achimnol (approved):

PR #1405, opened 2023-07-18
feat: roundrobin agent selection strategy

jopemachine:
This PR relates to https://github.com/lablup/backend.ai/issues/1384#event-9770189002.

Added `roundrobin` option for enforce manager process to disperse all pending sessions across agents.


src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -114,6 +118,11 @@
 _key_schedule_prep_tasks: Final = "scheduler.preptasks"
 
 
+def get_hash(value: str) -> str:
+    result = hashlib.md5(value.encode())
+    return result.hexdigest()
+
```
fregataa: this function is general but seems to be used in a very specific way.
how about specialize this function or make another function? such as `get_schedulable_group_id()`?

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -619,24 +628,95 @@ async def _schedule_single_node_session(
...
+                if scheduler.sgroup_opts.roundrobin:
+                    rr_state_str: str | None = await sched_ctx.registry.shared_config.get_raw(
+                        "roundrobin_state"
                     )
-                assert cand_agent is not None
-                agent_id = cand_agent
+
+                    if rr_state_str is not None:
+                        rr_state: Dict[str, Any] = json.loads(rr_state_str)
```
fregataa:
```suggestion
                        rr_state: dict[str, Any] = json.loads(rr_state_str)
```
As long as [typing.Dict](https://docs.python.org/3/library/typing.html#typing.Dict) is deprecated since Python 3.9, let's just use `dict`!

fregataa: Plus, it would be nice to show how `rr_state` looks like.
1. We can leave comments like below.
```python
# rr_state
# {
#     "architecture-1": {"schedulable_group_id": <ID>, "next_index": <INDEX>}
# }
```
2. Or, we can define a customized type like below.
```python
class RoundRobinState(TypedDict):
    schedulable_group_id: str
    next_index: int

RoundRobinStateDict: TypeAlias = dict[str, RoundRobinState]
```
Then we do not need to chain `cast`!

fregataa: Let's handle `JSONDecodeError` exception in case `rr_state_str` is a wrong formatted string.

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -619,24 +628,95 @@ async def _schedule_single_node_session(
...
+
+                            if schedulable_group_id == cast(
+                                Dict[str, Any], rr_state.get(requested_architecture)
+                            ).get("schedulable_group_id"):
+                                rr_index = cast(
+                                    int,
+                                    cast(Dict[str, Any], rr_state.get(requested_architecture)).get(
+                                        "next_index"
+                                    ),
+                                )
```
fregataa: This is just my opinion.
It is not bad to handle the branches through chaining methods. It looks like functional programming.
But I am not sure it is good in our code base cause I have never seen such `cast` chaining and I am worried about readability.
I just personally like to use walrus operator in such case.
```python
if (rr_state_data := rr_state.get(requested_architecture)) is not None:
    sg_id = rr_state_data.get("schedulable_group_id")
```
This is just my opinion, so please do not feel pressure to change this code. Let's talk about it later!

fregataa (changes requested):
Good! I left some comments.
And I have one question about round robin.
Is the `roundrobin` option orthogonal to [AgentSelectionStrategy.DISPERSED](https://github.com/lablup/backend.ai/pull/1394/files#diff-36ffd7bf35355ed19520f4bcce9eec8be45be53f280a0c43adfd198d85737d26R1133-R1136)?

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -619,24 +628,95 @@ async def _schedule_single_node_session(
...
-                        extra_msg=(
-                            "Could not find a contiguous resource region in any agent "
-                            "big enough to host the session "
-                            f"(id: {sess_ctx.id}, resource group: {sess_ctx.scaling_group_name})"
-                        ),
+                sorted_agents = sorted(compatible_candidate_agents, key=lambda agent: agent.id)
+
+                if scheduler.sgroup_opts.roundrobin:
+                    rr_state_str: str | None = await sched_ctx.registry.shared_config.get_raw(
+                        "roundrobin_state"
```
fregataa: In case of defining customized type, we can define new method in `shared_config`, such as `get_round_robin_state()` and then cast(or parse) the string into `dict` or customized type.

src/ai/backend/manager/config.py
```diff
@@ -416,6 +416,7 @@
             ).allow_extra("*")
             | t.Null
         ),
+        t.Key("roundrobin_state", default=None): t.Null | t.String,
```
fregataa: In case of defining customized type, we can validate the input string here. Then we do not even need to handle `JSONDecodeError` when load json from this value.

src/ai/backend/common/validators.py
```diff
@@ -696,3 +698,26 @@ def check_and_return(self, value: Any) -> float:
...
+
+class RoundRobinStatesJSONString(t.Trafaret):
+    def check_and_return(self, value: Any) -> RoundRobinStates:
+        if not isinstance(value, str):
+            self._failure("RoundRobinStatesJSONString must be JSON string", value=value)
+        else:
+            try:
+                rr_states_dict: dict[str, dict[str, Any]] = json.loads(value)
+            except (KeyError, ValueError):
+                self._failure("RoundRobinStatesJSONString is not a valid JSON string", value=value)
```
fregataa:
```suggestion
        try:
            rr_states_dict: dict[str, dict[str, Any]] = json.loads(value)
        except (KeyError, ValueError, json.decoder.JSONDecodeError):
            self._failure(f"Expected valid JSON string, got `{value}`. RoundRobinStatesJSONString should be a valid JSON string", value=value)
```
I think that we can *merge* the if state and try-catch state like this!

src/ai/backend/common/validators.py
```diff
@@ -696,3 +698,26 @@ def check_and_return(self, value: Any) -> float:
...
+
+            rr_states = {}
+            for arch, rr_state_dict in rr_states_dict.items():
+                if "next_index" not in rr_state_dict or "schedulable_group_id" not in rr_state_dict:
+                    self._failure("Invalid roundrobin states")
+
+                rr_states[arch] = RoundRobinState(
+                    schedulable_group_id=cast(str, rr_state_dict.get("schedulable_group_id")),
+                    next_index=cast(int, rr_state_dict.get("next_index")),
+                )
```
fregataa: I have one idea about this part.
if `RoundRobinState` inherits `JSONSerializableMixin`, then `RoundRobinState` can validate by itself.
```python
for arch, rr_state_dict in rr_states_dict.items():
    rr_states[arch] = RoundRobinState.from_json(rr_state_dict)
```
How do you think of it?

fregataa (commented):
Good!
I left some comments.

fregataa (commented):
Looks good in overall.
Does Round-robin work only for single node session?

jopemachine:
Yes I thought it could be too complicated to handle this feature on the multi node session.

fregataa (approved):
👍 good!

fregataa (changes requested):
Oops, please resolve the conflict and re-request issue!

jopemachine:
@fregataa I merged the PR with the main branch, and tested it again. It seems it still works as expected.

스크린샷 2023-09-05 오전 10 45 55

kyujin-cho (approved):

Issue #1398, opened 2023-07-14
Let's prevent the creation(or clone) of vfolder with duplicate names on different hosts


fregataa:
Currently, it is blocked to create(or clone) vfolders with duplicate names but only when the `hosts` are same.
Let's block to create vfolders with duplicate names even on different `hosts`.

PR #1407, opened 2023-07-21
fix: delete vfolder conditions of checking host while creating/cloning vfolders

NayeonKeum:
resolves: #1398 

**Done**

Originally `when creating/cloning vfolders` we `checked if two conditions are the same`, it's `name` and `host`.

To block creating/cloning vfolders with duplicate names even on different hosts, I `deleted a condition that checked whether if the host is same or not`. (Only conditions of checking names remains).

This will gradually make it impossible to create vfolder with names, whatever the host is.

**Test**

I executed postgres querys directly into it's docker container and checked if this works.

### DIFF
--- a/changes/1398.fix.md
+++ b/changes/1398.fix.md
@@ -1,1 +1,1 @@
-Prevent creating/cloneing vfolders with duplicate names on different hosts by deleting conditions checking host.
+Prevent creating/cloning vfolders with duplicate names on different hosts by deleting conditions checking host.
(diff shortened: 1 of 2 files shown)


fregataa (approved):
LGTM!
Let's wait for other's review

changes/1398.fix.md
```diff
@@ -0,0 +1 @@
+Prevent creating/cloneing vfolders with duplicate names on different hosts by deleting conditions checking host.
```
achimnol:
```suggestion
Prevent creating/cloning vfolders with duplicate names on different hosts by deleting conditions checking host.
```

achimnol (changes requested):
Just a typo. :wink:

achimnol (approved):

PR #1412, opened 2023-07-23
fix: re-map session status transition

fregataa:
1. Enable transit session status from `PULLING` to `CANCELLED` or `TERMINATED`
2. Remove `not in` syntax for better readability

### DIFF
--- a/src/ai/backend/manager/models/session.py
+++ b/src/ai/backend/manager/models/session.py
@@ -234,7 +234,6 @@ SESSION_STATUS_TRANSITION_MAP: Mapping[SessionStatus, set[SessionStatus]] = {
     SessionStatus.PREPARING: {
         SessionStatus.PULLING,
         SessionStatus.RUNNING,
-        SessionStatus.RESTARTING,
         SessionStatus.RUNNING_DEGRADED,
         SessionStatus.TERMINATING,
         SessionStatus.TERMINATED,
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/models/session.py
```diff
@@ -208,43 +208,37 @@ class SessionStatus(enum.Enum):
...
-        not in (
-            SessionStatus.PREPARING,
-            SessionStatus.PENDING,
-            SessionStatus.SCHEDULED,
-            SessionStatus.TERMINATED,
-            SessionStatus.CANCELLED,
-        )
+        SessionStatus.PULLING,
+        SessionStatus.RUNNING,
+        SessionStatus.RESTARTING,
```
achimnol: RESSTARTING should be removed from here because RESTARTING may occur only with RUNNING sessions.

src/ai/backend/manager/models/session.py
```diff
@@ -253,31 +247,21 @@ class SessionStatus(enum.Enum):
...
-        )
+        SessionStatus.RUNNING,
+        SessionStatus.TERMINATING,
+        SessionStatus.TERMINATED,
+        SessionStatus.ERROR,
     },
     SessionStatus.TERMINATING: {SessionStatus.TERMINATED, SessionStatus.ERROR},
     SessionStatus.TERMINATED: set(),
-    SessionStatus.ERROR: set(),
+    SessionStatus.ERROR: {SessionStatus.TERMINATED},
```
achimnol, line 265: Good catch!

achimnol (changes requested):
Minor fix!

src/ai/backend/manager/models/session.py
fregataa: Okay, I removed this part!

achimnol (approved):

PR #1417, opened 2023-07-25
feat: add `max_vfolder_count` to `ProjectResourcePolicy` and migrate same option to `UserResourcePolicy`

### LINKED ISSUE #1392: Add `max_vfolder_count` to `ProjectResourcePolicy` and migrate the same option to `UserResourcePolicy`
Let's add `max_vfolder_count` limit option to `ProjectResourecPolicy` added in #1354, following the implementation of `KeyPairResourcePolicy`.

Also, we need to migrate it from `KeyPairResourcePolicy` to `UserResourcePolicy` **because vfolders are owned by users, not keypairs.**  In the aspect, this issue supersedes #384.

Let's deprecate `max_vfolder_size` options in resource policies as we now use proper per-user/per-project quota implementation _instead of_ per-vfolder size limits.

* Update the database schema with alembic migrations:
  - [x] Add `max_vfolder_count` column to both the `project_resource_policies`, `user_resource_policies` tables, with the default of "unlimited".
  - [x] Remove `max_vfolder_size` and `max_vfolder_count` columns in the `keypair_resource_policies` table.
    - Also remove `max_vfolder_size` and `max_vfolder_count` CLI options in `keypair-resource-policy`.
  - [x] Rename `max_vfolder_size` columns in both `project_resource_policies` and `user_resource_policies` table to `max_quota_scope_size`.
    - FYI: In the future, `max_quota_scope_file_count` may be added as well.
    - Let's preserve `max_vfolder_size` as an alias in GraphQL queries for backward compatibility.
* [x] Update the related GraphQL queries and mutations to include the above fields.
* [x] Update the vfolder creation logic to apply the new options appropriately, as a follow-up of #1391.
* [x] Update the vfolder clone logic to apply the new options appropriately, as a follow-up of #1400.

pderer:
This PR resolves #1392.

Since I wrote the migration script, you need to do the alembic upgrade.


src/ai/backend/manager/api/vfolder.py
```diff
@@ -2422,6 +2429,28 @@ async def clone(request: web.Request, params: Any, row: VFolderRow) -> web.Respo
     if source_proxy_name != target_proxy_name:
         raise InvalidAPIParameters("proxy name of source and target vfolders must be equal.")
 
+    async with root_ctx.db.begin_session() as sess:
+        if row["group"]:
+            log.debug("row[group]:{}", row["group"])
```
fregataa: Let's improve this logging.
Such as `VFolder group: {}`, or we can log it better!

src/ai/backend/manager/api/vfolder.py
```diff
@@ -2422,6 +2429,28 @@ async def clone(request: web.Request, params: Any, row: VFolderRow) -> web.Respo
...
+            query = (
+                sa.select(GroupRow)
+                .where((GroupRow.domain_name == domain_name) & (GroupRow.id == row["group"]))
+                .options(selectinload(GroupRow.resource_policy_row))
```
fregataa: Just one little tip.
If you want to load only `max_vfolder_count` from this relationship, you can use `load_only` option.
https://github.com/lablup/backend.ai/blob/main/src/ai/backend/manager/models/session.py#L766-L768

fregataa (commented):
Good!
Thank you for your commitment.
I left a few comments that are not critical. Please check them and re-request review

src/ai/backend/manager/api/vfolder.py
pderer: Thanks!
I will fix it.

src/ai/backend/manager/api/vfolder.py
pderer: Thank you for helping me study ORM.
I will fix it.

fregataa (approved):
👍

kyujin-cho (approved):

PR #1418, opened 2023-07-25
fix: Ensure sudo for docker socket accessing scripts (check-docker.py)

achimnol:
- fix: Run `make compile` for the first time when installing the webui
  as editable


achimnol:
Confirmed working on a fresh new machine.

PR #1419, opened 2023-07-25
feat: Update the open-source version of CUDA plugin for CUDA v12

achimnol:
- feat: Allow using CUDA 12 series and add missing repr of device objects
- doc: Add news fragment


achimnol:
CUDA v12's `cudaDeviceProp` struct seems to have some incompatible layout changes (e.g., `luid` related fields are removed). Let's add the new v12 definition of it.

PR #1423, opened 2023-07-27
fix: unable to terminate a compute session when wsproxy address is not properly

pderer:
There was a bug that the session was not deleted when the wsproxy address of the scaling group was invalid.

This PR resolves this bug.

This PR is related to this PR [#1819](https://github.com/lablup/backend.ai-webui/pull/1819) in backend.ai-webui repository. 

For http status code issue, this PR [#1833](https://github.com/lablup/backend.ai-webui/pull/1833) is also related.


fregataa (commented):
This can make side effects because `query_wsproxy_status()` is also used from `manager.api.session.start_service()`.
How about handling `aiohttp.ClientConnectorError` in `get_wsproxy_version()`?
And let's add error popup or anything to webUI, which notices users that wsproxy cannot be connected.

pderer:
> This can make side effects because `query_wsproxy_status()` is also used from `manager.api.session.start_service()`.
> How about handling `aiohttp.ClientConnectorError` in `get_wsproxy_version()`?
> And let's add error popup or anything to webUI, which notices users that wsproxy cannot be connected.

Thanks for reviewing!
I didn't think of that part.
I will fix this code and Web UI part.
I will attach PR link in backend.ai-webui repository.

src/ai/backend/manager/api/scaling_group.py
```diff
@@ -93,13 +94,24 @@ async def get_wsproxy_version(request: web.Request, params: Any) -> web.Response
...
+                        )
+                    except aiohttp.ClientConnectorError:
+                        return web.HTTPInternalServerError(
+                            text=json.dumps(
+                                {
+                                    "type": "https://api.backend.ai/probs/internal-server-error",
+                                    "title": "Wsproxy of scaling group is invalid.",
+                                }
+                            ),
+                            content_type="application/problem+json",
```
fregataa: wow, could you give me any reference about this? seems interesting.

changes/1423.fix.md
```diff
@@ -0,0 +1 @@
+Unable to terminate a compute session when wsproxy address is not properly
```
fregataa:
```suggestion
Enable to terminate a compute session when wsproxy server has any connection issue.
```

fregataa (commented):
LGTM!
I left a comment about news fragment

changes/1423.fix.md
pderer: Thanks for reviewing! :)

src/ai/backend/manager/api/scaling_group.py
pderer: Thanks for reviewing!
I just saw this code in repo.
https://github.com/lablup/backend.ai/blob/main/src/ai/backend/web/proxy.py#L262-#L270

And in Web UI,
https://github.com/lablup/backend.ai-webui/blob/main/src/lib/backend.ai-client-esm.ts#L290-#L479
This code receive status code.

achimnol: Please use `ai.backend.manager.api.exceptions.URLNotFound` instead of manually creating `web.HTTPNotFound` with the RFC78007-style data.

achimnol: Please use `ai.backend.manager.api.exceptions.URLNotFound` instead of manually creating `web.HTTPNotFound` with the RFC7807-style data.

changes/1423.fix.md
```diff
@@ -0,0 +1 @@
+Enable to terminate a compute session when wsproxy server has any connection issue.
```
achimnol: The sentence looks strange. "has any connection issue"??

src/ai/backend/manager/api/scaling_group.py
```diff
@@ -93,13 +94,29 @@ async def get_wsproxy_version(request: web.Request, params: Any) -> web.Response
...
+                        wsproxy_status = await query_wsproxy_status(wsproxy_addr)
+                        wsproxy_version = wsproxy_status["api_version"]
+                        return web.json_response(
+                            {
+                                "wsproxy_version": wsproxy_version,
+                            }
+                        )
+                    except aiohttp.ClientConnectorError:
+                        log.error(
+                            "scaling group({}) uses INVALID WSProxy address: ({})",
```
achimnol: It's an already error message. There is no need to capitalize the description.
Also, let's distinguish the validation of wsproxy addr formats and actual availability of the wsproxy addr.

```suggestion
                            "Failed to query the wsproxy {1} configured for sg:{0}",
```

achimnol (changes requested):

src/ai/backend/manager/api/scaling_group.py
pderer: Thanks for reviewing!
I will fix it.

changes/1423.fix.md
pderer: Thanks for reviewing!
I will fix it.

src/ai/backend/manager/api/scaling_group.py
pderer: Thanks for reviewing!
I committed.

src/ai/backend/manager/api/scaling_group.py
```diff
@@ -93,13 +93,21 @@ async def get_wsproxy_version(request: web.Request, params: Any) -> web.Response
...
+                                "wsproxy_version": wsproxy_version,
+                            }
+                        )
+                    except aiohttp.ClientConnectorError:
+                        log.error(
+                            "Failed to query the wsproxy {1} configured for sg:{0}",
+                            scaling_group_name,
+                            wsproxy_addr,
+                        )
+                        return URLNotFound()
```
achimnol: Hm... sorry for requesting changes here again, but returning `URLNotFound` would be indistiguishable from the absence of `get_wsproxy_version` API itself.

Since this exception block is executed when the wsproxy address is invalid or inaccessible, I think we should wrap this using `ServerMisconfiguredError`.

achimnol (changes requested):
Sorry again, but almost done!

src/ai/backend/manager/api/scaling_group.py
pderer: Thanks for reviewing!
I will fix it.
And I fixed web ui code too.

achimnol (approved):

Issue #1408, opened 2023-07-21
Better handle "no such process" error when measuring process stats


achimnol:
A sample exception logging record that fails to be unpickled (`psutil.NoSuchProcess` is not pickle-safe):
```
LogRecord(
    name='ai.backend.agent.stats',
    msg='collect_per_container_process_stat(): gather_process_measures() error',
    args=('process no longer exists',),
    levelname='ERROR',
    levelno=40,
    pathname='/home/bai/.pyenv/versions/bai-23.03-agent/lib/python3.11/site-packages/ai/backend/common/logging_utils.py',
    filename='logging_utils.py',
    module='logging_utils',
    exc_info=(
        psutil.NoSuchProcess,
        'process no longer exists',
        (),
    ),
    stack_info=[
        ('aiomonitor.monitor', '/home/bai/.pyenv/versions/bai-23.03-agent/lib/python3.11/site-packages/aiomonitor/monitor.py', 12, 'gather_process_measures'),
        ('ai.backend.agent.docker.intrinsic', '/home/bai/.pyenv/versions/bai-23.03-agent/lib/python3.11/site-packages/ai/backend/agent/docker/intrinsic.py', 15, 'gather_process_measures'),
        ('psutil._pslinux', '/home/bai/.pyenv/versions/bai-23.03-agent/lib/python3.11/site-packages/psutil/_pslinux.py', 208, 'cpu_times'),
        ('psutil.__init__', '/home/bai/.pyenv/versions/bai-23.03-agent/lib/python3.11/site-packages/psutil/__init__.py', 232, 'cpu_times'),
    ],
    created=1654542763.860625,
    msecs=540.625,
    relativeCreated=1654542763.806,
    thread=MainThread,
    threadName='MainThread',
    processName='MainProcess',
    process=1336,
)
```

This seems to be caused by either trials to collect stats of terminated containers or existence of frequently (less than 1 sec) created and terminated processes in the user workloads.

Let's improve the exception handling to avoid unpickling errors and make it looking safer when this is a legitimate transient error.

PR #1424, opened 2023-07-27
fix: add type check conditions when awaiting gather_process_measures …

NayeonKeum:
resolves: #1408

I initially wanted to handle **NoSuchProcess** error at the function `gather_process_measures` but it already exsisted.

https://github.com/lablup/backend.ai/blob/f346699c0110f40ce78f3fcdbcf2c3e3a2f85354/src/ai/backend/agent/docker/intrinsic.py#L764-L774

So, I added a if/else condition at `src/ai/backend/agent/stats.py to check if the result of gather_process_measures task is psutil.NoSuchProcess` or not, and added a debug message of it's result.

This is before handling.
image

This is after handling.
image2

### DIFF
--- a/src/ai/backend/agent/stats.py
+++ b/src/ai/backend/agent/stats.py
@@ -7,6 +7,7 @@ Reference: https://www.datadoghq.com/blog/how-to-collect-docker-metrics/
 import asyncio
 import enum
 import logging
+import psutil
 import sys
 import time
 from decimal import Decimal
@@ -525,10 +526,14 @@ class StatContext:
             updated_cids: Set[ContainerId] = set()
             for result in results:
                 if isinstance(result, Exception):
-                    log.error(
-                        "collect_per_container_process_stat(): gather_process_measures() error",
-                        exc_info=result,
-                    )
+                    if type(result) == psutil.NoSuchProcess:    
+                        log.debug(f"{result}")
+                        pass
+                    else:
+                        log.error(
+                            "collect_per_container_process_stat(): gather_process_measures() error",
+                            exc_info=result,
+                        )
                     continue
                 for proc_measure in result:
                     metric_key = proc_measure.key


src/ai/backend/agent/stats.py
```diff
@@ -525,10 +526,14 @@ async def collect_per_container_process_stat(
...
-                        exc_info=result,
-                    )
+                    if type(result) == psutil.NoSuchProcess:    
+                        log.debug(f"{result}")
+                        pass
+                    else:
+                        log.error(
+                            "collect_per_container_process_stat(): gather_process_measures() error",
+                            exc_info=result,
+                        )
```
fregataa: I think we can use pattern matching here!
Let's try it.
```python
match result:
    case psutil.NoSuchProcess():
        log.debug(str(result))
    case Exception():
        log.error()
        ...
    case _:
        log.error()
        ...
```

fregataa (commented):
👍 
thanks! I left a comment.

src/ai/backend/agent/stats.py
NayeonKeum: Applied it right away 🚀

achimnol: Just FYI: This kind of `asyncio.gather(..., return_exceptions=True)` pattern will be replaced with [`aiotools.as_completed_safe()`](https://aiotools.readthedocs.io/en/latest/aiotools.utils.html#aiotools.utils.as_completed_safe) in the future.

For now, Python's pattern matching syntax would be a good fit here.

src/ai/backend/agent/stats.py
```diff
@@ -524,12 +525,17 @@ async def collect_per_container_process_stat(
...
-                    continue
+                match result:
+                    case psutil.NoSuchProcess():
+                        log.debug(str(result))
+                        continue
+                    case Exception():
+                        log.error(
+                            "collect_per_container_process_stat(): gather_process_measures() error",
+                            exc_info=result,
+                        )
```
fregataa:
```suggestion
                    case Exception():
                        log.error(
                            "collect_per_container_process_stat(): gather_process_measures() error",
                            exc_info=result,
                        )
                        continue
```
oops! miss the `continue` here

fregataa (changes requested):
Oops!

fregataa (approved):
👍 good!

src/ai/backend/agent/stats.py
```diff
@@ -524,12 +525,18 @@ async def collect_per_container_process_stat(
...
             for result in results:
-                if isinstance(result, Exception):
-                    log.error(
-                        "collect_per_container_process_stat(): gather_process_measures() error",
-                        exc_info=result,
-                    )
-                    continue
+                match result:
+                    case psutil.NoSuchProcess():
+                        log.debug(str(result))
```
achimnol: I think `psutil` is an implementation detail of the intrinsic CPUPlugin, so we should avoid importing it in the outer-level interface here.

Let's put matching of `psutil.NoSuchProcess` inside the `src/ai/backend/agent/docker/intrinsic.py`.

NOTE: We also need to backport this PR to the activator plugin that overrides the CPUPlugin in enterprise setups. I'll take care of that after this PR is done.

achimnol (changes requested):

src/ai/backend/agent/stats.py
NayeonKeum: 👌 I will work on it

NayeonKeum:
I found out that `psutil.NoSuchProcess` was already catched inside a try-catch exception, but the error turned out to be leaking outside it(in CPUPlugin, to agent stats). 
However, getting process id with `psutil.Process` in MemoryPlugin was catching the error perfectly, and it was synchronus.
I made a ThreadPoolExecutor but it needed more than 100 max_workers to catch CPUPlugin's `NoSuchProcess` error.
So, just changed CPUPlugin's `psutil.Process` same as MemoryPlugin.

achimnol (approved):

PR #1431, opened 2023-07-27
setup: Upgrade the base Python version from 3.11.3 to 3.11.4

inureyes:
Bump the base Python version (3.11.3 → 3.11.4) to resolve potential bugs.

Also:
- Upgrade redis-py (4.5.5 → 4.6.0)
- Upgrade lark (lark-parser 0.11.3 → lark 1.1.5; package is renamed)
- Stabilize flakky test cases (especially `tests/common/redis`) by ensuring the TCP ports used by fixture containers to become free
  - They are allocated using the parallel execution slot numbers and theorectically there should be no conflicts. But in GitHub Actions, they seem to remain open for a short period of time (randomly) even after terminating the containers. It may be because Docker uses a proxy process for each bound container port and repeating creation/destruction of containers for many times in paralllel seems to cause unexpected random delays to clear up them.


achimnol (approved):

achimnol:
`require:db-migration` is a bogus label because this PR's update changes the black formatting style affecting some files in the alembic migrations folder.

Issue #1388, opened 2023-07-12
Cannot add AWS ECR registry to use custom container images


iv0rish:
### What Operating System(s) are you seeing this problem on?

Linux (x86-64)

### Backend.AI version

23.03.9.dev0

### Describe the bug

Not only AWS ECR, I cannot add container registry which includes numeric or special characters like hypen or underscore due to below pattern at `configure-registry-url` textfield. 

```
^(https?)://(([a-zA-Zd.]{2,}).([a-zA-Z]{2,})|(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3})(:((6553[0-5])|(655[0-2])|(65[0-4][0-9]{2})|(6[0-4][0-9]{3})|([1-5][0-9]{4})|([0-5]{0,5})|([0-9]{1,4})))?$
```

### To Reproduce

Try to add registry in webui or desktop app

achimnol:
Thanks for the report. I think we should improve the input validation for the registry URLs.

achimnol:
Done!

PR #1432, opened 2023-07-28
refactor: Update URL parsing and validation in the `URL` class

kimjinmyeong:
related to #1388.
As a result of this [PR](https://github.com/lablup/backend.ai-webui/pull/1806), regular expressions are no longer used for URL validation. 
So, I updated to reflect this change in URL validation at backend.ai manager.

**Improvement**
- Modified the URL scheme validation to avoid using regular expressions.
- Exception handling codes have been modified to catch possible exceptions raised by the yarl library
The yarl library provides appropriate error messages for each exception. ([yarl GitHub repo](https://github.com/aio-libs/yarl/blob/1f94e1a7833a27203d63f8618d01d4d4fa2d3520/yarl/_url.py))
- Added test case to validate AWS ECR Url


src/ai/backend/common/validators.py
```diff
@@ -657,19 +655,19 @@ def __init__(
...
-        except ValueError as e:
-            self._failure(f"cannot convert the given value to URL (error: {e!r})", value=value)
+            parsed_url = yarl.URL(value)
+            if self.scheme_required:
+                if parsed_url.scheme == "":
+                    self._failure(
+                        "The given value does not have the scheme (protocol) part", value=value
+                    )
+        except Exception as e:
+            self._failure(f"{e}", value=value)
```
fregataa:
```suggestion
            self._failure(repr(e), value=value)
```
this would be better

fregataa (commented):
Good! I left one comment

src/ai/backend/common/validators.py
```diff
@@ -657,19 +655,19 @@ def __init__(
...
-            return yarl.URL(value)
-        except ValueError as e:
-            self._failure(f"cannot convert the given value to URL (error: {e!r})", value=value)
+            parsed_url = yarl.URL(value)
+            if self.scheme_required:
+                if parsed_url.scheme == "":
+                    self._failure(
+                        "The given value does not have the scheme (protocol) part", value=value
+                    )
+        except Exception as e:
```
achimnol: Let's narrow down this to `ValueError`.
See https://github.com/aio-libs/yarl/blob/master/yarl/_url.py

achimnol (changes requested):

achimnol (approved):

PR #1437, opened 2023-07-28
feat: add GraphQL queries to support paginated vFolder listing

### LINKED ISSUE #1386: Support vfolder list pagination API
### Main idea

Let's add a new gql query for vfolder list pagination. Three separate queries are required. And each query needs to support pagination and queryfilter.
```[tasklist]
- [x] Query own folders
- [x] Query invited folders
- [x] Query project folders
```

### Alternative ideas

_No response_

### Anything else?

_No response_

Choi-Jiwon-38:
This PR resolves #1386 

I add three queries
- `vfolder_own_list` 
- `vfolder_invited_list`
- `vfolder_project_list`


src/ai/backend/manager/models/vfolder.py
```diff
@@ -1210,6 +1210,172 @@ async def batch_load_by_user(
...
+        graph_ctx: GraphQueryContext,
+        limit: int,
+        offset: int,
+        *,
+        domain_name: str = None,
+        group_id: uuid.UUID = None,
+        user_id: uuid.UUID = None,
+        filter: str = None,
+        order: str = None,
+    ) -> Sequence[VirtualFolder]:
```
fregataa:
```suggestion
    ) -> list[VirtualFolder]:
```
`typing.Sequence` is deprecated since Python 3.9, let's just use `list` instead!
https://docs.python.org/ko/3/library/typing.html#typing.Sequence

src/ai/backend/manager/models/vfolder.py
```diff
@@ -1210,6 +1210,172 @@ async def batch_load_by_user(
...
+        user_id: uuid.UUID = None,
+        filter: str = None,
+    ) -> int:
+        from .user import users
+
+        j = vfolders.join(
+            vfolder_permissions,
+            vfolders.c.id == vfolder_permissions.c.vfolder,
+            isouter=True,
+        ).join(users, vfolders.c.user == users.c.uuid, isouter=True)
```
fregataa:
```suggestion
        j = sa.join(
            vfolders, vfolder_permissions,
            vfolders.c.id == vfolder_permissions.c.vfolder,
        )
```
we do not need multiple joins here since this query is only for counting number of rows.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -1210,6 +1210,172 @@ async def batch_load_by_user(
...
+    ) -> Sequence[VirtualFolder]:
+        from .group import groups
+        from .user import users
+
+        j = (
+            vfolders.join(
+                vfolder_permissions,
+                vfolders.c.id == vfolder_permissions.c.vfolder,
+                isouter=True,
+            )
```
fregataa:
```suggestion
           sa.join(
                vfolders, vfolder_permissions,
                vfolders.c.id == vfolder_permissions.c.vfolder,
            )
```
we do not need `isouter=True` here

src/ai/backend/manager/models/vfolder.py
```diff
@@ -1210,6 +1210,172 @@ async def batch_load_by_user(
...
+        filter: str = None,
+        order: str = None,
+    ) -> Sequence[VirtualFolder]:
+        from ai.backend.manager.models import association_groups_users as agus
+
+        from .group import groups
+        from .user import users
+
+        j = sa.join(agus, users, agus.c.user_id == users.c.uuid)
+        query = sa.select([agus.c.group_id]).select_from(j).where(agus.c.user_id == user_id)
```
fregataa:
```suggestion
        query = sa.select([agus.c.group_id]).select_from(agus).where(agus.c.user_id == user_id)
```
We do not need any join here because `association_group_users` table already has `user_id` and `group_id` columns.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -1210,6 +1210,172 @@ async def batch_load_by_user(
...
+        user_id: uuid.UUID = None,
+        filter: str = None,
+    ) -> int:
+        from ai.backend.manager.models import association_groups_users as agus
+
+        from .group import groups
+        from .user import users
+
+        j = sa.join(agus, users, agus.c.user_id == users.c.uuid)
+        query = sa.select([agus.c.group_id]).select_from(j).where(agus.c.user_id == user_id)
```
fregataa:
```suggestion
        query = sa.select([agus.c.group_id]).select_from(agus).where(agus.c.user_id == user_id)
```
Same here

src/ai/backend/manager/models/vfolder.py
```diff
@@ -1210,6 +1210,172 @@ async def batch_load_by_user(
...
+        async with graph_ctx.db.begin_readonly() as conn:
+            result = await conn.execute(query)
+        grps = result.fetchall()
+        group_ids = [g.group_id for g in grps]
+        j = vfolders.join(groups, vfolders.c.group == groups.c.id)
+        query = (
+            sa.select([vfolders, users.c.email, groups.c.name.label("groups_name")])
```
fregataa:
```suggestion
            sa.select([vfolders, groups.c.name.label("groups_name")])
```
We cannot fetch `users.c.email` column here because `group` owned vfolders have `null` value on `user` column.
Instead, we can use vfolder's `creator` field here.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -1210,6 +1210,172 @@ async def batch_load_by_user(
...
+            )
+            .join(
+                users,
+                vfolders.c.user == users.c.uuid,
+                isouter=True,
+            )
+            .join(groups, vfolders.c.group == groups.c.id, isouter=True)
+        )
+        query = (
+            sa.select([vfolders, users.c.email, groups.c.name.label("groups_name")])
```
fregataa, line 1275:
```suggestion
            sa.select([vfolders, groups.c.name.label("groups_name")])
```
Same here because users can be invitied to `group` owned vfolders. Then those `group` owned vfolders do not have valid `user` column value.

fregataa (changes requested):
Thank you for your work!
I think we need to discuss more about this issue.
Still, you can reflect some of the comments I left.

src/ai/backend/manager/models/vfolder.py
Choi-Jiwon-38: Currently, `load_slice_project`'s return value is related to class method `from_row`.
In `from_row`, `users_email` is required. 

>   File "/Users/choijiwon/bai-dev/src/ai/backend/manager/models/vfolder.py", line 1031, in from_row
>   user_email=row["users_email"],
>              ~~~^^^^^^^^^^^^^^^

So, I think we have to leave `users.c.email` as before.

Choi-Jiwon-38:
Thanks for review my PR! @fregataa 

I reflect on some comments that in below.
- Replace `Sequence` to `list` (only my edits).
- Remove not needed `join`, `isouter`.
- Add missing vfolder list's parameter `domain`.

src/ai/backend/client/func/vfolder.py
```diff
@@ -231,15 +312,21 @@ async def _download_file(
                                     ) as pbar:
                                         loop = current_loop()
                                         writer_fut = loop.run_in_executor(
-                                            None, self._write_file, file_path, file_mode, q.sync_q
+                                            None,
+                                            self._write_file,
+                                            file_path,
+                                            file_mode,
+                                            q.sync_q,
```
fregataa, line 319: please make sure that this branch is not behind of main or mistake on work around another branches.
Or is there any reason to change this part?

src/ai/backend/client/func/vfolder.py
```diff
@@ -231,15 +312,21 @@ async def _download_file(
...
                                         await asyncio.sleep(0)
                                         max_attempts = 10
                                         while True:
                                             try:
                                                 async for attempt in AsyncRetrying(
                                                     wait=wait_exponential(
-                                                        multiplier=0.02, min=0.02, max=5.0
+                                                        multiplier=0.02,
+                                                        min=0.02,
+                                                        max=5.0,
```
fregataa, line 329: same here

src/ai/backend/manager/models/vfolder.py
fregataa: This will return the requester's email or return no group-owned folders or just generate cartesian products that we must avoid.
So we should fix `from_row()` or test by replacing `users.c.email` to `vfolders.c.creator.label("users_email")` instead of leaving this part.
Could you test `vfolders.c.creator.label("users_email")` first?

fregataa (changes requested):

src/ai/backend/manager/models/vfolder.py
```diff
@@ -1020,13 +1020,19 @@ class Meta:
...
+            user_email = None
+
         return cls(
             id=row["id"],
             host=row["host"],
             quota_scope_id=row["quota_scope_id"],
             name=row["name"],
             user=row["user"],
-            user_email=row["users_email"],
+            user_email=user_email,
```
fregataa:
```suggestion
            user_email= row["users_email"] if "user_email" in row else None
```
Could you try this? Please remove the `try-catch` block if this works.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -1210,6 +1216,182 @@ async def batch_load_by_user(
...
+            )
+            .join(
+                users,
+                vfolders.c.user == users.c.uuid,
+                isouter=True,
+            )
+            .join(groups, vfolders.c.group == groups.c.id, isouter=True)
+        )
+        query = (
+            sa.select([vfolders, users.c.email, groups.c.name.label("groups_name")])
```
fregataa:
```suggestion
            .join(
                users,
                vfolders.c.user == users.c.uuid,
                isouter=True,
            )
        )
        query = (
            sa.select([vfolders, users.c.email])
```
Let's remove the `group` join because invited vfolders do not belong to any group.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -1210,6 +1216,182 @@ async def batch_load_by_user(
...
+        group_id: uuid.UUID = None,
+        user_id: uuid.UUID = None,
+        filter: str = None,
+    ) -> int:
+        from .user import users
+
+        j = vfolders.join(
+            vfolder_permissions,
+            vfolders.c.id == vfolder_permissions.c.vfolder,
+            isouter=True,
```
fregataa, line 1234:
```suggestion
```
We do not need outer join here

fregataa (changes requested):

src/ai/backend/manager/models/vfolder.py
Choi-Jiwon-38: I was approaching in the wrong way...😅
(ex: `row["users_email"] if row["users_email"] else None`)

And it(changed codes) works very well!

Choi-Jiwon-38:
@fregataa 
Thanks for review my PR 😉 
I reflect all of your comments.
But I caught some problems that I miss

1.  When use `./backend.ai vfolder list-invited`, Infinite calls occur.
스크린샷 2023-08-07 16 56 48

2. When invited users use `./backend.ai vfolder list-project`(not owner), Duplicate results appear. 
스크린샷 2023-08-07 16 56 08

changes/1437.doc.md
```diff
@@ -0,0 +1 @@
+Improve formatting and trafaret compatibility error reporting of the OpenAPI-based Manager REST API reference
```
fregataa: This description seems wrong. Please check it!

src/ai/backend/manager/models/vfolder.py
```diff
@@ -1210,6 +1211,178 @@ async def batch_load_by_user(
...
+        from .user import users
+
+        j = vfolders.join(
+            vfolder_permissions,
+            vfolders.c.id == vfolder_permissions.c.vfolder,
+        ).join(
+            users,
+            vfolders.c.user == users.c.uuid,
+            isouter=True,
+        )
```
fregataa:
```suggestion
        ).join(
            users,
            vfolder_permissions.c.user == users.c.uuid,
        )
```
This would be better!

src/ai/backend/manager/models/vfolder.py
```diff
@@ -1210,6 +1211,178 @@ async def batch_load_by_user(
...
```
fregataa:
```suggestion
        ).join(
            users,
            vfolder_permissions.c.user == users.c.uuid,
        )
```
Same here

fregataa (changes requested):
Almost there!

fregataa (approved):
Thanks!
Plus, I don't recommend using a formatter that only runs in your development environment.
The reason for using formatters is to keep everyone in the same coding style in one codebase. It will be confusing and not consistent if we use formatters on our own.

Sujin-Kim1 (approved):
LGTM

kyujin-cho (commented):
Update: After ongoing some internal discussions we have made a conclusion that it is better to split out query resolvers for every project-level and invited vfolders. It is now safe to ignore the comment below.

For me it feels like most part of the newly written resolvers (methods starting `load_`) share similar characteristics with existing `load_count` and `load_slice` resolvers and thus do not have particular reason to be created as a separated resolver instead of extending. Please rewrite `VFolder` class (and also other affected parts) so that existing `load_count()` and `load_slice()` methods can contain logics to limit selection to only invited and project shared folders.

PR #1441, opened 2023-08-01
setup: Auto-enable `--editable-webui` for `install-dev.sh` in the main branch

achimnol:
There are two ways to run the web UI with this option:

1. Use the npm commands (as instructed at the end of `install-dev.sh` execution) to run the full-fledged development servers of the webui
2. Use the standard `./py -m ai.backend.web.server` command to use the local version compiled from the checked out source.

NOTE: The prerequisite for `--editable-webui`: the latest stable version of NodeJS and `npm`


inureyes (approved):
LGTM

PR #1454, opened 2023-08-07
feature: add new plugin hook on predicate check

fregataa:
There is a demand for a feature to restrict session creation based on complex policies defined outside the system.
We implement this feature as a plugin to minimize code modifications to the Backend.AI core and allow the plugin to inspect policies during the predicate check phase.
For example, the current implementation, if you want to limit GPU resources to only interactive sessions for all user accounts in the system, you need to set the keypair resource policy or project resource policy individually and add new code in Backend.AI core. If the predicate plugin feature is added, you can implement the plugin according to the policy you want to restrict.


src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -471,6 +472,38 @@ async def _check_predicates() -> List[Tuple[str, Union[Exception, PredicateResul
                     )
                     has_failure = True
 
+            async def _check_predicates_hook() -> HookResult:
+                async with self.db.begin_session() as db_sess:
```
achimnol: Could we make this a read-only session if it does not violate any intended use case?

achimnol (approved):

PR #1455, opened 2023-08-07
refactor: replace `kernels.role` field with `sessions.is_system_session`

### LINKED ISSUE #1446: Rename `KernelRole` to something else with clarification that it's for sessions
Naming is important.
`ClusterRole` and `KernelRole` are confusing to each other.

kyujin-cho:
Resolves #1446.


achimnol (changes requested):
I have dumped my concerns in the chat about the original direction about having multiple different session types: `ComputeSession`, `InferenceSession`, and `SystemSession`, instead of distinguishing them using yet another field or flag.

Could you adapt the existing implementation to remedy these concerns?

achimnol:
We are going to use `SessionTypes` for `INTERACTIVE`, `BATCH`, `INFERENCE`, and `SYSTEM` (new) and remove `kernels.role`. Closing this PR in favor of #1587.

PR #1457, opened 2023-08-07
feat: add authentication headers when proxying pipeline websocket requests

rapsealk:
This PR updates the websocket proxy handler for the pipeline service, in the same fashion as #1350, to support WebSocket-based GraphQL subscription.


src/ai/backend/web/proxy.py
```diff
@@ -399,6 +400,25 @@ async def websocket_handler(request, *, is_anonymous=False) -> web.StreamRespons
...
+                        "iss": "Backend.AI Webserver",
+                        "iat": now,
+                        # Private claims
+                        "aiohttp_session": aiohttp_session,
+                        "access_key": api_session.config.access_key,
+                        # "secret_key": api_session.config.secret_key,
+                    }
+                    sso_token = jwt.encode(claims, key=jwt_secret, algorithm="HS256")
+                api_rqst.headers["X-BackendAI-SSO"] = sso_token
+                if session_id := (request_headers.get("X-BackendAI-SessionID") or aiohttp_session):
```
fregataa, line 422:
```suggestion
                if session_id := request_headers.get("X-BackendAI-SessionID", aiohttp_session):
```
There is no such big deal but how about this?

src/ai/backend/web/proxy.py
```diff
@@ -399,6 +400,25 @@ async def websocket_handler(request, *, is_anonymous=False) -> web.StreamRespons
                 content_type=request.content_type,
                 override_api_version=request_api_version,
             )
+            if proxy_path == "pipeline":
+                aiohttp_session = request.cookies.get("AIOHTTP_SESSION")
+                if not (sso_token := request.headers.get("X-BackendAI-SSO")):
+                    jwt_secret = request.app["config"]["pipeline"]["jwt"]["secret"]
+                    now = datetime.now(tz=timezone(timedelta(hours=9)))
```
fregataa: What does that `timedelta(hours=9)` mean??
Does it mean that this will be expired in 9 hours or something else?

src/ai/backend/web/proxy.py
fregataa: Plus, is it possible that `aiohttp_session` can be `None` here?

fregataa (commented):
I left some comments here

src/ai/backend/web/proxy.py
rapsealk: This is to set the timezone to KST(UTC+09:00).

src/ai/backend/web/proxy.py
rapsealk: 1. This is to avoid unexpected results: `request_headers.get("X-BackendAI-SessionID", aiohttp_session)` will return `None` rather than `aiohttp_session` when `request_headers = {"X-BackendAI-SessionID": None}`.
2. If `aiohttp_session` is `None`, which means a session is not authenticated, then the pipeline service will invalidate and guide an user to log in.

src/ai/backend/web/proxy.py
fregataa: where is the counterpart of this? which component uses this jwt token?

rapsealk: The token is for external services, such as the pipeline service. (https://github.com/lablup/backend.ai-fasttrack/blob/7b3832163a7fe0debd2c5c0c88614aa6b1c3cf28/config/asgi.py#L61-L86)

src/ai/backend/web/proxy.py
```diff
@@ -399,6 +400,25 @@ async def websocket_handler(request, *, is_anonymous=False) -> web.StreamRespons
                 content_type=request.content_type,
                 override_api_version=request_api_version,
             )
+            if proxy_path == "pipeline":
+                aiohttp_session = request.cookies.get("AIOHTTP_SESSION")
+                if not (sso_token := request.headers.get("X-BackendAI-SSO")):
+                    jwt_secret = request.app["config"]["pipeline"]["jwt"]["secret"]
```
kyujin-cho, line 408: Please update the config schema of webserver (`ai/backend/webserver/config.py`) accordingly.

src/ai/backend/web/proxy.py
kyujin-cho: As Backend.AI is not only served on countries with KST (yes, we are international!) it is a bad idea to hardcode the timezone like this. When adding `.astimezone()` to the end of `now()` call automatically detects time timezone set on host machine and updates the datetime object ([ref](https://docs.python.org/3.11/library/datetime.html#datetime.datetime.astimezone)), let's update this code to adapt these behavior.

kyujin-cho (changes requested):

src/ai/backend/web/proxy.py
rapsealk: Thank you for catching this!

src/ai/backend/web/config.py
```diff
@@ -86,7 +86,12 @@
         t.Key("pipeline"): t.Dict(
             {
                 t.Key("endpoint", default=None): t.Null | tx.URL,
-            }
+                t.Key("jwt"): t.Dict(
+                    {
+                        t.Key("secret"): t.String,
+                    },
+                ),
+            },
```
kyujin-cho: If I am guessing correctly, this `jwt` directive is for the traffics from FastTrack. FastTrack is a private project and thus only limited number of users who operate Backend.AI webserver will be required to fill out this field. When taking account into this situation I don't think this whole `pipeline` configuration body should be a required config values. Please refactor the validator to either mark the `pipeline` configuration as an Optional value or at lease leave `jwt` as nullable and dynamically check its existence when the request actually comes from the FastTrack side.

kyujin-cho (changes requested):

rapsealk:
Closing as support for WebSocket is not needed yet.

PR #1460, opened 2023-08-08
feat: CRUD `wsproxy_addr`, `wsproxy_api_token` in CLI

Sujin-Kim1:
- Add `wsproxy_api_token` column to manager/model.
- Make it possible to CRUD `wsproxy_addr` and `wsproxy_api_token` in CLI.
follows https://github.com/lablup/backend.ai/pull/1278
related: https://github.com/lablup/backend.ai-control-panel/issues/660

### DIFF
--- a/src/ai/backend/client/func/scaling_group.py
+++ b/src/ai/backend/client/func/scaling_group.py
@@ -1,6 +1,6 @@
 import json
 import textwrap
-from typing import Iterable, Mapping, Sequence
+from typing import Iterable, Mapping, Optional, Sequence
 
 from ai.backend.client.output.fields import scaling_group_fields
 from ai.backend.client.output.types import FieldSpec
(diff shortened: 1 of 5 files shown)


adrysn (approved):

src/ai/backend/client/func/scaling_group.py
```diff
@@ -123,6 +127,8 @@ async def create(
         scheduler: str = None,
         scheduler_opts: Mapping[str, str] = None,
         use_host_network: bool = False,
+        wsproxy_addr: str = None,
+        wsproxy_api_token: str = None,
```
kyujin-cho: These fields allow `None` as a valid value so the type annotation should be `Optional`.
```suggestion
        wsproxy_addr: Optional[str] = None,
        wsproxy_api_token: Optional[str] = None,
```

src/ai/backend/client/func/scaling_group.py
```diff
@@ -168,6 +176,8 @@ async def update(
...
```
kyujin-cho: Same here.

kyujin-cho (changes requested):

src/ai/backend/client/func/scaling_group.py
achimnol: FYI: All new codes should explicitly define `| None` or `Optional[]` when annotating nullable values in function arguments.
See https://adamj.eu/tech/2022/10/18/python-type-hints-implicit-optional-types/

kyujin-cho (approved):

PR #1462, opened 2023-08-08
feat: add redis ping command at manager CLI

hyoinandout:
This command checks whether the redis is healthy or not by executing PING command to every database inside redis. Resolves lablup/backend.ai-ossca-2023#11.


fregataa (changes requested):
Thanks!
We have to write a news fragment markdown file for every PR.
Please check [this PR](https://github.com/lablup/backend.ai/pull/1442) and [the news fragment file](https://github.com/lablup/backend.ai/blob/0b5f590ae6375b864c490464d8598df06b5de693/changes/1442.misc.md), and push an appropriate file!

src/ai/backend/manager/cli/redis.py
```diff
@@ -0,0 +1,67 @@
...
+                log.exception("REDIS_IMAGE_DB is not ok")
+            try:
+                await redis_helper.execute(
+                    redis_conn_set.stream,
+                    lambda r: r.execute_command("PING"),
+                )
+                log.info("REDIS_STREAM_DB is ok")
+            except Exception:
+                log.exception("REDIS_STREAM_DB is not ok")
+
```
fregataa: How about implement an inner function like below?
```python
async def _ping(redis_conn) -> None:
    try:
        await redis_helper.execute(redis_conn, lambda r: r.execute_command("PING"))
    except Exception:
        log.exception(f"{redis_conn} is not ok")
```

then this part can be...
```python
await _ping(redis_conn_set.live)
await _ping(redis_conn_set.stat)
await _ping(redis_conn_set.image)
await _ping(redis_conn_set.stream)
```

fregataa (changes requested):
Thank you for your commitment!
I left a comment.

src/ai/backend/manager/cli/redis.py
hyoinandout: Agreed

hyoinandout:
https://github.com/lablup/backend.ai-ossca-2023/pull/14#issuecomment-1682390006

kyujin-cho (approved):

PR #1469, opened 2023-08-09
fix: allow `ResourceSlotColumn` to return `None`

fregataa:
follow up #856.
in #856, we did hotfix to check `None` value when null value is stored in `ResourceSlotColumn`. After that, an empty `ResourceSlot` value is returned if a null value is stored in a `ResourceSlotColumn`. This makes it impossible to distinguish between an empty `ResourceSlot` value and a `None` value, which can be problematic.

So, we enable `ResourceSlotColumn` to return `None` since we need to distinguish between empty `ResourceSlot` value and `None`.
And alter `kernels.requested_slots` column into not nullable since the value of the column should not be null logically. Plus a migration file updates null `kernels.requested_slots` columns in case of any test data has been inserted.


src/ai/backend/manager/models/base.py
```diff
@@ -208,9 +208,9 @@ def process_bind_param(
             return value.to_json()
         return value
 
-    def process_result_value(self, raw_value: Dict[str, str], dialect) -> ResourceSlot:
+    def process_result_value(self, raw_value: Dict[str, str], dialect) -> ResourceSlot | None:
```
kyujin-cho: This should be `Optional`.
```suggestion
    def process_result_value(self, raw_value: Dict[str, str] | None, dialect) -> ResourceSlot | None:
```

kyujin-cho (changes requested):

achimnol (approved):

PR #1471, opened 2023-08-09
feat: Add container registry CRUD API using GraphQL

### LINKED ISSUE #1450: When deleting a registry, other registries with the same hostname prefix are also being deleted

macOS (Apple Silicon)

### Backend.AI version

23.03.10

### Describe the bug

When attempting to delete a registry, other registries containing the same prefix in their hostname are being deleted as well.

# Video
https://github.com/lablup/backend.ai/assets/84405002/c3e84e69-91ee-438d-846a-d964141d8e7d

### To Reproduce

1. Create registries that contains the same prefix word in hostname (e.g. "test", "test1", "test2")
2. Delete a registry
3. Observe that registries with same prefix hostname are also affected by the deletion process.

### Expected Behavior

The deletion process should only target the registry with the exact hostname that the user has chosen to delete. 

### Anything else?

_No response_

kimjinmyeong:
This PR did the implementation of a CRUD API for container registry using GraphQL, 
and also resolves #1450. 

# GraphQL Implementation 
**Scheme**
```graphql
type ContainerRegistry {
    hostname: String!
    config: ContainerRegistryConfig!
}

type ContainerRegistryConfig {
    url: String!
    type: String!
    project: List(String)
    username: String
    password: String
	ssl_verify: Boolean
}
```
**Get Registries**
query
```graphql
query {
    container_registries {
            hostname
            config {
                url
                type
                project
                username
                password
                ssl_verify
            }
    }
}
```
response
```json
{
    "container_registries": [
        {
            "hostname": "test1",
            "config": {
                "url": "https://test1.com",
                "type": "harbor",
                "project": [
                    "abc",
                    "bcd"
                ],
                "username": "test1",
                "password": "test1",
                "ssl_verify": true
            }
        },
        {
            "hostname": "test2",
            "config": {
                "url": "https://test2.com",
                "type": "harbor",
                "project": [
                    "abc",
                    "bcd"
                ],
                "username": "test2",
                "password": "test2"
                "ssl_verify": false
            }
        },
        {
            "hostname": "test3",
            "config": {
                "url": "https://test3.com",
                "type": "harbor",
                "project": [
                    "abc",
                    "bcd"
                ],
                "username": "test3",
                "password": "test3",
                "ssl_verify": true
            }
        }
    ]
}
```

**Create Container Registry** 
query
```graphql
mutation testCreateContainerRegistry ($hostname: String!, $props: CreateContainerRegistryInput!){
    create_container_registry(hostname: $hostname, props: $props)
    {
        result
    }
}
```
variables
```json
{
    "hostname": "test123",
    "props": {
        "url": "https://test.com",
        "type": "harbor",
        "project": "test1,test2",
        "username": "test",
        "password": "test",
        "ssl_verify": "true"
    }
}
```

> **Note**
> The `ssl_verify` field accepts stringified/integer boolean variables as specified in:
> https://trafaret.readthedocs.io/en/latest/api/trafaret.html#trafaret.ToBool
> The `project` field accepts both comma-separated list and actual JSON list of strings.

response
```json
{
    "create_container_registry": {
        "result": "ok"
    }
}
```

**Modify Container Registry**
query
```graphql
mutation testModifyContainerRegistry ($hostname: String!, $props: ModifyContainerRegistryInput!){
    modify_container_registry(hostname: $hostname, props: $props)
    {
        result
    }
}
```
variables (may use a partial set of fields)
```json
{
    "hostname": "test1",
    "props": {
        "url": "https://test2.com",
        "type": "docker",
        "project": ""
    }
}
```
response
```json
{
    "modify_container_registry": {
        "result": "ok"
    }
}
```
**Delete Container Registry**
query
```graphql
mutation testDeleteContainerRegistry($hostname: String!) {
    delete_container_registry(hostname: $hostname) {
        result
    }
}
```
variables
```json
{
    "hostname": "test123"
}
```
response
```json
{
    "delete_container_registry": {
        "result": "ok"
    }
}
```
**Get a Registry**
query
```graphql
query ($hostname: String!){
    container_registry (hostname: $hostname){
        hostname
        config {
            url
            type
            project
            username
            password
            ssl_verify
        }
    }
}
```
variables
```json
{
    "hostname": "test123"
}
```
response
```json
{
    "container_registry": {
        "hostname": "test123",
        "config": {
            "url": "https://test.com",
            "type": "docker",
            "project": [],
            "username": null,
            "password": null,
            "ssl_verify": true
        }
    }
}
```


src/ai/backend/manager/models/etcd.py
```diff
@@ -0,0 +1,257 @@
...
+        except KeyError:
+            raise ObjectNotFound(object_name=f"registry: {hostname}")
+        log.info(
+            "ETCD.DELETE_CONTAINER_REGISTRY (ak:{}, key: {}, hostname:{})",
+            ctx.access_key,
+            ETCD_CONTAINER_REGISTRY_KEY,
+            hostname,
+        )
+        await ctx.shared_config.etcd.delete_prefix(f"{ETCD_CONTAINER_REGISTRY_KEY}/{hostname}")
+        await ctx.shared_config.etcd.put_prefix(ETCD_CONTAINER_REGISTRY_KEY, registries)
```
fregataa: Let's think of any good way to reduce transactions!

src/ai/backend/manager/models/etcd.py
```diff
@@ -0,0 +1,257 @@
...
+        cls,
+        ctx: GraphQueryContext,
+    ) -> Sequence[ContainerRegistry]:
+        log.info(
+            "ETCD.LOAD_ALL_CONTAINER_REGISTRIES (ak:{}, key:{})",
+            ctx.access_key,
+            ETCD_CONTAINER_REGISTRY_KEY,
+        )
+        raw_registries = await ctx.shared_config.etcd.get_prefix_dict(ETCD_CONTAINER_REGISTRY_KEY)
+        registries = t.Mapping(t.String, container_registry_iv).check(raw_registries)
```
fregataa: Seems this code is used as the same in other functions.
How about implement a method like `get_container_registry()` in SharedConfig?

fregataa (commented):
I left a few comments!

src/ai/backend/manager/models/gql.py
```diff
@@ -570,6 +582,10 @@ class Queries(graphene.ObjectType):
         quota_scope_id=graphene.String(required=True),
     )
 
+    container_registry = graphene.Field(ContainerRegistry, hostname=graphene.String(required=True))
+
+    container_registries = graphene.Field(ContainerRegistries)
```
achimnol: This should be
```suggestion
    container_registries = graphene.List(ContainerRegistries)
```

achimnol (changes requested):

src/ai/backend/manager/models/etcd.py
```diff
@@ -0,0 +1,257 @@
...
+    "ModifyContainerRegistry",
+    "DeleteContainerRegistry",
+)
+
+ETCD_CONTAINER_REGISTRY_KEY = "config/docker/registry"
+
+
+class CreateContainerRegistryInput(graphene.InputObjectType):
+    url = graphene.String(required=True)
+    container_type = graphene.String(required=True)
```
achimnol: This should be `registry_type`.

src/ai/backend/manager/models/etcd.py
```diff
@@ -0,0 +1,257 @@
...
+                if k == "":
+                    inner_prefix = f"{ETCD_CONTAINER_REGISTRY_KEY}/{hostname}"
+                else:
+                    inner_prefix = f"{ETCD_CONTAINER_REGISTRY_KEY}/{hostname}/{k}"
+                if isinstance(v, Mapping):
+                    flatten(v)
+                else:
+                    updates[inner_prefix] = v
+
+        flatten(config)
```
achimnol: The calculated `config` seems not to be used anywhere. Please clean up the code here.

src/ai/backend/manager/models/etcd.py
```diff
@@ -0,0 +1,257 @@
...
```
achimnol: Even with inner functions, it's better to minimize the side-effects, particularly when the inner function mutates outer scope's values. Could we rewrite `flatten()` function to create and return `updates`? (The exception is when you write test codes to inspect inner activities.)

achimnol (changes requested):

src/ai/backend/manager/models/etcd.py
achimnol: I agree with @fregataa. Let's make the CUD methods reusable across other parts of our codebase by moving them to `SharedConfig`.

src/ai/backend/manager/models/etcd.py
achimnol: Currently there is no simple way to group multiple mutation calls to `ctx.shared_config.etcd` as a single transaction unless we break into our thin `ai.backend.common.etcd` abstraction layer.

As mutation of container registries rarely happen in production (I don't expect multiple superadmins would change this setting simultaneously...), I think this is fine for now. We could introduce context-based transaction support in [etcd-client-py](https://github.com/lablup/etcd-client-py) and the future `ai.backend.common.etcd` API later. (cc: @sanxiyn)

src/ai/backend/manager/models/etcd.py
bezzang-dev: @achimnol 
`config` removes keys with `None` values and uses it as an argument for the `flatten` function to create `updates`. Otherwise, it will be stored in etcd as `'None'`.(e.g. `username: 'None'`).
https://github.com/lablup/backend.ai/blob/da28c97afa488016d6c84e15b441675503e36bf2/src/ai/backend/manager/models/etcd.py#L194

src/ai/backend/manager/config.py
```diff
@@ -616,6 +616,11 @@ async def get_raw(self, key: str, allow_null: bool = True) -> Optional[str]:
             raise ServerMisconfiguredError("A required etcd config is missing.", key)
         return value
 
+    async def get_container_registry(self):
```
fregataa: please add a return type here!

src/ai/backend/manager/models/etcd.py
```diff
@@ -0,0 +1,225 @@
...
+)  # type: ignore[name-defined]
+
+__all__: Sequence[str] = (
+    "ContainerRegistry",
+    "CreateContainerRegistry",
+    "ModifyContainerRegistry",
+    "DeleteContainerRegistry",
+)
+
+ETCD_CONTAINER_REGISTRY_KEY = "config/docker/registry"
```
fregataa: `SharedConfig.get_container_registry()` also need to use this.

src/ai/backend/manager/models/etcd.py
bezzang-dev: How about this?? I used `set_if_set`
https://github.com/lablup/backend.ai/blob/e4e304234e6312e88ad28d775ed99a6fbea314de/src/ai/backend/manager/models/etcd.py#L118-L122

PR #1472, opened 2023-08-10
feat: enhance the logging for agent kernel registry and alloc maps

### LINKED ISSUE #1435: Enhance the logging for agent kernel registry and alloc maps
Recently we observe frequent mismatch of the agent's internal state (kernel registry) and the actual running containers. This often causes bogus `InsufficientResource` errors when creating sessions.

Let's dump the current state of the agent when it happens, and probably periodically as well.
The log should include `kernel_id`, `container_id`, and `occupied_slots`.

Let's also additionally print the allocation maps, where each compute plugin keeps track of the available slots per device.

Another idea: Let's auto-recalculate the alloc maps when `InsufficientResource` happens.

```[tasklist]
### Tasks
- [ ] https://github.com/lablup/backend.ai/pull/1472
```

pderer:
This PR is related to #1435.

This PR is only for logging.
I will make a new PR that auto-recalculate the alloc maps when `InsufficientResource` happens.

When `InsufficientResource` error is raised, we need to find out what went wrong in the resource allocation process.
There is the list to look out for.
- Manager side
  - occupied slots that are the sum of occupied_slots of the agent stored in the DB
- Agent side
  - allocation strategy
  - slot name to be assigned
  - amount of resource to be allocated
  - amount that the compute plugin can allocate to that slot name
  - alloc map of the compute plugin
  - occupied slots calculated from the compute plugin
  - occupied slots summed from resource.txt of containers managed by the agent
- - -
### Manager side
- occupied slots that are the sum of occupied_slots of the agent stored in the DB
스크린샷 2023-08-10 10 41 39

### Agent side
- allocation strategy (e.g. DiscretePropertyAllocMap: insufficient allocatable amount!)
- slot name to be assigned (e.g. with context tag: mem)
- amount of resource to be allocated (e.g. at allocating mem 335544320)
- amount that the compute plugin can allocate to that slot name (e.g. total allocatable amount: 3563917312)
- alloc map of the compute plugin (e.g. (before allocation) compute plugin alloc map:)
- occupied slots calculated from the compute plugin (e.g. (error during allocation) occupied slots from compute plugin alloc map:)
- occupied slots summed from resource.txt of containers managed by the agent (e.g. occupied slots from kernel resources: {'cpu': Decimal('2'), 'mem': Decimal('671088640')})
스크린샷 2023-08-10 10 41 06

## Process for outputting this log
This error could be a problem with the agent registry or a problem with the compute plugin alloc map.
Therefore, in order to output this log, an error situation must be deliberately created.
(_Be aware that the allocation strategy may differ depending on the resource config. The content below is only an example._)
1. create two session
2. modify code in src/ai/backend/agent/alloc_map.py
  - in DiscretePropertyAllocMap Class, _allocate_by_filling function.
  - check below img, line 303, 315-322
  - This makes it unconditionally raise `InsufficientResource` error when creating a new session
스크린샷 2023-08-10 11 31 13

3. restart manager, agent
4. create new session

This is why the error occurred even though the total allocatable amount was greater than the amount to be allocated.

## Conclusion
In any case, this log will help you figure out why the error occurred when you encounter `InsufficientResource` error.


src/ai/backend/agent/agent.py
```diff
@@ -1701,6 +1704,80 @@ async def create_kernel(
...
                             )
+                        except InsufficientResource as e:
+                            await self.produce_event(
+                                DoAgentResourceCheckEvent(
+                                    ctx.agent_id,
+                                )
+                            )
+                            plugin_alloc_map: MutableMapping[
+                                DeviceName,
+                                MutableMapping[SlotName, MutableMapping[DeviceId, Decimal]],
```
fregataa:
```suggestion
                            plugin_alloc_map: dict[
                                DeviceName,
                                dict[SlotName, MutableMapping[DeviceId, Decimal]],
```
https://docs.python.org/3/library/typing.html#typing.MutableMapping
Since we are just using `dict` here and `MutableMapping` has been deprecated since version 3.9, let's just use `dict`.
you can leave the type of existing `allocations`

src/ai/backend/agent/agent.py
```diff
@@ -1701,6 +1704,80 @@ async def create_kernel(
...
+                            ] = {
+                                dev: dict(self.computers[dev].alloc_map.allocations)
+                                for dev in ordered_dev_names
+                            }
+                            original_plugin_alloc_map = copy.deepcopy(plugin_alloc_map)
+                            for device_name, device_dict in plugin_alloc_map.items():
+                                for temp_slot_name, slot_dict in device_dict.items():
+                                    if temp_slot_name in e.args[5]:
```
fregataa: There are many codes like `e.args[i]` and it can be confusing.
How about initiate `InsufficientResource` with explicit arguments and refer them?

src/ai/backend/agent/alloc_map.py
```diff
@@ -499,9 +502,9 @@ def _allocate_by_filling(
         context_tag: Optional[str] = None,
         min_memory: Decimal = Decimal(0.01),
     ) -> Mapping[SlotName, Mapping[DeviceId, Decimal]]:
-        allocation = {}
+        allocation: MutableMapping[SlotName, MutableMapping[DeviceId, Decimal]] = {}
         for slot_name, alloc in requested_slots.items():
-            slot_allocation: MutableMapping[DeviceId, Decimal] = {}
+            slot_allocation: MutableMapping[DeviceId, Decimal] = defaultdict(Decimal)
```
fregataa: same here. we can use `dict` for the type of `allocation`

src/ai/backend/agent/agent.py
```diff
@@ -1701,6 +1704,80 @@ async def create_kernel(
...
+                            )
+                            plugin_alloc_map: MutableMapping[
+                                DeviceName,
+                                MutableMapping[SlotName, MutableMapping[DeviceId, Decimal]],
+                            ] = {
+                                dev: dict(self.computers[dev].alloc_map.allocations)
+                                for dev in ordered_dev_names
+                            }
+                            original_plugin_alloc_map = copy.deepcopy(plugin_alloc_map)
+                            for device_name, device_dict in plugin_alloc_map.items():
```
fregataa: How about just iterate allocations from the exception? I think that we do not need to iterate all the computer devices and copy all of them.
And if you want to iterate all computer devices here, it would be better to iterate the devices and set those key and value into `plugin_alloc_map` and `original_plugin_alloc_map` one by one rather than using list comprehension and copy them because we need at least two iterations here and `deepcopy()` causes additional recursions that can be slower than normal iterations.

fregataa (commented):
Thanks!
I left some comments.

src/ai/backend/agent/agent.py
```diff
@@ -1701,6 +1704,80 @@ async def create_kernel(
...
+                                " slots from kernel resources: {7}",
+                                e.args[0],
+                                e.args[1],
+                                e.args[2],
+                                e.args[3],
+                                e.args[4],
+                                pprint.pformat(original_plugin_alloc_map),
+                                pprint.pformat(plugin_alloc_map),
+                                occupied_slots_from_kernel,
+                            )
```
achimnol: Since `InsufficientResource` now has so many arguments, how about redefining it to use named arguments and fields with explicit type annotations?

Example:
```python
from dataclasses import dataclass

@dataclass
class CustomException(Exception):
    foo: str
    baz: int

    def __reduce__(self):  # make it pickle-safe
        return (self.__class__, (self.foo, self.baz))
```

Also, even when formatting with Black, the long-running multi-line strings could be split into more "semantic" lines for better readability, e.g.,
```suggestion
                            log.error(
                                "resource allocation failed (InsufficientResource)\n"
                                "{0}\n"
                                "with context tag: {1}\n"
                                "at allocating {2} {3}\n"
                                "total allocatable amount: {4}\n"
                                "-----------------------------------------------\n"
                                "(before allocation) compute plugin alloc map:\n"
                                " {5}\n"
                                "-----------------------------------------------\n"
                                "(error during allocation) occupied slots from "
                                "compute plugin alloc map:\n"
                                " {6}\n"
                                "-----------------------------------------------\n"
                                "occupied slots from kernel resources: {7}",
                                e.args[0],
                                e.args[1],
                                e.args[2],
                                e.args[3],
                                e.args[4],
                                pprint.pformat(original_plugin_alloc_map),
                                pprint.pformat(plugin_alloc_map),
                                occupied_slots_from_kernel,
                            )
```

achimnol (changes requested):

src/ai/backend/agent/agent.py
pderer: Thanks!
I will fix it together with other code 👍 
I have question about `Mapping`.
`Mapping` has been deprecated too.
So all `Mapping` can change to `dict`?
Or is it okay not to change it because there is nothing wrong with not changing it?

https://github.com/lablup/backend.ai#python-version-compatibility
According to the Python version compatibility, since backend.ai version 22.03 supports python 3.10 or higher, it seems possible to change both MutableMapping and Mapping to dict, is that correct?


`Mapping` example:
https://github.com/lablup/backend.ai/blob/main/src/ai/backend/agent/alloc_map.py#L181-#L187

src/ai/backend/agent/alloc_map.py
pderer: Thanks!
I will fix it 👍

src/ai/backend/agent/agent.py
pderer: This code attempts increment while assigning key and value one by one without using deepcopy.
```python
plugin_alloc_map: dict[
    DeviceName, dict[SlotName, MutableMapping[DeviceId, Decimal]]
] = {}
original_plugin_alloc_map: dict[
    DeviceName, dict[SlotName, MutableMapping[DeviceId, Decimal]]
] = {}
for dev in ordered_dev_names:
    plugin_alloc_map[dev] = dict(
        self.computers[dev].alloc_map.allocations
    )
    original_plugin_alloc_map[dev] = dict(
        self.computers[dev].alloc_map.allocations
    )
for temp_slot_name, temp_slot_dict in e.args[5].items():
    print(temp_slot_name)
    print(temp_slot_dict)
    print("------------")
    for device_id in temp_slot_dict.keys():
        if (
            device_id
            in plugin_alloc_map[dev_name][temp_slot_name].keys()
        ):
            print(
                "id(plugin_alloc_map[dev_name][temp_slot_name]): ",
                id(plugin_alloc_map[dev_name][temp_slot_name]),
            )
            print(
                "id(original_plugin_alloc_map[dev_name][temp_slot_name]): ",
                id(original_plugin_alloc_map[dev_name][temp_slot_name]),
            )
            print(
                "id(plugin_alloc_map[dev_name][temp_slot_name][device_id]): ",
                id(
                    plugin_alloc_map[dev_name][temp_slot_name][
                        device_id
                    ]
                ),
            )
            print(
                "id(original_plugin_alloc_map[dev_name][temp_slot_name][device_id]): ",
                id(
                    original_plugin_alloc_map[dev_name][temp_slot_name][
                        device_id
                    ]
                ),
            )
            print("----increment processing----")
            plugin_alloc_map[dev_name][temp_slot_name][
                device_id
            ] += temp_slot_dict[device_id]
            print(
                "id(plugin_alloc_map[dev_name][temp_slot_name]): ",
                id(plugin_alloc_map[dev_name][temp_slot_name]),
            )
            print(
                "id(original_plugin_alloc_map[dev_name][temp_slot_name]): ",
                id(original_plugin_alloc_map[dev_name][temp_slot_name]),
            )
            print(
                "id(plugin_alloc_map[dev_name][temp_slot_name][device_id]): ",
                id(
                    plugin_alloc_map[dev_name][temp_slot_name][
                        device_id
                    ]
                ),
            )
            print(
                "id(original_plugin_alloc_map[dev_name][temp_slot_name][device_id]): ",
                id(
                    original_plugin_alloc_map[dev_name][temp_slot_name][
                        device_id
                    ]
                ),
            )
            print("----------")
            print("original_plugin_alloc_map: ")
            pprint.pprint(original_plugin_alloc_map)
            print("plugin_alloc_map: ")
            pprint.pprint(plugin_alloc_map)
```
print results:
스크린샷 2023-08-15 01 57 12

It's example that allocate `{'cpu': {'2': Decimal('1')}`.
As you can see, `id(plugin_alloc_map[dev_name][temp_slot_name])` == `id(original_plugin_alloc_map[dev_name][temp_slot_name])`.
The id according to the device name is different because it is assigned according to the device name key, but the slot name points to the same id. (because it's nested dict)
So after increment, we can see that `original_plugin_alloc_map` and `plugin_alloc_map` have the same value.

This is the result when using deep copy.
스크린샷 2023-08-15 02 15 45
As you can see, `id(plugin_alloc_map[dev_name][temp_slot_name])` != `id(original_plugin_alloc_map[dev_name][temp_slot_name])`.
So after increment, we can see that `original_plugin_alloc_map` and `plugin_alloc_map` have different value. (cpu, device_id == 2)

So I decided to use deepcopy.
Please let me know if there is any other option!

pderer: Thanks!
Modified the code to use the least amount of deepcopy().
And just iterate allocations from the exception to efficiently construct the for loop.
plz check 👍

src/ai/backend/agent/agent.py
pderer: Thanks!
I admit that the readability is not good.
I modified `InsufficientResource` to have fields and type annotations.
plz check 👍

src/ai/backend/agent/agent.py
pderer: Thanks!
I modified `InsufficientResource` to have fields and type annotations.
And for readability, the black formattor was temporarily disabled.
plz check 👍

PR #1480, opened 2023-08-15
feature: add support for kmanila storage backend

fregataa:
https://github.com/lablup/giftbox/issues/480
Kmanila is an abbreviation for KTC's Manila NAS API.


src/ai/backend/storage/kmanila/__init__.py
```diff
@@ -0,0 +1,272 @@
...
+                    },
+                    "scope": {
+                        "project": {
+                            "domain": {"id": "default"},
+                            "name": rqst_id,
+                        }
+                    },
+                }
+            }
+            headers = {"Content-Type": "application/json", "Accept": "application/json"}
```
kyujin-cho: when passing request body with `json` parameter aiohttp automatically sets `Content-Type` header to `application/json`.
```suggestion
            headers = {"Accept": "application/json"}
```

src/ai/backend/storage/kmanila/__init__.py
```diff
@@ -0,0 +1,272 @@
...
+        return None
+
+    async def _create_volume(
+        self,
+        session: aiohttp.ClientSession,
+        quota_scope_id: QuotaScopeID,
+        options: QuotaConfig | None,
+        auth_info: dict[str, Any],
+        *,
+        name: str,
```
kyujin-cho, line 152: By the leading underscore character I guess that this function is only for internal use rather than a public interface. In this case is it really necessary to allow any kind of parameters? I mean, do we really need that asterisk?

src/ai/backend/storage/kmanila/__init__.py
```diff
@@ -0,0 +1,272 @@
...
+            }
+        }
+        async with session.post(
+            url=f"/d3/adm/manila/v2/{project_id}/shares", json=request_body
+        ) as resp:
+            if resp.status not in (200, 201, 204):
+                raise ExternalError(
+                    f"Got invalid status code when post data to API server. {resp.status = }"
+                )
+            await asyncio.sleep(0)
```
kyujin-cho: Could you explain what this sleep call is for?

src/ai/backend/storage/kmanila/__init__.py
```diff
@@ -0,0 +1,272 @@
...
+            self.mount_path,
+            self.etcd,
+            URL(self.config["api_base_url"]),
+            self.config["agent_addrs"],
+            access_to=self.config["access_to"],
+            access_level=access_level,
+            kmanila_requestor_id=self.config["user_id"],
+            kmanila_requestor_pwd=self.config["user_password"],
+            default_project_id=self.config.get("project_id"),
+            default_netword_id=self.config.get("netword_id"),
```
kyujin-cho: Just to double check, is `netword` a typo of `network` or is it intended?

src/ai/backend/storage/kmanila/__init__.py
```diff
@@ -0,0 +1,272 @@
...
+    async def get_user_volume_id(self, quota_scope_id: QuotaScopeID) -> str | None:
+        user_vol = await self.etcd.get_prefix(
+            f"{QUOTA_VOLUME_ID_MAP_KEY}/{quota_scope_id.scope_id}"
+        )
+        if not user_vol:
+            return None
+        return str(user_vol["volume_id"])
+
+    async def put_user_volume_id(self, quota_scope_id: QuotaScopeID, data: QuotaVolumeMap) -> None:
+        await self.etcd.put_prefix(f"{QUOTA_VOLUME_ID_MAP_KEY}/{quota_scope_id.scope_id}", data)  # type: ignore[arg-type]
```
kyujin-cho: `type: ignore` annotation should be avoided in best effort. If there is a critical reason to bypass the type checker please describe the reason at the line.

src/ai/backend/storage/kmanila/__init__.py
```diff
@@ -0,0 +1,272 @@
...
+                await self.put_user_volume_id(
+                    quota_scope_id, QuotaVolumeMap(volume_id=volume_id, volume_name=name)
+                )
+                return volume_id
+            await asyncio.sleep(1)
+            trial += 1
+            if trial > self.max_poll_count:
+                raise ExternalError(f"Poll trial exceeds the maximum trial count, {trial = }")
+
+    async def _create_access_control(
```
kyujin-cho, line 192: By only reading the method name I can't get what this function actually does. Please update the name to be more self-explanatory, perhaps like `_allow_access_to_volume()`?

src/ai/backend/storage/kmanila/__init__.py
```diff
@@ -0,0 +1,272 @@
...
+
+    async def _create_volume(
+        self,
+        session: aiohttp.ClientSession,
+        quota_scope_id: QuotaScopeID,
+        options: QuotaConfig | None,
+        auth_info: dict[str, Any],
+        *,
+        name: str,
+    ) -> str:
```
kyujin-cho: If you were to return value of intrinsic type the callee can be confused by what this function actually returns. Please write down the comment about what this actually does. This comment also applies to `_create_access_control()`.

src/ai/backend/storage/kmanila/__init__.py
```diff
@@ -0,0 +1,272 @@
...
+        project_id = auth_info["project_id"]
+        request_body = {
+            "share": {
+                "share_proto": "nfs",
+                "share_network_id": self.default_netword_id,
+                "name": name,
+                "is_public": False,
+                "size": options.limit_bytes if options is not None else DEFAULT_VOLUME_SIZE,
+                "availability_zone": "DX-DCN-CJ",
+                "share_type": "SSD",
```
kyujin-cho: Is it okay to leave these values hard-coded? Just asking.

src/ai/backend/storage/api/manager.py
```diff
@@ -168,13 +170,15 @@ class Params(TypedDict):
                     t.Key("qsid"): tx.QuotaScopeID(),
                     t.Key("options", default=None): t.Null | QuotaConfig.as_trafaret(),
                 },
-            ),
+            ).allow_extra("*"),
```
kyujin-cho: I think it would be much more better to explicitly define nullable `extra_args` unless there is a specific reason not to do so.

kyujin-cho (changes requested):
First batch of reviews; also please don't forget to update `install-dev.sh`'s `storage-proxy.toml` commenting logic.

src/ai/backend/storage/kmanila/__init__.py
fregataa: I forgot to delete this part

src/ai/backend/storage/kmanila/__init__.py
fregataa: wow what a huge mistypo. I fixed it!

src/ai/backend/storage/kmanila/__init__.py
fregataa: Since the API spec of the customer represent it as "Creating access control", I leave a docstring on this function rather than changing the name.
https://cloud.kt.com/docs/open-api-guide/d/storage/nas

src/ai/backend/storage/kmanila/__init__.py
fregataa: Move them to config.toml
thank you for checking it!

src/ai/backend/storage/kmanila/__init__.py
fregataa: it does not allow any parameters. Parameters after "*(asterisk)" or "*identifier" are keyword-only parameters and may only be passed by keyword arguments.
https://docs.python.org/3/reference/compound_stmts.html#function-definitions

kyujin-cho: Oh you can ignore this one. I did delete this review but this somehow managed to upload

src/ai/backend/storage/kmanila/__init__.py
kyujin-cho: Yup that would be a good reason to do so.

PR #1481, opened 2023-08-16
feat: add redis event client to storage proxy

fregataa:
Add the redis client to the server context because there may be enough cases where the agent consumes events, and the storage proxy can produce and consume events.

1. Add event producer/dispatcher to storage proxy
2. Add event dispatcher to agent


src/ai/backend/common/types.py
```diff
@@ -1125,3 +1126,12 @@ class AcceleratorMetadata(TypedDict):
...
     display_icon: str
+
+
+# Redis database IDs depending on purposes
+REDIS_STAT_DB: Final = 0
+REDIS_RLIM_DB: Final = 1
+REDIS_LIVE_DB: Final = 2
+REDIS_IMAGE_DB: Final = 3
+REDIS_STREAM_DB: Final = 4
+REDIS_STREAM_LOCK: Final = 5
```
kyujin-cho: Now that we have these DB definitions on a `common` package please remove duplicate declarations at `ai.backend.manager.defs` and migrate all usages referencing old constants under `.defs`  to these newer ones.

kyujin-cho (changes requested):

kyujin-cho (approved):

PR #1495, opened 2023-08-18
feat: add support for storage watcher

fregataa:
1. Mount volumes on Agent and Storage nodes via event.
2. Impl storage proxy watcher that delegate root permission and do privileged tasks such as mount volumes or chown.
3. Add "mount-path" config to agent to recognize and change mount point easily

**kmanila** storage backend has been migrated to another repo(plugin)


achimnol:
This PR will be moved into lablup/backend.ai-storage-kmanilla#1.

PR #1497, opened 2023-08-20
fix: add missing `kmanila_` prefix for configuration keys

chisacam:
- resolves https://github.com/lablup/backend.ai-ossca-2023/issues/18
- i'm not sure about i changed all of names as it need, please review it and told me it is require changes or not, or need more changes. i'll check and more edit or revert it.
- change default etcd port to i heard correct number.


fregataa (approved):
Thanks!
Plus, it will be nice to update `install-dev.sh` to replace etcd port in `storage-proxy.toml` from 8121(or 2379) into `ETCD_PORT`.
You can use `sed_inplace`.

fregataa (changes requested):
please add a news fragment file for timeline checker!

configs/storage-proxy/sample.toml
```diff
@@ -1,6 +1,6 @@
 [etcd]
 namespace = "local"                         # env: BACKEND_NAMESPACE
-addr = { host = "127.0.0.1", port = 2379 }  # env: BACKEND_ETCD_ADDR (host:port)
+addr = { host = "127.0.0.1", port = 8121 }  # env: BACKEND_ETCD_ADDR (host:port)
```
kyujin-cho: `2379` is de-facto standard for Etcd server's listening port number. Please just leave it as-is and let's create a separate issue to create a storage-proxy version of `halfstack.toml` config file.

kyujin-cho (changes requested):

configs/storage-proxy/sample.toml
chisacam: sure, i'll revert it..
how about changes in `install-dev.sh`? am i revert that too?

kyujin-cho: Yes please.

kyujin-cho:
Closing as we made a decision to lift up kmanila integration from the monorepo and store it as a separated plugin repository.

PR #1499, opened 2023-08-20
misc: add script to check commits not yet backported

kyujin-cho:
This PR introduces `check-backport-commits.py` script, which checks if there is a merge commit marked as a backport target but still yet did its job. This scripts `gh` CLI to be installed and configured at the machine.    
To run script simply boot up the python runtime with `python scripts/check-bacport-commits.py <Target version to be released very soon>`. 
image

### DIFF
--- a/scripts/check-backport-commits.py
+++ b/scripts/check-backport-commits.py
@@ -1,0 +1,23 @@
+import json
+import subprocess
+import sys
+
+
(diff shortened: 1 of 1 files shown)


kyujin-cho:
Though this is not a bugfix or guaranteed new feature, I am backporting this commit to ease the burden of switching branches when using this script to backport commits to 23.03.

PR #1510, opened 2023-08-25
fix: Ensure rollback of modified alloc maps when ResourceError occurs

achimnol:
This PR forifies and refactors out the allocation of per-device resource slots when creating a new kernel in the agent.

Previously, when there was a failure like `InsufficientResource` in the middle of the for-loop to allocate requested resource slots device by device, the prior devices' allocations were NOT rolled back to their initial state.

This has caused silent mismatch of the resource slots in the alloc maps and the occupation observed via the running containers in agents, resulting in continuous insufficient resource errors afterwards.

Now this PR provides a test to distinguish the behavioral difference depending on whether rollback is done or not, as well as refactoring out the pure logic part of allocation as `ai.backend.agent.resources.allocate()`.


achimnol:
22.09 backport is postponed because we need to backport additional features such as the dummy agent backend.

Issue #1443, opened 2023-08-01
Improve Redis sentinel support in the manager and agents


achimnol:
The underlying wrapper (`common.redis_helper`) is tested hard to run with Redis sentinels, but the current manager and agent's redis initialization routines are not perfectly compatible with it.

Let's allow manager & agent use the Redis sentinels.

achimnol:
This issue is replaced with rewriting the redis client: https://github.com/lablup/qedis

PR #1513, opened 2023-08-28
fix: Improve Redis sentinel support in the manager and agents (WIP)

jopemachine:
Aim to resolve https://github.com/lablup/backend.ai/issues/1443#event-10029601945.

WIP.


achimnol (changes requested):
92xx ports are used to run the test suite (`pants test tests/common/redis_helper::`).
We should use other range to avoid conflicts with locally executed tests.

src/ai/backend/common/redis_helper.py
```diff
@@ -485,34 +447,44 @@ def get_redis_object(
...
             sentinel_kwargs={
+                "password": password,
                 **kwargs,
             },
         )
+
         return RedisConnectionInfo(
-            client=sentinel,
-            service_name=redis_config.get("service_name"),
+            client=sentinel.master_for(service_name=service_name, **_default_conn_opts),
```
achimnol: I think we should preserve `socket_connect_timeout=reconnect_poll_interval` kwarg here.

src/ai/backend/storage/server.py
achimnol, line 1: 👍🏼 Good refactoring to use the standard way.
Note that we need to check whether the log arguments are picklable, though.  If the storage proxy is started well, there should be no problem in this case.

achimnol (changes requested):

achimnol (changes requested):
Let's place the `redis_helper` shared-config key under the `redis` shared-config key, to prevent changing every call to `get_redis_object()` in the entire codebase. Pleaes consider how to minimize such interface changes.

achimnol:
Due to PR/git branch mismatch issue, this is replaced with #1586.

PR #1517, opened 2023-08-29
fix: enforce `vfolder.delete_by_id()` to check type of `id` parameter

rapsealk:
1. `delete_by_id()` requires an superadmin privilege and accepts a string-typed parameter `id`.
https://github.com/lablup/backend.ai/blob/ca7c7eb3e70c1a923d6b21cbcc384a24f6d836ec/src/ai/backend/manager/api/vfolder.py#L639-L649

2. `ensure_vfolder_status()` accepts both `str` and `uuid.UUID` for `folder_id_or_name`, which equals to the `id` from above.
https://github.com/lablup/backend.ai/blob/ca7c7eb3e70c1a923d6b21cbcc384a24f6d836ec/src/ai/backend/manager/api/vfolder.py#L118-L127

3. The problem is that it compares the `folder_id_or_name` to `vfolders.c.name`, not `vfolders.c.id`.
https://github.com/lablup/backend.ai/blob/ca7c7eb3e70c1a923d6b21cbcc384a24f6d836ec/src/ai/backend/manager/api/vfolder.py#L134-L139

Therefore, this PR fixes to convert a string-typed parameter `id` to `uuid.UUID` type when calling `ensure_vfolder_status()`.

Furthermore, I found that some database executions do not work as expected. The query below returns just a single-column:
```python
query = (
    sa.select(
        [
            vfolders.c.quota_scope_id,
            vfolders.c.host,
        ]
    )
    .select_from(vfolders)
    .where(vfolders.c.id == params["id"])
)
quota_scope_id, folder_host = await conn.scalar(query)

#     quota_scope_id, folder_host = await conn.scalar(query)
#     ^^^^^^^^^^^^^^^^^
# TypeError: cannot unpack non-iterable QuotaScopeID object
```

So I changed the execution code as below:
```python
query = (
    sa.select(
        [
            vfolders.c.quota_scope_id,
            vfolders.c.host,
        ]
    )
    .select_from(vfolders)
    .where(vfolders.c.id == params["id"])
)
result = await conn.execute(query)
quota_scope_id, folder_host = result.first()
```

- [ ] Documentation


fregataa (approved):

src/ai/backend/manager/api/vfolder.py
```diff
@@ -646,7 +646,7 @@ async def list_folders(request: web.Request, params: Any) -> web.Response:
     ),
 )
 async def delete_by_id(request: web.Request, params: Any) -> web.Response:
-    await ensure_vfolder_status(request, VFolderAccessStatus.DELETABLE, params["id"])
+    await ensure_vfolder_status(request, VFolderAccessStatus.DELETABLE, uuid.UUID(params["id"]))
```
kyujin-cho: To get rid of similar human errors in the future, I guess it is better to enforce the type of "id" parameter as `tx.UUID` (instead of current plaintext-ish `t.String` validator). In that way, `params["id"]` itself will be delivered as a form of `UUID()` type instance.

kyujin-cho (changes requested):

src/ai/backend/manager/api/vfolder.py
rapsealk: I was worried about those which already are using this API. If it is okay, I agree that it would be better to change the type to tx.UUID.

kyujin-cho (approved):

PR #1465, opened 2023-08-09
fix: also check for mininum page size when paginating

chisacam:
- resolves https://github.com/lablup/backend.ai-ossca-2023/issues/15
- currently, fetch_paginated_result is not check minimum page_size aka limit.
- so it occur infinity fetch error, this PR is add check page_size(limit) variable block.
- also double check in execute_paginated_query function for safety.

PR #1518, opened 2023-08-29
fix: add default unit 'm' to 'mem' arg in session create cli

jjm2317:
resolves https://github.com/lablup/backend.ai-ossca-2023/issues/17

When running command `./backend.ai session create --help`, the following description of the resource is displayed.

```
 -r, --resources KEY=VAL         Set computation resources used by the session (e.g: -r cpu=2 -r mem=256 -r gpu=1).1 slot of cpu/gpu represents 1 core. The unit
                                  of mem(ory) is MiB.
```

With this pull request, `mem` argument of the --resource option in `./backend.ai session create` is passed with "m" unit when a `mem` is passed without 'm' or 'g'.

- [ ] Documentation


src/ai/backend/client/cli/session/execute.py
```diff
@@ -227,6 +227,8 @@ def format_stats(stats):
 def prepare_resource_arg(resources):
     if resources:
         resources = {k: v for k, v in map(lambda s: s.split("=", 1), resources)}
+        if resources["mem"] and "m" not in resources["mem"] and "g" not in resources["mem"]:
+            resources["mem"] += "m"
```
fregataa, line 231: Units of resources should be handled very strictly and commonly among components because many components shares such resource options and handling resources is critical to our system.

Also, if we also consider cases like shared memory (shmem), we need more general unit handling.
We defined [`BinarySize` type in common](https://github.com/lablup/backend.ai/blob/c0bbaa9012fa3120fa3c657bed963fd96835bc76/src/ai/backend/common/types.py#L434-L577). How about using it??

Plus, I think we can add type hints to this function!

fregataa (changes requested):
Please make news fragment file.
You can check how should file be in other PRs such as https://github.com/lablup/backend.ai/pull/1465/files.

src/ai/backend/client/cli/session/execute.py
jjm2317: Thank you for feedback. I will apply them to this CL!

achimnol:
Replaced by #1650.

PR #1521, opened 2023-08-30
feat: update `ContainerRegistry` mutations' response

rapsealk:
This is a follow-up of #1471
This PR enhances `ContainerRegistry`-related mutations to include affected field rather than `{"result": "ok"}`.

- [ ] Documentation


src/ai/backend/manager/models/etcd.py
```diff
@@ -56,6 +56,13 @@ class ContainerRegistry(graphene.ObjectType):
     hostname = graphene.String()
     config = graphene.Field(ContainerRegistryConfig)
 
+    class Meta:
+        interfaces = (graphene.relay.Node,)
+
+    # @classmethod
+    # def get_node(cls, info: graphene.ResolveInfo, id):
+    #     raise NotImplementedError
+
```
fregataa, line 67: Is this code for graphql node?

fregataa (commented):
Looks good!
I am a bit worried that every gql query will request to etcd, which could increase latency.

src/ai/backend/manager/models/etcd.py
rapsealk: `graphene` requires relay nodes to have `get_node()` class method to be implemented. This method is used to query the object directly by unique id. But yet we do not support this feature, which means it is not necessary.

Please refer: https://docs.graphene-python.org/en/latest/relay/nodes/#nodes

kyujin-cho, line 67: Please remove commented out code blocks. If you were meant to specify that this feature has to be implemented in near future then please describe about it with `TODO:` tag in front of the comment block.

kyujin-cho (changes requested):

Issue #1361, opened 2023-07-03
Easy dev installation script for k8s agent


inureyes:
### Main idea

The basic installation script, `install_dev.sh`, is based on the installation of the backend.ai native agent. However, depending on the situation, you may need to integrate it with a Kubernetes environment. In this case, providing an installation script that can be used without following a complex manual would be helpful in setting up a development environment.

### Alternative ideas

* Expand `install_dev.sh`
* Make another `install_dev_k8s.sh`

PR #1526, opened 2023-08-31
feat: introduce the k8s agent backend mode to install-dev.sh

NayeonKeum:
resolves: #1361 

This pr only includes introducing k8s agent backend mode with cli in install-dev.sh

The reason I added only k8s/kubernetes in the option is that the python kubernetes client uses local kubernetes context in `~/.kube/config`. So, using micro-k8s in local and any external kubernetes cluster makes no difference in configuration.

It just need's it's context added in `~/.kube/config`.

### DIFF
--- a/changes/1526.feature.md
+++ b/changes/1526.feature.md
@@ -1,0 +1,1 @@
+Introduce the k8s agent backend mode to install-dev.sh with --agent-backend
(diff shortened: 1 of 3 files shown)


fregataa (approved):
👍

kyujin-cho:
Closing in favor of #1527

PR #1525, opened 2023-08-30
docs: add instructions about generating PDF document

hanlee55:
Goal: Building pdf document

+ Add Sphinx extension to use SVG img in PDF https://github.com/lablup/backend.ai/commit/c2496fa9d70af564eaf86d890fa98c3d0c410ec9
+ Add guide for building PDF document https://github.com/lablup/backend.ai/commit/3294e3903d5f1f722f44bbe8ef35202bda212655
  - Command
  - Dependency

**Checklist:** (if applicable)

- [ ] Milestone metadata specifying the target backport version
- [ ] Mention to the original issue
- [ ] Documentation
- [ ] API server-client counterparts (e.g., manager API -> client SDK)
- [ ] Test case(s) to demonstrate the difference of before/after

PR #1529, opened 2023-08-31
fix: PDF not generated on macOS

kyujin-cho:
Follow-up PR of #1525. Updates docs/conf.py to switch PDF building engine to xelatex instead of pdflatex.

----
:books: Documentation preview :books:: https://sorna--1529.org.readthedocs.build/en/1529/

----
:books: Documentation preview :books:: https://sorna-ko--1529.org.readthedocs.build/ko/1529/


achimnol (approved):
👀👍

PR #1548, opened 2023-09-07
fix: install-dev.sh failing when node is not installed

kyujin-cho:
Follow-up PR of #1441. Fixed install-dev.sh script failing to execute when trying to execute the script on main branch and node.js is not installed on the system. From now on the script will automatically install NVM along with Node.js v18 when the script can't find the one from system.

- [ ] Documentation


scripts/install-dev.sh
```diff
@@ -435,6 +435,16 @@ install_system_pkg() {
   esac
 }
 
+install_nvm() {
+  show_info "Installing NVM and Node.js v18..."
+  curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.5/install.sh | bash
+  export NVM_DIR="$([ -z "${XDG_CONFIG_HOME-}" ] && printf %s "${HOME}/.nvm" || printf %s "${XDG_CONFIG_HOME}/nvm")"
+  [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" # This loads nvm
+
+  nvm install 18
```
achimnol: How about using `nvm install --lts`?

achimnol (changes requested):

scripts/install-dev.sh
achimnol: Or, we could read the recommended setting from the webui repository.

achimnol: Let's read `https://raw.githubusercontent.com/lablup/backend.ai-webui/main/.nvmrc`.
We could do this either issuing curl command against this URL, or read the local `.nvmrc` file after cloning the webui repository.

achimnol (approved):

PR #1551, opened 2023-09-07
feat: periodically update available slots

kyujin-cho:
This PR rewrites `detect_resources()` logic to split out plugin loading and slot detection algorithm, and then adds a new timer task to periodically scan available slots through all plugins and update if necessary.

- [ ] Documentation


achimnol (approved):
Good. I'd like to expand this PR in the future to handle a local watcher event to remove specific devices from scheduling as well.

PR #1554, opened 2023-09-08
fix: Raise error when fail to acquire `RedisLock`

fregataa:
https://github.com/redis/redis-py/blob/master/redis/asyncio/lock.py#L159-L187
According to `redis-py`, we need to use `redis.asyncio.Lock.__aenter__()` to acquire lock rather than `redis.asyncio.Lock.acquire()` because `acquire()` returns `True` or `False`, not raising an error.

### DIFF
--- a/src/ai/backend/common/lock.py
+++ b/src/ai/backend/common/lock.py
@@ -240,13 +240,13 @@ class RedisLock(AbstractDistributedLock):
             thread_local=False,
             sleep=self._lock_acquire_pause,
         )
-        await self._lock.acquire()
+        await self._lock.__aenter__()
         if self._debug:
             log.debug("RedisLock.__aenter__(): lock acquired")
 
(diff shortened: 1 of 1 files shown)


src/ai/backend/common/lock.py
```diff
@@ -240,13 +240,13 @@ async def __aenter__(self) -> None:
...
         )
-        await self._lock.acquire()
+        await self._lock.__aenter__()
         if self._debug:
             log.debug("RedisLock.__aenter__(): lock acquired")
 
     async def __aexit__(self, *exc_info) -> Optional[bool]:
         assert self._lock is not None
-        await self._lock.release()
+        await self._lock.__aexit__(*exc_info)
```
achimnol: To make it more transparent, we should take and return the return value of inner `__aexit__()` call.

image

achimnol (changes requested):

achimnol (approved):

PR #1560, opened 2023-09-10
feature: impl standalone watcher

fregataa:
implement standalone watcher component.
watcher runs with dynamic load of watcher plugin.
we can set the config file for the plugin by passing config file path as cmd argument.

- [ ] Documentation


fregataa:
Since control panel would be the one that requests to watcher servers, I made an issue about supporting HTTPS in control panel.
https://github.com/lablup/backend.ai-control-panel/issues/674

achimnol (changes requested):
Things to do:
* Migrate `src/ai/backend/agent/watcher.py` as an intrinsic `WatcherWebappPlugin`.
* Add an API to query the activated watcher plugins so that the control panel could show appropriate UI by each watcher.
* Apply #1640.

PR #1561, opened 2023-09-11
fix: modified to avoid repeated type casting with UUID

agatha197:
- follows https://github.com/lablup/backend.ai/pull/1517
- resolves https://github.com/lablup/backend.ai-control-panel/issues/672
- AttributeError occurs because folder id is typecast to UUID once again, even though folder id is entered as UUID.
- To Avoid AttributeError and make it possible to delete by vfolder id, I removed the repeated type casting with UUID.

- [ ] Documentation


rapsealk (approved):
LGTM!

kyujin-cho (approved):

PR #1570, opened 2023-09-12
fix: Let agents fire storage (un)mount completion events even when coexisting with storage-proxy

fregataa:
When there is any cohabiting storage with agents, agents should skip mount/umount task and just produce succeed event rather than just return.

If https://github.com/lablup/backend.ai/pull/1560 merged earlier than this PR, this should be closed.


achimnol:
Let's apply this first and then look into #1560.

PR #1577, opened 2023-09-14
feat: VAST Data storage backend

fregataa:


src/ai/backend/storage/vast/vastdata_client.py
```diff
@@ -0,0 +1,440 @@
...
+        current_dt = datetime.now()
+        if self._auth_token is None:
+            return await self._login()
+        elif self._auth_token["access_token"]["issued_dt"] + DEFAULT_ACCESS_TOKEN_SPAN > current_dt:
+            return
+        elif (
+            self._auth_token["refresh_token"]["issued_dt"] + DEFAULT_REFRESH_TOKEN_SPAN > current_dt
+        ):
+            return await self._refresh()
+        return await self._login()
```
kyujin-cho, line 181: VAST's API token is formed with JWT format so we can just evaluate `exp` value of JWT payload to determine if token is valid or not.

src/ai/backend/storage/context.py
```diff
@@ -39,6 +39,7 @@
 )
 from .purestorage import FlashBladeVolume
 from .types import VolumeInfo
+from .vast import VastVolume
```
kyujin-cho: We need to call them `VAST`, not `Vast` ([link to their official webpage](https://vastdata.com/)). Capital letters do matter!

src/ai/backend/storage/vast/__init__.py
```diff
@@ -0,0 +1,244 @@
...
+            api_version=self.config["vast_api_version"],
+            ssl=ssl_verify,
+        )
+
+    async def create_quota_model(self) -> VastQuotaModel:
+        return VastQuotaModel(self.mount_path, self.api_client)
+
+    async def get_capabilities(self) -> FrozenSet[str]:
+        return frozenset(
+            [CAP_VFOLDER, CAP_METRIC, CAP_QUOTA, CAP_FAST_FS_SIZE, CAP_FAST_SCAN, CAP_FAST_SIZE]
```
kyujin-cho: `CAP_FAST_SCAN` capability should only be advertised if the storage supports scanning number of files in a "fast" fashion. In here fast means we should not have to scan the whole filesystem and count the number of files one by one, usually by querying the API vender has exposed. But here I don't see any of such custom methods overridden. Purestorage backend's `RapidFileToolsFSOpModel.scan_tree_usage()` can be a good example of such feature.

src/ai/backend/storage/vast/vastdata_client.py
```diff
@@ -0,0 +1,440 @@
...
+                f"quotas/{vast_quota_id}/",
+                body,
+            )
+            data: Mapping[str, Any] = await response.json()
+            match response.status // 100:
+                case 2:
+                    pass
+                case 4 | 5:
+                    err_msg = data.get("detail", "Unkown error from vast API")
+                    raise VastInvalidParameterError(err_msg)
```
kyujin-cho: In theory `Invalid Parameter` error should only be represented by error code 400, not whole 4xx and 5xx status codes. We should discuss whether this kind of approach is legitimate or not.

kyujin-cho (changes requested):

src/ai/backend/storage/vast/vastdata_client.py
fregataa: I updated this part as handling 4xx and 5xx in the same way would be buggy.
The reason why I handled them in the same way is that some invalid API call with wrong parameter or invalid condition among parameters (such as, setting soft limit bigger than hard limit) would responses with status code 500 but it is client's fault.

kyujin-cho (approved):

PR #1579, opened 2023-09-15
fix: lazy raise of vFolder deletion failure

fregataa:
1. Raise `VFolderOperationFailed` after delete vfolders row of successfully deleted vFolders.
2. Pass task name to `PersistentTaskGroup.create_task()` in `initiate_vfolder_removal`

### DIFF
--- a/src/ai/backend/manager/models/vfolder.py
+++ b/src/ai/backend/manager/models/vfolder.py
@@ -931,7 +931,8 @@ async def initiate_vfolder_removal(
                     pass
             except aiohttp.ClientResponseError:
                 failed_deletion.append(vfolder_info)
-            successful_deletion.append(vfolder_info)
+            else:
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/models/vfolder.py
```diff
@@ -926,16 +930,23 @@ async def _delete():
                 ):
                     pass
             except aiohttp.ClientResponseError:
-                raise VFolderOperationFailed(extra_msg=str(folder_id))
+                failed_deletion.append(vfolder_info)
+            successful_deletion.append(vfolder_info)
```
achimnol: This should be put into `else:` block or the end of the `try` block.

achimnol (changes requested):

achimnol (approved):

PR #1582, opened 2023-09-17
fix: delete `vfolders` rows only when `successful_deletion` is not empty

fregataa:
follow-up #1579

1. in the #1579 PR, only successfully deleted vFolders is deleted in DB vfolders table.
let's delete the rows only when `successful_deletion` is not empty
2. delete the rows of not-found vfolders.


src/ai/backend/manager/models/vfolder.py
```diff
@@ -933,19 +933,20 @@ async def _delete():
...
-                await db_session.execute(sa.delete(vfolders).where(cond))
+            async def _delete_row() -> None:
+                async with db_engine.begin_session() as db_session:
+                    await db_session.execute(
+                        sa.delete(vfolders).where(vfolders.c.id.in_(vfolder_ids))
+                    )
 
-        await execute_with_retry(_delete_row)
+            await execute_with_retry(_delete_row)
+            log.debug("Successfully removed vFolders {}", [str(x) for x in vfolder_ids])
```
achimnol: Let's consistently use "vfolder" instead of "vFolders".

src/ai/backend/manager/models/vfolder.py
```diff
@@ -927,14 +929,19 @@ async def _delete():
...
                         "vfid": str(folder_id),
                     },
-                ):
+                ) as (_, resp):
                     pass
-            except aiohttp.ClientResponseError:
-                failed_deletion.append(vfolder_info)
+            except (VFolderOperationFailed, InvalidAPIParameters) as e:
+                if resp.status == 404:
+                    row_deletion_infos.append(vfolder_info)
```
achimnol: This indirect cross reference would confuse static analysis tools because `resp` may not be defined depending on the exact location where the error occurs. `e` already contains the status code as it inherits aiohttp's exception classes.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -890,6 +891,7 @@ async def _update_source_vfolder() -> None:
     return task_id, target_folder_id.folder_id
 
 
+@preserve_termination_log
```
achimnol: If this is a non-long-lived task, it should not be included `@preserve_termination_log` because it will make the termination log to be on the memory forever.  This will keep all of the call history of this method forever until restarting the manager, and as the user/admin tries to delete vfolders over time, this will become the source of memory leak.

achimnol (changes requested):

src/ai/backend/manager/models/vfolder.py
achimnol: In this case I'd advise you to use the standard logging facility upon unexpected exceptions.

achimnol (approved):

PR #1583, opened 2023-09-18
fix: handle redis LockError when release

fregataa:
If `RedisLock` fails to release the lock, such as if the lock has already been released or if it does not have the lock, it raises a `LockError`, as it does when it fails to acquire the lock.
Let's handle the `LockError` as we would when acquire the lock. 


src/ai/backend/common/lock.py
```diff
@@ -250,7 +250,14 @@ async def __aenter__(self) -> None:
...
         assert self._lock is not None
-        val = await self._lock.__aexit__(*exc_info)  # type: ignore[func-returns-value]
+        try:
+            val = await self._lock.__aexit__(*exc_info)  # type: ignore[func-returns-value]
+        except LockNotOwnedError:
+            log.exception("Lock no longer owned. Skip.")
+            return None
+        except LockError:
+            log.exception("Already unlocked. Skip.")
+            return None
```
achimnol: If we want to suppress this exception, how about returning a true value?
ref: https://docs.python.org/3.11/reference/compound_stmts.html#the-with-statement

achimnol: Or, we should check if the lock is properly acquired first and skip the unlocking operation entirely in the `__aexit__()` handler if it's not acquired.

fregataa: checking the lock is acquired in Redis lock would break the abstraction.
IMO, returning `True` makes sense!

achimnol (approved):

PR #1587, opened 2023-09-19
refactor: Merge `kernels.role` into `sessions.session_type`

### LINKED ISSUE #1446: Rename `KernelRole` to something else with clarification that it's for sessions

fregataa:
This PR resolves #1446 and supersedes #1455.

According to the predefined concept, the `kernels.role` field should be replaced by the `sessions.session_type` field.  Also check if the image's `ai.backend.role` label is compatible with the session type when the label is defined.

ref) https://docs.backend.ai/en/latest/concepts/computing.html#session-types

This PR also fixes wrong cases to decrement the per-access-key session concurrency when force-terminating non-running sessions, by skipping decrements for TERMINATED and CANCELLED sessions.

----
📚 Documentation preview 📚: https://sorna--1587.org.readthedocs.build/en/1587/

----
📚 Documentation preview 📚: https://sorna-ko--1587.org.readthedocs.build/ko/1587/


achimnol (changes requested):
Let's migrate the postgres enum to use the new `manager.models.base.StrEnumType` to avoid boilerplates when adding/removing enum items in the postgres schema. Refer #1936.

src/ai/backend/manager/models/kernel.py
```diff
@@ -303,6 +294,14 @@ def default_hostname(context) -> str:
 }
 
 
+SESSION_TYPE_IMAGE_ROLE_MAP: Mapping[SessionTypes, str] = {
+    SessionTypes.BATCH: "COMPUTE",
+    SessionTypes.INTERACTIVE: "COMPUTE",
+    SessionTypes.INFERENCE: "INFERENCE",
+    SessionTypes.SYSTEM: "SYSTEM",
+}
```
achimnol: Why is this needed if we migrate to `enum.StrEnum`?

src/ai/backend/manager/models/session.py
```diff
@@ -166,6 +168,8 @@ class SessionStatus(enum.Enum):
     )
 )
 
+PRIVATE_SESSION_TYPES = (SessionTypes.SYSTEM,)
```
achimnol, line 207: As a follow-up work, this should be refactored to handle "special type" sessions like SFTP containers which allow direct access without AppProxy.

achimnol (changes requested):

src/ai/backend/manager/models/kernel.py
fregataa: I don't get the reason or advantage of using `enum.StrEnum` here rather than `Mapping`.
As I understand, enum is to represent type, not to map data.

achimnol: `StrEnum` fields are string values that are also bidirectionally convertible from/to the Enum instances. There is **no need** to have an extraneous mapping from StrEnum instances to their str values **at all.**

achimnol: Read the manual:
- https://docs.python.org/3/library/enum.html#enum.StrEnum
- https://docs.python.org/3/library/enum.html#notes

achimnol: Ah, this mapping is for:
image

It seems to be an equivalent but **unnecessarily reversed** logical expression.
It should be a mapping from the label value to a `SessionTypes` value and compared with `session_type`.
It would be better to express the purpose and intention in the name of mapping, like `ALLOWED_...`.

Suggestion 1:
```python
            # Check if the image is available for a given session type.
            if (_img_role := labels.get("ai.backend.role")) is not None:
                if _img_role not in ALLOWED_IMAGE_ROLES_FOR_SESSION_TYPE[session_type]:
                    raise InvalidAPIParameters(
                        f"Cannot create {session_type} session with the given image. (img:"
                        f" {image_ref.name}, img role: {_img_role})"
                    )
```

Suggestion 2:
```python
            # Check if the image is available for a given session type.
            if (_img_role := labels.get("ai.backend.role")) is not None:
                if session_type not in ALLOWED_SESSION_TYPES_FOR_IMAGE_ROLE[_img_role]:
                    raise InvalidAPIParameters(
                        f"Cannot create {session_type} session with the given image. (img:"
                        f" {image_ref.name}, img role: {_img_role})"
                    )
```

achimnol: We also need to consider the cases when images have multiple roles.

achimnol:
We need to do a follow-up fix for https://github.com/lablup/backend.ai/issues/2746#issuecomment-2364138900 after merging the stack including #2848.

achimnol (approved):

PR #1589, opened 2023-09-21
feat: Improve the resource metadata API to return actually existent slots only

achimnol:
- feat: Include the resource metdata for only reported and available resource slots

- [ ] Documentation


agatha197 (approved):
LGTM!

PR #1592, opened 2023-09-22
fix: Add `pipeline.endpoint` default value to webserver halfstack.conf

rapsealk:
This PR adds a default value for `pipeline.endpoint` to configs/webserver/halfstack.conf to enable run right after fresh install without no changes.

Refs
- #1528

- [ ] Documentation


achimnol (changes requested):
Please add field descriptions to `configs/webserver/sample.conf`.

configs/webserver/sample.conf
```diff
@@ -95,8 +95,10 @@ max_file_upload_size = 4294967296
 #page = ""
 
 [pipeline]
+# Endpoint to the pipeline service
 #endpoint = "http://mlops.com:9500"
```
achimnol: It is not safe to use arbitrary URLs that may actually exist and get hooked by malicious entities.  Use `example.com`, `localhost`, or a URL with a non-existent TLDs.

achimnol (changes requested):

configs/webserver/sample.conf
achimnol: FYI:
- https://www.iana.org/help/example-domains
- https://www.iana.org/assignments/special-use-domain-names/special-use-domain-names.xhtml

achimnol (approved):

PR #1597, opened 2023-09-25
feature: revamp dummy agent to have compute plugins indivisually

fregataa:
follow-up: https://github.com/lablup/backend.ai/pull/1313

Let dummy agents to have compute plugins indivisually.

# How to use
## Set `agent.toml`
Check `configs/agent/sample-dummy.toml` for full content.
```toml
# File name "agent1.toml"
[agent]
mode = "dummy" # or backend = "dummy"
...

[dummy.agent.intrinsic.cpu]
core-indexes = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]

[dummy.agent.intrinsic.memory]
size = 50000000000

[dummy.agent.delay]
scan-image = 0.1
pull-image = 0.1
destroy-kernel = 0.1
clean-kernel = 0.1
create-network = 0.1
destroy-network = 0.1


[dummy.agent.image.already-have]
'cr.backend.ai/multiarch/python:3.9-ubuntu20.04' = ''


[dummy.kernel-creation-ctx.delay]
prepare-scratch = 0.1
prepare-ssh = 0.1
spawn = 0.1
start-container = 2.0
mount-krunner = 0.1


[dummy.kernel]
use-fake-code-runner = true


[dummy.kernel.delay]
check-status = 0.1
get-completions = 0.1
get-logs = 0.1
interrupt-kernel = 0.1
start-service = 0.1
start-model-service = 0.1
shutdown-service = 0.1
commit = 10.0
get-service-apps = 0.1
accept-file = 0.1
download-file = 0.1
download-single = 0.1
list-files = 0.1
```

Then run agent.
```sh
./backend.ai ag start-server --debug -f agent1.toml
```

## How to run multiple agents
If you want to run multiple dummy agents in one machine, you need multiple "agent.toml" files. Copy and paste the original "agent.toml" file and change `rpc-listen-addr`, `agent-sock-port`, `id` and `pid-file`.

```toml
# File name "agent2.toml"
[agent]
mode = "dummy"
rpc-listen-addr = { host = "127.0.0.1", port = 6012 }
agent-sock-port = 6008
id = "dummy-agent2"
pid-file = "./agent2.pid"
ipc-base-path = "/tmp/backend.ai/ipc"
var-base-path = "var/lib/backend.ai"
...
```
Then run agent.
```sh
./backend.ai ag start-server --debug -f agent2.toml
```

## How to set compute devices.
Dummy agents can set up heterogeneous computing devices using mock accelerators just like regular agents. For more information about how to use mock accelerators, see `src/ai/backend/accelerator/mock/README.md`.

# How it looks
image
Can check mock accelerator as "GPU" resource here.
It is possible to {create/delete} session with delays we set in `agent.toml` file.

# Expectation
- Easy load-test
- Easy to test GUI with multiple heterogeneous agents

- [x] Documentation


kyujin-cho:
Perhaps we can find a way to integrate our mock accelerator plugin with dummy agent.

agent1.toml
achimnol, line 1: Please remove this file if not intended to include.

src/ai/backend/agent/dummy/config.py
```diff
@@ -1,38 +1,39 @@
...
-            t.Key("cpu"): t.Dict({
-                t.Key("core-indexes", default=core_idx): tx.ToSet,
-            }),
-            t.Key("memory"): t.Dict({
-                t.Key("size", default=34359738368): t.Int,
-            }),
+            t.Key("already-have", default=None): t.Null
+            | t.Mapping(
+                t.String, t.String(allow_blank=True)
+            ),  # Key: a string of image canonical, Value: hash. it can be a random string.
```
achimnol: Please include the concrete example of this value structure in the sample toml.

configs/agent/sample.toml
```diff
@@ -370,3 +370,70 @@ size-limit = "64M"
...
+# spawn = 0.1
+# start-container = 2.0
+# mount-krunner = 0.1
+
+
+# Set dummy agent's image list.
+# [dummy.agent.image]
+# already-have = []
+# need-to-pull = []
+# missing = []
```
achimnol: `sample-dummy-config.toml` has
```toml
[agent.image.already-have]
'cr.backend.ai/multiarch/python:3.9-ubuntu20.04' = ''
```
and it breaks up the starting of the agent if I copy `sample-dummy-config.toml` to `agent.toml`.
I think it would be better to have an example configuration that can be copied and used out of the box.

```
> ./backend.ai ag start-server --debug
ConfigurationError: Validation of agent local config has failed:
{'agent': {'image': {'already-have': {0: 'value should be None',
                                      1: 'value is not a dict'}},
           'intrinsic': 'is required',
           'resource': 'resource is not allowed key'}}
Aborted!%
```

Please correct the sample if it needs to be updated.

achimnol (changes requested):

jopemachine:
@fregataa Dynamically generating multiple `agent.toml` files for running multiple agents could be a cumbersome task.

I think it would be convenient if a single `agent.toml` file could be used to run multiple agents by enabling the necessary values for each agent to be overridden with environment variables.
(`rpc-listen-addr`, `agent-sock-port`, `id` and `pid-file`.)

What do you think about it?

PR #1606, opened 2023-10-06
feat: check health status of model service actively

kyujin-cho:
This should help user to deal with cases which server process starts fine at first but dies at sudden point.    
This PR also includes some minor improvements to the model service:
- Server process now starts with cwd set to `model_path` defined under `model-definition.yml`
- `run_command` pre-start action accepts optional argument `echo`, which carries out the stdout/stderr outputs to container log. This is set to false by default.

### DIFF
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -87,7 +87,6 @@ from ai.backend.common.events import (
     ExecutionTimeoutEvent,
     KernelCreatingEvent,
     KernelHealthCheckFailedEvent,
-    KernelHealthyEvent,
     KernelLifecycleEventReason,
     KernelPreparingEvent,
     KernelPullingEvent,
@@ -187,6 +186,7 @@ class AbstractKernelCreationContext(aobject, Generic[KernelObjectType]):
     kernel_id: KernelId
     session_id: SessionId
     agent_id: AgentId
+    event_producer: EventProducer
     kernel_config: KernelCreationConfig
     local_config: Mapping[str, Any]
     kernel_features: FrozenSet[str]
@@ -201,6 +201,7 @@ class AbstractKernelCreationContext(aobject, Generic[KernelObjectType]):
         kernel_id: KernelId,
         session_id: SessionId,
         agent_id: AgentId,
+        event_producer: EventProducer,
         kernel_config: KernelCreationConfig,
         local_config: Mapping[str, Any],
         computers: MutableMapping[DeviceName, ComputerContext],
@@ -212,6 +213,7 @@ class AbstractKernelCreationContext(aobject, Generic[KernelObjectType]):
         self.kernel_id = kernel_id
         self.session_id = session_id
         self.agent_id = agent_id
+        self.event_producer = event_producer
         self.kernel_config = kernel_config
         self.image_ref = ImageRef(
             kernel_config["image"]["canonical"],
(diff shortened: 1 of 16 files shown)


src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -1381,7 +1381,7 @@ def _pipeline(r: Redis) -> RedisPipeline:
             desired_session_count = endpoint.desired_session_count
             if (
                 endpoint.lifecycle_stage == EndpointLifecycle.DESTROYING
-                and len(endpoint.routings) == 0
+                and len(non_error_routings) == 0
```
achimnol: Minor fix. I'd recommend renaming this variable to "active" voice, like "active_routings" or "successfully_created_routings" (more concise ideas are welcome).

src/ai/backend/agent/kernel.py
```diff
@@ -926,49 +940,62 @@ async def read_output(self) -> None:
...
+                            response = json.loads(msg_data)
+                            event: AbstractEvent
+                            if response["is_healthy"]:
+                                event = KernelHealthyEvent(
+                                    self.kernel_id, self.session_id, reason=response["model_name"]
+                                )
+                            else:
+                                event = KernelHealthCheckFailedEvent(
+                                    self.kernel_id, self.session_id, reason=response["model_name"]
+                                )
```
achimnol: IMO, it would be better to unify these two events and distinguish the status by a field of the event to avoid bloating the kind of events.

achimnol (changes requested):
Minor fixes.

Later, we need to add a watchdog mechanism to enforce timeouts on asynchronously delivered health check results for when the container completely hangs. Let's do this in a follow-up PR.

src/ai/backend/agent/kernel.py
achimnol: Q. Do you have plans to expand the usage of "kernel health" events in other contexts?
If so, we should add a prefix (e.g., "model:") to the `reason` field for future disambiguation, or add another structured field to specify the type of kernel health event.

src/ai/backend/manager/scheduler/dispatcher.py
kyujin-cho: `active_routings ` looks good.

src/ai/backend/agent/kernel.py
kyujin-cho: I've updated the implementation to both unify the event type and create a new event argument solely for representing model service health status.

src/ai/backend/common/events.py
```diff
@@ -320,12 +320,33 @@ class KernelCancelledEvent(KernelCreationEventArgs, AbstractEvent):
...
+            kernel_id=KernelId(uuid.UUID(value[0])),
+            session_id=SessionId(uuid.UUID(value[1])),
+            model_name=value[2],
+            is_healthy=value[3],
+        )
 
 
-class KernelHealthCheckFailedEvent(KernelCreationEventArgs, AbstractEvent):
-    name = "kernel_health_check_failed"
+class ModelServiceHealthStatusUpdatedEvent(ModelServiceHealthStatusEventArgs, AbstractEvent):
```
achimnol: Verbosity is good but it seems to be too long.
How about `ModelServiceStatusEvent`?

src/ai/backend/agent/agent.py
```diff
@@ -2093,14 +2096,11 @@ async def start_and_monitor_model_service_initial_health(
...
-            )
-        else:
-            await self.event_producer.produce_event(
-                KernelHealthyEvent(
-                    kernel_obj.kernel_id, kernel_obj.session_id, reason=model["name"]
+                ModelServiceHealthStatusUpdatedEvent(
+                    kernel_obj.kernel_id,
+                    kernel_obj.session_id,
+                    model["name"],
+                    False,
```
achimnol: I think something is missing here to utilize `result["status"]`.

achimnol (changes requested):

achimnol (changes requested):
Let's change `is_healthy` from a boolean field to a string-enum field for readability and extensibility for the future!

PR #1622, opened 2023-10-16
fix: replace stream connection API key from kernel id to session id

fregataa:
follow-up #576 
Currently we are using `kernel_id` to save access info of session streaming connection.
Let's replace the key from `kernel_id` to `session_id`.

### DIFF
--- a/src/ai/backend/manager/api/stream.py
+++ b/src/ai/backend/manager/api/stream.py
@@ -450,7 +450,7 @@ async def stream_proxy(
     assert myself is not None
     try:
         async with root_ctx.db.begin_readonly_session() as db_sess:
-            session = await asyncio.shield(
+            session: SessionRow = await asyncio.shield(
                 database_ptask_group.create_task(
                     SessionRow.get_session(
                         db_sess,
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/api/stream.py
```diff
@@ -463,11 +464,12 @@ async def stream_proxy(
...
-    kernel_id = kernel.id
-    stream_key = kernel_id
+    kernel_id = KernelId(kernel.id)
+    session_id = SessionId(session.id)
+    stream_key = session_id
     stream_id = uuid.uuid4().hex
-    app_ctx.stream_proxy_handlers[stream_key].add(myself)
-    defer(lambda: app_ctx.stream_proxy_handlers[stream_key].discard(myself))
+    app_ctx.stream_proxy_handlers[kernel_id].add(myself)
+    defer(lambda: app_ctx.stream_proxy_handlers[kernel_id].discard(myself))
```
achimnol: `stream_proxy_handlers` in other places (e.g., `handle_kernel_terminating()`) still uses `kernel.id` as the stream key. We should make the naming of `stream_key` consistent in two different usage scenarios in this module.

achimnol (changes requested):
I've done a little refactoring. Added another review to let you look once more.

Issue #1626, opened 2023-10-18
Hide sensitive fields in the config APIs


achimnol:
Let's hide sensitive fields like passwords in the config APIs and GraphQL introspection queries.

Note that frontends should not treat the placeholder values sent by the server as a new user input.
```[tasklist]
### Tasks
- [ ] https://github.com/lablup/backend.ai/pull/1627
- [ ] https://github.com/lablup/backend.ai/pull/1632
- [ ] Filter sensitive fields in the `/config/get` manager API
- [ ] Deprecate `/config/get` and `/config/set` manager APIs completely
```

Issue #407, opened 2022-03-28
Upgrade to Graphene 3.0


achimnol:
https://github.com/graphql-python/graphene/releases/tag/v3.0.0

achimnol:
Resolved in #1632.

Issue #328, opened 2021-12-22
Upgrade to Graphene v3.0


achimnol:
https://github.com/graphql-python/graphene/releases/tag/v3.0.0
It is released after a years-long beta period.
Let's upgrade.

achimnol:
Resolved in #1632.

PR #1632, opened 2023-10-18
enhance: Upgrade Graphene to v3 with various GQL fixes

### LINKED ISSUE #1028: Update graphene to version 3
backend.ai manager uses Graphene v2 currently.
Graphene released v3 and it seems stable.
Let's update the version and also replace aiodataloader-ng to aiodataloader since it released 0.3.0, which is newer than aiodataloader-ng

note: It is a huge task.

supersedes #407 and #328.

achimnol:
Part of #1626 and resolves #1028, #328, #407 and supersedes #1027.

- setup: Update graphene and graphql-core to the latest v3 series
- fix: Make it easier to script with GQL schem dump
- fix: container registries are global suepradmin-only configuration
- feat: Show the detailed path of GQL query errors for easier debugging
- fix: Adapt to the graphene/graphql v3 APIs
- doc: Add news fragment
- feat: Add a GQL validation step to disable introspection queries
- feat: Add shared_config option to control schema introspection (`config/api/allow-graphql-schema-introspection`, with default `no`)

- [ ] Documentation


achimnol:
image
Now it blocks introspection queries. This will be the default for new setups.

src/ai/backend/manager/api/admin.py
```diff
@@ -170,5 +176,5 @@ def create_app(
     # (graphene transforms our object model into a textual representation)
     # This is useful for writing documentation!
     schema = graphene.Schema(query=Queries, mutation=Mutations, auto_camelcase=False)
```
kyujin-cho: Main characteristic of this CLI invocation (which dumps GQL schema as a plaintext file) duplicates with `backend.ai mgr gql show` command, so I think it is better to just drop this whole execution block and encourage users to use `gql show` command instead.

kyujin-cho (changes requested):

src/ai/backend/manager/api/admin.py
achimnol: Applied!

PR #1644, opened 2023-10-23
fix: provide deprecation msg to `max_vfolder_size` graphene field

fregataa:
follow up #1397 

- As we can [provide deprecation message to graphene fields](https://docs.graphene-python.org/en/latest/types/scalars/#scalars), let's set the argument to `max_vfolder_size` field.
- Update VFolder update API to use renewed column name `max_quota_scope_size`.


yomybaby:
Currently, `max_vfolder_count` and `max_quota_scope_size` in `[Create|Modify]ProjectResourcePolicyInput` and `[Create|Modify]UserResourcePolicyInput` are required fields. However, it would be better to change them to optional fields unless there is a specific reason for them to be required. @fregataa 
The required field in the input type can break backward compatibility.

fregataa:
> Currently, `max_vfolder_count` and `max_quota_scope_size` in `[Create|Modify]ProjectResourcePolicyInput` and `[Create|Modify]UserResourcePolicyInput` are required fields. However, it would be better to change them to optional fields unless there is a specific reason for them to be required. @fregataa The required field in the input type can break backward compatibility.

I didn't check those parts. I updated the code, thank you!

achimnol (changes requested):
As #1643 is merged, please update the merge conflict. (`max_vfolder_count`'s GQL type is changed from `BigInt` to `grpahene.Int`)

fregataa:
> As #1643 is merged, please update the merge conflict. (`max_vfolder_count`'s GQL type is changed from `BigInt` to `grpahene.Int`)

Thank you, I resolved the part and pushed new commits.

achimnol (changes requested):
Multi-line news fragment will break the formatting when auto-generating changelogs. Please put them in a single line, and contract as one sentence.

yomybaby (approved):
LGTM!

GraphQL schema has changed as expected.
```
type ProjectResourcePolicy {
  id: ID!
  name: String!
  created_at: DateTime!
  max_vfolder_count: BigInt
  max_vfolder_size: BigInt @deprecated(reason: "Deprecated since 23.09.1")
  max_quota_scope_size: BigInt
}
```

PR #1645, opened 2023-10-23
fix: Raise error when a created docker container exposes no port

fregataa:
Docker containers do not expose their ports if they use macvlan.
Let's handle it for better debug.


achimnol (approved):
👍🏼

PR #1649, opened 2023-10-24
feat: increase container creation failure log verbosity

fregataa:
Let's provide more detail logs when a docker container fails to initiate.


achimnol (changes requested):
"Initiate" (i.e., trigger) is different from "Initialize".
I think for this case "create" would be better, as conceptually initialization is done after creation but before starting.

achimnol (approved):

PR #1652, opened 2023-10-24
feat: Allow privileged access for other's VFolder to superadmin

rapsealk:
This PR is for the pipeline service.
I have notified that VFolder delete API fails to find a VFolder with given ID; it raises `No such vfolder.`.
I found that `_delete()` fails to access other's VFolder when attempting with `superadmin`'s authority, because `query_accessible_vfolders()` adds a condition for checking `uuid`.

This PR **always** allows privileged access to `superadmin` to delete other's VFolder.

https://github.com/lablup/backend.ai/blob/2d4ff4e3f53b92258c26053ad9c5a4683b6d6d4c/src/ai/backend/manager/api/vfolder.py#L2195-L2212

https://github.com/lablup/backend.ai/blob/2d4ff4e3f53b92258c26053ad9c5a4683b6d6d4c/src/ai/backend/manager/models/vfolder.py#L406-L416

- [ ] Documentation


kyujin-cho (approved):

kyujin-cho:
For future references: This commit is reverted and thus should not be backported.

PR #1663, opened 2023-10-27
feat: Allow project names to contain Unicode characters

### LINKED ISSUE #1658: Support "unicode" on project(group) name
### Main idea

For now, we only provide alphabet and special characters (e.g. `-`, `_`, `@`) in project name but doesn't allow any unicode, such as Korean (e.g. "프로젝트"). It seems to be ["regex" originated error](https://github.com/lablup/backend.ai/blob/60e4bde7765dce35697a6a9dd21accc11794abd9/src/ai/backend/manager/models/group.py#L481-L482), which could be handled in [here](https://github.com/lablup/backend.ai/blob/60e4bde7765dce35697a6a9dd21accc11794abd9/src/ai/backend/manager/models/group.py#L78).

### Alternative ideas

_No response_

### Anything else?

_No response_

lizable:
This PR resolves #1658.
Before this PR every client-side got "400 Bad Request" error, which is originated from regex that only allows alphanumeric characters,`-`(hyphen) and `_`(underscore). 
After applying this PR will allow unicode characters, such as Korean(e.g. "프로젝트"), But not allowing any whitespaces.

- [ ] Documentation

### Screenshot(s)
Before
Screenshot 2023-10-27 at 3 45 29 PM

After
Screenshot 2023-10-27 at 3 33 27 PM


achimnol (changes requested):
* We need to add a `TypeDecorator` subclass (in the `manager.models.base` module) for database columns to enforce slug formats, with a new option `allow_unicode` to `common.validators.Slug`.
* Remove `_rx_slug` in the `manager.models.group` module. It should have been refactored and removed.
* Let's add unit tests to demonstrate what examples are accepted or denied.
* The regular expression should use `\w` (e.g., character class as `[-\w]`). Refer the docs. Fortunately, the unicode mode is the default in Python 3. What we want is to allow "word" characters while still disallowing non-whitespace special characters other than the underscore and hyphen.
image

src/ai/backend/manager/models/domain.py
```diff
@@ -52,7 +51,7 @@
 )
 
 MAXIMUM_DOTFILE_SIZE = 64 * 1024  # 61 KiB
-_rx_slug = re.compile(r"^[a-zA-Z0-9]([a-zA-Z0-9._-]*[a-zA-Z0-9])?$")
+# _rx_slug = re.compile(r"^[a-zA-Z0-9]([a-zA-Z0-9._-]*[a-zA-Z0-9])?$")
```
achimnol: Please remove commented out codes.

src/ai/backend/manager/models/group.py
```diff
@@ -76,7 +75,7 @@
 )
 
 MAXIMUM_DOTFILE_SIZE = 64 * 1024  # 61 KiB
-_rx_slug = re.compile(r"^[a-zA-Z0-9]([a-zA-Z0-9._-]*[a-zA-Z0-9])?$")
+# _rx_slug = re.compile(r"^[a-zA-Z0-9]([a-zA-Z0-9._-]*[a-zA-Z0-9])?$")
```
achimnol: Same in this file.

src/ai/backend/manager/models/base.py
```diff
@@ -660,6 +661,37 @@ def from_row(
...
+    A type wrapper for allowing unicode
+    """
+
+    impl = sa.types.Unicode
+    cache_ok = True
+
+    def __init__(self, *, length: int = 64, allow_unicode: bool = False) -> None:
+        self._allow_unicode = allow_unicode
+        self._length = length
+        self._rx_slug = re.compile(r"^[a-zA-Z0-9]([a-zA-Z0-9._-]*[a-zA-Z0-9])?$")
```
achimnol: How about using the `\w` pattern in place of `[a-zA-Z0-9]` and setting `re.ASCII` flag as the inverse of `allow_unicode`?

achimnol (changes requested):

lizable (commented):
LGTM.

achimnol:
Rewrote the implementation and test cases, to handle variable-length negative lookbehind patterns and forbid consecutive occurrences in the middle as a serial invocation of regex pattern matching.

- Allow alpha-numeric chars (including unicode alphabets like Hangul).
- Allow hyphen, underscore, space, and dot in the middle, _only when_ they appear as a single char between alpha-numeric chars.
- Disallow hyphen, underscore, space, and dot to appear in the head/tail.

Now `manager.models.base.SlugType` and `common.validators.Slug` accepts three options to control the behavior: `allow_dot`, `allow_space`, `allow_unicode`.
Hyphen and underscore are always allowed regardless of the options.

**Difference to the prior implementation:** It now forbids the values beginning with a single dot. (e.g., `".abc"` was allowed but it's not rejected.)

achimnol (approved):

PR #1670, opened 2023-10-30
feat: Allow explicit `null` and empty string to ContainerRegistry mutations

rapsealk:
This PR is a follow-up of #1471 

This PR enhances GraphQL mutations for ContainerRegistry, by allowing explicit `null` and empty string.

Please run test codes:
```sh
pants test tests/manager/models/test_etcd.py
```

Refs
- #1674 

- [ ] Documentation


achimnol:
We should use the explicitly distinguished `undefined` and `null` introduced in Graphene 3.3 after merging #1632, by refactoring `set_if_set()` in all places.

yomybaby (approved):
LGTM

.github/workflows/default.yml
```diff
@@ -190,6 +190,10 @@ jobs:
       with:
         named-caches-hash: ${{ hashFiles('python*.lock', 'tools/*.lock') }}
         cache-lmdb-store: 'true'
+    - name: Run halfstack etcd container
+      run: |
+        sed -i'' "s/- \"8120:2379\"/- \"8121:2379\"/g" docker-compose.halfstack-main.yml
+        docker-compose -f docker-compose.halfstack-main.yml run -d --service-ports backendai-half-etcd
```
achimnol: This should be removed and replaced with the `ai.backend.testutils.bootstrap.etcd_container` fixture to keep it parallelization-safe when running `pants test`. Please refer other use cases.

achimnol (changes requested):

tests/manager/models/test_etcd.py
```diff
@@ -0,0 +1,233 @@
...
+
+    response = await client.execute_async(query, variables=variables, context_value=context)
+    container_registry = response["data"]["create_container_registry"]["container_registry"]
+    assert container_registry["hostname"] == "cr.example.com"
+    assert container_registry["config"] == {
+        "url": "http://cr.example.com",
+        "type": "dockerhub",
+        "project": ["default"],
+        "username": "username",
+        "password": "*****",
```
achimnol: "*****" is defined as a constant at `ai.backend.manager.defs.PASSWORD_PLACEHOLDER`.
Let's use it.

tests/manager/models/test_etcd.py
```diff
@@ -0,0 +1,233 @@
...
+
+    # Then password is set to empty string
+    response = await client.execute_async(query, variables=variables, context_value=context)
+    container_registry = response["data"]["modify_container_registry"]["container_registry"]
+    assert container_registry["hostname"] == "cr.example.com"
+    assert container_registry["config"]["url"] == "http://cr.example.com"
+    assert container_registry["config"]["type"] == "harbor2"
+    assert container_registry["config"]["project"] == ["default", "example"]
+    assert container_registry["config"]["username"] == "username2"
+    assert container_registry["config"]["password"] == "*****"
```
achimnol: Also does this test case reveal that the password is actually set as an empty string? I think we need to directly access the etcd to validate the fact.

achimnol (changes requested):

achimnol (approved):

PR #1475, opened 2023-08-13
ci: Adopt Ruff for faster linting and formatting

achimnol:
This PR replaces flake8 and isort with [Ruff](https://beta.ruff.rs/docs/), which runs 100 times faster.
Fortunately, [Pantsbuild already has an integration for Ruff!](https://blog.pantsbuild.org/linting-python-at-warp-speed/)
(NOTE: You need to run `pants fix` instead of `pants fmt` to run Ruff's fixers.)
Still, black and mypy are required as separate tools.

This won't reduce the CI/CD time much because the most overheads come from the initialization of Pants, but will greatly improve the daily code writing experience with IDE and editors.

```
> time LOCKSET=ruff/3.11.4 ./py -m ruff check src
LOCKSET=ruff/3.11.4 ./py -m ruff check src  0.06s user 0.05s system 108% cpu 0.101 total
> time LOCKSET=flake8/3.11.4 ./py -m flake8 src
LOCKSET=flake8/3.11.4 ./py -m flake8 src  9.15s user 0.41s system 308% cpu 3.099 total
```

You may need to update your IDE and editor configurations, referring the updated developer guide in this PR.
(Once merged, it will be available at: https://docs.backend.ai/en/latest/dev/daily-workflows.html#using-ides-and-editors)

## Migration of an existing development setup

Run:
```shell
rm -rf dist/export/python/virtualenvs/flake8
rm -rf dist/export/python/virtualenvs/isort
pants export --resolve=ruff
```

And then reconfigure your IDE or editor settings.

----
:books: Documentation preview :books:: https://sorna--1475.org.readthedocs.build/en/1475/

----
:books: Documentation preview :books:: https://sorna-ko--1475.org.readthedocs.build/ko/1475/

PR #1671, opened 2023-10-30
ci: When models change, dump gql schema

Yaminyam:
To easily identify when the graphql schema has changed, a dump is created when files in the models path are modified.


achimnol (changes requested):
Let's generalize the workflow filename and the top-level `name` key to `update-api-schema`.
Let's rename the job to `graphql`.

This will give a concise yet extensible check name "update-api-schema / graphql" where we could add "update-api-schema / rest" in the future.

achimnol:
As of #1686, we should use `mgr api dump-gql-schema` instead of `mgr gql show`.

.github/workflows/update-api-schema.yml
```diff
@@ -0,0 +1,64 @@
...
+      if: ${{ env.REMOTE_CACHE_BACKEND_ENDPOINT != '' }}
+    - name: Bootstrap Pants
+      uses: ./actions/init-pants
+      # See: github.com/pantsbuild/actions/tree/main/init-pants/
+      # ref) https://github.com/pantsbuild/example-python/blob/main/.github/workflows/pants.yaml#L30-L49
+      with:
+        named-caches-hash: ${{ hashFiles('python*.lock', 'tools/*.lock') }}
+        cache-lmdb-store: 'true'
+    - name: Pants export
+      run: pants export '::'
```
achimnol: We don't need to export all venvs to run the `mgr` commands.
Let's change it to `pants export --resolve=python-default`.

.github/workflows/update-api-schema.yml
```diff
@@ -0,0 +1,64 @@
...
+    - name: Create GraphQL schema dump
+      run: |
+        ./backend.ai mgr gql show >> schema.graphql
```
achimnol: Please set the correct destination path as suggested in the discussion thread.

achimnol (changes requested):

.github/workflows/update-api-schema.yml
Yaminyam: This content will be reflected by modifying it to use the --output option after number 1695 is merged.

Yaminyam:
> * [x]  Let's also migrate the OpenAPI XML generator (`./py -m ai.backend.manager.openapi`) to `mgr openapi show` command.
As with graphql, it was decided to replace this with mgr api dump-openapi.

achimnol:
Please resolve the merge conflict. I think the conflict is due to the formatting rule change in #1475, so you may `git checkout --ours`, check if there are any missing attributes added in the main branch, and reformat it again using `pants fmt`.

PR #1677, opened 2023-10-31
fix: restore removed graphene fields and set them deprecated

fregataa:
Deprecated graphql fields should not be removed.
The fields to be deleted should be marked with a deprecation message and deleted after a specified support period.

### DIFF
--- a/src/ai/backend/manager/models/utils.py
+++ b/src/ai/backend/manager/models/utils.py
@@ -364,12 +364,15 @@ def is_db_retry_error(e: Exception) -> bool:
     return isinstance(e, DBAPIError) and getattr(e.orig, "pgcode", None) == "40001"
 
 
-def description_msg(version: str, detail: str | None = None):
+def description_msg(version: str, detail: str | None = None) -> str:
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/models/utils.py
```diff
@@ -362,3 +362,14 @@ def agg_to_array(column: sa.Column) -> sa.sql.functions.Function:
...
+
+
+def description_msg(version: str, detail: str | None = None):
+    val = f"Added since {version}."
+    if detail:
+        val = f"{val} {detail}"
+    return val
+
+
+def deprecation_reason_msg(version: str):
```
yomybaby: Let's add the `detail: str`to here alos.

yomybaby (changes requested):

src/ai/backend/manager/models/utils.py
yomybaby: Except for this, everything looks good to me.

fregataa: Updated it, thank you!

PR #1674, opened 2023-10-31
refactor: Explicit Undefined value for GraphQL mutation inputs

achimnol:
resolves #1647.

- refactor: Use the explicit `Undefined` value for unspecified mutation input fields
- doc: Add news fragment

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version
- [x] Mention to the original issue
- [ ] Documentation
- [ ] API server-client counterparts (e.g., manager API -> client SDK)
- [ ] Test case(s) to demonstrate the difference of before/after

Issue #1647, opened 2023-10-23
Utilize UNDEFINED introduced in Graphene v3.3


achimnol:
Previously there was no way to distinguish the `null` value and unspecified optional field value in GraphQL mutation.
This has a portion of responsibility to the "infinity value" problem—depending the fields in various object types, there are a variety of the special values: 0, -1, None—because setting `None` to specifiy the infinity (unlimited) value and updating the object without changing the value was indistinguishable.

After #1632 is done, let adopt graphql-python/graphene#1506 introduced in Graphene v3.3.0 release.

* [x] Let `ai.backend.manager.api.admin` to call `graphene.types.inputobjecttype.set_input_object_type_default_value(graphql.Undefined)`
* [x] Update `ai.backend.manager.models.base.set_if_set()` method to check `graphql.Undefined` instead of `None`.
* [ ] Test and fix any unexpected behaviors with the frontend...

PR #1688, opened 2023-11-03
fix: Make Client SDK and Manager GraphQL API aware of undefined

achimnol:
It's a follow-up of #1674.

### Server-side
* *Create* mutations
  - Set the default values to all input fields whenever possible. (A)
  - Apply `set_if_set()` to only the fields with `required=False` AND *no default* value.
    - It is best to eliminate this case, though.
  - Rename `XXXInput` to `CreateXXXInput`.
* *Modify* mutations
  - All input fields (except those to identify the target object) become `required=False`. **Remove the default values**.
  - Apply `ai.backend.manager.models.utils.base.set_if_set()` to all fields. It is alreay updated to be aware of `graphql.Undefined` in #1647.
### Client-side
* Functional wrappers
  - Set the default values of all optional arguments to `client.types.undefined`. _Omit_ those fields when making GraphQL input data using `ai.backend.client.types.set_if_set()`.
    - For creation, the server will fill them with appropriate default values due to (A).
  - Except the fields to identify the target objects, enforce keyword arguments.
* CLI
  - Use `client.cli.params.OptionalType` to represent unspecified options in `update` commands.
  - Use `client.cli.params.BoolExprType` to represent tri-state (true/false/unchanged) of boolean flags in `update` commands.
  - Disambiguate by removing some flag names that can be interpreted multiple ways.
    - `-p` →  `--public` or `--private`??
      - For creation commands, allow `--private` only. (e.g., `--private` to set the flag true)
      - For modification commands, use `--private` as a tri-state flag (e.g., `--private=true`, `--private=false`, or unspecified)
  - Add mid-length option names
    - `--description`, `-d` → `--desc`

> **Note**
> "undefined" constants are only used within the server-side, or within the client-side. It NEVER crosses the boundaries of the system via APIs as serialized values. When crossing the boundaries, it should be translated from/to absence of specific fields.

### Other fixes

* Add missing `max_concurrent_sftp_sessions` in the keypair resource policy mutations
* Rename `--allowed-vfolder-hosts` to `--vfolder-host-perms` (while keepting the original option name for CLI backward compatibility) and update the CLI integration tests.

- [ ] Documentation


src/ai/backend/manager/models/user.py
```diff
@@ -518,7 +518,7 @@ class UserInput(graphene.InputObjectType):
     domain_name = graphene.String(required=True, default="default")
     role = graphene.String(required=False, default=UserRole.USER)
     group_ids = graphene.List(lambda: graphene.String, required=False)
-    allowed_client_ip = graphene.List(lambda: graphene.String, required=False)
+    allowed_client_ip = graphene.List(lambda: graphene.String, required=False, defualt=None)
```
jopemachine: I found a typo here `defualt` -> `default`

agatha197 (approved):
LGTM

achimnol:
Thanks for reviews!

src/ai/backend/manager/models/resource_policy.py
```diff
@@ -317,28 +317,48 @@ async def batch_load_by_ak(
...
+        deprecation_reason=deprecation_reason_msg("23.09.4"),
+    )
 
 
 class ModifyKeyPairResourcePolicyInput(graphene.InputObjectType):
     default_for_unspecified = graphene.String(required=False)
     total_resource_slots = graphene.JSONString(required=False)
     max_session_lifetime = graphene.Int(required=False)
     max_concurrent_sessions = graphene.Int(required=False)
+    max_concurrent_sftp_sessions = graphene.Int(required=False)
```
yomybaby, line 346: @achimnol When you add a field, please don't forget to include a description with the version information. For example,
```py
graphene.Int(reqruied=False, description='This is something, added 24.03.0')
```

PR #927, opened 2022-12-11
feat,docs: Auto-upgrading the halfstack

achimnol:

PR #1689, opened 2023-11-06
deps: Upgrade halfstack (postgresql, redis, etcd)

inureyes:
postgresql: 15.2  -> 16.0
 redis:      7.0.5 -> 7.2.3
 etcd:       3.5.6 -> 3.5.10

- [ ] Documentation

### DIFF
--- a/docker-compose.halfstack-2309.yml
+++ b/docker-compose.halfstack-2309.yml
@@ -1,0 +1,71 @@
+version: "2.4"
+
+services:
+
+  backendai-half-db:
+    image: postgres:16.0-alpine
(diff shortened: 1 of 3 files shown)


achimnol (changes requested):
redis and etcd are fine as they are minor upgrades.
But for postgres, this will **BREAK** existing dev setups!
For that part, #927 must be done first. Before it, we should only upgrade the minor version of postgres.

achimnol:
I'm closing this. Please revisit this with a proper upgrade path for the postgres, at least as an explicit documentation.

PR #1692, opened 2023-11-06
feat: Add `frontend-endpoint` and `hide-side-menu-button` to webserver's pipeline configs

rapsealk:
This PR is a follow-up of #1576 

This PR adds additional configs `frontend-endpoint` and `hide-button` to webserver's pipeline configs.

1. `pipeline.frontend-endpoint` will be used for "FastTrack" shortcut button on WebUI. Currently, the button is using `pipeline.endpoint`, which references the pipeline service's API server. The value has to be the pipeline service's frontend app as the button is intended to be a shortcut for the pipeline service.

2. `hide-button` will provide an option to hide the "FastTrack" button or not.

image

- [x] Documentation


agatha197 (changes requested):
In https://github.com/lablup/backend.ai-webui/pull/1929, the fasttrack's endpoint is saved named `endpoint` not `fasttrack-endpoint`.

configs/webserver/halfstack.conf
```diff
@@ -33,6 +33,8 @@ max_file_upload_size = 4294967296
 
 [pipeline]
 endpoint = "http://127.0.0.1:9500"
+frontend-endpoint = "http://127.0.0.1:3000"
```
agatha197, line 36:
```suggestion
endpoint = "http://127.0.0.1:3000"
```

rapsealk: Thanks for the feedback! This was intended - `endpoint` has to be preserved since is has been already used in many places. Instead, this PR is to **add** new configs than to replace one.

agatha197 (approved):
LGTM!

configs/webserver/halfstack.conf
```diff
@@ -33,6 +33,8 @@ max_file_upload_size = 4294967296
...
+hide-button = false
```
yomybaby: `hide-button` is a general name. Let's make it more specific. @rapsealk

rapsealk: @yomybaby How about `hide-shortcut-button`? It would be `pipeline.hide-shortcut-button` on server side.

yomybaby: `hide-side-menu-button`? ;)

yomybaby:
After this PR merged, let's make a WebUI PR to use these options.

kyujin-cho (approved):

PR #1695, opened 2023-11-07
ci: Excluding lock files from size calculations

Yaminyam:
Exclude lock files from calculating size labels.
This leads to confusion about the size of pr.


achimnol:
Does it also exclude lock files in the `tools` directory?
(i.e., shouldn't the pattern be `**/*.lock`?)

achimnol (approved):

PR #1700, opened 2023-11-07
feature: fetch SessionRow after `PRE_START_SESSION` hook

fregataa:
If any Manager's plugin update SessionRow(or sibling KernelRow) data in `PRE_START_SESSION` hook, we should fetch the updated data from DB.

This is related to edu launcher, which appends a bootstrap script in `PRE_START_SESSION` hook.


fregataa:
Close this PR since we can use `POST_ENQUEUE_SESSION` hook in such cases

PR #1717, opened 2023-11-14
ci: Add sbom report to release artifact

Yaminyam:
When releasing the sbom workflow that was previously executed manually, it is included and added to the artifact.

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -401,9 +401,11 @@ jobs:
         path: .pants.d/pants.log
       if: always()  # We want the log even on failures.
 
+  build-sbom:
+    uses: ./.github/workflows/sbom.yml
 
   make-final-release:
-    needs: [build-scies, build-wheels]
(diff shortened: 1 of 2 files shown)


.github/workflows/default.yml
```diff
@@ -445,6 +445,13 @@ jobs:
       with:
         name: local-proxy
         path: dist
+    - name: SBOM report generation
+      uses: ./.github/workflows/sbom.yml
```
achimnol: Could we parallelize this by separating it out as a separate job?

achimnol (changes requested):

.github/workflows/default.yml
Yaminyam: I think that parallelizing simple tasks will take more time for runner init, so I tend not to separate them for non-complex tasks.
In the case of reusable workflow, runner setup is needed anyway, so it would be better to separate it!

achimnol (approved):

PR #1718, opened 2023-11-15
fix: check both success and termination for the dependency predicate

rapsealk:
Several cases have been reported that sessions with dependency are not executed in sequence; a session runs even its precedent session has not been terminated yet.

This PR resolves this by checking if a dependent session has both succeeded and terminated.

- [ ] Documentation


fregataa (commented):
This is good because it reduces confusion between session dependencies for users, but I think there should be more discussion about dependency check in general before we merge this.
How do you think of this? @achimnol

src/ai/backend/manager/scheduler/predicates.py
```diff
@@ -128,7 +130,7 @@ async def check_dependencies(
     rows = result.fetchall()
     pending_dependencies = []
     for row in rows:
-        if row.result != SessionResult.SUCCESS:
+        if row.result != SessionResult.SUCCESS or row.status != SessionStatus.TERMINATED:
```
kyujin-cho, line 133: Is it okay not to cover `CANCELLED` session status here?

kyujin-cho (changes requested):

src/ai/backend/manager/scheduler/predicates.py
rapsealk: I think those which depend on `CANCELLED` sessions are not supposed to be started.

rapsealk: They rather be notified and `CANCELLED` either?

kyujin-cho: Let's keep this as is for now and create a patch if same issue occurs for `CANCELLED` sessions.

kyujin-cho (approved):

PR #1721, opened 2023-11-16
fix: Restrict destroy of terminated sessions

fregataa:
destroy() API determines determines what to do based on the status of the session. It should not do anything if the session is already terminated.

### DIFF
--- a/src/ai/backend/manager/registry.py
+++ b/src/ai/backend/manager/registry.py
@@ -2144,6 +2144,10 @@ class AgentRegistry:
                     await self.event_producer.produce_event(
                         SessionTerminatingEvent(session_id, reason),
                     )
+                case SessionStatus.TERMINATED:
+                    raise GenericForbidden(
+                        "Cannot destroy sessions that has already been already terminated"
+                    )
                 case _:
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/registry.py
```diff
@@ -2144,6 +2144,10 @@ async def destroy_session(
                     await self.event_producer.produce_event(
                         SessionTerminatingEvent(session_id, reason),
                     )
+                case SessionStatus.TERMINATED:
+                    raise GenericForbidden(
+                        "Cannot destroy sessions that has already been already terminated"
+                    )
                 case _:
                     await SessionRow.set_session_status(
```
kyujin-cho: Just a question: is it okay to leave `CANCELLED` as is?

kyujin-cho (changes requested):

src/ai/backend/manager/registry.py
fregataa: good point! I updated this part to restrict destroy cancelled sessions also

kyujin-cho (approved):

PR #1727, opened 2023-11-21
feat: Use `ui.menu_blocklist` to hide pipeline menu button

rapsealk:
This PR is a counterpart against lablup/backend.ai-webui#2036 and partially reverts #1692 

This PR removes recently added `pipeline.hide-side-menu-button` and uses `ui.menu_blocklist` which already has been used to manage various menu buttons, to enhance maintenance.

- [ ] Documentation

### DIFF
--- a/changes/.feature.md
+++ b/changes/.feature.md
@@ -1,0 +1,1 @@
+Use `ui.menu_blocklist` to hide pipeline menu button and delete `pipeline.hide-side-menu-button`.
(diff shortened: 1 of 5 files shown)


achimnol:
Please update the installer code to set the added option.

src/ai/backend/install/context.py
```diff
@@ -509,6 +509,7 @@ async def configure_webserver(self) -> None:
                 if halfstack.redis_password:
                     redis_table["password"] = halfstack.redis_password
             data["session"]["redis"] = redis_table  # type: ignore
+            data["ui"]["menu_blocklist"] = "statistics,pipeline"  # type: ignore
```
rapsealk: This line will add `[ui].menu_blocklist` field with default value. Afterward, users can change the value from `webserver.conf`.

PR #1739, opened 2023-11-24
fix: do null check on `rate_limit`

fregataa:
`rate_limit` column of keypairs table is nullable.
we should do null check.


agatha197 (approved):
LGTM

achimnol (approved):

PR #1741, opened 2023-11-24
fix: convert redis config value

fregataa:
follow-up: #1585

`trafaret.Float()` does not convert string value into float. It just check whether it can be float or not.
We should rather use `trafaret.ToFloat()`.


jopemachine:
Thanks for letting me know! :)

PR #1745, opened 2023-11-28
fix: Correct misspelling of resource

Yaminyam:
I think it's a spelling error that could also affect the logic.


inureyes (approved):
LGTM

Issue #1729, opened 2023-11-22
Revamp End-to-End testing


kyujin-cho:
### Main idea

Since Backend.AI project is getting complicated even more these days, it is almost impossible for us to test core functionalities only with our hands. Therefore I think this is the right time to revamp our old fashioned integration test package and introduce a daily E2E testing CI pipeline.

- [x] Update (or rewrite) `ai.backend.test.cli_integration` package so that it can be ran and report errors
- [ ] Add new test cases for other uncovered CLI features
- [ ] Create a Github action pipeline to execute the integration on a daily basis
  ### Alternative ideas

_No response_

JIRA Issue: BA-175

rapsealk:
@mirageoasis You may start a new PR from #1659

PR #1747, opened 2023-11-28
fix : fixing e2e cli test

mirageoasis:
currently working on #1729


mirageoasis:
@achimnol finished with fix.md but cla seems not working how can I solve this? when I click I agree it does nothing

PR #1753, opened 2023-12-02
chore: enable raise-without-from-inside-except ruff rule

kyujin-cho:
This will help investigating root cause of the error more easier when inspecting the error log.

### DIFF
--- a/src/ai/backend/agent/docker/metadata/server.py
+++ b/src/ai/backend/agent/docker/metadata/server.py
@@ -143,7 +143,7 @@ class MetadataServer(aobject):
             structure_pointer = self.route_structure
             for component in components:
                 if structure_pointer.get(component) is None:
-                    raise web.HTTPNotFound
+                    raise web.HTTPNotFound from None
                 structure_pointer = structure_pointer[component]
             resources = []
             for k, v in structure_pointer.items():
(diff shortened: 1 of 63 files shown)


src/ai/backend/agent/docker/metadata/server.py
```diff
@@ -143,7 +143,7 @@ async def route_structure_fallback_middleware(
             structure_pointer = self.route_structure
             for component in components:
                 if structure_pointer.get(component) is None:
-                    raise web.HTTPNotFound
+                    raise web.HTTPNotFound from None
```
achimnol, line 146: In this case, I think we should simply reraise the original exception.
```suggestion
                    raise
```

src/ai/backend/client/func/vfolder.py
```diff
@@ -350,9 +350,9 @@ async def _download_file(
                                                                 chunk_size
                                                             )
                                                         except asyncio.TimeoutError:
-                                                            raise TryAgain
```
achimnol, line 409: Couldn't we just simply raise a "root" exception instead of specifying `from None` always for brevity?

achimnol (changes requested):
For the auto-generated `as ex` expressions, I think it would be better to unify the style: `as e`.

According to [the officla doc](https://docs.python.org/3/tutorial/errors.html#exception-chaining), Python automatically attach the chaining information when the `raise` statement is inside an `except` block.

We should force this rule **only when** the `raise` statement is not inside the `except` block.

HyeockJinKim:
It has now been applied to main. I'll close this PR.

PR #1756, opened 2023-12-04
feat: Allows field_ref to be used as a key value in fieldset

Yaminyam:
When using the --format option of session list, previously only the name specified as alt_name could be used.
However, it should be able to work not only with alt_name but also with the original name, field_ref.

### DIFF
--- a/src/ai/backend/client/output/types.py
+++ b/src/ai/backend/client/output/types.py
@@ -104,7 +104,9 @@ class FieldSpec:
 
 class FieldSet(UserDict, Mapping[str, FieldSpec]):
     def __init__(self, fields: Sequence[FieldSpec]) -> None:
-        super().__init__({f.alt_name: f for f in fields})
+        fields_set = {f.alt_name: f for f in fields}
+        fields_set.update({f.field_ref: fields_set[f.alt_name] for f in fields})
+        super().__init__(fields_set)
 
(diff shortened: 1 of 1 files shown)


tests/client/output/test_types.py
```diff
@@ -44,7 +44,7 @@ def test_fieldspec_init():
             ]
         ),
     )
-    assert f.field_ref == "key_foo { bar baz }"
+    assert f.field_ref == "key_foo { bar baz baz }"
```
achimnol, line 47: Shouldn't the addition be `bbb`?

Yaminyam: That line is a test to check the field_ref, so 
{baz: fieldspec{field_ref:baz,alt_name:bbb}, bbb: fieldspec{field_ref:baz,alt_name:bbb}}
It looks awkward because it is in the form of , but it is correct that field_ref has two baz.

Yaminyam: Since field_ref, not the key value, must be output, it is correct to have two baz.
`{baz: fieldspec{field_ref:baz, alt_name:bbb}, bbb: fieldspec{field_ref:baz, alt_name:bbb}}`

achimnol: Got it!

PR #1770, opened 2023-12-08
fix(BA-176): Compare image memory limit to sum of shmem and main mem

### LINKED ISSUE #1726: Make the shared memory setting for new sessions more intuitive
This is a late follow-up to lablup/backend.ai-webui#314.

Currently, the image label `ai.backend.resource.min.mem` is interpreted as the main memory size, _excluding_ the shared memory size.However, the web UI's resource configuration automatically sets the shared memory size ($S$) to 64 MiB for the main memory sizes ($M$) less than 4 GiB, while our scheduler allocates the sum of the main memory size and the shared memory size ($M+S$).

This makes a confusion when allocating the least amount of memory $M+S = 256\ \mathrm{MiB}$, for instance, because it _epic-fails_ as the web UI sends $M = 192\ \mathrm{MiB}$ and $S = 64\ \mathrm{MiB}$, while the manager's enqueue-session API handler compares the image label `ai.backend.resource.min.mem` with $M$ only and requires $M \ge 256\ \mathrm{MiB}$.

We are going to update the web UI to hide the detailed shared memory configuration for most use cases, and the memory resource slider will expose $M+S$ with auto-configured $S$ depending on the value of $M + S$.

To better support the above web UI update, let's change the enqueue-session API handler to:

- [x] Compare `ai.backend.resource.min.mem` with $M+S$ instead of $M$.
- [x] Raise an explicit input validation error if $S > M$, to prevent situations like $M = 0$ wihle $S = 1\ \mathrm{GiB}$.
  The Client SDK and CLI should still expose the raw configurations as the options. So, let's:

- [x] Update the CLI's help texts to be explicitly consistent with the above changes.

JIRA Issue: BA-176

fregataa:
resolves #1726 (BA-176)

This makes the behavior of the backend's interpretation of "mem" ($M$) and "shmem" ($S$) values consistent with the Web UI, which passes them as $(M-S, S)$ instead of $(M, S)$.

- Compare `ai.backend.resource.min.mem` with the sum of main memory size and shared memory size instead of main memory size.
- Configurable ratio between shared memory and memory.


achimnol:
I'm going to hold this PR because the calculation method change will break existing API users (now we have serious ones) and complicates which values to store in the scheduler.

src/ai/backend/manager/registry.py
```diff
@@ -1061,16 +1060,13 @@ async def enqueue_session(
...
-            # and cgroup's memory limit does not apply.
-            shmem = resource_opts.get("shmem", None)
-            if shmem is None:
-                shmem = labels.get("ai.backend.resource.preferred.shmem", "64m")
-            shmem = BinarySize.from_str(shmem)
+            # Do not including the shared memory when comparing the memory slot with image requiring resource slot.
+            shmem = BinarySize.from_str(
+                resource_opts.get("shmem")
+                or labels.get("ai.backend.resource.preferred.shmem")
+                or DEFAULT_SHARED_MEMORY_SIZE
```
achimnol: I think injection of the default value should be taken care of by the client side, and there should be a way to set `shmem` to zero intentionally when required. If we could distinguish "undefined" and "zero" here, setting the default value for "undefined" case would be fine.

src/ai/backend/manager/registry.py
```diff
@@ -1123,6 +1119,14 @@ async def enqueue_session(
                 if tpu is not None:
                     raise InvalidAPIParameters("Client upgrade required to use TPUs (v19.03+).")
 
+            # Check if the user has allocated an "imbalanced" shared memory amount.
+            if shmem >= requested_slots["mem"]:
```
achimnol: Let's make the ratio configurable, via etcd's scoped configurations (e.g., possible to override by resource groups).

achimnol (changes requested):

src/ai/backend/manager/defs.py
```diff
@@ -80,3 +81,7 @@ class LockID(enum.IntEnum):
 
 DEFAULT_KEYPAIR_RESOURCE_POLICY_NAME: Final = "default"
 DEFAULT_KEYPAIR_RATE_LIMIT: Final = 10000
+
+DEFAULT_SHARED_MEMORY_SIZE: Final = "64m"
+DEFAULT_MIN_MEM_SHARED_MEM_RATIO: Final = Decimal(1.0)
+MIN_MEM_SHARED_MEM_RATIO_KEY: Final = "manager/mem-shmem-ratio"
```
achimnol: Please update the module-level comment to describe the schema of this new etcd key in `src/ai/backend/manager/config.py`.

Also, the key and the variable should be renmaed to align with `resources/max-shm-ratio`. The ratio has the max limit, while allowing zero allocation of shm if intended (so no "min" limit). So the naming `MIN_...` may look confusing to the users/admins. I think you were confused to reverse the min/max concept here -- we are setting the relative size of shmem _out of_ the memory size. Conceptually, this is not a manager-component option but a resource-control option.

achimnol (changes requested):

src/ai/backend/manager/defs.py
fregataa: - I renamed the constants to `MAX_...` and related variables were also renamed. 
- Changed the comparative formula that checks too-large-shmem 
```python
# Before
if mem / shmem <= MIN_RATIO: raise

# After
if shmem >= mem * MAX_RATIO: raise
```
- The constants are relocated to `src/ai/backend/common.defs.py`.

Is the location of `MAX_..` constants proper? I am not sure which module is a good place

fregataa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/1770?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1770** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/1770?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2972** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2972?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

HyeockJinKim (approved):
LGTM

PR #1771, opened 2023-12-08
chore: Update Python development tools

achimnol:
* setup: Update python tool dependencies
  - black: 23.9.1 -> 23.11.0
  - mypy: 1.6.1 -> 1.7.1
  - ruff: 0.1.1 -> 0.1.7
  - pytest: 7.4.2 -> 7.4.3
* style: Apply the updated black and ruff rules
  - Now we can use more concise nested parenthesis/braces/brackets inside another one
    when there are no sibling items.
  - Actually this was my preferred style before adopting Black!
* fix: newly found typecheck errors
  - mypy found some potential bugs with its improved type checking!


achimnol:
NOTE: `require:db-migration` is a false-positive as black reformats some alembic migration scripts.

achimnol:
Currently I'd like to hold merging of this PR due to:
- https://github.com/astral-sh/ruff/issues/8678

achimnol:
Ah, we can use https://docs.astral.sh/ruff/settings/#format-preview !
Ruff now provides an intrinsic formatter that works almost in the same way that Black does, even including some subset of preview styles.

achimnol:
For now, let's stay on the 2023 stable style whereas both ruff and black almost identically works.

PR #1761, opened 2023-12-05
feature: Materialization of the "main" keypair concept

fregataa:
- Implement `main_keypair` which is unique for every user. the column name is `users.main_access_key`
- Superadmins may change a user's main keypair. But a keypair cannot be deleted while it is a main keypair.
- Update related the CLI and client SDK functions.

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version
- [ ] API server-client counterparts (e.g., manager API -> client SDK)

PR #1775, opened 2023-12-11
fix: Regression of purging user due to fkey constraint of `main_access_key`

fregataa:
follow-up #1761 

The `main_access_key` foreign key has `ondelete=RESTRICT` constraint, which disallow to delete any user record.
Replace the `ondelete` constraint on the `main_access_key` foreign key with `SET NULL` because the existing implementation deletes `keypair` record before deleting `user`.

- [x] Test case(s) to:


src/ai/backend/manager/models/keypair.py
```diff
@@ -624,7 +624,18 @@ async def mutate(
...
         ctx: GraphQueryContext = info.context
+        async with ctx.db.begin_readonly_session() as db_session:
+            user_query = sa.select(UserRow.uuid).where(UserRow.main_access_key == access_key)
+            uuid_list: list[str] = [
+                str(raw_uuid) for raw_uuid in (await db_session.scalars(user_query)).all()
+            ]
+            if uuid_list:
+                return DeleteKeyPair(
+                    False, f"the keypair is used as main access key by user(uuid: {uuid_list})"
+                )
```
kyujin-cho: Seems like these lines can be simplified by just evaluating result of `COUNT(*)` query, as the purpose of this block is to check if the user is utilizing this keypair as its main keypair.

kyujin-cho (approved):

PR #1778, opened 2023-12-12
test: Revamp the E2E CLI-based integration test suite

### LINKED ISSUE #1729: Revamp End-to-End testing

mirageoasis:
revied pr #1747  due to conflict issue
which will resolve #1729 

improved following

1. option.click type=bool -> type=OptionalType(BoolExprType)
2. moved params.py from client/cli to cli since it is used by admin/cli
3. moved Sentinel, Undefined class from client/cli/types.py to cli/types due to dependency from OptionalType in params.py
4. refactored test fixture

### DIFF
--- a/changes/1778.fix.md
+++ b/changes/1778.fix.md
@@ -1,4 +1,1 @@
-* option.click type=bool -> type=OptionalType(BoolExprType)
-* moved params.py from client/cli to cli since it is used by admin/cli
-* moved Undefined class from client/cli/types.py to cli/types due to dependency from OptionalType in params.py
-refactored test fixture
+option.click type fix, moved params.py and Undefined class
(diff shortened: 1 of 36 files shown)


changes/1778.fix.md
achimnol, line 1: News fragment should be a single sentence in one line, as it is inserted into [the changelog](https://github.com/lablup/backend.ai/blob/main/CHANGELOG.md). You may use the Markdown syntax here.

src/ai/backend/cli/params.py
```diff
@@ -6,7 +6,7 @@
 import click
 import trafaret
 
-from ..types import Undefined, undefined
+from ai.backend.cli.types import Undefined, undefined
```
achimnol: Inside the same package namespace, use the relative imports when possible.

achimnol (changes requested):
Overall, LGTM. Please revise the news fragment.

achimnol (approved):

PR #1779, opened 2023-12-12
fix: Deprecates querying projects from KeyPair query

fregataa:
follow-up #1022 

After #1022 , it has been possible to query `projects` from `KeyPair` GraphQL query, which shows what projects the owner of the user of the keypair are associated with.
BUT according to the reasons below, we decided to deprecate query `projects` in `KeyPair`.
- we decided `KeyPair` should be used only as authentication of users and not used as any other purpose.
- It causes too many left outer joins.


agatha197 (changes requested):
Sorry for checking late. Please resolve the conflicts.

PR #1780, opened 2023-12-12
docs: bundle external OpenAPI spec visualizer

kyujin-cho:
This PR adds a new OpenAPI spec renderer in place of sphinx-based generator.

- [x] Documentation


achimnol (approved):
There are a few issue to follow:
* The REST API reference won't be visible in PDF builds.
* We need to use a separate local web server (as added in docs/README) to have a working OpenAPI viewer.
  - The redoc (spec-viewer) seems to recognize `file://` URI scheme as the NodeJS environment while it's not.

PR #1785, opened 2023-12-13
fix: Update fixtures to include setting main keypairs

mirageoasis:
This PR introduces a new feature to `manager.base.populate_fixtures()` so that it could run as the "update" mode and splits out the fixture to set the main keypairs of the users using this update mode.

It also improves error messages when there are mistakes in the fixture files.


tests/manager/fixtures/example-resource-presets.json
```diff
@@ -1,29 +1,52 @@
...
+    {
+      "name": "cpu02-medium",
+      "resource_slots": {
+        "cpu": "4",
+        "mem": "8589934592"
+      },
+      "shared_memory": "1073741824"
+    },
+    {
+      "name": "cpu03-large",
```
fregataa: Why did resource-preset fixture change in this PR?

fregataa (commented):
I checked the fixtures work well.

tests/manager/fixtures/example-resource-presets.json
mirageoasis: think when this pr was first made I had old preset but when I merged main branch in this pr `example-resource-presents.json` has changed.
this change 
my pr 
old ---- merged   -----> new
main                  |
old  --- new

https://github.com/lablup/backend.ai/commit/8aaf27850ae87cec7572d95a0fe470f4b308f52b
between time when I worked on this branch and now there is pr from Joongi Kim which changed this fixture

scripts/install-dev.sh
```diff
@@ -955,7 +955,9 @@ configure_backendai() {
   # initialize the DB schema
   show_info "Setting up databases..."
   ./backend.ai mgr schema oneshot
+  ./backend.ai mgr fixture populate fixtures/manager/example-users.json
   ./backend.ai mgr fixture populate fixtures/manager/example-keypairs.json
+  ./backend.ai mgr fixture populate fixtures/manager/example-set-user-main-access-keys.json --override
```
achimnol: Even this works, it is better to keep the options in front of the arguments.
```suggestion
  ./backend.ai mgr fixture populate --override fixtures/manager/example-set-user-main-access-keys.json
```

You may experience very strange and hard-to-debug errors if you mix up options and arguments in gcc and linkers, for example. Always make your habits to conform with the standard convention.

tests/manager/fixtures/example-set-user-main-access-keys.json
achimnol, line 1: Here, you may leave the `uuid` (primary key) and `main_access_key` fields only, as you've implemented the in-place update. Let's remove duplication.

achimnol (changes requested):

achimnol (approved):

PR #1793, opened 2023-12-18
fix: `admin mgr agent info` is returning an incorrect value

Yaminyam:
Because results contain key values, delete keys that are not needed.
Modify mgr agent ping, which uses get_node_hwinfo(), which has not been implemented yet, to use ping().
It will be replaced when get_node_hwinfo() is implemented in the future.

- [ ] Test case(s) to:
- [ ] Documentation


achimnol (approved):
Rewrote several parts to match on par with the latest codebase.

PR #1796, opened 2023-12-19
doc: Update Backend.AI production installation guide doc

hoyajigi:
resolves https://github.com/lablup/giftbox/issues/583

Update Backend.AI production installation guide doc

- [x] Documentation

----
📚 Documentation preview 📚: https://sorna--1796.org.readthedocs.build/en/1796/

----
📚 Documentation preview 📚: https://sorna-ko--1796.org.readthedocs.build/ko/1796/

### DIFF
--- a/docs/install/install-from-package/install-agent.rst
+++ b/docs/install/install-from-package/install-agent.rst
@@ -87,7 +87,7 @@ would be:
    scratch-size = "1G"
 
    [watcher]
-   service-addr = { host = "bai-m1", port = 6009 }
+   service-addr = { host = "127.0.0.1"", port = 6009 }
    ssl-enabled = false
    target-service = "backendai-agent.service"
    soft-reset-available = false
--- a/docs/install/install-from-package/install-agent.rst
+++ b/docs/install/install-from-package/install-agent.rst
@@ -87,7 +87,7 @@ would be:
    scratch-size = "1G"
 
    [watcher]
-   service-addr = { host = "127.0.0.1"", port = 6009 }
+   service-addr = { host = "bai-a01"", port = 6009 }
    ssl-enabled = false
    target-service = "backendai-agent.service"
    soft-reset-available = false
--- a/docs/install/install-from-package/install-agent.rst
+++ b/docs/install/install-from-package/install-agent.rst
@@ -189,6 +189,9 @@ First, create a runner script at ``${HOME}/bin/run-agent.sh``:
       export HOME="/home/bai"
    fi
 
+   # -- If you have installed using static python --
+   source .venv/bin/activate
+
    # -- If you have installed using pyenv --
    if [ -z "$PYENV_ROOT" ]; then
       export PYENV_ROOT="$HOME/.pyenv"
--- a/docs/install/install-from-package/os-preparation.rst
+++ b/docs/install/install-from-package/os-preparation.rst
@@ -124,7 +124,7 @@ to a directory of your choice.
 
 .. code-block:: console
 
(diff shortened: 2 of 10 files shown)


docs/install/install-from-package/install-agent.rst
```diff
@@ -87,7 +87,7 @@ would be:
    scratch-size = "1G"
 
    [watcher]
-   service-addr = { host = "bai-m1", port = 6009 }
+   service-addr = { host = "127.0.0.1"", port = 6009 }
```
adrysn: Agent Watcher exposes API endpoints to Manager, therefore binding to 127.0.0.1 is not advisable since Manager and Agent can be installed in different hosts. Instead, bind to a host interface that the Manager can access, or revert to using the host alias.

docs/install/install-from-package/os-preparation.rst
```diff
@@ -113,11 +113,42 @@ information on the Python version compatibility can be found at
...
+   $ tar -xf "cpython-3.11.6+20231002-x86_64-unknown-linux-gnu-install_only.tar.gz"
+   $ mkdir -p "/home/bai/.static-python/versions"
+   $ mv python "/home/bai/.static-python/versions/3.11.6"
+
+Update your PATH depending on your shell, restart your shell for the PATH changes to take effect.
 
-Use pyenv to manually build and select a specific Python version
+.. code-block:: console
+
+   $ echo 'export PATH=".static-python/versions/${PYTHON_VERSION}/bin:$PATH"' >> ~/.bashrc
```
adrysn: I'm not sure about setting the statically built Python as the default for the user's shell. Instead, I suggest utilizing a script, such as the one found at https://github.com/lablup/backend.ai-ansible/blob/main/remotes/home/scripts/autoenv.sh, to activate the per-directory Python environment.

docs/install/install-from-package/prepare-cache-service.rst
```diff
@@ -50,9 +50,9 @@ Execute the following command to start the service container. The project
 
 .. code-block:: console
 
-   $ cd ${HOME}/halfstack
-   $ docker compose -f docker-compose.hs.redis.yaml -p ${USER} up -d
+   $ cd ${HOME}/halfstack/redis-cluster-default
+   $ docker compose -p ${USER} up -d
```
adrysn: The `-p ${USER}` part is no longer necessary. As the halfstack services are now separated by directory, there's no need to use a project-specific approach. Just like the following will be enough for every service:
```shell
cd <specific-service-directory>
docker compose down
docker compose up -d
docker compose logs -f
docker compose ps
...
```

adrysn (changes requested):
Good work. I have some comments.

docs/install/install-from-package/install-agent.rst
hoyajigi: With the all respect, I don't understand at all. This setting is used when launching aiohttp's TCPSite, and if you enter the Manager's hostname in the Agent machine, the server will not run.
Also, install from source document and sample tomls say to set to AGENT IP or 127.0.0.1.

docs/install/install-from-package/os-preparation.rst
hoyajigi: I will look more how to set Static Python and write more documentation setting env using venv.

hoyajigi: modifying systemd runner script using venv

docs/install/install-from-package/install-agent.rst
adrysn: I did not say to set this as the "Manager's hostname." Instead, I mentioned setting it to the (Agent's) host interface that the Manager can access or connect to. As I said, if the Manager and Agent are installed on separate servers, the Manager cannot send requests to Agent Wather which binds to 127.0.0.1.

adrysn (changes requested):

docs/install/install-from-package/install-agent.rst
hoyajigi: @adrysn reverted to use the host alias.

adrysn (approved):
LGTM.

Issue #409, opened 2022-04-01
Batching support in vfolder mkdir API


achimnol:
Let's make the vfolder mkdir API to accept multiple directories to create at once.

PR #1803, opened 2023-12-21
feat: Accept multiple paths in the vfolder mkdir API

mirageoasis:
This PR adds an extra feature to specify multiple directories in the vfolder mkdir API to create them via a single API call.

- Allow passing multiple relative paths of directores to create via `nargs` in `click.option`
- Add '-p' option to auto-create parent directories recursively

The feature itself is a relatively small addition, but **this PR demonstrates how we are going to design and implement multi-result (or batch) APIs.**

* Introduce `common.types.{ItemResult,ResultSet}` to distinguish success/failed results, with optional error messages attached to each item.
  - Now [the client SDK's output framework](https://github.com/lablup/backend.ai/issues/1925) also has `print_result_set()` API.
* The HTTP response codes will be segregated as follows:
  - 200: All success
  - 207: Partial success/failure (from WebDAV)
  - 422: All failure (semantic failure, not server error) (from WebDAV)

### Examples

Partial success (console output):
image

Partial success (JSON output):
image

All failed (JSON output):
image


src/ai/backend/client/cli/vfolder.py
```diff
@@ -445,17 +445,19 @@ def cp(filenames):
...
-def mkdir(name, path, parents, exist_ok):
+def mkdir(name, paths, parents, exist_ok):
     """Create an empty directory in the virtual folder.
 
     \b
     NAME: Name of a virtual folder.
-    PATH: The name or path of directory. Parent directories are created automatically
-          if they do not exist.
+    PATH: Mutliple arguments of name or path of directory seperated by space. Parent directories are created automatically
+          if they do not exist. (e.g. "dir1" "dir2" "dir3")
```
fregataa: Is this example placed here by mistake?
it is hard to understand what this example means

src/ai/backend/client/func/vfolder.py
```diff
@@ -522,11 +522,11 @@ async def _mkdir(
     @api_function
     async def mkdir(
         self,
-        path: Union[str, Path],
+        paths: List[str | Path],
```
fregataa:
```suggestion
        paths: Union[str, Path, List[Union[str, Path]]],
        # we can use `TypeAlias` if the typing looks too messy
```
I suggest to remain the name of the parameter as `path` and let users pass both `list` type and `str, Path` type for compatibility.

We should be very careful when we update APIs because there are many cases of using API other than Backendai CLI.

src/ai/backend/manager/api/vfolder.py
```diff
@@ -1228,7 +1228,7 @@ async def update_vfolder_options(
 @vfolder_permission_required(VFolderPermission.READ_WRITE)
 @check_api_params(
     t.Dict({
-        t.Key("path"): t.String,
+        t.Key("paths"): t.List(t.String),
```
```suggestion
        t.Key("path"): t.String | t.List(t.String),
```
Same here. Let's just add one more available type.

src/ai/backend/manager/api/vfolder.py
```diff
@@ -1239,11 +1239,11 @@ async def mkdir(request: web.Request, params: Any, row: VFolderRow) -> web.Respo
     folder_name = request.match_info["name"]
     access_key = request["keypair"]["access_key"]
     log.info(
-        "VFOLDER.MKDIR (email:{}, ak:{}, vf:{}, path:{})",
+        "VFOLDER.MKDIR (email:{}, ak:{}, vf:{}, paths:{})",
         request["user"]["email"],
         access_key,
         folder_name,
-        params["path"],
+        params["paths"],
```
```suggestion
        params["path"],
```
If we leave the name of the parameter, then this also should be the same

src/ai/backend/manager/api/vfolder.py
```diff
@@ -1253,7 +1253,7 @@ async def mkdir(request: web.Request, params: Any, row: VFolderRow) -> web.Respo
         json={
             "volume": volume_name,
             "vfid": str(VFolderID(row["quota_scope_id"], row["id"])),
-            "relpath": params["path"],
+            "relpaths": params["paths"],
```
```suggestion
            "relpaths": params["path"],
```
Same here. it is fine to change the name of storage-proxy API parameters because it is used only by BackendAI Managers

src/ai/backend/storage/api/manager.py
```diff
@@ -776,14 +777,18 @@ class Params(TypedDict):
...
-                    params["relpath"],
-                    parents=params["parents"],
-                    exist_ok=params["exist_ok"],
-                )
+            with handle_fs_errors(volume, vfid):
+                mkdir_tasks = [
+                    volume.mkdir(vfid, relpath, parents=parents, exist_ok=exist_ok)
+                    for relpath in relpaths
+                ]
+                await asyncio.gather(*mkdir_tasks)
```
fregataa: It is a good idea to use `asyncio.gather()` here.
How about passing `True` to [`return_exceptions`](https://docs.python.org/3/library/asyncio-task.html#asyncio.gather) parameter and handle errors here?

fregataa (commented):
Good!
I left some comments

src/ai/backend/client/cli/vfolder.py
mirageoasis: I agree it makes usage more confusing

src/ai/backend/client/func/vfolder.py
```diff
@@ -506,7 +506,7 @@ async def upload(
 
     async def _mkdir(
         self,
-        path: Union[str, Path],
+        path: str | Path | List[str | Path],
```
fregataa: Let's use `Union` instead of bitwise or operation because of compatibility. I didn't consider compatibility of this change when I said it is okay to use `|`(bitwise or) operator. Sorry for my early confusing reply!

src/ai/backend/client/func/vfolder.py
```diff
@@ -522,7 +522,7 @@ async def _mkdir(
     @api_function
     async def mkdir(
         self,
-        path: Union[str, Path],
+        path: str | Path | List[str | Path],
```
fregataa: same here

changes/1803.feature.md
```diff
@@ -0,0 +1 @@
+Add support for multi directory mkdir and fix type annotation accordingly
```
fregataa: This fragment will be used as a change log of release notes.
"What" we do here is kind of improving the feature of a CLI command, and that is what readers of release notes should know from this.
So, let's just inform readers the improvement and skip the "fixing type annotation" part which is "How". "How" should be in a description of PR.

src/ai/backend/storage/api/manager.py
fregataa: The reason why I suggest to use `return_exception` is that the current implement returns only one failure when multiple `mkdir` fails.

fregataa (commented):
some more comments!

src/ai/backend/storage/api/manager.py
```diff
@@ -776,14 +778,32 @@ class Params(TypedDict):
...
+
         async with ctx.get_volume(params["volume"]) as volume:
-            with handle_fs_errors(volume, params["vfid"]):
-                await volume.mkdir(
-                    params["vfid"],
-                    params["relpath"],
-                    parents=params["parents"],
-                    exist_ok=params["exist_ok"],
-                )
+            with handle_fs_errors(volume, vfid):
```
fregataa: we do not need `handle_fs_errors()` here because it cannot catches any errors since we handle error as a result of tasks of `gather()`

src/ai/backend/storage/api/manager.py
```diff
@@ -776,14 +778,32 @@ class Params(TypedDict):
...
+                    raise FileExistsError("None of directories created because they already exist")
+
+                for result_or_exception in result_group:
+                    if isinstance(result_or_exception, BaseException):
+                        log.error(
+                            "%s",
+                            repr(result_or_exception),
+                            exc_info=result_or_exception,
+                        )
+
```
fregataa: Let's refer the `handle_fs_errors()` function. It catches all ValueError and OSError and return the error message in response. How about return an appended error message in http response?

src/ai/backend/client/cli/vfolder.py
```diff
@@ -430,7 +430,7 @@ def cp(filenames):
 
 @vfolder.command()
 @click.argument("name", type=str)
-@click.argument("path", type=str)
+@click.argument("paths", type=str, nargs=-1)
```
mirageoasis, line 452: maybe should I change paths to path?

achimnol: Since this is an argument name, not the option name, I think it is safe to change the argument name without breaking existing client CLI users.

src/ai/backend/storage/api/manager.py
```diff
@@ -776,15 +779,53 @@ class Params(TypedDict):
     ) as params:
         await log_manager_api_entry(log, "mkdir", params)
         ctx: RootContext = request.app["ctx"]
+        vfid = params["vfid"]
+        parents = params["parents"]
+        exist_ok = params["exist_ok"]
+        relpath = params["relpath"]
+        relpaths = relpath if isinstance(relpath, list) else [relpath]
+        failure_tasks: List[str] = []
+        success_tasks: List[str] = []
```
fregataa:
```suggestion
        failure_tasks: list[str] = []
        success_tasks: list[str] = []
```

https://docs.python.org/3/library/typing.html#typing.List
`List` is deprecated since version 3.9

src/ai/backend/storage/api/manager.py
```diff
@@ -755,7 +757,7 @@ async def mkdir(request: web.Request) -> web.Response:
     class Params(TypedDict):
         volume: str
         vfid: VFolderID
-        relpath: PurePosixPath
+        relpath: PurePosixPath | List[PurePosixPath]
```
fregataa:
```suggestion
        relpath: PurePosixPath | list[PurePosixPath]
```
same here

fregataa (commented):
good!
I checked the functionality and it works well.
I left a few comment about typing

fregataa (approved):
LGTM 👍

mirageoasis:
@achimnol 
finished with issue #409 !

src/ai/backend/storage/api/manager.py
```diff
@@ -776,15 +779,53 @@ class Params(TypedDict):
...
+            ]
+            result_group = await asyncio.gather(*mkdir_tasks, return_exceptions=True)
+            exception_flags = [isinstance(res, BaseException) for res in result_group]
+
+            # if result group is all false throw error.
+            if all(exception_flags):
+                raise web.HTTPInternalServerError(
+                    body=json.dumps(
+                        {
+                            "msg": "FileExistsError None of directories created!",
```
achimnol: How are you sure that all exceptions are `FileExistsError`?
We should not "assume" any exception but always explicitly report the original information whenever possible.

In this case, I think we could sample the first exception instance, so that the user may retry the operation with revised inputs, though he/she may experience other remaining errors.

src/ai/backend/client/cli/vfolder.py
```diff
@@ -445,18 +445,23 @@ def cp(filenames):
...
     help="Skip an error caused by file not found",
 )
-def mkdir(name, path, parents, exist_ok):
+def mkdir(name, paths, parents, exist_ok):
     """Create an empty directory in the virtual folder.
 
     \b
     NAME: Name of a virtual folder.
-    PATH: The name or path of directory. Parent directories are created automatically
+    PATH: Multiple arguments of name or path of directory seperated by space. Parent directories are created automatically
```
achimnol: This is a multi-value arguments, not argument-separated-by-space. They are different one: `"value1 value2 value3"` vs. `value1 value2 value3`.
Let's simply reveal that "you may specify multiple paths to create at once."

achimnol (changes requested):

src/ai/backend/client/func/vfolder.py
achimnol: IMHO, this seems fine.
We are using Python 3.11 now and `|` was fully introduced in Python 3.10.
(ref: [PEP 604](https://peps.python.org/pep-0604/))

src/ai/backend/manager/api/vfolder.py
```diff
@@ -1257,9 +1257,18 @@ async def mkdir(request: web.Request, params: Any, row: VFolderRow) -> web.Respo
...
-    ):
-        pass
-    return web.Response(status=201)
+    ) as (_, storage_resp):
+        storage_reply = await storage_resp.json()
+        resp_code = storage_resp.status
+        resp = {
+            "failure_tasks": storage_reply["failure_tasks"],
+            "success_tasks": storage_reply["success_tasks"],
+        }
```
achimnol: The storage proxy implementation in this PR does NOT ALWAYS return these fields in the reply.
For example, it may return "msg" and "errno" when all given paths failed to be created.
We should define the consistent form of the API reply.

achimnol (changes requested):

src/ai/backend/manager/api/vfolder.py
achimnol: I revised and refactored this part by introducing reusable `common.types.{ItemResult,ResultSet}` definitions.

mirageoasis:
Totally different api from what I first wrote! amazing!
More fined grained reports!

PR #1804, opened 2023-12-21
fix: Embed ResponseModel converter middleware in service API

fregataa:
follow-up #1752 

Currently we have `check_api_params_v2` decorator function which check request's parameters and parse response before return it. But there are some API functions that isn't decorated by it and just returns `ResponseModel`, which occurs error. So,
- Embed a webapp middleware to parse the response of the functions.

Plus, fix some pydantic APIs usage when parse ResponseModel to web response.
- Change `TypeAdapter.dump_json()` to `TypeAdapter.dump_python()` because `dump_json()` returns serialized bytes.
- Change `BaseModel.model_dump_json()` to `BaseModel.model_dump()` because `model_dump_json()` returns JSON string, not `dict`.

- [ ] Test case(s) to:
- [ ] Documentation


lizable (changes requested):
I think returning pydantic instance for only those functions with `check_api_params_v2` decorators would be enough. I'm not sure we really need middleware for this issue. How about just reverting response type and value of delete function instead of adding middleware for converting responses?

I reverted and tried deleting service model and It worked fine.
https://github.com/lablup/backend.ai/pull/1752/files#diff-d21fbe306a9b579a2491f344fc4de5eff8407147aac0c18568f4ca95880b27c2L409-R577

fregataa:
> I think returning pydantic instance for only those functions with `check_api_params_v2` decorators would be enough. I'm not sure we really need middleware for this issue. How about just reverting response type and value of delete function instead of adding middleware for converting responses?
> 
> I reverted and tried deleting service model and It worked fine. https://github.com/lablup/backend.ai/pull/1752/files#diff-d21fbe306a9b579a2491f344fc4de5eff8407147aac0c18568f4ca95880b27c2L409-R577

Thank you for reviewing. But this is a good way to use middleware because all service APIs should be filtered by a certain function and this is not a global middleware.

kyujin-cho (approved):

PR #1810, opened 2023-12-26
fix: Key mismatch for registry URL in `ModifyContainerRegistry` mutation

### LINKED ISSUE #1808: Registry modification doesn't work properly

Linux (x86-64)

### Backend.AI version

24.03.0a1

### Describe the bug

Seems like changing the address of registry (etcd values) is not working. According to query requested from WebUI, value has been sent correctly, but I think there's something missing during processing the updated values.

### To Reproduce

You may try in backend.ai webserver process

1. Login with superadmin/admin account
2. Go to Environment page
3. Click 'gear' icon in Controls column
4. Change anything on input field. (including registry URL)
5. Click save button
6. See the error.

### Expected Behavior

Updating values including registry URL need to be applied.

### Anything else?

_No response_

rapsealk:
This PR is a follow-up of #1471 and resolves #1808.

An issue has been identified in the `ModifyContainerRegistry` mutation, where the `url` field was not updating as intended. The problem stemmed from a key mismatch: while the GraphQL input field uses `url`, the etcd uses an empty string ("") for the registry URL. This update addresses the issue by modifying the key from url to an empty string ("") if it exists.

- [x] Test case(s) to:
- [ ] Documentation


lizable (approved):
LGTM.

PR #1817, opened 2024-01-03
feature: improve Postgres connection mgmt

fregataa:
## Connection and transaction management
SQLAlchemy Engine (sqlalchemy.engine.Engine) can creates `Connection` and `Transaction`. 
When Connection creates Transaction by calling its `begin()`(connection.begin()), it **explicitly** initiate the Transaction under the connection.
When Engine creates Transaction by calling its `begin()` (engine.begin()), it **implicitly** creates a new Connection and initiate Transaction with the Connection.
Currently Backend.AI does not explicitly manage the lifecycle of Connections and it can cause connection burden especially when it runs `models.utils.execute_with_retry()` which retry entire DB connection.

Let's decrease the connection burden by managing a lifecycle of `Connection`s and retry DB transaction within a connection.

- https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Connection.begin
- https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Engine.connect

### Notice
Some of the code that retries DB calls through the `execute_with_retry()` API has been modified to prevent further retries because the DB calls will not be retried as intended.
- `get_*_occupancy()` APIs do not retry DB calls anymore because `execute_with_retry()` retries only when `SerializationError` occurs, which requires retry the entire transaction, not rolling back to `CHECKPOINT`.

"connection is closed" `DBAPIError` can occur if we set timeout for connections. This is because the pools of SQLAlchemy is not fully synced with real connections.
Let's set `pool-recycle` config to replace timeout connections
- https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.create_engine.params.pool_recycle
- https://docs.sqlalchemy.org/en/14/core/pooling.html#setting-pool-recycle
- https://community.fly.io/t/postgresql-connection-is-closed-error-after-a-few-minutes-of-activity/4768
- https://github.com/sqlalchemy/sqlalchemy/discussions/7657

- [x] Test case(s) to:

----
📚 Documentation preview 📚: https://sorna--1817.org.readthedocs.build/en/1817/

----
📚 Documentation preview 📚: https://sorna-ko--1817.org.readthedocs.build/ko/1817/


changes/1817.feature.md
```diff
@@ -0,0 +1 @@
+Improve Postgres connection management by retrying transactions instead of connections and add `pool-recycle` config to replace timeout connections.
```
achimnol:
```suggestion
Improve Postgres connection management by retrying transactions inside connections instead of retrying entire connections, and add `pool-recycle` config to drop and replace timed-out connections.
```

achimnol (changes requested):
It seems better to use more commonly used abbreviations:
`execute_with_trx_retry` → `execute_with_txn_retry`
`execute_with_cnx_retry` → `execute_with_conn_retry`

Also, if we want to replace `execute_with_retry()` in the entire codebase, please do so.

achimnol (changes requested):
Also, there may be unexpected side-effects or conflicts with nested `tenacity.AsyncRetrying` scopes as `execute_with_cnx_retry()` calls `execute_with_trx_retry()`.
Let's make each function to independently run the retry loop in the connection level and the transaction level inside a connection, respectively.

fregataa:
Turn this PR into a draft since #2088 and its related stack PRs will resolve the same issue.

fregataa:
close this PR since #2088 will cover this issue

PR #1822, opened 2024-01-05
feat: auditlogs 2 in vfolder

mirageoasis:
revived
#698 

----
📚 Documentation preview 📚: https://sorna--1822.org.readthedocs.build/en/1822/

----
📚 Documentation preview 📚: https://sorna-ko--1822.org.readthedocs.build/ko/1822/


src/ai/backend/manager/models/audit_logs.py
```diff
@@ -0,0 +1,266 @@
...
+    CHANGE = "CHANGE"
+    DELETE = "DELETE"
+
+
+class AuditLogTargetType(str, enum.Enum):
+    """ """
+
+    USER = "user"
+    KEYS = "keypairs"
+    VFOLDER = "vfolder"
```
fregataa: It is quite annoying to add enum value to existing Enum type in DB through alembic.
I recommend you to add all available target types here and update alembic migration file to reduce future annoying work!
Plus, please unify the name in singular or plural.

```suggestion
class AuditLogTargetType(str, enum.Enum):
    """ """

    USER = "user"
    GROUP = "group"
    KEYPAIRS = "keypair"
    VFOLDER = "vfolder"
    COMPUTE_SESSION = "compute_session"
```

src/ai/backend/manager/api/vfolder.py
```diff
@@ -316,6 +323,49 @@ async def _wrapped(request: web.Request, *args: P.args, **kwargs: P.kwargs) -> w
...
+        "created_at": datetime.utcnow(),
+        "success": False,
+        "rest_api_path": request.path,
+    })
+
+    try:
+        return await handler(request)
+    except Exception:
+        raise
+    finally:
```
```suggestion
    try:
        return await handler(request)
    finally:
```

you do not need to catch exception here unless you handle it

src/ai/backend/manager/api/vfolder.py
```diff
@@ -316,6 +323,49 @@ async def _wrapped(request: web.Request, *args: P.args, **kwargs: P.kwargs) -> w
...
+            log.info("AUDIT_LOG: {}", audit_log_data.get())
+            async with root_ctx.db.begin_session() as sess:
+                await sess.execute(sa.insert(audit_logs).values(audit_log_data.get()))
+        except Exception as e:
+            log.error("Failed to write audit log {}", e, exc_info=True)
+
+
+def updated_data(new_data: dict[str, Any]) -> dict[str, Any]:
+    current_audit_log_data = audit_log_data.get().copy()
+    return deep_update(current_audit_log_data, new_data)
```
```suggestion
def update_audit_log_data(new_data: dict[str, Any]) -> None:
    audit_log_data.set(
      {**audit_log_data.get(), **new_data}
    )
```

This will work.
And let's not use pydantic v1 API

fregataa (changes requested):
I will review in detail after we determine to write the audit logs to DB or file system (I personally prefer to write them on DB because it will be hard to track all audit logs under HA setup)

If we decide to use DB for audit logs, then there should be 2 major works.
1. update alembic migration files
2. use SQLAlchemy ORM rather than a raw table
3. change `AuditLog` graphene object to graphene relay object. I can do this work if you don't mind!

src/ai/backend/manager/api/vfolder.py
mirageoasis: maybe I think using this function is okay
CleanShot 2024-01-10 at 14 40 32@2x

https://docs.pydantic.dev/latest/migration/

this documents says it  `pydantic.v1.utils import deep_update` does not mean it has been deprecated!

src/ai/backend/manager/models/alembic/versions/41f6bbb4a04a_add_audit_logs_table.py
```diff
@@ -0,0 +1,65 @@
...
+        ),
+        sa.Column("target", sa.String(length=64), nullable=True),
+        sa.Column("error", sa.String(length=128), nullable=True),
+        sa.Column(
+            "created_at", sa.DateTime(timezone=True), server_default=sa.text("now()"), nullable=True
+        ),
+        sa.Column("success", sa.Boolean(), server_default=sa.text("true"), nullable=False),
+        sa.Column("rest_resource", sa.String(length=256), nullable=True),
+        sa.Column("gql_query", sa.String(length=1024), nullable=True),
+    )
```
fregataa: `id` column is missing here

src/ai/backend/manager/models/alembic/versions/41f6bbb4a04a_add_audit_logs_table.py
```diff
@@ -0,0 +1,65 @@
...
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_table("audit_logs")
+    auditlogaction.drop(op.get_bind())
```
fregataa: Need to drop unique key constraint if we have `id` column in the `audit_logs` table

src/ai/backend/manager/models/audit_logs.py
```diff
@@ -0,0 +1,234 @@
...
+        self.access_key = access_key
+        self.email = email
+        self.action = action
+        self.data = data
+        self.target_type = target_type.value
+        self.success = success
+        self.target = target
+        self.rest_resource = rest_resource
+        self.gql_query = gql_query
+        self.created_at = datetime.datetime.now()
```
fregataa: Is `__init__()` defined for type checking?
As [SQLAlchemy 1.4 doc](https://docs.sqlalchemy.org/en/14/orm/quickstart.html#create-objects-and-persist), this ORM class already has `__init__()` method.
Plus, it would be great if `id` and `created_at` are created by DB rather than by Python built-in modules.

fregataa (changes requested):
Looks good.
Can I update some codes that use SQLAlchemy APIs? We are planning to migrate to SQLAlchemy 2.0 and we should update some of the codes that are not compatible to 2.0  version

src/ai/backend/manager/models/alembic/versions/41f6bbb4a04a_add_audit_logs_table.py
fregataa: Nevermind. primary key columns have a `primary key` constraint and we don't need to add or delete the constraint.

achimnol (changes requested):
Remove `src/ai/__init__.py` and `src/ai/backend/__init__.py` as these are _namespace_ packages.

HyeockJinKim:
We're closing this PR because we've added audit log collection at the service layer.

PR #1823, opened 2024-01-07
fix: Ignore handling of 'transwindow' value in Dropbear

### LINKED ISSUE #1819: Fix or monkeypatch dropbear to send the data beyond 500MB for a single SFTP channel
After investigating a customer report, we found that a very specific combination of Dropbear as the SFTP server host and PuTTY-based SFTP clients (including FileZilla's `fzsftp.exe` and WinSCP) on Windows has a problem to cut the download in the middle when the streamed data size exceeds 500 MB. Note that there was also another report saying that OpenSSH's own SFTP client works well.

Backend.AI Agent auto-mounts the prebuilt executables of `dropbear` and `sftp-server` (from the openssh-portable project) to provide the SSH and SFTP capability to arbitrary user containers regardless of their base images.

In the past, we had some performance issues and PyCharm compatibility issues regarding the **_upload_** buffer management (lablup/backend.ai-agent#266). But this case is a _different_ one: the issue affects the SFTP **_download_** traffic only.

## Investigation

* Both FileZilla and WinSCP uses a vendored version of PuTTY's SFTP handler: `psftp.exe`. FileZilla applies additional patches to improve the performance and ships the modified version as `fzsftp.exe`. (ref: https://serverfault.com/questions/700780/why-is-filezilla-so-much-faster-than-psftp)
* Dropbear spawns the container's `sftp-server` binary (usually provided as a part of `openssh-server` package in Linux distros, but in our case we mount our own prebuilt version), if installed, to handle SFTP commands in an SSH session.
  - When sending, Dropbear uses [the hard-coded window size](https://github.com/mkj/dropbear/blob/41a6abca32dd3da273152c07dbded41fb6d70726/src/sysoptions.h#L217-L219), `500000000` (500 MB), and [decrement it by subtracting the chunk length transmitted out](https://github.com/mkj/dropbear/blob/41a6abca32dd3da273152c07dbded41fb6d70726/src/common-channel.c#L768). [It stops transmission when the window size reaches zero.](https://github.com/mkj/dropbear/blob/41a6abca32dd3da273152c07dbded41fb6d70726/src/common-channel.c#L715-L724)
* psftp's download loop implementation [simply stops](https://git.tartarus.org/?p=simon/putty.git;a=blob;f=psftp.c;h=a33bb35ab9bc014be9f9ed948f0f4aec77655371;hb=b846178443cf1a5dc7c5ea2079fd34fd465af497#l470) if the server stops transmission, while OpenSSH's sftp client implementation [requests the transmission again](https://github.com/openssh/openssh-portable/blob/1036d77b34a5fa15e56f516b81b9928006848cbd/sftp-client.c#L1781) if there is an insufficient number of bytes in the received buffer.

The combination of above implementation details resulted a very specific corner case that our customer has reported.

This wouldn't happen if we don't use Dropbear as the SSH provider of user containers, or if users don't use SFTP clients derived from PuTTY's implementation.

## Mitigation

We could simply increase the Dropbear's hard-coded window limit to a larger number as [a part of our custom patches](https://github.com/lablup/backend.ai/blob/4a19001f9d1ae12be7244e14b304d783da9ea4f9/scripts/agent/build-dropbear.sh#L51-L54). But this would not be an ultimate solution because our customers often wants to transfer many gigabytes or even terabytes of files.

Instead, we should patch the dropbear's channel implementation to be able to transmit the data without limits. Let's find out how we can minimally modify the source code of dropbear.

jung-min-seo:
resolves #1819


achimnol:
Recreating this PR as I cannot push LFS files of the new dropbear builds to a forked repository.

PR #1829, opened 2024-01-09
feature: Pass Root context of manager to plugins

fregataa:
follow-up #1699 

let's pass a root context to plugins instead of a plugin context for easy accessibility to db

### DIFF
--- a/changes/1829.feature.md
+++ b/changes/1829.feature.md
@@ -1,0 +1,1 @@
+Pass Root context of manager to plugins.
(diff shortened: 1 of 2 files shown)


achimnol (approved):

achimnol:
I think we need a revision PR to:
- [ ] Ensure backward-compatibility with existing hook plugins which use the `hook_plugin_ctx` argument
- [ ] Make the initialization step of both webapp/hook plugin contexts consistent
  - Currently `await ctx.init(...)` and `root_ctx.{webapp,hook}_plugin_context = ctx` are reversed to each other.

PR #1832, opened 2024-01-10
fix: update session's occupying slots when kernel starts

fregataa:
resolves https://github.com/lablup/giftbox/issues/602

image

- Update `sessions` table's `occupying_slots` field with kernel's actual resource allocation
- `convert_resource_spec_to_resource_slot()` method changed because it converts all values into `str`, which is unable to add to other resource slots. This WILL NOT change the value saved in DB because `ResourceSlotColumn` converts numeric values into string when ResourceSlot data is saved.

- [x] Test case(s) to:


src/ai/backend/manager/models/session.py
```diff
@@ -1082,6 +1082,68 @@ async def get_sgroup_managed_sessions(
         return result.scalars().all()
 
 
+async def finalize_running(
```
kyujin-cho: Most of the characteristics of this function are pretty simillar with those on `SessionRow.transit_session_status()`. I suppose we can just update implementations of the latter one instead of maintaining two duplicate functions.

kyujin-cho: Unlike `SessionRow.transit_session_status()` this hook does not have a safety gear to prevent session status from being updated to an illegal state. https://github.com/lablup/backend.ai/blob/bde8d203dd00223d962a3e11772d9a2319b770a9/src/ai/backend/manager/models/session.py#L762-L764 Is it okay to do so?

kyujin-cho (changes requested):

src/ai/backend/manager/registry.py
```diff
@@ -1581,13 +1582,8 @@ async def finalize_running(
                 ),
             }
             self._kernel_actual_allocated_resources[kernel_id] = actual_allocs
-            kernel_did_update = await KernelRow.update_kernel(
-                self.db, kernel_id, new_status, update_data=update_data
-            )
-            if not kernel_did_update:
```
kyujin-cho, line 1587: Is there a specific reason lifting up this condition?

kyujin-cho (changes requested):

src/ai/backend/manager/registry.py
fregataa: hm, I don't remember exactly. I reverted this part. thanks!

Issue #1323, opened 2023-06-08
Invalid vfolder mount target path


fregataa:
Currently, it is possible to assign mount target path that includes a comma such as `C:/hello`.
but such mount path occurs error because we use `:` as a delimeter of vfolder's source path, target path and permission like this `/src/hello:/trg/hello:rw`.
Let's block such target path or please suggest any idea.

achimnol:
Maybe we could force conversion to `pathlib.PosixPath` before interpolating the vfolder mount expression.

achimnol:
To distinguish if the given string is a Windows path or not, we could use the following sample: https://chat.openai.com/share/75f57f32-c9fe-4faa-b6af-ffb450fabed4

But, in this case, we need to generally *escape* the colons in the path components as colons are invalid in Windows filenames but valid in Linux filenames.

PR #1838, opened 2024-01-14
feat: Enhance VFolder mount with additional explicit and verbose options

### LINKED ISSUE #1836: Enable explicit permission override for vfolder mounts
### Main idea

Based on the existing implementation, the permission granted to a mounted virtual folder is derived from the user's permissions. For instance, if User-A possesses `Read-Write` permissions for a given virtual folder, such as `vf-0`, the mounted session for User-A will consistently reflect `Read-Write` permissions. It is important to note that explicit permission overrides are not supported by the current implementation.

https://github.com/lablup/backend.ai/blob/fbea3bee7c0be34ca7767e9a944b9451357596fd/src/ai/backend/manager/models/vfolder.py#L783-L791

References
- [Volumes | Docker Docs - Use a read-only volume](https://docs.docker.com/storage/volumes/#use-a-read-only-volume)
```sh
# --mount
$ docker run -d \
  --name=nginxtest \
  --mount source=nginx-vol,destination=/usr/share/nginx/html,readonly \
  nginx:latest

# --volume
$ docker run -d \
  --name=nginxtest \
  -v nginx-vol:/usr/share/nginx/html:ro \
  nginx:latest
```

### Alternative ideas

_No response_

### Anything else?

_No response_

rapsealk:
This PR resolves #1836 by introducing a new handler, `prepare_mount_arg_v2()`.

For instance, users can now use the following formats with two VFolders: `test-folder-808` with `READ_WRITE` permission and `test-folder-809` with `READ_ONLY` permission.

1. Users can use the same format as `NAME[=PATH]` or `NAME[:PATH]`, as shown below:
```sh
$ ./backend.ai run --rm --mount test-folder-808:/home/work/abc ...
```

2. Additionally, the format below is supported:
```sh
$ ./backend.ai run --rm --mount type=bind,source=test-folder-808,target=/home/work/abc ...

# with explicit READ_ONLY permission
$ ./backend.ai run --rm --mount type=bind,source=test-folder-809,target=/home/work/abc,permission=ro ...
$ ./backend.ai run --rm --mount type=bind,source=test-folder-808,target=/home/work/abc,perm=ro ...

# with explicit READ_WRITE permission
$ ./backend.ai run --rm --mount type=bind,source=test-folder-808,target=/home/work/abc,perm=rw ...
```

Each field separated by commas is optional, except for the `source` field, which is mandatory. If `permission` is ungiven, it will follow the permitted `VFolderPermission`.

For client, a new parameter `mount_options` has been added:
```python
mount_options: Mapping[str, Mapping[str, Any]] = {
  "vf-001": {
    "type": "bind",
    "permission": "ro",
  },
}
```

- [x] Test case(s) to:
- [x] Documentation


rapsealk:
Would it be better to relocate this to the `common` package for the purpose of reusability?

https://github.com/lablup/backend.ai/blob/44612158c9a4e23a0d215a5a859cf52c7a43cfc5/src/ai/backend/agent/resources.py#L450-L478

rapsealk:
Related to #1323

src/ai/backend/common/types.py
```diff
@@ -372,6 +375,44 @@ class MountTypes(str, enum.Enum):
...
+
+
+class MountExpression:
+    def __init__(self, expression: str, *, escape_map: Optional[Mapping[str, str]] = None) -> None:
+        self.expression = expression
+        self.escape_map = {
+            "\\,": ",",
+            "\\:": ":",
+            "\\=": "=",
+        }
```
achimnol, line 404: :+1:

tests/client/cli/test_mount.py
achimnol, line 1: Thanks for adding the test cases!

src/ai/backend/manager/api/session.py
```diff
@@ -186,6 +194,14 @@ def check_and_return(self, value: Any) -> object:
 creation_config_v5 = t.Dict({
     t.Key("mounts", default=None): t.Null | t.List(t.String),
     tx.AliasedKey(["mount_map", "mountMap"], default=None): t.Null | t.Mapping(t.String, t.String),
+    tx.AliasedKey(["mount_options", "mountOptions"], default=None): t.Null
+    | t.Mapping(
+        t.String,
+        t.Dict({
+            t.Key("type", default=MountTypes.BIND): MountTypes,
+            tx.AliasedKey(["permission", "perm"], default=None): t.Null | MountPermission,
```
achimnol: Though this issue is from a missing refactoring, `MountPermission` and `MountTypes` are not a trafaret. I think we need to make trafaret wrappers of these enums.  We could update the base type from `(str, enum.Enum)` to `enum.StrEnum` (introduced in 3.11) and wrap these with `t.Enum`.

> [!NOTE]
> A trafaret should call `self._failure()` to generate a properly structured `trafaret.DataError` to be reported as an input valiadtion error. See custom trafaret impl. examples in the `ai.backend.common.validators` module.

achimnol (changes requested):
Generally LGTM!
Just let's trafaret-fy `MountTypes` and `MountPermission` in the API IV checks.

rapsealk:
```sh
rror(\'{0: DataError(\\\'value should be None\\\'), 1: DataError(\\\'{\\\\\\\'pipeline-0-YynSBt\\\\\\\': DataError(\\\\\\\'{\\\\\\\\\\\\\\\'value\\\\\\\\\\\\\\\': DataError(\\\\\\\\\\\\\\\'{\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\'permission\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\': DataError("{0: DataError(\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\'value should be None\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\'), 1: DataError(\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\'value is not a valid member of MountPermission\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\')}")}\\\\\\\\\\\\\\\')}\\\\\\\')}\\\')}\')}')
```

achimnol (changes requested):
The new vfolder mount argument format should be also recognized by the `backend.ai session create` command as well!

src/ai/backend/client/cli/session/execute.py
```diff
@@ -245,11 +245,18 @@ def prepare_env_arg(env: Sequence[str]) -> Mapping[str, str]:
...
     """
     Parse the list of mount arguments into a list of
     vfolder name and in-container mount path pairs.
+
+    :param mount_args: A list of mount arguments such as
+        [
+            "vf-5194d5d8",
+            "vf-70b99ea5=/home/work/abc",
+            "vf-cd6c0b91:/home/work/zxc",
+        ]
```
kyujin-cho: When would be the case of this old `prepare_mount_arg()` being utilized on the new code base? I am in doubt of leaving old function on a new code.

kyujin-cho (changes requested):

src/ai/backend/client/cli/session/execute.py
rapsealk: I was considering keeping the old one for potential use cases. However, it would be reasonable to remove it since the new one (`prepare_mount_arg_v2()`) supersedes its functionality.

kyujin-cho: Yes removing old one looks ideal. Please update the PR accordingly.

kyujin-cho (approved):

PR #1851, opened 2024-01-19
feat: Add session tag in key-value format

### LINKED ISSUE #390: Add tags for compute sessions
It will be useful for users or admins to add, remove, or filter compute sessions with tags just like we can tag AWS EC2 instances.

Yaminyam:
close #390
The session tag function, which previously saved text like a regular memo, can now be saved in key-value format.
The key is limited to a maximum of 128 characters, and the value is limited to a maximum of 256 characters.
Input requires the form 'key1=value1, key2=value2, ...'.

- [ ] Test case(s) to:
- [ ] Documentation

----
📚 Documentation preview 📚: https://sorna--1851.org.readthedocs.build/en/1851/

----
📚 Documentation preview 📚: https://sorna-ko--1851.org.readthedocs.build/ko/1851/


src/ai/backend/client/utils.py
fregataa, line 1: How about using(or updating) `ai.backend.cli.params.CommaSeparatedKVListParamType`?

src/ai/backend/manager/models/session.py
fregataa, line 1: We have customized column schemas in `ai.backend.manager.models.base`.
It will be good for performance to set a constraint in SQL but how about making(or updating) a column schema?
`ai.backend.manager.models.base.StructuredJSONColumn` would work but it might need update.

PR #1858, opened 2024-01-23
feature: Apply modification on model service

### LINKED ISSUE #1848: Allow modification of inference session resource params preserving endpoints
Let's allow modification of the resource configuration (e.g., the number of CPU cores and GPUs) used by each container of existing service endpoints, without recreating the entire endpoints but just applying the new configuration from the newly spawned containers.

This way, we could allow our users to change the resource configs and applying them by shrinking/expanding the desired routing count on the fly.

For now, let's allow changing the resource slots only.
We could consider changing resource groups and other attributes as well, but we need to discuss potential side-effects of them and how we could present the scenarios to the users without intimidating them.

lizable:
This PR resolves #1848.

- [ ] Test case(s) to:
- [ ] Documentation


lizable:
Closing this PR again since too many and unnecessary merged commits from main branch is in this PR.

PR #1859, opened 2024-01-23
feature: mutation query for endpoint

lizable:

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/models/endpoint.py
+++ b/src/ai/backend/manager/models/endpoint.py
@@ -2,7 +2,7 @@ import datetime
 import logging
 import uuid
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Iterable, List, Mapping, Optional, Sequence
+from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Mapping, Optional, Sequence
 
 import graphene
 import jwt
@@ -15,7 +15,7 @@ from sqlalchemy.orm import relationship, selectinload
 from sqlalchemy.orm.exc import NoResultFound
 
 from ai.backend.common.logging_utils import BraceStyleAdapter
-from ai.backend.common.types import ClusterMode
+from ai.backend.common.types import ClusterMode, ResourceSlot
 from ai.backend.manager.defs import SERVICE_MAX_RETRIES
 
 from ..api.exceptions import EndpointNotFound, EndpointTokenNotFound
@@ -31,6 +31,8 @@ from .base import (
     PaginatedList,
     ResourceSlotColumn,
     URLColumn,
+    set_if_set,
+    simple_db_mutate,
 )
 from .image import ImageRow
 from .routing import RouteStatus, Routing
@@ -43,6 +45,7 @@ __all__ = (
     "Endpoint",
     "EndpointLifecycle",
     "EndpointList",
+    "ModifyEndpoint",
     "EndpointTokenRow",
     "EndpointToken",
     "EndpointTokenList",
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/models/endpoint.py
```diff
@@ -631,6 +645,86 @@ class Meta:
...
+
+
+class ModifyEndpoint(graphene.Mutation):
+    class Arguments:
+        endpoint_id = graphene.UUID(required=True)
+        props = ModifyEndpointInput(required=True)
+
+    ok = graphene.Boolean()
+    msg = graphene.String()
+    endpoint = graphene.Field(lambda: Endpoint, required=False)
```
yomybaby: Please add a description to introducing fileds for version compatibility like below:
https://github.com/lablup/backend.ai/blob/d8ac38b4752e40d37444438c54e0ddbb88875038/src/ai/backend/manager/models/group.py#L261

yomybaby (changes requested):

src/ai/backend/manager/models/endpoint.py
lizable: @yomybaby I think field `endpoint` should be applied in 23.09.9, rather than 24.03.0, since the milestone of this feature is 23.09. I'll apply it right away.

src/ai/backend/manager/models/endpoint.py
```diff
@@ -631,6 +645,87 @@ class Meta:
...
+            query = (
+                sa.select(EndpointRow)
+                .where(EndpointRow.id == row.id)
+                .options(
+                    joinedload(EndpointRow.image_row),
+                    selectinload(EndpointRow.routings),
+                    joinedload(EndpointRow.session_owner_row),
+                    joinedload(EndpointRow.created_user_row),
+                )
+            )
```
kyujin-cho: Please use `Endpoint.get()` ([ref](https://github.com/lablup/backend.ai/blob/0270a0fdbbdf3c6784f7c5d0d9dd3dd36c62ab34/src/ai/backend/manager/models/endpoint.py#L202-L240)) instead of directly constructing query.

src/ai/backend/manager/models/endpoint.py
```diff
@@ -631,6 +645,87 @@ class Meta:
...
+    cluster_mode = graphene.String()
+    cluster_size = graphene.Int()
+    desired_session_count = graphene.Int()
+    image = ImageRefType()
+    name = graphene.String()
+    resource_group = graphene.String()
+    open_to_public = graphene.Boolean()
+
+
+class ModifyEndpoint(graphene.Mutation):
```
kyujin-cho, line 664: There is no permission evaluation at all for this mutation - that means unprivileged users can use this mutation to update endpoint information of others. Please update the endpoint to include permission measurements.

kyujin-cho (changes requested):

src/ai/backend/manager/models/endpoint.py
```diff
@@ -631,6 +646,107 @@ class Meta:
...
+                    .where(EndpointRow.id == endpoint_id)
+                    .where(
+                        (EndpointRow.session_owner == user_id)
+                        | (EndpointRow.created_user == user_id)
+                    )
+                )
+                endpoint_row = (await db_session.scalars(stmt)).first()
+                if endpoint_row is None:
+                    raise EndpointNotFound
+
```
kyujin-cho: @fregataa Nice work, but still a little bit short to reach our goal - `ADMIN`s without `SUPERADMIN` privileges should not be able to alter configuration of endpoints from domains the account does not take control of.

kyujin-cho (changes requested):

src/ai/backend/manager/models/endpoint.py
```diff
@@ -379,6 +384,9 @@ class Meta:
 
     endpoint_id = graphene.UUID()
     image = graphene.String()
+    image_id = graphene.UUID(description="Added at 23.09.9")
+    image_row = graphene.Field(Image, description="Added at 23.09.9")
+    architecture = graphene.String()
```
yomybaby: Please add `architecture` description with the version

yomybaby (changes requested):

src/ai/backend/manager/models/endpoint.py
lizable: Bulls eye! I just applied the description. thx!

src/ai/backend/manager/models/endpoint.py
```diff
@@ -379,6 +384,9 @@ class Meta:
 
     endpoint_id = graphene.UUID()
     image = graphene.String()
+    image_id = graphene.UUID(description="Added at 23.09.9")
+    image_row = graphene.Field(Image, description="Added at 23.09.9")
```
kyujin-cho: Since we are now able to supply full information of a single `Image` info, including string reference and ID, I think it is better to retire both `image` / `image_id` field and instead encourage user to use `image_row` field.

kyujin-cho: `row` is a very specific noun, representing database record. Since the intension of serving this information is to expose full information of image "object", let's change the naming to do so.
```suggestion
    image = graphene.String()
    image_id = graphene.UUID(description="Added at 23.09.9")
    image_object = graphene.Field(Image, description="Added at 23.09.9")
```

kyujin-cho (changes requested):

src/ai/backend/manager/models/endpoint.py
lizable: You mean It's better to make `image` field be deprecated from version 23.09.9?

kyujin-cho (approved):

PR #1860, opened 2024-01-23
fix: add default group when cil signup and return groups info

### LINKED ISSUE #1625: User who doesn't have a project cannot log in.

Linux (x86-64)

### Backend.AI version

1419cf3

### Describe the bug

If you create a user without specifying the project, that user cannot log in.
Let's add default group when creating a user using cli or add a warning message before creating a new user.

### To Reproduce

_No response_

### Expected Behavior

_No response_

### Anything else?

This is a regression from #1022.

minseokey:
This PR resolves #1625 by add default-group when signup in cli-env

1. Now when user signup at cil-mode, add default-group
2. When signup at cil-env, it return user's group info

스크린샷 2024-01-23 오후 4 51 45

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/client/cli/admin/user.py
+++ b/src/ai/backend/client/cli/admin/user.py
@@ -255,6 +255,13 @@ def add(
                 allowed_client_ip=allowed_ip,
                 description=description,
                 sudo_session_enabled=sudo_session_enabled,
+                fields=(
+                    user_fields["domain_name"],
+                    user_fields["email"],
+                    user_fields["username"],
+                    user_fields["uuid"],
+                    user_fields["groups"],
+                ),
             )
         except Exception as e:
             ctx.output.print_mutation_error(
@@ -270,6 +277,9 @@ def add(
                 action_name="add",
             )
             sys.exit(ExitCode.FAILURE)
+
+        # preprocessing groups
+        data["user"]["groups"] = " ,".join(item["name"] for item in data["user"]["groups"])
         ctx.output.print_mutation_result(
             data,
             item_name="user",
--- a/src/ai/backend/manager/models/user.py
+++ b/src/ai/backend/manager/models/user.py
@@ -657,10 +657,18 @@ class CreateUser(graphene.Mutation):
             model_store_query = sa.select([groups.c.id]).where(
                 groups.c.type == ProjectType.MODEL_STORE
             )
+            default_query = sa.select([groups.c.id]).where(groups.c.name == "default")
+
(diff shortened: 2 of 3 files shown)


minseokey:
I changed the comma and space , so now
```Field        Value
-----------  ------------------------------------
ok           True
msg          success
domain_name  default
email        lms9999999@lablup.com
username     lms9999999@lablup.com
uuid         c9d92849-68ff-489e-9217-3a38962700aa
groups       default, model-store

File: src/ai/backend/manager/models/user.py [hunk 1]
```diff
@@ -657,10 +657,18 @@ async def _post_func(conn: SAConnection, result: Result) -> Row:
             model_store_query = sa.select([groups.c.id]).where(
                 groups.c.type == ProjectType.MODEL_STORE
             )
+            default_query = sa.select([groups.c.id]).where(groups.c.name == "default")
+
```
fregataa: It is not certain that the system has the "default" group named `default`.
We create a group named `default` for a development env for easy testing but not all systems have it

fregataa (commented):
I guess we can resolve #1625 by just printing (or logging) warning message after user creation

src/ai/backend/client/cli/admin/user.py
```diff
@@ -219,6 +219,17 @@ def list(ctx: CLIContext, status, group, filter_, order, offset, limit) -> None:
         "Note that this feature does not automatically install sudo for the session."
     ),
 )
+@click.option(
+    "-d",
+    "--default-group",
```
achimnol: The option just be a "group" with a default value, instead of "default group".

achimnol (changes requested):

src/ai/backend/manager/models/user.py
```diff
@@ -658,7 +660,19 @@ async def _post_func(conn: SAConnection, result: Result) -> Row:
...
+                    conn.scalar(sa.select([groups.c.id]).where(groups.c.name == name))
+                    for name in g_names
+                ])
+                for name, g_id in zip(g_names, g_ids):
+                    if g_id is None:
+                        raise GroupNotFound(f"Group '{name}' does not exist")
+                return g_ids
+
+            default_group_ids = await get_group_ids(conn, props.default_groups)
+            gids_to_join = [*default_group_ids, *group_ids, model_store_gid]
```
achimnol: We don't need to add new `default_groups` parameter in the API.
Reuse the `group_ids` parameter as intended.

achimnol (approved):

PR #1862, opened 2024-01-25
feature: Add a new client-side admin command group to manage quota scopes

### LINKED ISSUE #1614: CLI storage quota commands
### Main idea

Currently we do not have CLI command for storage quota.
Set quota, unset quota and get quota would be enough.
Let's implement it!

### Alternative ideas

_No response_

### Anything else?

```[tasklist]
### Tasks
- [x] remove `--quota` parameter from `vfolder create` CLI command (per-folder quota is deprecated for now)
- [x] implement new CLI command to manipulate quota value of eligible entities (either user or project)
```

SangwonYoon:
This PR resolves #1614 by adding new CLI command for get, set, and unset quota scope. Additionally, since per-folder quota has been deprecated, `--quota` parameter from `vfolder create` CLI command has been removed.

1. CLI command for get quota scope.

```
Usage: backend.ai admin quota-scope get [OPTIONS] HOST IDENTIFIER

  Get a quota scope.

  HOST: Name of the vfolder host (storage volume) to query the quota scope.
  IDENTIFIER: ID or email for user, ID or name for project.

Options:
  -d, --domain, --domain-name TEXT
                                  Specify the domain name
  -t, --type [user|project]       Specify per-user quota scope or per-project quota scope
  -h, --help                      Show this message and exit.
```

2. CLI command for set quota scope.

```
Usage: backend.ai admin quota-scope set [OPTIONS] HOST IDENTIFIER LIMIT_BYTES

  Set a quota scope.

  HOST: Name of the vfolder host (storage volume) to set the quota scope.
  IDENTIFIER: ID or email for user, ID or name for project.
  LIMIT_BYTES: Byte-size to be allocated to quota scope of a user or project. (e.g., 1t, 500g)

Options:
  -d, --domain, --domain-name TEXT
                                  Specify the domain name
  -t, --type [user|project]       Specify per-user quota scope or per-project quota scope
  -h, --help                      Show this message and exit.
```

3. CLI command for unset quota scope.

```
Usage: backend.ai admin quota-scope unset [OPTIONS] HOST IDENTIFIER

  Unset a quota scope.

  HOST: Name of the vfolder host (storage volume) to unset the given quota scope.
  IDENTIFIER: ID or email for user, ID or name for project.

Options:
  -d, --domain, --domain-name TEXT
                                  Specify the domain name
  -t, --type [user|project]       Specify per-user quota scope or per-project quota scope
  -h, --help                      Show this message and exit.
```


src/ai/backend/client/func/quota_scope.py
```diff
@@ -0,0 +1,156 @@
...
+            mutation($storage_host_name: String!, $quota_scope_id: String!, $input: QuotaScopeInput!) {
+                set_quota_scope(storage_host_name: $storage_host_name, quota_scope_id: $quota_scope_id, props: $input) {
+                    quota_scope {$fields}
+                }
+            }
+        """
+        )
+        query = query.replace("$fields", " ".join(f.field_ref for f in fields))
+        inputs: dict[str, Any] = {}
+        set_if_set(inputs, "hard_limit_bytes", hard_limit_bytes)
```
fregataa: we don't need to use `set_if_set` here because `hard_limit_bytes` should be `None` or `Undefined` here.

src/ai/backend/client/func/quota_scope.py
```diff
@@ -0,0 +1,156 @@
...
+_default_quota_scope_fields = (
+    quota_scope_fields["quota_scope_id"],
+    quota_scope_fields["storage_host_name"],
+)
+
+
+class QuotaScope(BaseFunction):
+    @api_function
+    @classmethod
+    async def get_user_qsid(
```
fregataa: I think of 3 options
1. Rename this function to `get_user_info`
2. Use a function in `client.func.user.User`. There will be a minor update to the function in `client.func.user.User`.
3. Return full quota scope id like `user:fda...` rather than returning only `id`. Because the name of this function is `get_qsid` not just `get_id`.

src/ai/backend/client/func/quota_scope.py
```diff
@@ -0,0 +1,156 @@
...
+        variables = {
+            "domain_name": domain_name,
+            "email": email,
+        }
+        data = await api_session.get().Admin._query(query, variables)
+        return data["user"]
+
+    @api_function
+    @classmethod
+    async def get_project_qsid(
```
fregataa: Same here. we could use a function in `client.func.group.Group` or consider other options

fregataa (commented):
Good!! I left some comments

src/ai/backend/client/func/quota_scope.py
SangwonYoon: > I think of 3 options
> 
> 1. Rename this function to `get_user_info`
> 2. Use a function in `client.func.user.User`. There will be a minor update to the function in `client.func.user.User`.
> 3. Return full quota scope id like `user:fda...` rather than returning only `id`. Because the name of this function is `get_qsid` not just `get_id`.

Thanks. I would change the name of this function to `get_user_info`.

src/ai/backend/client/func/quota_scope.py
SangwonYoon: > we don't need to use `set_if_set` here because `hard_limit_bytes` should be `None` or `Undefined` here.

I'm not sure I fully understand your comment, but isn't `hard_limit_bytes` not `None` or `Undefined` here? Because `hard_limit_bytes` is required parameter in set CLI command.

fregataa: Sorry for my misspelling. `hard_limit_bytes` is **NOT** `None` or `Undefined`

src/ai/backend/client/cli/admin/quota_scope.py
```diff
@@ -0,0 +1,195 @@
...
+    group_fields["name"],
+)
+
+
+@admin.group()
+def quota_scope():
+    """Quota scope administration commands."""
+
+
+def _get_qsid_from_identifier(type: str, domain_name: str, identifier: str, session: Session):
```
fregataa:
```suggestion
def _get_qsid_from_identifier(type: str, domain_name: str, identifier: str, session: Session) -> Optional[str]:
```

src/ai/backend/client/cli/admin/quota_scope.py
```diff
@@ -0,0 +1,195 @@
...
+                    name=identifier,
+                    fields=_project_query_fields,
+                )
+                if project_info is None:
+                    return None
+                project_qsid = f"project:{project_info['id']}"
+            else:
+                # In case identifier is the project ID
+                project_qsid = f"project:{project_id}"
+            return project_qsid
```
fregataa: How about returning `ai.backend.common.types.QuotaScopeID` object rather than `str`?
Because the `QuotaScopeID` type defines serialization

src/ai/backend/client/func/quota_scope.py
```diff
@@ -0,0 +1,156 @@
...
+            mutation($storage_host_name: String!, $quota_scope_id: String!, $input: QuotaScopeInput!) {
+                set_quota_scope(storage_host_name: $storage_host_name, quota_scope_id: $quota_scope_id, props: $input) {
+                    quota_scope {$fields}
+                }
+            }
+        """
+        )
+        query = query.replace("$fields", " ".join(f.field_ref for f in fields))
+        inputs: dict[str, Any] = {}
+        inputs["hard_limit_bytes"] = hard_limit_bytes
```
fregataa:
```suggestion
        inputs: dict[str, Any] = {
            "hard_limit_bytes": hard_limit_bytes,
        }
```
just a minor suggestion!

src/ai/backend/client/func/quota_scope.py
```diff
@@ -0,0 +1,156 @@
...
+        variables = {
+            "storage_host_name": host,
+            "quota_scope_id": str(qsid),
+            "input": inputs,
```
fregataa, line 130:
```suggestion
            "input": {
                "hard_limit_bytes": hard_limit_bytes,
            },
```
or how about this?

src/ai/backend/client/cli/admin/quota_scope.py
```diff
@@ -0,0 +1,202 @@
+import sys
+import uuid
+from typing import Optional
+
+import click
+
+from ai.backend.cli.types import ExitCode
+from ai.backend.client.session import Session
+
+from ....common.types import QuotaScopeID
```
kyujin-cho: Relative importing components outside current package is discouraged, where package scope is determined by the wheel artifact - that says, you can not relative import variables or functions provided by file which in production will be installed by another Python package. 
```suggestion
from ai.backend.common.types import QuotaScopeID
```

src/ai/backend/client/func/quota_scope.py
```diff
@@ -0,0 +1,157 @@
...
+        domain_name: str,
+        email: str,
+        fields: Sequence[FieldSpec] = _default_user_fields,
+    ):
+        query = textwrap.dedent(
+            """\
+            query($domain_name: String!, $email: String!) {
+                user(domain_name: $domain_name, email: $email) {$fields}
+            }
+        """
```
kyujin-cho, line 48: `domain_name` input is not required when querying individual `user` object. Requiring corresponding domain name to every query to CLI user can be cumbersome. Let's just remove it.

src/ai/backend/client/func/quota_scope.py
```diff
@@ -0,0 +1,157 @@
...
+        domain_name: str,
+        name: str,
+        fields: Sequence[FieldSpec] = _default_project_fields,
+    ):
+        query = textwrap.dedent(
+            """\
+            query($domain_name: String!, $name: String!) {
+                groups_by_name(domain_name: $domain_name, name: $name) {$fields}
+            }
+        """
```
kyujin-cho, line 71: Same here.

kyujin-cho (changes requested):

kyujin-cho (changes requested):
Could you refactor current CLI implementations to load available quota scope types from `ai.backend.common.types.QuotaScopeType`, instead of defining separate string literals from `@click.option()`?

achimnol (approved):

PR #1866, opened 2024-01-29
feature: implement customization of name displayed in the Graylog source field

### LINKED ISSUE #1849: Make graylog `source` field customizable
### Main idea

Follow-up issue of #1138. By default, graypy represents identifier of the log sender as the hostname of machine. When a single node serves multiple Backend.AI components, it is really hard to distinguish which line belongs to which Backend.AI component. When looking at the implementation of graylog library, it seems like we can pass `localname=` positional parameter to constructor of `BaseGELFHandler()` class. 

### Alternative ideas

_No response_

### Anything else?

_No response_

SangwonYoon:
This PR resolves #1849 by implement customization of name displayed in the Graylog source field through config file.
In the previous implementation, the name in Graylog's source field was displayed as the machine's hostname, which made it difficult to distinguish between Backend.AI components. 
It is now possible to customize the name displayed in Graylog's source field by setting the 'localname' in the config file (such as manager.toml, agent.toml, etc.).

## Before
스크린샷 2024-01-29 오후 4 56 57

## After
스크린샷 2024-01-29 오후 4 57 42


src/ai/backend/common/logging.py
```diff
@@ -88,6 +88,8 @@
         t.Key("ca-certs", default=None): t.Null | t.String(allow_blank=True),
         t.Key("keyfile", default=None): t.Null | t.String(allow_blank=True),
         t.Key("certfile", default=None): t.Null | t.String(allow_blank=True),
+        t.Key("fqdn", default=True): t.Bool,
+        t.Key("localname", default=None): t.Null | t.String(allow_blank=True),
```
rapsealk: Allowing the use of a blank `localname` does not appear to be secure, following the implementation details in the `graypy`:

https://github.com/severb/graypy/blob/a5a82c535002db6015473cb59a7476361060d358/graypy/handler.py#L236-L237

configs/agent/sample.toml
```diff
@@ -267,6 +267,10 @@ ca-certs = ""
 keyfile = ""
 certfile = ""
 
+# If you want the source field in Graylog to display a name that distinguishes between components rather than the hostname, uncomment the lines below and insert a value into localname.
+# fqdn = false
+# localname = ""
```
rapsealk: It would be better to set any non-empty string as a default value, as `localname` does not allow blank.
```suggestion
# localname = "agent"
```

configs/manager/sample.toml
```diff
@@ -197,6 +197,10 @@ ca-certs = ""
 keyfile = ""
 certfile = ""
 
+# If you want the source field in Graylog to display a name that distinguishes between components rather than the hostname, uncomment the lines below and insert a value into localname.
+# fqdn = false
+# localname = ""
```
rapsealk: It would be better to set any non-empty string as a default value, as `localname` does not allow blank.
```suggestion
# localname = "manager"
```

Yaminyam (commented):
Taking local-name as the manager and agent by default is a value that assumes that there is only one manager and agent by default.
So how about not commenting it out and enabling it by default?

kyujin-cho:
> Taking local-name as the manager and agent by default is a value that assumes that there is only one manager and agent by default. So how about not commenting it out and enabling it by default?

@Yaminyam The phrase "manager" and "agent" is to illustrate example configuration and will not set by default. On current implementation the default behavior is not to fill out `localname` parameter and instead use underlying OS' hostname.

kyujin-cho (approved):

PR #1749, opened 2023-11-29
feature: model metadata query

fregataa:
resolves #1066 

# Example `model-definition.yaml` file
```yaml
models:
  - name: "simple-http-server"
    model_path: "/models"
    service:
      pre_start_actions:
        - action: run_command
          args:
            command: ["pip", "install", "-r", "/models/requirements.txt"]
      start_command:
        - python
        - /models/server.py
      port: 8000
      health_check:
        path: /health
        max_retries: 5
    metadata:
      author: "John Doe"
      version: 3
      created: "2019-01-01T00:00:00Z"
      last_modified: "2019-12-31T00:00:00Z"
      description: "This is a simple HTTP server"
      title: "Simple http server. Human readable name."
      task: "HTTP server"
      category: "langauge"
      label: 
        - "language"
      license: "Apache-2.0"
      min_resource:
        cpu: 1
        memory: 256m
```

# Query Model Info connection(list)
```
query ModelInfoQuery {
    model_infos {
        edges {
            cursor
            node {
                name
                author
                title
                version
                created_at
                modified_at
                description
                task
                category
                label
                license
                min_resource
            }
        }
        count
        pageInfo {
            hasNextPage
            hasPreviousPage
        }
    }
}
```

# Query single Model Info node
```
query ModelInfoQuery {
    model_info(id: "TW9kZWxJbmZvOjQ4M2RiY2Q5LTgxOGItNGJhYS1iMWI1LTFlNTY0NDI1NTczMw==") {
        name
        author
        title
        version
        created_at
        modified_at
        description
        task
        category
        label
        license
        min_resource
    }
}
```

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version
- [ ] Mention to the original issue
- [ ] Documentation
- [ ] API server-client counterparts (e.g., manager API -> client SDK)
- [ ] Test case(s) to demonstrate the difference of before/after

Issue #1861, opened 2024-01-24
Keypair data and count mismatch issues


agatha197:
### Main idea

The `keypair_list` returns only 7 user data, but the count is 14.
It looks like because of the mismatch between keypair list's `items` and `total_count` query.
Let's sync up both data.
<img width="619" alt="image" src="https://github.com/lablup/backend.ai/assets/28584164/8f7ea3d1-2184-4381-9c14-a73f2d2d5d49">
<img width="232" alt="image" src="https://github.com/lablup/backend.ai/assets/28584164/0f3b6374-c765-4fe5-b5c6-48d5410dca2b">

related issue: https://github.com/lablup/backend.ai/issues/1703

### Alternative ideas

It can be resolved using [AsyncNode](https://github.com/lablup/backend.ai/blob/main/src/ai/backend/manager/models/gql_relay.py#L77).

rapsealk:
Resolved by #1869

PR #1869, opened 2024-01-30
fix: Keypair double-count across multiple projects

rapsealk:
This PR is a follow-up to #1749 and fixes double-count in keypairs across multiple projects.

This occurred when attempting to list keypairs using the command:

```sh
$ backend.ai admin keypair list
```

However, the command got stuck in an infinite loop. After investigating the code below:
https://github.com/lablup/backend.ai/blob/a4149f26bae3ff510ecf54e1b355f60b8206b481/src/ai/backend/client/output/console.py#L130-L139
I observed that `fetch_func()` returned an empty list (`len(result.items) == 0`). Despite having only 6 keypairs, `result.total_count` was 12, indicating duplication.

Further investigation revealed that `Keypair.load_count()` did not aggregate the results, leading to the duplication issue.

https://github.com/lablup/backend.ai/blob/7c2ab43d5f4787caf29631d7ee4c3599ca2dfc4a/src/ai/backend/manager/models/keypair.py#L324-L342

image

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/changes/.fix.md
+++ b/changes/.fix.md
@@ -1,0 +1,1 @@
+Fix double-count issue caused by keypairs belonging to multiple projects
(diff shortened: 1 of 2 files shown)


fregataa:
https://github.com/lablup/backend.ai/issues/1861
this is a related issue.
this PR resolves the infinite fetch, but still we cannot list keypairs whose user is not associated with any project.

achimnol (approved):

PR #1872, opened 2024-01-31
refactor: include the full image name in InvalidImageTag exception

### LINKED ISSUE #1456: Show the full container image name when there are unrecognized platform tags
When raising `InvalidImageTag` exception during validation of container image tags, let's include the full image name as well as the invalid tag component, for ease of debugging and error handling.

SangwonYoon:
This PR resolves #1456 by include the full container image name in `InvalidImageTag` exception.

## Before
스크린샷 2024-01-31 오후 12 13 42

## After
스크린샷 2024-01-31 오후 12 12 38


src/ai/backend/common/docker.py
```diff
@@ -360,17 +360,17 @@ class PlatformTagSet(Mapping):
     _data: dict[str, str]
     _rx_ver = re.compile(r"^(?P<tag>[a-zA-Z]+)(?P<version>\d+(?:\.\d+)*[a-z0-9]*)?$")
 
-    def __init__(self, tags: Iterable[str]):
+    def __init__(self, tags: Iterable[str], value: str = None):
```
achimnol: As we update this code, let's just add a missing type annotation as a part of small refactoring :wink::
```suggestion
    def __init__(self, tags: Iterable[str], value: str = None) -> None:
```

src/ai/backend/common/docker.py
```diff
@@ -414,17 +414,18 @@ def __init__(
         architecture: str = "x86_64",
         is_local: bool = False,
     ):
```
achimnol: Same here:
```suggestion
    ) -> None:
```

src/ai/backend/common/exception.py
```diff
@@ -59,7 +59,10 @@ class InvalidImageTag(ValueError):
     """
 
     def __str__(self) -> str:
-        return f"Invalid or duplicate image name tag: {self.args[0]}"
+        if len(self.args) > 1 and self.args[1] is not None:
+            return f"Invalid or duplicate image name tag: {self.args[0]}, full image name: {self.args[1]}"
+        else:
+            return f"Invalid or duplicate image name tag: {self.args[0]}"
```
achimnol: Instead of calculating `len(self.args)` like this, we could consider adding custom attributes to the exception class in a way to be safe with pickling and unpickling.

Check out the discussions here: https://stackoverflow.com/a/27334007/122423

achimnol (changes requested):

src/ai/backend/common/exception.py
```diff
@@ -58,11 +58,18 @@ class InvalidImageTag(ValueError):
     ```
     """
 
+    def __init__(self, tag: str, value: str | None = None) -> None:
```
achimnol: It is hard to see what "value" means here just reading this part of the code, because "value" is a generic word and usually in exceptions like this one (a subclass of `ValueError`) it indicates the value that is incorrect, i.e., "tag" here.

How about renaming it like "full_name" to reveal the intention more clearly (i.e., the contextual information where the errorneous value is taken from)?

src/ai/backend/common/exception.py
```diff
@@ -58,11 +58,18 @@ class InvalidImageTag(ValueError):
...
+        super(InvalidImageTag, self).__init__(tag, value)
```
achimnol: In Python 3, we can just do `super().__init__(tag, value)`.

achimnol (changes requested):

achimnol (approved):

PR #1884, opened 2024-02-04
feat: improve reliability & usability of VFolder removal API

### LINKED ISSUE #1797: Reliable VFolder deletion(purge)
Current version: main(before release 24.03.0)

Currently, the vfolder deletion task is run by `aiotools.PersistentTaskGroup`.The steps of the task is like below

1. Update status value of DB vfolders table to `PURGE_ONGOING`.
1. Request folder deletion to each storage proxy.
1. Delete records of DB vfolders table that has been successfully deleted.
   The clients do not get informed when any error occurs during this long-term tasks.There are several ways to resolve this issue.

- ~<sub>Use `BackgroundTask` and its server-side response</sub>~
- Convert folder deletion from blocking response to event-driven pattern (using Redis)
- Update vfolder name before delete-forever to allow creating a new vfolder with the same name
  Please leave any other idea.

Related issues:

- #767

fregataa:
resolves #1797 

~~blocked by #1892~~

~~we do not need to back-port Graphene relay node part to 23.09~~. ~~We are now able to backport relay node impl to 23.09~~
Let's not backport this PR

- Enable to track VFolder deletion task through Backgruond task. The progress data of background task will contain `log_level` field
- Apply `asyncio.Semaphore` to limit too many concurrent deletion task for one storage-proxy process and apply File-lock to limit concurrent deletion on the same directory
~~Add `move_to_trash_vfolder`, `restore_vfolder`, `delete_forever_vfolder` and `purge_vfolder` GraphQL endpoints~~
Adding GQL APIs will be handled in another PR

- [x] Update of end-to-end CLI integration tests in `ai.backend.test`
- [x] Test case(s) to:

----
📚 Documentation preview 📚: https://sorna--1884.org.readthedocs.build/en/1884/

----
📚 Documentation preview 📚: https://sorna-ko--1884.org.readthedocs.build/ko/1884/


achimnol:
Need to limit the concurrent deletion jobs for each vfolder to one and the maximum concurrent number of vfolder deletion jobs for each storage volume (configurable, defaults to 20).

src/ai/backend/client/cli/vfolder.py
```diff
@@ -185,41 +185,25 @@ def purge(name):
 
 @vfolder.command()
 @click.argument("name", type=str)
-def delete_trash(name):
+def delete_forever(name):
```
kyujin-cho, line 205: Please add an alias to old command name so that this change won't break the compatibility.

src/ai/backend/client/func/vfolder.py
```diff
@@ -284,15 +297,16 @@ async def restore(self):
         return await self._restore()
 
     @api_function
-    async def delete_trash(self) -> Mapping[str, Any]:
+    async def delete_forever(self) -> Mapping[str, Any]:
```
kyujin-cho: Same here.

src/ai/backend/common/bgtask.py
```diff
@@ -86,6 +88,7 @@ async def _pipe_builder(r: Redis) -> Pipeline:
                     "current": str(current),
                     "total": str(total),
                     "msg": message or "",
+                    "log_type": str(log_type),
```
kyujin-cho: In my opinion `LogLevel` (or `log_level`) better suits in this case.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -1587,6 +1548,74 @@ class Meta:
...
+                        "volume": volume_name,
+                        "vfid": str(folder_id),
+                    },
+                ) as (_, resp):
+                    pass
+            except (VFolderOperationFailed, InvalidAPIParameters) as e:
+                if e.status == 404:
+                    row_deletion_infos.append(vfolder_info)
+                    progress_msg = (
+                        f"VFolder not found, transit status to DELETE_COMPLETE (id: {folder_id})"
```
kyujin-cho: Storage proxy raising 404 means it did not manage to find VFolder's corresponding physical path from its storage backend, not VFolder itself. This message needs to be rewrited so that it can represent situation more accurately.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -1587,6 +1548,74 @@ class Meta:
...
 
+async def delete_vfolders(
+    requested_vfolders: Sequence[VFolderDeletionInfo],
+    *,
+    storage_manager: StorageSessionManager,
+    db: ExtendedAsyncSAEngine,
+    reporter: ProgressReporter | None = None,
+) -> None:
+    async def _task(reporter: ProgressReporter | None):
+        row_deletion_infos: list[VFolderDeletionInfo] = []
```
kyujin-cho: The name `row_deletion_infos` confuses reader as if it was meant to contain information about which folders to delete. Please update the variable identifier so that it can be shown as a list to store information of deleted folder.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -1587,6 +1548,74 @@ class Meta:
...
+        failed_deletion: list[tuple[VFolderDeletionInfo, str]] = []
```
kyujin-cho: Simillar situation here - the name `failed_deletion` itself can be considered to reader as a flag that indicates if the operation is failed or not.

src/ai/backend/storage/config.py
```diff
@@ -47,6 +47,7 @@
                     ),
                     t.Key("event-loop", default="asyncio"): t.Enum("asyncio", "uvloop"),
                     t.Key("scandir-limit", default=1000): t.Int[0:],
+                    t.Key("delete-concurrency", default=20): t.Int[1:],
```
kyujin-cho: I could not catch the meaning of this config key. For better representation the name has to be a phrase which more highlights maximum number of folders can be removed simultaneously.

kyujin-cho (changes requested):

src/ai/backend/manager/models/vfolder.py
```diff
@@ -1779,6 +1784,35 @@ async def delete_vfolders(
...
+    Spawn vfolder deletion tasks grouped by a name of storage proxies.
+    """
+
+    def _keyfunc(item: VFolderDeletionInfo) -> str:
+        proxy_name, _ = storage_manager.split_host(item.host)
+        return proxy_name
+
+    async with aiotools.TaskGroup() as tg:
+        for _, vfolder_iterator in itertools.groupby(
+            sorted(requested_vfolders, key=_keyfunc), key=_keyfunc
```
rapsealk, line 1804: In this case, using `typing.Iterable` seems more suitable than `typing.Sequence` for `requested_vfolders`.
(ref: https://docs.python.org/3/library/functions.html#sorted)

fregataa:
replaced by #2737, #2738, #2739

PR #1885, opened 2024-02-05
docs: add helper message for the allowed_session_types

### LINKED ISSUE #1293: Add error message when creating/modifying a scaling group without `allowed_session_type`
### Main idea

[The min length of allowed_session_type is 1](https://github.com/lablup/backend.ai/blob/cf60bd4d01ca4d296ada1d623c5c710cb33976a6/src/ai/backend/manager/models/scaling_group.py#L87). So it is not possible to create or modify a new scaling group without `allowed_session_type` and if the user tries to do that, it returns an inappropriate error message. 
Let's add an error message for this situation.
image

### Alternative ideas

_No response_

### Anything else?

_No response_

SangwonYoon:
This PR resolves #1293 by adding helper message for the `allowed_session_types` in create and modify commands of scaling group.
The commands work well even without `allowed_session_types`, but since `allowed_session_types` is set to the default values of `interactive` and `batch`, it may not operate as intended by the user. 
Therefore, a helper message has been added.

### DIFF
--- a/src/ai/backend/client/cli/admin/scaling_group.py
+++ b/src/ai/backend/client/cli/admin/scaling_group.py
@@ -100,7 +100,7 @@ def list(ctx: CLIContext) -> None:
     default="{}",
     help="""
         Set scheduler options as a JSON string.
-        For 'allowed_session_types', if the key does not exist, the values 'interactive' and 'batch' are set.
+        If the 'allowed_session_types' key is not specified, the policy defaults to accept both 'interactive' and 'batch'.
         """,
 )
 @click.option(
@@ -212,7 +212,7 @@ def add(
     default=undefined,
     help="""
         Set scheduler options as a JSON string.
-        For 'allowed_session_types', if the key does not exist, the values 'interactive' and 'batch' are set.
+        If the 'allowed_session_types' key is not specified, the policy defaults to accept both 'interactive' and 'batch'.
         """,
 )
 @click.option(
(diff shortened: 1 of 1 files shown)


src/ai/backend/client/cli/admin/scaling_group.py
```diff
@@ -98,7 +98,10 @@ def list(ctx: CLIContext) -> None:
     "--scheduler-opts",
     type=JSONParamType(),
     default="{}",
-    help="Set scheduler options as a JSON string.",
+    help="""
+        Set scheduler options as a JSON string.
+        For 'allowed_session_types', if the key does not exist, the values 'interactive' and 'batch' are set.
```
achimnol: Let's prefer the active voice.
```suggestion
        If the 'allowed_session_types' key is not specified, the policy defaults to accept both 'interactive' and 'batch'.
```

src/ai/backend/client/cli/admin/scaling_group.py
```diff
@@ -207,7 +210,10 @@ def add(
     "--scheduler-opts",
     type=OptionalType(JSONParamType),
     default=undefined,
-    help="Set scheduler options as a JSON string.",
+    help="""
+        Set scheduler options as a JSON string.
+        For 'allowed_session_types', if the key does not exist, the values 'interactive' and 'batch' are set.
```
achimnol: Same here.
```suggestion
        If the 'allowed_session_types' key is not specified, the policy defaults to accept both 'interactive' and 'batch'.
```

achimnol (changes requested):

fregataa (approved):

PR #1887, opened 2024-02-05
feature: add some exceptions when schedule computing session

### LINKED ISSUE #1814: Better log when no agent is available for schedule Computing session
When we try to schedule a `Computing session` but there is no available agent in the scaling group, the log is like below.

```
No agents found to be compatible with the image architecture (image[0]: ...)
```
The problem is that the above logging is very confusing when the reason why there is no actual available agent is not image mismatch.
Let's add more **comprehensible and detail** logs.

minseokey:
This PR resolves #1814 
add some exceptions when schedule computing session

1. When candidates_agent itself is empty. 
2. When requested_architecture is empty.
3. When the agent's architecture and the requested architecture are different

it is possible to express the error situation in detail

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -714,12 +714,16 @@ async def _schedule_single_node_session(
             raise GenericBadRequest(
                 "Cannot assign multiple kernels with different architectures' single node session",
             )
+        if requested_architectures is None:
+            raise GenericBadRequest("Requested session has no architecture information")
```
fregataa: This cannot happen because `requested_architectures` is always `set`, not `None`

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -714,12 +714,16 @@ async def _schedule_single_node_session(
...
         requested_architecture = requested_architectures.pop()
         compatible_candidate_agents = [
             ag for ag in candidate_agents if ag.architecture == requested_architecture
         ]
 
         try:
+            if candidate_agents is None:
+                raise InstanceNotAvailable(extra_msg=("No agents are registered with the manager"))
```
fregataa: Same here. `candidate_agents` is always a `list`, not `None`

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -1024,6 +1028,10 @@ async def _schedule_multi_node_session(
                         compatible_candidate_agents = [
                             ag for ag in candidate_agents if ag.architecture == kernel.architecture
                         ]
+                        if candidate_agents is None:
+                            raise InstanceNotAvailable(
+                                extra_msg=("No agents are registered with the manager")
+                            )
```
fregataa: Same here. `candidate_agents` cannot be `None`

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -997,7 +1001,7 @@ async def _schedule_multi_node_session(
                                     AgentRow.occupied_slots,
                                 ]).where(AgentRow.id == agent.id)
                             )
-                        ).fecthall()[0]
+                        ).fetchall()[0]
```
fregataa, line 1004: 👍
Could you test whether this fix works well?

fregataa (commented):
Thanks! I left some comments

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -714,12 +714,16 @@ async def _schedule_single_node_session(
             raise GenericBadRequest(
                 "Cannot assign multiple kernels with different architectures' single node session",
             )
+        if not requested_architectures:
+            raise GenericBadRequest("Requested session has no architecture information")
```
fregataa: `requested_architectures` can be empty when the computing session does not have any sibling computing kernels.
How about checking session's sibling kernels instead of `requested_architectures`?
like this.
```python
if not sess_ctx.kernels:
    raise GenericBadRequest(f"Given session does not have any sibling kernel. (id: {sess_ctx.id}")
```

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -714,12 +714,16 @@ async def _schedule_single_node_session(
...
         requested_architecture = requested_architectures.pop()
         compatible_candidate_agents = [
             ag for ag in candidate_agents if ag.architecture == requested_architecture
         ]
 
         try:
+            if not candidate_agents:
+                raise InstanceNotAvailable(extra_msg="No agents are registered with the manager")
```
fregataa: 👍
"No agents are available for scheduling" error msg would be more comprehensible

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -1024,6 +1028,10 @@ async def _schedule_multi_node_session(
                         compatible_candidate_agents = [
                             ag for ag in candidate_agents if ag.architecture == kernel.architecture
                         ]
+                        if not candidate_agents:
+                            raise InstanceNotAvailable(
+                                extra_msg="No agents are registered with the manager"
+                            )
```
fregataa: same here!

changes/1887.feature.md
```diff
@@ -0,0 +1 @@
+add some exceptions when schedule computing session
```
fregataa: This will be logged on release note, so we should consider how users will read this and understand.
Something like "Eager error handling when scheduling compute session" would be readable for readers. We could use chatGPT or ask to other seniors for it !

fregataa (commented):
Looks good!

src/ai/backend/manager/scheduler/dispatcher.py
fregataa: Could you check the error message again? The sentence "Registered with the manager" can confuse clients

fregataa: Could you check the error message again? The sentence "Registered with the manager" can confuse clients

src/ai/backend/manager/scheduler/dispatcher.py
fregataa: same here

fregataa (commented):
The new change log is good! I left some comments

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -714,12 +714,18 @@ async def _schedule_single_node_session(
             raise GenericBadRequest(
                 "Cannot assign multiple kernels with different architectures' single node session",
             )
+        if not sess_ctx.kernels:
+            raise GenericBadRequest(
+                f"Given session does not have any sibling kernel. (id: {sess_ctx.id}"
```
achimnol: A session is the "parent" of kernels. They are not siblings. :wink:

achimnol (approved):

PR #1891, opened 2024-02-06
feature: add validate_image_alias, validate_image_canonical CLI command

### LINKED ISSUE #1088: Programmatic image spec validation & kernel image verification tool
Let's add a simple CLI utility to validate a local/remote image's labels and platform tag formats for ease of development and potential reuse in Forklift.    
Backend.AI requires special tags to be added at docker image in order to be included at its ecosystem:
- `ai.backend.kernelspec`: kernel spec version identifier. fixed to `1`
- `ai.backend.features`: intrinsic features advertised by image as capable of. In most cases `"batch query uid-match user-input"` will be used
- `ai.backend.role`: type of image. Currently there are two defined values:
  - `INTERACTIVE`: for general (ML/DL workloads, AI inferences, ...) purpose
  - `SYSTEM`: for kernels spawned by Backend.AI system, not user; e.g. SFTP session 
- `ai.backend.base-distro`: Fundamental linux distribution identifier of the docker image. Should be formed like `<distro><version>`. For example:
  - `ubuntu20.04`
  - `centos8.7`
  - `alpine3.8`
- `ai.backend.resource.min.<device>`: minimum number (discrete or fraction) of devices this image requires. All images have to define minimum resource at lease for `cpu` and `mem` devices. Without appropriate minimum resource defined the image will not accept requested accelerator to be attached to target kernel. That says, when you are going to allow any amount of CUDA devices to be available for your image, then you have to set `ai.backend.resource.min.cuda.shares` as `0`, instead of skipping the label.
- `ai.backend.resource.max.<device>`: maximum number (discrete or fraction) of devices this image can handle. Omittable.
- `ai.backend.service-ports`: Defines web applications exposed by the container (along with its local port number) to the end user. Seperated by comma(,) when multiple entries are present. Each item should be formed as `<service name>:<port type>:<port number>`.
  - service name: name of the service
  - port type: `preopen`, `http` or `tcp`. `preopen` type is dedicated for apps those started immediately upon loading the container and thus does not require any arbitrary startup procedures. Backend.AI agent expects matching service definition file to be present at `/etc/backend.ai/service-defs/<service name>.json` inside container for `http` and `tcp` type apps. ref: https://github.com/lablup/backend.ai/blob/5cf6ef93b843a2633d3ebecaa0cfc00bb0a37397/src/ai/backend/kernel/service.py#L41-L50

A valid example of Backend.AI-compatible image is like:
```dockerfile
LABEL ai.backend.kernelspec="1" \
      ai.backend.envs.corecount="OPENBLAS_NUM_THREADS,OMP_NUM_THREADS,NPROC" \
      ai.backend.features="batch query uid-match user-input" \
      ai.backend.role="INTERACTIVE" \
      ai.backend.base-distro="ubuntu18.04" \
      ai.backend.accelerators="cuda" \
      ai.backend.resource.min.cpu="1" \
      ai.backend.resource.min.mem="1g" \
      ai.backend.resource.min.cuda.device=0 \
      ai.backend.resource.min.cuda.shares=0 \
      ai.backend.service-ports="triton-server-http:preopen:8000,chatbot-ui:http:9000" 
```

minseokey:
This PR resolves #1088
Implement src/common/docker.py validate_image_labal with the CLI command, and It support both canonical_name and alias

`./backend.ai mgr image -h`
스크린샷 2024-02-06 오후 5 04 57

`./backend.ai mgr image validate-image-alias -h`
스크린샷 2024-02-06 오후 5 05 39

`./backend.ai mgr image validate-image-canonical -h`
스크린샷 2024-02-06 오후 5 06 12

ex) 
스크린샷 2024-02-06 오후 5 08 44

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/manager/cli/image.py
```diff
@@ -24,11 +29,11 @@ def cli() -> None:
     pass
 
 
-@cli.command()
+@cli.command(name="list")
```
achimnol, line 31: Good find! 👍🏼

changes/1891.feature.md
```diff
@@ -0,0 +1,2 @@
+add validate-image-canonical and validate-image-alias cli command
+
```
achimnol: Remove the extra trailing newline here as it may destroy the changelog formatting. Also, let's start the sentence with a capital letter, as you see in [the changelog](https://github.com/lablup/backend.ai/blob/main/CHANGELOG.md).

src/ai/backend/manager/cli/image.py
```diff
@@ -110,3 +115,36 @@ def alias(cli_ctx, alias, target, architecture) -> None:
...
+        for key, value in result.items():
+            print(f"{key[11:]:<20}: ", end="")
+            print(*value) if isinstance(value, list) else print(value)
+    except Exception:
+        sys.exit(ExitCode.FAILURE)
+
+
+@cli.command()
+@click.argument("canonical")
+@click.argument("architecture")
```
achimnol: I think it would be better to make this as an option instead of an argument, which defaults to the current client's architecture.

src/ai/backend/manager/cli/image_impl.py
```diff
@@ -192,3 +192,41 @@ async def dealias(cli_ctx, alias):
             log.exception("Alias not found.")
             return
         await session.delete(alias_row)
+
+
+async def validate_image_alias(cli_ctx, alias: str):
```
achimnol:
```suggestion
async def validate_image_alias(cli_ctx, alias: str) -> None:
```

src/ai/backend/manager/cli/image_impl.py
```diff
@@ -192,3 +192,41 @@ async def dealias(cli_ctx, alias):
...
+                raise UnknownImageReference
+
+            return validate_image_labels(image_row.labels)
+        except UnknownImageReference:
+            log.error(f"No images were found with alias: {alias}")
+        except Exception:
+            log.exception("An error occurred.")
+
+
+async def validate_image_canonical(cli_ctx, canonical: str, architecture: str):
```
achimnol:
```suggestion
async def validate_image_canonical(cli_ctx, canonical: str, architecture: str): -> None:
```

src/ai/backend/manager/cli/image.py
```diff
@@ -110,3 +115,36 @@ def alias(cli_ctx, alias, target, architecture) -> None:
...
+@cli.command()
+@click.argument("alias")
+@click.pass_obj
+def validate_image_alias(cli_ctx: CLIContext, alias: str) -> None:
+    """Validate a local/remote image's labels and platform tag formats by image alias."""
+    try:
+        result = asyncio.run(validate_image_alias_impl(cli_ctx, alias))
+        for key, value in result.items():
+            print(f"{key[11:]:<20}: ", end="")
+            print(*value) if isinstance(value, list) else print(value)
```
achimnol: The intention of `xxx_impl()` functions are to make the command handler "async".
Please move these result-processing logic into the implementation function.

src/ai/backend/manager/cli/image.py
```diff
@@ -110,3 +115,36 @@ def alias(cli_ctx, alias, target, architecture) -> None:
...
+    except Exception:
+        sys.exit(ExitCode.FAILURE)
```
achimnol: I think it is better to just raise appropriate Click exceptions instead of manually catching it here.
(ref: https://click.palletsprojects.com/en/8.1.x/api/#exceptions)

src/ai/backend/manager/cli/image.py
```diff
@@ -110,3 +115,36 @@ def alias(cli_ctx, alias, target, architecture) -> None:
...
+
+    Write in the format registry/name:tag architecture
+    """
+    try:
+        result = asyncio.run(validate_image_canonical_impl(cli_ctx, canonical, architecture))
+        for key, value in result.items():
+            print(f"{key[11:]:<20}: ", end="")
+            print(*value) if isinstance(value, list) else print(value)
+    except Exception:
+        sys.exit(ExitCode.FAILURE)
```
achimnol: Same here.

src/ai/backend/manager/cli/image_impl.py
```diff
@@ -192,3 +192,41 @@ async def dealias(cli_ctx, alias):
...
+        db.begin_session() as session,
+    ):
+        try:
+            image_row = await session.scalar(
+                sa.select(ImageRow).where(
+                    (ImageRow.name == canonical) & (ImageRow.architecture == architecture)
+                )
+            )
+            if image_row is None:
+                raise UnknownImageReference(f"{canonical}:{architecture}")
```
achimnol: Although there is no standardized rule yet, but I think it would be better to notate arhictecture using a slash after the image tag as it's a commonly used convention.
```suggestion
                raise UnknownImageReference(f"{canonical}/{architecture}")
```

achimnol (changes requested):

src/ai/backend/manager/cli/image.py
minseokey: How can I get the architecture of the client? @achimnol

achimnol: Refer `ai.backend.common.arch` module!

src/ai/backend/manager/cli/image_impl.py
```diff
@@ -192,3 +194,58 @@ async def dealias(cli_ctx, alias):
             log.exception("Alias not found.")
             return
         await session.delete(alias_row)
+
+
+async def validate_image_alias(cli_ctx, alias: str) -> None:
+    async with (
+        connect_database(cli_ctx.local_config) as db,
+        db.begin_session() as session,
```
fregataa:
```suggestion
        db.begin_readonly_session() as session,
```
Since the DB session is used for `SELECT` only, let's start readonly session here.

src/ai/backend/manager/cli/image_impl.py
```diff
@@ -192,3 +194,58 @@ async def dealias(cli_ctx, alias):
...
+    ):
+        try:
+            image_row = await ImageRow.from_alias(session, alias)
+            if image_row is None:
+                raise UnknownImageReference
```
fregataa:
```suggestion
            image_row = await ImageRow.from_alias(session, alias)
```
Don't need null-check here because `UnknownImageReference` is raised inside of `ImageRow.from_alias()` function

src/ai/backend/manager/cli/image_impl.py
```diff
@@ -192,3 +194,58 @@ async def dealias(cli_ctx, alias):
...
+                if key[:11] == "ai.backend.":
+                    print(f"{key[11:]:<30}: ", end="")
+                else:
+                    print(f"{key:<30}: ", end="")
+                print(*value) if isinstance(value, list) else print(value)
+
+        except UnknownImageReference:
+            log.error(f"No images were found with alias: {alias}")
+        except Exception:
+            log.exception("An error occurred.")
```
fregataa, line 215:
```suggestion
            log.exception(f"An error occurred. (e: {str(e)})")
```
Let's log a little bit more info about the error

src/ai/backend/manager/cli/image_impl.py
```diff
@@ -192,3 +194,58 @@ async def dealias(cli_ctx, alias):
...
+    cli_ctx, canonical: str, architecture: Optional[str] = None
+) -> None:
+    async with (
+        connect_database(cli_ctx.local_config) as db,
+        db.begin_session() as session,
+    ):
+        try:
+            # Default arch is current arch
+            if architecture is None:
+                architecture = CURRENT_ARCH
```
fregataa:
```python
architecture = architecture or CURRENT_ARCH
```
a little tip!

src/ai/backend/manager/cli/image_impl.py
```diff
@@ -192,3 +194,58 @@ async def dealias(cli_ctx, alias):
...
+        except Exception:
+            log.exception("An error occurred.")
+
+
+async def validate_image_canonical(
+    cli_ctx, canonical: str, architecture: Optional[str] = None
+) -> None:
+    async with (
+        connect_database(cli_ctx.local_config) as db,
+        db.begin_session() as session,
```
fregataa:
```suggestion
        db.begin_readonly_session() as session,
```
Same here, we only do `SELECT` job from the DB session

src/ai/backend/manager/cli/image_impl.py
fregataa: How about fetching image-row without `architecture` condition if it is `None`? 
Then the result of DB fetch can be multiple image rows and we just print them all

src/ai/backend/manager/cli/image_impl.py
```diff
@@ -192,3 +194,58 @@ async def dealias(cli_ctx, alias):
...
+    ):
+        try:
+            image_row = await ImageRow.from_alias(session, alias)
+            if image_row is None:
+                raise UnknownImageReference
+
+            for key, value in validate_image_labels(image_row.labels).items():
+                if key[:11] == "ai.backend.":
```
fregataa: How about just printing the full name of key?
Is it requested to trim out `ai.backend.` when print labels?

src/ai/backend/manager/cli/image_impl.py
```diff
@@ -192,3 +194,58 @@ async def dealias(cli_ctx, alias):
...
+                    print(f"{key[11:]:<30}: ", end="")
+                else:
+                    print(f"{key:<30}: ", end="")
+                print(*value) if isinstance(value, list) else print(value)
```
fregataa:
```python
if isinstance(value, list):
    value = f'[{", ".join(value)}]'
```
How about parse the list into readable string and print it?

fregataa (commented):
Thanks! I left some comments

src/ai/backend/manager/cli/image_impl.py
minseokey: 스크린샷 2024-02-14 오후 3 24 20
I didn't receive any requests, but I tried to trim unnecessary parts in order to look neat when printing.

src/ai/backend/manager/cli/image_impl.py
minseokey: Joong-gi said that the basic architecture would be better to follow the architecture of the user machine currently using, so I made it like this!
How about add click.option for all architecture's image info?

fregataa: I see. this is good enough for me!

minseokey: @fregataa I've added a  click.option  Please check it out :)

fregataa (approved):
LGTM 👍

src/ai/backend/manager/cli/image_impl.py
```diff
@@ -194,39 +196,64 @@ async def dealias(cli_ctx, alias):
...
+                            value = f'{", ".join(value)}'
+                        print(value)
 
-            return validate_image_labels(image_row.labels)
         except UnknownImageReference as e:
             log.error(f"{e}")
-        except Exception:
-            log.exception("An error occurred.")
+        except Exception as e:
+            log.exception(f"An error occurred. (e: {str(e)})")
```
achimnol: `log.exception()` inside an `except` block automatically includes the traceback. We may omit the explicit interpolation of `e` here.

src/ai/backend/manager/cli/image_impl.py
```diff
@@ -194,39 +196,64 @@ async def dealias(cli_ctx, alias):
...
+                    value = f'[{", ".join(value)}]'
+                print(value)
 
-            return validate_image_labels(image_row.labels)
         except UnknownImageReference:
             log.error(f"No images were found with alias: {alias}")
-        except Exception:
-            log.exception("An error occurred.")
+        except Exception as e:
+            log.exception(f"An error occurred. (e: {str(e)})")
```
achimnol: Same here.

achimnol (changes requested):

src/ai/backend/manager/cli/image_impl.py
achimnol: This is not required as per https://github.com/lablup/backend.ai/pull/1891#pullrequestreview-1887533685

achimnol (changes requested):
Overall, we should use the output framework to make it testable via using `--output=json` in [the integration tests](https://github.com/lablup/backend.ai/blob/main/src/ai/backend/test/cli_integration/admin/test_image.py).

Refer other recently written CLI module examples:
- https://github.com/lablup/backend.ai/blob/main/src/ai/backend/client/cli/service.py#L84
- https://github.com/lablup/backend.ai/blob/main/src/ai/backend/client/cli/admin/session.py#L266-L277

achimnol (approved):
I'm going to split fixing the CLI commands to use the output framework consistently as a separate issue.

PR #1894, opened 2024-02-06
fix: Modify help text for delete and purge domain

Yaminyam:
Domain's purge help text explains delete.
Of course, delete and purge are words used with similar meanings, but there is a command called domain delete, which can confuse users.
```
Delete the given virtual folder.
Purge the given virtual folder.
```
I modified it by referring to the vfolder delete and purge help phrases above.

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/client/cli/admin/domain.py
+++ b/src/ai/backend/client/cli/admin/domain.py
@@ -228,7 +228,7 @@ def update(
 @click.argument("name", type=str, metavar="NAME")
 def delete(ctx: CLIContext, name: str) -> None:
     """
-    Inactive an existing domain.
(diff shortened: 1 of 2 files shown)


achimnol (changes requested):
I think we need to elaborate what deletion and purging are here, in addition to unifying the terminology.

Yaminyam:
> I think we need to elaborate what deletion and purging are here, in addition to unifying the terminology.

This is what I agree with.
Since purge and delete are not often used together, I think it is necessary to inform users of this in detail.

achimnol (approved):

PR #1898, opened 2024-02-07
docs: modify helper message for mkdir command in the vfolder

SangwonYoon:
This PR is for modifying two helper message for mkdir command in the vfolder.

### 1. PATH's helper message
__Previous helper message__:
``` Parent directories are created automatically if they do not exist. ```

Unlike the helper message, if the -p option is not provided, the parent directory is not created automatically, and an error raised.

### 2. exist-ok option's helper message
__Previous helper message__:
``` Skip an error caused by file not found ```

__Changed helper message__:
``` Skip an error caused by file already exists ```


fregataa (changes requested):
LGTM !
Could you add the news fragment file to check timeline?

PR #1719, opened 2023-11-15
feature: impl graphql relay resolver for user and group

fregataa:
Implement async compatible graphql relay Node and Connection types.
Implement user/group relay Node and Connection.

# Why re-invent GraphQL relay?
Current Graphene(v 3.3.0) has a relay implementation but it does not support asynchronous resolver functions. Additionally, Graphene Connection resolver is implemented to only accept one complete array(`Iterable` value) without considering pagination, which is a huge performance issue.
Here, the newly "invented" Connection took pagination into account. But, pagination order of this Connection can only be either forward or backward in a single query, meaning that queries such as `first` and `last` together or `after` and `before` together are limited.

# Query all groups
```gql
# query all groups
query GroupsQuery {
  group_nodes {
    edges {
      node {
        id
        name
      }
    }
  }
}
```

# Query group with relay id
```gql
query GroupsQuery($id: String!) {
    group_node(id: $id) {
        id
        name
        user_nodes(order: "created_at", offset: 2, first: 3) {
            edges {
                node {
                    id
                    username
                    created_at
                }
            }
            count
            pageInfo {
                hasNextPage
                hasPreviousPage
            }
        }
    }
}

# variables
# {
#    "id": "R3JvdXBOb2RlOjJkZTJiOTY5LTFkMDQtNDhhNi1hZjE2LTBiYzhhZGIzYzgzMQ=="
# }
```

# Query node with node's global id
```gql
{
    node(id: "R3JvdXBOb2RlOjJkZTJiOTY5LTFkMDQtNDhhNi1hZjE2LTBiYzhhZGIzYzgzMQ==") {
        ... on GroupNode {
            name
            description
        }
    }
}

```

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version
- [ ] Documentation
- [ ] API server-client counterparts (e.g., manager API -> client SDK)
- [ ] Test case(s) to demonstrate the difference of before/after

PR #1916, opened 2024-02-16
fix: GQL Relay node query builder does not have `queryfilter_fieldspec` and `queryorder_colmap`

### LINKED ISSUE #1923: bug: `user_nodes` query does not get data correctly

macOS (Apple Silicon)

### Backend.AI version

24.03.0a2

### Describe the bug

If you use `user_node` GQL Relay query to get user nodes, you can't get data currectly.
I tested to use filter option only, and I got correctly data.
I think that manager has a issue about option as filter, offset, and first.

### To Reproduce

_No response_

### Expected Behavior

_No response_

### Anything else?

_No response_

fregataa:
follow-up #1719
resolves #1923

Clients cannot query relay nodes with "filter" and "order" argument properly.
Add `queryfilter_fieldspec` and `queryorder_colmap` as Relay node class variable and initiate `QueryFilterParser` or `QueryOrderParser` with them.

### DIFF
--- a/changes/1916.fix.md
+++ b/changes/1916.fix.md
@@ -1,0 +1,1 @@
+Fix GQL Relay node resolver to parse `filter` and `order` argument into SQL query properly.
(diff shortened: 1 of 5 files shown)


gee05053:
@fregataa 
Please check this issue too.🙏🏻

achimnol (approved):

Issue #1907, opened 2024-02-15
Rebase container registry config storage to relational database


kyujin-cho:
To define relations between container registries and user/project, let's move container registry configuration storage from etcd to relational database. In addition to every existing configurable values, new table should have a new column representing its global visibility - If this value represents positive meaning, every users can utilize the registry. Otherwise only users with eligible membership will be able to see it.

PR #1917, opened 2024-02-19
feat: Migrate container registry config storage from `Etcd` to `PostgreSQL`

jopemachine:
Resolves https://github.com/lablup/backend.ai/issues/1907.

## Overview

* Migrate container registry schema and data from `Etcd` to `PostgreSQL`
* Add `is_global` column to the `ContainerRegistry` table
* This PR does not modify the GraphQL schema, but just migrates the storage of container registry configurations.
* `/config/container-registries` API is now an alias to `/resource/container-registries`.

---

- [x] Test case(s) to:


src/ai/backend/manager/models/base.py
```diff
@@ -1178,7 +1186,13 @@ async def populate_fixture(
...
             continue
         assert not isinstance(rows, str)
-        table: sa.Table = getattr(models, table_name)
+
+        # TODO: Remove this try statement after all the table upgrade is done
+        try:
+            table: sa.Table = getattr(models, table_name)
+        except AttributeError:
+            table = await load_table(engine, table_name)
+
```
kyujin-cho: Could you describe more about what this block is for?

kyujin-cho (changes requested):
Current implementation does not have a flag field at `container_registry` table which indicates if this registry is available for every users. Please update the branch accordingly.

src/ai/backend/manager/models/base.py
jopemachine: In this block, the below tables data will be loaded by the `table: sa.Table = getattr(models, table_name)`.

```
domains,
user_resource_policies,
project_resource_policies,
groups,
scaling_groups,
sgroups_for_domains,
users,
association_groups_users,
keypair_resource_policies,
keypairs,
users,
resource_presets
```

You can see the specified tables are defined like the below form.

```
domains = sa.Table(
...
)
user_resource_policies = sa.Table(
...
)
project_resource_policies = sa.Table(
...
)

...
``` 

That's why they can be loaded through `getattr(models, table_name)`.

*However, we did not define the ContainerRegistry table in this way.* 

Since I cannot retrieve the table information through `getattr(models, table_name)`, I loaded and accessed the DB directly to load the table information.

There might be a better way to load table information that I am not aware of.
If you know any, I would appreciate it if you could share it with me.

jopemachine: Resolved in https://github.com/lablup/backend.ai/pull/1917/commits/ad881b22223cf90f0f3f0366fbe665721766a52f

src/ai/backend/manager/models/alembic/versions/1d42c726d8a3_add_container_registry_tables.py
```diff
@@ -0,0 +1,241 @@
...
+            async def _take_container_registries(etcd: AsyncEtcd):
+                result = await etcd.get_prefix(ETCD_CONTAINER_REGISTRY_KEY)
+                try:
+                    backup(result)
+                except Exception as e:
+                    return e
+                await etcd.delete_prefix(ETCD_CONTAINER_REGISTRY_KEY)
+                return result
+
+            etcd = get_async_etcd()
```
achimnol: It is safer to initialize any asyncio-depending objects inside the async function where the event loop is already initialized (i.e., `asyncio.get_running_loop()` returns successfully), because otherwise there may be two or more event loops implicitly created depending on how the libraries initialize themselves. Let's move this line into the inner async function.

src/ai/backend/manager/models/alembic/versions/1d42c726d8a3_add_container_registry_tables.py
```diff
@@ -0,0 +1,241 @@
...
+            queue.sync_q.put(asyncio.run(_take_container_registries(etcd)))
+
+        executor.submit(take_etcd_container_registries, queue)
+
+    maybe_registries = queue.sync_q.get()
+
+    if isinstance(maybe_registries, Exception):
```
achimnol: Is any unhandled exception inside `asyncio.run()` stored as a queue item?

src/ai/backend/manager/models/alembic/versions/1d42c726d8a3_add_container_registry_tables.py
```diff
@@ -0,0 +1,241 @@
...
+    return AsyncEtcd(
+        local_config["etcd"]["addr"],
+        local_config["etcd"]["namespace"],
+        scope_prefix_map,
+        credentials=etcd_credentials,
+    )
+
+
+def migrate_data_etcd_to_psql():
+    queue = janus.Queue()
```
achimnol: Let's add explicit type annotations here for future readers and editors.

achimnol: In your implementation, you don't interface this queue between non-async and async codes, but uses the queue to store the results from _non-async_ code in separate executor threads running in parallel.
As such, we can just use the stdlib's queue for simpler code!

src/ai/backend/manager/models/alembic/versions/1d42c726d8a3_add_container_registry_tables.py
```diff
@@ -0,0 +1,241 @@
...
+                result = await etcd.get_prefix(ETCD_CONTAINER_REGISTRY_KEY)
+                try:
+                    backup(result)
+                except Exception as e:
+                    return e
+                await etcd.delete_prefix(ETCD_CONTAINER_REGISTRY_KEY)
+                return result
+
+            etcd = get_async_etcd()
+            queue.sync_q.put(asyncio.run(_take_container_registries(etcd)))
```
achimnol: Since this particular line is not an async context (it's outside `asyncio.run()`!), you may use `queue` instead of `janus`.

tests/manager/conftest.py
```diff
@@ -470,6 +464,7 @@ async def clean_fixture():
                 await conn.execute((users.delete()))
                 await conn.execute((scaling_groups.delete()))
                 await conn.execute((domains.delete()))
+                await conn.execute(((await load_table(engine, "container_registries")).delete()))
```
achimnol: Why does this line use a specially crafted `load_table()` function?

achimnol (changes requested):

src/ai/backend/manager/models/alembic/versions/1d42c726d8a3_add_container_registry_tables.py
jopemachine: The exception is stored in the queue *only when the backup method fails* currently.
Should we catch all exceptions occurring in the thread and put them into the queue?

tests/manager/conftest.py
jopemachine: Since the `container_registries` variable was not defined as a table (unlike others), I wrote it this way to import the ContainerRegistry table object.

It would be nice if we could remove `load_table` if possible.

src/ai/backend/manager/models/alembic/versions/1d42c726d8a3_add_container_registry_tables.py
```diff
@@ -0,0 +1,241 @@
...
+
+import sqlalchemy as sa
+import trafaret as t
+from alembic import op
+
+from ai.backend.common import validators as tx
+from ai.backend.common.etcd import AsyncEtcd, ConfigScopes
+from ai.backend.manager.config import load
+from ai.backend.manager.models.base import IDColumn, StrEnumType, convention
+from ai.backend.manager.models.container_registry import ContainerRegistryRow, ContainerRegistryType
```
fregataa:
```suggestion
from ai.backend.manager.models.container_registry import ContainerRegistryType
```

It is highly recommended to define a table in migration files rather than to import a table because `ContainerRegistryRow` can change but DB schemas could remain outdated.
It would be also good to define enum types in migration files.

fregataa (changes requested):

src/ai/backend/manager/models/image.py
```diff
@@ -100,13 +95,10 @@ async def rescan_images(
...
-        registry_config_iv = t.Mapping(t.String, container_registry_iv)
-        latest_registry_config = cast(
-            dict[str, Any],
-            registry_config_iv.check(
-                await etcd.get_prefix("config/docker/registry"),
-            ),
-        )
+        async with db.begin_readonly_session() as session:
+            result = await session.execute(sa.select(ContainerRegistryRow))
+            latest_registry_config = {row.registry_name: row for row in result.scalars().all()}
```
fregataa:
```suggestion
            latest_registry_config = cast(dict[str, ContainerRegistryRow], {row.registry_name: row for row in result.scalars().all()})
```

Line no 89, we declared `registries` as a `dict[str, dict[str, Any]]`.
`get_container_registry_cls()` will raise error when `local=True` on line 131.
Let's add type casting here to remove type confusing

src/ai/backend/manager/container_registry/harbor.py
```diff
@@ -28,7 +30,15 @@ async def fetch_repositories(
...
         api_url = self.registry_url / "api"
-        registry_projects = self.registry_info["project"]
+
+        async with self.db.begin_readonly_session() as db_sess:
+            result = await db_sess.execute(
+                sa.select(ContainerRegistryRow.project).where(
+                    ContainerRegistryRow.registry_name == self.registry_info.registry_name
+                )
+            )
+            registry_projects = result.scalars().all()
```
fregataa:
```suggestion
            registry_projects = cast(list[str | None], result.scalars().all())
```
Let's add typing here.
Plus, `ContainerRegistryRow.registry_name` is nullable. Is it intended to fetch `null` registry name's project?

src/ai/backend/manager/container_registry/harbor.py
```diff
@@ -130,7 +140,15 @@ async def fetch_repositories(
...
         api_url = self.registry_url / "api" / "v2.0"
-        registry_projects = self.registry_info["project"]
+
+        async with self.db.begin_readonly_session() as db_sess:
+            result = await db_sess.execute(
+                sa.select(ContainerRegistryRow.project).where(
+                    ContainerRegistryRow.registry_name == self.registry_info.registry_name
+                )
+            )
+            registry_projects = result.scalars().all()
```
fregataa: Same here

src/ai/backend/manager/container_registry/base.py
```diff
@@ -53,15 +54,15 @@ def __init__(
...
+        registry_info: ContainerRegistryRow,
         *,
         max_concurrency_per_registry: int = 4,
         ssl_verify: bool = True,
     ) -> None:
         self.db = db
         self.registry_name = registry_name
         self.registry_info = registry_info
-        self.registry_url = registry_info[""]
+        self.registry_url = yarl.URL(registry_info.url)
```
fregataa, line 67: `registry_info.url` is nullable and `yarl.URL(None)` raises `TypeError`.
The argument should be string

fregataa (changes requested):
I left some comments about typing

src/ai/backend/manager/container_registry/harbor.py
jopemachine: No, `url`, `registry_name`, `type` should be required. I will fix it, thanks!

src/ai/backend/manager/container_registry/base.py
jopemachine: `url` should be required

src/ai/backend/manager/models/alembic/versions/1d42c726d8a3_add_container_registry_tables.py
```diff
@@ -0,0 +1,304 @@
...
+        ).scalar()
+
+        if registry_id is not None:
+            db_connection.execute(
+                sa.update(ImageRow)
+                .values(registry_id=str(registry_id))
+                .where(ImageRow.id == image_id)
+            )
+        else:
+            print(f"ContainerRegistry row not found for image {image_name}", file=sys.stderr)
```
fregataa: Let's iterate all `ContainerRegistries` and save their `id`. Such as `registry_name/project` key and `id` value dictionary.
Then we can bulk update image rows here.
Below is an example bulk update using `bindparam`. you can try and let me know how it works!
```python
stmt = (
    sa.update(ImageRow)
    .where(
   
 (ImageRow.registry_name.like(sa.bindparam("registry_name"))))
    )
    .values(ImageRow.registry_id == sa.bindparam("registry_id"))
)
value = [
{
    "registry_name": f'{d["registry_name"]}/{d["project"]}%',
    "registry_id: d["registry_id"],
} for d in data
]
db_connection.execute(stmt, value)
```

src/ai/backend/manager/models/alembic/versions/1d42c726d8a3_add_container_registry_tables.py
```diff
@@ -0,0 +1,304 @@
...
+    ContainerRegistry = get_container_registry_row_schema()
+
+    image_infos = db_connection.execute(sa.select([ImageRow.id, ImageRow.name])).fetchall()
+
+    for image_info in image_infos:
+        image_id, image_name = image_info
+
+        parts = image_name.split("/")
+        registry_name = parts[0]
+        project = parts[1]
```
fregataa:
```suggestion
        registry_name, _, project = image_name.partition("/")
```
Not a big deal, but we prefer to use `partition` if we are sure that the string should be divided into two strings.

src/ai/backend/manager/models/image.py
```diff
@@ -161,6 +155,7 @@ class ImageRow(Base):
     )
     tag = sa.Column("tag", sa.TEXT)
     registry = sa.Column("registry", sa.String, nullable=False, index=True)
+    registry_id = sa.Column("registry_id", sa.String, nullable=True)
```
fregataa:
```suggestion
    registry_id = sa.Column("registry_id", GUID, nullable=True)
```
Let's define this column as customized UUID type

src/ai/backend/manager/models/alembic/versions/1d42c726d8a3_add_container_registry_tables.py
jopemachine: We can't use `partition` like this.

For example, 
> image_name = "registry_name/project/something"
> image_name.partition("/")
('registry_name', '/', 'project/something')

'project/something' should be just project

src/ai/backend/common/docker.py
```diff
@@ -279,53 +272,63 @@ async def login(
...
-    if username is not None:
-        creds["username"] = username
-    password = item.get("password")
-    if password is not None:
-        creds["password"] = password
-    return yarl.URL(registry_addr), creds
+def is_known_registry(
+    registry_name: str,
+    project: str,
+    known_registries: Optional[Mapping[str, Mapping[str, Any]] | Sequence[str]] = None,
```
fregataa:
```suggestion
    known_registries: dict[str, dict[str, Any] | list[str]] | None,
```
if `known_registries` can be one of (`dict`, `list`, `None`), this would be a better type hint

src/ai/backend/common/docker.py
```diff
@@ -279,53 +272,63 @@ async def login(
...
+def is_in_known_library_dict(
+    registry_name: str,
+    project: str,
+    known_registries: Optional[Mapping[str, Mapping[str, Any]] | Sequence[str]],
+) -> bool:
+    return (
+        isinstance(known_registries, Mapping)
+        and project in known_registries.keys()
+        and registry_name in known_registries[project]
+    )
```
fregataa:
```suggestion
    return (
        isinstance(known_registries, dict)
        and project in known_registries
        and registry_name in known_registries[project]
    )
```

src/ai/backend/common/docker.py
```diff
@@ -410,7 +413,7 @@ class ImageRef:
     def __init__(
         self,
         value: str,
-        known_registries: Optional[Mapping[str, Any] | Sequence[str]] = None,
+        known_registries: Optional[Mapping[str, Mapping[str, Any]] | Sequence[str]] = None,
```
fregataa:
```suggestion
        known_registries: Mapping[str, Mapping[str, Any]] | Sequence[str] | None = None,
```
same here

src/ai/backend/manager/registry.py
```diff
@@ -192,6 +194,54 @@
...
+            )
+        ).fetchall()[0]
+
+        if not result:
+            raise UnknownImageRegistry(registry_id)
+
+        url, username, password = result
+        creds = {"username": username, "password": password}
+
+        return yarl.URL(url), creds
```
fregataa: How about locating those functions in `models/container_registry.py` file? just a question!

src/ai/backend/common/docker.py
```diff
@@ -279,53 +272,63 @@ async def login(
...
-        return True
-    if known_registries is not None and val in known_registries:
-        return True
+def is_dockerhub_library_registry(
+    project: str,
+    known_registries: Optional[Mapping[str, Mapping[str, Any]] | Sequence[str]],
+) -> bool:
+    if project == "":
+        if isinstance(known_registries, list) and "index.docker.io" in known_registries:
+            return True
```
fregataa:
```suggestion
        if isinstance(known_registries, list):
            return "index.docker.io" in known_registries
```
let's make it similar with `dict` branch

src/ai/backend/common/docker.py
```diff
@@ -279,53 +272,63 @@ async def login(
...
+        if isinstance(known_registries, Mapping):
+            return "index.docker.io" in known_registries["library"]
```
fregataa: is it sure that "library" key is always in `known_registries` dictionary? if not, we should handle the case of not having "library" key

fregataa (commented):
Just a little suggestion about type hint in `is_known_registry()`. How about accepting only `dict` (or `Mapping`) type as `known_registries` parameter?
Please enlighten me how the `known_registries` looks like in list type and dict type

jopemachine:
> Just a little suggestion about type hint in `is_known_registry()`. How about accepting only `dict` (or `Mapping`) type as `known_registries` parameter? Please enlighten me how the `known_registries` looks like in list type and dict type

For example,

```
{'stable': {'cr.backend.ai': URL('https://cr.backend.ai')}, 'community': {'cr.backend.ai': URL('https://cr.backend.ai')}, 'multiarch': {'cr.backend.ai': URL('https://cr.backend.ai')}}
```

src/ai/backend/common/docker.py
jopemachine: No, key name could be different with `library`, so I think it should be 

```py
        if isinstance(known_registries, Mapping):
			return "index.docker.io" in known_registries.get("library", [])
```

Thanks!

src/ai/backend/manager/registry.py
jopemachine: These functions are only used in this file, so I defined them here.

If we move these functions, we need to update `test_handle_heartbeat`.

https://github.com/lablup/backend.ai/blob/e2ef6f399885055cbd43fa7a412e7088cdb3bd9f/tests/manager/test_registry.py#L18C11-L29C1

fregataa: then how about renaming the functions to specify that they get data about container registry?
it would be confusing because this file is `registry.py` and we cannot understand what `registry` means in those name at the first glance

fregataa (approved):
good!
in the future, we can apply more strict typing about container image data such as `ImageRef`, `ImageRegistry` or `ContainerRegistry` etc, and revamp codes

kyujin-cho:
I am postponing milestone to 24.09. We managed to meet our internal deadline for per-user image customization feature (#1973). From now on, at this PR, let's focus on forming concrete definitions and stabilize things rather than rushing for the merge approval.

src/ai/backend/manager/models/container_registry.py
```diff
@@ -0,0 +1,425 @@
...
+    is_global = graphene.Boolean(description="Added in 24.09.0.")
+    project = graphene.String(description="Added in 24.09.0.")
+    username = graphene.String(description="Added in 24.09.0.")
+    password = graphene.String(description="Added in 24.09.0.")
+    ssl_verify = graphene.Boolean(description="Added in 24.09.0.")
+
+
+class ContainerRegistry(graphene.ObjectType):
+    class Meta:
+        interfaces = (AsyncNode,)
```
fregataa:
```suggestion
    class Meta:
        interfaces = (AsyncNode,)
        description = "Added in 24.09.0."
```

src/ai/backend/manager/models/container_registry.py
```diff
@@ -0,0 +1,425 @@
...
+
+    id = graphene.String(
+        required=True,
+        description="Object id. Can be either global id or object id. Added in 24.09.0.",
+    )
+
+
+class ContainerRegistryConfig(graphene.ObjectType):
+    url = graphene.String(required=True, description="Added in 24.09.0.")
+    type = ContainerRegistryTypeField(required=True, description="Added in 24.09.0.")
```
fregataa: As gql inspector checked, `type` field changed from `string` to `ContainerRegistryTypeField`.
We can deprecate this field and add a new `ContainerRegistryTypeField` field.

src/ai/backend/manager/models/container_registry.py
```diff
@@ -0,0 +1,425 @@
...
+    registry_name = graphene.String(required=True, description="Added in 24.09.0.")
+    is_global = graphene.Boolean(description="Added in 24.09.0.")
+    project = graphene.String(description="Added in 24.09.0.")
```
fregataa: Same here.

src/ai/backend/manager/models/gql.py
```diff
@@ -2193,19 +2208,54 @@ async def resolve_quota_scope(
     async def resolve_container_registry(
         root: Any,
         info: graphene.ResolveInfo,
-        hostname: str,
+        id: graphene.UUID,
```
fregataa:
```suggestion
        id: graphene.UUID,
        hostname: str,
```
Should not remove field argument.
Let clients use the field and this resolver can simply ignore it.

jopemachine:
* **#1917** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/1917?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/ai/backend/client/func/resource.py
```diff
@@ -40,11 +40,11 @@ async def check_presets(cls, group: str | None = None, scaling_group: str | None
 
     @api_function
     @classmethod
-    async def get_docker_registries(cls):
```
fregataa, line 43: Let's leave the existing API function and add docstring like "this API function is deprecated".

src/ai/backend/manager/api/etcd.py
achimnol, line 1: Let's keep the API `/config/container-registries` available, by calling the new API handler inside it, while adding a deprecation warning log.

src/ai/backend/install/context.py
achimnol, line 1: `INSERT INTO container_registries ...` in `populate_images()` looks like a duplication of loading `example-container-registries-harbor.json` in `load_fixtures()`. I think it would be better to leave the fixture only and leave a small comment saying that the container registry configuration is included in the fixture.

achimnol (changes requested):

src/ai/backend/manager/container_registry/base.py
```diff
@@ -142,27 +131,44 @@ async def commit_rescan_result(self) -> None:
...
+                        ]).where(ContainerRegistryRow.registry_name == image_ref.registry)
+
+                        registries = (await session.execute(query)).fetchall()
+                        registry_cache[image_ref.registry] = registries
+                    else:
+                        registries = registry_cache[image_ref.registry]
+
+                    for registry in registries:
+                        if image_ref.name.startswith(registry.project):
+                            if not image_ref.name.split(registry.project)[1].startswith("/"):
```
jopemachine: The reason for writing this branch logic is to handle cases where, for example, `image_ref.name` is "namespace/project" and `registry.project` is "namespace/project2".

Of course, in this case, since the projects are different, this `registry_id` should not be selected.

src/ai/backend/manager/api/session.py
```diff
@@ -1074,33 +1075,43 @@ async def convert_session_to_image(
...
     registry_hostname = project.container_registry["registry"]
     registry_project = project.container_registry["project"]
-    registry_conf = await root_ctx.shared_config.get_container_registry(registry_hostname)
-    if not registry_conf:
-        raise InvalidAPIParameters(f"Registry {registry_hostname} not found")
-    if registry_project not in registry_conf.get("project", ""):
-        raise InvalidAPIParameters(f"Project {registry_project} not found")
 
-    base_image_ref = session.main_kernel.image_ref
+    async with root_ctx.db.begin_readonly() as db_session:
```
fregataa:
```suggestion
    async with root_ctx.db.begin_readonly_session() as db_session:
```
`begin_readonly()` yields an `AsyncConnection` object and it does work the same as an `AsyncSession` object.

src/ai/backend/manager/api/session.py
```diff
@@ -1074,33 +1075,43 @@ async def convert_session_to_image(
...
+            .options(
+                load_only(
+                    ContainerRegistryRow.url,
+                    ContainerRegistryRow.username,
+                    ContainerRegistryRow.password,
+                )
+            )
+        )
+
+        registry_conf = (await db_session.execute(query)).fetchone()
```
fregataa:
```suggestion
        registry_conf = cast(ContainerRegistryRow | None, await db_session.scalar(query))
```

src/ai/backend/manager/registry.py
```diff
@@ -204,6 +206,58 @@
...
+            if project not in result:
+                result[registry_row[0]] = {}
+
+            result[project][registry_name] = yarl.URL(url)
+
+        return result
+
+
+async def get_container_registry_info(
+    db: ExtendedAsyncSAEngine, registry_id: str
```
fregataa:
```suggestion
    db: ExtendedAsyncSAEngine, registry_id: uuid.UUID
```
id column of `ContainerRegistryRow` has UUID type

src/ai/backend/manager/registry.py
```diff
@@ -204,6 +206,58 @@
...
+
+        for registry_row in registries:
+            project, registry_name, url = registry_row
+
+            if project not in result:
+                result[registry_row[0]] = {}
+
+            result[project][registry_name] = yarl.URL(url)
+
+        return result
```
fregataa:
```suggestion
    async with db.begin_readonly_session() as db_session:
        
        query_stmt = (
          sa.select(ContainerRegistryRow)
          .options(load_only(
                    ContainerRegistryRow.project,
                    ContainerRegistryRow.registry_name,
                    ContainerRegistryRow.url
                    ))
        )
        registries = cast(list[ContainerRegistryRow], (await db_session.scalars(query_stmt).all())
        result: MutableMapping[str, MutableMapping[str, yarl.URL]] = {}

        for registry_row in registries:
            project = registry_row.project
            registry_name = registry_row.registry_name
            url = registry_row.url

            if project not in result:
                result[project] = {}

            result[project][registry_name] = yarl.URL(url)

        return result
```

src/ai/backend/manager/registry.py
```diff
@@ -204,6 +206,58 @@
...
+                ]).where(ContainerRegistryRow.id == registry_id)
+            )
+        ).fetchall()
+
+        if not results:
+            raise UnknownImageRegistry(registry_id)
+
+        result: tuple[str, Optional[str], Optional[str]] = results[0]
+
+        url, username, password = result
```
fregataa:
```suggestion
    async with db.begin_readonly_session() as db_session:
       query_stmt = (
       sa.select(ContainerRegistryRow)
       .where(ContainerRegistryRow.id == registry_id)
       .options(load_only(ContainerRegistryRow.url,
                    ContainerRegistryRow.username,
                    ContainerRegistryRow.password,))
       )
       registry_row = cast(ContainerRegistryRow | None, await db_session.scalar(query_stmt))

        if registry_row is None:
            raise UnknownImageRegistry(registry_id)
        url = registry_row.url
        username = registry_row.username
        password = registry_row.password
```

src/ai/backend/manager/registry.py
```diff
@@ -1381,9 +1435,11 @@ async def start_session(
                 image_infos[str(image_ref)] = resolved_image_info
                 if not resolved_image_info.image_ref.is_local:
                     is_local_image = False
-                    registry_url, registry_creds = await get_registry_info(
-                        self.shared_config.etcd, image_ref.registry
+
+                    registry_url, registry_creds = await get_container_registry_info(
+                        self.db, cast(GUID, resolved_image_info.registry_id)
```
fregataa:
```suggestion
                    registry_url, registry_creds = await get_container_registry_info(
                        session, cast(GUID, resolved_image_info.registry_id)
```
readonly session is already opened in this context. Opening a new DB session is redundant and also could raise an error.
Let's pass the opened DB session to `get_container_registry_info()` and update code inside the function.

src/ai/backend/manager/container_registry/base.py
```diff
@@ -130,33 +131,59 @@ async def commit_rescan_result(self) -> None:
...
-                        config_digest=v["config_digest"],
-                        size_bytes=v["size_bytes"],
-                        type=ImageType.COMPUTE,
-                        accelerators=v.get("accels"),
-                        labels=v["labels"],
-                        resources=v["resources"],
-                    )
-                    for k, v in _all_updates.items()
-                ])
+                registry_cache = {}
```
fregataa:
```suggestion
                registry_cache: dict[str, list[ContainerRegistryRow]] = {}
```

src/ai/backend/manager/container_registry/base.py
```diff
@@ -130,33 +131,59 @@ async def commit_rescan_result(self) -> None:
...
+
+                for image_ref, v in _all_updates.items():
+                    if image_ref.registry not in registry_cache:
+                        query = sa.select([
+                            ContainerRegistryRow.id,
+                            ContainerRegistryRow.project,
+                        ]).where(ContainerRegistryRow.registry_name == image_ref.registry)
```
fregataa:
```suggestion
                        query = sa.select(
                            ContainerRegistryRow
                        ).where(ContainerRegistryRow.registry_name == image_ref.registry).options(load_only(ContainerRegistryRow.id, ContainerRegistryRow.project))
```

src/ai/backend/manager/models/container_registry.py
```diff
@@ -0,0 +1,294 @@
...
+    DOCKER = "docker"
+    HARBOR = "harbor"
+    HARBOR2 = "harbor2"
+    LOCAL = "local"
+
+
+class ContainerRegistryRow(Base):
+    __tablename__ = "container_registries"
+    id = IDColumn()
+    url = sa.Column("url", sa.String(length=512), index=True, nullable=False)
```
fregataa, line 53: Just a comment. We can use `ai.backend.manager.models.base.URLColumn` for this column. But it is UnicodeText type and updating existing "str to yarl.URL" codes would be bothersome.. so, it is up to you.

src/ai/backend/manager/registry.py
fregataa: Just a comment. how about adding this function to `ContainerRegistryRow` as a classmethod?

src/ai/backend/manager/registry.py
```diff
@@ -204,6 +206,58 @@
...
+
+            if project not in result:
+                result[registry_row[0]] = {}
+
+            result[project][registry_name] = yarl.URL(url)
+
+        return result
+
+
+async def get_container_registry_info(
```
fregataa: We can use `ContainerRegistryRow.get()` method here. And this function also could be in `ContainerRegistryRow` class as a classmethod.

fregataa (changes requested):

PR #1918, opened 2024-02-19
ci: new test cases for vfolder CLI commands

### LINKED ISSUE #1729: Revamp End-to-End testing

SangwonYoon:
This PR resolves #1729 by adding new test cases for CLI commands below.

1. `vfolder upload`
2. `vfolder rename-file`
3. `vfolder download`
4. `vfolder mkdir`
5. `vfolder mv`


src/ai/backend/test/cli_integration/conftest.py
```diff
@@ -140,6 +141,16 @@ def keypair_resource_policy() -> str:
     return fake.unique.word()
 
 
+@pytest.fixture
+def make_txt_file() -> Generator[TextIOWrapper, None, None]:
```
rapsealk: It would be better to use the noun form rather than the verb form for this fixture.

src/ai/backend/test/cli_integration/conftest.py
```diff
@@ -140,6 +141,16 @@ def keypair_resource_policy() -> str:
...
+    FILE_NAME = "test.txt"
+    with open(FILE_NAME, "w") as f:
+        f.write("This file is for testing.")
+    yield f
+
+    os.remove(FILE_NAME)
```
rapsealk: Please use lowercases for local variables.
```suggestion
    filepath = "test.txt"
    with open(filepath, "w") as f:
        f.write("This file is for testing.")
    yield f

    os.remove(filepath)
```

src/ai/backend/test/cli_integration/user/test_vfolder.py
```diff
@@ -71,6 +73,123 @@ def test_rename_vfolder(run_user: ClientRunnerFunc):
...
+    Otherwise, it will raise an error.
+    """
+
+    VFOLDER_NAME = "test_folder2"
+    FILE_NAME = make_txt_file.name
+
+    # Upload the file to vfolder
+    with closing(run_user(["vfolder", "upload", VFOLDER_NAME, FILE_NAME])) as p:
+        p.expect(EOF)
+        assert "Done." in p.before.decode(), "File upload failed."
```
rapsealk, line 94: It is unsafe to check raw messages from the response. I suggest to use `--output=json` option and check `ok` field from the response. Please refer [this](https://github.com/lablup/backend.ai/blob/9df3058fd9981e54737a5905f6e12b83e133c35c/src/ai/backend/test/cli_integration/admin/test_keypair.py#L18-L40) as an example.

rapsealk (changes requested):

src/ai/backend/test/cli_integration/user/test_vfolder.py
SangwonYoon: I wanted to implement in this way, but in vfolder, even if I use the `--output=json` option, the output is not displayed in json. Is there other way to output in json format?

rapsealk: @SangwonYoon Sorry, you may leave this as-is now. OutputHandler and `upload()` are not ready for this yet and need to be improved in further PRs. Thank you!

src/ai/backend/test/cli_integration/user/test_vfolder.py
```diff
@@ -71,6 +73,123 @@ def test_rename_vfolder(run_user: ClientRunnerFunc):
     assert bool(test_folder3), "Test folder 3 doesn't exist!"
 
 
+def test_upload_file(run_user: ClientRunnerFunc, txt_file: TextIOWrapper):
```
rapsealk, line 81: If execution order is crucial, you may use `pytest-dependency`. Please refer the code below or the [docs](https://pytest-dependency.readthedocs.io/en/stable/usage.html).
https://github.com/lablup/backend.ai/blob/9df3058fd9981e54737a5905f6e12b83e133c35c/tests/manager/models/test_etcd.py#L95-L97

rapsealk (changes requested):

PR #1922, opened 2024-02-21
feat: Add the non-intercative mode to the TUI installer

### LINKED ISSUE #1920: Add the non-interactive mode to the TUI installer
- Use the mode option as given instead of highlighting it as the default menu.
- Make the target path option applied if specified, instead of taking the user input (the default `~/backendai`)

achimnol:
resolves #1920

It adds the `--non-interactive` flag that:
- Auto-selects the mode specified via the cli option
- Skips the input dialog to get an alternative target-path when it already exists
- Auto-exits when the installation process is done or has failed by simulating the "q" key press


achimnol:
We'll open a follow-up issue if there are other hang-up issues in the headless (non-tty) mode.

PR #1924, opened 2024-02-22
refactor: Enhance typing of some modules

achimnol:
* refactor: Enhance typing of the `ai.backend.manager.models.base` module
* refactor: Make typing/arg-names consistent in the `ai.backend.storage` package
  - `recursive` is now an explicitly keyword-only argument.
  - The default value of `recursive` is now consistent between the concrete implementation classes and the abstract base class.
  - `options: QuotaConfig` is now consistently `config: QuotaConfig`.
  - `target_path` arguments in the concrete classes are renamed to `path` to silence pyright warning about mismatch with argument names in the base classes.
* refactor: Use `enum.{IntEnum,StrEnum}` when possible
  - https://docs.python.org/3/library/enum.html#enum.IntEnum
  - https://docs.python.org/3/library/enum.html#enum.StrEnum
  - They can be serialized via `json.dumps()` as their values without any extra conversion.
  - In Click command handlers:  
    `click.Choice([*LogSeverity.__members__.keys()], case_sensitive=False)`  
    → `click.Choice([*LogSeverity], case_sensitive=False)`
    - Click preserves the enum type when processing case-insensitive match.
  - In alembic migrations:  
    `list(map(lambda v: v.value, VFolderPermission))`  
    → `list(map(str, VFolderPermission))`

### DIFF
--- a/src/ai/backend/accelerator/mock/defs.py
+++ b/src/ai/backend/accelerator/mock/defs.py
@@ -1,6 +1,6 @@
 import enum
 
 
-class AllocationModes(str, enum.Enum):
+class AllocationModes(enum.StrEnum):
     DISCRETE = "discrete"
     FRACTIONAL = "fractional"
--- a/src/ai/backend/agent/server.py
+++ b/src/ai/backend/agent/server.py
@@ -868,15 +868,15 @@ async def server_main(
 )
 @click.option(
     "--log-level",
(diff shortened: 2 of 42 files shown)


achimnol:
NOTE: This PR updates some existing migration codes, so `require:db-migration` is a false-positive.

PR #1934, opened 2024-02-27
fix: CLI `agent info` related bugs

fregataa:
related to #1793 

* Replace `HardwareMetadata` to `dict` when class check because `TypedDict` does not support such functionality
* Add a missing parameter to default metric value formatter in the Client SDK's output framework.
  - Also make it detectable via static type checks by improving type annotations.
* Refine and clarify the description of `MetricTypes`, with fixes of the `cuda_open` and `mock` plugin metric types.
  - Rename `MetricTypes.USAGE` to `MetricTypes.GAUGE` to distinguish better from `MetricTypes.ACCUMULATION`.
  - This renaming does not affect existing compute plugins:
    - `MetricTypes.USAGE` is preserved for backward compatibility and evaluated as the same to `MetricTypes.GAUGE`.
    - `metric.type` attribute does not change the processing/display of metric values at all. Frontend formatters use `unit_hint` as the primary key.

### DIFF
--- a/changes/1934.fix.md
+++ b/changes/1934.fix.md
@@ -1,0 +1,1 @@
+Fix CLI `agent info` related issues by replacing `HardwareMetadata` to `dict` when class check and adding parameter to default metric value formatter.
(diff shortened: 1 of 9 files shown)


achimnol (approved):
Thanks for the initial fix.
I've expanded your changes to make the original bug detectable via static type checks and improve clarity of related codes.

PR #1935, opened 2024-02-27
fix: unreasonable `ondelete` restrictions of certain foreign keys

fregataa:
- Change `endpoints.model` to nullable and set ondelete="NULL"
- Set `endpoint_tokens.endpoint` ondelete="NULL"


src/ai/backend/manager/models/endpoint.py
```diff
@@ -82,7 +82,10 @@ class EndpointRow(Base):
         "image", GUID, sa.ForeignKey("images.id", ondelete="RESTRICT"), nullable=False
     )
     model = sa.Column(
-        "model", GUID, sa.ForeignKey("vfolders.id", ondelete="RESTRICT"), nullable=False
+        "model",
+        GUID,
+        sa.ForeignKey("vfolders.id", ondelete="SET NULL"),
+        nullable=True,
```
kyujin-cho, line 88: Unless updating API implementation to aware of nullable `model` column, leaving model service which depends on now removed VFolders alive will cause lots of query errors. Please either update on delete rule to `CASCADE` or update API implementations to skip querying model services without `model` column set.

kyujin-cho (changes requested):

PR #1941, opened 2024-02-29
fix: modify the code logic to check for the existence of min/max in `resource_limits`.

agatha197:
Modify the code logic to check for the existence of min/max in the `resource_limits` of the `ModifyImage` mutation.
From this PR, the admin can modify the image metadata without the below error.

Original error log:
```shell
ADMIN.GQL Exception: {'message': "object of type 'UndefinedType' has no len()", 'locations': [{'line': 6, 'column': 5}], 'path': ['modify_image']}
2024-02-29 14:05:15.598 DEBUG ai.backend.manager.api.admin [49674] Traceback (most recent call last):
  File "/backend.ai/dist/export/python/virtualenvs/python-default/3.11.6/lib/python3.11/site-packages/graphql/execution/execute.py", line 528, in await_result
    return_type, field_nodes, info, path, await result
                                          ^^^^^^^^^^^^
  File "/backend.ai/src/ai/backend/manager/models/image.py", line 1007, in mutate
    if limit_option.max is not None and len(limit_option.max) > 0:
                                        ^^^^^^^^^^^^^^^^^^^^^
graphql.error.graphql_error.GraphQLError: object of type 'UndefinedType' has no len()

GraphQL request:6:5
5 | ) {
6 |     modify_image(
  |     ^
7 |         target: $target,
```

- [ ] Test case(s) to:
- [ ] Documentation


fregataa (commented):
How about checking `Undefined` of `limit_option`?

PR #1962, opened 2024-03-22
fix: Disable krunner's user site to avoid potential conflicts with user-installed packages

### LINKED ISSUE #1959: Disable user site packages (`~/.local`) for specific in-container Python components
When users install custom Python packages into the user-site directory (`~/.local`) which is provided as an auto-mounted docker volume in containers, sometimes they break the dependency of Jupyter making it to crash when starting the app.

We could minimize the impact of dependency breakage due to user-installed packages as follows:
image

However, it is not possible to 100% prevent such problems because we cannot disable user-site directories for specific packages for a single Python process running both ipykernel and user programs.

> [!NOTE]
> _Some futuristic idea:_ We could make ipykernel to use subinterpreters in Python 3.12+ to isolate the package import namespaces (with different user-site settings) within a single process.

Still, I think it would be better to explicitly disable user-sites for the designated components to minimize the problem's surface area.

achimnol:
resolves #1959

This PR disables the user-site functionality for the krunner python commands to improve security and potential conflicts with the user-installed packages that may override some packages in the krunner.

Though, it does not disable it for the user-executed service-port apps that use the container's own Python runtime such as Jupyter notebooks (both the frontend server and the kernel processes). This is why I've put `-s` options individually instead of setting the `PYTHONNOUSERSITE=1` environment variable which may propagate to all sub-processes.

### DIFF
--- a/changes/1962.fix.md
+++ b/changes/1962.fix.md
@@ -1,0 +1,1 @@
+Explicitly disable the user-site package detection in the krunner python commands to avoid potential conflicts with user-installed packages in `.local` directories
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -1920,6 +1920,7 @@ class AbstractAgent(
                     krunner_opts.append("--debug")
                 cmdargs += [
                     "/opt/backend.ai/bin/python",
+                    "-s",
                     "-m",
                     "ai.backend.kernel",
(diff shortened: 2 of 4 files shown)


achimnol:
Tested the behavior as follows.

Without the "-s" flag (user-site enabled):
image

With the "-s" flag (user-site disabled):
image

PR #1970, opened 2024-03-27
chore: Wrap `callosum.AuthenticationError` to custom error

fregataa:
`callosum.AuthenticationError` is raised when manager's RPC peer object fails to initiate zmq connection due to several reasons like...
- Ping timeout
- Unable to access agent address
- etc

Let's wrap `callosum.AuthenticationError` to our custom error so that we don't misunderstand that it is an **"authentication error"**


src/ai/backend/manager/exceptions.py
```diff
@@ -25,6 +25,26 @@ class InvalidArgument(Exception):
     pass
 
 
+class RPCError(RuntimeError):
+    """
+    An exception class to represent any error caused in RPC functions.
+    """
+
```
kyujin-cho, line 32: The idea to wrap AuthenticationError with a more meaningful `RPCError` looks great. However this error should contain more information about the cause (think of the facts you've mentioned at the PR description!). Please update the implementation to contain more details.

kyujin-cho (changes requested):

src/ai/backend/manager/exceptions.py
```diff
@@ -25,6 +25,29 @@ class InvalidArgument(Exception):
...
+    def __init__(
+        self,
+        agent_id: AgentId,
+        agent_addr: str,
+        exc_repr: str,
+    ) -> None:
+        super().__init__(agent_id, agent_addr, exc_repr)
+        self.agent_id = agent_id
+        self.agent_addr = agent_addr
+        self.exc_repr = exc_repr
```
kyujin-cho: To match convention I think this variable is better to be named as `extra_msg`.
```suggestion
        self.extra_msg = extra_msg
```

kyujin-cho (changes requested):

kyujin-cho (approved):

PR #1973, opened 2024-03-28
feat: Per-user container image creation from running sessions

kyujin-cho:
This PR introduces new API ` POST /session/{session_id}/image`, which commits ongoing session and pushes artifact to designated registry.    
To make this feature work, administrator first has to update project setting to fill out `per_user_image_storage_registry` column. `per_user_image_storage_registry` should contain a JSON string formed like `{"registry": "<registry name>", "project": "<registry project name>"}`, where `<registry name>` refers registry definition (`config/docker/registries/<registry name>`) and `<registry project name>` for registry project defined under `config/docker/registries/<registry name>/projects`.


src/ai/backend/manager/registry.py
```diff
@@ -457,6 +458,10 @@ async def create_session(
                     ],
                 )
             requested_image_ref = image_row.image_ref
+            if (
+                _owner_uuid := image_row.labels.get("ai.backend.image-owner")
+            ) and _owner_uuid != str(user_scope.user_uuid):
```
achimnol: We plan to expand this feature to support per-project custom images.
Let's use quota-scope-like prefixes to distinguish whether the given owner ID is a user or a project.

src/ai/backend/manager/registry.py
```diff
@@ -3162,9 +3167,56 @@ async def get_commit_status(
...
+        extra_labels: dict[str, str] = {},
+    ) -> Mapping[str, Any]:
+        """
+        Commit a main kernel's container of the given session.
+        """
+
+        kernel: KernelRow = session.main_kernel
+        if kernel.status != KernelStatus.RUNNING:
+            raise InvalidAPIParameters(
+                f"Unable to commit since kernel(id: {kernel.id}) of session(id: {session.id}) is"
```
achimnol: Though this is not a strictly documented convention, I'd suggest the following to keep consistency with existing log formats:
```suggestion
                f"Unable to commit since the kernel k:{kernel.id} (of s:{session.id}) is"
```

src/ai/backend/agent/server.py
```diff
@@ -608,20 +610,52 @@ async def commit(
...
+    async def push_image(
+        self,
+        canonical: str,
+        architecture: str,
+        registry_conf: ImageRegistry,
+        *,
+        is_local: bool = False,
+    ) -> dict[str, Any]:
+        log.info("rpc::push_image(c:{})", canonical)
+        bgtask_mgr = self.local_config["background_task_manager"]
```
achimnol: This shouldn't be at `local_config`.
Put it in the `AbstractAgent` or similar things.

src/ai/backend/agent/server.py
```diff
@@ -608,20 +610,52 @@ async def commit(
...
         kernel_id: str,
         subdir: str,
-        filename: str,
+        *,
+        canonical: str | None = None,
+        filename: str | None = None,
+        extra_labels: dict[str, str] = {},
     ) -> dict[str, Any]:
         log.info("rpc::commit(k:{})", kernel_id)
         bgtask_mgr = self.local_config["background_task_manager"]
```
achimnol: Same here.

achimnol (changes requested):
Initial batch of review!

PR #1985, opened 2024-04-01
fix: inconsistency of `kernels.session_name` column

fregataa:
resolves https://github.com/lablup/giftbox/issues/646

- update session rename API to update session's sibling kernels.session_name
- sync mismatch between `kernels.session_name` with `sessions.name`


kyujin-cho (approved):

kyujin-cho:
Although I did reckon the DB migration script from this PR does not involve any kind of DB schema alteration, since we did agree to freeze whole datbase migration tree from prior major release, this PR should not be backported to 23.09.

PR #1988, opened 2024-04-01
fix: Revert vfolder APIs response to raw `web.Response`

fregataa:


kyujin-cho (approved):
For now this PR will be approved, but please keep in mind of of allowing setting arbitrary response headers along with using pydantic as response schema validator.

PR #1989, opened 2024-04-02
feat: Clear `error_logs` records in the `clear-history` command

### LINKED ISSUE #1974: Clear the old `error_logs` records in the `mgr clear-history` command
https://github.com/lablup/backend.ai/blob/1c34a01ecac777cf2aba08519d74b7a9e874566c/src/ai/backend/manager/cli/__main__.py#L222
This function should also clean up the old records of the `error_logs` table.

jopemachine:

- [ ] Test case(s) to:
- [ ] Documentation

Fixed #1974.


src/ai/backend/manager/cli/__main__.py
```diff
@@ -305,10 +306,6 @@ async def _clear_terminated_sessions():
                 )
                 deleted_count = result.rowcount
 
-                vacuum_sql = "VACUUM FULL" if vacuum_full else "VACUUM"
-                log.info(f"Perfoming {vacuum_sql} operation...")
-                await conn.exec_driver_sql(vacuum_sql)
-
```
kyujin-cho, line 311: Could you tell me why this block has been lifted?

kyujin-cho (changes requested):

src/ai/backend/manager/cli/__main__.py
jopemachine: I moved the logic to the bottom function (`_clear_old_error_logs`) because I find it easier to identify the vacuum-related logic at the end rather than somewhere in the middle.

kyujin-cho: @jopemachine  Since `VACCUM` operation itself does not depend on neither session clear nor error log clearing job, I think it is better to separate it as an independent job package rather than appending it to prior jobs.

kyujin-cho (approved):

PR #1991, opened 2024-04-02
feat: Enable robust DB connection handling

fregataa:
Considering situations such as DB HA settings, there may be many situations where DB processes are restarted. Optimistic connection pool handling(pool-recycle, #1877) does not fully cope with these frequent DB restarts, so we introduce pessimistic DB connection pool handling, `pool-pre-ping`.

`pool-pre-ping` setting eliminates DB error due to stale pooled connections by ping at the start of each connection pool checkout.
It adds a small overhead to the connection checkout process.

ref: https://docs.sqlalchemy.org/en/14/core/pooling.html#disconnect-handling-pessimistic

- [ ] Test case(s) to:

### DIFF
--- a/configs/manager/halfstack.toml
+++ b/configs/manager/halfstack.toml
@@ -12,7 +12,7 @@ name = "backend"
 user = "postgres"
 password = "develove"
 pool-recycle = 50
-pool-pre-ping = true
+pool-pre-ping = false
 
 
 [manager]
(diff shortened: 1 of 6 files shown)


configs/manager/halfstack.toml
```diff
@@ -12,6 +12,7 @@ name = "backend"
 user = "postgres"
 password = "develove"
 pool-recycle = 50
+pool-pre-ping = true
```
achimnol: Since the default value is false, I think this should be also false.
Though, we should recommend setting this option for HA setups.

achimnol (approved):

PR #1993, opened 2024-04-03
chore: fix incorrect version notation of GQL Field

fregataa:
blocked by #2000 

# Change
- Correct Added/Deprecated version of Resource policy GQL schema fields
- Unify description/deprecation message to `Added in {VERSION}` and `Deprecated since {VERSION}` for consistency.
- Unify version notation to end with `.(dot)`. e.g. "23.09.10" -> "23.09.10."
- Fix fields labeled as added in version "24.03.1" that were actually added in version "24.03.0"

# Why are the versions so different?
### KeyPairResourcePolicy
- `max_quota_scope_size` field was added by mistake in version "23.09.6", when #1677 was backported (https://github.com/lablup/backend.ai/commit/a4bad32e09c131833262d0ad59bac199d0084ebe)

### UserResourcePolicy, ProjectResourcePolicy
- `max_vfolder_size` was actually deprecated since "23.09.2", when #1644 was backported (https://github.com/lablup/backend.ai/commit/2a1cc67790d6449b9e3b17540b33a87d59b03cc4)

### {Create/Modify}UserResourcePolicyInput, ProjectResourcePolicyInput
- Restore `max_vfolder_size` field that was removed when #1644 was backported to "23.09.2" (https://github.com/lablup/backend.ai/commit/2a1cc67790d6449b9e3b17540b33a87d59b03cc4)

# Warning
in 23.09, `max_vfolder_count` field of {User/Project}ResourcePolicy was removed in "23.09.3." by mistake and restored in "23.09.6.". So, the version the field was added in is "23.09.6." 
- The commit that removed `max_vfolder_count` from {User/Project}ResourcePolicy: https://github.com/lablup/backend.ai/commit/f4d6b51b3b4a833820a5eeae5c7ad26817dcfc26
- The commit that restored the field: https://github.com/lablup/backend.ai/commit/3d9a886d7352b80afd9bfb03f11706bf7c62c455

Some fields has a different version in each major version. e.g. {User/Project}ResourcePolicy has `max_vfolder_count` field and it was added in "24.03.1" and "23.09.6".

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version
- [ ] API server-client counterparts (e.g., manager API -> client SDK)
- [ ] Documentation
  - Contents in the `docs` directory
  - docstrings in public interfaces and type annotations

----
📚 Documentation preview 📚: https://sorna--1993.org.readthedocs.build/en/1993/

----
[...]

PR #2000, opened 2024-04-05
ci: Modified to operate even in deprecation reason changed situations

Yaminyam:
- https://github.com/lablup/backend.ai/actions/runs/8564982449?pr=1993
'||' The part that could not handle undefined because it operates as an operator is '??' Modify to use operators.
- Upgrade stefanzweifel/git-auto-commit-action@v4 using node 16 version to v5 version.

- [ ] Test case(s) to:
- [ ] Documentation


fregataa (commented):
Can we change the regex pattern that check field description?
we decided version notation formats as "Added in xx.xx.x." and "Deprecated since xx.xx.x" in #1993

PR #2005, opened 2024-04-07
fix: pre-commit hook raising false alarm

kyujin-cho:
Follow-up PR of #1998. This PR explicitly marks `alembic` as a third party package on ruff's isort config to avoid ruff from think that the package pretending to be a first-party one. Those confusion is caused by our `ai.backend.manager.models.alembic` package having identical name with the third-party `alembic`.


inureyes:
LGTM

achimnol:
Maybe related to pantsbuild/pants#18410.

PR #2006, opened 2024-04-08
ci: Upgrade Pants to 2.21.0.dev5 and remove no longer used ruff_preview plugin

achimnol:
* Remove `tools/pants-plugins/ruff_preview` plugin as Pants 2.21 comes with the intrinsic one.
* Upgrade Pantsbuild from 2.21.0.dev4 to 2.21.0.dev5 ([Release note](https://github.com/pantsbuild/pants/releases/tag/release_2.21.0.dev5))

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version

PR #2008, opened 2024-04-08
hotfix: revert PR #2006

kyujin-cho:
This PR reverts commit 7f63e82e3b8b2fe59a2f7526a750f9b6afae7988, as it has just been discovered that latest pants (2.21.0.dev5) release being shipped with broken dependency packed (ref: https://github.com/pantsbuild/pants/issues/20765). This commit will be restored right after upstream issue is fixed and ready to be shipped.


kyujin-cho:
Skipping backport since originated PR itself is also yet applied to target branch (24.03).

achimnol (approved):

PR #2023, opened 2024-04-16
chore: migrate `subprocess.call()` usages to `.run()`

mirageoasis:
[link](https://docs.python.org/3/library/subprocess.html#older-high-level-api)

it says `subprocess.call` is order version of `subprocess.run` 

so I changed `subprocess.call` to `subprocess.run`

since changed code does not use return value of I think it is find to change functions


achimnol (approved):
Good catch on the modern recommended API use.

PR #2026, opened 2024-04-16
ci: Modify artifact name to avoid duplication

Yaminyam:
https://github.com/lablup/backend.ai/actions/runs/8709104568/job/23888120282
Starting from artifact@v4, artifacts with duplicate names are not allowed.
So, we add the OS name of each matrix running in parallel to make it unique.

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -292,7 +292,7 @@ jobs:
     - name: Upload local-proxy binaries
       uses: actions/upload-artifact@v4
       with:
-        name: local-proxy
+        name: ${{ matrix.os }}-local-proxy
         path: dist-local-proxy/*
     - name: Bootstrap Pants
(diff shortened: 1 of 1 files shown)


.github/workflows/default.yml
```diff
@@ -311,12 +311,12 @@ jobs:
...
       uses: actions/upload-artifact@v4
       with:
-        name: scies-${{ matrix.os }}
+        name: ${{ matrix.os }}-scies
         path: dist/*
     - name: Upload pants log
       uses: actions/upload-artifact@v4
       with:
-        name: pants.deploy.log
+        name: ${{ matirx.os }}-pants.deploy.log
```
kyujin-cho: There's a typo here
```suggestion
        name: ${{ matrix.os }}-pants.deploy.log
```

kyujin-cho (changes requested):

PR #2030, opened 2024-04-17
fix: Set keypairs.rate_limit to not nullable

fregataa:
Currently `keypairs.rate_limit` column is nullable and it causes error when a manager middleware checks request rate limit.
- The middleware should do null-check
- Or, the column should not be nullable

I choose the second option. Please give me any suggestion

- [ ] Test case(s) to:
- [ ] Documentation

----
📚 Documentation preview 📚: https://sorna--2030.org.readthedocs.build/en/2030/

----
📚 Documentation preview 📚: https://sorna-ko--2030.org.readthedocs.build/ko/2030/


achimnol (approved):

achimnol (changes requested):
Need to check why the `X-RateLimit-{Limit,Remaining}` header values are divided by 10.

src/ai/backend/manager/models/keypair.py
```diff
@@ -79,7 +79,13 @@
         onupdate=sa.func.current_timestamp(),
     ),
     sa.Column("last_used", sa.DateTime(timezone=True), nullable=True),
-    sa.Column("rate_limit", sa.Integer),
+    sa.Column(
+        "rate_limit",
+        sa.Integer,
+        nullable=False,
+        default=DEFAULT_KEYPAIR_RATE_LIMIT,
+        server_default=str(DEFAULT_KEYPAIR_RATE_LIMIT),
```
achimnol: Shouldn't we remove `str()` here as the database field is an integer and the original constant is also an int?

achimnol (changes requested):

src/ai/backend/manager/models/keypair.py
fregataa: `server_default=DEFAULT_KEYPAIR_RATE_LIMIT,` will raise error when we run migration. The error log is like below.
```
sqlalchemy.exc.ArgumentError: Argument 'arg' is expected to be one of type '<class 'str'>' or '<class 'sqlalchemy.sql.elements.ClauseElement'>' or '<class 'sqlalchemy.sql.elements.TextClause'>', got '<class 'int'>'
```
And as I tested, the stringified server default value is injected as integer.
image

achimnol: Ah, so it expects a string to be interpolated into the actual SQL statement.
I think it would be better to use `sa.text()` or similar stuffs to reveal that this is an SQL-injected text.

agatha197 (commented):
When I created a user whose `rate_limit` is null, I got this error as I expected.

But the user is created successfully.

Is it right to operate like the above?

src/ai/backend/manager/models/keypair.py
fregataa: `sa.text(DEFAULT_KEYPAIR_RATE_LIMIT)` raises error like below
```
TypeError: expected string or bytes-like object, got 'int'
```
so I just wrap it like this `sa.text(str(DEFAULT_KEYPAIR_RATE_LIMIT))`

src/ai/backend/manager/models/alembic/versions/a0b457d41018_set_keypairs_rate_limit_to_not_nullable.py
```diff
@@ -0,0 +1,44 @@
...
+def upgrade():
+    conn = op.get_bind()
+    sync_stmt = textwrap.dedent(
+        f"""
+        UPDATE keypairs
+        SET rate_limit = {default_rate_limit}
+        WHERE rate_limit is NULL;
+        """
+    )
+    conn.execute(text(sync_stmt))
```
achimnol, line 39: Although `server_default` expects a SQL string, this statement could be safely parametrized like:
```suggestion
    sync_stmt = textwrap.dedent(
        f"""
        UPDATE keypairs
        SET rate_limit = :rate_limit
        WHERE rate_limit is NULL;
        """
    )
    conn.execute(text(sync_stmt).bindparams(rate_limit=default_rate_limit))
```
See https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.TextClause.bindparams

achimnol (changes requested):

fregataa:

PR #2036, opened 2024-04-18
feature: Prevent deleting mounted folders

### LINKED ISSUE #776: Inhibit mounted vfolder deletion

macOS (Apple Silicon)

### Backend.AI version

21.03

### Describe the bug

For now, there's no explicit error that comes out when deleting virtual folder that is mounted to one or more sessions, which would bring a serious security hole. 

### To Reproduce

1. Create a virtual folder
2. Create a session with the folder created in step 1 mounted
3. Delete the folder created in step 1
4. Try to access the folder inside the session created in step 2 (with ttyd, etc.)
5. See the error.

### Expected Behavior

When virtual folder is mounted to one or more sessions, then deleting the mounted folder shall not be allowed.

### Anything else?

It's been handled in former repo (https://github.com/lablup/backend.ai-manager/pull/417), but It was on stale status for a while, and now It needs to be resolved.

fregataa:
resolves #776 

- The mount lookup happens only when we **"delete"** vfolders, not when we **"delete-forever"** them since it is already disallowed to delete-forever the READY vfolders
- Apply [GIN index](https://www.postgresql.org/docs/current/datatype-json.html#JSON-INDEXING) to `sessions.vfolder_mounts` column

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/models/base.py
+++ b/src/ai/backend/manager/models/base.py
@@ -417,6 +417,9 @@ class StructuredJSONObjectListColumn(TypeDecorator):
         super().__init__()
         self._schema = schema
 
+    def coerce_compared_value(self, op, value):
+        return JSONB()
+
     def process_bind_param(self, value, dialect):
         return [self._schema.to_json(item) for item in value]
 
--- a/src/ai/backend/manager/models/vfolder.py
+++ b/src/ai/backend/manager/models/vfolder.py
@@ -6,7 +6,7 @@ import os.path
 import uuid
 from datetime import datetime
 from pathlib import PurePosixPath
-from typing import TYPE_CHECKING, Any, Final, List, Mapping, NamedTuple, Optional, Sequence
+from typing import TYPE_CHECKING, Any, Final, List, Mapping, NamedTuple, Optional, Sequence, cast
 
 import aiohttp
 import aiotools
@@ -23,7 +23,7 @@ from sqlalchemy.dialects import postgresql as pgsql
 from sqlalchemy.engine.row import Row
 from sqlalchemy.ext.asyncio import AsyncConnection as SAConnection
 from sqlalchemy.ext.asyncio import AsyncSession as SASession
-from sqlalchemy.orm import selectinload
+from sqlalchemy.orm import load_only, selectinload
 
 from ai.backend.common.bgtask import ProgressReporter
 from ai.backend.common.config import model_definition_iv
@@ -32,6 +32,7 @@ from ai.backend.common.types import (
     MountPermission,
     QuotaScopeID,
     QuotaScopeType,
+    SessionId,
     VFolderHostPermission,
     VFolderHostPermissionMap,
     VFolderID,
@@ -72,6 +73,7 @@ from .gql_relay import AsyncNode, Connection, ConnectionResolverResult
 from .group import GroupRow, ProjectType
 from .minilang.ordering import OrderSpecItem, QueryOrderParser
 from .minilang.queryfilter import FieldSpecItem, QueryFilterParser, enum_field_getter
+from .session import DEAD_SESSION_STATUSES, SessionRow
 from .user import UserRole
 from .utils import ExtendedAsyncSAEngine, execute_with_retry, sql_json_merge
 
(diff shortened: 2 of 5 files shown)


src/ai/backend/manager/models/vfolder.py
```diff
@@ -1213,6 +1234,63 @@ async def ensure_quota_scope_accessible_by_user(
...
+        def from_json(cls, obj: Mapping[str, Any]) -> SimpleVFolderMount:
+            return cls(**cls.as_trafaret().check(obj))
+
+        @classmethod
+        def as_trafaret(cls) -> t.Trafaret:
+            return t.Dict({
+                t.Key("vfid"): tx.VFolderID,
+            })
+
+    class SimpleSessionRow(Base):
```
achimnol: Is this kind of specialization necessary to "minimize the queried columns"?
Could you compare the actual stringification results of the SQLAlchemy query object when using this class or the original class?

If you have added this inner class due to circular import issues, you could just import the session model classes inside the function.

achimnol (changes requested):

src/ai/backend/manager/models/vfolder.py
fregataa: I left comments on each new class.
It is because of the query condition `vfolder_mounts.contains([VFolderMount(vfid=vfolder_id)])`.
`VFolderMount` requires all name, vfid, vfsubpath, host_path etc fields to initialize and it is impossible to query with a condition like this `vfolder_mounts.contains([{"vfid": vfolder_id)])`.
So I just made a simplified type and an ORM class for this query.

This is the original query that I need.
```sql
SELECT id
FROM sessions
WHERE vfolder_mounts @> '[{"vfid": "vfolder_id"}]';
```

achimnol: So, the problem is that we cannot specify a partially initialized `VFolderMount` object for a SELECT context. (When INSERT-ing, we need all fields and our `JSONSerializableMixin` expects this.)

Could we find a way to detect the query context when `process_bind_param()` is called and consider introduction of a new optional flag `allow_partial=True|False` to `JSONSerializableMixin.to_json()`?

achimnol: As you've researched, it seems that it's not trivial to detect the query type inside type decorators.
Let's conclude this design as-is: adding a partial declaration of `SessionRow` as `SimpleSessionRow` to allow raw JSONB column queries.

achimnol:
We'll revisit the removal of `SimpleSessionRow` once we get some answers or ideas from https://github.com/sqlalchemy/sqlalchemy/discussions/11369.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -1213,6 +1230,48 @@ async def ensure_quota_scope_accessible_by_user(
     raise InvalidAPIParameters
 
 
+async def get_sessions_by_mounted_folder(
+    db_session: SASession, vfolder_id: VFolderID
+) -> tuple[uuid.UUID]:
```
achimnol: How about returning `common.types.SessionID`?
When possible, it is better to use meaningful types.

achimnol (changes requested):

src/ai/backend/manager/models/base.py
```diff
@@ -417,6 +417,9 @@ def __init__(self, schema: Type[JSONSerializableMixin]) -> None:
         super().__init__()
         self._schema = schema
 
+    def coerce_compared_value(self, op, value):
+        return JSONB()
+
```
kyujin-cho, line 422: What is the purpose of this interface?

kyujin-cho (changes requested):

src/ai/backend/manager/models/base.py
fregataa: Here is the full context https://github.com/sqlalchemy/sqlalchemy/discussions/11369
In short, this overridden method allows us to set a query condition on `StructuredJSONObjectListColumn` type column with simple JSON value.
```python
# If we don't override `coerce_compared_value()`, this will raise error
vfolder_mount_value = [{"vfid": "VFOLDER_ID"}]
query = sa.select(SessionRow).where(SessionRow.vfolder_mounts.contains(vfolder_mount_value))
```

PR #2041, opened 2024-04-22
feat: bespoke redis client and event dispatcher

kyujin-cho:
Let's hope this can resolve our long-lasting redis pool bug...


src/ai/backend/common/redis_client.py
```diff
@@ -0,0 +1,267 @@
...
+            raise RuntimeError("Redis with sentinel not supported for this library")
+
+        redis_url = self.redis_config.get("addr")
+        assert redis_url is not None
+
+        host = str(redis_url[0])
+        port = redis_url[1]
+        password = self.redis_config.get("password")
+
+        async with async_timeout.timeout(self.socket_connect_timeout):
```
fregataa: How about using `asyncio.timeout()` here?
https://docs.python.org/ko/3/library/asyncio-task.html#asyncio.timeout

src/ai/backend/common/redis_client.py
```diff
@@ -0,0 +1,267 @@
...
+            ["SELECT", self.db], reconnect_poll_interval=self.reconnect_poll_timeout
+        )
+
+        return client
+
+    async def __aenter__(self) -> RedisClient:
+        return await self.connect()
+
+    async def disconnect(self) -> None:
+        self.writer.close()
```
fregataa: `self.writer` should be StreamWriter object here but how about assigning `None` as default to `self.writer` and doing null-check here?

fregataa (commented):
As @adrysn mentioned, let's test it with read-only replicas

PR #2050, opened 2024-04-23
feat: Add GraphQL field `rolling_count` to check current rate limit without increasing

### LINKED ISSUE #2039: Include the rolling count of the keypair rate limit counter in the keypair GQL query
Currently, the number of API queries of a keypair displays the total accumulated number of API queries tracked via Redis.
For rate-limiting, we are using a sliding-window mechanism to track the number of API queries for the last 15 minutes, so to check the rate limits, it would be better to know this number of queries within the current sliding window.

It can be simply calculated using `ZCARD {access_key}` Redis command against the rate-limit database. Let's add this 'rolling count' field to the GraphQL query schema of KeyPair and display it in the Control Panel in addition to the accumulated query count.
```[tasklist]
### Tasks
- [ ] https://github.com/lablup/backend.ai/pull/2050
- [ ] https://github.com/lablup/backend.ai-control-panel/pull/770
```

rapsealk:
This PR resolves #2039 by adding a new field `rolling_count` to KeyPair. This field will only execute `ZCARD` redis command without increasing any counter.

----
📚 Documentation preview 📚: https://sorna--2050.org.readthedocs.build/en/2050/

----
📚 Documentation preview 📚: https://sorna-ko--2050.org.readthedocs.build/ko/2050/


src/ai/backend/manager/models/keypair.py
```diff
@@ -60,6 +64,15 @@
...
 
+_rolling_count_script = """
+local access_key = KEYS[1]
+local now = tonumber(ARGV[1])
+local window = tonumber(ARGV[2])
+if redis.call('EXISTS', access_key) == 1 then
+    redis.call('ZREMRANGEBYSCORE', access_key, 0, now - window)
+end
+return redis.call('ZCARD', access_key)
+"""
```
achimnol: I think it should be a just simple call of `ZCARD` instead of running a full script, as the API request to get this value is already taken into account for rate limiting.

achimnol (changes requested):

src/ai/backend/manager/models/keypair.py
```diff
@@ -228,6 +234,20 @@ async def resolve_num_queries(self, info: graphene.ResolveInfo) -> int:
...
+        redis_rlim = redis_helper.get_redis_object(
+            ctx.shared_config.data["redis"], name="ratelimit", db=REDIS_RLIM_DB
+        )
+        ret = await redis_helper.execute_script(
+            redis_rlim,
+            "rollingcount",
+            _rolling_count_script,
+            [self.access_key],
+            [],
+        )
```
achimnol: Since the script now contains only a single command ZCARD, there is no need to use the complex `execute_script()` API. :wink:

achimnol (changes requested):

achimnol (approved):

Issue #2044, opened 2024-04-22
[Docs: Install from Package Guide] To add the manager's RPC client certificate entry to the guide.


studioego:
### Main idea

To add the manager's RPC client certificate entry to the guide.
Following the merged PR https://github.com/lablup/backend.ai/pull/887 , `manager.toml` config file's `rpc-auth-manager-keypair` filed is always required,  even when no agents require authentication and encryption.

As a result, Let's add the manager's RPC client certificate entry to the guide.

For the first time, generate for any new setup for security using the `backend.ai mgr generate-rpc-keypair` command.
```shell
$ backend.ai mgr generate-rpc-keypair . 'manager'
2024-04-19 07:41:10.859 INFO ai.backend.manager.cli [43943] Generating a RPC keypair...
Public Key: ]Y?Lvl9sdak3c)3?y(UoA.)6PZDFNu{@hGT>bGdm (stored at ./test.key)
Secret Key: YUX*C4X6VluV7Ah.mVb@uKVd[EbSOTgei)EtCy}3 (stored at ./test.key_secret)
```
and then, checking the [Local configuration](https://docs.backend.ai/en/latest/install/install-from-package/install-manager.html#local-configuration)
after generating new keypairs, modify `rpc-auth-manager-keypair` filed  on `manager.toml` file. 
rpc-auth-manager-keypair = "fixtures/manager/manager.key_secret"

and then pass the [Global configuration](https://docs.backend.ai/en/latest/install/install-from-package/install-manager.html#global-configuration)

PR #887, opened 2022-11-20
feat: Implement encrypted RPC channel between manager and agent

achimnol:
This PR adds authentication and encryption to our manager-to-agent RPC connections, using ZeroMQ's Curve asymmetric encryption algorithm. (refs: http://api.zeromq.org/master:zmq-curve, http://curvezmq.org/)

The motivation behind this feature is to allow distributed but secure deployment of Backend.AI agents across public networks. One of the example scenarios is deployment of micro-Backend.AI federation cluster.

Agents register their public keys to the manager via heartbeats, and the manager uses them to self-authenticate RPC calls to the agents.  The agent also checks the incoming connection's client public key to decide whether to accept or not, so that only the legitimate connections from a pre-configured manager identity can control the agent.

The manager caches each agent's address and public key in the memory (`ai.backend.manager.agent_cache`) to avoid excessive database queries for every RPC call, which is invalidated when it detects changes in the agent heartbeats.

There is no noticeable performance impacts, but this encryption and authentication adds some handshake delays (like 0.001s → 0.006s in my setup).  I expect that doing large file transfers through manager-agent RPC will be slower.

When combined with TLS and explicit credentials for Redis and etcd, the cluster admins can secure the manager-to-agent control plane when it reaches beyond public networks.  Full encryption of the container app traffic (wsproxy to agents) is planned as a future work.

### New `manager.toml` configs (mandatory)

```toml
# Set the manager's RPC client certificate (public key and secret key).
# This is a mandatory option even when no agents require authentication.
# It must be generated for any new setup for security using
# the `backend.ai mgr generate-rpc-keypair` command.
rpc-auth-manager-keypair = "fixtures/manager/manager.key_secret"
```

This field is always required, even when no agents require authentication and encryption.
Installers should automatically generate the manager keypair (aka certificate).

For easier adoption in existing development setups, the default value is set to read the working copy's fixture dircetory.

### New `agent.toml` configs (optional)

```toml
# Set the agent's private and public keypair to sign the RPC responses.
# The agent's public key is sent to the manager via heartbeats.
# [default: unset]
rpc-auth-agent-keypair = "agent.key_secret"

# Set the manager's public key to validate and accept RPC calls from the manager.
# Only effective when rpc-auth-agent-keypair is set.
# [default: unset]
rpc-auth-manager-public-key = "manager.key"
```

If these options are not set, the agent will run in the "plain-text" mode like before.

### Introduction of `AgentRPCCache`

[...]

PR #2052, opened 2024-04-23
docs: add the manager's RPC client certificate entry to the guide

studioego:
To add the manager's RPC client certificate entry to the guide.

Following the merged PR #887 , manager.toml config file's rpc-auth-manager-keypair filed is always required, even when no agents require authentication and encryption.

As a result, add the manager's RPC client certificate entry to the guide.

fix https://github.com/lablup/backend.ai/issues/2044

- [x] Documentation

----
📚 Documentation preview 📚: https://sorna--2052.org.readthedocs.build/en/2052/

----
📚 Documentation preview 📚: https://sorna-ko--2052.org.readthedocs.build/ko/2052/

### DIFF
--- a/changes/2052.doc.md
+++ b/changes/2052.doc.md
@@ -1,1 +1,1 @@
-docs: add the manager's RPC client certificate entry on "install from package" document
+Updated the package installation documentation to include instructions on adding the manager's RPC key pair.
--- a/changes/2052.doc.md
+++ b/changes/2052.doc.md
@@ -1,0 +1,1 @@
+docs: add the manager's RPC client certificate entry on "install from package" document
--- a/changes/2052.doc.md
+++ b/changes/2052.doc.md
@@ -1,1 +1,1 @@
-Updated the package installation documentation to include instructions on adding the manager's RPC key pair.
+Update the package installation documentation to include instructions on adding the manager's RPC key pair.
--- a/docs/install/install-from-package/install-manager.rst
+++ b/docs/install/install-from-package/install-manager.rst
@@ -18,8 +18,8 @@ If you want to install a specific version:
 
    $ pip install -U backend.ai-manager==${BACKEND_PKG_VERSION}
 
-You may need to generate an RPC keypair for the first setup. This key pair is used for authentication and encryption in our manager-to-agent RPC connections.
-Even if you don't intend to encrypt the RPC channel, the key pair should still be present.
+You may need to generate an RPC keypair for the first setup. This keypair is used for authentication and encryption in our manager-to-agent RPC connections.
+Even if you don't want to encrypt the RPC channel, you still need to have a keypair.
 
 
 .. code-block:: console
@@ -32,7 +32,7 @@ Even if you don't intend to encrypt the RPC channel, the key pair should still b
     Public Key: >B-mF}N{WygT92d&=Kceix$7cWzg!dT])rIc39=S (stored at fixtures/manager.key)
     Secret Key: g.4&?*b&0#oRRC9?DMO[SUXikjKZ7nYj!bzFJN92 (stored at fixtures/manager.key_secret)
 
-You generated an RPC keypair for the manager. The public key is stored in ``fixtures/manager.key`` and the secret key is stored in ``fixtures/manager.key_secret``.
+You have generated an RPC keypair for the manager. The public key is stored in ``fixtures/manager.key`` and the secret key is stored in ``fixtures/manager.key_secret``.
 
 Local configuration
 -------------------
--- a/docs/install/install-from-package/install-manager.rst
+++ b/docs/install/install-from-package/install-manager.rst
@@ -66,8 +66,8 @@ would be:
    ssl-enabled = false
 
    heartbeat-timeout = 40.0
-   rpc-auth-manager-keypair = "/home/bai/manager/fixtures/manager.key_secret"
-   pid-file = "/home/bai/manager/manager.pid"
+   rpc-auth-manager-keypair = "fixtures/manager.key_secret"
+   pid-file = "manager.pid"
    disabled-plugins = []
    hide-agents = true
    # event-loop = "asyncio"
--- a/docs/install/install-from-package/install-manager.rst
+++ b/docs/install/install-from-package/install-manager.rst
@@ -200,7 +200,7 @@ issuing SQL statement directly inside the PostgreSQL container:
 .. code-block:: console
 
    $ vfolder_host_val='{"bai-m1:local": ["create-vfolder", "modify-vfolder", "delete-vfolder", "mount-in-session", "upload-file", "download-file", "invite-others", "set-user-specific-permission"]}'
-   $ docker compose -f halfstack/postgres-cluster-default exec -it backendai-half-db psql -U postgres -d backend \
+   $ docker compose -f "$HOME/halfstack/postgres-cluster-default" exec -it backendai-half-db psql -U postgres -d backend \
          -c "UPDATE domains SET allowed_vfolder_hosts = '${vfolder_host_val}' WHERE name = 'default';"
 
 
--- a/docs/install/install-from-package/install-manager.rst
+++ b/docs/install/install-from-package/install-manager.rst
@@ -18,17 +18,21 @@ If you want to install a specific version:
 
    $ pip install -U backend.ai-manager==${BACKEND_PKG_VERSION}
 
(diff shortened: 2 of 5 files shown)


changes/2052.doc.md
```diff
@@ -0,0 +1 @@
+docs: add the manager's RPC client certificate entry on "install from package" document
```
adrysn: I think change logs is not a commit message so prefixes like `docs:` are not required. Might be changed something like:
> Updated the package installation documentation to include instructions on adding the manager's RPC key pair.

docs/install/install-from-package/install-manager.rst
```diff
@@ -18,6 +18,17 @@ If you want to install a specific version:
 
    $ pip install -U backend.ai-manager==${BACKEND_PKG_VERSION}
 
+For the first time, you need to generate rpc keypair. This keypair is used for authentication and encryption to our manager-to-agent RPC connections.
```
adrysn: My suggestion:
> You may need to generate an RPC keypair for the first setup. This key pair is used for authentication and encryption in our manager-to-agent RPC connections. Even if you don't intend to encrypt the RPC channel, the key pair should still be present.

docs/install/install-from-package/install-manager.rst
```diff
@@ -18,6 +18,17 @@ If you want to install a specific version:
...
+
+.. code-block:: console
+
+   $ cd "${HOME}/manager"
+   $ mkdir -p fixtures/manager
+   $ backend.ai mgr generate-rpc-keypair fixtures/manager 'manager'
```
adrysn: This is a minor one, but I prefer to generate the key pair under `fixtures/` not `fixtures/manager/`. It already is evident the `fixtures` will contain manager-only configurations here.

docs/install/install-from-package/install-manager.rst
```diff
@@ -51,6 +62,7 @@ would be:
    ssl-enabled = false
 
    heartbeat-timeout = 30.0
+   rpc-auth-manager-keypair = "/home/bai/manager/fixtures/manager/manager.key_secret"
```
adrysn: No need to include `/home/bai/manager`. It would be enough to just write `rpc-auth-manager-keypair = fixtures/manager.key_secret`.

docs/install/install-from-package/install-manager.rst
```diff
@@ -51,6 +62,7 @@ would be:
    ssl-enabled = false
 
    heartbeat-timeout = 30.0
```
adrysn: Let's update `heartbeat-timeout` to `40.0` since the heart beat period from agent is increased in 23.09.

adrysn (changes requested):
I added some comments.

docs/install/install-from-package/install-manager.rst
studioego: I'll change `fixtures/manager` to `fixtures/`

docs/install/install-from-package/install-manager.rst
studioego: fixed it.

docs/install/install-from-package/install-manager.rst
studioego: [The document](https://docs.backend.ai/en/latest/install/install-from-package/install-manager.html#local-configuration) specifies the location of `manager.toml` as `${HOME}/.config/backend.ai/manager.toml`

So, I modify as follows according to the above modifications. 
rpc-auth-manager-keypair =  "/home/bai/manager/fixtures/manager.key_secret"

studioego:
I fixed typos for the latest version.

docs/install/install-from-package/install-manager.rst
adrysn: Sometimes, we cannot install the manager under the `bai` account. Depending on the situation, a different OS account may be used. In such cases, it's better to remove hardcoded references if possible. By specifying `fixtures/manager.key_secret`, the `fixtures` directory will be recognized as a relative path under the directory where the manager is installed. This allows for a more generalized configuration.

studioego: I modified that removed hardcoded references(absolute path) and modified relative path on the example of manager.toml.
It ran. As a result. I fix it

YEONFEEL96 (changes requested):
Modified some explanation to make document better.

docs/install/install-from-package/install-manager.rst
```diff
@@ -18,6 +18,21 @@ If you want to install a specific version:
 
    $ pip install -U backend.ai-manager==${BACKEND_PKG_VERSION}
 
+You may need to generate an RPC keypair for the first setup. This key pair is used for authentication and encryption in our manager-to-agent RPC connections.
+Even if you don't intend to encrypt the RPC channel, the key pair should still be present.
```
YEONFEEL96, line 22:
```suggestion
You may need to generate an RPC keypair for the first setup. This keypair is used for authentication and encryption in our manager-to-agent RPC connections.
Even if you don't want to encrypt the RPC channel, you still need to have a keypair.
```

YEONFEEL96:
```suggestion
You may need to generate an RPC keypair for the first setup. This keypair is used for authentication and encryption in our manager-to-agent RPC connections.
Even if you don't want to encrypt the RPC channel, you still need to have a keypair.
```

docs/install/install-from-package/install-manager.rst
```diff
@@ -18,6 +18,21 @@ If you want to install a specific version:
...
+
+   $ cd "${HOME}/manager"
+   $ mkdir fixtures
+   $ backend.ai mgr generate-rpc-keypair fixtures 'manager'
+
+   2024-04-23 12:06:31.913 INFO ai.backend.manager.cli [21024] Generating a RPC keypair...
+    Public Key: >B-mF}N{WygT92d&=Kceix$7cWzg!dT])rIc39=S (stored at fixtures/manager.key)
+    Secret Key: g.4&?*b&0#oRRC9?DMO[SUXikjKZ7nYj!bzFJN92 (stored at fixtures/manager.key_secret)
+
+You generated an RPC keypair for the manager. The public key is stored in ``fixtures/manager.key`` and the secret key is stored in ``fixtures/manager.key_secret``.
```
YEONFEEL96:
```suggestion
You have generated an RPC keypair for the manager. The public key is stored in ``fixtures/manager.key`` and the secret key is stored in ``fixtures/manager.key_secret``.
```

YEONFEEL96 (changes requested):
made some changes in description, such as key pair to kaypair, etc. Was only able to change the newly added part.

docs/install/install-from-package/install-manager.rst
studioego: modified it.

docs/install/install-from-package/install-manager.rst
studioego: Also, I modified the description.

docs/install/install-from-package/install-manager.rst
```diff
@@ -183,7 +199,7 @@ issuing SQL statement directly inside the PostgreSQL container:
 .. code-block:: console
 
    $ vfolder_host_val='{"bai-m1:local": ["create-vfolder", "modify-vfolder", "delete-vfolder", "mount-in-session", "upload-file", "download-file", "invite-others", "set-user-specific-permission"]}'
-   $ docker exec -it bai-backendai-pg-active-1 psql -U postgres -d backend \
+   $ docker exec -it $(docker ps | grep backendai-pg | awk '{print $1}') psql -U postgres -d backend \
```
achimnol: Whenever possible, we should prefer the exact reference mechanism.
Here, we could use `docker compose exec -it backendai-pg psql -U postgres -d backend` to exactly refer to the target postgres container of the current compose stack, as long as the compose stack is initialized via the cwd's `docker-compose.yml`. Otherwise, we should specify `-p` and/or `-f` options explicitly in the `docker compose` command.

achimnol (changes requested):

docs/install/install-from-package/install-manager.rst
achimnol: There is no need to take care of the "not beautiful" container naming scheme of Docker Compose—it's just an auto-generated one. We should use the identity of compose stack and the service name to refer to the actual container, instead of their auto-generated names.

PR #2056, opened 2024-04-23
fix: Wrong usage of dataloader in GQL group resolver

fregataa:

----
📚 Documentation preview 📚: https://sorna--2056.org.readthedocs.build/en/2056/

----
📚 Documentation preview 📚: https://sorna-ko--2056.org.readthedocs.build/ko/2056/

### DIFF
--- a/src/ai/backend/manager/models/gql.py
+++ b/src/ai/backend/manager/models/gql.py
@@ -916,7 +916,9 @@ class Queries(graphene.ObjectType):
                 ctx,
                 "Group.by_user",
             )
-            client_groups = await loader.load(client_user_id, type=[ProjectType[t] for t in type])
+            client_groups = [
+                group for group in await loader.load(client_user_id) if group.type in type
+            ]
             if group.id not in (g.id for g in client_groups):
                 raise InsufficientPrivilege
         else:
--- a/src/ai/backend/manager/models/group.py
+++ b/src/ai/backend/manager/models/group.py
@@ -378,8 +378,13 @@ class Group(graphene.ObjectType):
         cls,
         graph_ctx: GraphQueryContext,
         user_ids: Sequence[uuid.UUID],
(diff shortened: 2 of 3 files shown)


agatha197 (approved):
LGTM

src/ai/backend/manager/models/gql.py
```diff
@@ -913,7 +913,9 @@ async def resolve_group(
                 ctx,
                 "Group.by_user",
             )
-            client_groups = await loader.load(client_user_id, type=[ProjectType[t] for t in type])
+            client_groups = [
+                group for group in await loader.load(client_user_id) if group.type in type
```
jopemachine, line 917: On `groups` table definition, `type` column is non-nullable. May I ask why `null` check on `group.type` is necessary here?

jopemachine:
LGTM, but we also need to edit `resolve_groups`.
https://github.com/lablup/backend.ai/blob/056510689bdec33303dd85d1d0323f5777d5bfdf/src/ai/backend/manager/models/gql.py#L1005

src/ai/backend/manager/models/group.py
```diff
@@ -378,7 +378,8 @@ async def batch_load_by_user(
         cls,
         graph_ctx: GraphQueryContext,
         user_ids: Sequence[uuid.UUID],
-        type: list[ProjectType] = [ProjectType.GENERAL],
+        *,
+        type: list[ProjectType] | None = None,
```
kyujin-cho, line 382: Is there a specific reason you decided to expose every types of project by default here? When forming the model store sprint (which is directly related to this `type` parameter) we decided to reveal such projects only if the type was explicitly specified by requester.

kyujin-cho (changes requested):

src/ai/backend/manager/models/group.py
fregataa: I forgot to add this. it is restored now

PR #2062, opened 2024-04-24
fix: Resource usage wrong calculation

fregataa:
Currently resource usage APIs calculates all project/session/kernel resource in a wrong way.
1. We did not deep-copy `ResourceUsage` objects when we register it to project/session/kernel resource usage and that causes referenced fields of `ResourceUsage` are shared with unrelated project/session usages. e.g. kernel-1 has `nfs={"/vfroot/foo"}` and kernel-2 has `nfs={"/vfroot/bar"}` but the usage API results kernel-1 has `nfs={"/vfroot/foo", "/vfroot/bar"}`.
2. `total_usage` is doubled or tripled in project/session resource usages because `total_usage` is initiated when it is registered as a child resource group and project/session resource usages add the `total_usage` again after initiate a child resource group. Let's initiate child resource usages with "zero" ResourceUsage.

### DIFF
--- a/changes/2062.fix.md
+++ b/changes/2062.fix.md
@@ -1,0 +1,1 @@
+Fix wrong calculation of resource usage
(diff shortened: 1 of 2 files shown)


agatha197 (approved):
LGTM 👍

PR #2072, opened 2024-04-26
fix: VFolder file listing not working

kyujin-cho:
Resolves lablup/giftbox#661. This PR fixes VFolder I/O operation APIs not working when user tries to call API to a VFolder which shares name with deleted one, which is caused by `vfolder_permission_required()` decorator selecting already removed VFolder instead of live one. Adding these kind of extra filters will make the codebase coarsened, so our ultimate goal is to retire every VFolder APIs utilizing VFolder name as its key to resolve target VFolder.

----
📚 Documentation preview 📚: https://sorna--2072.org.readthedocs.build/en/2072/

----
📚 Documentation preview 📚: https://sorna-ko--2072.org.readthedocs.build/ko/2072/

----
📚 Documentation preview 📚: https://sorna--2072.org.readthedocs.build/en/2072/

----
📚 Documentation preview 📚: https://sorna-ko--2072.org.readthedocs.build/ko/2072/

----
📚 Documentation preview 📚: https://sorna--2072.org.readthedocs.build/en/2072/

----
📚 Documentation preview 📚: https://sorna-ko--2072.org.readthedocs.build/ko/2072/

----
📚 Documentation preview 📚: https://sorna--2072.org.readthedocs.build/en/2072/

----
📚 Documentation preview 📚: https://sorna-ko--2072.org.readthedocs.build/ko/2072/

----
📚 Documentation preview 📚: https://sorna--2072.org.readthedocs.build/en/2072/

### DIFF
--- a/changes/2072.fix.md
+++ b/changes/2072.fix.md
@@ -1,0 +1,1 @@
+Fix VFolder file operation not working when user has been granted access to shared but deleted VFolder which has same name with the normal one
--- a/src/ai/backend/manager/api/vfolder.py
+++ b/src/ai/backend/manager/api/vfolder.py
@@ -8,7 +8,7 @@ import math
 import stat
 import uuid
 from datetime import datetime
-from enum import StrEnum
+from enum import Enum, StrEnum
 from pathlib import Path
 from types import TracebackType
 from typing import (
@@ -68,7 +68,6 @@ from ..models import (
     UserRole,
     UserRow,
     UserStatus,
-    VFolderAccessStatus,
     VFolderCloneInfo,
     VFolderDeletionInfo,
     VFolderInvitationState,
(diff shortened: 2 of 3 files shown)


achimnol (changes requested):
### Good point
* Making an explicit mapping on the top-level is a good choice. Now I can see more clearly that `VFolderAccessStatus` is NOT yet another state-machine.
### Change requests
* `VFolderAccessStatus` is not a state-machine but an alias mapping to `VFolderOperationStatus`. Let's rename it like `VFolderStatusSet` or something similar to make the meaning and intention to stand out.
* Split `vfolder_state_required()` to:
  - `vfolder_permission_required()`: the original *permission* check decorator
  - `vfolder_status_check()`: the new *status* check decorator converted from `ensure_vfolder_status()`

PR #2077, opened 2024-04-26
fix: clear-history sessions as well as kernels

fregataa:
follow-up #576 
Let the `backend.ai mgr clear-history` command clears session records as well as kernel records.

### DIFF
--- a/src/ai/backend/manager/cli/__main__.py
+++ b/src/ai/backend/manager/cli/__main__.py
@@ -5,8 +5,10 @@ import logging
 import pathlib
 import subprocess
 import sys
+import uuid
 from datetime import datetime
 from functools import partial
+from typing import cast
 
 import click
 from more_itertools import chunked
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/cli/__main__.py
```diff
@@ -302,21 +304,28 @@ async def _clear_terminated_sessions():
...
+                curs = await conn.execute(sa.select([sa.func.count()]).select_from(SessionRow))
                 if ret := curs.fetchone():
                     table_size = ret[0]
                     log.info(
-                        "The number of rows of the `kernels` tables after cleanup: {}",
+                        "The number of rows of the `sessions` tables after cleanup: {}",
                         table_size,
                     )
         log.info(
             "Cleaned up {:,} database records older than {}.",
```
achimnol, line 327: As we have changed the implementation, I think this line should be also updated to reflect that we removed N "session records" instead of "database records".

achimnol (changes requested):

src/ai/backend/manager/cli/__main__.py
fregataa: There is a info log telling the number of deleted records below so it would be duplicate. Does it need to change?

kyujin-cho: L323 already states amount of deleted sessions so I think current implementation would be sufficient for users to understand what's under the hood.

kyujin-cho (approved):

PR #2081, opened 2024-04-29
feat: cache volume objects in `RootContext.get_volume`

### LINKED ISSUE #2078: Make storage-proxy a little more stateful
Some storage-proxy backend implementations have "heavy" initialization steps.
Let's update storage-proxy's `RootContext.get_volume()` method to memoize the instances of configured volumes and add an explicit shutdown procedure of those memoized volumes upon the storage-proxy shutdown.

jopemachine:

Fixes #2078.

### DIFF
--- a/src/ai/backend/storage/context.py
+++ b/src/ai/backend/storage/context.py
@@ -94,6 +94,7 @@ def _init_subapp(
 
 
 class RootContext:
+    volumes: dict[str, AbstractVolume]
     pid: int
     etcd: AsyncEtcd
     local_config: Mapping[str, Any]
@@ -115,6 +116,7 @@ class RootContext:
         watcher: WatcherClient | None,
         dsn: Optional[str] = None,
     ) -> None:
+        self.volumes = {}
         self.pid = pid
         self.pidx = pidx
         self.node_id = node_id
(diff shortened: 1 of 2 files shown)


src/ai/backend/storage/context.py
```diff
@@ -170,28 +172,37 @@ def list_volumes(self) -> Mapping[str, VolumeInfo]:
...
+                local_config=self.local_config,
+                mount_path=Path(volume_config["path"]),
+                options=volume_config["options"] or {},
+                etcd=self.etcd,
+                event_dispathcer=self.event_dispatcher,
+                event_producer=self.event_producer,
+            )
+
+            await volume_obj.init()
+            self.volumes[name] = volume_obj
```
rapsealk, line 203: Does it work with high availability setups?

jopemachine: @rapsealk No, in the HA setup, each cache will be stored separately on the storage proxy servers.
Do all storage proxy servers need to be implemented to use the same cache?
If so, which storage should be used as the cache?

kyujin-cho: In terms of memoization I don't think this objects do not have to (and also cannot be) shared among every nodes consisting HA setup.

kyujin-cho (approved):

PR #2082, opened 2024-04-29
chore: Allow 'DELETE' method for vfolder delete-from-trash API

fregataa:
Currently only `POST` method is used for vfolder delete-from-trash API.
Let's allow `DELETE` method for easy usage.

----
📚 Documentation preview 📚: https://sorna--2082.org.readthedocs.build/en/2082/

----
📚 Documentation preview 📚: https://sorna-ko--2082.org.readthedocs.build/ko/2082/

### DIFF
--- a/changes/2092.misc.md
+++ b/changes/2092.misc.md
@@ -1,0 +1,1 @@
+Allow `DELETE` method for vfolder delete-from-trash API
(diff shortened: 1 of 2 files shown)


fregataa:
Close this PR because it will be confusing to add endpoint `folders/_/delete-from-trash`, which is different from `folders/delete-from-trash`.

PR #2084, opened 2024-04-29
fix: timeout on `compute_session_list` GQL Query

kyujin-cho:
This PR fixes querying `compute_session_list` with `commit_status` field raising timeout when there are large enough amount of kernels residing in the query scope. On current implementation every single `commit_status` resolver performed single followed by a database query, resulting in bursting request to both Redis and PostgreSQL. To solve these bottlenecks this PR refactored implementations of both `AgentRegistry.get_commit_status()` and `ComputeSession.resolve_commit_status()` methods so that it can benefit from redis' pipeline feature.

----
📚 Documentation preview 📚: https://sorna--2084.org.readthedocs.build/en/2084/

----
📚 Documentation preview 📚: https://sorna-ko--2084.org.readthedocs.build/ko/2084/

### DIFF
--- a/changes/2084.fix.md
+++ b/changes/2084.fix.md
@@ -1,0 +1,1 @@
+Fix `compute_session_list` GQL query not responding on an abundant amount of sessions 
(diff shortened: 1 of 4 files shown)


adrysn (approved):
I confirmed it dramatically reduces the session list query response time.

PR #2088, opened 2024-04-30
feature: enhanced PostgreSQL connection mgmt

fregataa:
This is a stacked PR migrated from #1817

Improve PostgreSQL connection management.
The point is replacing connection retry function `execute_with_retry()` to transaction retry function `execute_with_txn_retry()`.

## Connection and transaction management
`sqlalchemy.Connection` object and `sqlalchemy.Engine` object can open DB transactions.
We can open multiple transactions in a single DB connection by calling `Connection.begin()` while calling `Engine.begin()` opens a new DB connection for every transaction.

Currently Backend.AI Manager opens a new DB connection for every DB transaction and it can cause connection burden especially when it runs `models.utils.execute_with_retry()` which retries entire DB connection.

Let's decrease the connection burden by managing a lifecycle of `Connection`s and retry DB transaction within a connection.

- https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Connection.begin
- https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Engine.connect

- [x] Test case(s) to:

### DIFF
--- a/src/ai/backend/manager/models/utils.py
+++ b/src/ai/backend/manager/models/utils.py
@@ -4,6 +4,7 @@ import asyncio
 import functools
 import json
 import logging
+from contextlib import AbstractAsyncContextManager as AbstractAsyncCtxMgr
 from contextlib import asynccontextmanager as actxmgr
 from typing import (
     TYPE_CHECKING,
@@ -14,6 +15,7 @@ from typing import (
     Mapping,
     Tuple,
     TypeVar,
+    overload,
 )
 from urllib.parse import quote_plus as urlquote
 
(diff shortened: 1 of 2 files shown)


fregataa:
* **#2088** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2088?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/ai/backend/manager/models/utils.py
```diff
@@ -4,6 +4,7 @@
 import functools
 import json
 import logging
+from contextlib import _AsyncGeneratorContextManager as AbstractAsyncCtxMgr
```
achimnol: You may use `contextlib.AbstractAsyncContextManager` as an abstract type instead of this concrete implementation class for type checks, though usually lsp like pyright/pylance suggests this one.

src/ai/backend/manager/models/utils.py
```diff
@@ -173,9 +237,67 @@ async def advisory_lock(self, lock_id: LockID) -> AsyncIterator[None]:
...
+            f"DB serialization failed after {max_attempts} retry transactions"
+        )
+    assert result is not Sentinel.token
+    return result
+
+
 def create_async_engine(
     *args,
     _txn_concurrency_threshold: int = 0,
+    _conn_timeout: int = 0,
```
achimnol: According to the configuration schema, this should be `float`.

src/ai/backend/manager/models/utils.py
```diff
@@ -63,79 +66,140 @@ def __init__(self, *args, **kwargs) -> None:
...
-                    "The number of concurrent generic transactions ({}) "
-                    "looks too high (warning threshold: {}).",
-                    self._generic_txn_count,
-                    self._txn_concurrency_threshold,
-                    stack_info=False,
-                )
+            self._check_generic_txn_cnt()
             try:
-                yield conn
+                async with asyncio.timeout(self.conn_timeout):
```
achimnol: According to the usage of `conn_timeout`, it seems that semantically it is **not a connection timeout** because it _does not_ guard the actual database connection establishment inside connection pools. The connection establishment is done when entering the outer outer `async with engine.connect()` block. Moreover, this timeout does not cover the transaction's beginning and conclusion, the outer `async with conn.begin()` block.

What is the exact meaning of this timeout and what is it guarding?
For me, it just seems to be enforcing timeouts of the client-side database interfacing and data processing _inside_ transactions, _neither transactions nor connections_.
Shouldn't it include the transaction starts and shutdowns, i.e., the outer `async with conn.begin()` block, at least if you want to set a transaction timeout?

src/ai/backend/manager/models/utils.py
```diff
@@ -63,79 +66,140 @@ def __init__(self, *args, **kwargs) -> None:
         self._readonly_txn_count = 0
         self._generic_txn_count = 0
         self._sess_factory = sessionmaker(self, expire_on_commit=False, class_=SASession)
+        self._readonly_sess_factory = sessionmaker(self, class_=SASession)
+
+    def _check_generic_txn_cnt(self) -> None:
```
achimnol, line 70: Good refactoring! 👍🏼

achimnol (changes requested):

src/ai/backend/manager/models/utils.py
fregataa: I reverted implementation of `conn_timeout` setting because it is not strongly related to handling a lifecycle of SQLAlchemy connection management. I will rather open a new PR when we need to enforce timeouts of client-side interfacing.

achimnol (approved):

PR #2093, opened 2024-04-30
fix: VFolder invitation not accepted

kyujin-cho:
This PR fixes VFolder invitation not being accepted when invited user has record of deleted folder which shares name with new inviting one. This PR also updates `query_accessible_folders()` so that it won't return records of non-accessible (that means `delete-complete`d folders).

----
📚 Documentation preview 📚: https://sorna--2093.org.readthedocs.build/en/2093/

----
📚 Documentation preview 📚: https://sorna-ko--2093.org.readthedocs.build/ko/2093/


achimnol (changes requested):
Need to update other queries in the `query_accessible_vfolders()` function to make the new filter consistent.

achimnol (approved):
LGTM!

achimnol (changes requested):
I think it would be better to define a new alias set `MOUNTABLE` to include `READY`, `MOUNTED`, `PERFORMING`, and `CLONING`.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -172,16 +177,79 @@ class VFolderOperationStatus(enum.StrEnum):
...
+vfolder_status_map: Final[dict[VFolderStatusSet, set[VFolderOperationStatus]]] = {
+    VFolderStatusSet.READABLE: {
+        VFolderOperationStatus.READY,
+        VFolderOperationStatus.PERFORMING,
+        VFolderOperationStatus.CLONING,
+        VFolderOperationStatus.MOUNTED,
+        VFolderOperationStatus.ERROR,
+        VFolderOperationStatus.DELETE_PENDING,
+    },
+    VFolderStatusSet.READABLE: {
```
achimnol: This seems like a typo of `MOUNTABLE`! 😉

achimnol (approved):
Discussed in depth separately. Please follow-up after this PR, @kyujin-cho !

kyujin-cho:
After the internal discussion, we decided to defer changing `query_accessible_vfolders()`' blacklist-based (`.not_in(INACCESSIBLE)`) query behaviors to whitelist-based (`.in_(MOUNTABLE)`) but still retain declarations about `VFolderStatusSet.MOUNTABLE` state.

PR #2095, opened 2024-05-02
fix: `images` GQL query returning installed images only

kyujin-cho:
Follow-up of #2010. Fixes `images` GQL query returning sets of `installed` flagged images only regardless of `is_installed` request variable.


kyujin-cho:
Closing in favor of #2136.

PR #2105, opened 2024-05-03
feat: Introduce Raftify and RaftGlobalTimer to replace DistributedGlobalTimer

jopemachine:

- [ ] Test case(s) to:
- [ ] Documentation


jopemachine:
* **#2105** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2105?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

requirements.txt
```diff
@@ -95,3 +95,4 @@ backend.ai-krunner-alpine==5.2.0
 backend.ai-krunner-static-gnu==4.2.0
 
 etcd-client-py==0.3.0
+raftify==0.1.65
```
kyujin-cho: To keep the order let's move this line to the upper block in an appropriate order

src/ai/backend/common/distributed.py
```diff
@@ -16,7 +18,77 @@
...
+        event_producer: EventProducer,
+        event_factory: Callable[[], AbstractEvent],
+        interval: float = 10.0,
+        initial_delay: float = 0.0,
+    ) -> None:
+        self._event_producer = event_producer
+        self._event_factory = event_factory
+        self._stopped = False
+        self.interval = interval
+        self.initial_delay = initial_delay
```
kyujin-cho: Since these parameters are considered essential among every other timers inheriting `AbstractGlobalTimer` please consider moving these initialization steps to the initializer of superclass.

src/ai/backend/common/distributed.py
```diff
@@ -16,7 +18,77 @@
...
+
+    async def leave(self) -> None:
+        self._stopped = True
+        await asyncio.sleep(0)
+        if not self._tick_task.done():
+            try:
+                self._tick_task.cancel()
+                await self._tick_task
+            except asyncio.CancelledError:
+                pass
```
kyujin-cho: Definitions here are copy & pasted from `DistributedLockGlobalTimer`. I suggest you to just directly inherit `DistributedLockGlobalTimer` instead of `AbstractGlobalTimer` so that the `RaftGlobalTimer` can benefit from already defined `join()` and `leave()` functions.

src/ai/backend/manager/api/logs.py
```diff
@@ -246,14 +250,24 @@ async def init(app: web.Application) -> None:
...
+        )
+    else:
+        app_ctx.log_cleanup_timer = DistributedLockGlobalTimer(
+            root_ctx.distributed_lock_factory(LockID.LOCKID_LOG_CLEANUP_TIMER, 20.0),
+            root_ctx.event_producer,
+            lambda: DoLogCleanupEvent(),
+            20.0,
+            initial_delay=17.0,
+            task_name="log_cleanup_task",
+        )
```
kyujin-cho: These kind of flag-like approach can be a big burdensome when adapting a third lock backend - please refactor this to match our match-case selection convention ([ref](https://github.com/lablup/backend.ai/blob/main/src/ai/backend/manager/server.py#L716-L761)).

configs/manager/halfstack.toml
```diff
@@ -35,6 +35,11 @@ hide-agents = true
 # The order of agent selection.
 agent-selection-resource-priority = ["cuda", "rocm", "tpu", "cpu", "mem"]
 
+[raft]
+heartbeat-tick = 3
+election-tick = 10
+log-dir = "./logs"
+
```
kyujin-cho, line 44: From my understanding raft backend is only enabled when this directive is specified at the configuration file. This kind of approach can be too implicit from system manager's perspective; Please refactor activation mechanism so that config writer can explicitly mention which timer backend to be used.

src/ai/backend/manager/cli/__main__.py
```diff
@@ -341,6 +352,93 @@ async def _clear_old_error_logs():
...
+        })
+
+        peers: dict[str, Any] | None = None
+        for intial_peer in initial_peers.to_dict().values():
+            raft_client = await RaftServiceClient.build(intial_peer.get_addr())
+            try:
+                resp = await raft_client.get_peers()
+                peers = json.loads(resp)
+            except Exception as e:
+                print(f"Failed to getting peers from {intial_peer.get_addr()}: {e}")
```
kyujin-cho: To make message more sense:
```suggestion
                print(f"Failed to obtain peer information from {intial_peer.get_addr()}: {e}")
```

src/ai/backend/manager/config.py
```diff
@@ -307,11 +308,58 @@
...
+                    t.Key("host"): t.String,
+                    t.Key("port"): t.Int,
+                    t.Key("role", default=RaftNodeInitialRole.VOTER): tx.Enum(RaftNodeInitialRole),
+                })
+            )
+            | t.Null
+        ),
+    }),
+}).allow_extra("*")
+
```
kyujin-cho, line 374: Is there a specific reason to split out raft cluster related configuration as a separate file?

src/ai/backend/manager/raft/logger.py
```diff
@@ -0,0 +1,25 @@
+from typing import Any
+
+
+class Logger:
```
kyujin-cho, line 4: Is there any reason to retain separate Logger class for raft APIs only?

src/ai/backend/manager/raft/state_machine.py
```diff
@@ -0,0 +1,50 @@
...
+        self.key = key
+        self.value = value
+
+    def encode(self) -> bytes:
+        return pickle.dumps(self.__dict__)
+
+    @classmethod
+    def decode(cls, packed: bytes) -> "SetCommand":
+        unpacked = pickle.loads(packed)
+        return cls(unpacked["key"], unpacked["value"])
```
kyujin-cho, line 22: Most of the general concepts of this class can be achieved by just using `dataclass` library.

src/ai/backend/manager/raft/utils.py
```diff
@@ -0,0 +1,108 @@
...
+from raftify import (
+    Raft,
+    set_confchange_context_deserializer,
+    set_confchangev2_context_deserializer,
+    set_entry_context_deserializer,
+    set_entry_data_deserializer,
+    set_fsm_deserializer,
+    set_log_entry_deserializer,
+    set_message_context_deserializer,
+    set_snapshot_data_deserializer,
```
kyujin-cho, line 15: It is more of an upstream problem, but since we also has the key of `raftify` project: why does every context have its own `_deserializer()` API function? Won't it be better to merge these into one gateway function?

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -283,52 +311,33 @@ def _pipeline(r: Redis) -> RedisPipeline:
         )
 
         try:
-            # The schedule() method should be executed with a global lock
-            # as its individual steps are composed of many short-lived transactions.
-            async with self.lock_factory(LockID.LOCKID_SCHEDULE, 60):
-                async with self.db.begin_readonly_session() as db_sess:
-                    # query = (
```
kyujin-cho: Where did this lock context go?

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -1253,91 +1263,90 @@ def _pipeline(r: Redis) -> RedisPipeline:
             known_slot_types,
         )
         try:
-            async with self.lock_factory(LockID.LOCKID_PREPARE, 600):
-                now = datetime.now(tzutc())
+            now = datetime.now(tzutc())
 
-                async def _mark_session_preparing() -> Sequence[SessionRow]:
-                    async with self.db.begin_session() as db_sess:
```
kyujin-cho: Same here.

kyujin-cho (changes requested):

src/ai/backend/manager/types.py
```diff
@@ -41,3 +41,18 @@ class UserScope:
 
 class DistributedLockFactory(Protocol):
     def __call__(self, lock_id: LockID, lifetime_hint: float) -> AbstractDistributedLock: ...
+
+
+class RaftNodeInitialRole(str, enum.Enum):
+    LEADER = "leader"
+    VOTER = "voter"
+    LEARNER = "learner"
```
rapsealk: If raftify or raft-rs define the behaviors for each role, how about having those roles in raftify?

src/ai/backend/manager/config.py
jopemachine: Thanks for your correction. I think there is a accident while merging with main branch.
I will revert this.

src/ai/backend/manager/scheduler/dispatcher.py
jopemachine: Thanks for your correction, I think there was an accident while merging with the main branch.
I will revert them.

src/ai/backend/manager/raft/utils.py
jopemachine: I agree with your point. It would be more helpful for code readability to register callbacks all at once, rather than the current method.

I will make improvements in the PR below.
https://github.com/lablup/raftify/pull/101

src/ai/backend/manager/raft/state_machine.py
jopemachine: Although I haven't yet found a proper way to express the association with Rust types, this class is used on the Rust side as the `LogEntry` type. (Refer to the `PyLogEntry` implementation below)

https://github.com/lablup/raftify/blob/main/binding/python/src/bindings/state_machine.rs#L27-L31

This means that the `encode` and `decode` method should be callable from Rust.

If we try to refactor `SetCommand` like below, 
```py
@dataclass
class SetCommand:
    key: str
    value: str
```

We will encounter the following error when we call them from Rust.

```
2024-05-10 11:48:16,101 - ERROR    - Error handling request
Traceback (most recent call last):
  File "/Users/jopemachine/.pyenv/versions/3.12.2/lib/python3.12/site-packages/aiohttp/web_protocol.py", line 452, in _handle_request
    resp = await request_handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jopemachine/.pyenv/versions/3.12.2/lib/python3.12/site-packages/aiohttp/web_app.py", line 543, in _handle
    resp = await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jopemachine/Desktop/raftify/binding/python/examples/web_server_api.py", line 52, in put
    await raft_node.propose(message.encode())
                            ^^^^^^^^^^^^^^
AttributeError: 'SetCommand' object has no attribute 'encode'

```

If there is still a way to improve the `SetCommand` using `dataclass`, I would appreciate your advice. If there is room for improvement, I will reflect it immediately.

src/ai/backend/manager/raft/logger.py
jopemachine: The logs printed on the Rust side should be passed to the Python logger instead of being directly output to stdout or a file.

In other words, the methods of this class are called from Rust. That's why the logs output by `raft-rs` are formatted to match the Backend.AI logger format.

You can see the details of this class in the code below.
Ref: https://github.com/lablup/raftify/blob/main/binding/python/src/bindings/logger.rs

src/ai/backend/manager/config.py
jopemachine: The contents of `raft-config-cluster.toml` may be updated automatically in the future. Specifically, when new members are added to or removed from the cluster, this information should be reflected in `raft-config-cluster`. 

Therefore, it would be appropriate to manage these in different files to update only the information about the cluster members that need updating instead of overwriting the entire `manager.toml`.

However, if there is no need for the functionality to dynamically add or remove cluster members, this may be an unnecessary consideration, and it might be preferable to keep all settings in `manager.toml`.

src/ai/backend/common/distributed.py
jopemachine: To have `RaftGlobalTimer` inherit from `DistributedLockGlobalTimer` instead of `AbstractGlobalTimer`, it is necessary to define `RaftDistributedLock`. This seems to conflict with the intent to replace the existing `DistributedLock`, and to implement this, it is essential to decide how the `RaftDistributedLock` should work.

IMHO, if our purpose here is simply to reuse `join()` and `leave()`, it would be appropriate to move the implementation of these two functions to `AbstractGlobalTimer`.

src/ai/backend/manager/api/logs.py
jopemachine: I refactored the related codes in 5eb1e9f to reflect the feedback, and this issue is resolved in the commit.

In this commit, the `use_raft()` checks are replaced with match-case statements.

configs/manager/halfstack.toml
jopemachine: I refactored the related codes in 5eb1e9f to reflect the feedback, and this issue is resolved in the commit.

In this commit, I created `global-timer` option to `manager.toml` (manager local config). 

I think the `global-timer` option allows to specify whether to use `raft` or `distributed-lock` more explicitly.

May I ask for a review for the commit?

jopemachine:
CI test is failing while downloading raftify package, this should be fixed after setting up raftify package deployment workflow

src/ai/backend/manager/types.py
jopemachine: In fact, in Raftify, `InitialRole` type is already defined on the Rust side, and this type exists in the Python bindings as well. 

The reason I define and use a separate Enum with the same contents here is because the type is not an actual Enum in Python.

For instance, below type validating code produce the following erorr.

```py
from raftify import InitialRole
...
	t.Key("role", default=InitialRole.VOTER): tx.Enum(InitialRole),
```

```
  File "/home/jopemachine/backend.ai/dist/export/python/virtualenvs/python-default/3.12.2/lib/python3.12/site-packages/trafaret/base.py", line 1143, in transform
    for k, v, names in key(value, context=context):
  File "/home/jopemachine/backend.ai/dist/export/python/virtualenvs/python-default/3.12.2/lib/python3.12/site-packages/trafaret/base.py", line 972, in __call__
    result = self.trafaret(self.get_data(data, default), context=context)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jopemachine/backend.ai/dist/export/python/virtualenvs/python-default/3.12.2/lib/python3.12/site-packages/trafaret/base.py", line 152, in __call__
    return self.check(val, context=context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jopemachine/backend.ai/dist/export/python/virtualenvs/python-default/3.12.2/lib/python3.12/site-packages/trafaret/base.py", line 115, in check
    return self.check_and_return(value)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jopemachine/backend.ai/src/ai/backend/common/validators.py", line 216, in check_and_return
    return self.enum_cls(value)
           ^^^^^^^^^^^^^^^^^^^^
TypeError: No constructor defined
```

There might be a way to handle Enum types more wisely in writing PyO3 bindings. 
There are several aspects for improvement currently existing on the Python bindings side of Raftify like this.

Ref: https://github.com/lablup/raftify/issues/91

PR #2117, opened 2024-05-07
fix: Filter out host permissions of model store

fregataa:
"Model store" projects have `allowed_host_permission` values and it is used when determining folder permissions.
We should not use host permission of model stores because it is out of the purpose of the model store project.


fregataa:
Closed this PR because it is too old to maintain and we will add a new permission layer, also known as RBAC

PR #2132, opened 2024-05-08
ci: Independent distribution for more frequently distributed webui

### LINKED ISSUE #2017: Allow more frequent release of webui and webserver
[Our webui](https://github.com/lablup/backend.ai-webui) has faster release paces as it involves many UI-level bug fixes and enhancements.

Let's allow releasing it more frequently than the core, without duplicating all other core packages to increment the version of webui.

Since the webui release snapshot is embedded inside the webserver package (`src/ai/backend/web/static` at `backend.ai-webserver`), we need to be able to update and re-release the `backend.ai-webserver` package only. For instance, we keep the core version to "24.03.1" while incrementing the webserver and webui version to "24.03.1.1".

- [ ] Add a new GitHub Actions workflow that can be manually triggered to fetch the given tag name (e.g., "24.03.1.1") of the webui and build the webserver wheel and scie using it.\* [ ] Overwrite the `*-webserver` scie executables and their shasum hashes in the GitHub release (e.g., "24.03.1").
- [ ] Upload the PyPI wheel `backend.ai-webserver` only, with the incremented version (e.g., "24.03.1.1") because we cannot overwrite already published wheels.

Yaminyam:
webui has more changes patched faster than the main repo.
Therefore, the monorepo's distribution cycle cannot keep up with this speed, so only the webui package can be distributed more frequently through fourth versioning.
close #2017


.github/workflows/webserver-release.yml
```diff
@@ -0,0 +1,178 @@
...
+jobs:
+  build-scies:
+    strategy:
+      fail-fast: false
+      matrix:
+        # ubuntu-latest: intel
+        # linux-aarch64: aarch64 (self-hosted)
+        # macos-12: intel
+        # macos-13: apple silicon
+        os: [ubuntu-latest, linux-aarch64, macos-13-xlarge, macos-12-large]
```
HyeockJinKim: It's a minor thing, but it bothers me that the order of macos-12 and macos-13 is different from the comment.

HyeockJinKim (commented):
Have you verified that the github action is working?

.github/workflows/webserver-release.yml
Yaminyam: This is taken from ci.yml.
I will modify ci.yml in the future as well.

Yaminyam:
> �Have you verified that the github action is working?

It's hard to actually run it because you have to check if the pypi and github release are uploaded.
You can create a fork repository with https://test.pypi.org/ and fork repository to verify, but you'll also need to set several secrets and environment variables.

HyeockJinKim:
> > �Have you verified that the github action is working?
> 
> It's hard to actually run it because you have to check if the pypi and github release are uploaded. You can create a fork repository with [test.pypi.org](https://test.pypi.org/) and fork repository to verify, but you'll also need to set several secrets and environment variables.

Is there a way to run the github action job pr before it is merged, without affecting production, using an option like `--dry-run`?
I want to check once before merging the PR whether there is any problem when the github action is actually triggered.

but.. I guess twine hasn't implemented dry-run mode yet.. 😢 
https://github.com/pypa/twine/issues/207

Yaminyam:
> > > �Have you verified that the github action is working?
> > 
> > 
> > It's hard to actually run it because you have to check if the pypi and github release are uploaded. You can create a fork repository with [test.pypi.org](https://test.pypi.org/) and fork repository to verify, but you'll also need to set several secrets and environment variables.
> 
> Is there a way to run the github action job pr before it is merged, without affecting production, using an option like `--dry-run`? I want to check once before merging the PR whether there is any problem when the github action is actually triggered.
> 
> but.. I guess twine hasn't implemented dry-run mode yet.. 😢 [pypa/twine#207](https://github.com/pypa/twine/issues/207)

Fortunately, the action doesn't modify any existing logic, and the trigger is also manually triggered.
We'll use it manually to avoid impacting existing releases and automate it once we know it's stable.

.github/workflows/webserver-release.yml
```diff
@@ -0,0 +1,178 @@
...
+jobs:
+  build-scies:
+    strategy:
+      fail-fast: false
+      matrix:
+        # ubuntu-latest: intel
+        # linux-aarch64: aarch64 (self-hosted)
+        # macos-12: intel
+        # macos-13: apple silicon
+        os: [ubuntu-latest, linux-aarch64, macos-12-large, macos-13-xlarge]
```
HyeockJinKim: I think you should update os version. @Yaminyam

Yaminyam:
This became unnecessary when we started doing release cycles every two weeks.

PR #2145, opened 2024-05-14
fix: avoid duplicate `set_quota` action

fregataa:
In all inherited `QuotaModel`, both `create_quota_scope()` and `update_quota_scope()` methods create quota scope.
`PATCH /quota-scope API` conditionally calls both methods and it causes duplicate quota creating jobs.
Let's return right after calling `create_quota_scope()`


src/ai/backend/storage/ddn/__init__.py
```diff
@@ -159,7 +156,8 @@ async def create_quota_scope(
         except CalledProcessError as e:
             raise RuntimeError(f"'lfs project -p {project_id}' command failed: {e.stderr}")
```
kyujin-cho, line 157: <i class='graphite__hidden'>[Re: lines 144 to 157]<br/><br/></i>

Could you tell me more about changing logic here in a way of calling `lfs`?<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2145?utm_source=unchanged-line-comment">Graphite</a>.</span>

kyujin-cho (changes requested):

src/ai/backend/storage/ddn/__init__.py
fregataa: `lfs project -p ...` command maps DDN Project ID to a directory. This enables us to set a quota to a directory, so it should be done in `create_quota_scope()` even there is no specified quota option.

PR #2153, opened 2024-05-16
feat: Add parameters to check-presets SDK

fregataa:
check-presets API in manager side have `group` and `scaling_group` parameters but the counterpart function in SDK does not have them. Let's add those parameters


adrysn (approved):
LGTM.

PR #2155, opened 2024-05-17
fix: Ensure that all pending sessions are picked by schedulers

fregataa:
Currently, a scheduler fetches one list of all pending session in a given scaling group and it iterates them to schedule pending sessions one by one.
If one of the pending sessions raises `InstanceNotAvailable` or `GenericBadRequest` exceptions (which is normal), the scheduler skips all trailing pending sessions in the list.
The scheduler should keep iterating, not stop scheduling.

### DIFF
--- a/changes/2155.fix.md
+++ b/changes/2155.fix.md
@@ -1,0 +1,1 @@
+Ensure that all pending sessions are picked by schedulers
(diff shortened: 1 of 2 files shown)


jopemachine:
I confirmed that after creating a session with arbitrarily oversized resources, resulting in a `no-available-instances` state's pending session, creating another session skips the previous session and schedules the next available schedulable session.

kyujin-cho (approved):

PR #2157, opened 2024-05-17
fix: Correctly register and deregister terminating kernels

fregataa:
`Agent` registers kernel ids to its `terminating_kernels` so that it can track kernels under "terminate" lifecycle.
There are some holes that kernel ids cannot be deleted from the `terminating_kernels`, which block destroying or cleaning the kernels by `sync_container_lifecycle` task.


fregataa:
Continue this work on #2178 , close this PR

PR #2170, opened 2024-05-22
fix: Netapp `create_qtree` timing issue

fregataa:
There might be a slight delay between qtree-creation-job success response and the directory creation.
Let's poll after `create_qtree`

### DIFF
--- a/src/ai/backend/storage/netapp/__init__.py
+++ b/src/ai/backend/storage/netapp/__init__.py
@@ -18,6 +18,14 @@ from typing import (
 
 import aiofiles
 import aiofiles.os
+from tenacity import (
+    AsyncRetrying,
+    RetryError,
+    TryAgain,
+    retry_if_exception_type,
+    stop_after_attempt,
+    wait_fixed,
+)
 
 from ai.backend.common.logging import BraceStyleAdapter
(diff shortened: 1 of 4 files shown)


src/ai/backend/storage/netapp/__init__.py
```diff
@@ -87,8 +115,10 @@ async def create_quota_scope(
                 options,
             )
             self.netapp_client.check_job_result(result, [])
-            result = await self.netapp_client.enable_quota(self.volume_id)
-            self.netapp_client.check_job_result(result, ["5308507"])  # pass if "already on"
+            if not self.quota_enabled:
+                result = await self.netapp_client.enable_quota(self.volume_id)
+                self.netapp_client.check_job_result(result, ["5308507"])  # pass if "already on"
```
kyujin-cho: Could you tell me about the meaning of literal `5308507`?

kyujin-cho (changes requested):

src/ai/backend/storage/netapp/__init__.py
fregataa: `530857` code means quota is enabled in the given volume. I defined an enum for it

kyujin-cho (approved):

PR #2174, opened 2024-05-24
fix: `check-docker.py` script not working

agatha197:
This PR resolves `install-dev.sh` script failing due to recent changes in `requests` library by lifting up dependencies to `requests` and instead benefiting from `aiohttp`'s native Unix Domain Socket connector feature.

### DIFF
--- a/scripts/check-docker.py
+++ b/scripts/check-docker.py
@@ -1,4 +1,5 @@
 import argparse
+import asyncio
 import base64
 import functools
 import hashlib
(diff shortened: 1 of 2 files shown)


kyujin-cho:
As noted at https://github.com/psf/requests/issues/6707#issuecomment-2123385729 requests won't rollback the regression, so I think it is up to us to refactor the root cause from our side, rather than just keep pinning request version forever.

PR #2185, opened 2024-05-27
feat: Add centos8 dropbear binaries

jopemachine:
Partially fix https://github.com/lablup/giftbox/issues/677.


jopemachine:
* **#2185** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2185?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

jopemachine:
The version `centos8.0` may not be correct because higher compatible versions like `centos8.1` exist. 
However, when using the name `centos8`, the current `find_artifacts` function finds this value invalid and skips.

https://github.com/lablup/backend.ai/blob/be3a3056efccae7ae28cf9377e932150820bcddf/src/ai/backend/agent/agent.py#L417-L423

Therefore, I think maybe we should use names like `centos8.0` or `centos8.x`, or modify the implementation of the `find_artifact`.

jopemachine:
Since each PR is very small, I think it would be more concise to handle it as a single PR instead of a stack of PR in this case.
Therefore, I will handle these changes in a single PR.

Ref: #2220.

PR #2186, opened 2024-05-27
chore: fix incorrect version notation of GQL Field (#1993)

lablup-octodog:
This is an auto-generated backport PR of #1993 to the 24.03 release.


kyujin-cho:
Already backported manually.

PR #2220, opened 2024-05-30
feat: Add CentOS 8.0 binaries

jopemachine:
Partially fix https://github.com/lablup/giftbox/issues/677.

Due to the difference in the glibc version, we cannot run Backend.AI on the CentOS8 image. 
So adding CentOS8 krunner binary is needed.

And this PR also adds build scripts for `backend.ai-jail` and `backend.ai-hook`.

### version issue 

The version `centos8.0` may not be correct because higher compatible versions like `centos8.1` exist. 
However, when using the name `centos8`, the current `find_artifacts` function finds this value invalid and skips.

https://github.com/lablup/backend.ai/blob/be3a3056efccae7ae28cf9377e932150820bcddf/src/ai/backend/agent/agent.py#L417-L423

Therefore, I think maybe we should use names like `centos8.0` or `centos8.x`, or modify the implementation of the `find_artifact`.

### zlib upgrade

It appears that ZLIB 1.2.11 is no longer distributed. 
Therefore, the `build-sftpserver.sh` for CentOS 8.0 was built with ZLIB version 1.3.1. 
But this PR does not include binary updates for other distributions.

### Switching repository to the *vault.centos.org*

CentOS 8 support ended in late 2021, disabling its main repositories. 
Switching to the `vault.centos.org` repository enables continued access to packages and updates for unsupported versions.

```
FROM centos:centos8
RUN sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-Linux-*
```

---


jopemachine:
* **#2220** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2220?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

scripts/agent/build-jail.sh
```diff
@@ -0,0 +1,26 @@
+#! /bin/bash
```
kyujin-cho: Jail already ships its own [binary distribution](https://github.com/lablup/backend.ai-jail/releases) so this script is redundant.

scripts/agent/build-libbaihook.sh
```diff
@@ -0,0 +1,26 @@
+#! /bin/bash
```
kyujin-cho: Since we are also in charge of `backend.ai-hook` project it will be better to implement release pipeline of its own instead of just adding a single script in the dependent repository.

kyujin-cho (changes requested):

PR #2251, opened 2024-06-04
fix: Robust check on pqsql enum type name

fregataa:
`vfolders.status` field was `pgsql.Enum` type field before migrating it from enum to string and the enum name was `vfolderoperationstatus`. But in some machines, the enum name is `vfolderstatus`
Let's handle such DB schema divergent


fregataa:
Closed this PR because #3829 has resolved the issue

Issue #301, opened 2021-10-11
View print_pagination_list beautifully


mindgitrwx:
**Is your feature request related to a problem? Please describe.**`/client-py/src/ai/bakcend/client/output/console.py` has print_pagenation_function. This function generates a table that describes given input from the client's side command line.

But if too many columns have to be fetched, table alignment broke.This picture shows `backend.ai ps` generates readability!https://user-images.githubusercontent.com/37054916/136814635-1c5522b7-8c56-4c94-ac29-d599ca1b8393.png!

Small size terminal shows much more indicative output.![https://user-images.githubusercontent.com/37054916/136817156-f60a49fe-88d5-460b-9156-40b12b4a4da6.png](https://user-images.githubusercontent.com/37054916/136817156-f60a49fe-88d5-460b-9156-40b12b4a4da6.png)

\*Describe the solution you'd like\*A clear and concise description of what you want to happen.

In `/client-py/src/ai/bakcend/client/cli/pagenation.py`, `echo_via_pager` function has been defined. I modified the functionwith ['rich'](https://github.com/willmcgugan/rich) package.  The results are as followshttps://github.com/willmcgugan/rich!https://user-images.githubusercontent.com/37054916/136815698-6c6c60f1-9880-475e-ae5f-03fca17f9fed.png!

But several problems still remain here.

1. With `no_wrap  = True` option, some of the columns can not be displayed.

1. With `no_wrap  = False` option, the alignment of the table is still broken.

(no_wrap is rich's internal option)

\*Describe alternatives you've considered\*A clear and concise description of any alternative solutions or features you've considered.

1.Default `backend.ai ps` fetches too many columns to shows output in a small shell. Can the number of default columns be reduced?

1. Just leave the `echo_via_pager`, make a new function for pretty formatting
**Additional context**

['rich'](https://github.com/willmcgugan/rich) package seems to have a wide variety of uses for this project.

JIRA Issue: BA-286

PR #2254, opened 2024-06-05
feat: GQL resolvers for querying agents' total resource allocation

jopemachine:
Fix https://github.com/lablup/giftbox/issues/324.

This PR adds per-resource-group GQL resolvers to query the status of total resource allocation.

----
📚 Documentation preview 📚: https://sorna--2254.org.readthedocs.build/en/2254/

----
📚 Documentation preview 📚: https://sorna-ko--2254.org.readthedocs.build/ko/2254/


jopemachine:
* **#2254** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2254?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/ai/backend/manager/models/scaling_group.py
```diff
@@ -315,6 +315,17 @@ class ScalingGroup(graphene.ObjectType):
     scheduler_opts = graphene.JSONString()
     use_host_network = graphene.Boolean()
 
+    # Dynamic fields.
+    agent_count = graphene.Int(description="Added in 24.03.0.")
+
+    async def resolve_agent_count(self, info: graphene.ResolveInfo) -> int:
+        from .agent import Agent
```
jopemachine, line 346: FYI, to prevent cyclic imports between `agent.py` and `scaling_group.py`, lazy import is used here.

src/ai/backend/manager/models/scaling_group.py
```diff
@@ -315,6 +320,45 @@ class ScalingGroup(graphene.ObjectType):
...
+
+        graph_ctx = info.context
+        async with graph_ctx.db.begin_readonly() as conn:
+            result = (
+                await conn.execute(
+                    sa.select([agents.c.occupied_slots, agents.c.available_slots]).where(
+                        agents.c.scaling_group == self.name
+                    )
+                )
+            ).fetchall()
```
fregataa: Let's use SQLAlchemy ORM API here rather than Core API.
```suggestion
        async with graph_ctx.db.begin_readonly_session() as db_session:
            query_stmt = sa.select(AgentRow).where(AgentRow.scaling_group == self.name)
            result = await db_session.scalars(query_stmt)
            agent_rows = cast(list[AgentRow], result)
            for row in agent_rows:
                # row.occupied_slots and row.available_slots are ResourceSlot type
            ...
```

src/ai/backend/manager/models/scaling_group.py
```diff
@@ -315,6 +320,45 @@ class ScalingGroup(graphene.ObjectType):
     scheduler_opts = graphene.JSONString()
     use_host_network = graphene.Boolean()
 
+    # Dynamic fields.
+    agent_count = graphene.Int(description="Added in 24.03.0.")
+    agent_total_resource_slots = graphene.JSONString(description="Added in 24.03.0.")
```
fregataa: The version those fields will be added should be `24.03.5` or later than that. Please check the latest release version and match the version notations.

fregataa (changes requested):

src/ai/backend/manager/models/scaling_group.py
```diff
@@ -315,6 +320,42 @@ class ScalingGroup(graphene.ObjectType):
...
+        self, info: graphene.ResolveInfo
+    ) -> Mapping[str, Any]:
+        from .agent import AgentRow
+
+        graph_ctx = info.context
+        async with graph_ctx.db.begin_readonly() as db_session:
+            query_stmt = sa.select([AgentRow.occupied_slots, AgentRow.available_slots]).where(
+                AgentRow.scaling_group == self.name
+            )
+            result = (await db_session.execute(query_stmt)).fetchall()
```
fregataa: `ExtendedAsyncSAEngine.begin_readonly()` returns `Connection` object and it allows SQLAlchemy Core APIs only. 
How about using `ExtendedAsyncSAEngine.begin_readonly_session()` to open SQLAlchemy session and uses ORM APIs?

jopemachine: I never noticed that, thank you for the review! 👍

src/ai/backend/manager/models/scaling_group.py
```diff
@@ -315,6 +320,42 @@ class ScalingGroup(graphene.ObjectType):
...
+        self, info: graphene.ResolveInfo
+    ) -> Mapping[str, Any]:
+        from .agent import AgentRow
+
+        graph_ctx = info.context
+        async with graph_ctx.db.begin_readonly_session() as db_session:
+            query_stmt = sa.select([AgentRow.occupied_slots, AgentRow.available_slots]).where(
+                AgentRow.scaling_group == self.name
+            )
+            result = (await db_session.execute(query_stmt)).fetchall()
```
fregataa:
```suggestion
            query_stmt = sa.select(AgentRow).where(
                AgentRow.scaling_group == self.name
            ).options(load_only(AgentRow.occupied_slots, AgentRow.available_slots))
            result = (await db_session.scalars(query_stmt)).all()
            result = cast(list[AgentRow], result)
```
We can use more ORM patterns like this

src/ai/backend/manager/models/scaling_group.py
```diff
@@ -315,6 +320,42 @@ class ScalingGroup(graphene.ObjectType):
...
+
+            total_occupied_slots = ResourceSlot()
+            total_available_slots = ResourceSlot()
+
+            for occupied, available in result:
+                total_occupied_slots += occupied
+                total_available_slots += available
```
fregataa:
```suggestion
            for agent_row in result:
                total_occupied_slots += agent_row. occupied_slots
                total_available_slots += agent_row. available_slots
```

fregataa (changes requested):
As I checked, this does not fix https://github.com/lablup/giftbox/issues/301.
The issue of #301 is to show the number of containers in an agent, not the number of agents in a scaling group. Please let me know if I missed something about the issue

jopemachine:
> As I checked, this does not fix [lablup/giftbox#301](https://github.com/lablup/giftbox/issues/301). The issue of #301 is to show the number of containers in an agent, not the number of agents in a scaling group. Please let me know if I missed something about the issue

Sorry, I got it confused with another issue. I will address this issue in a separate PR.

src/ai/backend/manager/models/scaling_group.py
jopemachine: I didn't understand what you meant `more ORM pattern`.
Could you let me know what advantage we can take from this query?
I'm curious to know when and for what purpose I should use the `options`, and `load_only` instead of just using a select statement.

agatha197 (changes requested):
There is a mismatch in the agent count. Even though the connected agent count is 1, I got 2 when I tested on the local machine.

jopemachine:
> There is a mismatch in the agent count. Even though the connected agent count is 1, I got 2 when I tested on the local machine.

@agatha197 Could you check if only one agent is displayed on the WebUI compute resources page (like the below screenshot)?




If only one agent is displayed but the query below returns 2, that would be very strange..

```
# query 
query ($name: String!) {
  scaling_group(name: $name) {
    agent_count
  }
}

# variables
{"name": "default"}
```

src/ai/backend/manager/api/schema.graphql
```diff
@@ -693,6 +693,12 @@ type ScalingGroup {
   scheduler: String
   scheduler_opts: JSONString
   use_host_network: Boolean
+
+  """Added in 24.03.5."""
+  agent_count: Int
```
agatha197:
```suggestion
  alive_agent_count: Int
```

agatha197 (changes requested):

src/ai/backend/manager/api/schema.graphql
```diff
@@ -693,6 +693,12 @@ type ScalingGroup {
...
+
+  """Added in 24.03.5."""
+  agent_total_resource_slots: JSONString
```
agatha197:
```suggestion
  alive_agent_total_resource_slots: JSONString
```

agatha197 (changes requested):

src/ai/backend/manager/models/scaling_group.py
fregataa: it is not what we must do but it is recommended to use ORM pattern in this system because of the following reasons.
1. It is too hard to distinguish which API to use among `execute()`, `scalar()` and `scalars()`. The result of those APIs are different between SQLAlchemy Core API and ORM API, which can cause error. We can reduce such confusion by using only ORM APIs.
2. Easy to guess type. Using ORM APIs ensure that the result of queries are always ORM objects.

We can simply query like this `sa.select([AgentRow.available_slots, AgentRow. occupied_slots])` as you said.
Using ORM patterns is not what we must do. But strong type checking is what we should do. So, let's apply type casting and let the type-checker checks thoroughly

src/ai/backend/manager/models/scaling_group.py
```diff
@@ -315,6 +320,43 @@ class ScalingGroup(graphene.ObjectType):
...
+        self, info: graphene.ResolveInfo
+    ) -> Mapping[str, Any]:
+        from .agent import AgentRow, AgentStatus
+
+        graph_ctx = info.context
+        async with graph_ctx.db.begin_readonly_session() as db_session:
+            query_stmt = sa.select([AgentRow.occupied_slots, AgentRow.available_slots]).where(
+                AgentRow.scaling_group == self.name and AgentRow.status == AgentStatus.ALIVE
+            )
+            result = (await db_session.execute(query_stmt)).fetchall()
```
fregataa:
```suggestion
            result = (await db_session.execute(query_stmt)).fetchall()
            result = cast(list[tuple[ResourceSlot, ResourceSlot]], result)
```

fregataa (changes requested):
Looks good in general. I left one comment

src/ai/backend/manager/models/scaling_group.py
```diff
@@ -315,6 +320,59 @@ class ScalingGroup(graphene.ObjectType):
     scheduler_opts = graphene.JSONString()
     use_host_network = graphene.Boolean()
 
+    # Dynamic fields.
+    agent_count_by_status = graphene.Field(
+        graphene.Int,
+        description="Added in 24.03.5.",
+        status=graphene.String(default_value="ALIVE"),
```
fregataa:
```suggestion
        description=f"Added in 24.03.5. Should be one of {[s.name for s in AgentStatus]}",
        status=graphene.String(default_value=AgentStatus.ALIVE.name),
```
With such detail description, clients will be able to know what values are available for this field.
And `Enum` type values have `name` string field. Let's use it

src/ai/backend/manager/models/scaling_group.py
```diff
@@ -315,6 +320,59 @@ class ScalingGroup(graphene.ObjectType):
...
+    )
+
+    agent_total_resource_slots_by_status = graphene.Field(
+        graphene.JSONString,
+        description="Added in 24.03.5.",
+        status=graphene.String(default_value="ALIVE"),
```
fregataa: Same here

src/ai/backend/manager/models/scaling_group.py
```diff
@@ -315,6 +320,59 @@ class ScalingGroup(graphene.ObjectType):
...
+    )
+
+    async def resolve_agent_count_by_status(
+        self, info: graphene.ResolveInfo, status: str = "ALIVE"
```
fregataa:
```suggestion
        self, info: graphene.ResolveInfo, status: str = AgentStatus.ALIVE.name
```

src/ai/backend/manager/models/scaling_group.py
```diff
@@ -315,6 +320,59 @@ class ScalingGroup(graphene.ObjectType):
...
+        from .agent import Agent
+
+        return await Agent.load_count(
+            info.context,
+            raw_status=status,
+            scaling_group=self.name,
+        )
+
+    async def resolve_agent_total_resource_slots_by_status(
+        self, info: graphene.ResolveInfo, status: str = "ALIVE"
```
fregataa:
```suggestion
        self, info: graphene.ResolveInfo, status: str = AgentStatus.ALIVE.name
```

fregataa (changes requested):

agatha197 (changes requested):
We need to add `status` param to `scaling_group`, `scaling_groups`, `scaling_groups_for_xxx` queries.

fregataa (approved):
LGTM

agatha197 (approved):
LGTM

PR #2178, opened 2024-05-26
fix: Keep `sync_container_lifecycles()` bgtask alive in a loop.

fregataa:
### TL; DR
`sync_containers_lifecycle()` should work

### Detail
We have 3 different sources of Kernel / Container data
1. Manager side DB
2. Agent side `kernel_registry`
3. Agent side actual docker containers

I tried to synchronize ALL Agent's data to Manager's DB but synchronizing actual containers to `kernel_registry` or manager's data needs to process container destruction tasks which take a long time.
Since `sync_containers_lifecycle()` task synchronizes actual containers to `kernel_registry` periodically in background, resource-sync API only needs to sync `kernel_registry` to Manager side DB if it is ensured that the background task stays alive and works as we expect.

### Fix
- Keep `sync_container_lifecycles()` bgtask alive in a loop by handling any exceptions raised in the task.
- In the try-except context of container creation, **do not** use unbound local variable

### Enhance
- Fetch all containers eagerly in `sync_container_lifecycles()`

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version

PR #2263, opened 2024-06-07
refactor: fetch all containers eagerly when sync containers

fregataa:
`AbstractAgent.enumerate_containers()` fetches containers whose status matches with `status` parameter.
Currently, `sync_container_lifecycles()` fetches containers twice by setting different status option.
Let's fetch containers all at once.


fregataa:
#2178 will fix the same issue

PR #2274, opened 2024-06-13
ci: Fix misjudgment as edge when release branch is missing

Yaminyam:

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -462,13 +462,13 @@ jobs:
       # wheels after buildling them to add arch-specific tags.
       run: |
         twine upload dist/*.whl dist/*.tar.gz
-    - name: Extract stable release version
-      id: extract_stable_release_version
+    - name: Extract minor version
+      id: extract_minor_version
       run: |
         release_version=$(awk -F'.' '{print $1"."$2}' <<< "${{ github.ref_name }}")
         echo "RELEASE_VERSION=$release_version" >> $GITHUB_OUTPUT
     - name: Update stable installer shorten URL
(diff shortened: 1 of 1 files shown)


.github/workflows/default.yml
```diff
@@ -463,10 +463,18 @@ jobs:
     - name: Extract edge release version
       id: extract_edge_release_version
       run: |
-        release_version=$(git branch -r | grep -E 'origin/[0-9]{2}\.[0-9]{2}$' | awk -F'/' '{print $2}' | sort -V | tail -n 1)
-        echo "RELEASE_VERSION=$release_version" >> $GITHUB_OUTPUT
+        edge_release=$(git ls-remote --heads | grep -E 'refs/heads/[0-9]{2}\.[0-9]{2}$' | awk -F'/' '{print $3}' | sort -V | tail -n 1)
+        echo "EDGE_RELEASE=$edge_release" >> $GITHUB_OUTPUT
+        git ls-remote --heads | grep -q "refs/heads/${{ steps.extract_minor_version.outputs.RELEASE_VERSION }}\$"
+        if [ $? -eq 0 ]; then
+          is_undistributed=false
```
achimnol, line 470: "Undistributed" looks like a strange konglish.
Please remember the principle: _Use negations sparingly._
**Don't make an extra step of indirection and reasoning in the code reader's head.**

In this case, you may just inverse the branches to rename it "is_distributed", or even better, "is_released" to be consistent with the command above the line.

achimnol (changes requested):

.github/workflows/default.yml
achimnol: Also, I'd like to be sure that the string value "false" is really evaluated as false in the followed conditional expression.

PR #2353, opened 2024-06-30
ci: Update the regex and list of the PR title prefix checker

achimnol:
This PR updates the PR title prefix checker as follows:

- Allow `chore:` in addition to the `chore(...):` pattern.
- Allow `deps:`.
- Allow `doc:` in addition to `docs:`.
- Align inner indentation of the error message for better readability.

### DIFF
--- a/.github/workflows/pr-prefix.yml
+++ b/.github/workflows/pr-prefix.yml
@@ -10,18 +10,21 @@ jobs:
     steps:
       - name: Check PR title prefix
         run: |
-          if echo "${{ github.event.pull_request.title }}" | tr ':' '\n' | head -n 1 | grep -qE 'feat|fix|docs|refactor|ci|chore(\([^)]+\)|$)';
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-prefix.yml
```diff
@@ -10,18 +10,21 @@ jobs:
     steps:
       - name: Check PR title prefix
         run: |
-          if echo "${{ github.event.pull_request.title }}" | tr ':' '\n' | head -n 1 | grep -qE 'feat|fix|docs|refactor|ci|chore(\([^)]+\)|$)';
+          if echo "${{ github.event.pull_request.title }}" | tr ':' '\n' | head -n 1 | grep -qE '^(feat|fix|docs?|refactor|ci|chore(\([^)]+\))?|deps)$';
```
Yaminyam, line 13: In fact, the chore: form was already possible.
https://github.com/lablup/backend.ai/actions/runs/9676696895/job/26696837465?pr=2310
But I think your way is a better way!

Yaminyam (approved):

.github/workflows/pr-prefix.yml
achimnol: Ah, yes it works, but I think the new regex is less confusing.

Issue #2045, opened 2024-04-22
GraphQL rate limiting


achimnol:
GraphQL clients may make a wide range of queries that may affect the database and server performance. It's recommended to have "rate limiting" to prevent excessive queries.

Let's add the followings:
- Check availability of graphene's internal query validators to limit the maximum number of nodes and depth of nested queries. Apply them if available, or write our own ones.
- Apply the maximum query execution timeout in the executor level.

Yaminyam:
Currently, the pagination of the relay connection does not enforce the first or last argument, so it is impossible to predict how many nodes the connection will call.
Force the first or last argument to avoid making too large a request by calling all nodes at once, and make it possible to predict the number of nodes to be called in the gql query.

achimnol:
Resolved in #2709 (somehow it did not close this issue automatically...).

PR #2356, opened 2024-07-01
feat: Set timeout in graphql api

Yaminyam:
relation #2045 
Since graphql queries can be written in various forms, queries that place a lot of load on the server may be created.
Since users using the API cannot know how each query operates, it is best to prevent queries that are too costly on the server.
So, set a timeout to prevent queries that take too long.

### DIFF
--- a/src/ai/backend/manager/api/admin.py
+++ b/src/ai/backend/manager/api/admin.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import logging
 import traceback
 from typing import TYPE_CHECKING, Any, Iterable, Tuple
@@ -45,6 +46,9 @@ class GQLLoggingMiddleware:
         return next(root, info, **args)
 
 
+GQL_TIMEOUT = 60
+
+
 async def _handle_gql_common(request: web.Request, params: Any) -> ExecutionResult:
     root_ctx: RootContext = request.app["_root.context"]
     app_ctx: PrivateContext = request.app["admin.context"]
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/api/admin.py
```diff
@@ -77,17 +81,20 @@ async def _handle_gql_common(request: web.Request, params: Any) -> ExecutionResu
...
-        None,  # root
-        variable_values=params["variables"],
-        operation_name=params["operation_name"],
-        context_value=gql_ctx,
-        middleware=[
-            GQLLoggingMiddleware(),
-            GQLMutationUnfrozenRequiredMiddleware(),
-            GQLMutationPrivilegeCheckMiddleware(),
-        ],
+    result = await asyncio.wait_for(
```
achimnol, line 84: Now `async with asyncio.timeout(): ...` is the recommended timeout pattern.
https://docs.python.org/3/library/asyncio-task.html#timeouts
Also, `asyncio.wait_for()` raises `TimeoutError` instead of `asyncio.TimeoutError` as of Python 3.11.  Have you checked the timeout handler actually catches the timeout by injecting an arbitrary sleep?

src/ai/backend/manager/api/admin.py
```diff
@@ -110,8 +117,11 @@ async def _handle_gql_common(request: web.Request, params: Any) -> ExecutionResu
     })
 )
 async def handle_gql(request: web.Request, params: Any) -> web.Response:
-    result = await _handle_gql_common(request, params)
-    return web.json_response(result.formatted, status=200)
+    try:
+        result = await _handle_gql_common(request, params)
+        return web.json_response(result.formatted, status=200)
+    except asyncio.TimeoutError:
+        return web.json_response(status=504)
```
achimnol, line 124: We should have a structured error output following `ai.backend.manager.api.exceptions` module.

achimnol (changes requested):

PR #2344, opened 2024-06-27
ci: octodog performs graphql schema dump commit

Yaminyam:
By default, events performed by github-actions-bot do not trigger other actions.
So, after the graphql action was executed, there was a problem where prs could not be merged because the action was not triggered.

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version

PR #2358, opened 2024-07-01
ci: octodog assign pr number to news fragment

Yaminyam:
follow-up #2344 
It was resolved in #1296 pr, but reverted in #1498 pr.


achimnol:
Shouldn't we restore as `secrets.WORKFLOW_PAT` instead of `secrets.OCTODOG`?

Yaminyam:
> Shouldn't we restore as `secrets.WORKFLOW_PAT` instead of `secrets.OCTODOG`?

In the future, our Git automation tasks will use OCTODOG accounts rather than personal PATs.

PR #2372, opened 2024-07-03
feat: Python native WSProxy

kyujin-cho:
This PR introduces new variant of WSProxy, which in turn replaces our nodejs based `local-proxy`. After this PR is merged, developers can now launch local proxy by `backend.ai wsproxy start-server` command - just like we start other Backend.AI core components.    
To start the WSProxy you need to set up a valid configuration file (`wsproxy.toml`). Check out `configs/wsproxy` for detailed configuration directives. 


kyujin-cho:
In the near future we can think of migrating contents of both `wsproxy.api.utils` and `wsproxy.exceptions` to our `common` repository so that it can be reused across other Backend.AI Core components.

PR #2070, opened 2024-04-25
feat: Add `backend.ai plugin scan` command

achimnol:
resolves #1683

Add a new CLI command `backend.ai plugin scan` to inspect the plugin entrypoint scan results.

```console
$ ./backend.ai plugin scan backendai_accelerator_v21
Source       Name    Module Path
-----------  ------  ---------------------------------------
buildscript  cuda    ai.backend.accelerator.cuda_open.plugin
buildscript  mock    ai.backend.accelerator.mock.plugin

$ ./backend.ai plugin scan backendai_accelerator_v20
No plugins found for the entrypoint 'backendai_accelerator_v20'

$ ./backend.ai plugin scan backendai_cli_v10
Source       Name                  Module Path
-----------  --------------------  -------------------------------
buildscript  ag                    ai.backend.agent.cli
buildscript  ag.start-server       ai.backend.agent.server
buildscript  _                     ai.backend.client.cli.main
buildscript  install               ai.backend.install.cli
buildscript  mgr                   ai.backend.manager.cli.__main__
buildscript  mgr.start-server      ai.backend.manager.server
buildscript  plugin                ai.backend.plugin.cli
buildscript  storage               ai.backend.storage.cli
buildscript  storage.start-server  ai.backend.storage.server
buildscript  test                  ai.backend.test.cli.__main__
buildscript  web                   ai.backend.web.cli
buildscript  web.start-server      ai.backend.web.server
```

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version
- [x] Mention to the original issue

Issue #1683, opened 2023-11-02
Improve visibility of plugin detection


achimnol:
`ai.backend.plugin.entrypoint` implements 3 different entrypoint scan sources: 1) package metadata (sharing the same venv), 2) `BUILD` files in the package directory, 3) `setup.cfg` files in a designated directory (e.g., `plugins/`).

Currently there is no way to visually see the scanned results for arbitrary entrypoint group names.

Let's add:
* [x] a CLI command to show the scanned results with the entrypoint provider's package name, path, and the detected source as a table.
* [x] #2359 
  - This will be tracked separately as we need to make it compatible with pex/scie-based installation scenarios.
  - https://pantsbuild.slack.com/archives/C046T6T9U/p1714049948561409

This will help our fieldops members could diagnose plugin recognition issues.

PR #2377, opened 2024-07-04
feat: Scanning plugin entrypoints of external packages

### LINKED ISSUE #2359: Custom Python package loader for the plugin subsystem with scie distributions and local devs
Python has a versatile, highly customizable module loading subsystem.

Example:
image

Let's use them to auto-detect and load plugin packages:

- Local dev setups: scan a designated directory to get the entrypoints and import paths
  - Currently we are doing a similar thing for `plugins` directory in `pants.toml`, but it is just manually parsing(!) `setup.cfg` and `BUILD` files found inside it.
  - I'd like to let it use the standard package entrypoint.
- scie-based distributions: scan the current working directory and/or the location of the executable binary to list `.whl` files and load them with their entrypoints.

This is a prerequisite for the "neo" enterprise TUI installer.

**Potential useful references:**
- https://docs.python.org/3/library/importlib.html
- https://docs.python.org/3/library/zipimport.html
- https://docs.python.org/3/library/site.html
- https://docs.python.org/3/library/modulefinder.html
- https://packaging.python.org/en/latest/specifications/section-installation-metadata/

(cc: @xyloon @adrysn)

jopemachine:
Fix #2359, refs #1683.

## Wheel distributions

This PR implements an automated process to extract `.whl` files found under the `{base-dir}/wheelhouse` directory and add their source directories to `PYTHONPATH`, so that they could be loaded without additional installation steps.

**Example:**

* Download a wheel package into `./wheelhouse` directory.
* Upon scanning, it will be extracted as sub-directories of `./wheelhouse`.
* The package appears in the scan result from the "python-package" source.

> [!TIP]
> To distribute and install plugins that have external dependencies that the core does not use, you may utilize the `pip wheel` command to create a collection of wheel files and put it under the `wheelhouse` directory, as the plugin loader will load all of them into `PYTHONPATH`.

> [!NOTE]
> When a Python wheel file is extracted, it contains:
> - The top-most package directory (For our packages, it is usually `ai`.)
> - `{pkg-name}-{version}.dist-info/`
>   - `entry_points.txt`
>   - `METADATA`
>   - `RECORDS`
>   - `top_level.txt`
>   - `WHEEL`
> - `{pkg-name}.libs/` (empty)
>
> We may simply insert the extracted directory into `sys.path` to make it importable, assuming that all dependencies of the wheel is already available in the venv.

## Plugin source checkouts

It also enhances the `backend.ai plugin scan` command (#2070) to explicitly display the plugin-checkout sources while **skipping** plugin-checkout sources when actually loading the plugins.  Also added more detailed guides such as `[agent].allowed-compute-plugins` checks when the group name contains "accelerator".

You may enable the debug logging of the scan command itself by adding `--debug` option to the `plugin` subcommand, like `./backend.ai plugin --debug scan ...`.

**Example:**

* Clone a private `cuda` plugin into the `./plugins` directory.
* Run `./backend.ai plugin scan backendai_accelerator_v21`:
image

* Run `./py -m pip install -e plugins/cuda`.
* Run `./backend.ai plugin scan backendai_accelerator_v21`:
image

---

----
📚 Documentation preview 📚: https://sorna--2377.org.readthedocs.build/en/2377/

----
📚 Documentation preview 📚: https://sorna-ko--2377.org.readthedocs.build/ko/2377/


jopemachine:
* **#2377** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2377?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

achimnol (changes requested):
Um.. we need to support the scenario for developing plugins by cloning the repositories into `plugins/` directory.

jopemachine:
> Um.. we need to support the scenario for developing plugins by cloning the repositories into `plugins/` directory. Do you have any better idea to it?

Even if `scan_entrypoint_from_plugin_checkouts` is removed from the development environment, each plugin's path under the plugins directory is added to `PYTHONPATH` from `./py`, so it operates the same as before. 

That is, all plugins under the `plugins` directory are still recognized as *plugins*, and we can install and develop plugins in an editable way through the `./scripts/install-plugin.sh` script in the development environment.

achimnol:
I'm also considering to fix scie-based executables suffer from missing imports when executed inside the working copy, while they work fine when executed outside.

achimnol:
> > Um.. we need to support the scenario for developing plugins by cloning the repositories into `plugins/` directory. Do you have any better idea to it?
> 
> Even if `scan_entrypoint_from_plugin_checkouts` is removed from the development environment, each plugin's path under the plugins directory is added to `PYTHONPATH` from `./py`, so it operates the same as before.
> 
> That is, all plugins under the `plugins` directory are still recognized as _plugins_, and we can install and develop plugins in an editable way through the `./scripts/install-plugin.sh` script in the development environment.

This is not true.

image

achimnol:
Okay, I'm going to preserve the plugin-checkout type for debugging purpose (the CLI), while skipping it in the loader mechanism.

achimnol (approved):

PR #2378, opened 2024-07-04
chore(install-dev): block until DB container stabilizes

Yaminyam:
image
When starting codespace, redis-related settings are not being put into etcd.
Since other etcd put requests were processed normally, it is expected that the problem occurred because the put request was executed before etcd was ready.

So, after performing a health check of the halfstack service, proceed to the next step.

image
When using install-dev.sh in the provisioning stage, such as codespace, script execution may be faster than executing docker daemon.
The script runs after the docker daemon's health check.


achimnol (changes requested):
Let's refactor the repeated spin-wait codes as a bash function.

achimnol (approved):

PR #2038, opened 2024-04-21
feat: UNIX socket server to report agent status

kyujin-cho:
This PR allows agent to host a UNIX domain socket which produces up-to-date snapshot of agent's `kernel_registry` and `computers` property. By exposing these informations admins should be able to monitor discrepancies of accelerator allocation status between agent and manager easier. The server can be accessed by `nc` ing socket file (e.g. `nc -U /tmp/backend.ai/ipc/agent-registry-snapshot.sock`).

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version

PR #2392, opened 2024-07-05
fix: Skip waiting cancelled task to shutdown agent properly

fregataa:
follow-up #2038 

Cleanup agent process has not worked properly since cancelled `debug_server_task` is awaited.
`asyncio.Task` should not be awaited if it has already been cancelled.
ref: https://docs.python.org/3.12/library/asyncio-task.html#task-cancellation

### DIFF
--- a/changes/2392.fix.md
+++ b/changes/2392.fix.md
@@ -1,0 +1,1 @@
+Shutdown agent properly by removing a code that waits a cancelled task.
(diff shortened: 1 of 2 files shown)


achimnol (commented):
This is related to the discussion in https://github.com/python/cpython/issues/103486.

Proposing to do something like:
```python
self.debug_server_task.cancel()
await asyncio.sleep(0)
if not self.debug_server_task.done():
    await self.debug_server_task
```

achimnol (approved):

PR #2407, opened 2024-07-09
ci: Invoke the SBOM report generation only when required

achimnol:
Invoke the SBOM report generation workflow when either:
- a tagged commit is pushed for releases. (Since the release workflow depends on the SBOM generation, this condition should be included.)
- a PR is merged.

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -357,6 +357,7 @@ jobs:
       if: always()  # We want the log even on failures.
 
   build-sbom:
+    if: ${{ github.event.pull_request.merged == true || (github.event_name == 'push' && contains(github.ref, 'refs/tags/')) }}
     uses: ./.github/workflows/sbom.yml
 
   make-final-release:
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -357,6 +357,7 @@ jobs:
       if: always()  # We want the log even on failures.
 
   build-sbom:
+    needs: [lint, typecheck, test]
     if: ${{ github.event_name == 'push' && contains(github.ref, 'refs/tags/') }}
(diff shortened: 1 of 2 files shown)


.github/workflows/default.yml
```diff
@@ -357,6 +357,7 @@ jobs:
       if: always()  # We want the log even on failures.
 
   build-sbom:
+    if: ${{ github.event.pull_request.merged == true || (github.event_name == 'push' && contains(github.ref, 'refs/tags/')) }}
```
Yaminyam: ```labeled, unlabeled, opened, synchronize, reopened```
Currently, the default workflow is executed for the above types.
The pull_request.merged value exists only when it is of type 'closed'.
https://docs.github.com/ko/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request_target-workflow-when-a-pull-request-merges

Therefore, similar to the other jobs above, it would be a good idea to execute it only when a tag is pushed.
```suggestion
    if: ${{ github.event_name == 'push' && contains(github.ref, 'refs/tags/') }}
```

.github/workflows/default.yml
```diff
@@ -357,6 +357,7 @@ jobs:
       if: always()  # We want the log even on failures.
 
   build-sbom:
```
Yaminyam, line 359: A release that fails the test cannot be considered a meaningful release.
Therefore, build-sbom and build-wheels also have test as a dependency.
I think sbom is similar to this, so how about having test as a dependency?

Additionally, similar to build-sbom and build-wheels, jobs responsible for creating artifacts required for release can be identified at a glance in the same layer.
```suggestion
  build-sbom:
    needs: [lint, typecheck, test]
```

PR #2411, opened 2024-07-09
fix: Handle all possible exceptions when scheduling single node session

fregataa:
follow-up #643 #2155 
refs https://github.com/lablup/giftbox/issues/691

manager's scheduler does not handle some of `GenericBadRequest` exceptions raised in `_schedule_single_node_session()`, which make status information empty for some pending sessions.


achimnol (approved):

achimnol:
I don't think this will fix all "pending-but-no-reason" cases, but the logic of fix is correct. Thanks!

PR #2419, opened 2024-07-10
feat: support for Pure Storage RapidFiles Toolkit v2

kyujin-cho:
This PR updates implementation of storage proxy to support newest RapidFiles Toolkit version along with older v1 binaries.

By upgrading to RapidFile Toolkit v2, it also resolves a missing support of `-T` option in the `pcp` command which skips subdirectory creation if the target directory already exists, which has caused a malformed vfolder directory structure when performing vfolder clones.


achimnol:
Does this PR include the fix for path manipulation issues when cloning the vfolder?

kyujin-cho:
@achimnol That specific issue has been resolved when upgrading the toolkit itself to v2. `pcp` (or `pcopy` in v2) binary seems to support `-T` flag, which prevents subdirectory from being created when target directory exists. But specifying such flag with v1 `pcp` binary did not achieve the original objective, whereas v2's `pcopy` binary did its job when specifying `T`.

PR #2436, opened 2024-07-12
ci: Change package manager from npm to pnpm

Yaminyam:
https://github.com/lablup/backend.ai-webui/pull/2515
If the above pr is applied, webui will use pnpm as the package manager from now on.
So, in the webui build script part, we also use pnpm.


achimnol:
Please write the news fragment. I will add it.

PR #2456, opened 2024-07-15
feat: Provide extending login session expiration

lizable:
This PR provides extending login session expires timestamp by new API `extend-login-session`.
It only affects when user is already logged in and it's extends based on the time logged in and `max_age`.
Here's the scenario that sets the time to extend login session.

1. Set `login_session_extension_sec` configuration in Webserver.
2. Start Webserver process.
3. Client(User) logins.
4. Set `expires` value(as int) to Redis Storage and store it to session cookie in ISO8601 format string.
    - `expires` value is sum of current time of Redis server and `max_age` value set in Webserver configuration.
6. Client(User) requests extend login session time.
7. Compares `expires` value and the current time of Redis server. If the current time of Redis server is bigger than `expires` value, then update `expires` value in Redis Storage.
8. Same as Step 4.
9. Return `expires` value to requested Client(User).

- [ ] Test case(s) to:
- [ ] Documentation

----
📚 Documentation preview 📚: https://sorna--2456.org.readthedocs.build/en/2456/

----
📚 Documentation preview 📚: https://sorna-ko--2456.org.readthedocs.build/ko/2456/

### DIFF
--- a/src/ai/backend/common/web/session/__init__.py
+++ b/src/ai/backend/common/web/session/__init__.py
@@ -163,7 +163,10 @@ async def update_expires(request: web.Request) -> int:
     if storage is None:
         raise RuntimeError("Install aiohttp_session middleware in your aiohttp.web.Application")
     session: Session = await get_session(request)
-    session.expires = int(storage.expires)
+    if storage.expires is None:
+        await storage.update_expires(session)
+    else:
+        session.expires = int(storage.expires)
     return storage.login_session_extend_time
 
 
--- a/src/ai/backend/common/web/session/__init__.py
+++ b/src/ai/backend/common/web/session/__init__.py
@@ -158,16 +158,19 @@ SESSION_KEY = "aiohttp_session"
 STORAGE_KEY = "aiohttp_session_storage"
 
 
-async def update_expires(request: web.Request) -> int:
+async def update_expires(request: web.Request) -> str:
     storage = request.get(STORAGE_KEY)
     if storage is None:
         raise RuntimeError("Install aiohttp_session middleware in your aiohttp.web.Application")
     session: Session = await get_session(request)
     if storage.expires is None:
         await storage.update_expires(session)
+    if session is not None:
+        expires_timestamp = int(storage.expires) if storage.expires is not None else 0
+        session.expires = expires_timestamp
+        return datetime.fromtimestamp(timestamp=expires_timestamp, tz=timezone.utc).isoformat()
     else:
-        session.expires = int(storage.expires)
-    return storage.login_session_extend_time
+        raise RuntimeError("Session is None")
 
 
 async def get_session(request: web.Request) -> Session:
--- a/src/ai/backend/common/web/session/__init__.py
+++ b/src/ai/backend/common/web/session/__init__.py
@@ -4,8 +4,9 @@ import logging
 import logging.config
 import sys
 import time
+from collections.abc import MutableMapping
 from datetime import datetime, timezone
-from typing import Any, Awaitable, Callable, Dict, Iterator, MutableMapping, Optional, Union, cast
+from typing import Any, Awaitable, Callable, Dict, Iterator, Optional, Union, cast, override
 
 import trafaret as t
 from aiohttp import web
@@ -41,6 +42,7 @@ class _CookieParams(TypedDict, total=False):
 
 class SessionData(TypedDict, total=False):
     created: int
+    expiration_dt: int
     session: Dict[str, Any]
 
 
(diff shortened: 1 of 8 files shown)


src/ai/backend/common/web/session/__init__.py
```diff
@@ -98,6 +101,21 @@ def created(self) -> int:
...
+    def expires(self) -> Optional[str]:
+        return self._expires
+
+    @expires.setter
+    def expires(self, value: Optional[str]) -> None:
+        if value is None and self.max_age is not None:
+            expires_from_now = datetime.fromtimestamp(
+                time.time() + int(self.max_age), tz=timezone.utc
+            )
+            self._expires = expires_from_now.strftime("%a, %d-%b-%Y %H:%M:%S GMT")
```
achimnol: Could we use locale-aware datetime representation?
Or, could we use some standard format like ISO8601 and let the frontend convert it to a locale-aware representation?
Any new string messages generated in the server-side should be aware of localization and transformation in the frontend side.

achimnol (changes requested):

src/ai/backend/common/web/session/__init__.py
```diff
@@ -146,6 +158,20 @@ def __delitem__(self, key: str) -> None:
...
+    if storage is None:
+        raise RuntimeError("Install aiohttp_session middleware in your aiohttp.web.Application")
+    session: Session = await get_session(request)
+    await storage.update_expires(session)
+    if session is not None:
+        expires_timestamp = int(storage.expires) if storage.expires is not None else 0
+        session.expires = expires_timestamp
+        return datetime.fromtimestamp(timestamp=expires_timestamp, tz=timezone.utc).isoformat()
+    else:
+        raise RuntimeError("Session is None")
```
fregataa: `get_session()` ensures that returned value is not null.
is there any possibility of the returned `session` is `None`?

src/ai/backend/common/web/session/__init__.py
```diff
@@ -321,7 +358,7 @@ def __init__(
         super().__init__(
             cookie_name=cookie_name,
             domain=domain,
-            max_age=max_age,
+            # max_age=max_age,
```
fregataa: I guess this should be deleted

src/ai/backend/common/web/session/redis_storage.py
```diff
@@ -20,13 +20,18 @@
 class RedisStorage(AbstractStorage):
     """Redis storage"""
 
+    @property
+    def login_session_extend_time(self) -> Optional[int]:
+        return self._login_session_extend_time
```
fregataa: This name is confusing. It is hard to understand that this variable is used to be as a flag that allows to update expiration time or is used as a default value to extend the expiration time or a default value of max age of sessions.
Let's give a better name to it

src/ai/backend/common/web/session/__init__.py
```diff
@@ -35,7 +36,7 @@ class _CookieParams(TypedDict, total=False):
     secure: Optional[bool]
     httponly: bool
     samesite: Optional[str]
-    expires: str
+    expires: Optional[str]
```
fregataa, line 40: This name is also confusing.
It is hard to understand the usage of `expires` and `login_session_extend_time`.

src/ai/backend/common/web/session/__init__.py
```diff
@@ -146,6 +158,20 @@ def __delitem__(self, key: str) -> None:
 STORAGE_KEY = "aiohttp_session_storage"
 
 
+async def update_expires(request: web.Request) -> str:
+    storage = request.get(STORAGE_KEY)
+    if storage is None:
+        raise RuntimeError("Install aiohttp_session middleware in your aiohttp.web.Application")
+    session: Session = await get_session(request)
+    await storage.update_expires(session)
```
fregataa: A `RedisStorage` object updates expiration time of a session because the expiration time is calculated by Redis.
How about letting the storage calculate time to extend and implements `update_expires()` (or `update_expiration_time()`) in Session class?
Please give me any comment if you want and I think I can update this part of codes.

fregataa (changes requested):
Left some comments.
Mainly, it is about naming and some questions about how the expiration time and extending it works

src/ai/backend/common/web/session/__init__.py
lizable: Bulls eye!

src/ai/backend/common/web/session/__init__.py
lizable: I added conditional statement in order to pass the mypy checks, but I'll fix it.

src/ai/backend/common/web/session/redis_storage.py
lizable: Good point. The usage of this values is to store the amount of expiration time of login session cookie. How about login_session_extend_time_in_seconds ?

src/ai/backend/common/web/session/__init__.py
lizable: expires is as mentioned in [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires), the date/time of response is expired. Therefore I don't think it need to be changed.

fregataa (approved):

PR #2471, opened 2024-07-15
fix: typos in install and concepts docs

why-arong:
This PR corrects several typos in the install and concepts sections of the documentation.

In addition to correcting typos, I made changes to improve clarity:

> Make sure every mission critical workloads are shut down **`when`** performing a rolling upgrade.
> -> 
> Make sure every mission critical workloads are shut down **`before`** performing a rolling upgrade.
> 

Please double-check the changes for accuracy.

### DIFF
--- a/docs/concepts/computing.rst
+++ b/docs/concepts/computing.rst
@@ -27,7 +27,7 @@ Kernel roles in a cluster session
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 In a cluster session with multiple kernels, each kernel has a role.
-By default, the first container takes the "main" role while others takes the "sub" role.
+By default, the first container takes the "main" role while others take the "sub" role.
 All kernels are given unique hostnames like "main1", "sub1", "sub2", ..., and "subN" (the cluster size is N+1 in this case).
 A non-cluster session has one "main1" kernel only.
 
@@ -91,7 +91,7 @@ Interactive compute session
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Interactive compute sessions are used to run various interactive applications and development tools,
-such as Jupyter Notebooks, web-based terminals, and etc.
+such as Jupyter Notebooks, web-based terminals, etc.
 It is expected that the users control their lifecycles (e.g., terminating them)
 while Backend.AI offers configuration knobs for the administrators to set idle timeouts with various criteria.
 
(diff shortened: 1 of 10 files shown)


docs/concepts/services.rst
```diff
@@ -11,9 +11,9 @@ Manager and Webserver
 ^^^^^^^^^^^^^^^^^^^^^
 :raw-html-m2r:`<span style="background-color:#fafafa;border:1px solid #ccc;display:inline-block;width:16px;height:16px;margin:0;padding:0;"></span>`
 
-Backend.AI manager is the central governor of the cluster.
+Backend.AI manager is the central governor of  the cluster.
```
rapsealk:
```suggestion
Backend.AI manager is the central governor of the cluster.
```

why-arong: thanks!

docs/concepts/computing.rst
```diff
@@ -91,7 +91,7 @@ Interactive compute session
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Interactive compute sessions are used to run various interactive applications and development tools,
-such as Jupyter Notebooks, web-based terminals, and etc.
+such as Jupyter Notebooks, web-based terminals, etc.
```
Yaminyam: When writing multiple numbers in English, I think it is common to write and at the end.
The comma before and may or may not be indicated, but I have seen more cases where it is not present.

why-arong: I thought "and etc." was redundant, so I corrected it. However, I have seen many documents using "and etc" also. If you think my correction was unnecessary, I'll revert it back.

why-arong: I fixed it!

Yaminyam: It seems better to follow @YEONFEEL96 opinion on this!

docs/install/install-from-source.rst
```diff
@@ -97,4 +97,4 @@ Configuring Overlay Networks for Multi-node Training (Optional)
 Currently the cross-node inter-container overlay routing is controlled via Docker Swarm's overlay networks.
 In the manager, you need to `create a Swarm <https://docs.docker.com/engine/swarm/swarm-tutorial/create-swarm/>`_.
 In the agent nodes, you need to `join the Swarm <https://docs.docker.com/engine/swarm/swarm-tutorial/add-nodes/>`_.
-Then restart all manager and agent daemons to make it working.
+Then restart all manager and agent daemons to make it work.
```
achimnol: `make A to {verb}` or `make A {adjective}`.
It should be a gerund (동명사) "working".

achimnol (changes requested):

docs/install/install-from-source.rst
why-arong: I fixed it!

PR #2480, opened 2024-07-16
feat: add tests for _ensure_slot method in PortFrontend

why-arong:
### Summary

This PR adds unit tests for the `_ensure_slot` method in the `PortFrontend` class. The tests cover the following scenarios:
- Unregistered port
- No circuit available for the registered port
- Successful handling with proper circuit and backend setup

---

The `_ensure_slot` function in the PortFrontend class is designed to ensure that a specific port has an associated circuit and backend. If the port is not registered or the circuit is not available, it raises an error. If everything is valid, it adds the circuit and backend to the request object.

### Changes
- Created `DummyRequest` class to simulate `aiohttp.web.Request` object.
- Added `test_ensure_slot_unregistered_port`.
- Added `test_ensure_slot_no_circuit`.
- Added `test_ensure_slot_success`.

### Checklist
- [x] Test case(s) to:
  - [ ] Demonstrate the difference of before/after
  - [ ] Demonstrate the flow of abstract/conceptual models with a concrete implementation


why-arong:
I don't think there is a wsproxy related label, should I do it with approxy?

achimnol:
> I don't think there is a wsproxy related label, should I do it with approxy?

Yes! They are the same thing. Historically, we used to call it "websocket proxy", but now we are moving to a more generic name, "app proxy".

tests/BUILD
```diff
@@ -1 +0,0 @@
-python_test_utils()
```
kyujin-cho: Could you tell me more about the removal of this file?

tests/wsproxy/conftest.py
```diff
@@ -0,0 +1,72 @@
...
+    def __setitem__(self, key: str, value: Any):
+        self._state[key] = value
+
+
+@pytest.fixture
+def dummy_request():
+    def _dummy_request(app_data: Dict[str, Any]):
+        return DummyRequest(app_data)
+
+    return _dummy_request
```
kyujin-cho: Unlike other fixtures this returns a function which just initializes an instance without any runtime modifications - isn't it better to declare a new helper function which does the same job with old inner function and use it directly from the test case?

kyujin-cho (changes requested):

tests/wsproxy/conftest.py
why-arong: - Removed specific fixtures from conftest.py, leaving only the general **create_circuit** fixture. 
(They can be moved back if needed in the future.)

- Modified the `create_circuit` function to accept more general parameters, allowing it to create a `Circuit` with customizable values. 
(Future test cases can pass the appropriate arguments based on their specific requirements.)

kyujin-cho: I still don't recognize any benefits of leaving `create_circuit` as a fixture populating an inner function. What would be the upsides of doing so instead of just exposing inner function itself as a role of helper function?

why-arong: The `create_circuit` function is currently only used in `test_port.py`, so I will include it there for now.

If this function is needed in other test code in the future, we can move it to `utils.py`. (However, if it’s added under the tests directory, we'll need to address the **PYTHONPATH** issue!)

Issue #2300, opened 2024-06-17
Let containers use the host timezone config


achimnol:
Let's read-only mount /etc/localtime and /etc/timezone into the containers if present to keep the timezone config in sync with the host.

Good first issue for summer interns.

JIRA Issue: BA-117

PR #2503, opened 2024-07-18
feat: mount host's /etc/localtime and /etc/timezone

ggstargame45:
**Feature**
This is a follow up of https://github.com/lablup/backend.ai/issues/2300.
When the agent executes create_kernel(*), it mounts the /etc/localtime and /etc/timezone to the container if the host system platform is linux-based.

**Limitations** 
- It is valid only if the host OS is a linux-based system. 
Mac OS doesn't have a /etc/localtime to mount (Has to neither access the system variable nor /usr/share/zoneinfo manualy)
- The sessions that are made beforehand will not get affected.


why-arong:
Labels such as size:~ and self-assignments do not need to be manually added, as they are handled automatically! 😆

ggstargame45:
> You don't need to manually add size labels or do self-assignments as they are handled automatically! 😆
 
I'll do it next time. Thx!

PR #2506, opened 2024-07-18
fix: Check null value of user mutation by `Undefined`

fregataa:
`modify_user` mutation checks `is_active` input value.
All input values that client does not specify get `Undefined` sentinel value rather than `None`.
Let's correct this check logic.


kmkwon94 (approved):
LGTM!

PR #2511, opened 2024-07-18
feat: update selected_route method and add scenario tests

why-arong:
### Background
```python
for ... # Original code
    ceiling = ranges[0]
    if (i == 0 and rand < ceiling) or (ranges[i - 1] <= rand and rand < ceiling):
        selected_route = routes[i]
        break
else:
	...
```
[was intended to distribute traffic based on traffic_ratio.](https://github.com/lablup/backend.ai/blob/9140bed4d7ae578709152cb128c24fa8b1832dc5/src/ai/backend/wsproxy/proxy/backend/http.py#L28C5-L53C30) However, ceiling = ranges[0] was fixed, causing an issue where, for example, if ranges[0] = 0.4 and rand = 0.5, the code would always fall into the else clause, selecting the last route regardless of other ratios.

### Changes 
1. **Refactored `selected_route` Method:**

2. **Added Comprehensive Scenario Tests:**

## **Checklist:** 

- [ ] Test case(s) to:

### DIFF
--- a/src/ai/backend/wsproxy/proxy/backend/http.py
+++ b/src/ai/backend/wsproxy/proxy/backend/http.py
@@ -44,8 +44,8 @@ class HTTPBackend(AbstractBackend):
                 ranges.append(ratio_sum)
             rand = random.random() * ranges[-1]
             for i in range(len(ranges)):
-                ceiling = ranges[0]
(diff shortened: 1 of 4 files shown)


HyeockJinKim:
wsproxy is deprecated.

PR #2514, opened 2024-07-19
ci: Makes if conditions written in multi-line work properly

Yaminyam:
Refactored the syntax with multiple if conditions to multi-line in #2499 to make the code more readable.
However, when using `${{ }}` as multi-line in github actions does not work correctly.
To use a grammar like `!` in an if condition, it would conflict with the existing keyword in YAML, so originally the `${{ }}` grammar, but using multi-line grammars like `|` or `>` is fine because all subsequent values are treated as literal.

To make sure it is labeled skip:ci, the kind of event must be pull_request.
So the ci is not being skipped in the case of ci fired by push trigger, and it is not merged in the case of required.
And there is also a problem that if you push a commit to the pr branch, the pull_request trigger and push trigger occur at the same time, so the ci is performed twice.
So we make sure that the push trigger is only executed on the main branch and release branch, which performs the commit directly without PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
 
 
   lint:
-    if: >
+    if: |
       !(contains(github.event.pull_request.labels.*.name, 'skip:ci') || needs.optimize-ci.outputs.skip == true)
       && !contains(fromJSON('["flow:merge-queue", "flow:hotfix"]'), github.event.label.name)
       && github.event.pull_request.merged == false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,12 +26,11 @@ jobs:
 
 
   lint:
-    if: |
-      ${{ !contains(github.event.pull_request.labels.*.name, 'skip:ci')
-        && !contains(fromJSON('["flow:merge-queue", "flow:hotfix"]'), github.event.label.name)
-        && github.event.pull_request.merged == false
-        && needs.optimize-ci.outputs.skip == false
-      }}
+    if: >
+      !contains(github.event.pull_request.labels.*.name, 'skip:ci')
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -200,11 +204,10 @@ jobs:
 
   test:
     if: |
-      ${{ !contains(github.event.pull_request.labels.*.name, 'skip:ci')
-        && !contains(fromJSON('["flow:merge-queue", "flow:hotfix"]'), github.event.label.name)
-        && github.event.pull_request.merged == false
-        && needs.optimize-ci.outputs.skip == false
-      }}
+      !contains(github.event.pull_request.labels.*.name, 'skip:ci')
```
achimnol: Missing update for the condition check.

.github/workflows/ci.yml
```diff
@@ -102,11 +106,10 @@ jobs:
 
   check-alembic-migrations:
     if: |
-      ${{ contains(github.event.pull_request.labels.*.name, 'require:db-migration')
-        && !contains(fromJSON('["flow:merge-queue", "flow:hotfix"]'), github.event.label.name)
-        && github.event.pull_request.merged == false
-        && needs.optimize-ci.outputs.skip == false
-      }}
+      contains(github.event.pull_request.labels.*.name, 'require:db-migration')
```
achimnol, line 109: Need to resolve the race issue with the auto-labeler workflow.
Let's embed the auto-labeler job in this workflow.

achimnol (changes requested):

.github/workflows/ci.yml
achimnol: This will be handled in a separate PR.

achimnol (approved):

PR #2525, opened 2024-07-22
fix: Delete quota scope when deleting vfolder

fregataa:
There are a few storage backends that require deleting quota scope before deleting a path of a vfolder, such as GPFS.
Let's call `AbstractQuotaModel.delete_quota_scope()` when deleting vfolder.

> [!NOTE]
> Calling `AbstractQuotaModel.delete_quota_scope()` raises `NotEmptyError` if the vfolder is not empty. We should empty the folder before deleting quota scope.

### DIFF
--- a/changes/2525.fix.md
+++ b/changes/2525.fix.md
@@ -1,0 +1,1 @@
+Delete quota scope when deleting vfolder.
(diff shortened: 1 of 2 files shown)


fregataa:
Close this PR since it is wrong

PR #2529, opened 2024-07-22
feat: Impl Resource sync interface in agent side

fregataa:
Agents's sync-and-get-kernels() API
The API that synchronizes agent's kernels to kernel information specified by API parameters (preparing_kernels, pulling_kernels, running_kernels, terminating_kernels). It assumes that the kernel information given by the parameter is the "truth".
If any of kernel information mismatch between kernel_registry and running_kernels(or terminating_kernels), agent injects termination event to terminate the kernel.

sync-and-get-kernels() API returns actual { running, terminating, terminated } kernels (which is not used for now). actual_terminated_kernels contains terminated kernels specified as running_kernels by API parameter.

### DIFF
--- a/src/ai/backend/common/types.py
+++ b/src/ai/backend/common/types.py
@@ -1263,3 +1263,29 @@ MODEL_SERVICE_RUNTIME_PROFILES: Mapping[RuntimeVariant, ModelServiceProfile] = {
     ),
     RuntimeVariant.CMD: ModelServiceProfile(name="Predefined Image Command"),
 }
+
+
+@dataclass
+class KernelStatusCollection(JSONSerializableMixin):
+    KernelTerminationInfo = tuple[KernelId, str]
+
+    actual_existing_kernels: list[KernelId]
+    actual_terminating_kernels: list[KernelTerminationInfo]
+    actual_terminated_kernels: list[KernelTerminationInfo]
+
+    def to_json(self) -> dict[str, list[KernelId]]:
+        return dataclasses.asdict(self)
+
+    @classmethod
+    def from_json(cls, obj: Mapping[str, Any]) -> KernelStatusCollection:
+        return cls(**cls.as_trafaret().check(obj))
+
+    @classmethod
+    def as_trafaret(cls) -> t.Trafaret:
+        from . import validators as tx
+
+        return t.Dict({
+            t.Key("actual_existing_kernels"): tx.ToList(tx.UUID),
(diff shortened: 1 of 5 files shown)


fregataa:
* **#2530** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2530?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2529** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2529?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/ai/backend/common/validators.py
```diff
@@ -711,6 +711,19 @@ def check_and_return(self, value: Any) -> set:
             self._failure("value must be Iterable")
 
 
+class ToList(t.List):
```
achimnol, line 714: Could we just reuse [`trafaret.base.List`](https://github.com/Deepwalker/trafaret/blob/4ac577155c93e5a3b024645abd64b1442fd4d550/trafaret/base.py#L876)?

src/ai/backend/common/types.py
```diff
@@ -1279,3 +1279,29 @@ class ModelServiceProfile:
     ),
     RuntimeVariant.CMD: ModelServiceProfile(name="Predefined Image Command"),
 }
+
+
+@dataclass
+class SimpleKernelRegistry(JSONSerializableMixin):
```
achimnol: Since this does not manage the states but is just a collection of serializable information about the kernels, how about renaming it like a more data-structure-like one? e.g., `KernelRecords`, `KernelStatusCollection`, ...

achimnol (changes requested):
Some early reviews!

src/ai/backend/common/types.py
fregataa: I want to allow any iterable data type here since sequential data fetched from RPC response get deserialized into `tuple`.

src/ai/backend/common/validators.py
fregataa: [`trafaret.base.List`](https://github.com/Deepwalker/trafaret/blob/4ac577155c93e5a3b024645abd64b1442fd4d550/trafaret/base.py#L876) allows list type data only and I want to allow any iterable data type here since sequential data fetched from RPC response get deserialized into `tuple`.

PR #2550, opened 2024-07-24
feat: ensure correct nodejs version and add pyenv environment check

### LINKED ISSUE #2526: Verify that the node.js version is correct in `install-dev.sh`
### Main idea

https://github.com/lablup/backend.ai/blob/19622222d4938eb3031f62f8cb745a0e29d30c6c/scripts/install-dev.sh#L489Similar to install_node, let's make sure that node.js is not only installed, but also that the correct version is installed.

### Alternative ideas

_No response_

### Anything else?

_No response_

why-arong:
Fixes #2526

#### correct nodejs version

This PR addresses an issue where the script did not check if the currently installed Node.js version matches the required version. 

If Node.js was already installed but with a different version, the script would not update it to the required version. 

I have written logic to check if the Node version is `exactly a specific version`, not just above a certain version. Let me know if you find anything wrong with this!

---

####  pyenv enviornment check
The part of setting environment variables was refactored into a function and generalized for future reuse!
(add "config/fish/config.fish" for fish shell user too!)
I also eliminated the possibility of duplicating the script_content by using grep.
```shell
append_to_profiles() {
  script_content="$1"
  PROFILE_FILES=("zshrc" "bashrc" "profile" "bash_profile" "config/fish/config.fish")

  for PROFILE_FILE in "${PROFILE_FILES[@]}"; do
    FULL_PATH="${HOME}/.${PROFILE_FILE}"

    if [ -e "$FULL_PATH" ] && ! grep -qF "$script_content" "$FULL_PATH"; then
      echo "$script_content" >>"$FULL_PATH"
    fi
  done
}
```

and I considered a case where pyenv is installed but the environment variables are not set.

```shell
if ! type "pyenv" >/dev/null 2>&1 && [ ! -d "$HOME/.pyenv" ]; then
```

### 

**Checklist:** 


Yaminyam (approved):

HyeockJinKim:
I will close this PR because it has been applied with another PR.

PR #2554, opened 2024-07-25
docs: Add a dollar sign to indicate shell script

Yaminyam:

### DIFF
--- a/docs/install/install-from-package/os-preparation.rst
+++ b/docs/install/install-from-package/os-preparation.rst
@@ -259,7 +259,7 @@ you may pull the following starter images:
 
 .. code-block:: bash
 
-   docker pull cr.backend.ai/stable/filebrowser:21.02-ubuntu20.04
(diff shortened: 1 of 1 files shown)


achimnol:
Um... this is intentional.

There are two code-block types:
- `shell`, `bash`: The shell script code or the command snippet.
- `console`: The log of alternating occurrences of commands and their outputs, distinguishing those by the presence of the prompt sign like dollar.

I used to use `console` with the prompts, but now I'm inclined to use `shell` or `bash` when I omit the command outputs but just list the commands, so that they can be copy-and-pasted at once.

This is a `shell` block:
```shell
docker run blah-blah $myvar
hello world!  # this is treated as a part of the shell code
```

This is a `console` block:
```console
$ docker run blah-blah $myvar \
>    --extended-lines \
>    --oops
hello world!  # highlighted as plain text
```

PR #2555, opened 2024-07-25
fix: rename Image Filter to Image Type

### LINKED ISSUE #2535: Refine namings of `images()`, `customized_images()` GraphQL queries
Before it's too late, change as follows:

* `filter` argument of `images()` (already there) and `user_created_images()` (to be added):
  - **Every `filter` argument should be [a queryfilter expression.](https://github.com/lablup/backend.ai/blob/2b81b796392c41586e2101d0d5a5f48b58158856/src/ai/backend/manager/models/minilang/queryfilter.py#L16-L43)**
  - To add other high-level filters, use specific names.
    - `filter` of `images()` is a list of [`ImageLoadFilter` enum constants](https://github.com/lablup/backend.ai/blob/3db5a0f574074a5afcb4009b53b3c7e88e181256/src/ai/backend/manager/models/image.py#L94).
    - In this case, we could use `types` as an alternative name as it is used to categorize the source of images ("CUSTOMIZED" for user-created, "OPERATIONAL" for system-provided).

### Additional information:

* Clarification of the naming of `customized_images()`
  - _custom_: built for the user-specific requirements from the beginning
  - _customized_: modified for the user-specific requirements → This applies here since it's the user's own customization on existing container images.
  - In this case, the subject of customization is the user.
  - We also need to consider there will be project-level customized images in the future (#1906).

kyujin-cho:
Resolves #2535. Ideas what we mentioned as `Image Filter` will now be called `Image Type` to avoid conflict between `filter` concept of GraphQL queries.


src/ai/backend/manager/models/image.py
```diff
@@ -84,14 +84,14 @@
...
-class PublicImageLoadFilter(enum.StrEnum):
+class PublicImageLoadType(enum.StrEnum):
     OPERATIONAL = "operational"
     """Include operational images."""
     CUSTOMIZED = "customized"
     """Include customized images owned or accessible by API callee."""
 
 
-class ImageLoadFilter(enum.StrEnum):
+class ImageLoadType(enum.StrEnum):
```
achimnol: Why not just `ImageType`?

kyujin-cho: `ImageType` is already being used [here](https://github.com/lablup/backend.ai/blob/4446f594d654febde7c2481cc3bbb653b7d59385/src/ai/backend/manager/models/image.py#L167-L171).

kyujin-cho:
Since there are no better name to describe the characteristic, I and @achimnol have made a conclusion to leave the name `ImageLoadFilter` as is, and instead add a comment describing the intent and limits of itself to the Enum implementation.

achimnol:
`ImageLoadFilter` is a kind of predefined set of filter conditions, and it will be combined with user-provided custom (query) `filter` expression via "AND".

PR #2570, opened 2024-07-26
fix: skip resolving malformed `ModelCard` GQL item

fregataa:
`ModelCard` GQL query resolver returns a model card object with `error_msg` field if the client is the creator or is admin. Else, returns model card with `id`, `row_id` and `name` fields.


agatha197 (approved):
LGTM

PR #2572, opened 2024-07-26
fix: Silent failure of `DockerAgent`'s `push_image()`, `pull_image()`

jopemachine:
This PR modifies the behavior of the `DockerAgent` so that when an error occurs during the execution of `push_image()` or `pull_image()`, it does not fail silently but instead informs the user that an issue has occurred.

## Test 

Tested manually.
For example, when incorrect Docker credentials are used, it will raise an error like below.

```
❯ ./backend.ai session convert-to-image 7a6b642b-22f4-42fd-b6fb-bbdf2a818ba3 new-img
∙ Request to commit Session(name or id: 7a6b642b-22f4-42fd-b6fb-bbdf2a818ba3)
 33%|███████████████████████████████████████████████████████████                                                                                                                      | 1/3 [00:02<00:04,  2.01s/it]
✗ Error during the operation: <BackendError(-1): General Backend API Error. (RuntimeError("Failed to push image: unauthorized: HTTP Basic: Access denied. If a password was provided for Git authentication, the password was incorrect or you're required to use a token instead of a password. If a token was provided, it was either incorrect, expired, or improperly scoped. See https://gitlab.com/help/user/profile/account/two_factor_authentication.md#troubleshooting"))>
```

- [x] Test case(s) to:


jopemachine:
* **#2572** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2572?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

jopemachine:
In addition to modifying the `push_image` code, it appears that other aiodocker API calls, such as `pull_image`, also need to be revised.

jopemachine:
In this PR, I found several incorrect return types in `aiodocker`, which led me to create the PR below. 
https://github.com/aio-libs/aiodocker/pull/909.

Until the above PR is merged, I’ll manually cast these types to `list` and leave TODO comments as reminders.

src/ai/backend/agent/docker/agent.py
```diff
@@ -1487,7 +1487,11 @@ async def push_image(self, image_ref: ImageRef, registry_conf: ImageRegistry) ->
             }
 
         async with closing_async(Docker()) as docker:
-            await docker.images.push(image_ref.canonical, auth=auth_config)
+            # TODO: Remove this useless type casting after resolving https://github.com/aio-libs/aiodocker/pull/909.
+            result = cast(list, await docker.images.push(image_ref.canonical, auth=auth_config))
+
+            if error := result[-1].get("error"):
+                raise RuntimeError(f"Failed to push image: {error}")
```
HyeockJinKim: `docker query` won't return an empty list, right?

jopemachine: I think instead of making assumptions about how this API work, it would be better to add exception handling.


```py
            if not result:
                raise RuntimeError("Failed to push image: unknown error")
            elif error := result[-1].get("error"):
                raise RuntimeError(f"Failed to push image: {error}")
```

Thank you for the review!

src/ai/backend/agent/docker/agent.py
```diff
@@ -1487,7 +1487,13 @@ async def push_image(self, image_ref: ImageRef, registry_conf: ImageRegistry) ->
             }
 
         async with closing_async(Docker()) as docker:
-            await docker.images.push(image_ref.canonical, auth=auth_config)
+            # TODO: Remove this useless type casting after resolving https://github.com/aio-libs/aiodocker/pull/909.
+            result = cast(list, await docker.images.push(image_ref.canonical, auth=auth_config))
+
+            if not result:
+                raise RuntimeError("Failed to push image: unknown error")
```
HyeockJinKim: It seems better to indicate that an unexpected result was returned rather than an unknown error.

HyeockJinKim (approved):
lgtm

Issue #411, opened 2022-04-08
LocalDockerDaemon container registry type for ease of development/single-node setup


achimnol:
Currently we need to explicitly configure the container registries and perform 'rescan' of them to get the image metadata and either pull the images to the agent nodes manually or let the session creation process pull automatically.

**Other container registries:**

```mermaid
flowchart TD
	subgraph manager-node
		Manager
	end
	CR[Container Registry]
	subgraph agent-node
		Agent
	end
    Manager -->|rescan| CR
    Agent -->|pull images| CR
```

**LocalDockerDaemon container registry:**

```mermaid
flowchart TD
	subgraph all-in-one-node
		Manager
		CR[local Docker daemon]
		Agent
	end
    Manager -->|rescan| CR
    Agent -.->|just use images| CR
```

In single-node / development setup, it might be better to make the manager to auto-detect any locally available Docker images and self-rescan them, so that we could just pull (or even build) the image to test it.

This may be also useful for building a CI/CD for our image builds because a single-node setup of Backend.AI cluster could be used to run an integration test of newly built container images, with less efforts to setup a separate registry and rescan it.

- [ ] Add a local config option (boolean flag) whether to use this feature or not: `manager.enable-local-docker-images` in `manager.toml`\* In production/multi-node setups, this should be disabled.

- [ ] Add a new container registry backend `LocalDockerDaemonRegistry` which uses the docker API to enumerate and fetch the images and their archs/labels from the local Docker daemon.
- [ ] Trigger rescanning with this registry when:\* The manager starts up.
- The manager receives an image pull event from the local docker daemon.\* This could be implemented as a cleanup context manager in `manager/server.py` to spawn and shutdown a separate async task to monitor Docker daemon events (like in agent).
- The Docker daemon event fetch loop could be refactored and migrated from the agent to the common.

JIRA Issue: BA-270

achimnol:
Let's revisit this issue after #1917, by making the local registry as a `container_registries` table record and reuse all existing code paths, instead of the special `--local` option implemented in #1089.

Issue #293, opened 2021-09-30
Invalid image tag completely blocks the new session creation due to the failure of fetching the image list


adrysn:
There is an [internal rule](https://github.com/lablup/backend.ai-common/blob/34c3442ed97a3e3f6c3cae393e91f55c34f3e83d/src/ai/backend/common/docker.py#L151) in the format of an image tag. If an image with a tag violates the rule, the image is considered to have an "invalid tag".

The problem is when there are image metadata in Etcd with an invalid tag, the request from a client that fetches the available image list raises an error, without returning any image list. This means that users cannot create any compute session due to single image metadata with an invalid tag.

This confuses many customers who build their own images. I think we at least need to return the normal image list even if there are some mistagged images.

### How to reproduce the problem.

* `cd` into your Manager directory.
* Save the contents below as `image_with_invalid_tag.json`.
```json
{
    "": "1",
    "21.08-tf2-tf231_py36_backend": {
        "": "sha256:5baf2f180a51dac1bb6938c733bdac1cab55daa8791b6a0359c6d5b8ecdf1837",
        "labels": {
            "ai.backend.base-distro": "ubuntu16.04",
            "ai.backend.envs.corecount": "OPENBLAS_NUM_THREADS,OMP_NUM_THREADS,NPROC",
            "ai.backend.features": "batch query uid-match user-input",
            "ai.backend.kernelspec": "1",
            "ai.backend.resource.min.cpu": "1",
            "ai.backend.resource.min.cuda.device": "1",
            "ai.backend.resource.min.cuda.shares": "0.1",
            "ai.backend.resource.min.mem": "1g",
            "ai.backend.runtime-path": "/usr/local/bin/python3",
            "ai.backend.runtime-type": "python",
            "ai.backend.service-ports": "jupyter:http:8080"
        },
        "resource": {
            "cpu": {
                "min": "1"
            },
            "cuda.device": {
                "min": "1"
            },
            "cuda.shares": {
                "min": "0.1"
            },
            "mem": {
                "min": "1g"
            }
        },
        "size_bytes": "7284871419"
    }
}
```
* Insert the image information into Etcd. It doesn't matter the registry exists or not.
```bash
backend.ai mgr etcd put-json images/10.20.30.10%3A7080/lablup%2Fimage-with_invalid-tag image_with_invalid_tag.json
```
* Check if the information is inserted correctly.
```bash
[...]

PR #2582, opened 2024-07-27
feat: Allow running images without explicit metadata labels

### LINKED ISSUE #758: Reduce the required information to import a new container image
(initial idea by @hephaex)

To streamline support for arbitrary Docker/container images, let's reduce the required information to import a container image to Backend.AI.

Currently, there are three ways to import a container image:
1. Manually write a Dockerfile to add required `ai.backend.*` labels and copy service-definition files.
   ([Refer the specification](https://docs.backend.ai/en/latest/dev/adding-kernels.html))
2. Use [the image import API in the manager](https://github.com/lablup/backend.ai/blob/main/src/ai/backend/manager/api/image.py)
3. Use the Forklift service

In all cases, users need to provide [many label options](https://docs.backend.ai/en/latest/dev/adding-kernels.html#metadata-labels).

Since 2019, almost all our customers are relying on the service ports (aka container applications) and startup commands of the batch-type sessions, rather than the REPL-based code execution (usually used by `backend.ai run ...` CLI commands).

So, let's simplify the kernel label specification.

Suggestions:
* Introduce `ai.backend.kernelspec` version 2.
  - Remove all `ai.backend.resource.*`.
  - Remove `ai.backend.runtime-type` by introducing the unified kernel runner which only supports running service ports and the startup script of batch-type sessions.
  - Remove `ai.backend.features` and assume `uid-match` by default.
  - Replace `ai.backend.base-distro` with `ai.backend.base-platform` which follows [the Python's wheel platform tag specification](https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/) such as "manylinux". (See also: [PEP-600](https://peps.python.org/pep-0600/))
    - This could be automatically discovered from the contents of the imported image.
  - Replace `ai.backend.runtime-path` with `ai.backend.python.base` to specify the base path of the image-provided Python runtime.
    - Let's specify only the `sys.prefix` (e.g., `/usr/local`) instead of the full path to the Python executable (e.g., `/usr/local/bin/python`)
    - We may add `ai.backend.{other-language}.base` in the same fashion in the future, which may coexist with the base paths of multiple language runtimes in a single image.
  - Preserve `ai.backend.accelerators`, `ai.backend.service-ports` as-is.
* Assume a default `policy.yml` and avoid requiring copying it for all images.
* Service definitions remain same.
* Automate population of the image-provided bootstrap script
  - NOTE: This may not be possible for "any arbitrary" image, but at least we could try full automation for NGC-based images.
  - Let's _reuse_ the image-provided entrypoint and/or command by passing it to our `entrypoint.sh`.
  - We need to determine whether to run it in the background or foreground.  Maybe we could detect it by whether the image has exposed port numbers or not.
* Make the batch-type session default in `backend.ai run` and avoid reusing existing sessions, which is a legacy from CodeOnWeb.  Guide the users to use `backend.ai start` and `backend.ai ssh` for such REPL use cases.

### LINKED ISSUE #1684: Auto-detect the base-distro (aka glibc version) of container images
Currently we rely on `ai.backend.base-distro` label to choose the appropriate build of statically built in-container applications injected by the agent, such as ttyd and dropbear.

Sometimes, there are human errors to misconfigure this lable or omit it, causing unexpected app launch failures after creating session containers.

Could we read the container image (without lauching it?) to determine the glibc version inside it to automate this process?

kyujin-cho:
This PR:
- resolves #758
- resolves #1684

Let's allow running images without Backend.AI-specific labels, particularly the distro labels to mount compatible kernel-runner binaries.

## Approach 1: Statically built kernel-runner binaries

To ease the burden of managing many per-distro per-version kernel-runner binaries, this PR replaces existing kernel-runner binaries with fully statically built versions, using latest Alpine Linux (3.20).

When we install both `zlib-dev` and `zlib-static` in Alpine Linux, we can statically link zlib. `libressl-dev` already contains the static version `/usr/lib/{libcrypto,libssl,libtls}.a`.

| Component | Version |
|---|---|
| Dropbear | 2020.81 → 2024.85 |
| OpenSSH portable (SFTP and SCP)<br>OpenSSL<br>LibreSSL | 8.1p1 → 9.8p1<br>1.1.1i (removed)<br>3.8.4 (new) |
| tmux | 3.0a → 3.4 |

### Backporting statically built binaries

This PR will not be backported to older releases. Instead, #2625 will target 24.03 and 23.09 releases to partially backport the static builds of the kernel-runner binaries.

## Approach 2: Allowing use of container images without Backend.AI-specific metadata labels

This PR adds a new abstract `Agent` interface: `resolve_image_distro()` to determine the distro-version combination to select the kernel-runner binaries to mount.
Statically building most intrinsic executables makes our life easier, but still `libbaihook.so` and the syscall jail requires to be built against specific libc versions.

`DockerAgent` implements this method by executing `ldd --version` command in a temporary container to get the exact libc (glibc or musl) versions, mapping the libc version to a distro version, and caching this result in Redis for faster resolution at the next time. The cache key is the image ID (digest).

`DummyAgent` falls back to "ubuntu16.04" always.

`KuebernetesAgent` first tries to read the `ai.backend.base-distro` label and raises NotImplementedError if it does not exist.

When scanning an image that does not have metadata labels, the intrinsic slots (`cpu` and `mem`) are filled with `manager.defs.INTRINSIC_SLOTS_MIN` (1 core + 1 GiB). All other min values are assumed to be zero and max values to be infinity. The supported accelerator list is initialized as "*" where frontend implementations should allow setting arbitrary resource slots present in the target resource group.


achimnol:
Confirmed working of the new tmux, dropbear, and scp binaries.

achimnol:
I've changed the default value of `ai.backend.runtime-path` from "/opt/backend.ai/bin/python" to "/usr/bin/python" because the `/opt/backend.ai` is reserved for the kernel-runner package only, not intended to host user-submitted codes.

achimnol (changes requested):
* Let's invalidate the cache when the image digest is updated upon rescanning container registries.
* Could we use a separate Redis database for this kind of caching data instead of sharing the "stat" database?

kyujin-cho:
> * Let's invalidate the cache when the image digest is updated upon rescanning container registries.
I don't quite get the first point - image digest (what we usually call as image hash) is identical to each atomic image and thus cannot be shared by multiple image even if newer image has been pushed with same image canonical. Is it really necessary to add the invalidation logic?

achimnol:
> > Let's invalidate the cache when the image digest is updated upon rescanning container registries.
> 
> @achimnol I don't quite get the first point - image digest (what we usually call as image hash) is identical to each atomic image and thus cannot be shared by multiple image even if newer image has been pushed with same image canonical. Is it really necessary to add the invalidation logic?

We may rebuild images. Usually, their base-distro are not changed, but for images like `:latest`, it may change. In a long time ago, we decided to **relax** _the tag formatting rule_.

The problem is that if we just allow arbitrary images that does not have any metadata labels nor our specific tag rule, when combined with #411, there will be random container images pulled in the agents. (e.g., `redis:7.2-alpine`, `postgres:16-alpine`, ... on a local all-in-one setup) Currently we are just ignoring "malformed" image tags like `:latest` during rescan since #293.

We also need to check #758 on the default values of the label fields.

achimnol:
* **#2582** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2582?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @kyujin-cho and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

achimnol (approved):

PR #2612, opened 2024-08-02
feat: add an explicit deprecation notices

### LINKED ISSUE #2560: Deprecate all non-paginated list queries in GraphQL
As our system scale grows, we need to retire non-paginated queries of agents, users, session, kernels and vfolders, where the table size can grow over hundreds and thousands of rows.

- [ ] Add an explicit warning-level logs when they are used.
- [ ] Add deprecation notices to the GraphQL schema.

why-arong:
Drafting....

fixes #2560


why-arong:
currently, it's implemented as middleware, but I plan to switch this using a validator

PR #2645, opened 2024-08-05
feat: add base_context fixture for improved testing setup

why-arong:
### Description:
This PR introduces a new `base_context` fixture in `conftest.py` to facilitate GraphQL query testing. The base_context fixture allows tests to easily override specific context parameters, which is useful when multiple test files require context but with different parameter values.

This change is separated from the original work in [PR #2612](https://github.com/lablup/backend.ai/pull/2612)

### Changes:

Added a base_context fixture in` conftest.py` to provide a flexible context for GraphQL queries.
The `deep_update` function has been included to allow for recursive updates to the context parameters.
- Where is the best place to put this function? (I mean 'file location')


why-arong:
and please review on whether to include annotation or not!!

tests/manager/models/test_etcd.py
```diff
@@ -23,42 +22,20 @@
 
 
 @pytest.fixture(scope="module")
-def client() -> Client:
-    return Client(Schema(query=Queries, mutation=Mutations, auto_camelcase=False))
-
-
-@pytest.fixture(scope="module")
-def context(etcd_container) -> GraphQueryContext:  # noqa: F811
+def etcd_context(base_context, etcd_container) -> GraphQueryContext:  # noqa: F811
```
rapsealk: There is no need to understand the **backend** or **detailed implementation**, whether it's `etcd` or something else. Therefore, I recommend keeping the name `context` as it is.
```suggestion
def context(base_context, etcd_container) -> GraphQueryContext:  # noqa: F811
```

tests/manager/models/test_etcd.py
```diff
@@ -110,7 +87,8 @@ async def test_modify_container_registry(client: Client, context: GraphQueryCont
         },
     }
 
-    response = await client.execute_async(query, variables=variables, context_value=context)
+    response = await client.execute_async(query, variables=variables, context_value=etcd_context)
+    print(response)
```
rapsealk:
```suggestion
```

tests/manager/models/BUILD
```diff
@@ -1 +1,5 @@
 python_tests(name="tests")
+
+python_test_utils(
+    name="test_utils",
+)
```
rapsealk: No new line at the end of the file!
Ref: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_195
```suggestion
)

```

tests/manager/models/conftest.py
```diff
@@ -0,0 +1,65 @@
...
+
+from ai.backend.manager.models.gql import GraphQueryContext, Mutations, Queries
+
+
+@pytest.fixture(scope="module", autouse=True)
+def client() -> Client:
+    return Client(Schema(query=Queries, mutation=Mutations, auto_camelcase=False))
+
+
+@pytest.fixture(scope="module")
```
rapsealk: Maybe you want it to have different fixture per each **function**?
Ref: https://pytest-with-eric.com/fixtures/pytest-fixture-scope/#Pytest-Fixture-Scope-%E2%80%94-Function
```suggestion
@pytest.fixture(scope="function")
```

why-arong: I implemented it this way assuming that the same context would be used on a per-file basis, but it seems that each function within the file may require a different context.

tests/manager/models/test_etcd.py
```diff
@@ -22,38 +21,16 @@
 """
 
 
-@pytest.fixture(scope="module")
-def client() -> Client:
-    return Client(Schema(query=Queries, mutation=Mutations, auto_camelcase=False))
-
-
-@pytest.fixture(scope="module")
```
why-arong, line 30: Since the context function in this file is used as read-only, shouldn't the scope be set to module?

tests/manager/models/conftest.py
```diff
@@ -0,0 +1,65 @@
...
+            "registry": None,
+            "idle_checker_host": None,
+        }
+        updated_params = deep_update(default_params, overrides)
+        return GraphQueryContext(**updated_params)
+
+    return _get_base_context
+
+
+def deep_update(original: dict, updates: dict) -> dict:
```
achimnol: We already have `ai.backend.agent.utils.update_nested_dict()`. Maybe we could move it to `ai.backend.common.utils`.
I think it is also better *not* to return the original value when there is a side-effect to mutate the given argument (`original`).

achimnol (changes requested):
Good work! Just a minor comment.

tests/manager/models/conftest.py
why-arong: I make some changes:

1. Moved `ai.backend.agent.utils.update_nested_dict()` -> `ai.backend.common.utils`
2. Relocated the test code for `update_nested_dict()`from **agent** to the **common** module.
3. Removed `deep_update`and replace its usage with `update_nested_dict` function

tests/manager/models/conftest.py
```diff
@@ -0,0 +1,47 @@
...
+from ai.backend.manager.models.gql import GraphQueryContext, Mutations, Queries
+
+
+@pytest.fixture(scope="module", autouse=True)
+def client() -> Client:
+    return Client(Schema(query=Queries, mutation=Mutations, auto_camelcase=False))
+
+
+@pytest.fixture(scope="function")
+def get_base_context():  # noqa: F811
```
why-arong, line 18: "Is the name get_base_context() suitable for this function, or would you suggest a different name?"

src/ai/backend/agent/docker/utils.py
```diff
@@ -10,9 +10,10 @@
 from aiodocker.exceptions import DockerError
 
 from ai.backend.common.logging import BraceStyleAdapter
+from ai.backend.common.utils import update_nested_dict
```
why-arong, line 13: I used absolute path `from ai.backend.common.utils` instead of the relative path `from ...common.utils `

HyeockJinKim:
We're going to work on removing the root context, so let's close it.

PR #2651, opened 2024-08-05
feat: add deprecated query validator & notice

why-arong:
### Summary:
In this update, I have introduced a new validator (`GQLDeprecatedQueryValidator`) to handle the deprecation of non-paginated GraphQL queries. This validator logs warning message whenever a non-paginated query is used. Additionally, I have marked specific queries with the `deprecation_reason` attribute to indicate that these fields are deprecated.

### Details:

1. Added Validator for Non-Paginated Queries:

- I created a validator named GQLDeprecatedQueryValidator that checks if a field in a GraphQL query is deprecated due to being non-paginated.
- The deprecated queries are listed in the deprecated_queries tuple. These include queries that:
   - End with **-s** (indicating a plural form).
   - Do not have `limit` and `offset` arguments.
    - Are not implemented using the `Relay` specification.

- If a query matches **all these conditions**, the validator logs a warning message using the log.warning method.

2. Deprecation Reason Added to Queries:

- For each deprecated query, I have added a `deprecation_reason` attribute. This helps to clearly indicate that the query is deprecated and provides the reason why.

Example:
```
users = graphene.List(  # legacy non-paginated list
    User,
    domain_name=graphene.String(),
    group_id=graphene.UUID(),
    is_active=graphene.Boolean(),
    status=graphene.String(),
    deprecation_reason="This query is deprecated due to lack of pagination support and will be removed in future versions.",
)
```

### Review Request:

1. Validation of Deprecated Queries:

- Please verify that the list of `deprecated_queries` accurately reflects the queries that should be deprecated. 

2. Deprecation Reason:

- The `deprecation_reason` for each deprecated field is currently set to indicate that the query is deprecated due to lack of pagination support. If there is a **more appropriate** message or additional context that should be provided, please suggest it.
   

**Checklist:** 

### DIFF
--- a/src/ai/backend/manager/api/validators.py
+++ b/src/ai/backend/manager/api/validators.py
@@ -1,0 +1,41 @@
+import logging
+
+from graphql.language import FieldNode
+from graphql.validation import ValidationRule
+
+from ai.backend.common.logging import BraceStyleAdapter
+
+log = BraceStyleAdapter(logging.getLogger(__spec__.name))  # type: ignore[name-defined]
+
+__all__ = ["GQLDeprecatedQueryValidator"]
+
+deprecated_queries = (
+    "agents",
+    "domains",
+    "groups_by_name",
+    "groups",
+    "images",
+    "customized_images",
+    "users",
+    "keypairs",
+    "keypair_resource_policies",
+    "user_resource_policies",
+    "resource_presets",
+    "scaling_groups",
+    "scaling_groups_for_domain",
+    "scaling_groups_for_user_group",
+    "scaling_groups_for_keypair",
+    "vfolders",
+    "container_registries",
+)
+
+
+def is_deprecated_query(query_name: str) -> bool:
+    return query_name.lower() in deprecated_queries
+
+
+class GQLDeprecatedQueryValidator(ValidationRule):
+    def enter_field(self, node: FieldNode, *_args):
+        field_name = node.name.value
+        if is_deprecated_query(field_name):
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/api/validators.py
```diff
@@ -0,0 +1,41 @@
...
+
+def is_deprecated_query(query_name: str) -> bool:
+    return query_name.lower() in deprecated_queries
+
+
+class GQLDeprecatedQueryValidator(ValidationRule):
+    def enter_field(self, node: FieldNode, *_args):
+        field_name = node.name.value
+        if is_deprecated_query(field_name):
+            log.warning("Non-paginated query '{}' is being used.", field_name)
```
kyujin-cho, line 41: Referencing deprecated query outside of the query definition can cause confusion to future readers in a high chance. Please check about possibilities of flagging each query as deprecated within the query implementation.

src/ai/backend/manager/models/gql.py
```diff
@@ -289,6 +289,7 @@ class Queries(graphene.ObjectType):
         Agent,
         scaling_group=graphene.String(),
         status=graphene.String(),
+        deprecation_reason="deprecated_reason",
```
kyujin-cho, line 292: When talking about the deprecation it is important to also mention about alternative solutions. Please update every deprecation reason comments to do so.

kyujin-cho (changes requested):

PR #2661, opened 2024-08-06
fix: let GET requests contain data in param rather than body

fregataa:


agatha197 (approved):
LGTM

PR #2664, opened 2024-08-07
docs: Updated package-based installation documentation for version 24.03.6

Yaminyam:
Based on the current 24.03.6 version of the package, we'll walk through the installation of docs in order to reflect what's missing or required for the latest version.

- [x] Documentation

----
📚 Documentation preview 📚: https://sorna--2664.org.readthedocs.build/en/2664/

----
📚 Documentation preview 📚: https://sorna-ko--2664.org.readthedocs.build/ko/2664/


docs/install/install-from-package/install-agent.rst
achimnol, line 1: Assuming a multi-node setup, I think this shouldn't become 127.0.0.1.

achimnol (changes requested):
Mostly LGTM except `install-agent.rst`.

docs/install/install-from-package/install-agent.rst
Yaminyam: That was one of the things I struggled with the most, what value to put in there.
I decided to go with the values recommended in this file.
https://github.com/lablup/backend.ai/blob/9965d566537a8fb74b1b94dac855f12ad8333e05/configs/agent/sample.toml#L141
Please let me know if you think there are better values!

kyujin-cho: Strategy for setting bound and announced IPs are too complicated to mention in a single paragraph of explanation so I think it is better to split out whole section as a new documentation page.

Yaminyam: If so, I think it would be better to write a single-node based description for the current example and a separate multi-node based description.

PR #2681, opened 2024-08-09
fix: Update `webserver` logout API to respond with HTTP 200 OK

rapsealk:
Follow-up on commit https://github.com/lablup/backend.ai-webserver/commit/867b0f2ad781e723c0590b4803599e26bd4b2c17.

This pull request updates the `webserver` logout API to return an **HTTP 200 OK** status instead of **HTTP 201 CREATED**. The current response code of **201 CREATED** is not suitable for logout operations, as it is typically used to indicate resource creation, which does not align with the intent of a logout request.

This update aligns with common practices observed in popular web frameworks such as `Django` and `Spring`, which typically use **HTTP 200 OK** for successful logout responses. In cases where a redirect is involved, **HTTP 302 Found** is also commonly used.

Using **HTTP 200 OK** for successful logout responses is more intuitive and consistent with standard web development practices. It avoids confusion and ensures that the API response is in line with what developers expect when handling logout operations.

Refs
- [Django](https://github.com/django/django/blob/69aa13ffb92f6a7c62661c616a8c7b0f515ea43d/django/contrib/auth/views.py#L139-L146)
  ```python
    def post(self, request, *args, **kwargs):
        """Logout may be done via POST."""
        auth_logout(request)
        redirect_to = self.get_success_url()
        if redirect_to != request.get_full_path():
            # Redirect to target page once the session has been cleared.
            return HttpResponseRedirect(redirect_to)
        return super().get(request, *args, **kwargs)
  ```
- [Spring](https://github.com/spring-projects/spring-security/blob/6da485ae2036ddb9c3a915507078e43b7954c762/web/src/main/java/org/springframework/security/web/authentication/logout/HttpStatusReturningLogoutSuccessHandler.java#L54-L56)
  ```java
	public HttpStatusReturningLogoutSuccessHandler() {
		this.httpStatusToReturn = HttpStatus.OK;
	}
  ```


achimnol (approved):
According to the following references, this shouldn't affect existing client behaviors.

- https://github.com/lablup/backend.ai/blob/87c6e4162ece3e081bd19d721e9b031d9c8fffa5/src/ai/backend/client/request.py#L570
- https://github.com/aio-libs/aiohttp/blob/bbe71972b4452cd45ed94f5546722a31860a5dda/aiohttp/client_reqrep.py#L1016
- https://github.com/lablup/backend.ai-webui/blob/main/src/lib/backend.ai-client-esm.ts#L340
- https://developer.mozilla.org/ko/docs/Web/API/Fetch_API/Using_Fetch#%EC%9D%91%EB%8B%B5_%EA%B0%9D%EC%B2%B4

PR #2696, opened 2024-08-11
fix: Scan parent directory of created qtree for robust check

fregataa:
follow-up #2170 

When we create a quota after creating Netapp qtree, sometimes Netapp API responds not found error.
#2170 is to avoid such timing issue but it causes another error because Python's Path implementation cache directory data and it takes long to update the cache with newly created qtree directory.

So, let's scan directories from the parent directory of created qtree.


adrysn (approved):
LGTM.

PR #2706, opened 2024-08-13
fix:let vfolder GET requests contain data in params rather than body

why-arong:
After #2483  , we should let GET requests contain data in params rather than body!

Similar to this PR: #2661.

so, let's fix client's vfolder api.
and it could solve the cli-integration failure issue!! (related to: https://github.com/lablup/backend.ai/pull/2703)

---
The existing `build_url` function in **test_vfolders.py** had an issue where it was incorrectly recognizing query parameters as part of the path. 

This has been changed to append query parameters to the URL 
similar to the [`_build_url` method in the Request class.](https://github.com/lablup/backend.ai/blob/5ed63d28a6840b11029e61168baddea4af770438/src/ai/backend/client/request.py#L281C6-L291C19)


src/ai/backend/client/func/vfolder.py
```diff
@@ -607,10 +607,7 @@ async def delete_files(self, files: Sequence[Union[str, Path]], recursive: bool
 
     @api_function
     async def list_files(self, path: Union[str, Path] = "."):
-        rqst = Request("GET", "/folders/{}/files".format(self.name))
-        rqst.set_json({
-            "path": path,
-        })
+        rqst = Request("GET", "/folders/{}/files".format(self.name), params={"path": str(path)})
```
why-arong, line 610: In the Request class, the params attribute [should only accept mappings where the keys are str and the values are either str or int. ](https://github.com/lablup/backend.ai/blob/5ed63d28a6840b11029e61168baddea4af770438/src/ai/backend/client/request.py#L129C4-L136C54)

why-arong: so I convert `path` to `str`

kyujin-cho (approved):

PR #2705, opened 2024-08-13
chore: Remove session scheduler related types no longer used

jopemachine:
## Overview

During the process of refactoring tasks of different characters, I discovered that these types are no longer in use.
It seems advisable to remove these types for maintenance.

### Details

- Remove the following types no longer used.
    - `AgentContext` 
        - Replaced with `AgentRow` in #576.
    - `ExistingSession`
        - Replaced with `SessionRow` in #576.
    - `PendingSession`
        - Replaced with `SessionRow` in #576.
    - `SchedulingPredicate`
        - Created in https://github.com/lablup/backend.ai-manager/pull/192, but I couldn't find where `SchedulingPredicate` is used.
    - `StartTaskArgs`
        - Created in https://github.com/lablup/backend.ai-manager/pull/376. In that PR, there was code that used the `StartTaskArgs` type, but it is no longer present in version 20.09.

- Replace the `KernelInfo` type argument of `assign_agent_for_kernel` with `KernelRow`.
The `KernelInfo` type is not actually used, and at the call site of `assign_agent_for_kernel`, a value of type `KernelRow` is passed to that parameter. You can see the `kernel` passed to the `assign_agent_for_kernel` is `KernelRow` type in the below code. https://github.com/lablup/backend.ai/blob/62eca72efffa3de057534b7ebaaf0cb55d1a8060/src/ai/backend/manager/scheduler/dispatcher.py#L1027-L1094

> [!NOTE]
> Unlike other types, I couldn't find any evidence that the `SchedulingPredicate` type was used in the past, so there might be some misunderstanding on my part. 
> If that's the case, please let me know. If it is a useful type somewhere, we can fix it by replacing the removed `PendingSession` type with `SessionRow`.

---

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version

PR #2643, opened 2024-08-05
feat: Enable custom descriptions for GQL relay edge types

rapsealk:
This pull request is a follow-up of #1719 and #2165.

This will enable custom descriptions for GraphQL `relay` connection types.

```diff
class GroupConnection(Connection):
    class Meta:
        node = GroupNode
+       description = "Added in 24.03.0"  # <-- Description comes here!
```

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version

PR #2615, opened 2024-08-02
feat: Implement ID based client workflow to `ContainerRegistry` graphql API

jopemachine:
## Overview

* The existing `ContainerRegistry` graphql API based on `hostname` will continue to be maintained for compatibility, but it will be deprecated due to breaking changes.

* The new `ContainerRegistry` API has been renamed to `ContainerRegistryNode` and features an updated client workflow based on `ID` instead of `hostname`. Please refer to the test code for usage instructions.

* `hostname` is changed into `registry_name`.

---

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version
- [x] Update of end-to-end CLI integration tests in `ai.backend.test`
- [x] API server-client counterparts (e.g., manager API -> client SDK)
- [x] Test case(s) to:
  - Demonstrate the difference of before/after
  - Demonstrate the flow of abstract/conceptual models with a concrete implementation

----
📚 Documentation preview 📚: https://sorna--2615.org.readthedocs.build/en/2615/

----
📚 Documentation preview 📚: https://sorna-ko--2615.org.readthedocs.build/ko/2615/

PR #2707, opened 2024-08-13
feat: Add `project` column to the `images` table and refactor `ImageRef` related logics

jopemachine:
## Overview

- Adds `project` column to the `images` table.
- Refactors `ImageRef` to be a simple type that only holds data (*dataclass*).
    - Add `project` field. Previously, the `name` field mixed the `project` and image names, but now `ImageRef.name` only contains the image name, and the `project` value is separately in the `ImageRef.project` field.

## Why

Saving the `project` value in the DB is necessary because the string corresponding to `project` may differ depending on the type of container repository, making it impossible to separate the `project` from the image canonical string.

The `project` value should also be included in the `ImageRef` field, and `ImageRef` should be refactored into a simple struct that only holds data, separated from the existing complex logic of parsing image strings.

## Refactoring details

This PR performs relatively large-scale refactoring to add a `project` field to the `ImageRef` corresponding to the `ImageRow`.

### `ParsedImageStr` type

- `ParsedImageStr` is a namedtuple that represents the parsed result of an image string, containing properties such as `canonical`.
- `ParsedImageStr` type is similar to the `ImageRef` type, but it cannot distinguish between the project and image name and does not include *architecture* information, so it does not correspond to a single `ImageRow`.
- `ImageRef.parse_image_str()` returns a `ParsedImageStr` type, which can be used to create an `ImageRef`. But you can also create an `ImageRef` instance directly using `ImageRef.from_image_str()`.

### Comparison of `ImageRef.parse_image_str()` with the existing `ImageRef()` constructor

- In the existing code, `known_registries`* which contains contextual information about registries, was separately provided and used to parse the `image_str (value)`
- However, in this refactoring PR, `ImageRef` is now just a simple dataclass, so the parsing logic previously handled by the `ImageRef` constructor is now handled by `ImageRef.parse_image_str()`.
- `ImageRef.parse_image_str()` can only accept a single registry as an argument, unlike the existing code, which allows for multiple registries through `known_registries`.
- The image string parsing logic still behaves the same as before. For example, the behavior when passing `['*']` to the `known_registries` parameter in the original code can be replicated by passing `"*"` to the registry argument in `ImageRef.parse_image_str()`. For more details, please refer to the comments in the `ImageRef.parse_image_str()` function and the `test_docker.py` test code.

### `ImageIdentifier` type

- A tuple of *image canonical string*, and *architecture*.
- Uniquely corresponds one-to-one with an `ImageRow`.
- `ImageRow.resolve()` can now be operated using `ImageIdentifier`, in addition to `ImageAlias` and `ImageRef`.

### Add `IMAGE_REF` type to `ExtTypes`

- Adds the `ImageRef` type to `msgpack` to enable the exchange of `ImageRef` type objects in RPC communication between the managers and agents.

### Etcs

- Modify variables name `image_ref` that are not of the `ImageRef` type (but are actually `str`).
- This PR includes the content of #2705 to avoid unnecessary updates.

---

- [x] Migration script for filling the `Image.project` column
- [x] Test case(s) to:

----
📚 Documentation preview 📚: https://sorna--2707.org.readthedocs.build/en/2707/

----
📚 Documentation preview 📚: https://sorna-ko--2707.org.readthedocs.build/ko/2707/


jopemachine:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2707?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2707** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2707?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1917** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/1917?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#2615](https://github.com/lablup/backend.ai/pull/2615) <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2615?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

achimnol (commented):
- Could we make `ImageRef` also a dataclass?
- Could you elaborate why we need a separate `ParsedImageStr`?

jopemachine:
1. Could we make `ImageRef` also a `dataclass`?

I am currently working on refactoring Images to change the format so that `ImageRef` only holds decomposed values, although this has not yet been reflected in this draft PR.

2. Could you elaborate why we need a separate `ParsedImageStr`?

`ParsedImageStr` is a type created to store the parsed form of the canonical image without the project for now, but I plan to remove it if it is no longer needed after the refactoring is completed.

jopemachine:
We need to fix the CI that is currently broken after the PR below is merged.
https://github.com/lablup/backend.ai/pull/2643

jopemachine:
> [!NOTE]
> This PR includes the contents of PR 2705 to avoid updating code that contains types removed in PR 2705.

src/ai/backend/common/docker.py
```diff
@@ -398,83 +341,163 @@ def __eq__(self, other):
         return self._data == other
 
 
+class ParsedImageStr(namedtuple("ParsedImageStr", ["registry", "project_and_image_name", "tag"])):
```
fregataa: Just a comment.
How about using [typing.NamedTuple](https://docs.python.org/3/library/typing.html#typing.NamedTuple)?
`collections.namedtuple` does not support type hint of attributes. In this case, we cannot set type hint for `registry`, `project_and_image_name` and `tag`

src/ai/backend/common/docker.py
```diff
@@ -398,83 +341,163 @@ def __eq__(self, other):
...
             raise InvalidImageName("Empty image repository/name")
-        if ("/" not in image) and using_default_registry:
-            image = default_repository + "/" + image
-        return image, tag
+        if ("/" not in image_str) and using_default_repository:
+            image_str = default_repository + "/" + image_str
+        return image_str, tag
+
+    @classmethod
+    def parse_image_str(cls, image_str: str, registry: str | None = None) -> ParsedImageStr:
```
fregataa: Is there any counterpart of this function?
like, parse `ImageRef` from `ParsedImageStr`.

src/ai/backend/manager/models/image.py
```diff
@@ -64,6 +64,7 @@
     from .gql import GraphQueryContext
 
 log = BraceStyleAdapter(logging.getLogger(__spec__.name))
+ImageIdentifier = namedtuple("ImageIdentifier", ["canonical", "architecture"])
```
fregataa: Same here, `typing.NamedTuple` would be better.

src/ai/backend/manager/models/image.py
```diff
@@ -272,6 +282,29 @@ async def from_alias(
...
+    async def from_image_identifier(
+        cls,
+        session: AsyncSession,
+        identifier: ImageIdentifier,
+        load_aliases: bool = True,
+    ) -> ImageRow:
+        query = sa.select(ImageRow).where(
+            ImageRow.name == identifier.canonical
+            and ImageRow.architecture == identifier.architecture
+        )
```
fregataa:
```suggestion
        query = sa.select(ImageRow).where(
            (ImageRow.name == identifier.canonical)
            and (ImageRow.architecture == identifier.architecture)
        )
```
As I remember, trailing condition statement (like `Row.id == row_id and Row.name == row_name`) raises syntax error.
This would not raise error since two condition statements are divided by a new line but let's make it safe just in case.

fregataa (commented):
1. As your comment, `ParsedImageStr` is used to store image reference that does not contain project data. Then how about renaming it to represent that it is for images that doesn't have project data?

2. It would be good to leave description comments on `ai.backend.manager.models.image.ImageIdentifier`. And is it should be in `manager/models/image` or `common/docker`?

jopemachine:
> As your comment, ParsedImageStr is used to store image reference that does not contain project data. Then how about renaming it to represent that it is for images that doesn't have project data?

Do you have any ideas for a better name? 
IMHO, if we try to reflect the fact that this type doesn't distinguish between project and image names in the type name, it might end up being too verbose and lengthy.

> It would be good to leave description comments on ai.backend.manager.models.image.ImageIdentifier. And is it should be in manager/models/image or common/docker?

At the moment, `ImageIdentifier` is used to find image records and is only utilized in the manager component, so I believe it is appropriate to place it in `models/image.py`

src/ai/backend/common/docker.py
jopemachine: No, in my opinion, it wouldn’t be necessary since we can directly create an `ImageRef` object from an image string using `ImageRef.from_image_str`.

PR #2709, opened 2024-08-13
feat: Configurable GQL query depth limit & Connection page limit

fregataa:
1. To set GQL query depth limit, set etcd's `config/api/max-gql-query-depth` value to integer bigger than 1. Default value is null, which means there is no depth limit.
2. To set GQL Connection page size limit, set etcd's `config/api/max-gql-connection-page-size` value to integer bigger than 1. Default value is null, which means there is no page size limit.
3. Default GQL Connection page size is 10.

----
📚 Documentation preview 📚: https://sorna--2709.org.readthedocs.build/en/2709/

----
📚 Documentation preview 📚: https://sorna-ko--2709.org.readthedocs.build/ko/2709/

### DIFF
--- a/changes/2709.feature.md
+++ b/changes/2709.feature.md
@@ -1,0 +1,3 @@
+* Add query depth limit config of GQL.
+* Add page size limit config of GQL Connection.
+* Set default page size of GQL Connection to 10.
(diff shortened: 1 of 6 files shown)


yomybaby (approved):
LGTM

PR #2716, opened 2024-08-14
ci(BA-542): Verify that the current model schema and the migration file match

Yaminyam:
Validate that the alembic revision file matches the current model schema by checking that no changes were made when it was created.
The require:db-migration changes the trigger condition of check-alembic-migration because it is a label that we attach if there is a change in the actual revision file.


.github/workflows/ci.yml
```diff
@@ -150,6 +150,36 @@ jobs:
...
+    - name: Prepare database schema
+      run: ./backend.ai mgr schema oneshot
+    - name: Create a new revision
+      id: create-revision
+      run: |
+        output=$(./py -m alembic revision --autogenerate)
+        revision_file=$(echo "$output" | grep -oP '(?<=Generating ).*\.py')
+        echo "REVISION_FILE=$revision_file" >> $GITHUB_OUTPUT
+    - name: Verify that revision is empty
+      run: python scripts/check-alembic-revision.py ${{ steps.create-revision.outputs.REVISION_FILE }}
```
achimnol, line 210: Shouldn't we check the exit code to distinguish success and failure?

.github/workflows/ci.yml
```diff
@@ -150,6 +150,36 @@ jobs:
...
+        cache-lmdb-store: 'true'
+    - name: Prepare DB
+      run: docker compose -f docker-compose.halfstack-main.yml up -d backendai-half-db --wait
+    - name: Pants export
+      run: pants export --resolve=python-default
+    - name: Prepare the alembic configuration file
+      run: cp configs/manager/halfstack.alembic.ini alembic.ini
+    - name: Prepare database schema
+      run: ./backend.ai mgr schema oneshot
+    - name: Create a new revision
```
achimnol: How about renaming this step to "Try creating alembic migration" to clearly indicate that this is not an actual revision but for a check, and make it clear that the revision is about alembic migrations (not a general term)?

scripts/check-alembic-revision.py
```diff
@@ -0,0 +1,43 @@
...
+    for node in ast.walk(tree):
+        if isinstance(node, ast.FunctionDef):
+            if node.name == "upgrade":
+                for stmt in node.body:
+                    if isinstance(stmt, ast.Pass):
+                        upgrade_has_pass = True
+            elif node.name == "downgrade":
+                for stmt in node.body:
+                    if isinstance(stmt, ast.Pass):
+                        downgrade_has_pass = True
```
achimnol: We should check if the only valid statement is `pass`, not just inclusion.

For instance,
```python
def upgrade():
    if something:
        pass
    else:
        some_operation()
```
could also satisfy this logic.

achimnol (changes requested):

HyeockJinKim (approved):
lgtm

achimnol:
This is not what I meant....
I'd like to exclude the possibility of false-positive when there is a `pass` statement with a long body of `downgrade()` or `upgrade()` functions auto-generated.
My intention was to check if `downgrade()` and `upgrade()` functions are ***actually empty***, not just they contain `pass` _somewhere inside_.

achimnol (approved):

PR #2720, opened 2024-08-16
ci: Skip `timeline-check` on draft PR

Yaminyam:
CI like lint and test, which the PR author needs to check while writing the PR, and timeline-check, which the admin needs to check before merging the PR, do not need to be checked in draft, but only in the final version.

### DIFF
--- a/.github/workflows/timeline-check.yml
+++ b/.github/workflows/timeline-check.yml
@@ -9,7 +9,12 @@ permissions:
 
 jobs:
   assign-pr-number:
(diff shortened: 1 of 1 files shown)


rapsealk (approved):
LGTM!

According to the [GitHub Actions document](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds), the `towncrier` job will not be triggered if the `assign-pr-number` is skipped.
```
If a job fails or is skipped, all jobs that need it are skipped \
unless the jobs use a conditional expression \
that causes the job to continue.
```

rapsealk:
Please write a changelog before merge. 📝

PR #2730, opened 2024-08-19
fix: Assign the initial value of `occupying_slots` to the value of `requested_slots`

Yaminyam:
In #1115, we did something to initialize occupied_slots to the value of requested_slots when creating the kernel.
The session's data value also ensures that occupying_slots is not empty, so that it has a value even when the session is in PREPARE or PENDING state.


src/ai/backend/manager/registry.py
```diff
@@ -1273,6 +1273,7 @@ async def _enqueue() -> None:
 
                     session_data["environ"] = environ
                     session_data["requested_slots"] = session_requested_slots
+                    session_data["occupying_slots"] = session_requested_slots
```
fregataa, line 1276: This will double the value of `sessions.occupying_slots` because the value of actual allocated resource is added to this in [`finalize_running()`](https://github.com/lablup/backend.ai/blob/main/src/ai/backend/manager/registry.py#L1617-L1618).
We should update the `finalize_running()` or change this code.

fregataa (changes requested):

src/ai/backend/manager/registry.py
Yaminyam: Patched to show the occupying slot as the requested slot if it is empty on the frontend.
Since this is before actual resources are allocated, it seems like a good idea to keep it the way it is now.

PR #2760, opened 2024-08-22
refactor: Split out `ai.backend.logging` package

### LINKED ISSUE #2759: Wrong Implicit log level overridding when `--log-level` argument is not provided

Linux (x86-64)

### Backend.AI version

main

### Describe the bug

Even if I set the `log-level` to `DEBUG` in the toml file, the `log-level` is always overridden by the default value (`INFO`) from the CLI, resulting in the logging level always being set to `INFO` regardless of the config file's `log-level`.

As a result, it is not possible to change the `log-level` when using commands other than `start-server` that do not include the `--log-level` argument.

### To Reproduce

1. Set the logging level to debug in the toml configuration file.
2. Run a command that does not include the `--log-level` argument, such as any command other than `start-server`.
3. Observe that the logging level is set to `INFO` instead of `DEBUG`.

### Expected Behavior

The `log-level` set in the toml file should be respected, and commands that do not include the `--log-level` argument should use the `log-level` specified in the toml configuration.

### Anything else?

_No response_

achimnol:
This PR splits out `ai.backend.common.{logging,logging_utils}` modules to a separate package `ai.backend.logging`.  It will allow fellow developers to more freely import the logging subsystem in simpler CLI packages without importing the entire common packages.

The new `ai.backend.logging` package does NOT have any dependency to other `ai.backend.*` packages by allowing a little bit of duplication of custom trafarets in its config module and making `ai.backend.logging.Logger` to take the msgpack options to enable custom extended types as a runtime initialization parameter.

This PR replaces `ai.backend.common.types.LogSeverity` with `ai.backend.logging.LogLevel` in all places, and also resolves #2759 by actively utilizing `LogLevel.NOTSET` as the default value of the `--log-level` CLI option.

For backward-compatibility with existing plugins, it preserves the importability of `from ai.backend.common.{logging,logging_utils} import BraceStyleAdapter`:
```python
# Kept for backward-compatibility of the plugins relying on the old common.logging module.
import warnings

__all__ = ("BraceStyleAdapter",)

warnings.warn(
    "Please import BraceStyleAdapter from ai.backend.logging "
    "instead of ai.backend.common.logging",
    DeprecationWarning,
)
```

Note that `ai.backend.kernel.logging` is also preserved because the kernel-runner package is a standalone one to be placed inside user containers.

Later, we will split out the common packages into more smaller packages to improve reusability across CLI/client and server components.

- [x] Test case(s) updated
- Package builds and dependencies
  - [x] `pants dependencies src/ai/backend/manager:: | grep src/ai/backend/logging`
  - [x] `pants --tag=wheel package src/ai/backend/logging::`
  - [x] `pants --tag=scie --tag=+lazy package src/ai/backend/manager::`
- Working of the log-level configurations in various commands using the logging subsystem
  - [x] `./backend.ai mgr start-server` (with/without `--debug` option)
  - [x] `./backend.ai mgr gql show` (with/without `manager.toml`) – #1686
- Backward compatibility
  - [x] Support plugins using `ai.backend.common.{logging,logging_utils}.BraceStyleAdater`
  - [x] Show an explicit deprecation warning when importing `ai.backend.common.{logging,logging_utils}`


src/ai/backend/logging/types.py
```diff
@@ -0,0 +1,15 @@
...
+    ERROR = "ERROR"
+    WARNING = "WARNING"
+    INFO = "INFO"
+    DEBUG = "DEBUG"
+    NOTSET = "NOTSET"
+
+
+class LogFormat(enum.StrEnum):
+    SIMPLE = enum.auto()
+    VERBOSE = enum.auto()
```
jopemachine: It seems when defining `StrEnum`, there are cases where some variants are defined as uppercase strings (`LogLevel`) and others use `enum.auto()` (`LogFormat`)

Unless there's a specific reason for the difference, wouldn't it be better to standardize on one method?

src/ai/backend/storage/server.py
```diff
@@ -239,15 +247,15 @@ async def server_main(
 )
 @click.option(
     "--log-level",
-    type=click.Choice([*LogSeverity], case_sensitive=False),
-    default=LogSeverity.INFO,
+    type=click.Choice([*LogLevel], case_sensitive=False),
+    default=LogLevel.INFO,
```
jopemachine: It seems that the default value of `LogLevel` should be changed to `NOTSET`.

```py
...
type=click.Choice([*LogLevel], case_sensitive=False),
default=LogLevel.NOTSET,
...
```

src/ai/backend/logging/types.py
achimnol: The stdlib doc says `enum.auto()` in `StrEnum` defaults to the lowercased name of the field. This is fine for most cases while `LogLevel` is _an exception_ to use the uppercased names as the values.

achimnol: I'll test if the lowercased values work fine in `LogLevel`.

src/ai/backend/logging/logger.py
```diff
@@ -0,0 +1,349 @@
...
+            if rec is None:
+                break
+            if console_handler:
+                console_handler.emit(rec)
+            try:
+                if file_handler:
+                    file_handler.emit(rec)
+                if logstash_handler:
+                    logstash_handler.emit(rec)
+                    print("logstash")
```
jopemachine: It seems maybe that the print statement may have been accidentally included as unnecessary code.

src/ai/backend/logging/config.py
```diff
@@ -0,0 +1,151 @@
...
+        t.Key("protocol", default="tcp"): t.Enum("zmq.push", "zmq.pub", "tcp", "udp"),
+        t.Key("ssl-enabled", default=True): t.ToBool,
+        t.Key("ssl-verify", default=True): t.ToBool,
+        # NOTE: logstash does not have format option.
+    }).allow_extra("*"),
+    t.Key("graylog", default=None): t.Null
+    | t.Dict({
+        t.Key("host"): t.String,
+        t.Key("port"): t.ToInt[1024:65535],
+        t.Key("level", default="INFO"): loglevel_iv,
```
jopemachine: I think it would be better to use Enum value directly instead of string.

```py
...
t.Key("level", default=LogLevel.INFO): loglevel_iv,
...
```

jopemachine (changes requested):
https://github.com/lablup/backend.ai/blob/7ab0ebaa9a28cd5d8bcd7cf087d4f1e6fcc06d64/src/ai/backend/manager/cli/context.py#L54

We also need to apply the logging configuration to commands other than `start-server`. 

For example, we can see that when using the `./backend mgr image rescan` command, the log level is always fixed at `INFO`.

To do so, we should pass the logging configuration as a constructor argument to `LocalLogger` in this line.

```py
        if click_ctx.invoked_subcommand != "start-server":
            self._logger = LocalLogger(self.local_config["logging"])
            self._logger.__enter__()
```

src/ai/backend/logging/types.py
achimnol: I've added `CIStrEnum` and `CIUpperStrEnum` classes to handle this case, as the stdlib enum offers [`_generate_next_value_()` method that can be overriden.](https://docs.python.org/3/library/enum.html#enum.Enum._generate_next_value_)

achimnol:
> https://github.com/lablup/backend.ai/blob/7ab0ebaa9a28cd5d8bcd7cf087d4f1e6fcc06d64/src/ai/backend/manager/cli/context.py#L54
> 
> We also need to apply the logging configuration to commands other than `start-server`. For example, we can see that when using the `./backend mgr image rescan` command, the log level is always fixed at `INFO`.
> 
> To resolve this, we should pass the logging configuration as a constructor argument to `LocalLogger` in this line.
> 
> ```python
>         if click_ctx.invoked_subcommand != "start-server":
>             self._logger = LocalLogger(self.local_config["logging"])
>             self._logger.__enter__()
> ```
> 
> And in the same context, the constructor call for `LocalLogger` in `wsproxy/cli/context.py` should also be updated.

Eh..... this is actually due to #1686.
I think we could keep the use of `self.local_config` optional, but use it if `manager.toml` presents.
I'll update #1686's implementation.

achimnol:
I'm going to postpone setting the local logging config in `wsproxy.cli.context.CLIContext` because the wsproxy package uses pydantic-based config schema while all other components use trafaret-based config schema.  We are planning to migrate to pydantic-based ones in the future, but this refactoring is out of scope of this PR.

PR #2771, opened 2024-08-26
fix: Make the config file be treated as multiline by re

Yaminyam:
The re package, which deals with regular expressions in python, is not in Multiline mode by default.
This can cause problems with syntax that checks for line-related items such as ^ when dealing with multi-line files.
Therefore, we set the re.M mode when creating a pattern object


achimnol (approved):
Thanks!

PR #2785, opened 2024-08-29
chore: change vastdata API token span

fregataa:
Change Vast data client's API token life span since there are some API call failure due to outdate tokens.


adrysn (approved):
LGTM.

PR #2806, opened 2024-09-04
fix: improper variable declaration position on service creation SDK

fregataa:


leksikov:
LGTM. That fix should solve the issue

PR #2819, opened 2024-09-09
feat: Insert container id to kernel obj of registry

fregataa:
follow-up #2178 

Some kernel objects in agent's registry does not have container id.
Let `sync_container_lifecycles()` insert container id to such kernel objects.


fregataa:

PR #2825, opened 2024-09-10
feat: Support setting health check interval for model service

### LINKED ISSUE #2817: Support setting health check interval for model service runtimes
### Main idea

On current implementation we do not have a way to set interval of health check requests periodically sent by Kernel Runner (`ai.backend.kernel`). Let's extend the runner to support controlling the value.

- [ ] Add a new key definition on `ai.backend.common.config.model_definition_iv` which describes the interval of health check request
- [ ] Update kernel runner (and agent also) implementation to resepct the health check interval value set by user

### Alternative ideas

_No response_

### Anything else?

_No response_

jopemachine:
Fix #2817.

The existing code does not allow adjusting the `interval` between health check requests, and the `max_wait_time` is always used as both the interval and the response timeout value.

This PR keeps the `max_wait_time` as the value that triggers a timeout and adds a separate setting that specifies the health check `interval`.

## Detail

For example, if a health check request is successful, the next health check request will be sent after the `interval`.

If the health check request fails, meaning the server responds but with an incorrect response code, the next health check request will be sent after waiting for the `interval`.

If the server does not respond, the health check request will be sent after waiting for the `max_wait_time`.

---

- [ ] Determine a sensible default value for the `interval`


jopemachine:
* **#2825** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2825?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

kyujin-cho (commented):
Please refactor the health check process to utilize `aiotools.create_timer()`.

jopemachine:
Remaining tasks: We need to decide on the default value for the interval and the milestones.

PR #2834, opened 2024-09-12
feat: Configure the logging module's config file to use Pydantic

Yaminyam:

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/logging/config_pydantic.py
+++ b/src/ai/backend/logging/config_pydantic.py
@@ -1,0 +1,158 @@
+import enum
+from pathlib import Path
+from typing import Annotated
+
+from pydantic import BaseModel, ByteSize, ConfigDict, Field
+
+from .types import LogFormat, LogLevel
+
+default_pkg_ns = {"": "WARNING", "ai.backend": "DEBUG", "tests": "DEBUG", "aiohttp": "INFO"}
+
+
+class LogDriver(str, enum.Enum):
+    CONSOLE = "console"
+    LOGSTASH = "logstash"
+    FILE = "file"
+    GRAYLOG = "graylog"
+
+
+class LogstashProtocol(str, enum.Enum):
+    ZMQ_PUSH = "zmq.push"
+    ZMQ_PUB = "zmq.pub"
+    TCP = "tcp"
+    UDP = "udp"
+
+
+class BaseSchema(BaseModel):
+    model_config = ConfigDict(
+        populate_by_name=True,
+        from_attributes=True,
+        use_enum_values=True,
+    )
+
+
+class HostPortPair(BaseSchema):
+    host: Annotated[str, Field(examples=["127.0.0.1"])]
+    port: Annotated[int, Field(gt=0, lt=65536, examples=[8201])]
+
+    def __repr__(self) -> str:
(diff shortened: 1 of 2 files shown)


configs/wsproxy/halfstack.toml
```diff
@@ -1,12 +1,12 @@
...
+advertised-api-port = 5050
 
 # replace these values with your passphrase
-jwt_encrypt_key = "QRX/ZX2nwKKpuTSD3ZycPA"
-permit_hash_key = "gHNAohmntRV0t9zlwTVQeQ"
+jwt-encrypt-key = "QRX/ZX2nwKKpuTSD3ZycPA"
+permit-hash-key = "gHNAohmntRV0t9zlwTVQeQ"
 
-api_secret = "v625xZLOgbMHhl0s49VuqQ"
+api-secret = "v625xZLOgbMHhl0s49VuqQ"
```
HyeockJinKim: Do you need to change the config file?
I think you can make it backwards compatible.

Yaminyam: Backward compatibility!

achimnol (changes requested):
- Let's use Pydantic's alias generator to accept hyphenated field names as well.
- Let's avoid using `Annotated` whenever possible to reduce verbosity.

src/ai/backend/logging/config_pydantic.py
```diff
@@ -0,0 +1,125 @@
...
+    def __str__(self) -> str:
+        return self.__repr__()
+
+    def __getitem__(self, *args) -> int | str:
+        if args[0] == 0:
+            return self.host
+        elif args[0] == 1:
+            return self.port
+        else:
+            raise KeyError(*args)
```
HyeockJinKim: One thing I’m wondering is why the `HostPortPair` class is inside the `logging` package.
It seems like it should be in a different package.

Yaminyam: We support several logging tools and may support more in the future.
(ex, graylog, logstash) 
Since these logging tools require the host address by default, we've just separated that part into a common setting.
ref. https://github.com/lablup/backend.ai/blob/a86c2fb6e1afd7353772758142d72fb013fb38ce/src/ai/backend/logging/config.py#L66

HyeockJinKim:
Please resolve the conflict. @Yaminyam

src/ai/backend/common/metrics/profiler.py
```diff
@@ -6,9 +6,9 @@
 @dataclass
 class PyroscopeArgs:
     enabled: bool
-    app_name: str
-    server_address: str
-    sample_rate: int
+    app_name: str | None
+    server_address: str | None
+    sample_rate: int | None
```
HyeockJinKim: Apply `Optional`

HyeockJinKim (approved):

HyeockJinKim:
Please resolve conflict @Yaminyam

HyeockJinKim:
Can you review again? @achimnol

achimnol (changes requested):
I'd like to signify that config options should defined via BaseSchema by improving the name of BaseSchema class, to prevent confusion with BaseModel.

PR #2852, opened 2024-09-20
feat: Add support for setting a timeout when pulling Docker images

fregataa:
related to https://github.com/aio-libs/aiodocker/pull/900

- [Upgrade `aiodocker` to version 0.23.0.](https://github.com/aio-libs/aiodocker/releases/tag/v0.23.0)
- Add support for setting a timeout when pulling Docker images.
  - Default value is 2 hours.
  - `None` value indicates no timeout is set.
  - Session creation will be cancelled if pulling timed out.

image


src/ai/backend/agent/agent.py
```diff
@@ -1839,11 +1845,24 @@ async def create_kernel(
                 kernel_config["image"]["digest"],
                 AutoPullBehavior(kernel_config.get("auto_pull", "digest")),
             )
+            timeout = cast(
```
kyujin-cho: To properly illustrate purpose of this timeout value please rename the variable.
```suggestion
            image_pull_timeout = cast(
```

src/ai/backend/agent/config.py
```diff
@@ -147,11 +147,25 @@
...
+}
+
 agent_etcd_config_iv = t.Dict({
     t.Key("container-logs", default=default_container_logs_config): t.Dict({
         t.Key("max-length", default=default_container_logs_config["max-length"]): tx.BinarySize(),
         t.Key("chunk-size", default=default_container_logs_config["chunk-size"]): tx.BinarySize(),
     }).allow_extra("*"),
+    t.Key("docker", default=default_docker_config): t.Dict({
+        t.Key("api", default=default_docker_config["api"]): t.Dict({
+            t.Key("pull-timeout", default=default_docker_config["api"]["pull-timeout"]): tx.ToNone
```
kyujin-cho: Could you describe more about the reason of config variable residing under `api/` prefix?

kyujin-cho (changes requested):

src/ai/backend/agent/config.py
fregataa: it is because the variable sets configuration about docker API and I thought of any possibility of adding more config variables under `docker/`. But it is not necessary.

kyujin-cho (approved):

PR #2869, opened 2024-09-26
docs: Update readme for the cuda_open plugin in the monorepo

Yaminyam:
CleanShot 2024-09-26 at 15 37 33@2x
Since the cuda_open plugin is integrated into the monorepo and uses the same version, there is no need to mark version compatibility.

- [ ] Documentation

### DIFF
--- a/src/ai/backend/accelerator/cuda_open/README.md
+++ b/src/ai/backend/accelerator/cuda_open/README.md
@@ -6,23 +6,8 @@ This will allow the agents to detect CUDA devices on their hosts and make them
 available to Backend.AI kernel sessions.
 
 ```console
-$ pip install backend.ai-accelerator-cuda
+$ pip install backend.ai-accelerator-cuda-open
 ```
 
 This open-source edition of CUDA plugins support allocation of one or more CUDA
 devices to a container, slot-by-slot.
-
-Compatibility Matrix
---------------------
-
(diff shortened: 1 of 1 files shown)


src/ai/backend/accelerator/cuda_open/README.md
```diff
@@ -6,23 +6,8 @@ This will allow the agents to detect CUDA devices on their hosts and make them
...
-|:----------------------------:|:----------------:|
-|  20.09.x                     |  2.0.0           |
-|  20.03.9 ~                   |  2.0.0           |
-|  20.03.0 ~ 20.03.8           |  0.14.x          |
-|  19.09.17 ~                  |  0.13.x          |
-|  19.06.x, 19.09.0 ~ 19.09.16 |  0.11.x, 0.12.x  |
-|  19.03.x                     |  0.10.x          |
-
-In the versions released after the above matrix, the agent will set the required
-version range of this plugin as an extra requirements set "cuda".
```
HyeockJinKim, line 28: Even if we don't represent the Compatibility Matrix ourselves, it seems like a good idea to provide a reference link so people can find it.

Yaminyam: We should of course have a compatibility guide in our docs documentation, but I think we should also be able to see compatible versions right in the pip package.

HyeockJinKim (approved):
LGTM

PR #2870, opened 2024-09-26
chore:  Remove license section from configuration files

Yaminyam:
https://github.com/lablup/backend.ai/blob/733bdf6881ca72ac6815befee6bcc90dfcbcd635/configs/agent/sample.toml#L377

Similar to the agent.toml setup, the manager.toml requires license access information.


configs/manager/sample.toml
```diff
@@ -243,3 +243,7 @@ enhanced-aiomonitor-task-info = false
 
 # Periodically sync container stat from Redis to the kernels.last_stat column.
 periodic-sync-stats = false
+
+
+[license]
```
jopemachine: In the development environment, this setting will cause an error indicating that it cannot connect to the license server. 
Wouldn't it be better to either comment out this setting only in `install-dev.sh` or find an alternative approach?

Yaminyam: If the license-related setting exists, but the associated plugin doesn't exist, no error is thrown.
In practice, we've been using the license-related setting in agent for a while now.

jopemachine: You're correct, please don't mind.

configs/manager/sample.toml
```diff
@@ -243,3 +243,7 @@ enhanced-aiomonitor-task-info = false
...
+addr = { host = "127.0.0.1", port = 6099 }
```
HyeockJinKim: I haven't found the config code for the license, could you point me to where it is declared?

HyeockJinKim (approved):
LGTM

PR #2878, opened 2024-09-30
fix: Missing CLIContext and log-level config in wsproxy and manager CLI (#2698)

lablup-octodog:
This is an auto-generated backport PR of #2698 to the 24.03 release.


achimnol:
Marking #2698 as 24.09 target.

PR #2311, opened 2024-06-18
refactor: revamp update mechanism of session & kernel status

fregataa:
resolves #2266 

### Update kernel status to {PREPARING,PULLING,RUNNING,TERMINATING}
1. When a manager consumes `Kernel{PREPARING,PULLING,RUNNING,TERMINATING}Event`, set the status of kernel to the given status and save `session_id` of the kernel to Redis. - `SADD` command is used here
2. Among managers, `update_session_status()` task is triggered by a global timer. The task fetches `session_id`s from Redis that saved in step 2.
3. The task fetches session rows & binding kernel rows from DB and calculates the session status.
4. Update the calculated session status to DB.

### Update kernel status to else status ...
Just set the status of session & kernels to the given status.

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version
- [x] Mention to the original issue
- [x] API server-client counterparts (e.g., manager API -> client SDK)

PR #2884, opened 2024-10-02
fix: Missing batch execution call after session starts

fregataa:
follow-up #2327 #2311 

Manager does not call batch execution after batch-type session starts because the caller part is missing.


agatha197 (approved):
LGTM🚀

PR #2888, opened 2024-10-02
fix: Wrong container registries migration script

jopemachine:
Follow-up of #1917.


src/ai/backend/manager/models/alembic/versions/1d42c726d8a3_migrate_container_registries_from_etcd_to_psql.py
```diff
@@ -346,6 +454,7 @@ def upgrade():
     )
 
     insert_registry_id_to_images()
+    insert_registry_id_to_images_with_black_registry_id()
```
achimnol: Seems to be like a typo: `black` → `blank`?
More precisely, it is not "blank" because there is original value corresponding to the project part in the image ref but just there is no corresponding registry information. "missing" would fit better here.

achimnol (changes requested):
Still it fails:
```
  File "/home/joongi/bai-edge/src/ai/backend/manager/models/alembic/versions/1d42c726d8a3_migrate_container_registries_from_etcd_to_psql.py", line 457, in upgrade
    insert_registry_id_to_images_with_black_registry_id()
  File "/home/joongi/bai-edge/src/ai/backend/manager/models/alembic/versions/1d42c726d8a3_migrate_container_registries_from_etcd_to_psql.py", line 381, in insert_registry_id_to_images_with_black_registry_id
    registry_info = dict(registry_info)
                    ^^^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not iterable
```

jopemachine:
The current PR has become larger than initially intended, making it difficult to review. So, I will close this PR and recreate it as a PR in Graphite.

PR #2892, opened 2024-10-07
fix: Regression preventing creation of dot-prefixed vfolders

achimnol:
We should allow a single prefix dot in vfolder names to make them auto-mounted.
It's a regression in #1663.

- [x] Test case(s) to:


ironAiken2 (approved):
LGTM

PR #2898, opened 2024-10-11
fix: Disallow `None` id encoding in `AsyncNode.to_global_id()`

jopemachine:
`AsyncNode.to_global_id()` encodes the value as base64 and uses it as a global ID even when `None` is passed as the id.

I think this is clearly a mistake rather than the developer's intention, and it can be easily corrected by adding simple exception handling.

By adding this exception handling, we can discover that the logic to pass the `id` argument to `ContainerRegistryNode.from_row()` is missing.

Since this is a separate issue, it will be addressed in a follow-up PR, and I will add the `skip:ci` tag to this PR.

---

### DIFF
--- a/changes/2898.fix.md
+++ b/changes/2898.fix.md
@@ -1,1 +1,1 @@
-Prevent null id parsing in `AsyncNode.to_global_id`.
+Prevent null id encoding in `AsyncNode.to_global_id()`.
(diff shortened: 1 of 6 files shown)


jopemachine:
* **#2898** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/2898?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

fregataa (changes requested):
The target milestone is `23.09` but this PR changes an annotation of `container_registry` schema. Is it intended?

jopemachine:
> The target milestone is `23.09` but this PR changes an annotation of `container_registry` schema. Is it intended?

I'll remove the change in the backport PRs.

PR #2906, opened 2024-10-12
chore: Describe the extension of the config file to use TOML

Yaminyam:
We used to use config files with the conf extension, but now we use toml files, so we also describe the use of toml in the CLI help text.


fregataa (approved):
LGTM
Is there any other CLI help message that has the same issue?

Yaminyam:
> LGTM
Is there any other CLI help message that has the same issue?
@fregataa Thanks for pointing that out
I have the same problem with agent watcher, I will fix it.

PR #2914, opened 2024-10-16
ci: Modify to accommodate the changed release naming convention

Yaminyam:
We decided to have a faster release cycle going forward.
This has led to a wider variety of versions in rules, where minor versions were only available in 03,09.

https://docs.github.com/ko/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

The github actions pattern match syntax does not provide a way to match only numbers, so we chose the current way without using *.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - 'main'
-      - '[0-9][0-9].0[39]'
+      - '[0-9][0-9].[0-9]'
+      - '[0-9][0-9].[0-9][0-9]'
     tags:
-      - '[0-9][0-9].0[39].*'
+      - '[0-9][0-9].[0-9].*'
+      - '[0-9][0-9].[0-9][0-9].*'
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - 'main'
-      - '[0-9][0-9].0[39]'
+      - '[0-9][0-9].[0-9]'
+      - '[0-9][0-9].[0-9][0-9]'
     tags:
-      - '[0-9][0-9].0[39].*'
+      - '[0-9][0-9].[0-9].*'
+      - '[0-9][0-9].[0-9][0-9].*'
```
HyeockJinKim, line 11: I'll have to revisit this PR after the release policy is finalized this week.

achimnol: I think it is fine to apply this PR anyway, for 24.12 release.
It is easy enough to update this regex if needed.

achimnol (approved):

PR #2921, opened 2024-10-18
fix: Rename column from `occupying` to `occupied`

Yaminyam:

- [ ] Test case(s) to:
- [ ] Documentation


HyeockJinKim (commented):
Does `occupying_slots` in `docs/locales/en/LC_MESSAGES/client/cli/sessions.po` not need to be modified? @Yaminyam

HyeockJinKim (changes requested):
I’d prefer not to create a DB migration if possible.
How about just renaming the internal variables?

PR #2944, opened 2024-10-23
feat: Allow container user to join extra Linux groups

### LINKED ISSUE #2850: Extend accelerator plugin architecture to allow container user to join extra Linux groups
### Main idea

There are cases where accelerator driver requires Linux user to join on specific groups (e.g. [ROCm](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/install/prerequisites.html)). To support such cases we can think of extending current accelerator plugin architecture to expose extra set of gids which container user (`work` user) will be joined at. 

### Alternative ideas

_No response_

### Anything else?

This issue can work as a keystone job of #2592, as the implementation of this feature will allow way to make container user join extra groups other than default GID and `44` (`shadow`).

jopemachine:
Resolves #2850.

This PR modifies the `AbstractComputePlugin` interface by adding `get_additional_gids()`, allowing additional GIDs that the work user will belong to be specified through the accelerator plugin configuration.

### DIFF
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -522,6 +522,10 @@ class AbstractKernelCreationContext(aobject, Generic[KernelObjectType]):
                 computer_ctx.instance,
                 device_alloc,
             )
+
+            additional_gids = computer_ctx.instance.get_additional_gids()
+            environ["ADDITIONAL_GIDS"] = ",".join(map(str, additional_gids))
+
             for mount_info in accelerator_mounts:
                 _mount(mount_info.mode, mount_info.src_path, mount_info.dst_path.as_posix())
             alloc_sum = Decimal(0)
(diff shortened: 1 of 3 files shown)


src/ai/backend/agent/agent.py
```diff
@@ -522,6 +522,10 @@ def mount_static_binary(filename: str, target_path: str) -> None:
                 computer_ctx.instance,
                 device_alloc,
             )
+
+            additional_gids = computer_ctx.instance.get_additional_gids()
+            environ["ADDITIONAL_GIDS"] = ",".join(map(str, additional_gids))
+
```
kyujin-cho: This logic will make GIDs only from last accelerator last, not from every accelerators.

kyujin-cho (changes requested):

src/ai/backend/agent/agent.py
jopemachine: Thanks for correction!

PR #2959, opened 2024-10-24
fix: Missing notification of bgtask cancellation or failure when shutting down the server (#2579)

lablup-octodog:
This is an auto-generated backport PR of #2579 to the 23.09 release.


jopemachine:
I manually backported and tested it.

```
❯ ./backend.ai vfolder clone abc abc2 local:volume1
✗ Error during the operation: Exception('test')
```

PR #2965, opened 2024-10-24
feat: limit max num of session to prepare in one tick

fregataa:


fregataa:
Add an "on-hold" label to this PR because we suspected that too few DB connections were causing this issue.

PR #2980, opened 2024-10-25
feat: Improve local type container registry handling

jopemachine:

- [ ] Test case(s) to:
- [ ] Documentation


HyeockJinKim:
Is there a reference that can give me the motivation for this work?

jopemachine:
> Is there a reference that can give me the motivation for this work?

Unfortunately, there is no separate reference compiled. 
But, as explained in the PR description, when rescanning and using a local image (such as when creating a session), it would be much more intuitive to use only the image name rather than assigning arbitrary `registry_name` and `project` name.

jopemachine:
If we decide to apply this PR, creating a migration script to update existing local image records seems necessary so that the Images.image column has the same value as the Images.name column.

jopemachine:
To include the `registry_name` in the `canonical` even for local type strings as in the current implementation, a separate `container_registry` row must be created for each local image with a different registry part. 
And even with this approach, images that do not include a `registry_name` cannot be rescanned at all.

My proposal is to exclude `registry_name` from the `canonical` of Local type images.
https://github.com/lablup/backend.ai/pull/2980/files#diff-b2ff25f56c8729a1b42b80c61cfbbfb9ae08f5e9fe8d9675321f4c3f311404b8R586-R588

@achimnol I would appreciate it if you could review this idea.
I think it will be helpful for us to implement more general image rescan logic.

src/ai/backend/manager/models/image.py
```diff
@@ -269,12 +270,31 @@ def trimmed_digest(self) -> str:
...
             _, tag = ImageRef.parse_image_tag(self.name.split(f"{self.registry}/", maxsplit=1)[1])
         else:
-            image_and_tag = self.name.split(f"{self.registry}/{self.project}/", maxsplit=1)[1]
+            join = functools.partial(join_non_empty, sep="/")
+            if self.is_local:
+                image_and_tag = self.name
+            else:
+                image_and_tag = self.name.split(
+                    f"{join(self.registry, self.project)}/", maxsplit=1
+                )[1]
```
HyeockJinKim: It seems like a good idea to define `join` only where you use it.
Additionally, I think `removeprefix` would be clearer.

src/ai/backend/manager/models/image.py
```diff
@@ -269,12 +270,31 @@ def trimmed_digest(self) -> str:
...
+
             image_name, tag = ImageRef.parse_image_tag(image_and_tag)
```
HyeockJinKim: Right now, only the tag value parsing behavior is different, but the `ImageRef` creation works in 2 places, so I need to check if other values go in differently as well.
How about separating the function that only parses tag differently and calling the constructor only once?

jopemachine: Thanks!

HyeockJinKim (commented):
please check migration script.

HyeockJinKim:
It might be faster for you to start a new PR than to continue this one. I'll close this PR.

PR #2985, opened 2024-10-28
fix: Allow compute session mutation without the priority input field

fregataa:
Currently, `modify_compute_session` does not work if there is no `priority` field in input argument.


HyeockJinKim (approved):
LGTM

PR #3000, opened 2024-10-30
feat(BA-613): tmux version from 3.4 to 3.5a

inureyes:
Terminal emulator in JupyterLab sometimes shows garbage string. This is related with rxvt and tmux.

https://github.com/tmux/tmux/issues?q=is%3Aissue+rxvt+rgb+

To solve the issue, we need to check
 * [x] Update tmux version
 * [ ] Check JupyterLab's terminal 

This PR addresses the first.

- [ ] Test case(s) to:
- [ ] Documentation


inureyes:
We need to build the binaries.

scripts/agent/build-tmux.sh
```diff
@@ -24,8 +24,8 @@ curl -LO https://github.com/libevent/libevent/releases/download/release-2.0.22-s
 tar -zxvf libevent-2.0.22-stable.tar.gz
 curl -LO https://ftp.kaist.ac.kr/gnu/ncurses/ncurses-6.4.tar.gz
 tar zxvf ncurses-6.4.tar.gz
-curl -LO https://github.com/tmux/tmux/releases/download/3.4/tmux-3.4.tar.gz
-tar zxvf tmux-3.4.tar.gz
+curl -LO https://github.com/tmux/tmux/releases/download/3.5a/tmux-3.5a.tar.gz
+tar zxvf tmux-3.5a.tar.gz
```
HyeockJinKim, line 28: Is there a specific reason you applied version `3.5a` instead of `3.5`?

inureyes: @HyeockJinKim  3.5a is the recent stable version of tmux. ref) https://github.com/tmux/tmux/wiki

inureyes:
```
CHANGES FROM 3.5 TO 3.5a

* Do not translate BSpace as Unicode with extended keys.

* Fix so that keys with Shift are represented correctly with extended keys.

* Revert to using /bin/sh for #() and run-shell and if-shell; the change to use
  default-shell only applies now to popups.

* Fix grey colour without a number suffix in styles.
```

HyeockJinKim (approved):

PR #3019, opened 2024-11-04
feat(BA-80): Implement Fine-grained seccomp profile

### LINKED ISSUE #2931: Fine-grained seccomp profile managed by Backend.AI Agent
### Main idea

Since it is well known that some newly released accelerators rely on bleeding-edge syscalls, which are often blocked by docker's default seccomp profile for various reasons. To deal with such issues we can think of a serial of ideas:
- Add a copy of Docker's [latest seccomp profile](https://github.com/moby/moby/blob/master/profiles/seccomp/default.json) JSON into Backend.AI Agent's distribution artifact 
- Add a new interface to `AbstractAccelerator` which tells set of syscalls to be allowed inside container
  - scratch directory fits good to store the file
- Each time kernel spawns make Agent create a copy of the profile JSON and utilizes it as a new seccomp profile JSON for the container
- If the attached accelerator requires extra syscalls to operate (which are determined by step 2) alter the single-use seccomp profile to allow it

jopemachine:
Resolves #2931 ([BA-80](https://lablup.atlassian.net/browse/BA-80)).

Now we can implement `get_additional_syscalls()` in the accelerator implementation to allow additional system calls in the containers.

> [!TIP]
> I think it would be a good idea to automate the process of checking if the `default-seccomp.json` profile is up to date and updating it through CI.
> 
> Although it is not directly related to this PR, I think referring to the PR below will be helpful for automating the update of the `default-seccomp.json` file.
> https://github.com/lablup/backend.ai-jail/pull/18

## Reference

- `default-seccomp.json` is copy-pasted from https://github.com/moby/moby/blob/master/profiles/seccomp/default.json

## How to Test

We can test this PR using the following method.

1. Let's try blocking some essential system calls for session creation in the default-seccomp.json file.
Then, session creation will fail as shown below.

```
❯ ./backend.ai session create python
✗ Session ID 1c8844a4-0e41-4b3b-9098-1dab7cdc97e9 has an error during scheduling/startup or cancelled.
```

2. Next, let’s implement the `get_additional_syscalls()` method in the CUDA `MockPlugin` class to return the blocked system calls.

3. Lastly, let’s try creating the session again with the mock plugin resource options. This time, we could see that the session creation is successful.

```
❯ ./backend.ai session create -r cuda.shares=1 python
∙ Session ID 81dc4784-2271-48a6-a512-a3342032f53b is created and ready.
∙ This session provides the following app services: sshd, ttyd, jupyter, jupyterlab
```

---

[BA-80]: https://lablup.atlassian.net/browse/BA-80?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jopemachine:
* **#3019** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/3019?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/ai/backend/agent/docker/agent.py
```diff
@@ -876,6 +881,28 @@ async def start_container(
                 },
             },
         }
+
+        seccomp_profile_pth = self.config_dir / "seccomp.json"
+
+        if seccomp_profile_pth.exists():
+            with open(seccomp_profile_pth, "r") as seccomp_file:
+                seccomp_profile = json.load(seccomp_file)
```
kyujin-cho: We should not rely on sync-based IO operations.

src/ai/backend/agent/agent.py
```diff
@@ -203,6 +203,7 @@ class AbstractKernelCreationContext(aobject, Generic[KernelObjectType]):
     kernel_features: FrozenSet[str]
     image_ref: ImageRef
     internal_data: Mapping[str, Any]
+    additional_syscalls: List[str]
```
kyujin-cho: To make things more obvious, let's update the variable name a little bit:
```suggestion
    additional_allowed_syscalls: List[str]
```

src/ai/backend/agent/docker/agent.py
```diff
@@ -876,6 +881,28 @@ async def start_container(
...
+                additional_syscalls = self.additional_syscalls
+
+                additional_syscall_rule = {
+                    "names": additional_syscalls,
+                    "action": "SCMP_ACT_ALLOW",
+                    "args": [],
+                    "comment": "Additionally allowed syscalls by Backend.AI Agent",
+                }
+
+                seccomp_profile["syscalls"].append(additional_syscall_rule)
```
kyujin-cho: Wouldn't it be better to just update the copied seccomp.json file in-place and reference it further inside the `container_config`?

src/ai/backend/agent/resources.py
```diff
@@ -428,6 +428,13 @@ def get_additional_gids(self) -> list[int]:
         """
         return []
 
+    def get_additional_syscalls(self) -> list[str]:
```
kyujin-cho: Same here.
```suggestion
    def get_additional_allowed_syscalls(self) -> list[str]:
```

kyujin-cho (changes requested):
Along with the comments, please write down a shell script under `scripts/` folder which downloads up-to-date seccomp profile json from upstream.

src/ai/backend/agent/docker/agent.py
jopemachine: The seccomp option can only be provided as an inline JSON string, and (unlike docker CLI) it does not seem to accept a file path.

So, we've decided to pass an inline JSON string here.

Issue Reference:
https://gitlab.com/gitlab-org/gitlab-runner/-/issues/27235

src/ai/backend/agent/docker/agent.py
```diff
@@ -769,6 +770,31 @@ def protected_services(self) -> Sequence[str]:
...
+    async def _apply_seccomp_profile(self, container_config: MutableMapping[str, Any]) -> None:
+        default_seccomp_pth = self.resolve_krunner_filepath("runner/default-seccomp.json")
+
+        if not default_seccomp_pth.exists():
+            log.warning(
+                "Default seccomp profile file not found in the expected path! Skipped the application of additional syscalls."
+            )
+            return
+
+        async with aiofiles.open(default_seccomp_pth, mode="r") as fp:
```
jopemachine: TODO: If `additional_allowed_syscalls` is empty, do not run this

HyeockJinKim (approved):
LGTM

HyeockJinKim:
@kyujin-cho can you review again?

src/ai/backend/agent/docker/agent.py
```diff
@@ -769,6 +770,31 @@ def protected_services(self) -> Sequence[str]:
             case ResourceGroupType.STORAGE:
                 return ("ttyd",)
 
+    async def _apply_seccomp_profile(self, container_config: MutableMapping[str, Any]) -> None:
+        default_seccomp_pth = self.resolve_krunner_filepath("runner/default-seccomp.json")
```
kyujin-cho: Is this a typo of `path`? I have not seen any common cases of shortening `path` as `pth`...

src/ai/backend/agent/resources.py
```diff
@@ -428,6 +428,13 @@ def get_additional_gids(self) -> list[int]:
         """
         return []
 
+    def get_additional_allowed_syscalls(self) -> list[str]:
+        """
+        Returns system calls allowed within the container.
+        These system calls will be additionally allowed in addition to those allowed by the default seccomp profile.
```
kyujin-cho, line 434: It will be good if we can make a reference to the examples here, either as a string or a URL of syscall names.

kyujin-cho (changes requested):

src/ai/backend/agent/docker/agent.py
jopemachine: I will change it to a more common name,`default_seccomp_path`.

jopemachine:
@kyujin-cho Could you review the PR again and confirm if it's ready to be merged?

HyeockJinKim:
Could you review again? @kyujin-cho

kyujin-cho (approved):
Looks good to me

PR #3028, opened 2024-11-05
fix: Customized image tag parsing error

jopemachine:

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/common/docker.py
+++ b/src/ai/backend/common/docker.py
@@ -312,6 +312,10 @@ class PlatformTagSet(Mapping):
         self._data = dict()
         rx = type(self)._rx_ver
         for tag in tags:
+            if re.match(r"^customized_[0-9a-fA-F]{32}$", tag):
+                self._data["bai_customized_image_tag"] = tag
+                continue
+
             match = rx.search(tag)
             if match is None:
                 raise InvalidImageTag(tag, value)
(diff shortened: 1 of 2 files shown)


jopemachine:
Let’s avoid embedding special keywords or gimmicks directly into the parsing logic.
Instead of adding special keywords like `bai_customized_image_tag`, let’s consider a better approach.

src/ai/backend/common/docker.py
```diff
@@ -312,6 +312,10 @@ def __init__(self, tags: Iterable[str], value: Optional[str] = None) -> None:
         self._data = dict()
         rx = type(self)._rx_ver
         for tag in tags:
+            if re.match(r"^customized_[0-9a-fA-F]{32}$", tag):
+                self._data["bai_customized_image_tag"] = tag
```
HyeockJinKim, line 316: Shouldn't you remove `customized_` from `tag`?

jopemachine: In this PR, I mappped the tag itself to the special keyword `bai_customized_image_tag`.
If we’re only mapping the UUID of the custom image, it might be better to rename the keyword to something like `bai_customized_image_id`.

HyeockJinKim (approved):
LGTM

jopemachine:
> LGTM

This is "on hold" state, I will change this PR's state to draft for avoiding confusion

PR #3031, opened 2024-11-06
fix: Prevent redis password from being logged

jopemachine:
Let's mask the password when outputting the Redis config.

### DIFF
--- a/src/ai/backend/common/types.py
+++ b/src/ai/backend/common/types.py
@@ -1142,6 +1142,13 @@ class EtcdRedisConfig(TypedDict, total=False):
     redis_helper_config: RedisHelperConfig
 
 
+def safe_print_redis_config(config: EtcdRedisConfig) -> str:
+    safe_config = config.copy()
+    if "password" in safe_config:
+        safe_config["password"] = "********"
+    return str(safe_config)
+
+
 class RedisHelperConfig(TypedDict, total=False):
     socket_timeout: float
     socket_connect_timeout: float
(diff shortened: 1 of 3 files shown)


src/ai/backend/common/types.py
```diff
@@ -1142,6 +1142,13 @@ class EtcdRedisConfig(TypedDict, total=False):
     redis_helper_config: RedisHelperConfig
 
 
+def safe_print_redis_config(config: EtcdRedisConfig) -> str:
+    safe_config = config.copy()
+    if "password" in safe_config:
+        safe_config["password"] = "********"
+    return str(safe_config)
```
HyeockJinKim, line 1149: The way you implemented it seems that safe_print_redis_config is optional.
How about processing the above logic in the `__str__` function of `EtcdRedisConfig` to make safe_print_redis_config the default behavior?
I think that password should be masked by default unless there is a special reason.

jopemachine: > The way you implemented it seems that safe_print_redis_config is optional.
> How about processing the above logic in the `__str__` function of `EtcdRedisConfig` to make safe_print_redis_config the default behavior?
> I think that password should be masked by default unless there is a special reason.

As mentioned in the Teams thread, since `EtcdRedisConfig` is a `TypedDict`, we cannot add `__repr__` or `__str__` methods to this type.

So, as an alternative, I created this function.

HyeockJinKim (approved):
LGTM

kyujin-cho:
For future references - we can think of generalizing the `safe_print_redis_config()` function in a way of covering all other sensitive informations, not just for redis.

PR #3054, opened 2024-11-08
fix: Include legacy logging module

jopemachine:
When the code of a plugin loaded at runtime is outdated, it attempt to import `BraceStyleAdapter` from `ai.backend.common.logging`.

(This module has been moved to `ai.backend.logging`, See https://github.com/lablup/backend.ai/pull/2760.)

When packaging a pex file, any unused source code in the component is automatically excluded. 
Therefore, in the case of an updated component that no longer imports the legacy logging module, the legacy logging module will be excluded from the pex.

This causes a ModuleNotFound error when an outdated legacy module is dynamically imported.

So let's update out BUILD configuration to ensure that the legacy logging module is always included in the pex, even if it's not imported.

---

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/agent/BUILD
+++ b/src/ai/backend/agent/BUILD
@@ -13,6 +13,7 @@ python_sources(
         "src/ai/backend/agent/kubernetes:src",  # not auto-inferred
         "src/ai/backend/runner:src",
         "src/ai/backend/helpers:src",
+        "src/ai/backend/common/logging.py:src",  # not auto-inferred (due to lazy-loading plugins)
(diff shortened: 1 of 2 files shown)


jopemachine:
For now, I’ve only included it in the manager and agent settings. If it needs to be included in all components, please let me know, and I’ll update the PR accordingly.

jopemachine:
> For now, I’ve only included it in the manager and agent settings. If it needs to be included in all components, please let me know, and I’ll update the PR accordingly.

Configuration update applied to the webserver and storage-proxy components.

achimnol (approved):

PR #3061, opened 2024-11-09
fix: Change vfolder name when deleted to allow reusing the name

fregataa:
resolves #3041 

Update vfolder name before delete-forever.
The deletion date time format is `%Y-%m-%dT%HH%MM%SS%z`, update the name to `{ORIGINAL}_deleted_{DATE-TIME}`.

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version
- [x] Mention to the original issue
- [x] Update of end-to-end CLI integration tests in `ai.backend.test`

PR #3074, opened 2024-11-12
fix: Change vfolder name when deleted to allow reusing the name (#3061)

lablup-octodog:
This is an auto-generated backport PR of #3061 to the 24.03 release.


jopemachine:
@fregataa Lint failed on CI.

PR #3101, opened 2024-11-15
fix: Invalid use of `subproceess.Popen` API with wrapped file objects

kyujin-cho:
Fix `LZMAFile` and `GZipFile` instances incorrectly fed to `asyncio.create_subprocess_exec()`'s `stdin` parameter.

The following code looks legitimately working and it actually executes:
```python
with lzma.open(...) as f:
    await asyncio.create_subprocess_exec(..., stdin=f)
```
Under the hood, the `subprocess.Popen` constructor calls `f.fileno()` method that returns the file descriptor representing the raw file stream _before decompression_.

These confusing behavior made `create_subprocess_exec()` not to hand out the extracted tarball to the `docker load` command, but rather passing original uncompressed `.xz` (or `.gz`) file, as `Popen` tries to locate make use of the file descriptor (which returned by `fileno()` method) rather than using `IOBase`'s high-level `read()` API where `LZMAFile` and `GZipFile` performs decompression.

We should have done either:
```python
with lzma.open(...) as f:
    proc = await asyncio.create_subprocess_exec(..., stdin=asyncio.subprocess.PIPE)
    await proc.communicate(f.read())
```
(for files whose sizes are controlled by us), or:
```python
with lzma.open(...) as f:
    proc = await asyncio.create_subprocess_exec(..., stdin=asyncio.subprocess.PIPE)
    while True:
        chunk = f.read(CHUNK_SIZE)
        if not chunk:
            break
        await proc.stdin.write(chunk)
```
(for files that may have arbitrarily large sizes).

Even though it is definitely not an intended situation, as `docker load` also accepts tarballs compressed with either XZ or GZip algorithm ([ref](https://docs.docker.com/reference/cli/docker/image/load/#description)) - which has effectively hidden this long-standing bug, at least for more than 5 years...


src/ai/backend/agent/docker/utils.py
```diff
@@ -104,10 +107,17 @@ async def install_latest(self) -> None:
...
                 stdout=subprocess.DEVNULL,
                 stderr=subprocess.PIPE,
             )
+            assert proc.stdin
+            while True:
+                chunk = reader.read(IMAGE_CHUNK_SIZE)
+                if not chunk:
+                    break
+                proc.stdin.write(chunk)
+            await proc.stdin.drain()
```
achimnol: I think it will be better to drain at each chunk, and increase the chunk size a little bit (at least megabytes).

achimnol (changes requested):

achimnol (approved):

PR #3115, opened 2024-11-19
fix: Correctly display os_info in install report

Yaminyam:
Fixes an issue where the part that detects OS information was not executed when the installer was run with the --show-guie option, causing messages to be printed incorrectly.

[before]
CleanShot 2024-11-25 at 16 12 45@2x
[after]
image


src/ai/backend/install/cli.py
```diff
@@ -226,9 +226,10 @@ def cancel_config(self) -> None:
 
 
 class InstallReport(Static):
-    def __init__(self, install_info: InstallInfo, **kwargs) -> None:
+    def __init__(self, install_info: InstallInfo, os_info, **kwargs) -> None:
```
jopemachine: Let's add type hint to `os_info`

HyeockJinKim: Is it difficult to add type hints? @Yaminyam

src/ai/backend/install/cli.py
```diff
@@ -542,14 +562,7 @@ def compose(self) -> ComposeResult:
...
         if self._args.show_guide:
-            try:
-                install_info = InstallInfo(**json.loads((Path.cwd() / "INSTALL-INFO").read_bytes()))
-                yield InstallReport(install_info)
-            except IOError as e:
-                log = SetupLog()
-                log.write("Failed to read INSTALL-INFO!")
-                log.write(e)
-                yield log
+            asyncio.create_task(self.show_guide())
```
achimnol: Please DO NOT use "fire-and-forget" pattern whenever possible.

Also, in the Textual framework, the `mount()` method should be invoked using `await` to avoid potential race conditions in the screen rendering while in the async context (see #2867) or keep `yield Widget(...)` inside the `compose()` generator function consistently.

If you want to refactor the logic, please **separate** the widget composition and the actual business logic, rather than simply copy-and-paste out as a function.

src/ai/backend/install/cli.py
```diff
@@ -242,7 +251,7 @@ def compose(self) -> ComposeResult:
         - Username: `admin@lablup.com`
         - Password: `wJalrXUt`
 
-        To see this guide again, run './backendai-install-<platform> install --show-guide'.
+        To see this guide again, run './backendai-install-{self.os_info.platform} install --show-guide'.
```
achimnol: Could we replace this the actual execution command from `sys.argv` instead of composing the executable name manually?
For example, what if we change the CI workflow and the scie binary naming scheme? How could a future developer notice that he/she should update here, too?

achimnol (changes requested):

src/ai/backend/install/cli.py
HyeockJinKim: Please review this once. @Yaminyam

PR #3125, opened 2024-11-21
fix: Outdated image string join logic

jopemachine:
Follow-up to #2978.

Since `project` value was changed to be nullable in PR #2978, the `image` string join logic also should be updated.

### DIFF
--- a/src/ai/backend/manager/models/image.py
+++ b/src/ai/backend/manager/models/image.py
@@ -35,6 +35,7 @@ from ai.backend.common.types import (
     ImageRegistry,
     ResourceSlot,
 )
+from ai.backend.common.utils import join_non_empty
 from ai.backend.logging import BraceStyleAdapter
 from ai.backend.manager.models.container_registry import ContainerRegistryRow
 
@@ -264,7 +265,8 @@ class ImageRow(Base):
             image_name = ""
             _, tag = ImageRef.parse_image_tag(self.name.split(f"{self.registry}/", maxsplit=1)[1])
         else:
-            image_and_tag = self.name.split(f"{self.registry}/{self.project}/", maxsplit=1)[1]
+            join = functools.partial(join_non_empty, sep="/")
+            image_and_tag = self.name.split(f"{join(self.registry, self.project)}/", maxsplit=1)[1]
             image_name, tag = ImageRef.parse_image_tag(image_and_tag)
 
         return ImageRef(


fregataa (approved):
Good. After this PR, let's refactor this codes by replacing `split()` to `partition()` and adding type hints to the `join_non_empty()` function

jopemachine:
> Good. After this PR, let's refactor this codes by replacing `split()` to `partition()` and adding type hints to the `join_non_empty()` function

@fregataa You might be referring to the following code.
By the way, is there a specific reason we prefer `partition()` over `split()`?
I don't see significant readability difference between code that uses `split()` and code that uses `partition()`.

```py
image_and_tag = self.name.partition(f"{join(self.registry, self.project)}/")[2]
```

fregataa:
> @fregataa You might be referring to the following code. By the way, is there a specific reason we prefer `partition()` over `split()`? I don't see significant readability difference between code that uses `split()` and code that uses `partition()`.
> 
> ```python
> image_and_tag = self.name.partition(f"{join(self.registry, self.project)}/")[2]
> ```

`partition()` always returns a tuple of three strings but `split()` returns a list of strings.
you can use `split()` if you expect the string value will be divided into more than 2 parts. but if you are sure that the value is divided into 2 parts, using `partition()` makes it easy to understand how the result comes from the original value.
```python
image_and_tag = self.name.partition(f"{join(self.registry, self.project)}/")[2] # This is okay but
_, _, image_and_tag = self.name.partition(f"{join(self.registry, self.project)}/") # I recommend to use this pattern with partition()
```

jopemachine:
> > @fregataa You might be referring to the following code. By the way, is there a specific reason we prefer `partition()` over `split()`? I don't see significant readability difference between code that uses `split()` and code that uses `partition()`.
> > ```python
> > image_and_tag = self.name.partition(f"{join(self.registry, self.project)}/")[2]
> > ```
> 
> `partition()` always returns a tuple of three strings but `split()` returns a list of strings. you can use `split()` if you expect the string value will be divided into more than 2 parts. but if you are sure that the value is divided into 2 parts, using `partition()` makes it easy to understand how the result comes from the original value.
> 
> ```python
> image_and_tag = self.name.partition(f"{join(self.registry, self.project)}/")[2] # This is okay but
> _, _, image_and_tag = self.name.partition(f"{join(self.registry, self.project)}/") # I recommend to use this pattern with partition()
> ```

Thanks for your description! I will reflect this.

src/ai/backend/manager/models/image.py
```diff
@@ -264,7 +265,8 @@ def image_ref(self) -> ImageRef:
             image_name = ""
             _, tag = ImageRef.parse_image_tag(self.name.split(f"{self.registry}/", maxsplit=1)[1])
         else:
-            image_and_tag = self.name.split(f"{self.registry}/{self.project}/", maxsplit=1)[1]
+            join = functools.partial(join_non_empty, sep="/")
+            _, _, image_and_tag = self.name.partition(f"{join(self.registry, self.project)}/")
```
HyeockJinKim: str have `removeprefix`

HyeockJinKim (approved):
LGTM

PR #3133, opened 2024-11-25
feat: show inference metrics from `Endpoint` and `Routing` queries

kyujin-cho:
This PR brings following updates:
- Huggingface TGI runtime variant support
- Show inference metric from GQL query
  - This feature will only be effective when using Enterprise AppProxy, as we do not have plans to support such feature on OSS AppProxy

## Inference metric format
```json
{
  # ...
  "vllm_request_params_n": {
    "__type": "HISTOGRAM",
    "current": {
      "1.0": "3832319",
      "2.0": "3832319",
      "5.0": "3832319",
      "10.0": "3832319",
      "20.0": "3832319",
      "+Inf": "3832319"
    },
    "threshold_unit": "le",
    "count": 3832319,
    "sum": "3832319"
  },
  "vllm_request_params_best_of": {
    "__type": "HISTOGRAM",
    "current": {
      "1.0": "3832319",
      "2.0": "3832319",
      "5.0": "3832319",
      "10.0": "3832319",
      "20.0": "3832319",
      "+Inf": "3832319"
    },
    "threshold_unit": "le",
    "count": 3832319,
    "sum": "3832319"
  },
  "vllm_cache_config_info": {
    "__type": "GAUGE",
    "current": "1",
    "capacity": null,
    "pct": "0.00",
    "unit_hint": "count"
  },
  # ...
}
```

Metrics from inference frameworks can be distinguished by three types: `GAUGE`, `COUNTER` and `HISTOGRAM`. The identifier `__type` bundled at the root of every metrics represents its type. `GAUGE`, `COUNTER` follows Backend.AI's session metrics format, whereas `HISTOGRAM` has its own new defintion.
### `HISTOGRAM` metric format
On `HISTOGRAM` type, the distribution will be illustrated under `current` key. `count` and `sum` are optional, and will be supplied only if inference framework supports.

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/agent/agent.py
```diff
@@ -2275,6 +2275,22 @@ async def load_model_definition(
...
+                    "service": {
+                        "start_command": image_command,
+                        "port": MODEL_SERVICE_RUNTIME_PROFILES[RuntimeVariant.HUGGINGFACE_TGI].port,
+                        "health_check": {
+                            "path": MODEL_SERVICE_RUNTIME_PROFILES[
+                                RuntimeVariant.HUGGINGFACE_TGI
+                            ].health_check_endpoint,
+                        },
+                    },
+                }
```
achimnol: Just for conciseness and **to prevent potential mistakes**, you could refer the matched value as follows.
The same applies to other `case` branches.

```suggestion
            case RuntimeVariant.HUGGINGFACE_TGI:
                _model = {
                    "name": "tgi-model",
                    "model_path": model_folder.kernel_path.as_posix(),
                    "service": {
                        "start_command": image_command,
                        "port": MODEL_SERVICE_RUNTIME_PROFILES[runtime_variant].port,
                        "health_check": {
                            "path": MODEL_SERVICE_RUNTIME_PROFILES[
                                runtime_variant
                            ].health_check_endpoint,
                        },
                    },
                }
```

achimnol (changes requested):

src/ai/backend/manager/models/endpoint.py
```diff
@@ -711,6 +715,50 @@ def from_enum(cls, enum: RuntimeVariant) -> "RuntimeVariantInfo":
...
+    @classmethod
+    async def batch_load_by_endpoint(
+        cls,
+        ctx: "GraphQueryContext",
+        endpoint_ids: Sequence[uuid.UUID],
+    ) -> Sequence[Optional[Mapping[str, Any]]]:
+        async def _build_pipeline(redis: Redis) -> Pipeline:
+            pipe = redis.pipeline()
+            for endpoint_id in endpoint_ids:
+                await pipe.get(f"inference.{endpoint_id}.app")
```
HyeockJinKim: Maybe I don't know the python Redis library well enough, but why is `await` used for the get operation when it's a pipelined request?

kyujin-cho: Looks like the await is redundant here. Thanks.

HyeockJinKim (approved):
lgtm

achimnol (approved):

PR #3137, opened 2024-11-26
fix: Allow admins to delete other user's vfolders

fregataa:
resolves https://github.com/lablup/giftbox/issues/773

----
📚 Documentation preview 📚: https://sorna--3137.org.readthedocs.build/en/3137/

----
📚 Documentation preview 📚: https://sorna-ko--3137.org.readthedocs.build/ko/3137/

### DIFF
--- a/src/ai/backend/manager/api/vfolder.py
+++ b/src/ai/backend/manager/api/vfolder.py
@@ -2423,6 +2423,7 @@ async def get_vfolder_id(request: web.Request, params: IDRequestModel) -> Compac
         entries = await query_accessible_vfolders(
             db_session.bind,
             user_uuid,
+            allow_privileged_access=True,
             user_role=user_role,
             domain_name=domain_name,
             allowed_vfolder_types=allowed_vfolder_types,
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/api/vfolder.py
```diff
@@ -2315,7 +2317,11 @@ async def delete_by_id(request: web.Request, params: DeleteRequestModel) -> web.
         folder_id,
     )
 
-    row = (await resolve_vfolder_rows(request, VFolderPermission.OWNER_PERM, folder_id))[0]
+    row = (
+        await resolve_vfolder_rows(
+            request, VFolderPermission.OWNER_PERM, folder_id, allow_privileged_access=True
+        )
+    )[0]
```
jopemachine, line 2324: @fregataa If the number of vfolders must be exactly one, let's either raise an error or at least display a warning message when it isn't.

fregataa: Thats a good point, we should refactor all use case of this. But I am not going to change any other codes since this is a hotfix.
Please make a new issue for this

jopemachine: Agreed, we only need to fix this in this PR.
I made a separate epic issue. https://github.com/lablup/backend.ai/issues/3159

PR #3139, opened 2024-11-26
fix: Swap the ports that client and manager are looking at

### LINKED ISSUE #3138: The api port in the storage-proxy setting created by the installer is the opposite of sample

Linux (x86-64)

### Backend.AI version

main

### Describe the bug

https://github.com/lablup/backend.ai/blob/ea9ea02476c64acf090c2c21b292264f8c232583/configs/storage-proxy/sample.toml#L63https://github.com/lablup/backend.ai/blob/ea9ea02476c64acf090c2c21b292264f8c232583/src/ai/backend/install/context.py#L841

In sample.toml and everywhere else, the client-facing port is 6021 and the manager-facing port is 6022.

The installer has them reversed.This is fine behaviorally, but it can be confusing for future changes to the config file, so we want to unify it.

### To Reproduce

_No response_

### Expected Behavior

_No response_

### Anything else?

_No response_

JIRA Issue: BA-49

Yaminyam:
close #3138 


src/ai/backend/install/context.py
```diff
@@ -838,11 +838,11 @@ def hydrate_install_info(self) -> InstallInfo:
...
             agent_ipc_base_path="ipc/agent",
             agent_var_base_path="var/agent",
-            storage_proxy_manager_facing_addr=ServerAddr(HostPortPair("127.0.0.1", 6021)),
             storage_proxy_client_facing_addr=ServerAddr(
-                bind=HostPortPair(public_component_bind_address, 6022),
-                face=HostPortPair(public_facing_address, 6022),
+                bind=HostPortPair(public_component_bind_address, 6021),
+                face=HostPortPair(public_facing_address, 6021),
             ),
+            storage_proxy_manager_facing_addr=ServerAddr(HostPortPair("127.0.0.1", 6022)),
```
HyeockJinKim, line 845: Can you please check if the code currently using ports 6021 and 6022 in backend.ai is okay? @Yaminyam

Yaminyam: This part is irrelevant to the behavior because it is the part that generates the config file.
As mentioned in the issue, we change it to be consistent with the port we use by default.

HyeockJinKim (approved):

PR #3154, opened 2024-11-27
refactor: Split config/halfstack/global contexts from root context

achimnol:
This PR splits `RootContext` into three categories of contextual objects:

- Configurations (those from etcd and local TOML files, runtime identifier like `pidx`)
- Halfstack connection pools (postgres and redis)
- Global stateful objects (event bus, background task manager, storage proxy connector, etc.)


achimnol:
* **#3155** <a href="https://app.graphite.com/github/pr/lablup/backend.ai/3155?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3154** <a href="https://app.graphite.com/github/pr/lablup/backend.ai/3154?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai/3154?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

HyeockJinKim (commented):
In my opinion, even if the root context is divided into substructures, the significance of refactoring diminishes if the root context is still being accessed directly. 
It seems necessary to explicitly pass only the required configuration values to handlers or other modules.
Are there any tasks you are considering regarding this? @achimnol

achimnol:
> In my opinion, even if the root context is divided into substructures, the significance of refactoring diminishes if the root context is still being accessed directly. It seems necessary to explicitly pass only the required configuration values to handlers or other modules. Are there any tasks you are considering regarding this? @achimnol

Ultimately yes, but this PR is an interim step towards that goal.
I'd like to first signify whether the root context attributes are either configurations (`c`), global singletons (`g`), or database connection pools (`h`).

Each category would have different sharing scopes. For instance:
* `config`: can be shared and accessed from anywhere
  - Later, we may need to add an interface to reload/live-update the config values using etcd watch for the shared configs.
* `halfstack`: need to pass exactly required ones but we need to first complete #3183 to categorize and make explicit dependency relationships between the global singletons whether each instance belongs to either _service_ or _repository_ layers. We may need to split existing classes to achieve this.
* `global`: need to pass exactly required ones

I think introducing an explicit hierarchy to the root context will help us to automatically replace the references across the entire manager codebase.

HyeockJinKim:
Since the dependency on the root context has now been removed, I will close this PR.

PR #3176, opened 2024-11-29
fix: allow purge vfolders by passing status to precond check function

fregataa:


adrysn (changes requested):
I applied the changes, but the issue persists. Could it be that the superadmin's email doesn't match the vfolder owner's email?

fregataa:
> I applied the changes, but the issue persists. Could it be that the superadmin's email doesn't match the vfolder owner's email?

Do you get the same 404 error after this change?

kyujin-cho (approved):

PR #3186, opened 2024-12-02
fix: Do not multiply session occupying slots

### LINKED ISSUE #3187: Compute session's occupying slots are stored as abnormal multiples

Linux (x86-64)

### Backend.AI version

24.09.2

### Describe the bug

There is an intermittent issue where the occupying slots value of a compute session is stored as an abnormal multiple. This issue occurs sporadically when creating a compute session.

### To Reproduce

1. Create a compute session.
2. Observe the occupying slots value.
3. Note that occasionally, the value is stored as an abnormal multiple.

### Expected Behavior

The occupying slots value should be the same as the requested slots value and should not be stored as a multiple.

### Anything else?

_No response_

fregataa:
resolves #3187 

Whenever `_transit_session_status()` is called, it accumulates session's occupying slots

### DIFF
--- a/src/ai/backend/manager/models/session.py
+++ b/src/ai/backend/manager/models/session.py
@@ -1309,7 +1309,7 @@ class SessionLifecycleManager:
             transited = session_row.determine_and_set_status(status_changed_at=now)
 
             def _calculate_session_occupied_slots(session_row: SessionRow):
-                session_occupying_slots = ResourceSlot.from_json({**session_row.occupying_slots})
+                session_occupying_slots = ResourceSlot()
                 for row in session_row.kernels:
                     kernel_row = cast(KernelRow, row)
                     kernel_allocs = kernel_row.occupied_slots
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/models/session.py
```diff
@@ -1309,7 +1309,7 @@ async def _get_and_transit(
             transited = session_row.determine_and_set_status(status_changed_at=now)
 
             def _calculate_session_occupied_slots(session_row: SessionRow):
-                session_occupying_slots = ResourceSlot.from_json({**session_row.occupying_slots})
+                session_occupying_slots = ResourceSlot()
```
HyeockJinKim, line 1312: The current code seems a bit tricky to write test code... 😢 
Have you tried verifying that it works?

fregataa: yeah I tested it

PR #3201, opened 2024-12-05
feat: Replace `sessions`, `kernels`'s `status_history`'s type `dict` with `list`

jopemachine:

- [ ] Test case(s) to:
- [ ] Documentation


jopemachine:
* **#3201** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/3201?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lablup/backend.ai/3201?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

jopemachine:
@HyeockJinKim How can we specifically implement a length restriction?
We need to decide whether the length restriction should be configurable, whether older logs in the `status_history_log` list should simply be deleted when the limit is exceeded, and whether a warning log should be generated during the deletion process.

src/ai/backend/manager/models/utils.py
```diff
@@ -550,3 +562,34 @@ async def vacuum_db(
...
+def get_lastest_timestamp_for_status(
+    status_history_log: list[dict[str, str]],
+    status: str,
+) -> Optional[datetime]:
+    """
+    Get the last occurrence time of the given status from the status history records.
+    If the status is not found, return None.
+    """
+
+    for item in list(reversed(status_history_log)):
```
jopemachine: TODO: Fix this using `range`

jopemachine: TODO: check `reversed` vs reverse order range

src/ai/backend/manager/models/utils.py
```diff
@@ -550,3 +562,34 @@ async def vacuum_db(
...
+) -> Optional[datetime]:
+    """
+    Get the last occurrence time of the given status from the status history records.
+    If the status is not found, return None.
+    """
+
+    for item in reversed(status_history_log):
+        if item["status"] == status:
+            return dtparse(item["timestamp"])
+    return None
```
jopemachine, line 579: @HyeockJinKim `reversed()` return **Iterator**, so it allows iterating from the end of the list without copying the entire list, eliminating the need to use `range()`.
But we should use `reversed()` alone, not `list(reversed(...))` which copy the entire list.

jopemachine:
Broken CI will be fixed in #3235.

src/ai/backend/client/cli/session/lifecycle.py
```diff
@@ -844,48 +846,63 @@ def logs(session_id: str, kernel: str | None) -> None:
             sys.exit(ExitCode.FAILURE)
 
 
-@session.command("status-history")
+@session.command()
```
HyeockJinKim, line 894: I don't know much about `session.command`, but can you subtract the internal string "status-history"?

src/ai/backend/manager/models/minilang/ordering.py
```diff
@@ -56,6 +56,10 @@ def _get_col(self, col_name: str) -> sa.Column:
                     case JSONFieldItem(_col, _key):
                         _column = get_col_from_table(self._sa_table, _col)
                         matched_col = _column.op("->>")(_key)
+                    case JSONArrayFieldItem(_col_name, _conditions, _key_name):
+                        # TODO: Implement this.
+                        pass
+                        # ...
```
HyeockJinKim, line 62: Was it handled in a follow-up PR?

src/ai/backend/client/cli/session/lifecycle.py
jopemachine: https://github.com/lablup/backend.ai/blob/42b4030e9aecd32e297e8b5b08fad4729a2ea2a9/src/ai/backend/client/cli/session/lifecycle.py#L59-L61

Looking at the code, it seems that the decorator itself does not perform any operations on the arguments.
So, I think it would be better to leave the arguments empty, as is currently done in most of the other code.

src/ai/backend/manager/models/minilang/ordering.py
jopemachine: This code requires discussion before going ahead.
I created a Teams thread, but I haven’t received any responses yet.

src/ai/backend/client/cli/session/lifecycle.py
HyeockJinKim: I think this is using `overload` to overload the function for the last implementation, can you double check? @jopemachine

jopemachine: If no value is passed as an argument, the decorated function name is automatically used as the CLI command.
If we pass an arbitrary string, that string can be used as the CLI command instead of changing the function name.
So, let's leave the arguments empty here, just like with the other functions.

src/ai/backend/manager/models/minilang/ordering.py
jopemachine: @achimnol Could I get feedback on this code? 
(Please refer to the Teams thread for more detail about `JSONArrayFieldItem`.)

src/ai/backend/manager/api/schema.graphql
```diff
@@ -306,6 +306,9 @@ type ComputeContainer implements Item {
   registry: String
   status: String
   status_changed: DateTime
+
+  """Added in 24.12.0."""
```
HyeockJinKim, line 310: Update version please.

HyeockJinKim:
I am going to completely separate the history table related to session status and kernel status, so I will close this PR.

PR #3222, opened 2024-12-08
fix: Broken session CLI commands due to invalid initialization of `ComputeSession`

### LINKED ISSUE #3202: Session `rename`, `commit` CLI commands not working due to invalid initialization

Linux (x86-64)

### Backend.AI version

main

### Describe the bug

```Java
❯ ./backend.ai session rename 534d186c-f73b-4e1d-b96d-846b51accabc name
✘ BackendAPIError: 404 Not Found
  No such session.
  ➜ Message: Session (id=None) does not exist.
```

The "Rename" command is not working.This is because the API should receive `session_name`, but since the `ComputeSession` is created via `from_session_id`, `session_name` is set to `None`.

### To Reproduce

Use the CLI command.

### Expected Behavior

Rename should work

### Anything else?

None

jopemachine:
Fix #3202 ([BA-19](https://lablup.atlassian.net/browse/BA-19))

This PR fixes following broken session CLI commands.

* `session rename`
* `session set-priority`
* `session events`
* `session commit`

## About milestone (24.09)

Starting from version `24.09`, the code uses `session_id` (UUID) as an argument to create a `ComputeSession` object by calling `from_session_id`. However, this approach does not populate the `ComputeSession.name` field. As a result, it sends API calls like `/session/None (session_id_or_name)/rename?name=new-name`, leading to incorrect behavior.

So this PR only needs to be backported to version `24.09`.

---

- [x] Test case(s) to:

[BA-19]: https://lablup.atlassian.net/browse/BA-19?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ai/backend/client/cli/session/lifecycle.py
+++ b/src/ai/backend/client/cli/session/lifecycle.py
@@ -963,7 +963,7 @@ def rename(session_id_or_name: str, new_name: str) -> None:
 @session.command()
 @click.argument("session_id_or_name", metavar="SESSION_ID_OR_NAME")
 @click.argument("priority", metavar="PRIORITY", type=int)
-def set_priority(session_id: str, priority: int) -> None:
+def set_priority(session_id_or_name: str, priority: int) -> None:
     """
     Sets the scheduling priority of the session.
 
@@ -974,7 +974,7 @@ def set_priority(session_id: str, priority: int) -> None:
 
     async def cmd_main() -> None:
         async with AsyncSession() as api_sess:
-            session = api_sess.ComputeSession(session_id)
+            session = api_sess.ComputeSession(session_id_or_name)
             resp = await session.update(priority=priority)
             item = resp["item"]
             print_done(f"Session {item["name"]!r} priority is changed to {item["priority"]}.")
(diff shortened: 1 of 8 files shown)


src/ai/backend/client/cli/session/lifecycle.py
```diff
@@ -1265,11 +1265,8 @@ def events(session_name_or_id, owner_access_key, scope):
 
         async def _run_events():
             async with AsyncSession() as session:
-                try:
-                    session_id = uuid.UUID(session_name_or_id)
-                    compute_session = session.ComputeSession.from_session_id(session_id)
-                except ValueError:
-                    compute_session = session.ComputeSession(session_name_or_id, owner_access_key)
```
jopemachine, line 1272: The intention of the code appears to be to raise a `ValueError` if `session_name_or_id` is a str that is not a `UUID`, and to call `from_session_id` if it is a session ID. Otherwise, the default constructor is intended to be called.

However, when `from_session_id` is called, the `name` field is not populated, which causes the code to not function correctly.

Additionally, it seems to be a mistake that `owner_access_key` is not set when `session_id` is of UUID type.

HyeockJinKim (approved):
lgtm

PR #3240, opened 2024-12-11
fix: Wrong default value of `total_resource_slots` column

fregataa:
`total_resource_slots` column of `domains` and `groups` tables has `"{}"` default value as a string not a JSON data, which causes error when initialize `ResourceSlot()` object from the value.


fregataa:
Closed this because it is already resolved

PR #3268, opened 2024-12-18
chore: Update gitignore to exclude alembic.ini for account manager

fregataa:
ref #2689 

- Update `.gitignore` to exclude `alembic-accountmgr.ini` config file
- Update `install-dev.sh` to install `alembic-accountmgr.ini` instead of `am-alembic.ini` using a more clear file name
- Delete `am-alembic.ini` config file


achimnol:
I don't understand why we have to add this specific pattern to gitignore.
Should we add all patterns like these?:
```
/*account-manager.toml
/*manager.toml
/*agent.toml
/*storage-proxy.toml
/*webserver.conf
/*cuda-mock.toml
/*mock-accelerator.toml
/*.pants.env
/*.pants.bootstrap
/*.pants.rc
/*docker-compose.halfstack.current.yml
/*alembic.ini
/*dev.etcd.volumes.json
/*dev.etcd.installed.json
/*env-*.sh
/*dummy*.toml
/*wsproxy.toml
```

achimnol:
After inspecting the code history, I found that `am-alembic.ini` was a part of #2688.
Let's clarify that it's part of account manager for first readers by renaming it to `alembic-accountmgr.ini`.

As future work, we would need to move the location of generated config files to a sub-directory to avoid conflicts of the same configuration files used in different service daemons.

Issue #3271, opened 2024-12-19
Fix string formatting of BinarySize with very small fractions


achimnol:
We need to ensure the size-suffixed format should use the floating point notation, not the exponents (scientific) notation, to ease parsing the numbers in the clients.

For example, a very subtle fraction like 1048576123 (1000.0001173019409) should be just `"1000m"`, not "1E+3m".

If we run `str(BinarySize(1048576123))`:

- Before: `"1E+3 MiB"`
- Fixed: `"1000 MiB"`

If we run `f"{BinarySize(1048576123):m}"`:

- Before: `"1E+3"`
- Fixed: `"1000m"`

PR #3277, opened 2024-12-19
feat(BA-96): Metric-based model service autoscaling

### LINKED ISSUE #2659: Support auto scaling on Model Service
### Main idea

Support for auto scaling is crucial when serving production-grade model service APIs. We can think of supporting the feature by two approaches: Metric-based and Schedule-based.

```Java
### Tasks
- [ ] Show session metric from model service status page
- [ ] implement Metric-based auto scaling (see the subissues)
- [ ] implement Schedule-based auto scaling
```

### Alternative ideas

_No response_

### Anything else?

_No response_

JIRA Issue: BA-96

kyujin-cho:
Resolves #2659 (BA-96)

## What's changed
- This PR adds the user-configurable auto-scaling rules associated with each model-service endpoint.
- The `scale_services()` function now interprets the configured rules and applies their decisions by changing the desired replica count.
  - The original `scale_services()` logic that fills the difference of the current replica count to the desired replica count remains the same.
- Each rule can either **increase** or **decrease** the desired number of replicas by positive/negative `step_size`, meaning that it is a single-direction trigger. To auto-scale the replicas in bidirectional ways, users must define _at least two rules_.
  - There is no explicit validation or warnings for the auto-scaling rules if they contradict each other. It is the user's responsibility to configure a consistent set of rules for a single metric or combinations of them.

```mermaid
flowchart TD
    A1["Auto-scaling rule 1 (GREATER_THAN...)"] -->|"(&plus;) step_size"| Count(desired replica count)
    A2["Auto-scaling rule 2 (LESS_THAN...)"] -->|"(&minus;) step_size"| Count

    Count -->|apply difference to current replica count| E[Endpoint]
    
    U[User] -->|manually set| Count
```

### Other changes
- This PR separates the aiodataloader handler from the bulk loading mechanism in both `EndpointStatistics` and `KernelStatistics` to enable reuse of the bulk loader. 

## How it works
- Every endpoint (model service) can have one or more auto scaling rules
- Auto scaling rule is defined as:
  - Metric source - inference runtime or kernel
    - inference framework: average value taken from every replicas. Supported only if both AppProxy reports the inference metrics. Check Backend.AI Enterprise guide for more details.
    - kernel: average value taken from every kernels backing the endpoint
  - Metric name (e.g. `cuda.shares` or `vllm_avg_prompt_throughput_toks_per_s`)
  - Comparator - method to compare live metrics with threshold value.
    - `LESS_THAN`: Rule triggered when current metric value goes below the threshold defined
    - `LESS_THAN_OR_EQUAL`: Rule triggered when current metric value goes below or equals the threshold defined
    - `GREATER_THAN`: Rule triggered when current metric value goes above the threshold defined
    - `GREATER_THAN_OR_EQUAL`: Rule triggered when current metric value goes above or equals the threshold defined
  - Step size: size of step of the replica count to be changed when rule is triggered. Can be represented as both positive and negative value - when defined as negative, the rule will decrease number of replicas.
  - Cooldown Seconds: Durations in seconds to skip reapplying the rule right after rule is first triggered.
  - Minimum Replicas: Sets a minimum value for the replica count of the endpoint. Rule will not be triggered if the potential replica count gets below this value.
  - Maximum Replicas: Sets a maximum value for the replica count of the endpoint. Rule will not be triggered if the potential replica count gets above this value.

----
📚 Documentation preview 📚: https://sorna--3277.org.readthedocs.build/en/3277/

----
📚 Documentation preview 📚: https://sorna-ko--3277.org.readthedocs.build/ko/3277/


achimnol (changes requested):
For consistency with existing dataloader usecases, let's rename `bulk_load_xxx` to `batch_load_xxx` (also in `gql_model/image.py`).

To avoid confusion with internal implementation and the primary interface invoked from the dataloader, I **strongly object** to use English word differences like "bulk" vs. "batch". Let's use a more sensible suffix like "_impl".

src/ai/backend/manager/models/endpoint.py
```diff
@@ -494,6 +583,95 @@ async def get(
...
+        query = sa.select(EndpointAutoScalingRuleRow).filter(EndpointAutoScalingRuleRow.id == id)
+        if load_endpoint:
+            query = query.options(selectinload(EndpointAutoScalingRuleRow.endpoint_row))
+        result = await session.execute(query)
+        row = result.scalar()
+        if not row:
+            raise ObjectNotFound("endpoint_auto_scaling_rule")
+        return row
+
+    def __init__(
```
achimnol: Is this constructor strictly required as it's already an ORM class?

src/ai/backend/manager/models/gql.py
achimnol, line 1: I wonder if it's strictly necessary to put "Added in 24.12.0" in all *fields* when the object type itself is a newly added one.  Couldn't we just put it in the `Meta` class of new object types?

achimnol (commented):
I think these comments may be addressed later, but we need to consider them as immediate follow-ups.

src/ai/backend/manager/models/base.py
```diff
@@ -1656,3 +1657,30 @@ def generate_sql_info_for_gql_connection(
...
+
+    impl = sa.VARCHAR
+    cache_ok = True
+
+    def process_bind_param(
+        self,
+        value: Optional[Decimal],
+        dialect: Dialect,
+    ) -> Optional[str]:
+        return str(value) if value else None
```
achimnol: Like #3271 (PR #3272), shouldn't we use an explicit floating format here, or does postgres/sqlalchemy accept scientific notations as well?

src/ai/backend/manager/models/gql.py
achimnol: After discussion with @yomybaby, we decided to ignore the current validation errors and update the rule later. I'll remove those non-necessary description fields.

src/ai/backend/manager/models/base.py
achimnol: I fixed it.

src/ai/backend/manager/models/endpoint.py
kyujin-cho: The purpose of this constructor is to avoid manually assigning individual properties after creating the object.

achimnol: I think it should just work even without this method.

To avoid boilerplate constructor methods for SQLAlchemy models, we could consider using [SQLModel](https://sqlmodel.tiangolo.com/) and similar libraries. I'm going to leave this as a future exploration.

achimnol: In 58dbb9d6ebef3773640aff3bbefe0b1a962e383c, it's just enough to explicit keyword arguments.

achimnol (changes requested):
As @yomybaby has requested, please update the _GraphQL_ schema to use explicit Enum types for `AutoScalingMetricSource` and `AutoScalingMetricComparator`.

achimnol (changes requested):
image
image
There are some async-database handling issues as well.

kyujin-cho:
Well, seems like graphene's `Enum` type expects enum names (not values) as a field input... :( 
image

HyeockJinKim:
Could you review it again? @achimnol

achimnol:
Updated "chore" parts of the changed/affected codes:

- Use `click.UUID`.
- Add missing `case_sensitive=False` and corresponding type annotation in click command arguments.
- Use modern Python features like `@deprecated` and `typing.Self`.
- Explicitly type `click.ParamType` subclasses.
- Fix a few copy-and-paste mistakes from the network management CLI module.
- Fix minor Python exception handling patterns.

achimnol (approved):

PR #3282, opened 2024-12-20
feature: restart model service process

kyujin-cho:
This PR enhances model service reloading experience by enabling user to restart model process only, not the whole container. Since this only stops and reloads the process itself, any changes applied to container spec (e.g. image, resource request, env var, ...) will not be effective.


HyeockJinKim:
Regarding process restart, integration with sokovan is required, and since many aspects are intertwined overall, we'll revisit this if needed later. I'll close the PR.

PR #3299, opened 2024-12-25
refactor: Resolve some pyright errors

### LINKED ISSUE #3298: Remove some type errors found by pyright
Although we are using mypy as the primary type checker, sometimes pyright gives more useful inspection results.

This PR does not exhaustively cover all cases, but just some cases found during the work of BA-10 (#3253). We could repeat this with other tasks.

achimnol:
resolves #3298 (BA-412).

- Fix type definition mismatch of `common/enum_extenion.{py,pyi}`
- Fix possibly unbound variables in multiple places
- Use explicit enum types when possible and fix inconsistent use of enum/str values for `MountPermission` in `get_runner_mount()`
- Use `typing.Self` in `aobject.new()`

### DIFF
--- a/src/ai/backend/agent/docker/agent.py
+++ b/src/ai/backend/agent/docker/agent.py
@@ -1101,8 +1101,8 @@ class DockerKernelCreationContext(AbstractKernelCreationContext[DockerKernel]):
                     )
                     sport["host_ports"] = created_host_ports
 
-        assert repl_in_port != 0, "repl_in_port should have bee assigned."
-        assert repl_out_port != 0, "repl_out_port should have bee assigned."
+        assert repl_in_port != 0, "repl_in_port should have been assigned."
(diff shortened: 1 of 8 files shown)


src/ai/backend/agent/docker/agent.py
```diff
@@ -1100,6 +1101,8 @@ async def _rollback_container_creation() -> None:
                     )
                     sport["host_ports"] = created_host_ports
 
+        assert repl_in_port != 0, "repl_in_port should have bee assigned."
+        assert repl_out_port != 0, "repl_out_port should have bee assigned."
```
rapsealk:
```suggestion
        assert repl_in_port != 0, "repl_in_port should have been assigned."
        assert repl_out_port != 0, "repl_out_port should have been assigned."
```

PR #3337, opened 2024-12-31
fix(BA-433): Misuse of the walrus operator when taking ModifyEndpoint input

### LINKED ISSUE #3336: The replica count of model service endpoints is always set to 1 regardless of user input
## Summary  

Regardless of the actual input value, modifying the replica count of a model service endpoint always sets the count to 1.

### Steps to Reproduce  

1. Create a model service.
1. Change the replica count using GUI or CLI.
1. Reload the model service information page.

### Expected Behavior  

- The replica count should be set to the input value.

### Actual Behavior  

- It is set to 1 always.

## Root cause analysis

There is a mistake in use of the walrus operator (`:=`) when renaming the `desired_session_count` field to `replicas`. We need to use parenthesis carefully to evaluate the result of walrus operator expressions.

refs) [https://peps.python.org/pep-0572/#relative-precedence-of](https://peps.python.org/pep-0572/#relative-precedence-of|smart-link)

achimnol:
resolves #3336 (BA-433)

### DIFF
--- a/changes/3337.fix.md
+++ b/changes/3337.fix.md
@@ -1,0 +1,1 @@
+Fix a regression that modifying a model service endpoint's replica count always sets it to 1 regardless of the user input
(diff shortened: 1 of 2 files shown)


achimnol:
@ironAiken2 reported this issue and has confirmed the issue is fixed.

PR #3341, opened 2025-01-02
fix(BA-438): Libc version not parsed on image without metadata label

kyujin-cho:
Follow-up PR of #3173 and #2582. On #3173 we forced `Entrypoint` value as `sh` to force override the value set from the image, but it caused another shell script evaluation error. Along with that it seems that we can just let container use docker system's default value by passing `[""]` as the `Entrypoint` config.

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/changes/3341.fix.md
+++ b/changes/3341.fix.md
@@ -1,0 +1,1 @@
+Fix image without metadata label not working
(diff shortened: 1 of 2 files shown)


HyeockJinKim (approved):
lgtm

PR #3363, opened 2025-01-04
fix(BA-432): Enforce VFolder name length restriction through the API schema

### LINKED ISSUE #3334: Allow Deletion of Vfolders with Names Longer Than 40 Characters
## Summary  

- The manager fails to delete a vfolder when its name exceeds 40 characters.
  ## History

- Previously, an issue occurred where attempting to create a new vfolder during the deletion of an existing vfolder with the same name would result in a failure due to name conflicts. To resolve this, manager was updated to modify the vfolder’s name during deletion.
- However, this solution introduced a new issue: if the original vfolder name is within the 60-character limit, appending a postfix during deletion can cause the name to exceed the limit.
- With this fix, the manager should still allow the creation of new vfolders with the same name during the deletion process, even if deletion takes a longer time.
- [https://github.com/lablup/backend.ai/blob/aa97730d72f7578bb844b5b6b40e75144e3f84e2/src/ai/backend/manager/models/vfolder.py#L1074-L1075](https://github.com/lablup/backend.ai/blob/aa97730d72f7578bb844b5b6b40e75144e3f84e2/src/ai/backend/manager/models/vfolder.py#L1074-L1075|smart-link)
  ## Steps to Reproduce  

1. Create a vfolder with a name longer than 40 characters.
1. Attempt to delete the vfolder.
1. Observe the failure and the delete-error state, leaving the vfolder undeletable.
   ## Expected Behavior  

- Backend.ai should successfully delete vfolders with names between 40 and 60 characters.
  ## Actual Behavior  

- Deletion fails when the vfolder name exceeds 40 characters. The vfolder remains in a delete-error state and cannot be deleted.
  ## Logs/Errors  

-

fregataa:
resolves #3334 

Enforce VFolder name length restriction through the API schema, not by the DB column constraint.
- Increase the length constraint of the `vfolders.name` column from 64 to 128.
- Set the maximum length of the VFolder name to 64 characters in the create/rename API.
- Refactor the VFolder create/rename API schema definition using Pydantic models.

- [ ] Test case(s) to:

----
📚 Documentation preview 📚: https://sorna--3363.org.readthedocs.build/en/3363/

----
📚 Documentation preview 📚: https://sorna-ko--3363.org.readthedocs.build/ko/3363/


src/ai/backend/common/typed_validators.py
```diff
@@ -156,3 +158,26 @@ def session_name_validator(s: str) -> str:
...
 """Validator with extended re.ASCII option to match session name string literal"""
+
+
+def _vfolder_name_validator(name: str) -> str:
+    """
+    Although the length constraint of the `vfolders.name` column is 128,
+    we limit the length to 64 in the create/rename API
+    because we append a timestamp of deletion to the name when VFolders are deleted.
+    """
+    if (name_len := len(name)) > 64:
```
HyeockJinKim: Rather than using hardcoded constant values, it would be better to declare and use them as Const values.
This improves the readability of the logic. Additionally, if the specifications change in the future, it will be easier to make consistent updates.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -512,6 +513,12 @@ def vfid(self) -> VFolderID:
         return VFolderID(self.quota_scope_id, self.id)
 
 
+# Although the length constraint of the `vfolders.name` column is 128,
+# we limit the length to 64 in the create/rename API
+# because we append a timestamp of deletion to the name when VFolders are deleted.
+VFolderNameTrafaret = tx.Slug(allow_dot=True, max_length=64)
```
HyeockJinKim: Please update all `64` constants representing the length of the vfolder name to use a defined constant.
Also, kindly replace all 128 magic number with a properly defined constant as well. @fregataa

fregataa: I removed this type definition since it is not used anywhere

src/ai/backend/common/defs.py
```diff
@@ -10,3 +11,33 @@
...
+    "/srv",
+    "/sys",
+    "/tmp",
+    "/usr",
+    "/var",
+    "/home",
+]
+RESERVED_VFOLDER_PATTERNS = [re.compile(x) for x in _RESERVED_VFOLDER_PATTERNS]
+API_VFOLDER_LENGTH_LIMIT: Final[int] = 128
+MODEL_VFOLDER_LENGTH_LIMIT: Final[int] = 64
```
HyeockJinKim: Aren’t the two values set in reverse? @fregataa

fregataa: thanks! it was my mistake

HyeockJinKim (approved):
lgtm

PR #3387, opened 2025-01-06
fix(BA-459): password based SSH login not working

### LINKED ISSUE #3386: Password based authentication not working when connecting to container with SSH
## Summary  

- Can't log in with random generated password when establishing SSH connection to kernel on certain images
  ## Steps to Reproduce  

1. Start a session from NGC 23.09 PyTorch image
1. Try to connect to shell via SSH with password on Backend.AI Desktop app
   ## Expected Behavior  

- Login should work fine
  ## Actual Behavior  

- Login fails with "Incorrect password" error
  ## Root Causes  

- Unlike OpenSSH, dropbear does not support authenticating linux users whose password is hashed with `yescrypt` method. Since certain NGC images defaults password hash algorithm to `yescrypt` (check `/etc/pam.d/common-password`) this blocked users from logging in to the shell via password.We can work around this limitation by forcing `chpassword` to use hash algorithms supported by dropbear (e.g. `SHA256` or `SHA512`) when updating user password on `entrypoint.sh`.

kyujin-cho:
Resolves #3386.

### DIFF
--- a/src/ai/backend/runner/entrypoint.sh
+++ b/src/ai/backend/runner/entrypoint.sh
@@ -55,7 +55,7 @@ if [ $USER_ID -eq 0 ]; then
     /opt/backend.ai/bin/python -s /opt/kernel/fantompass.py > "$HOME/.password"
     export ALPHA_NUMERIC_VAL=$(cat $HOME/.password)
     chmod 0644 "$HOME/.password"
-    echo "work:$ALPHA_NUMERIC_VAL" | chpasswd
+    echo "work:$ALPHA_NUMERIC_VAL" | chpasswd -c SHA512
   fi
 
   echo "Executing the main program..."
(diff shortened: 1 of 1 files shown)


src/ai/backend/runner/entrypoint.sh
```diff
@@ -55,7 +55,7 @@ if [ $USER_ID -eq 0 ]; then
     /opt/backend.ai/bin/python -s /opt/kernel/fantompass.py > "$HOME/.password"
     export ALPHA_NUMERIC_VAL=$(cat $HOME/.password)
     chmod 0644 "$HOME/.password"
-    echo "work:$ALPHA_NUMERIC_VAL" | chpasswd
+    echo "work:$ALPHA_NUMERIC_VAL" | chpasswd -c SHA512
```
HyeockJinKim, line 58: In the case where the issue occurred, there shouldn’t be a problem since there were no previously stored hash values, right?
In the normal operation scenario, hash values would have been stored, but changing the hashing algorithm shouldn’t cause any issues?

kyujin-cho: I can find that Ubuntu introduced SHA512 password hashing algorithm at [8.10](https://jonmccune.wordpress.com/2009/07/28/1-vs-6-in-etcshadow/) and RHEL from [5](https://access.redhat.com/solutions/221403). So yes, at least for the Linux distros supported by Backend.AI, it will be fine to fix the password hashing method as SHA512.

HyeockJinKim (approved):
lgtm

PR #3391, opened 2025-01-07
fix(BA-450): do not set timeout on docker push API execution

kyujin-cho:
Resolves https://github.com/lablup/backend.ai-internal/issues/118.

### DIFF
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -1610,7 +1610,12 @@ class AbstractAgent(
         self.images = await self.scan_images()
 
     @abstractmethod
-    async def push_image(self, image_ref: ImageRef, registry_conf: ImageRegistry) -> None:
+    async def push_image(
+        self,
+        image_ref: ImageRef,
+        registry_conf: ImageRegistry,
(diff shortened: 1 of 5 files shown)


src/ai/backend/agent/agent.py
```diff
@@ -1610,7 +1610,13 @@ async def _scan_images_wrapper(self, interval: float) -> None:
         self.images = await self.scan_images()
 
     @abstractmethod
-    async def push_image(self, image_ref: ImageRef, registry_conf: ImageRegistry) -> None:
+    async def push_image(
+        self,
+        image_ref: ImageRef,
+        registry_conf: ImageRegistry,
+        *,
+        timeout: float | None | Sentinel = Sentinel.TOKEN,
```
HyeockJinKim, line 1618: Why is the Sentinel type necessary?

kyujin-cho: It is to distinguish the parameter's state between "Not defined" and "Defined as None". aiodocker, the ultimate consumer of this parameter, takes both state into account and let sets timeout as globally defined value when callee does not fill in this parameter.

HyeockJinKim (approved):
lgtm

PR #3411, opened 2025-01-09
fix(BA-478): Accept container images that has the 'null' labels field in their manifests

### LINKED ISSUE #3409: Fix scanning container images with no labels at all
In [https://github.com/lablup/backend.ai/pull/2582](https://github.com/lablup/backend.ai/pull/2582|smart-link) (24.09), we began to support scanning and running container images that do not have Backend.AI-specific metadata labels. However, we missed that some image manifests may have `null` in the image labels field. Let’s handle such images seamlessly.

achimnol:
resolves #3409 (BA-478)

This PR lets the container registry implementations coalesce `null` values in the manifest `"labels"` field to an empty dict to prevent breakage.

It also adds explicit `typing.override` decorators to mark overriden methods in the `HarborRegistry_v2` class. Though, we need a follow-up refactoring to make the code structure to tell the reader that: 'Harbor and Docker registry share the image manifest query and pull APIs while Harbor has its specific repository listing API.'

This PR is a follow-up fix to #2582.


HyeockJinKim (approved):
lgtm

PR #3412, opened 2025-01-09
feat(BA-468): Add timeout conf for image push

### LINKED ISSUE #3401: Add timeout configuration for Docker image push
## Motivation  

The `push_image()` function in the agent does not allow configuring the timeout for Docker image push operations.

Let’s Introduce a configurable timeout parameter for the `push_image()` function to allow users to specify a suitable timeout duration for their Docker image push operations.

## Additional Information

- Affected function:  Agent’s `push_image()`
- Default timeout: infinite

fregataa:
resolves #3401 

The default timeout value is `None`, which means infinite timeout

### DIFF
--- a/src/ai/backend/agent/config.py
+++ b/src/ai/backend/agent/config.py
@@ -150,10 +150,9 @@ default_container_logs_config = {
 }
 
 DEFAULT_PULL_TIMEOUT = 2 * 60 * 60  # 2 hours
+DEFAULT_PUSH_TIMEOUT = None  # Infinite
 
-default_api_config = {
-    "pull-timeout": DEFAULT_PULL_TIMEOUT,
-}
+default_api_config = {"pull-timeout": DEFAULT_PULL_TIMEOUT, "push-timeout": DEFAULT_PUSH_TIMEOUT}
 
 agent_etcd_config_iv = t.Dict({
     t.Key("container-logs", default=default_container_logs_config): t.Dict({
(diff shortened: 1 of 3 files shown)


src/ai/backend/agent/server.py
```diff
@@ -815,11 +815,15 @@ async def push_image(
         log.info("rpc::push_image(c:{})", image_ref.canonical)
         bgtask_mgr = self.agent.background_task_manager
 
+        image_push_timeout = cast(
+            Optional[float], self.local_config["agent"]["api"]["push-timeout"]
+        )
```
HyeockJinKim, line 820: Is type casting necessary? Can’t we set the timeout value stored in local_config as Optional[float]?

fregataa: the value is stored as Optional[float]. I used [`typing.cast`](https://docs.python.org/3/library/typing.html#typing.cast) to specify the expected type of the value to the type checker

src/ai/backend/agent/config.py
```diff
@@ -163,6 +162,10 @@
     t.Key("api", default=default_api_config): t.Dict({
         t.Key("pull-timeout", default=default_api_config["pull-timeout"]): tx.ToNone
         | t.ToFloat[0:],  # Set the image pull timeout in seconds
+        t.Key("push-timeout", default=default_api_config["push-timeout"]): tx.ToNone
+        | t.ToFloat[
+            0:
+        ],  # Set the image push timeout in seconds. 'None' indicates an infinite timeout.
```
HyeockJinKim, line 168: It doesn’t seem right to have the push-timeout value under the api section.

HyeockJinKim: However, since the existing pull-timeout is also in the same location, it seems like this will require a follow-up task to reorganize the config structure.

HyeockJinKim (approved):

PR #3451, opened 2025-01-14
fix(BA-515): Let the vfolder REST API accept vfolder UUIDs as intended

### LINKED ISSUE #3450: Missing UUID conversion when accepting vfolder "name or ID" in the vfolder REST APIs
## Summary  

The intention of `resolve_vfolder_rows()` in the `manager.api.vfolder` module is to accept either an UUID or a human-friendly vfolder name. However, the primary caller, `@with_vfolder_rows_resolved` decorator did not try to convert the match-info value to UUID and has always passed it as `str`.

The result is that even the client places UUID in the vfolder REST APIs (`/folders/{here}/...`), it was always interpreted as vfolder names, not IDs.

We had a plan to add a new set of vfolder REST APIs as `/folder/{id}/...`, but I think we can just embrace the ID-based queries by fixing `@with_vfolder_rows_resolved` to work as originally intended.

achimnol:
resolves #3450 (BA-515)

- [x] Test case(s) to:

----
📚 Documentation preview 📚: https://sorna--3451.org.readthedocs.build/en/3451/

----
📚 Documentation preview 📚: https://sorna-ko--3451.org.readthedocs.build/ko/3451/

### DIFF
--- a/src/ai/backend/manager/api/vfolder.py
+++ b/src/ai/backend/manager/api/vfolder.py
@@ -943,7 +943,7 @@ async def list_allowed_types(request: web.Request) -> web.Response:
 async def get_info(request: web.Request, row: VFolderRow) -> web.Response:
     root_ctx: RootContext = request.app["_root.context"]
     resp: Dict[str, Any] = {}
-    folder_name = request.match_info["name"]
+    folder_name = row["name"]
     access_key = request["keypair"]["access_key"]
     log.info(
         "VFOLDER.GETINFO (email:{}, ak:{}, vf:{})",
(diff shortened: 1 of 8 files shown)


HyeockJinKim (approved):
lgtm

src/ai/backend/manager/api/vfolder.py
```diff
@@ -294,9 +294,17 @@ def _wrapper(
...
+            try:
+                folder_name_or_id = uuid.UUID(piece)
+            except ValueError:
+                folder_name_or_id = piece
             return await handler(
-                request, await resolve_vfolder_rows(request, perm, folder_name), *args, **kwargs
+                request,
+                await resolve_vfolder_rows(request, perm, folder_name_or_id),
+                *args,
+                **kwargs,
```
fregataa: How about parsing folder id value to UUID in `resolve_vfolder_rows()` rather than in `with_vfolder_rows_resolved()`?
There are some handler functions that call `resolve_vfolder_rows()`

achimnol: Either way is possible, but I thought raw input interpretation should be done in the handler layer while type-based branching should be done inside the logic layer.

PR #3460, opened 2025-01-15
fix(BA-514): the required state output logic in the openopi reference doc

### LINKED ISSUE #3447: Create openapi reference action keeps repeating
## Summary  

The OPEN API documentation is iterative and changing.
This is interfering with normal PR operations.

{width=864 height=216}

## Steps to Reproduce  

1. 
1. 
1. 

## Expected Behavior  

- 

## Actual Behavior  

- 

## Logs/Errors  

-

Yaminyam:
resolves #3447 (BA-514)
The handling of multiple states and one state was different.
And since the state is a set, we need to use sorted to guarantee the order so that we can always generate the same document.

- [x] Documentation

----
📚 Documentation preview 📚: https://sorna--3460.org.readthedocs.build/en/3460/

----
📚 Documentation preview 📚: https://sorna-ko--3460.org.readthedocs.build/ko/3460/


fregataa (approved):
Good, but the openapi schema has been updated again

src/ai/backend/manager/openapi.py
```diff
@@ -271,14 +271,14 @@ def generate_openapi(subapps: list[web.Application], verbose=False) -> dict[str,
...
-                    if len(manager_status) > 0:
+                    if len(manager_status) == 1:
                         preconds.append(
                             f"Manager status required: {list(manager_status)[0].value.upper()}"
                         )
                     else:
                         preconds.append(
                             "Manager status required: one of "
-                            f"{', '.join([e.value.upper() for e in manager_status])}"
+                            f"{', '.join([e.value.upper() for e in sorted(manager_status)])}"
```
fregataa, line 283:
```suggestion
                            f"{', '.join([e.value.upper() for e in sorted(manager_status, key=lambda s: s.value)])}"
```
It will raise error if the `manager_status` is a list of enum values.

PR #3511, opened 2025-01-21
feat(BA-572): Add pydantic-only API hander decorator

### LINKED ISSUE #3512: Implement Pydantic Handling Decorator for Request/Response Validation for new VFolder API
## Motivation  

- Current pydantic request/response handling decorator  lacks consistency due to mixed usage of aiohttp's web.Request and Pydantic models. So we need to standardize and maintain consistency in API interfaces
- Current decorator combines both request/response handling and error handling, creating a need for separation of concerns

## Required Features

- Implement Pydantic middleware to handle all request components (body, query params, path variables) through Pydantic models instead of web.Request
- Deploy as subapp middleware to isolate changes within new vFolder API application without affecting other API endpoints

## Implementation (Detail)

**As-Is**

```python
# 1. Global Auth Middleware
@web.Middleware
async def auth_middleware(request, handler):
    # 1) Initialize auth state in request
    request.update({
        "is_authorized": False,
        "user": None,
        "keypair": None
    })

    # 2) Check if handler requires auth
    if not handler.auth_required:
        return await handler(request)

    # 3) Process authorization and stor info in request obj
    if authenticated:
        request.update({
            "is_authorized": True,
            "user": user_info,
            "keypair": keypair_info
        })

# 2. Pydantic Param handler
def pydantic_params_api_handler(param_model, query_model = None):
    async def wrapped(request):
        try:
            # 1) Validate body/query params
            if request.has_body:
                params = param_model.validate(request.body)
                if query_model:
                    query_model = query_model.validate(request.query)
            else:
                params = param_model.valdiate(request.query)
            
            # 2) Exec handler
            result = await handler(request, params)
        
        # 3) If validation failed make custom msg and raise InvalidAPIParameters error(Custom error)
        except ValidationError as e:
            message = make_custom_msg(e)
            raise InvalidAPIParameters(message)
        
        # 4) Convert Pydantic Response model into aiohttp web.Response
        return web.json_response(result.model_dump(mode="json"))

    return wrapped

# 3. Usage Example
@auth_required  
@pydantic_params_api_handler(RequestModel)
async def handler(request, params):
   user = request["user"]  # Set from auth_middleware
   data = params  # Validated by pydantic
```

**To-Be**

## Impact  

- Consistent Request/Response processing logic across all API endpoints
- Easy API documentation generation by using Pydantic

## Testing Scenarios  

- Verify all request components (body, params, path vars) are correctly processed through Pydantic models and validate error handling for each component

seedspirit:
resolves #3512 (BA-572)

## Description
Implement Pydantic api handler decorator to handle all request components (body, query params, path variables) through Pydantic models instead of web.Request

## Implementation
```python
1. Request Body:
    @pydantic_api_handler
    async def handler(body: BodyParam[UserModel]):  # UserModel is a Pydantic model
        user = body.parsed                          # 'parsed' property gets pydantic model you defined
        return BaseResponse(data=YourResponseModel(user=user.id))

2. Query Parameters:
    @pydantic_api_handler
    async def handler(query: QueryParam[QueryPathModel]):
        parsed_query = query.parsed
        return BaseResponse(data=YourResponseModel(search=parsed_query.query))

3. Headers:
    @pydantic_api_handler
    async def handler(headers: HeaderParam[HeaderModel]):
        parsed_header = headers.parsed
        return BaseResponse(data=YourResponseModel(data=parsed_header.token))

4. Path Parameters:
    @pydantic_api_handler
    async def handler(path: PathModel = PathParam(PathModel)):
        parsed_path = path.parsed
        return BaseResponse(data=YourResponseModel(path=parsed_path))
```

To use `MiddlewareParam`
1. Define a class that inherits from MiddlewareParam
2. Implement the from_request class method to extract required data from the request object

```python
5. Middleware Parameters:
    # Need to extend MiddlewareParam and implement 'from_request'
    class AuthMiddlewareParam(MiddlewareParam):
        user_id: str
        user_email: str
        @classmethod
        def from_request(cls, request: web.Request) -> Self:
            # Extract and validate data from request
            user_id = request["user"]["uuid"]
            user_email = request["user"]["email"]
            return cls(user_id=user_id)

    @pydantic_api_handler
    async def handler(auth: AuthMiddlewareParam):  # No generic, so no need to call 'parsed'
        return BaseResponse(data=YourResponseModel(author_name=auth.name))
```

**Multiple Parameters**
```python 
    @pydantic_api_handler
    async def handler(
        user: BodyParam[UserModel],  # body
        query: QueryParam[QueryModel],  # query parameters
        headers: HeaderParam[HeaderModel],  # headers
        auth: AuthMiddleware,  # middleware parameter
    ):
        return BaseResponse(data=YourResponseModel(
                user=user.parsed.user_id,
                query=query.parsed.page,
                headers=headers.parsed.auth,
                user_id=auth.user_id
            )
        )
```

## Test
- Test code with [pytest aiohttp plugins](https://github.com/aio-libs/pytest-aiohttp) which provides fixtures for aiohttp test server and client creation
- Tested BodyParam, QueryParam, HeaderParam, PathParam, MiddlewareParam work, and raise exception if invalid input are in

- [x] Test case(s) to:
- [ ] Documentation

----
📚 Documentation preview 📚: https://sorna--3511.org.readthedocs.build/en/3511/

----
📚 Documentation preview 📚: https://sorna-ko--3511.org.readthedocs.build/ko/3511/

### DIFF
--- a/src/ai/backend/common/api_handlers.py
+++ b/src/ai/backend/common/api_handlers.py
@@ -120,7 +120,7 @@ JSONDict = dict[str, Any]
 
 
 @dataclass
-class ApiResponse:
+class APIResponse:
     _status_code: int
     _data: Optional[BaseResponseModel]
 
@@ -220,9 +220,9 @@ async def _parse_and_execute_handler(request: web.Request, handler, signature) -
 
     response = await handler(**handler_params.get_all())
 
-    if not isinstance(response, ApiResponse):
+    if not isinstance(response, APIResponse):
         raise InvalidAPIParameters(
-            f"Only Response wrapped by ApiResponse Class can be handle: {type(response)}"
+            f"Only Response wrapped by APIResponse Class can be handle: {type(response)}"
         )
 
     return web.json_response(
@@ -240,25 +240,25 @@ def api_handler(handler):
         async def handler(body: BodyParam[UserModel]):  # UserModel is a Pydantic model
             user = body.parsed                          # 'parsed' property gets pydantic model you defined
             # Response model should inherit BaseResponseModel
-            return ApiResponse.build(status_code=200, response_model=YourResponseModel(user=user.id))
+            return APIResponse.build(status_code=200, response_model=YourResponseModel(user=user.id))
 
     2. Query Parameters:
         @api_handler
         async def handler(query: QueryParam[QueryPathModel]):
             parsed_query = query.parsed
-            return ApiResponse.build(status_code=200, response_model=YourResponseModel(search=parsed_query.query))
+            return APIResponse.build(status_code=200, response_model=YourResponseModel(search=parsed_query.query))
 
     3. Headers:
         @api_handler
         async def handler(headers: HeaderParam[HeaderModel]):
             parsed_header = headers.parsed
-            return ApiResponse.build(status_code=200, response_model=YourResponseModel(data=parsed_header.token))
+            return APIResponse.build(status_code=200, response_model=YourResponseModel(data=parsed_header.token))
 
     4. Path Parameters:
         @api_handler
         async def handler(path: PathModel = PathParam(PathModel)):
             parsed_path = path.parsed
-            return ApiResponse.build(status_code=200, response_model=YourResponseModel(path=parsed_path))
+            return APIResponse.build(status_code=200, response_model=YourResponseModel(path=parsed_path))
 
     5. Middleware Parameters:
         # Need to extend MiddlewareParam and implement 'from_request'
@@ -274,7 +274,7 @@ def api_handler(handler):
 
         @api_handler
         async def handler(auth: AuthMiddlewareParam):  # No generic, so no need to call 'parsed'
-            return ApiResponse(status_code=200, response_model=YourResponseModel(author_name=auth.name))
+            return APIResponse(status_code=200, response_model=YourResponseModel(author_name=auth.name))
 
     6. Multiple Parameters:
         @api_handler
@@ -284,7 +284,7 @@ def api_handler(handler):
             headers: HeaderParam[HeaderModel],  # headers
             auth: AuthMiddleware,  # middleware parameter
         ):
-            return ApiResponse(
+            return APIResponse(
                 status_code=200,
                 response_model=YourResponseModel(
                     user=user.parsed.user_id,
@@ -296,7 +296,7 @@ def api_handler(handler):
 
     Note:
     - All parameters must have type hints or wrapped by Annotated
-    - Response class must be ApiResponse and your response model should inherit BaseResponseModel
+    - Response class must be APIResponse and your response model should inherit BaseResponseModel
     - Request body is parsed must be json format
     - MiddlewareParam classes must implement the from_request classmethod
     """
--- a/src/ai/backend/common/api_handlers.py
+++ b/src/ai/backend/common/api_handlers.py
@@ -150,6 +150,10 @@ class APIResponse:
     def to_json(self) -> Optional[JSONDict]:
         return self._data.model_dump(mode="json") if self._data else None
 
+    @property
+    def status_code(self) -> int:
+        return self._status_code
+
 
 _ParamType: TypeAlias = BodyParam | QueryParam | PathParam | HeaderParam | MiddlewareParam
 
@@ -252,13 +256,14 @@ async def _parse_and_execute_handler(
 
     return web.json_response(
         response.to_json,
-        status=response._status_code,
+        status=response.status_code,
     )
 
 
 def api_handler(handler: BaseHandler) -> ParsedRequestHandler:
     """
     This decorator processes HTTP request parameters using Pydantic models.
+    NOTICE: API hander methods must be classmethod. It handlers are not class methods it will not work as intended
 
     1. Request Body:
         @api_handler
@@ -330,13 +335,6 @@ def api_handler(handler: BaseHandler) -> ParsedRequestHandler:
 
     @functools.wraps(handler)
     async def wrapped(first_arg: Any, *args, **kwargs) -> web.Response:
-        if isinstance(first_arg, web.Request):
-            return await _parse_and_execute_handler(
-                request=first_arg, handler=handler, signature=original_signature
-            )
-
-        # If handler is method defined in class
-        # Remove 'self' in parameters
         instance = first_arg
         sanitized_signature = original_signature.replace(
             parameters=list(original_signature.parameters.values())[1:]
--- a/src/ai/backend/common/api_handlers.py
+++ b/src/ai/backend/common/api_handlers.py
@@ -192,7 +192,7 @@ class _HandlerParameters:
         return self._params
 
 
-async def _pydantic_handler(request: web.Request, handler, signature) -> web.Response:
+async def _parse_and_execute_handler(request: web.Request, handler, signature) -> web.Response:
     handler_params = _HandlerParameters()
     for name, param in signature.parameters.items():
         # If handler has no parameter, for loop is skipped
@@ -224,31 +224,31 @@ async def _pydantic_handler(request: web.Request, handler, signature) -> web.Res
     )
 
 
-def pydantic_api_handler(handler):
+def api_handler(handler):
     """
     This decorator processes HTTP request parameters using Pydantic models.
 
     1. Request Body:
-        @pydantic_api_handler
+        @api_handler
         async def handler(body: BodyParam[UserModel]):  # UserModel is a Pydantic model
             user = body.parsed                          # 'parsed' property gets pydantic model you defined
             # Response model should inherit BaseResponseModel
             return ApiResponse.build(status_code=200, response_model=YourResponseModel(user=user.id))
 
     2. Query Parameters:
-        @pydantic_api_handler
+        @api_handler
         async def handler(query: QueryParam[QueryPathModel]):
             parsed_query = query.parsed
             return ApiResponse.build(status_code=200, response_model=YourResponseModel(search=parsed_query.query))
 
     3. Headers:
-        @pydantic_api_handler
+        @api_handler
         async def handler(headers: HeaderParam[HeaderModel]):
             parsed_header = headers.parsed
             return ApiResponse.build(status_code=200, response_model=YourResponseModel(data=parsed_header.token))
 
     4. Path Parameters:
-        @pydantic_api_handler
+        @api_handler
         async def handler(path: PathModel = PathParam(PathModel)):
             parsed_path = path.parsed
             return ApiResponse.build(status_code=200, response_model=YourResponseModel(path=parsed_path))
@@ -265,12 +265,12 @@ def pydantic_api_handler(handler):
                 user_email = request["user"]["email"]
                 return cls(user_id=user_id)
 
-        @pydantic_api_handler
+        @api_handler
         async def handler(auth: AuthMiddlewareParam):  # No generic, so no need to call 'parsed'
             return ApiResponse(status_code=200, response_model=YourResponseModel(author_name=auth.name))
 
     6. Multiple Parameters:
-        @pydantic_api_handler
+        @api_handler
         async def handler(
             user: BodyParam[UserModel],  # body
             query: QueryParam[QueryModel],  # query parameters
@@ -299,7 +299,7 @@ def pydantic_api_handler(handler):
     @functools.wraps(handler)
     async def wrapped(request: Any, *args, **kwargs) -> web.Response:
         if isinstance(request, web.Request):
-            return await _pydantic_handler(request, handler, original_signature)
+            return await _parse_and_execute_handler(request, handler, original_signature)
 
         # If handler is method defined in class
         # Remove 'self' in parameters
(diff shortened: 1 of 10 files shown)


src/ai/backend/common/exception.py
```diff
@@ -110,3 +113,81 @@ class VolumeUnmountFailed(RuntimeError):
     """
     Represents a umount process failure.
     """
+
+
+class BackendError(web.HTTPError):
+    """
+    An RFC-7807 error class as a drop-in replacement of the original
+    aiohttp.web.HTTPError subclasses.
+    """
```
HyeockJinKim, line 122: Ah, BackendError wasn’t in common package… 
For now, I’ll apply it this way, and I’ll create a separate issue to organize exceptions and refactor later.

src/ai/backend/common/pydantic_handlers.py
```diff
@@ -0,0 +1,265 @@
...
+            except ValidationError:
+                raise MiddlewareParamParsingFailed(f"Failed while parsing {parsed_signature.name}")
+
+        # If origin type name is BodyParam/QueryParam/HeaderParam/PathParam
+        origin_name = get_origin(param_type).__name__
+        pydantic_model = get_args(param_type)[0]
+        param_instance = param_type(pydantic_model)
+
+        match origin_name:
+            case "BodyParam":
```
HyeockJinKim: Is it difficult to use type information instead of `origin_name`?

src/ai/backend/common/pydantic_handlers.py
```diff
@@ -0,0 +1,265 @@
...
+        handler_params.add(name, value)
+
+    response = await handler(**handler_params.get_all())
+
+    if not isinstance(response, BaseResponse):
+        raise InvalidAPIParameters(
+            f"Only Response wrapped by BaseResponse Class can be handle: {type(response)}"
+        )
+
+    return web.json_response(response.data.model_dump(mode="json"), status=response.status_code)
```
HyeockJinKim: Shouldn’t it be a private function?

tests/common/test_pydantic_handlers.py
```diff
@@ -0,0 +1,304 @@
...
+    data = await resp.json()
+    assert data["authorization"] == "Bearer token123"
+
+
+class UserPathModel(BaseModel):
+    user_id: str
+
+
+class UserPathResponse(BaseModel):
+    user_id: str
```
HyeockJinKim: Please make it clear in the naming that the mocked request and response are for testing purposes.

seedspirit: I'll add prefix 'Test' in request and response models

src/ai/backend/common/pydantic_handlers.py
seedspirit: Yes. I will change function name

src/ai/backend/common/exception.py
seedspirit: I'll keep it in mind until the next refactoring

src/ai/backend/common/pydantic_handlers.py
seedspirit: @HyeockJinKim 
I think it can be achieved by `if origin_type is BodyParam`. However, since it is not recommended to use `is` to compare between classes, how about make enum, register it as a class variable and compare them? 

```python
class _ParamType(Enum):
    BODY = auto()
    QUERY = auto()
    HEADER = auto()
    PATH = auto()
    MIDDLEWARE = auto()

class BodyParam(Generic[T]):
    _param_type: _ParamType = _ParamType.BODY
    
 
async def _extract_param_value(
    request: web.Request, parsed_signature: _ParsedSignature
) -> Optional[Any]:
    try:
        input_param_type = parsed_signature.input_param_type

        origin_type = get_origin(input_param_type)

        if not hasattr(origin_type, "_param_type"):
            raise UnknownRequestParamType

        match origin_type._param_type:
            case _ParamType.BODY:
                return param_instance.from_body(body)
            case _ParamType.QUERY:
                return param_instance.from_query(request.query)
            case _ParamType.HEADER:
                return param_instance.from_header(request.headers)
            case _ParamType.PATH:
                return param_instance.from_path(request.match_info)
```

HyeockJinKim: I think using a match case would make it easier to distinguish types. Why is it necessary to introduce an enum?
for example:
```python
def match_case_example(param):
    match param:
        case BodyParam():
            param.from_body()
        case QueryParam():
            param.from_query()
        case PathParam():
            param.from_path()
        case HeaderParam():
            param.from_header()
```

src/ai/backend/common/pydantic_handlers.py
```diff
@@ -0,0 +1,254 @@
...
+    @classmethod
+    @abstractmethod
+    def from_request(cls, request: web.Request) -> Self:
+        pass
+
+
+@dataclass
+class BaseResponse:
+    data: BaseModel
+    status_code: int = 200
```
HyeockJinKim: I think it would be better not to set a default value and instead explicitly specify the status code for each handler.

src/ai/backend/common/pydantic_handlers.py
```diff
@@ -0,0 +1,254 @@
...
+                f"Parameter '{input_param_type}' must use one of QueryParam, PathParam, HeaderParam, MiddlewareParam, BodyParam"
+            )
+
+    except ValidationError as e:
+        raise InvalidAPIParameters(str(e))
+
+
+class _HandlerParameters:
+    def __init__(self) -> None:
+        self.params: dict[str, Any] = {}
```
HyeockJinKim: Please declare all the fields used in the class.

src/ai/backend/common/pydantic_handlers.py
```diff
@@ -0,0 +1,254 @@
...
+5. Middleware Parameters:
+    # Need to extend MiddlewareParam and implement 'from_request'
+    class AuthMiddlewareParam(MiddlewareParam):
+        user_id: str
+        user_email: str
+        @classmethod
+        def from_request(cls, request: web.Request) -> Self:
+            # Extract and validate data from request
+            user_id = request["user"]["uuid"]
+            user_email = request["user"]["email"]
```
HyeockJinKim: This comment should be added as a comment in the function that will be used as a decorator.

src/ai/backend/common/pydantic_handlers.py
```diff
@@ -0,0 +1,254 @@
...
+
+class BodyParam(Generic[T]):
+    def __init__(self, model: Type[T]) -> None:
+        self.model = model
+        self._parsed: Optional[T] = None
+
+    @property
+    def parsed(self) -> T:
+        if not self._parsed:
+            raise ValueError("Data not yet parsed")
```
HyeockJinKim: How about defining a custom exception internally instead of using a ValueError?

src/ai/backend/common/pydantic_handlers.py
HyeockJinKim: If the value should not be accessed outside the class, please set it as private.

src/ai/backend/common/pydantic_handlers.py
seedspirit: Okay, I will replace them to custom exceptions

src/ai/backend/common/pydantic_handlers.py
seedspirit: I also think defining specific status code for each handler would be more explicit and prevent unintended behavior. I will fix it

src/ai/backend/common/api_handlers.py
```diff
@@ -0,0 +1,316 @@
...
+    response = await handler(**handler_params.get_all())
+
+    if not isinstance(response, ApiResponse):
+        raise InvalidAPIParameters(
+            f"Only Response wrapped by ApiResponse Class can be handle: {type(response)}"
+        )
+
+    return web.json_response(
+        response._data.model_dump(mode="json") if response._data else {},
+        status=response._status_code,
```
HyeockJinKim: Please prevent external access to private values. It would be better to use a property for that.

HyeockJinKim: If {} is used as the default value, it won’t guarantee no content.

seedspirit: I have encapsulated it using `@property`. Could you take a look?

src/ai/backend/common/api_handlers.py
```diff
@@ -0,0 +1,323 @@
...
+            )
+
+        handler_params.add(name, value)
+
+    response = await handler(**handler_params.get_all())
+
+    if not isinstance(response, APIResponse):
+        raise InvalidAPIParameters(
+            f"Only Response wrapped by APIResponse Class can be handle: {type(response)}"
+        )
```
fregataa, line 255: We don't need such checking if we apply type hint to the handler function!

src/ai/backend/common/api_handlers.py
```diff
@@ -0,0 +1,323 @@
...
+            raise InvalidAPIParameters(
+                f"Type hint or Annotated must be added in API handler signature: {param.name}"
+            )
+
+        value = await _extract_param_value(request=request, input_param_type=param.annotation)
+
+        if not value:
+            raise InvalidAPIParameters(
+                f"Type hint or Annotated must be added in API handler signature: {param.name}"
+            )
```
fregataa, line 246: Is this needed? `value` should be `_ParamType`

src/ai/backend/common/api_handlers.py
```diff
@@ -0,0 +1,323 @@
...
+
+    3. Headers:
+        @api_handler
+        async def handler(headers: HeaderParam[HeaderModel]):
+            parsed_header = headers.parsed
+            return APIResponse.build(status_code=200, response_model=YourResponseModel(data=parsed_header.token))
+
+    4. Path Parameters:
+        @api_handler
+        async def handler(path: PathModel = PathParam(PathModel)):
```
fregataa:
```suggestion
        async def handler(path: PathParam[PathModel]):
```
I think this needs a correction 👀

fregataa (commented):
1. How about allowing `api_handler` decorator to class methods only? Allowing both class methods and general functions makes the decorator do weird signature manipulation like `self = request` after if-branch.
2. We can apply better type hints on the api handler function
3. The model field of `...Param` types is restricted to pydantic object types. It seems like the type for the model field could also be generalized - are there any plans for this?
4. Why should the api handler functions have `Keypair` as a parameter?

src/ai/backend/common/api_handlers.py
seedspirit: Even if we put in a type hint, shouldn't we put in some defense against using classes other than APIResponse in the api handler?

src/ai/backend/common/api_handlers.py
seedspirit: I put the defense logic in just in case (the message is wrong), but I also think it's unnecessary since we already prevent the case of no value in `_extract_param_value` and `_ParamType`

src/ai/backend/common/api_handlers.py
seedspirit: Yes, It should be fixed

src/ai/backend/common/api_handlers.py
HyeockJinKim: It seems that _status_code is still in use.

src/ai/backend/common/api_handlers.py
HyeockJinKim: In MiddlewareParam, since it’s implemented via inheritance, it seems possible that None could be returned.

HyeockJinKim (approved):

src/ai/backend/common/api_handlers.py
seedspirit: also encapsulated using `@property`

src/ai/backend/common/api_handlers.py
```diff
@@ -0,0 +1,348 @@
...
+    @classmethod
+    def build(cls, status_code: int, response_model: BaseResponseModel) -> Self:
+        return cls(_status_code=status_code, _data=response_model)
+
+    @classmethod
+    def no_content(cls, status_code: int) -> Self:
+        return cls(_status_code=status_code, _data=None)
+
+    @property
+    def to_json(self) -> Optional[JSONDict]:
```
rapsealk, line 150: This doesn't necessarily need to be a `property`, particularly when the method name is in verb form, like `to_json()`.

src/ai/backend/common/api_handlers.py
```diff
@@ -0,0 +1,348 @@
...
+class APIResponse:
+    _status_code: int
+    _data: Optional[BaseResponseModel]
+
+    @classmethod
+    def build(cls, status_code: int, response_model: BaseResponseModel) -> Self:
+        return cls(_status_code=status_code, _data=response_model)
+
+    @classmethod
+    def no_content(cls, status_code: int) -> Self:
```
rapsealk, line 146: The name `no_content` is confusing. I'd suggest `empty`.
Refs
- https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/ResponseEntity.html#noContent()
- https://learn.microsoft.com/ko-kr/dotnet/api/microsoft.aspnetcore.http.httpresults.nocontent?view=aspnetcore-9.0

rapsealk: Alternatively, you could allow `None` for `data` in `build()` and remove `no_content()`.

PR #3404, opened 2025-01-08
chore(BA-469): Rename variables `api_rqst` to `api_request` for clarity

rapsealk:
Jira ticket: BA-469

This pull request focuses on renaming the variable `api_rqst` to `api_request` across multiple files to improve code readability and maintain consistency. The changes are primarily in functions handling web requests and responses.

Variable renaming for consistency:

* [`src/ai/backend/client/cli/app.py`](diffhunk://#diff-b3521a04216dddc50de36cc5b730ee7fdc39ceb87d84b20e3b1fa4976d2cbe62L60-R61): Renamed `api_rqst` to `api_request` in the `run` method.
* [`src/ai/backend/client/cli/proxy.py`](diffhunk://#diff-60899f6e10ee0689703de40a634b6b5c0d791abe8dd405fe5c070e34018d266aL124-R135): Renamed `api_rqst` to `api_request` in the `web_handler` and `websocket_handler` methods. [[1]](diffhunk://#diff-60899f6e10ee0689703de40a634b6b5c0d791abe8dd405fe5c070e34018d266aL124-R135) [[2]](diffhunk://#diff-60899f6e10ee0689703de40a634b6b5c0d791abe8dd405fe5c070e34018d266aL167-R175)
* [`src/ai/backend/client/cli/session/app.py`](diffhunk://#diff-2f4cf0e53a688560153dd84875c45b544a22e7638bde977875a9374814e0475bL58-R59): Renamed `api_rqst` to `api_request` in the `run` method.
* [`src/ai/backend/client/func/session.py`](diffhunk://#diff-e680c622eecfe90ccf44a6b880343a40b88739e06ebfb295462ddf0452093a30L1054-R1059): Renamed `api_rqst` to `api_request` in the `stream_app_info` method.
[...]

PR #3522, opened 2025-01-22
chore(BA-469): Rename variables `api_rqst` to `api_request` for clarity (#3404)

lablup-octodog:
This is an auto-generated backport PR of #3404 to the 24.03 release.


HyeockJinKim:
This is a variable name change, so it will not be applied to the previous version.

PR #3377, opened 2025-01-05
chore(BA-458): Upgrade pants to 2.23.1 and replace the scie plugin with intrinsic pex option

achimnol:
resolves #3385 (BA-458): This PR upgrades Pantsbuild from 2.21 to 2.23 release, and also upgrades PEX from 2.10 to 2.28.1 alongside.  Thanks to https://docs.pex-tool.org/scie.html ([a-scie/lift#116](https://github.com/a-scie/lift/issues/116#issuecomment-2571394474)), we can now use PEX to build scie binaries.

The new BUILD configs preserve the same command format, so we don't have to update the CI and documentation. Just run the same way: `pants --tag=scie --tag=+lazy package ::`.

The most significant advantage is that now pants tracks the file changes properly and the scie binaries will be rebuilt if you modify the source files in the working copy, which greatly eases repeated testing of the scie binaries in development setups.

refs #3285 (BA-4): Upgrading Pantsbuild also enables Python 3.13 support added in Pants v2.22.

Additionally, this PR does:
- Strip out unimplemented details from the towncrier plugin to adapt with the pants plugin API changes.
- Update the `build-scies` job in the GHA workflow to properly suffix the scie binaries.
  - The latest PEX provides `--scie-name-style` option to automate this suffix attachment, but it confuses Pantsbuild to pick up the generated output file.  Hence, we do the required suffixing by oursevles.
  - Add `scripts/get-platform-suffix.py` for use in the workflow.
- Let the `build-scies` job merge the checksums into a single file (`checksum.txt`) to reduce the number of GitHub release artifacts.
- Refactor the BUILD files and move the common scie-specific config options for `pex_binary` targets to `tools/build-macros.py`.
- Add missing cross-import visibility rule for `ai.backend.account_manager`.

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version
- [x] Mention to the original issue

PR #3535, opened 2025-01-23
ci: Unify suffixes in release names the same way as before

### LINKED ISSUE #3534: Unify suffixes in release names the same way as before
## Motivation  

- [before]
  {width=1824 height=142}
  [current]
  {width=1804 height=258}
  The suffix for kernel information is different than before.

## Objective  

- Fix it to spell macos the same as before.

## Details  

1. 
1. 
1. 

## Impact  

-

Yaminyam:
resolves #3534 (BA-605)
I will also need to rename all of the release files for the 25.1.1 version that were incorrectly registered in this PR.

### DIFF
--- a/scripts/get-platform-suffix.py
+++ b/scripts/get-platform-suffix.py
@@ -11,7 +11,13 @@ arch_name_aliases = {
     "i686": "x86",  # Windows
 }
 
-platform_kernel = sys.platform
+kernel_name_aliases = {
+    "linux": "linux",
+    "darwin": "macos",
+    "win32": "windows",
+}
+
+platform_kernel = kernel_name_aliases.get(sys.platform, sys.platform)
 platform_arch = arch_name_aliases.get(platform.machine(), platform.machine())
 
 print(f"{platform_kernel}-{platform_arch}")


scripts/get-platform-suffix.py
```diff
@@ -11,7 +11,13 @@
...
 }
 
-platform_kernel = sys.platform
+kernel_name_aliases = {
+    "linux": "linux",
+    "darwin": "macos",
+    "win32": "windows",
+}
+
+platform_kernel = kernel_name_aliases.get(sys.platform, sys.platform)
```
HyeockJinKim, line 20: This seems likely to resolve the issue.
However, in the past, it seems that the name was matched without an alias. Do you know why the name has been changed?

Yaminyam: Previously, we were building with the SCIE plugin, but recently PEX included SCIE-based builds to replace that functionality.(ref. #3377)
There was an issue where if pantsbuild varied the output filename specified, pants would not recognize that it had output normally and would clean it up.
So we added scripts/get-platform-suffix.py to append the suffix after pantsbuild has finished.

HyeockJinKim: Could you please run the test and share the results?

Yaminyam: CleanShot 2025-01-24 at 15 20 11@2x
Yes I just ran a test locally.

HyeockJinKim (approved):
lgtm

PR #3555, opened 2025-01-30
feat(BA-615): Collect metrics for the RPC server

### LINKED ISSUE #3549: Add Prometheus metrics for RPC server in agent component
Implement Prometheus metrics collection for the RPC server within the agent component. This will involve identifying key performance indicators and setting up the necessary instrumentation to capture these metrics.

HyeockJinKim:
resolves #3549 (BA-615)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/agent/server.py
+++ b/src/ai/backend/agent/server.py
@@ -213,7 +213,7 @@ def _collect_metrics(observer: RPCMetricObserver) -> Callable:
                     duration=duration,
                 )
                 return res
-            except Exception as e:
+            except BaseException as e:
                 duration = time.perf_counter() - start_time
                 observer.observe_rpc_request_failure(
                     method=meth.__name__,
--- a/src/ai/backend/agent/server.py
+++ b/src/ai/backend/agent/server.py
@@ -12,6 +12,7 @@ import shutil
 import signal
 import ssl
 import sys
+import time
 from collections import OrderedDict, defaultdict
 from datetime import datetime, timezone
 from ipaddress import IPv4Address, IPv6Address, ip_network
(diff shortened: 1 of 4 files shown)


src/ai/backend/agent/server.py
```diff
@@ -195,6 +200,33 @@ async def _inner(self_: AgentRPCServer, request: RPCMessage) -> Any:
...
+            start_time = time.perf_counter()
+            try:
+                res = await meth(self, *args, **kwargs)
+                duration = time.perf_counter() - start_time
+                observer.observe_rpc_request_success(
+                    method=meth.__name__,
+                    duration=duration,
+                )
+                return res
+            except Exception as e:
```
fregataa: How about handling `BaseException` here?

src/ai/backend/agent/server.py
```diff
@@ -162,15 +164,18 @@ async def _inner(self: AgentRPCServer, *args, **kwargs):
 
 class RPCFunctionRegistry:
     functions: Set[str]
+    _metric_observer: RPCMetricObserver
 
     def __init__(self) -> None:
         self.functions = set()
+        self._metric_observer = RPCMetricObserver.instance()
```
fregataa, line 171: Is it used as a singleton object? Is it expected to use this observer in other context?

src/ai/backend/agent/server.py
```diff
@@ -195,6 +200,33 @@ async def _inner(self_: AgentRPCServer, request: RPCMessage) -> Any:
         return _inner
 
 
+def _collect_metrics(observer: RPCMetricObserver) -> Callable:
+    def decorator(meth: Callable) -> Callable[[AgentRPCServer, RPCMessage], Any]:
+        @functools.wraps(meth)
+        async def _inner(self: AgentRPCServer, *args, **kwargs) -> Any:
```
fregataa, line 206: We can apply more strict type hints here but then we have to update the type hints of `RPCFunctionRegistry.__call__()`.
Let's leave it as a minor issue

src/ai/backend/agent/server.py
HyeockJinKim: This code was added to handle a case where two or more gauges are initialized, which causes panic in prometheus_client.

src/ai/backend/agent/server.py
HyeockJinKim: Can you review again?

PR #3570, opened 2025-02-03
fix(BA-630): Create `./volumes/grafana-data` directory with 757 permissions

### LINKED ISSUE #3568: Permission denied when creating a directory for Grafana on Ubuntu Linux
## Summary  

During installation [Backend.AI](http://Backend.AI) dev on Ubuntu Linux, a permission denied error is raised as below:

{width=2908 height=1546}

## Steps to Reproduce  

On Ubuntu Linux 22.04:

1. $ git clone https://github.com/lablup/backend.ai
1. $ cd backend.ai
1. $ ./script/install-dev.sh

## Expected Behavior  

Installation succeeds with no error.

## Actual Behavior  

Grafana container dies.

## Logs/Errors  

```
mkdir: can't create directory '/var/lib/grafana/plugins': Permission denied
```

rapsealk:
resolves #3568 (BA-630)

It has been reported that the Grafana container fails to bootstrap because it cannot create `/var/lib/grafana/plugins` directory due to a lack of proper permissions.

https://github.com/lablup/backend.ai/blob/a10db5d776219690162140c2c2bb2faacf008dab/docker-compose.halfstack-2409.yml#L78

In the current `docker-compose` configuration, the `/var/lib/grafana` directory inside the container is mapped to the `./volumes/grafana-data` directory on the host machine. This results in permission issues, as the container is unable to create necessary directories or files in the host-mounted volume.

Therefore, this pull request modifies the `scripts/install-dev.sh` script to ensure that the `./volumes/grafana-data` directory on the host machine is created with **757**(rwxr-xrwx) permissions, which grants the necessary privileges for the Grafana container to successfully create and write to the required directories.


scripts/install-dev.sh
```diff
@@ -849,6 +849,7 @@ setup_environment() {
   mkdir -p "${HALFSTACK_VOLUME_PATH}/postgres-data"
   mkdir -p "${HALFSTACK_VOLUME_PATH}/etcd-data"
   mkdir -p "${HALFSTACK_VOLUME_PATH}/redis-data"
+  mkdir -p "${HALFSTACK_VOLUME_PATH}/grafana-data" -m 757
```
HyeockJinKim, line 852: What is the reason for setting it to 757?

rapsealk: The default permission for the directory was set to **755**, but using **--7** was necessary to create the new directory `/var/lib/grafana/plugins` from within the container. However, manually setting permissions for a directory introduces potential security risks. Instead, I plan to configure the user directly in the Docker Compose file as an alternative solution.

Reference: [Grafana Docker Installation with Persistent Storage Permission Issue](https://community.grafana.com/t/new-docker-install-with-persistent-storage-permission-problem/10896/13)

HyeockJinKim (approved):
This approach seems simpler compared to other.

PR #3576, opened 2025-02-04
feat(BA-114): Restore unmanaged vfolder functionality

### LINKED ISSUE #2313: Revisit the unmanaged vfolder with the latest storage-proxy architecture
Unmanaged vfolders allow superadmins to map a random host path to a vfolder, in case that users need to cooperate with external systems which use a specific host path or sub-path in shared network volumes.

image

Issue: unmanaged vfolders do not work with the current backend-aware storage-proxy implementation with explicit vfolder host configurations, as the vfolder-host information is missing for them.

Let's update the implementation as follows:

- If the host path of an unmanaged vfolder contains a known vfolder host's mount path, let's follow the configuration of the matched vfolder host.
- Otherwise, let's introduce a concept of "ghost vfolder host", which provides a default vfolder-host permission configurations.

fregataa:
resolves #2313 (BA-114)


fregataa:
Close this PR since the issue has been redefined as an epic

Issue #3575, opened 2025-02-04
The checksum file is integrated and the installer is not working


Yaminyam:
## Motivation  

[Why is this Task important? Explain the context and reasons for this task.]  

- [https://github.com/lablup/backend.ai/blob/6556f26630f1068de37cc568e36bd331d00b3630/src/ai/backend/install/context.py#L996](https://github.com/lablup/backend.ai/blob/6556f26630f1068de37cc568e36bd331d00b3630/src/ai/backend/install/context.py#L996|smart-link) 
- Previously, checksum files were downloaded for each package, but since version 25, the checksum files have been consolidated into a single file, which is not working.
  ![CleanShot 2025-02-04 at 14.52.42@2x-20250204-055246.png](CleanShot 2025-02-04 at 14.52.42@2x-20250204-055246.png){width=1175 height=347}

## Objective  

[Clearly state the purpose or goal of this task.]  

- Change the installer to download and use the combined checksum.

## Details  

[Outline the specific steps or changes needed to complete this task.]  

1. 
1. 
1. 

## Impact  

[Describe how this task may indirectly influence other services, if at all.]  

-

PR #3577, opened 2025-02-04
ci: Merge checksum files into one

Yaminyam:
#3575 (BA-635)

Since the build-scies job runs as a matrix, it will be overwritten by checksum.txt when uploading the artifact with all the same names.
We need to keep all the values, so we use platform-suffix to separate the files and have them merged into one file in the final step.


HyeockJinKim (approved):
lgtm

PR #3583, opened 2025-02-05
fix(BA-641): Check intrinsic time files exist before mount

### LINKED ISSUE #3582: Cannot create kernels on RHEL system agents
## Summary  

Cannot create compute kernels on agents installed in RHEL OS machines.

It is because an agent inject intrinsic files (`/etc/timezone` and `/etc/localtime`) but RHEL OS does not have the files.

## How to fix

Let’s check if the machine has the files

fregataa:
resolves #3582 (BA-641)


HyeockJinKim (approved):
This is a temporary measure, but I will merge it for now.

PR #3585, opened 2025-02-05
feat: Apply nonce template for requests with specified file extensions

### LINKED ISSUE #3578: Inject a nonce for CSP in the web server.
## Motivation  

- 

## Required Features

- 

## Impact  

- 

## Testing Scenarios  

-

HyeockJinKim:
This PR must be merged only after all the nonce injection work on the web has been reflected.

resolves #3578 (BA-639)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/web/server.py
+++ b/src/ai/backend/web/server.py
@@ -2,6 +2,7 @@ import asyncio
 import json
 import logging
 import logging.config
+import mimetypes
 import os
 import re
 import socket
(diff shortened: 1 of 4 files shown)


src/ai/backend/web/server.py
```diff
@@ -94,6 +109,9 @@ async def static_handler(request: web.Request) -> web.StreamResponse:
             }),
             content_type="application/problem+json",
         )
+    if file_path.suffix in template_extension_list:
```
yomybaby, line 123: I have checked the following two cases:

1. When a request is made without an extension, such as `http://127.0.0.1:8090/`, the response should be HTML. However, it seems that the nonce injection logic is not being applied in this case.
2. When accessing `http://127.0.0.1:8090/index.html`,  extension check condition is not passing. If I remove the condition, it works properly.

PR #3648, opened 2025-02-11
fix(BA-594): Handle cancel and timeout when creating kernels

### LINKED ISSUE #3523: Fix to prevent session hang, even with long bootstrap.
## Summary  

- When the bootstrap script takes too long, the computation session remains stuck in the PREPARING state indefinitely.
- The root cause will be analyzed, and a fix will be implemented.

## Steps to Reproduce  

1. If the bootstrap process takes more than 1 minute, the computation session hangs.

## Expected Behavior  

- Even if the bootstrap process takes a long time, the computation session should not hang.

## Actual Behavior  

- When the bootstrap process takes too long, the computation session hangs and does not function properly.

##

fregataa:
resolves #3523 (BA-594)


src/ai/backend/manager/exceptions.py
```diff
@@ -102,7 +103,7 @@ class ErrorStatusInfo(TypedDict):
 
 
 def convert_to_status_data(
-    e: Exception,
+    e: Exception | asyncio.CancelledError,
     is_debug: bool = False,
```
HyeockJinKim: I think it would be better to receive BaseException and create a status_data.

PR #3679, opened 2025-02-13
fix(BA-725): Increase Backend.AI Kernel's app startup timeout (#3672)

### LINKED ISSUE #3672: Increase Backend.AI Kernel's app startup timeout
## Summary  

- The timeout value should be increased substantially. Determining the ideal value will be controversial, but current setting (5 seconds) is too small.

## Steps to Reproduce  

1. Try to start the  JupyterLab from NGC Torch 24.09 image on Lablup’s Internal Cloud.

## Expected Behavior 

- App should start fine

## Actual Behavior  

- App does not start - and server returns `500` on `start-service`

## Logs/Errors  

`N/A`

MintCat98:
resolves #3672  (BA-725)

Increased the kernel-timeout value in `kernel/base.py`:
- prev: `5.0s`
- curr: `30.0s`

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/kernel/base.py
+++ b/src/ai/backend/kernel/base.py
@@ -862,7 +862,7 @@ class BaseRunner(metaclass=ABCMeta):
                         self.services_running[service_info["name"]] = proc
                         asyncio.create_task(self._wait_service_proc(service_info["name"], proc))
                         if not do_not_wait:
-                            async with asyncio.timeout(20.0):
+                            async with asyncio.timeout(30.0):
                                 await wait_local_port_open(service_info["port"])
                         log.info(
                             "Service {} has started (pid: {}, port: {})",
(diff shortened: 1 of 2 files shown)


src/ai/backend/kernel/base.py
```diff
@@ -862,7 +862,7 @@ async def _start_service(
                         self.services_running[service_info["name"]] = proc
                         asyncio.create_task(self._wait_service_proc(service_info["name"], proc))
                         if not do_not_wait:
-                            async with asyncio.timeout(5.0):
+                            async with asyncio.timeout(20.0):
```
kyujin-cho: Could you update the value a little bit bigger, like 30 seconds?

kyujin-cho (approved):

src/ai/backend/kernel/base.py
MintCat98: Ok I'll revise it right away.

kyujin-cho (approved):

PR #3681, opened 2025-02-13
feat(BA-499): implement api for downloading multiple files

### LINKED ISSUE #3445: Implement an API to stream a ZIP file based on a JWT from storage-proxy
## Motivation  

- Currently, even though there are many files, only one file can be downloaded at a time. For convenience, it seems necessary to implement a feature that allows multiple files to be downloaded at once.

## Required Features

- When multiple files are selected, I plan to send a single `ZIP` file containing those files to the client (there’s an existing function that streams ZIP files, so I intend to reuse that).
- To compress multiple files, we need a common parent directory for the files, so I intend to change the folder structure as follows:
  - Target
    folder/file.png
  - Current
    folder.zip ~~extract~~> file.png
  - Updated
    folder.zip ~~extract~~> folder/file.png

## Impact  

- Using the token received from `storage`, I intend to update the existing `download_directory_as_archive` function so that it can handle not only directories but also multiple files when creating a ZIP file.

## Testing Scenarios  

- Test if a single file/folder is returned correctly files as individual files and folders as ZIP files.
- Test if multiple files (files + folders) are returned in a ZIP file named after the common parent folder.

JIRA Issue: BA-499

nowgnuesLee:
resolves #3445 [(BA-499)](https://lablup.atlassian.net/browse/BA-499)

**changes**
* add `download_multi` function to `storage/api/client.py` to stream an archive file on the fly.

### DIFF
--- a/src/ai/backend/storage/api/client.py
+++ b/src/ai/backend/storage/api/client.py
@@ -301,60 +301,58 @@ async def download_multi(request: web.Request) -> web.StreamResponse:
                 vfpath = Path(token_data["unmanaged_path"])
             else:
                 vfpath = volume.mangle_vfpath(token_data["vfid"])
-                parent_dir = vfpath
-                if (dst_dir := params["dst_dir"]) is not None:
-                    parent_dir = vfpath / dst_dir
-                zf = zipstream.ZipFile(compression=zipstream.ZIP_DEFLATED)
-                for path in token_data["relpathList"]:
-                    try:
-                        file_path = parent_dir / path
-                        file_path.resolve().relative_to(vfpath)
-                        if not file_path.exists():
-                            raise FileNotFoundError
-                    except (ValueError, FileNotFoundError):
-                        raise web.HTTPNotFound(
-                            body=json.dumps(
-                                {
-                                    "title": "File not found",
-                                    "type": "https://api.backend.ai/probs/storage/file-not-found",
-                                },
-                            ),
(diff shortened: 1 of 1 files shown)


src/ai/backend/storage/api/client.py
```diff
@@ -213,6 +243,120 @@ class Params(TypedDict):
...
+    ) as params:
+        async with ctx.get_volume(params["token"]["volume"]) as volume:
+            token_data = params["token"]
+            if token_data["unmanaged_path"] is not None:
+                vfpath = Path(token_data["unmanaged_path"])
+            else:
+                vfpath = volume.mangle_vfpath(token_data["vfid"])
+            parent_dir = vfpath
+            if (dst_dir := params["dst_dir"]) is not None:
+                parent_dir = vfpath / dst_dir
```
HyeockJinKim, line 306: Define `target_vfolder_path` method.

src/ai/backend/storage/api/client.py
```diff
@@ -213,6 +243,120 @@ class Params(TypedDict):
...
+            zf = zipstream.ZipFile(compression=zipstream.ZIP_DEFLATED)
```
HyeockJinKim, line 307: Don't use abbreviations

HyeockJinKim: Add compression level to config

PR #3693, opened 2025-02-13
feat(BA-442): Distribute model service replica containers whenever possible

### LINKED ISSUE #3349: Develop logic to distribute model service replica containers across different GPU servers whenever possible
## Motivation  

If a scaling group is used for both training and inference, the current pending session scheduling policy of the AgentSelector alone is not sufficient.

For example, there was a customer who wanted to use the **ConcentratedAgentSelector** for training to maximize the number of available GPUs by default, while selectively using the **Dispersed** policy for inference to ensure availability.

## Required Features

- One way to implement the requirement is to add additional branching in `SchedulerDispatcher._schedule_in_sgroup` based on a specific flag or option. For example, instead of simply loading the scheduler corresponding to the `AgentSelectionStrategy`, we could load a **Dispersed** type scheduler if the `PendingSession` is of the **Inference** type (Additionally we could apply further optimizations such as applying Concentrated strategy if there are no related RoutingRows.)

### **IMPACT**  

- However, adding branching can increase the complexity of the scheduling logic.It might be a better choice to introduce an appropriate abstraction layer or explore alternative solutions.

### **TESTING SCENARIOS**  

- Appropriate test code should be added to `test_scheduler.py` to test this PR.

jopemachine:
Resolves #3349 ([BA-442](https://lablup.atlassian.net/browse/BA-442)).

[BA-442]: https://lablup.atlassian.net/browse/BA-442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ai/backend/manager/scheduler/agent_selector.py
+++ b/src/ai/backend/manager/scheduler/agent_selector.py
@@ -164,7 +164,10 @@ class ConcentratedAgentSelector(BaseAgentSelector[NullAgentSelectorState]):
     @property
     @override
     def config_iv(self) -> t.Dict:
-        return t.Dict({"dispersion": t.Dict(t.String, t.Int())}).allow_extra("*")
+        return t.Dict({
+            t.Key("dispersions", optional=True, default=None): t.Null
+            | t.Mapping(t.String, t.Int()),
+        }).allow_extra("*")
 
     @override
     @classmethod
@@ -195,7 +198,7 @@ class ConcentratedAgentSelector(BaseAgentSelector[NullAgentSelectorState]):
             ],
         )
 
-        if dispersions_at_endpoint := self.config.get("dispersion"):
+        if dispersions_at_endpoint := self.config.get("dispersions"):
             agents = sorted(agents, key=lambda agent: dispersions_at_endpoint.get(agent.id, 0))
 
         return agents[0].id
--- a/src/ai/backend/manager/scheduler/agent_selector.py
+++ b/src/ai/backend/manager/scheduler/agent_selector.py
@@ -189,24 +189,27 @@ class ConcentratedAgentSelector(BaseAgentSelector[NullAgentSelectorState]):
         resource_priorities = sort_requested_slots_by_priority(
             requested_slots, self.agent_selection_resource_priority
         )
-        agents = sorted(
+
+        # When not using enforce_spreading_endpoint_replica, treat all agent kernel counts as 0.
+        kernel_counts_at_same_endpoint = (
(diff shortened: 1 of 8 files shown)


src/ai/backend/manager/scheduler/agent_selector.py
```diff
@@ -189,7 +199,14 @@ async def select_agent(
...
             ],
         )
-        return chosen_agent.id
+
+        if self.sgroup_opts.enforce_spreading_endpoint_replica:
+            if kernel_counts_at_same_endpoint := self.config.get("kernel_counts_at_same_endpoint"):
+                agents = sorted(
+                    agents, key=lambda agent: kernel_counts_at_same_endpoint.get(agent.id, 0)
+                )
+
```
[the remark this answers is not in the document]
jopemachine: The two sorting operations performed here are intentional, prioritizing availability over the Concentrated strategy.
However, if availability does not always have to take precedence over the Concentrated strategy, it should be changed to a single sorting operation as suggested by Copilot, and a concrete strategy should be established.

tests/manager/scheduler_utils.py
```diff
@@ -152,6 +152,7 @@ def create_mock_session(
     status_data: dict[str, Any] | None = None,
     kernel_opts: Sequence[KernelOpt] | None = None,
     priority: int = SESSION_PRIORITY_DEFAULT,
+    type: SessionTypes = SessionTypes.BATCH,
```
HyeockJinKim: Don't use `type` name.

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -408,25 +461,59 @@ def _pipeline(r: Redis) -> RedisPipeline:
...
+            global_scheduler_opts = self.shared_config["plugins"]["scheduler"].get(
+                scheduler_name, {}
+            )
+        scheduler_config = {**global_scheduler_opts, **sgroup_opts.config}
+
+        return load_scheduler(scheduler_name, sgroup_opts, scheduler_config)
+
+    async def _load_agent_selector(self, args: LoadAgentSelectorArgs) -> AbstractAgentSelector:
+        db_sess = args.db_sess
+        _scheduler_name, sgroup_opts = await self._get_scaling_group_data(db_sess, args.sgroup_name)
```
HyeockJinKim: Do you need to read `sgroup_opts` every time?

Issue #3418, opened 2025-01-09
Revamp ContainerRegistry API


jopemachine:
Follow-up to #1917.

## Motivation  

- In PR #1917, there was some confusion as the `ContainerRegistryNode` API was changed to no longer use `graphene.InputObjectType`. Let's revert to using `graphene.InputObjectType` for accepting arguments.
- Additionally, PR #3067 implements an API to establish the relationship between `container_registries` and `groups`, and discussions have been held for a more intuitive API design. Let's update the API according to the decision. This will be a breaking change.

## Objective  

- Let's update the API according to the decision.

## Details  

1. First, let's update the `ContainerRegistry` API.
1. Next, let's update APIs for adding and removing associations.

## Impact  

- The WebUI and ControlPanel client code must be updated accordingly.

Mizokuiam:
Thank you for raising this issue! I'll look into it and try to help if I can.

jopemachine:
> Thank you for raising this issue! I'll look into it and try to help if I can.

I appreciate your interest in the issue. 
But, this issue will include breaking changes and needs to be addressed promptly as a prerequisite for other PRs.

So, I would appreciate it if you could look into other issues instead.

PR #3697, opened 2025-02-13
fix(BA-749): Handle `None` arguments properly in `ContainerRegistryNode` GQL API

### LINKED ISSUE #3698: `None` value fields are not being handled correctly in `ContainerRegistryNode` GQL
## Summary  

- Values in the database should not be changed when no value is provided.

## Steps to Reproduce  

1. Update the registry information in the WebUI and call the ModifyContainerRegistryNode API.
1. Check how the values of fields, when not provided, are changed in the database

## Expected Behavior  

- Values in the database should not be changed when no value is provided.

## Actual Behavior  

- Fields not provided, are changed to `NULL` in the database.

## Logs/Errors  

None

JIRA Issue: BA-749

jopemachine:
Resolves #3698 ([BA-749](https://lablup.atlassian.net/browse/BA-749)).

[BA-749]: https://lablup.atlassian.net/browse/BA-749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/3697.feature.md
+++ b/changes/3697.feature.md
@@ -1,0 +1,1 @@
+Handle `None` arguments properly in `ContainerRegistryNode` GQL API.
--- a/src/ai/backend/manager/models/gql_models/container_registry.py
+++ b/src/ai/backend/manager/models/gql_models/container_registry.py
@@ -331,12 +331,12 @@ class CreateContainerRegistryNode(graphene.Mutation):
         url: str,
         type: ContainerRegistryType,
         registry_name: str,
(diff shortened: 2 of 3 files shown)


jopemachine:
If we treat null as a missing value, it causes an issue where we can't insert null into nullable columns like DB `username` and `password`.

This issue can be resolved by using the new V2 mutation introduced in #3418.
So I'll close this PR as not planned.

PR #3714, opened 2025-02-16
feat(BA-759): Revamp api handler

### LINKED ISSUE #3713: Revamp 'api_handler()'
## Motivation  

In version 25.02, we introduced the `ai.backend.common.api_handlers` module to standardize HTTP API handler functions. However, the module has several issues that need to be addressed:

- The name of a generic type variable `T`  does not follow Python naming conventions
- The `_parse_and_execute_handler()` function only accepts `ai.backend.common.APIResponse` objects
- The module does not support `aiohttp.web.StreamResponse` as a return type for API handlers, despite `aiohttp.web.Response` being a subclass of `StreamResponse`
- The type checking implementation for API handler responses is incorrect
- **API handler signatures are validated at runtime when handling requests**, rather than during static analysis or server initialization
- Even when the request is valid, an 'Invalid parameter' error may be raised **if the API handler is incorrectly defined**.

fregataa:
resolves #3713 (BA-759)


src/ai/backend/common/api_handlers.py
```diff
@@ -156,6 +184,7 @@ def status_code(self) -> int:
 
 
 _ParamType: TypeAlias = BodyParam | QueryParam | PathParam | HeaderParam | MiddlewareParam
+_ValidatorType: TypeAlias = BodyParam | QueryParam | PathParam | HeaderParam | type[MiddlewareParam]
```
seedspirit: Just a question, what is the difference between two?

src/ai/backend/common/api_handlers.py
```diff
@@ -249,15 +276,70 @@ async def _parse_and_execute_handler(
...
+
+def _register_parameter_validator(
+    signature: Signature,
+    *,
+    handler_name: str,
+) -> dict[str, _ValidatorType]:
+    signature_validator_map: dict[str, _ValidatorType] = {}
+    for name, param in signature.parameters.items():
+        if param.annotation is inspect.Parameter.empty:
+            raise ValueError(f"Empty type hint is not allowed (handler:{handler_name},name:{name})")
```
seedspirit: Defining and using a custom error rather than simply using ValueError would be better for future error convention modification

src/ai/backend/common/api_handlers.py
```diff
@@ -249,15 +276,70 @@ async def _parse_and_execute_handler(
...
+        signature_validator_map[name] = param_instance
+    return signature_validator_map
+
+
+async def _serialize_parameter(
+    request: web.Request, param_instance_or_class: _ValidatorType
+) -> _ParamType:
+    param_instance: _ParamType
+    match param_instance_or_class:
+        case PathParam() | BodyParam() | HeaderParam() | QueryParam():
```
seedspirit: I would like to ask why you changed from comparing classes to comparing instances?

src/ai/backend/common/api_handlers.py
```diff
@@ -249,15 +276,70 @@ async def _parse_and_execute_handler(
...
+                param_instance = await param_instance_or_class.validate_request(request)
+            except ValidationError as e:
+                raise InvalidAPIParameters(str(e))
+        case _:
+            try:
+                param_instance = param_instance_or_class.from_request(request)
+            except ValidationError:
+                raise MiddlewareParamParsingFailed(
+                    f"Failed while parsing {param_instance_or_class}"
+                )
```
seedspirit: I think there might be some confusion because MiddlewareParamParsingFailed is thrown in all cases of non_ValidatorType - are we trusting the type checker to assume a non_ValidatorType class or instance is coming in?

src/ai/backend/common/api_handlers.py
```diff
@@ -249,15 +276,70 @@ async def _parse_and_execute_handler(
...
+
+def _parse_response(response: ResponseType) -> web.StreamResponse:
+    match response:
+        case APIResponse():
+            return web.json_response(
+                data=response.to_json,
+                status=response.status_code,
+            )
+        case _:
+            return response
```
seedspirit: The initial intent was to only use `APIResponse`, so that only responses of a certain shape (the shape that has a Pydantic Model and a status_code field) are returned. Would it be better to be a little more flexible and allow other kinds of responses to be received?

src/ai/backend/common/api_handlers.py
fregataa: You can initialize all `...Param` types before evaluating request data except `MiddlewareParam`, which requires input data for initializing. `_ValidatorType` represents the types that can evaluate and validate input data.

src/ai/backend/common/api_handlers.py
fregataa: This error occurs when bootstrapping the server, not on runtime if any API handler function is defined incorrectly. You can test it

src/ai/backend/common/api_handlers.py
fregataa: As type checkers check, only ValidatorType input is allowed here. Thats why we use type checkers.

src/ai/backend/common/api_handlers.py
HyeockJinKim: Isn't it safer to only receive APIResponse types, rather than other return values?

fregataa: If none of API handler function will handle `aiohttp.web.StreamResponse` or `aiohttp.web.FileResponse` etc, the current state is okay.
Else, we should change it.
ref: https://docs.aiohttp.org/en/v3.7.3/web_reference.html?highlight=streamresponse#aiohttp.web.StreamResponse

src/ai/backend/common/api_handlers.py
fregataa: It is recommended to use pattern matching rather than using multiple `isinstance()` if it is possible
https://peps.python.org/pep-0636/#adding-a-ui-matching-objects

fregataa:
Todo
1. Do not allow `aiohttp` respose types in api handler
2. Do not allow any other signature types in api handler functions

src/ai/backend/common/api_handlers.py
fregataa: I added an exception for this case

src/ai/backend/common/api_handlers.py
```diff
@@ -26,25 +26,32 @@
...
+TModel = TypeVar("TModel", bound=BaseModel)
 
 
-class BodyParam(Generic[T]):
-    _model: Type[T]
-    _parsed: Optional[T]
+class _HTTPRequestDataValidator:
+    @abstractmethod
+    async def validate_request(self, request: web.Request) -> Self:
+        pass
```
HyeockJinKim: Remove this.

HyeockJinKim (approved):

PR #3725, opened 2025-02-17
feat(BA-24): Split redis config for each connection pool

### LINKED ISSUE #3195: Split Redis configuration for each connection pool by database
## Synopsis

In some customer sites, we have experienced:

- Too high CPU usage of the Redis service
- Excessive disk usage by AOF files, which may be a Redis bug or some subtle config issues:\* redis/redis#10806
- redis/redis#10742

To mitigate these issues, i.e., to prevent them from impacting the SLA and reduce the chance of happening, let's split the Redis instances in our "halfstack" so that we could apply different levels of HA configuration depending on the purpose.

Since we already split the connection pools for each Redis database by different purposes as follows, it is easy to just change the initialization routine of those connection pools to use different connection parameters.

### Redis databases

```python
# src/ai/backend/common/defs.py
REDIS_STAT_DB: Final = 0
REDIS_RLIM_DB: Final = 1
REDIS_LIVE_DB: Final = 2
REDIS_IMAGE_DB: Final = 3
REDIS_STREAM_DB: Final = 4
```

I do not plan to pre-define how to group these Redis databases but leave it to the our solution architect team to determine the desired setup per site.

Though, here are some references to understand the background:

**Data persistency requirements**

- Needed:\* `STAT` (statistics): if lost, it impacts the usage accounting critical for billing and etc.
- `LIVE` (agent liveness, idle checkers): if lost, it impacts the scheduler performance and mis-terminate the running sessions depending on the idle-checker config

- Not much needed:\* `RLIM` (rate limit): volatile information, effective only for up to 15 minutes, no need to enforce it strictly on failover
- `STREAM` (event bus): events are transient and often they are retried periodically or have another mean of synchronization.
- `IMAGE` (per-agent image availability map): if lost, automatically reconstructed from agent heartbeats

**Main load patterns**

- `STAT`, `LIVE`: proportional to the number of sessions
- `RLIM`: proportional to the volume of client API requests
- `IMAGE`: proportional to the number of cluster nodes
- `STREAM`: proportional to the number of cluster nodes & the number of sessions
  ## Current Implementation

The configuration of Redis connection parameters are stored in etcd and shared across the entire cluster nodes including Manager, Agent, Storage Proxy, and App Proxy.

Currently, the Redis configuration can specify only one Redis instance (either a single `host:port` address or a list of sentinel `host:port` addresses).

```python
redis_helper_config_iv = t.Dict({
    t.Key("socket_timeout", default=5.0): t.ToFloat,
    t.Key("socket_connect_timeout", default=2.0): t.ToFloat,
    t.Key("reconnect_poll_timeout", default=0.3): t.ToFloat,
}).allow_extra("*")

redis_config_iv = t.Dict({
    t.Key("addr", default=redis_default_config["addr"]): t.Null | tx.HostPortPair,
    t.Key(  # if present, addr is ignored and service_name becomes mandatory.
        "sentinel", default=redis_default_config["sentinel"]
    ): t.Null | tx.DelimiterSeperatedList(tx.HostPortPair),
    t.Key("service_name", default=redis_default_config["service_name"]): t.Null | t.String,
    t.Key("password", default=redis_default_config["password"]): t.Null | t.String,
    t.Key(
        "redis_helper_config",
        default=redis_helper_default_config,
    ): redis_helper_config_iv,
}).allow_extra("*")
```

## Proposed Addition

Let's extend the configuration format to additionally specify per-db instance addresses (either a single `host:port` address or a list of sentinel `host:port` addresses).

We could just allow configuring optional, additional mappings from DB index to address settings (either single or sentinels), while non-specified ones falling back to the base address settings.  This way, we could keep backward compatibility with existing setups.

**An example split:**

- Instance 1 for `STAT`, `LIVE`: persistence with AOF for HA
- Instance 2 for `STREAM`, `IMAGE`: no persistence with HA
- Instance 3 for `RLIM`: no persistence with HA

seedspirit:
resolves #3195 (BA-24)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/common/defs.py
+++ b/src/ai/backend/common/defs.py
@@ -11,7 +11,7 @@ REDIS_STREAM_DB: Final = 4
 REDIS_STREAM_LOCK: Final = 5
 
 
-class RedisTarget(StrEnum):
+class RedisRole(StrEnum):
     STAT = "stat"
     RLIM = "rlim"
     LIVE = "live"
--- a/src/ai/backend/common/defs.py
+++ b/src/ai/backend/common/defs.py
@@ -3,8 +3,8 @@ from enum import StrEnum
 from typing import Final
 
 # Redis database IDs depending on purposes
-REDIS_STAT_DB: Final = 0
-REDIS_RLIM_DB: Final = 1
+REDIS_STATISTICS_DB: Final = 0
+REDIS_RATE_LIMIT_DB: Final = 1
 REDIS_LIVE_DB: Final = 2
 REDIS_IMAGE_DB: Final = 3
 REDIS_STREAM_DB: Final = 4
@@ -12,8 +12,8 @@ REDIS_STREAM_LOCK: Final = 5
 
 
 class RedisRole(StrEnum):
-    STAT = "stat"
-    RLIM = "rlim"
+    STATISTICS = "statistics"
+    RATE_LIMIT = "rate_limit"
     LIVE = "live"
     IMAGE = "image"
     STREAM = "stream"
--- a/src/ai/backend/common/defs.py
+++ b/src/ai/backend/common/defs.py
@@ -1,4 +1,5 @@
 import re
+from enum import StrEnum
 from typing import Final
 
 # Redis database IDs depending on purposes
@@ -8,6 +9,15 @@ REDIS_LIVE_DB: Final = 2
 REDIS_IMAGE_DB: Final = 3
 REDIS_STREAM_DB: Final = 4
 REDIS_STREAM_LOCK: Final = 5
+
+
+class RedisTarget(StrEnum):
+    STAT = "stat"
+    RLIM = "rlim"
+    LIVE = "live"
+    IMAGE = "image"
+    STREAM = "stream"
+    STREAM_LOCK = "stream_lock"
 
 
 DEFAULT_FILE_IO_TIMEOUT: Final = 10
--- a/src/ai/backend/common/types.py
+++ b/src/ai/backend/common/types.py
@@ -1294,7 +1294,7 @@ def _stringify_number(v: Union[BinarySize, int, float, Decimal]) -> str:
 
 
 @dataclass
-class RedisConfig:
+class _RedisConfig:
     addr: Optional[HostPortPair] = None
     sentinel: Optional[Union[str, List[HostPortPair]]] = None
     service_name: Optional[str] = None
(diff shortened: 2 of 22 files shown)


src/ai/backend/common/types.py
```diff
@@ -1293,12 +1293,70 @@ def _stringify_number(v: Union[BinarySize, int, float, Decimal]) -> str:
...
+@dataclass
+class RedisConfig:
+    addr: Optional[HostPortPair] = None
+    sentinel: Optional[Union[str, List[HostPortPair]]] = None
+    service_name: Optional[str] = None
+    password: Optional[str] = None
+    redis_helper_config: Optional[RedisHelperConfig] = None
+
+    def get(self, key: str, default: Any = None) -> Any:
+        return getattr(self, key, default)
```
HyeockJinKim: Make the class private if it is not used publicly.

seedspirit: Fixed it! Thanks

src/ai/backend/common/types.py
```diff
@@ -1293,15 +1294,129 @@ def _stringify_number(v: Union[BinarySize, int, float, Decimal]) -> str:
...
-    redis_helper_config: RedisHelperConfig
+@dataclass
+class RedisConfig:
+    STANDARD_FIELDS = frozenset({
+        "addr",
+        "sentinel",
+        "service_name",
+        "password",
+        "redis_helper_config",
+    })
```
HyeockJinKim: Why is this setting necessary?

src/ai/backend/common/defs.py
```diff
@@ -10,6 +11,15 @@
 REDIS_STREAM_LOCK: Final = 5
 
 
+class RedisRole(StrEnum):
+    STAT = "stat"
+    RLIM = "rlim"
+    LIVE = "live"
+    IMAGE = "image"
+    STREAM = "stream"
+    STREAM_LOCK = "stream_lock"
```
HyeockJinKim: Avoid using abbreviations like `RLIM`.

src/ai/backend/common/types.py
seedspirit: As redis config allows extra fields, we need to add what is standard and what is not standard fields.
The only way I know of to add a field that is not predefined in the dataclass is to have it as a set or dict, as it is currently. Is there any other way?

```python
redis_config_iv = t.Dict({
    t.Key("addr", default=redis_default_config["addr"]): t.Null | tx.HostPortPair,
    ...
    t.Key("override_configs", default=None): t.Null
    | t.Mapping(
        t.String,
        t.Dict({
            t.Key("addr", default=redis_default_config["addr"]): t.Null | tx.HostPortPair,
            ...
        }).allow_extra("*"),
    ),
}).allow_extra("*")
```

src/ai/backend/common/defs.py
seedspirit: Changed it. Thank you

src/ai/backend/common/types.py
seedspirit: Removed it as `_extra` fields do nothing in this period

PR #3767, opened 2025-02-19
fix(BA-802): Set default domain name value of vfolders

### LINKED ISSUE #3765: Fail to update vfolders table schema to add `domain_name` field
Sometimes alembic migration upgrade fails to run `d7df3baf3779` script because of null violation

fregataa:
resolves #3765 (BA-802)

----
📚 Documentation preview 📚: https://sorna--3767.org.readthedocs.build/en/3767/

----
📚 Documentation preview 📚: https://sorna-ko--3767.org.readthedocs.build/ko/3767/


src/ai/backend/common/defs.py
```diff
@@ -44,3 +44,5 @@
 
 DEFAULT_VFOLDER_PERMISSION_MODE: Final[int] = 0o755
 VFOLDER_GROUP_PERMISSION_MODE: Final[int] = 0o775
+
+DEFAULT_DOMAIN_NAME: Final[str] = "default"
```
HyeockJinKim, line 48: I think it would be great if the default domain name was set to a less commonly used name that's not as user-friendly.

fregataa: we already use this as a default value. Change the default name to another should be a separate epic

HyeockJinKim (approved):

PR #3785, opened 2025-02-21
feat(BA-625): Register v2 volume handler to router

### LINKED ISSUE #3559: Add structure to easily register class handlers in storage-proxy
## Motivation  

- In the existing structure, when performing ⁠create_app, the handler was registered without upper-level information.
- However, now that we are creating a ⁠handler class and registering it with the router, we need information from the ⁠root_context.

## Required Features

- Structure to create and register services when running create_app

## Impact  

- Handlers can operate through ⁠service objects without accessing ⁠Root_context, allowing each ⁠Handler class to handle values separately.

## Testing Scenarios  

- Test if handler creation is possible.

HyeockJinKim:
resolves #3559 (BA-625)

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/storage/api/manager.py
```diff
@@ -1158,6 +1159,47 @@ async def _shutdown(app: web.Application) -> None:
             await task
 
 
+def init_v2_volume_app(service_ctx: ServiceContext) -> web.Application:
+    app = web.Application()
+    handler = VFolderHandler(service_ctx.volume_service)
+    app.router.add_route("GET", "/volumes", handler.get_volumes)
+    app.router.add_route("GET", "/volumes/{volume_id}", handler.get_volume)
+    app.router.add_route(
+        "GET", "/volumes/{volume_id}/quota/{scope_type}/{scope_uuid}", handler.get_quota_scope
```
MintCat98: It’s a minor issue, but I’m leaving this comment because I’m not entirely sure what you meant.

If you were referring to quota as the correct term for individually assigned limits on multiple volumes, then a revision may be necessary, as quotas would be the appropriate term in that case.

MintCat98 (commented):
Other than that, there are no major issues.
Thank you for the past two months.

src/ai/backend/storage/api/manager.py
HyeockJinKim: I've reflected it, please check it again when you have time. :)

MintCat98 (approved):
lgtm 👍

PR #3794, opened 2025-02-21
fix(BA-822): Wrong Python interpreters included in the scie builds (#3793)

lablup-octodog:
This is an auto-generated backport PR of #3793 to the 24.03 release.


achimnol:
This is not required to be backported because 24.09 and 24.03 has a custom pants plugin `scie` which specifies the exact yyyymmdd PBS version.

PR #3796, opened 2025-02-22
feat(BA-817): Better error log during server context initialization

### LINKED ISSUE #3788: Better error log when a server fails to bootstrap
Bootstrap failure error messages of our components have a bad readability

{width=1285 height=983}

fregataa:
resolves #3788 (BA-817)


src/ai/backend/manager/server.py
```diff
@@ -438,6 +448,11 @@ async def event_dispatcher_ctx(root_ctx: RootContext) -> AsyncIterator[None]:
         root_ctx.shared_config.data["redis"],
         db=REDIS_STREAM_DB,
     )
+    try:
+        await root_ctx.event_producer.ping()
+    except (Exception, asyncio.CancelledError) as e:
+        log.error("Failed to initiate event producer (error: {0})", repr(e))
+        raise
```
HyeockJinKim, line 471: Umm… Rather than inserting `log.error` in every place where an Exception might occur, it seems better to convey exceptions meaningfully and handle them in a common place.

HyeockJinKim (commented):
The direction of subsequent work should be towards applying it to the common structure, but this PR will apply it as a temporary measure for now.
Is it still in the draft state and not yet completed? @fregataa

src/ai/backend/manager/server.py
fregataa: Which place should the error handler be? I want the logger logs the exception raised in bootstrap steps

HyeockJinKim (changes requested):
I understand the issue, but I'm not sure if this is the way to implement it.

HyeockJinKim:
While applying the aiotools update PR, related improvement work was performed, so I will close this PR.

PR #3817, opened 2025-02-24
fix(BA-842): Use correct global lock ID

fregataa:
resolves #3811 (BA-842)

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version
- [x] Mention to the original issue

PR #3819, opened 2025-02-25
fix(BA-842): Use correct global lock ID (#3817)

lablup-octodog:
This is an auto-generated backport PR of #3817 to the 24.09 release.


HyeockJinKim:
Could you please check this? @fregataa

fregataa:
Close this PR because there is no duplicate global ID values in 24.09 version.

PR #3829, opened 2025-02-25
fix(BA-848): Wrong alembic scripts

### LINKED ISSUE #3828: Smell in alembic migration scripts
There are several bugs in the Alembic migration scripts. For example, some scripts import modules from `ai.backend.manager` or `ai.backend.common`, which can raise errors because the database schema might be different from the current code schema

fregataa:
resolves #3828 (BA-848)

### DIFF
--- a/src/ai/backend/manager/models/alembic/versions/85984c98b90f_update_endpoint_and_routing_table.py
+++ b/src/ai/backend/manager/models/alembic/versions/85984c98b90f_update_endpoint_and_routing_table.py
@@ -6,13 +6,14 @@ Create Date: 2023-05-11 12:40:09.197522
 
 """
 
+import enum
+
 import sqlalchemy as sa
 from alembic import op
 from sqlalchemy.dialects import postgresql
 
 from ai.backend.common.types import ClusterMode
 from ai.backend.manager.models.base import GUID, URLColumn
-from ai.backend.manager.models.routing import RouteStatus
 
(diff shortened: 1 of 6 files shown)


src/ai/backend/manager/models/alembic/versions/85984c98b90f_update_endpoint_and_routing_table.py
```diff
@@ -6,21 +6,31 @@
...
 down_revision = "857bdec5abda"
 branch_labels = None
 depends_on = None
 
-routestatus_choices = [v.value for v in RouteStatus]
+
+class LegacyRouteStatus(enum.Enum):
+    HEALTHY = "healthy"
+    UNHEALTHY = "unhealthy"
+    TERMINATING = "unhealthy"
```
HyeockJinKim, line 28: Is `TERMINATING` also `unhealthy`?

fregataa: In #1278, the enum type was defined with duplicate values by mistake.
And in #2191, the enum type was corrected by adding a new "terminating" value.

The migration script `38741c` made in #2191 raises an error saying 'terminating value already exists' because `RouteStatus` is imported directly in `85984`, which leads to create the `routestatus` enum with correct values.
So, I leave the typo by purpose.

HyeockJinKim (approved):

PR #3871, opened 2025-03-02
fix(BA-887): Rescan all tags in `HarborRegistryV2`

### LINKED ISSUE #3872: Rescan only the latest tag in `HarborRegistryV2`
## Summary  

- When rescanning images in the `HarborRegistryV2` type registry, the rescan operation occurs only for a single tag that is considered as "latest."
- Even the ignored tags are not being logged. This is not the behavior the administrator expects.

## Steps to Reproduce  

1. Run a Harbor registry for testing.
1. Push the test image to the registry with the tags `0.1` and `0.2`.
1. Run the image rescan command for the testing registry.

## Expected Behavior  

- The rescan should work for all tags, or at the least, the ignored tags should be logged.

## Actual Behavior  

- Only version `0.2` is recognized, while version `0.1` is neither rescanned nor logged.

## Logs/Errors  

- None

jopemachine:
Resolves #3872 ([BA-887](https://lablup.atlassian.net/browse/BA-887)).

> [!NOTE]
> This bug occurs only in HarborRegistryV2 (it does not occur in V1 or other registries).

---

[BA-887]: https://lablup.atlassian.net/browse/BA-887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ai/backend/manager/container_registry/harbor.py
+++ b/src/ai/backend/manager/container_registry/harbor.py
@@ -316,8 +316,6 @@ class HarborRegistry_v2(BaseContainerRegistry):
             resp.raise_for_status()
             resp_json = await resp.json()
             async with aiotools.TaskGroup() as tg:
-                tag = resp_json["tags"][0]["name"]
-
                 match resp_json["manifest_media_type"]:
                     case self.MEDIA_TYPE_OCI_INDEX:
                         await self._process_oci_index(tg, sess, rqst_args, image, tag, resp_json)
--- a/src/ai/backend/manager/container_registry/harbor.py
+++ b/src/ai/backend/manager/container_registry/harbor.py
@@ -260,24 +260,26 @@ class HarborRegistry_v2(BaseContainerRegistry):
                             if not image_info["tags"] or len(image_info["tags"]) == 0:
                                 skip_reason = "no tag"
                                 continue
-                            tag = image_info["tags"][0]["name"]
-                            match image_info["manifest_media_type"]:
-                                case self.MEDIA_TYPE_OCI_INDEX:
-                                    await self._process_oci_index(
-                                        tg, sess, rqst_args, image, tag, image_info
-                                    )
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/container_registry/harbor.py
```diff
@@ -260,24 +260,26 @@ async def _scan_image(
...
-                                    )
-                                case self.MEDIA_TYPE_DOCKER_MANIFEST:
-                                    await self._process_docker_v2_image(
-                                        tg, sess, rqst_args, image, tag, image_info
-                                    )
-                                case _ as media_type:
-                                    raise RuntimeError(
-                                        f"Unsupported artifact media-type: {media_type}"
-                                    )
+                            tags = [item["name"] for item in image_info["tags"]]
```
HyeockJinKim, line 263: The name field will always be present, right?

jopemachine: I searched for relevant information in the Harbor API Swagger and API specification file but couldn't find any evidence that Tag.name is non-optional.
In the API specification below, we can find the definition of Tags, but the `required` array is not defined.
- https://github.com/goharbor/harbor/blob/main/api/v2.0/swagger.yaml

It’s just that, based on experience, there has never been a case where a Tag object didn’t have a name.

src/ai/backend/manager/container_registry/harbor.py
```diff
@@ -314,8 +316,6 @@ async def _scan_tag(
             resp.raise_for_status()
             resp_json = await resp.json()
             async with aiotools.TaskGroup() as tg:
-                tag = resp_json["tags"][0]["name"]
```
HyeockJinKim, line 317: 👍

HyeockJinKim (approved):

PR #3891, opened 2025-03-05
feat(BA-15): Add GQL Exception middleware

### LINKED ISSUE #3213: Implement Exception Handling by Graphene Middleware
### Main idea

In the current code, each mutation requires manually injecting `ok` and `msg` in the form of `cls(ok="...", msg="...")` every time.This approach is prone to mistakes, and since it involves code related to the API interface, mistakes can be difficult to roll back.

We can refactor it to simply throw an error when an error occurs, allowing the middleware to handle it instead.It seems this middleware could be added at the following location.

https://github.com/lablup/backend.ai/blob/eadddf67c138523ced275fa1bbd647424f669671/src/ai/backend/manager/api/admin.py#L97-L101

HyeockJinKim:
resolves #3213 (BA-15)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/models/gql.py
+++ b/src/ai/backend/manager/models/gql.py
@@ -1,6 +1,7 @@
 # ruff: noqa: E402
 from __future__ import annotations
 
+import time
 import uuid
 from collections.abc import Mapping, Sequence
 from typing import TYPE_CHECKING, Any, Optional
@@ -8,9 +9,10 @@ from typing import TYPE_CHECKING, Any, Optional
 import attrs
 import graphene
 from graphene.types.inputobjecttype import set_input_object_type_default_value
-from graphql import OperationType, Undefined
+from graphql import GraphQLError, OperationType, Undefined
 from graphql.type import GraphQLField
 
+from ai.backend.common.metrics.metric import GraphQLMetricObserver
 from ai.backend.manager.plugin.network import NetworkPluginContext
 from ai.backend.manager.service.base import ServicesContext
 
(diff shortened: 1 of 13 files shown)


HyeockJinKim:
image

src/ai/backend/manager/models/gql.py
```diff
@@ -2860,3 +2863,48 @@ def resolve(self, next, root, info: graphene.ResolveInfo, **args) -> Any:
...
+                success=True,
+                duration=time.perf_counter() - start,
+            )
+        except BaseException as e:
+            graph_ctx.metric_observer.observe_request(
+                operation_type=operation_type,
+                field_name=field_name,
+                parent_type=parent_type,
+                operation_name=operation_name,
+                success=True,
```
fregataa: Is the value of `success` supposed to be `True` here?

src/ai/backend/manager/models/gql.py
```diff
@@ -2860,3 +2863,48 @@ def resolve(self, next, root, info: graphene.ResolveInfo, **args) -> Any:
...
+        return res
+
+
+class GQLMetricMiddleware:
+    def resolve(self, next, root, info: graphene.ResolveInfo, **args) -> Any:
+        graph_ctx: GraphQueryContext = info.context
+        operation_type = info.operation.operation
+        field_name = info.field_name
+        parent_type = info.parent_type.name
+        operation_name = info.operation.name.value if info.operation.name else "anonymous"
```
fregataa: if info.operation.name is not None else "anonymous"

fregataa (approved):

PR #3895, opened 2025-03-05
fix(BA-910): Resolve vfolder mount correctly when ID is given

### LINKED ISSUE #3893: Cannot create model service
Cannot create model service. The manager does not mount model folders

fregataa:
resolves #3893 (BA-910)

### DIFF
--- a/src/ai/backend/manager/models/vfolder.py
+++ b/src/ai/backend/manager/models/vfolder.py
@@ -888,22 +888,23 @@ async def prepare_vfolder_mounts(
     requested_mounts: list[str] = [
         name for name in requested_mount_references if isinstance(name, str)
     ]
-    requested_mount_map: dict[str, str] = {
+    requested_mount_name_map: dict[str, str] = {
         name: path for name, path in requested_mount_reference_map.items() if isinstance(name, str)
     }
-    requested_mount_options: dict[str, dict[str, Any]] = {
+    requested_mount_map: dict[uuid.UUID, str] = {
+        name: path
+        for name, path in requested_mount_reference_map.items()
+        if isinstance(name, uuid.UUID)
+    }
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/models/vfolder.py
```diff
@@ -888,22 +888,23 @@ async def prepare_vfolder_mounts(
     requested_mounts: list[str] = [
         name for name in requested_mount_references if isinstance(name, str)
     ]
-    requested_mount_map: dict[str, str] = {
+    requested_mount_name_map: dict[str, str] = {
         name: path for name, path in requested_mount_reference_map.items() if isinstance(name, str)
```
HyeockJinKim, line 892: The only difference between `requested_mount_references` and `requested_mount_reference_map` is that one is a list and the other is a dict, right?

src/ai/backend/manager/models/vfolder.py
```diff
@@ -888,22 +888,23 @@ async def prepare_vfolder_mounts(
...
     }
-    requested_mount_options: dict[str, dict[str, Any]] = {
+    requested_mount_map: dict[uuid.UUID, str] = {
+        name: path
+        for name, path in requested_mount_reference_map.items()
+        if isinstance(name, uuid.UUID)
```
HyeockJinKim: Why use the `name` name when it's a UUID?

src/ai/backend/manager/models/vfolder.py
fregataa: The value of `requested_mount_reference_map` dictionary indicates the destination directory of vfolder mounts

HyeockJinKim: It seems to assume that the two values are the same. Maybe we should refactor that next time. 😢

HyeockJinKim (approved):

PR #3906, opened 2025-03-06
fix(BA-840): Apply Image RBAC to `create_session()`

### LINKED ISSUE #3809: Resolve Image RBAC Non-Application Logics
## Summary 

- In the "Per project customized image" PR, the Image RBAC interface and RBAC-based ImageNode and ImageNodes GQL resolver have been implemented.
- However, the RBAC interface has not yet been applied to the entire codebase. This issue introduces several CLI command cases where Image RBAC has not been applied.

### Create a session using other user’s custom image

1. User1 commits a specific image to a specific non-global registry.
1. Assume that User2 belongs to the same group as User1. 
1. After applying User2's API token, they execute the `session commit` command to use the custom image committed by User1.
1. Then, the following error can be observed.

```
✘ BackendAPIError: 404 Not Found
  No such environment image.
```

This is because the RBAC interface has not yet been applied to the logic that determines the image visibility in the `session commit` command.

Specifically, in the above case, the `session commit` command fails to find the custom image due to the following logic. To implement project image visibility, we need to replace this logic with Image RBAC as well.

[https://github.com/lablup/backend.ai/blob/4517b0c189c10ad8ceb2c7c3294a0e0f2f5683ea/src/ai/backend/manager/registry.py#L508-L511](https://github.com/lablup/backend.ai/blob/4517b0c189c10ad8ceb2c7c3294a0e0f2f5683ea/src/ai/backend/manager/registry.py#L508-L511)

The above example only introduces the `session commit` command’s case, but other commands (such as `rescan`) also need Image RBAC applied.

## Expected Behavior  

- User2 should be able to use the image committed by User1.

## Actual Behavior  

- User2 cannot use the image committed by User1.

## Logs/Errors  

```
✘ BackendAPIError: 404 Not Found
  No such environment image.
```

JIRA Issue: BA-840

jopemachine:
Resolves #3809 (BA-840).

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/registry.py
+++ b/src/ai/backend/manager/registry.py
@@ -123,6 +123,9 @@ from ai.backend.common.types import (
 )
 from ai.backend.common.utils import str_to_timedelta
 from ai.backend.logging import BraceStyleAdapter
+from ai.backend.manager.models.rbac import SystemScope
+from ai.backend.manager.models.rbac.context import ClientContext
+from ai.backend.manager.models.rbac.permission_defs import ImagePermission
 from ai.backend.manager.plugin.network import NetworkPluginContext
 from ai.backend.manager.utils import query_userinfo
 
@@ -182,7 +185,7 @@ from .models import (
     verify_vfolder_name,
 )
 from .models.container_registry import ContainerRegistryRow
-from .models.image import ImageIdentifier, bulk_get_image_configs
+from .models.image import ImageIdentifier, bulk_get_image_configs, get_permission_ctx
 from .models.session import (
     COMPUTE_CONCURRENCY_USED_KEY_PREFIX,
     SESSION_KERNEL_STATUS_MAPPING,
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/registry.py
```diff
@@ -510,10 +513,26 @@ async def create_session(
...
+                client_ctx = ClientContext(
+                    db=self.db,
+                    domain_name=user_scope.domain_name,
+                    user_id=user_scope.user_uuid,
+                    user_role=UserRole(user_scope.user_role),
+                )
+
+                perm_ctx = await get_permission_ctx(
+                    db_conn, client_ctx, SystemScope(), ImagePermission.CREATE_CONTAINER
+                )
```
jopemachine, line 527: Note: To share images between teams, use SystemScope. To ensure that only the user who committed the image (the owner) can use it according to the existing logic, use UserScope.

jopemachine: IMHO, It would be nice if users could set the sharing scope to either "private" or "team-wide" when committing an image.
For now, I will use `UserScope` to maintain the existing logic.

PR #3923, opened 2025-03-08
fix(BA-932): Unload removed docker images from Redis

### LINKED ISSUE #3924: Unload purged docker images
## Summary  

- When images are removed using the `PurgeImages` mutation, its cache is not cleared from Redis, so it is still displayed as "Installation Complete."

## Steps to Reproduce  

1. Call `PurgeImages` mutation
1. Navigate to the WebUI execution environment page.
1. Check that the deleted images are still displayed as "Installation Complete."

## Expected Behavior  

- Deleted images should no longer be displayed as "Installation Complete."

## Actual Behavior  

- Deleted images are still displayed as "Installation Complete."

## Logs/Errors  

- None

jopemachine:
Resolves #3924 (BA-927, BA-932)

Previously, there was no feature to remove images, so the absence of the Unload logic does not seem like a bug. 
So I think we don't need to backport this PR. So I will set the milestone to 25Q1.

### DIFF
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -26,6 +26,7 @@ from collections.abc import (
     MutableSequence,
     Sequence,
 )
+from dataclasses import dataclass
 from decimal import Decimal
 from io import SEEK_END, BytesIO
 from pathlib import Path
@@ -79,6 +80,7 @@ from ai.backend.common.events import (
     AbstractEvent,
     AgentErrorEvent,
     AgentHeartbeatEvent,
+    AgentPurgeImagesEvent,
     AgentStartedEvent,
     AgentTerminatedEvent,
     DoAgentResourceCheckEvent,
@@ -223,6 +225,12 @@ def update_additional_gids(environ: MutableMapping[str, str], gids: Iterable[int
     else:
         additional_gids = set(gids)
     environ["ADDITIONAL_GIDS"] = ",".join(map(str, additional_gids))
+
+
+@dataclass
+class ScanImagesResult:
+    scanned_images: Mapping[str, str]
+    removed_images: Mapping[str, str]
 
 
 class AbstractKernelCreationContext(aobject, Generic[KernelObjectType]):
@@ -772,7 +780,7 @@ class AbstractAgent(
         self.affinity_map = AffinityMap.build(all_devices)
 
         if not self._skip_initial_scan:
-            self.images = await self.scan_images()
+            self.images = (await self.scan_images()).scanned_images
             self.timer_tasks.append(aiotools.create_timer(self._scan_images_wrapper, 20.0))
             await self.scan_running_kernels()
 
@@ -1679,7 +1687,7 @@ class AbstractAgent(
             )
 
     @abstractmethod
-    async def scan_images(self) -> Mapping[str, str]:
+    async def scan_images(self) -> ScanImagesResult:
         """
         Scan the available kernel images/templates and update ``self.images``.
         This is called periodically to keep the image list up-to-date and allow
@@ -1687,7 +1695,11 @@ class AbstractAgent(
         """
 
     async def _scan_images_wrapper(self, interval: float) -> None:
-        self.images = await self.scan_images()
+        result = await self.scan_images()
+        self.images = result.scanned_images
+        await self.produce_event(
+            AgentPurgeImagesEvent(image_canonicals=list(result.removed_images.keys()))
+        )
 
     @abstractmethod
     async def push_image(
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -227,10 +227,16 @@ def update_additional_gids(environ: MutableMapping[str, str], gids: Iterable[int
     environ["ADDITIONAL_GIDS"] = ",".join(map(str, additional_gids))
 
 
+ImageCanonical = str
+ImageDigest = str
+
+ScannedImage = Mapping[ImageCanonical, ImageDigest]
+
+
 @dataclass
 class ScanImagesResult:
-    scanned_images: Mapping[str, str]
-    removed_images: Mapping[str, str]
+    scanned_images: ScannedImage
+    removed_images: ScannedImage
 
 
 class AbstractKernelCreationContext(aobject, Generic[KernelObjectType]):
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -1703,9 +1703,10 @@ class AbstractAgent(
     async def _scan_images_wrapper(self, interval: float) -> None:
         result = await self.scan_images()
         self.images = result.scanned_images
-        await self.produce_event(
-            AgentImagesRemoveEvent(image_canonicals=list(result.removed_images.keys()))
-        )
+        if result.removed_images:
+            await self.produce_event(
+                AgentImagesRemoveEvent(image_canonicals=list(result.removed_images.keys()))
+            )
 
     @abstractmethod
     async def push_image(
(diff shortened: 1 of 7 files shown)


src/ai/backend/manager/registry.py
```diff
@@ -3149,19 +3151,21 @@ async def _update() -> None:
...
+            prev_image_canonicals = self.agent_installed_images.get(agent_id, set())
 
             async def _pipe_builder(r: Redis):
                 pipe = r.pipeline()
-                for image, _ in loaded_images:
-                    try:
-                        await pipe.sadd(ImageRef.parse_image_str(image, "*").canonical, agent_id)
-                    except ValueError:
-                        # Skip opaque (non-Backend.AI) image.
-                        continue
```
jopemachine, line 3171: I intentionally omitted the logic to skip image caching when a ValueError occurs while parsing the image string since I think excluding the loaded images sent from the agent from only redis caching is meaningless.

jopemachine:
I'm concerned that WebUI will remain displayed as "Installation Complete" until the next `scan_image` is called. 
However, if `purge_images` directly deletes the Redis cache, there could be a timing issue where the deleted image cache gets restored, so I will not remove the Redis cache in `purge_images`.

src/ai/backend/agent/docker/agent.py
```diff
@@ -1240,7 +1240,6 @@ class DockerAgent(AbstractAgent[DockerKernel, DockerKernelCreationContext]):
     monitor_docker_task: asyncio.Task
     agent_sockpath: Path
     agent_sock_task: asyncio.Task
-    scan_images_timer: asyncio.Task
```
jopemachine, line 1268: The timer is created in the line below, but this variable is not used.

https://github.com/lablup/backend.ai/blob/1f77864bd706111fc55d6f3733fdf58439b0c446/src/ai/backend/agent/agent.py#L755

src/ai/backend/manager/registry.py
```diff
@@ -3149,19 +3151,21 @@ async def _update() -> None:
                 )
 
             # Update the mapping of kernel images to agents.
-            loaded_images = msgpack.unpackb(zlib.decompress(agent_info["images"]))
+            images = msgpack.unpackb(zlib.decompress(agent_info["images"]))
+            image_canonicals = set(img_info[0] for img_info in images)
```
HyeockJinKim, line 3165: I assume the 0th value means canonical, but that's not going to change, right?

src/ai/backend/manager/registry.py
```diff
@@ -237,6 +237,7 @@ class AgentRegistry:
     pending_waits: set[asyncio.Task[None]]
     database_ptask_group: aiotools.PersistentTaskGroup
     webhook_ptask_group: aiotools.PersistentTaskGroup
+    agent_installed_images: dict[AgentId, set[str]]
```
HyeockJinKim: we are caching image info to redis, should we cache locally again?

jopemachine: Yes, this is necessary to detect images that need to be removed.

src/ai/backend/manager/registry.py
jopemachine: Yes, FYI,
Tuple's 0th value: image canonical.
Tuple's 1st value: image digest.

src/ai/backend/agent/agent.py
```diff
@@ -202,6 +204,19 @@ def update_additional_gids(environ: MutableMapping[str, str], gids: Iterable[int
...
+    def __setitem__(self, image_canonical: str, image_digest: str) -> None:
+        if not isinstance(image_canonical, str) or not isinstance(image_digest, str):
+            raise TypeError("image_canonical, image_digest must be str")
+        super().__setitem__(image_canonical, image_digest)
+
+
+@dataclass
+class ScanImagesResult:
+    scanned_images: ImageCanonicalDigestDict
+    removed_images: ImageCanonicalDigestDict
```
HyeockJinKim: Is there a reason you made it a `UserDict`? I think you could have just used a dict for this part.
My current thought is to create a dataclass with Image canonical and image digest as fields and save it as a set.

HyeockJinKim: @dataclass
class ImageMeta
  canonical: str
  digest: str
`dict[str, ImageMeta]`

jopemachine:
@HyeockJinKim 
Two personal concerns: 

1. Could event reception be missed due to a network (RPC call) failure?
In the current structure, removed images are not detected in the heartbeat, so if the AgentImagesRemoveEvent processing is missed, the deleted images will continue to be displayed as successfully installed.

2. How about adding `AgentImagesAddEvent` just like `AgentImagesRemoveEvent`?
In the current structure, Backend.AI users have to wait up to an additional 20 seconds (scan_image timer) on top of the image installation time. I think if `AgentImagesAddEvent` handling is added, users will feel that the image is installed much faster.

src/ai/backend/agent/agent.py
```diff
@@ -1632,15 +1646,19 @@ def _rm(path: Path) -> None:
...
         manual image addition and deletions by admins.
         """
 
     async def _scan_images_wrapper(self, interval: float) -> None:
-        self.images = await self.scan_images()
+        result = await self.scan_images()
+        self.images = result.scanned_images
+        await self.produce_event(
+            AgentImagesRemoveEvent(image_canonicals=list(result.removed_images.keys()))
+        )
```
HyeockJinKim: Does it always raise an event?

src/ai/backend/manager/registry.py
```diff
@@ -3151,16 +3153,13 @@ async def _update() -> None:
...
             async def _pipe_builder(r: Redis):
                 pipe = r.pipeline()
-                for image, _ in loaded_images:
-                    try:
-                        await pipe.sadd(ImageRef.parse_image_str(image, "*").canonical, agent_id)
-                    except ValueError:
-                        # Skip opaque (non-Backend.AI) image.
-                        continue
+                for image in image_canonicals:
+                    await pipe.sadd(image, agent_id)
```
HyeockJinKim: The variable in the for loop should also be `image_canonical`.

src/ai/backend/manager/registry.py
```diff
@@ -3170,6 +3169,17 @@ async def _pipe_builder(r: Redis):
             (agent_id, sgroup, available_slots),
         )
 
+    async def handle_agent_images_remove(
+        self, agent_id: AgentId, image_canonicals: list[str]
+    ) -> None:
+        async def _pipe_builder(r: Redis):
+            pipe = r.pipeline()
+            for image in image_canonicals:
```
HyeockJinKim: this too.

jopemachine:
> @HyeockJinKim Two personal concerns:
> 
> 1. Could event reception be missed due to a network (RPC call) failure (or other reasons...)?
>    In the current structure, removed images are not detected in the heartbeat, so if the AgentImagesRemoveEvent processing is missed, the deleted images will continue to be displayed as successfully installed.
> 2. How about adding `AgentImagesAddEvent` just like `AgentImagesRemoveEvent`?
>    In the current structure, Backend.AI users have to wait up to an additional 20 seconds (scan_image timer) on top of the image installation time. I think if `AgentImagesAddEvent` handling is added, users will feel that the image is installed much faster.

Let's address this in a separate follow-up PR.

src/ai/backend/agent/agent.py
```diff
@@ -751,7 +765,7 @@ async def _pipeline(r: Redis):
         self.affinity_map = AffinityMap.build(all_devices)
 
         if not self._skip_initial_scan:
-            self.images = await self.scan_images()
+            self.images = (await self.scan_images()).scanned_images
```
HyeockJinKim: Whenever possible, write only one action per line.

PR #3939, opened 2025-03-11
fix(BA-954): Customized image visibility issue

### LINKED ISSUE #3938: Customized image visibility issue
## Summary  

- In the current Image RBAC implementation, there is an issue where customized images are included in SystemScope and DomainScope.
  A custom image should only be accessible to the user who committed it and should only be used to create sessions by that user.
  So, custom images should be visible only to the user who owns them.

## Steps to Reproduce  

1. Create a session with an arbitrary image and then commit it.
1. Log in as a user other than the one who created the session.
1. Call the Image Nodes GQL resolver with System Scope.

## Expected Behavior  

- Custom images that the user did not create should not be included in the query results.

## Actual Behavior  

- Custom images that the user did not create are included in the query results.

## Logs/Errors  

- None

jopemachine:
Resolves #3938 (BA-954)

### DIFF
--- a/src/ai/backend/manager/models/image.py
+++ b/src/ai/backend/manager/models/image.py
@@ -1025,6 +1025,19 @@ class ImagePermissionContextBuilder(
             return False
         return True
 
+    def _is_image_accessible_for_user_in_user_scope(
+        self,
+        image: ImageRow,
+        allowed_registries: set,
+        user_id: UUID,
+        has_admin_privilege: bool,
+    ) -> bool:
+        if has_admin_privilege:
+            return image.registry in allowed_registries
+        return image.is_customized_by(user_id) and self._is_image_accessible_for_user(
+            image, allowed_registries, user_id
+        )
+
     async def _in_user_scope(
         self,
         ctx: ClientContext,
@@ -1045,16 +1058,10 @@ class ImagePermissionContextBuilder(
 
         for row in await self.db_session.scalars(img_query_stmt):
             image_row = cast(ImageRow, row)
-            if has_admin_privilege:
-                if image_row.registry not in allowed_registries:
-                    continue
-            else:
-                if not image_row.is_customized_by(
-                    scope.user_id
-                ) or not self._is_image_accessible_for_user(
-                    image_row, allowed_registries, scope.user_id
-                ):
-                    continue
+            if not self._is_image_accessible_for_user_in_user_scope(
+                image_row, allowed_registries, ctx.user_id, has_admin_privilege=has_admin_privilege
+            ):
+                continue
 
             image_id_permission_map[image_row.id] = permissions
 
--- a/src/ai/backend/manager/models/image.py
+++ b/src/ai/backend/manager/models/image.py
@@ -939,7 +939,14 @@ class ImagePermissionContextBuilder(
             ctx, user_accessible_project_scopes
         )
         perm_ctx.merge(non_global_project_scopes_perm_ctx)
-        user_scope_perm_ctx = await self._in_user_scope(ctx, UserScope(ctx.user_id))
+
+        if ctx.user_role in [UserRole.ADMIN, UserRole.SUPERADMIN]:
+            user_scope_perm_ctx = await self._in_user_scope(
+                ctx, UserScope(ctx.user_id), has_admin_privilege=True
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/models/image.py
```diff
@@ -1011,8 +1011,10 @@ def _is_image_accessible_for_user(
         if image.registry not in allowed_registries:
             return False
         labels = cast(dict[str, str], image.labels)
-        if labels.get("ai.backend.customized-image.owner") != f"user:{user_id}":
-            return False
+        owner_label = labels.get("ai.backend.customized-image.owner")
+        if owner_label is not None:
+            if owner_label != f"user:{user_id}":
+                return False
```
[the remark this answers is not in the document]
HyeockJinKim: This seems like it could be handled in a single if statement with `and`.

src/ai/backend/manager/models/image.py
```diff
@@ -1034,7 +1036,11 @@ async def _in_user_scope(
 
         for row in await self.db_session.scalars(_img_query_stmt):
             image_row = cast(ImageRow, row)
-            if not self._is_image_accessible_for_user(image_row, allowed_registries, scope.user_id):
+            if image_row.labels.get(
+                "ai.backend.customized-image.owner"
+            ) is None or not self._is_image_accessible_for_user(
```
HyeockJinKim: Wouldn't `image_row.is_customized_image()` be more readable than `image_row.labels.get("ai.backend.customized-image.owner")`?

src/ai/backend/manager/models/image.py
```diff
@@ -1011,31 +1020,42 @@ def _is_image_accessible_for_user(
...
+            if has_admin_privilege:
+                if image_row.registry not in allowed_registries:
+                    continue
+            else:
+                if not image_row.is_customized_by(
+                    scope.user_id
+                ) or not self._is_image_accessible_for_user(
+                    image_row, allowed_registries, scope.user_id
+                ):
+                    continue
```
jopemachine: This logic is verbose, complex, and contains many redundant parts, so it seems to need refactoring.
Rather than reorganizing the if conditions, it would be better to rewrite the function.

src/ai/backend/manager/models/image.py
```diff
@@ -1011,31 +1020,42 @@ def _is_image_accessible_for_user(
         if image.registry not in allowed_registries:
             return False
         labels = cast(dict[str, str], image.labels)
-        if labels.get("ai.backend.customized-image.owner") != f"user:{user_id}":
+        owner_label = labels.get("ai.backend.customized-image.owner")
+        if owner_label is not None and owner_label != f"user:{user_id}":
```
HyeockJinKim: if not image.is_cusomized_by(): ?

src/ai/backend/manager/models/image.py
```diff
@@ -1010,32 +1019,50 @@ def _is_image_accessible_for_user(
     ) -> bool:
         if image.registry not in allowed_registries:
             return False
-        labels = cast(dict[str, str], image.labels)
-        if labels.get("ai.backend.customized-image.owner") != f"user:{user_id}":
+        if image.labels.get(
+            "ai.backend.customized-image.owner"
+        ) is not None and not image.is_customized_by(user_id):
```
HyeockJinKim: This feels unnatural, is there another way?

src/ai/backend/manager/models/image.py
```diff
@@ -892,6 +893,33 @@ async def calculate_final_permission(self, rbac_obj: ImageRow) -> frozenset[Imag
...
+        return True
+
+    def is_accessible_customized_image(self, image: ImageRow) -> bool:
+        """
+        Check if the image is a customized image accessible by the user.
+        """
+        if image.labels.get("ai.backend.customized-image.owner") is None:
+            return False
+
+        return image.is_customized_by(self.user_id) and self.is_accessible_image(image)
```
HyeockJinKim: It seems like all you just need to do is create the `customized()`, `is_owner(user_id)` methods on the ImageRow and `is_accessible(image_row)` on the ImageAccessCriteria.

HyeockJinKim (approved):

PR #3951, opened 2025-03-12
feat(BA-997): Add `image purge` CLI command for hard deleting ImageRow

### LINKED ISSUE #3979: Add `image purge` CLI command for hard deleting ImageRow from DB
## Motivation  

- While the commands `mgr image forget` and `mgr image purge` exist,
  only the `image forget` command is available without a corresponding `image purge` command, which is confusing.

## Required Features

- Add `image purge` CLI command.

## Impact  

- Provide a more user-friendly CLI interface.

## Testing Scenarios  

- Manual test

jopemachine:
Resolves #3979 (BA-997).

----
📚 Documentation preview 📚: https://sorna--3951.org.readthedocs.build/en/3951/

----
📚 Documentation preview 📚: https://sorna-ko--3951.org.readthedocs.build/ko/3951/

### DIFF
--- a/docs/manager/graphql-reference/schema.graphql
+++ b/docs/manager/graphql-reference/schema.graphql
@@ -2882,7 +2882,7 @@ input PurgeImageOptions {
 }
 
 """
-Deprecated since 25.10.0. Use `purge_image_by_id`, and `remove_from_registry` option instead.
+Deprecated since 25.10.0. Use `purge_image_by_id` with `remove_from_registry` option instead.
 """
 type UntagImageFromRegistry {
   ok: Boolean
--- a/docs/manager/graphql-reference/schema.graphql
+++ b/docs/manager/graphql-reference/schema.graphql
@@ -2270,7 +2270,7 @@ type Mutations {
     image_id: String!
 
     """Added in 25.5.0."""
-    options: PurgeImageByIdOptions = {remove_from_registry: false}
+    options: PurgeImageOptions = {remove_from_registry: false}
   ): PurgeImageById
 
   """Added in 24.03.1"""
@@ -2874,7 +2874,7 @@ type PurgeImageById {
 }
 
 """Added in 25.5.0."""
-input PurgeImageByIdOptions {
+input PurgeImageOptions {
   """Only available in the HarborV2 registry."""
   remove_from_registry: Boolean = false
 }
--- a/docs/manager/graphql-reference/schema.graphql
+++ b/docs/manager/graphql-reference/schema.graphql
@@ -2269,7 +2269,7 @@ type Mutations {
   purge_image_by_id(
     image_id: String!
 
-    """Added in 25.6.0."""
+    """Added in 25.10.0."""
     options: PurgeImageOptions = {remove_from_registry: false}
   ): PurgeImageById
 
@@ -2880,7 +2880,7 @@ input PurgeImageOptions {
 }
 
 """
-Deprecated since 25.6.0. Use `purge_image_by_id`, and `remove_from_registry` option instead.
+Deprecated since 25.10.0. Use `purge_image_by_id`, and `remove_from_registry` option instead.
 """
 type UntagImageFromRegistry {
   ok: Boolean
--- a/docs/manager/graphql-reference/schema.graphql
+++ b/docs/manager/graphql-reference/schema.graphql
@@ -2873,7 +2873,7 @@ type PurgeImageById {
   image: ImageNode
 }
 
-"""Added in 25.6.0."""
+"""Added in 25.10.0."""
 input PurgeImageOptions {
   """Only available in the HarborV2 registry."""
   remove_from_registry: Boolean = false
--- a/docs/manager/graphql-reference/schema.graphql
+++ b/docs/manager/graphql-reference/schema.graphql
@@ -2269,7 +2269,7 @@ type Mutations {
   purge_image_by_id(
     image_id: String!
 
-    """Added in 25.5.0."""
+    """Added in 25.6.0."""
     options: PurgeImageOptions = {remove_from_registry: false}
   ): PurgeImageById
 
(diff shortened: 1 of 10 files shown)


src/ai/backend/client/func/image.py
```diff
@@ -157,6 +157,22 @@ async def forget_image_by_id(cls, image_id: str):
...
         return data["forget_image_by_id"]
 
+    @api_function
+    @classmethod
+    async def purge_image_by_id(cls, image_id: str):
+        q = _d("""
+            mutation($image_id: String!) {
+                purge_image_by_id(image_id: $image_id) {
+                    __typename
+                }
```
jopemachine: It would be nice to have an alternative way to indicate that no value should be returned.

jopemachine: Does this function need an `ImageNode` as its return type?
If so, we should first add a new type corresponding to the Client. @HyeockJinKim

src/ai/backend/client/cli/image.py
```diff
@@ -92,3 +92,40 @@ def forget(reference_or_id, arch):
...
+    REFERENCE_OR_ID: Canonical string of image (<registry>/<project>/<name>:<tag>)"""
+    with Session() as session:
+        try:
+            image_id = get_image_id(session, reference_or_id, architecture=arch)
+        except BackendAPIError:
+            print_fail("Could not find image.")
+            if not arch:
+                print_warn(
+                    "`arch` parameter not passed. If you are trying to resolve image via its canonical string you have to specify which architecture are you trying to manipulate with."
+                )
```
HyeockJinKim, line 113: I feel like this would be an inappropriate place to warn when there is no arch, is there a reason to warn at this point?

src/ai/backend/client/cli/image.py
```diff
@@ -92,3 +92,40 @@ def forget(reference_or_id, arch):
...
+            sys.exit(ExitCode.FAILURE)
+        try:
+            if remove_from_registry:
+                result = session.Image.untag_image_from_registry(image_id)
+                if not result["ok"]:
+                    print_fail("Image untag has failed: {0}".format(result["msg"]))
+                    sys.exit(ExitCode.FAILURE)
+            session.Image.purge_image_by_id(image_id)
```
HyeockJinKim: I guess it should be managed as a transaction, but is it okay if only one of the actions (remove_from_registry, purge_image_by_id) executes and the other fails?

jopemachine: If `untag_image_from_registry` succeeds but `purge_image_by_id` fails, an inappropriate ImageRow will remain in the `images` table.

Unless the `purge_image` mutation is explicitly called for the corresponding `ImageRow`, it will continue to exist.

Ideally, a soft deletion process should be added during the image rescan to handle images that have already been removed from the registry.

See BA-969 for this

src/ai/backend/client/cli/image.py
jopemachine: As stated in the warning message, canonical alone is not sufficient to identify an `ImageRow`, so a warning should be issued if the arch value is not provided.

The reason arch is not required is that `image_id` can be used as an argument instead of canonical.

jopemachine: This should be refactored as following work.

docs/manager/graphql-reference/schema.graphql
```diff
@@ -1975,7 +1975,12 @@ type Mutations {
   forget_image(architecture: String = "x86_64", reference: String!): ForgetImage @deprecated(reason: "Deprecated since 25.4.0. Use `forget_image_by_id` instead.")
 
   """Added in 25.4.0"""
-  purge_image_by_id(image_id: String!): PurgeImageById
+  purge_image_by_id(
+    image_id: String!
+
+    """Added in 25.5.0."""
+    options: PurgeImageByIdOptions = {remove_from_registry: false}
```
jopemachine: Should a single type like `image_id` also be wrapped in a new Key type? @HyeockJinKim 
(If so, that PR should also be considered a breaking change.)

docs/manager/graphql-reference/schema.graphql
```diff
@@ -2560,7 +2565,15 @@ type PurgeImageById {
   image: ImageNode
 }
 
-"""Added in 24.03.1"""
+"""Added in 25.5.0."""
+input PurgeImageByIdOptions {
```
HyeockJinKim: PurgeImageOptions seems to be sufficient.

src/ai/backend/manager/cli/image.py
```diff
@@ -59,10 +59,11 @@ def forget(cli_ctx, canonical_or_alias, architecture) -> None:
 @cli.command()
 @click.argument("canonical_or_alias")
 @click.argument("architecture")
+@click.option("--remove-from-registry", is_flag=True, help="Remove the image from the registry.")
 @click.pass_obj
-def purge(cli_ctx, canonical_or_alias, architecture) -> None:
+def purge(cli_ctx, canonical_or_alias, architecture, remove_from_registry) -> None:
```
HyeockJinKim: Can't you add a type in a function of click?

docs/manager/graphql-reference/schema.graphql
jopemachine: Since the mutation is named `PurgeImageById`, I think `PurgeImageByIdOptions` is a more intuitive name.
`PurgeImageOptions` makes it look like there's a mutation called `PurgeImage`.

src/ai/backend/client/func/image.py
jopemachine: The fields that the client wants to receive in the response should also be passed as arguments.

src/ai/backend/manager/cli/image.py
jopemachine: It’s simply because the existing code lacked type hints. Let’s add type hints.

docs/manager/graphql-reference/schema.graphql
HyeockJinKim: In my opinion, the fields in option appear to be options for the `purge image` behavior, which has nothing to do with `by id`.

jopemachine: In second thought, since the same option may be used in the `PurgeImages` mutation, I will change it to `PurgeImageOptions`.

HyeockJinKim (approved):

PR #3925, opened 2025-03-10
fix(BA-926): Properly handle zero-value unknown resource limits when creating session

jopemachine:
Resolves #3918 (BA-926).

## Manual test 

```
❯ ./backend.ai session create cr.backend.ai/stable/python:3.10-ubuntu22.04-amd64
∙ Session ID c504dbf7-feef-4ffd-a7a3-29702575cc84 is created and ready.
∙ This session provides the following app services: sshd, ttyd, jupyter, jupyterlab, vscode
```

---

**Checklist:** (if applicable)

- [x] Mention to the original issue

PR #3961, opened 2025-03-12
fix(BA-926): Properly handle zero-value unknown resource limits when creating session (#3925)

lablup-octodog:
This is an auto-generated backport PR of #3925 to the 23.09 release.


HyeockJinKim:
I think you don't need to apply 23.09, but please double check it. @jopemachine

PR #3972, opened 2025-03-12
fix: `PurgeImageById` mutation not working when an `image_alias` is present

### LINKED ISSUE #3952: `PurgeImageById` mutation not working when an `image_alias` is present
## Summary  

- An error occurs when calling the `PurgeImageById` mutation on an image that has an `image_alias`.

## Steps to Reproduce  

1. An alias for the image is created before deleting the image row.
1. Call `PurgeImageById` mutation on image that has an alias.

## Expected Behavior  

- Both `ImageAliasRow` and `ImageRow` are deleted together.

## Actual Behavior  

- Following error occurs.

## Logs/Errors  

```
2025-03-12 02:57:20.640 ERROR ai.backend.manager.api.admin [126119] ADMIN.GQL Exception: {'message': '(sqlalchemy.dialects.postgresql.asyncpg.IntegrityError) <class \'asyncpg.exceptions.NotNullViolationError\'>: null value in column "image" of relation "image_aliases" violates not-null constraint\nDETAIL:  Failing row contains (38c4723d-6098-4f57-a446-0c3cd6a00185, python, null).\n[SQL: UPDATE image_aliases SET image=%s WHERE image_aliases.id = %s]\n[parameters: (None, UUID(\'38c4723d-6098-4f57-a446-0c3cd6a00185\'))]\n(Background on this error at: https://sqlalche.me/e/14/gkpj)', 'locations': [{'line': 62, 'column': 3}], 'path': ['purge_image_by_id']}
```

jopemachine:
Resolves #3952 (BA-975)

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/manager/models/gql_models/image.py
```diff
@@ -806,6 +806,9 @@ async def mutate(
             if client_role != UserRole.SUPERADMIN:
                 if not image_row.is_customized_by(ctx.user["uuid"]):
                     raise GenericForbidden("Image is not owned by your account.")
+            for alias in image_row.aliases:
+                await session.delete(alias)
```
jopemachine, line 810: If possible, it would be better to execute this as a single query using bulk delete instead of this code.
However, executing the following query results in the following error.

```py
            await session.execute(
                sa.delete(ImageAliasRow)
                .where(ImageAliasRow.image_id == image_id)
            )
```

```
2025-03-12 04:41:54.578 ERROR ai.backend.manager.api.admin [234212] ADMIN.GQL Exception: {'message': "UPDATE statement on table 'image_aliases' expected to update 1 row(s); 0 were matched.", 'locations': [{'line': 62, 'column': 3}], 'path': ['purge_image_by_id']}
```

jopemachine: It seems to be this issue caused by inconsistency between the database and ORM because `ImageRow.get()` is called first, caching `ImageAliasRow` due to the `load_alias=True` option.

jopemachine: Do you happen to know a good solution for this situation? @fregataa

HyeockJinKim (approved):

PR #3984, opened 2025-03-13
example(BA-994): action architecture code example

HyeockJinKim:
resolves (BA-994)

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/manager/actions/processor.py
```diff
@@ -11,9 +13,16 @@
...
 
+class ActionValidator(ABC):
+    async def validate(self, action: BaseAction) -> None:
+        pass
+
+
 class ActionProcessor(Generic[TAction, TActionResult]):
     _monitors: list[ActionMonitor]
+    validates: list[ActionValidator]
     _func: Callable[[TAction], TActionResult]
```
seedspirit, line 24: Since most of the functions in Backend.AI are async functions, it would be better to type func as `Callable[[TAction], TActionResult | Awaitable[TActionResult]]`. 

Also, if the object that gives the following result is also awaitable, it would be better to have `result = await self._func(action)` instead of `result = self._func(action)`

HyeockJinKim: I'm just scratching the surface right now, so there's a lot of things I haven't thought through to work with.
Think of `_func` as working to always return awaitable.

HyeockJinKim: Please update once. @seedspirit

src/ai/backend/manager/services/image/actions.py
jopemachine, line 1: In this example PR, all actions are declared in a single file, but In actual work, each file should be separated by action and processed individually.
@fregataa @seedspirit @HyeockJinKim

src/ai/backend/manager/actions/processor.py
seedspirit: Updated it in https://github.com/lablup/backend.ai/pull/3999 PR 🚀

HyeockJinKim:
Now that we're almost done, I'll close this example code.

PR #3987, opened 2025-03-13
feat(BA-820): Add `force`, `noprune` options to `PurgeImages` GQL API

### LINKED ISSUE #3988: Add `force`, `noprune` option to `PurgeImages` GQL API
## Motivation  

When the administrator runs "purge images," there may be cases where certain options need to be used.

## Required Features

The purge images CLI command and GQL arguments will include the force and noprune options.

## Impact  

The administrator will be able to modify these options when running the purge images command.

## Testing Scenarios  

For example, when calling purge images on an image that is currently in use, an error occurs because the default value of the force option is False, preventing the image from being deleted.

Check that when the force option is set to True, the error does not occur and the in-use image is deleted immediately.

jopemachine:
Resolves #3988 (BA-820).

> [!WARNING]
> This PR includes a breaking change.

----
📚 Documentation preview 📚: https://sorna--3987.org.readthedocs.build/en/3987/

----
📚 Documentation preview 📚: https://sorna-ko--3987.org.readthedocs.build/ko/3987/


docs/manager/graphql-reference/schema.graphql
```diff
@@ -1984,7 +1984,7 @@ type Mutations {
   clear_images(registry: String): ClearImages
 
   """Added in 25.4.0"""
-  purge_images(agent_id: String!, images: [ImageRefType]!): PurgeImages
+  purge_images(agent_id: String!, images: [ImageRefType]!, props: PurgeImageProps): PurgeImages
```
HyeockJinKim: QUERY: func(key KeyStruct, options Options (filter, order, query filter...))
---
CREATE: func(props CreateProps) pattern
UPDATE: func(key KeyStruct, updateProps UpdateProps) pattern
DELETE: func(key KeyStruct, options OptionProps) pattern
---
Now we need to fit it into the pattern above.

docs/manager/graphql-reference/schema.graphql
```diff
@@ -1984,7 +1984,7 @@ type Mutations {
   clear_images(registry: String): ClearImages
 
   """Added in 25.4.0"""
-  purge_images(agent_id: String!, images: [ImageRefType]!): PurgeImages
+  purge_images(key: PurgeImagesKey!, options: PurgeImagesOptions): PurgeImages
```
yomybaby: It's better to change the response type name from `PurgeImages` to `PurgeImagesPayload` or another name.

yomybaby (changes requested):

src/ai/backend/manager/registry.py
```diff
@@ -3653,13 +3654,9 @@ async def get_agent_local_config(
         async with self.agent_cache.rpc_context(agent_id) as rpc:
             return await rpc.call.get_local_config()
 
-    async def purge_images(
-        self,
-        agent_id: AgentId,
-        images: list[str],
-    ) -> PurgeImageResponses:
+    async def purge_images(self, agent_id: AgentId, request: PurgeImagesReq) -> PurgeImageResponses:
```
jopemachine: Would `PurgeImagesRequest` be better than `PurgeImagesReq`?
Let’s standardize it one way or the other. @HyeockJinKim 
(And I think this should apply the same to `Response` and `Resp`.)

src/ai/backend/agent/docker/agent.py
```diff
@@ -1687,18 +1688,23 @@ async def pull_image(
             elif error := result[-1].get("error"):
                 raise RuntimeError(f"Failed to pull image: {error}")
 
-    async def _purge_image(self, docker: Docker, image: str) -> PurgeImageResponse:
+    async def _purge_image(
+        self, docker: Docker, image: str, force: bool, noprune: bool
+    ) -> PurgeImageResponse:
```
HyeockJinKim: Do we really need to `unpack` arguments here?

src/ai/backend/manager/models/gql_models/image.py
```diff
@@ -1129,26 +1130,59 @@ def __str__(self) -> str:
         return f"PurgeImagesResult:\n  Reserved Bytes: {self.reserved_bytes}\n  Results:\n  {results_str}"
 
 
+class PurgeImagesKey(graphene.InputObjectType):
+    """
+    Added in 25.5.0.
+    """
+
+    agent_id = graphene.String(required=True)
+    images = graphene.List(ImageRefType, required=True)
```
HyeockJinKim, line 1147: Wouldn't it be nice to get a class with agent_id, image_id as a list?

src/ai/backend/manager/registry.py
HyeockJinKim: Let's standardize all as `Req`, `Resp`.

jopemachine: In that case, for `PurgeImageResponseList`, should it be written as `PurgeImageResponses` or `PurgeImageResps`?

Personally, I think using `PurgeImageResponses` for the plural form is easier to understand.

HyeockJinKim: `PurgeImagesResponse` (Not multiple responses but multiple images.)

jopemachine: What I intended is a list of responses.

```py
@dataclass
class PurgeImageResponses(AbstractAgentResp):
    responses: list[PurgeImageResp]

    @override
    def as_dict(self) -> dict:
        return asdict(self)
```

HyeockJinKim: I believe this is a response to an action called `purge images`, not a summary of multiple `purge image` responses.

src/ai/backend/agent/server.py
```diff
@@ -927,9 +928,12 @@ async def _push_image(reporter: ProgressReporter) -> None:
...
     @collect_error
-    async def purge_images(self, images: list[str]) -> PurgeImageResponses:
-        log.info("rpc::purge_images(images:{0})", images)
-        return await self.agent.purge_images(images)
+    async def purge_images(
+        self, image_canonicals: list[str], force: bool, noprune: bool
+    ) -> PurgeImagesResp:
+        request = PurgeImagesReq(images=image_canonicals, force=force, noprune=noprune)
+        log.info("rpc::purge_images(request:{0})", request)
+        return await self.agent.purge_images(request)
```
HyeockJinKim: Hmm... req shouldn't be enclosed here...

src/ai/backend/agent/server.py
```diff
@@ -935,9 +936,18 @@ async def _push_image(reporter: ProgressReporter) -> None:
...
+    ) -> PurgeImagesResp:
+        log.info(
+            "rpc::purge_images(images:{0}, force:{1}, noprune:{2})",
+            image_canonicals,
+            force,
+            noprune,
+        )
+        return await self.agent.purge_images(
+            PurgeImagesReq(images=image_canonicals, force=force, noprune=noprune)
+        )
```
HyeockJinKim, line 950: this wasn't the intention, but it seems to be a minor detail that keeps delaying things, so I'll let it slide this time.

HyeockJinKim (approved):

yomybaby (approved):
I reviewed GraphQL schema only. LGTM!

PR #3989, opened 2025-03-13
feat(BA-922): Add internal address config

### LINKED ISSUE #3905: Separate Prometheus ports and adding options
## Motivation  

- The prometheus metrics can be accessed from a common port, allowing server metrics to be visible to all users.
- It would be nice to have a flag to turn metric collection on/off, as it might affect metric collection.

## Required Features

- Add a config to allow you to specify the prometheus port.
- Add a config to turn prometheus metrics collection on/off.

HyeockJinKim:
resolves #3905 (BA-922)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/storage/config.py
+++ b/src/ai/backend/storage/config.py
@@ -100,7 +100,7 @@ local_config_iv = (
                             t.Key("service-addr"): tx.HostPortPair(
                                 allow_blank_host=True,
                             ),
-                            t.Key("internal-addr"): tx.HostPortPair(
+                            t.Key("internal-addr", default=("127.0.0.1", 6023)): tx.HostPortPair(
                                 allow_blank_host=True,
                             ),
                             t.Key("ssl-enabled"): t.ToBool,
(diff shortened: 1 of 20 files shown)


fregataa (commented):
1. I checked the #3905 and the issue mentions about "a config to turn prometheus metrics collection on/off". Is it included here?
2. storage proxy does not run
image

src/ai/backend/storage/config.py
```diff
@@ -100,6 +100,9 @@
                             t.Key("service-addr"): tx.HostPortPair(
                                 allow_blank_host=True,
                             ),
+                            t.Key("internal-addr"): tx.HostPortPair(
+                                allow_blank_host=True,
+                            ),
```
fregataa: How about setting a default value here?

HyeockJinKim: I added it. @fregataa

fregataa (approved):
LGTM

PR #3996, opened 2025-03-17
fix(BA-1006): superadmin query agent nodes

### LINKED ISSUE #3995: Superadmin cannot query all GQL agent_nodes types
Superadmin sometimes cannot query all GQL agent_nodes types. This is because of wrong permission check

fregataa:
resolves #3995 (BA-1006)

### DIFF
--- a/src/ai/backend/manager/models/gql_models/agent.py
+++ b/src/ai/backend/manager/models/gql_models/agent.py
@@ -31,6 +31,7 @@ from ai.backend.common.types import (
 from ai.backend.logging.utils import BraceStyleAdapter
 
 from ..agent import (
+    ADMIN_PERMISSIONS,
     AgentRow,
     AgentStatus,
     agents,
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/models/gql_models/agent.py
```diff
@@ -305,21 +306,28 @@ async def get_connection(
...
+                    return ADMIN_PERMISSIONS
+
+                permission_getter = all_permissions
             async with graph_ctx.db.begin_readonly_session(db_conn) as db_session:
                 agent_rows = (await db_session.scalars(query)).all()
                 total_cnt = await db_session.scalar(cnt_query)
         result: list[AgentNode] = [
-            cls.parse(info, row, await permission_ctx.calculate_final_permission(row))
-            for row in agent_rows
+            cls.parse(info, row, await permission_getter(row)) for row in agent_rows
```
HyeockJinKim, line 330: Let's just go ahead because it's a hotfix, and then when we refactor our code with action architecture, let's change it so that if you're super admin, it works without permission checks in validator layer.

HyeockJinKim (approved):

PR #3999, opened 2025-03-17
feat(BA-1008): Make action processor work with async functions

### LINKED ISSUE #3998: Support async functions in ActionProcessor
## Motivation  

Current ActionProcessor only supports syncronous functions. Since most of the functions in [http://Backend.AI](http://Backend.AI|smart-link)  are async functions, it would be better to type func as `Callable[[TAction], TActionResult | Awaitable[TActionResult]]`.

Also, if the object that gives the following result is also awaitable, it would be better to have `result = await self._func(action)` instead of `result = self._func(action)`

## Objective  

- Make ActionProcessor support async function calling

## Details  

1. Change type hint to `Callable[[TAction], TActionResult | Awaitable[TActionResult]]`
1. if func result is awaitable, call`result = await self._func(action)`

## Impact   

Avoid weird function behavior during the introduction of Action-Process structures and enable asynchronous programming

seedspirit:
resolves #3998 (BA-1008)

Current processor only support sync function processing. But as most of Backend.AI functions are async functions, it needs to changed to process async func processing

Previous:
```python
class ActionProcessor(Generic[TAction, TActionResult]):
    _monitors: list[ActionMonitor]
    _func: Callable[[TAction], TActionResult]

    def __init__(
        self,
        func: Callable[[TAction], TActionResult],
        monitors: Optional[list[ActionMonitor]] = None,
    ) -> None:
        self._func = func
        self._monitors = monitors or []

    async def _run(self, action: TAction) -> ProcessResult[TActionResult]:
        started_at = datetime.now()
        status: str
        try:
            result = self._func(action)
            status = "success"
            description = "Success"
```

After:
```python
class ActionProcessor(Generic[TAction, TActionResult]):
    _monitors: list[ActionMonitor]
    _func: Callable[[TAction], Awaitable[TActionResult]]

    def __init__(
        self,
        func: Callable[[TAction], Awaitable[TActionResult]],
        monitors: Optional[list[ActionMonitor]] = None,
    ) -> None:
        self._func = func
        self._monitors = monitors or []

    async def _run(self, action: TAction) -> ProcessResult[TActionResult]:
        started_at = datetime.now()
        status: str
        try:
            result = await self._func(action)
            status = "success"
            description = "Success"
```

### DIFF
--- a/src/ai/backend/manager/actions/processor.py
+++ b/src/ai/backend/manager/actions/processor.py
@@ -1,6 +1,7 @@
 import asyncio
 from datetime import datetime
-from typing import Callable, Generic, Optional
+from inspect import isawaitable
+from typing import Awaitable, Callable, Generic, Optional
 
 from .action import (
     BaseActionResultMeta,
@@ -13,11 +14,11 @@ from .monitors.monitor import ActionMonitor
 
 class ActionProcessor(Generic[TAction, TActionResult]):
     _monitors: list[ActionMonitor]
-    _func: Callable[[TAction], TActionResult]
+    _func: Callable[[TAction], TActionResult | Awaitable[TActionResult]]
 
     def __init__(
         self,
-        func: Callable[[TAction], TActionResult],
+        func: Callable[[TAction], TActionResult | Awaitable[TActionResult]],
         monitors: Optional[list[ActionMonitor]] = None,
     ) -> None:
         self._func = func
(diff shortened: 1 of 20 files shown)


src/ai/backend/manager/actions/processor.py
```diff
@@ -13,11 +14,11 @@
 
 class ActionProcessor(Generic[TAction, TActionResult]):
     _monitors: list[ActionMonitor]
-    _func: Callable[[TAction], TActionResult]
+    _func: Callable[[TAction], TActionResult | Awaitable[TActionResult]]
 
     def __init__(
         self,
-        func: Callable[[TAction], TActionResult],
+        func: Callable[[TAction], TActionResult | Awaitable[TActionResult]],
```
HyeockJinKim: Can you make it so that only the async return value goes in?

seedspirit: Sure, I will fix it

PR #4001, opened 2025-03-17
feat(BA-1009): Add `AuditLog` GQL interface

### LINKED ISSUE #4000: Add AuditLog GQL interface
## Motivation  

- We need to create a GQL interface that matches the newly added AuditLog.

## Required Features

- Pagination should be supported.
- It should allow flexible filtering and querying through each field.

## Impact  

- The client will be able to freely query the AuditLog table.

## Testing Scenarios  

- Manual test

jopemachine:
Resolves #4000 (BA-1009).

----
📚 Documentation preview 📚: https://sorna--4001.org.readthedocs.build/en/4001/

----
📚 Documentation preview 📚: https://sorna-ko--4001.org.readthedocs.build/ko/4001/


docs/manager/graphql-reference/schema.graphql
```diff
@@ -285,6 +288,90 @@ interface Node {
...
+  cursor: String!
+}
+
+"""Added in 25.5.0."""
+type AuditLogNode implements Node {
+  """The ID of the object"""
+  id: ID!
+  row_id: UUID!
+  entity_type: AuditLogEntityType!
+  operation: ImageAuditLogOperationType!
```
jopemachine: Currently, the only Entity type is Image, so the operation type is also limited to Image, which feels a bit awkward. It would be good to find a way to make the operation type more clearly defined.

docs/manager/graphql-reference/schema.graphql
```diff
@@ -9,6 +9,37 @@ type Queries {
...
+    """Queries the `last` number of results from the query result from last."""
+    after: String
+
+    """
+    Queries the `first` number of results from the query result from first.
+    """
+    first: Int
+
+    """If the given value is provided, the query will start from that value."""
+    last: Int
```
HyeockJinKim, line 41: Is it possible to make them types and bundle them together instead of listing them separately?

jopemachine: Yes, in order to do that, I think we need to define `AsyncPaginatedConnectionV2Field`.

https://github.com/lablup/backend.ai/blob/560953388c41f91596c6095aa60f53edd3c7a56c/src/ai/backend/manager/models/base.py#L1420-L1425

I’ll try making the change.

jopemachine: Looking again, it seems that `IterableConnectionField`, the parent class of `AsyncListConnectionField` (a type from graphene), dynamically adds `before`, `after`, `first`, and `last`.

It looks like we need to define and use our own version of `IterableConnectionField`.

HyeockJinKim: Isn't this a work in progress? @jopemachine

HyeockJinKim (approved):

PR #4021, opened 2025-03-20
fix: Improve handling of process termination races in the krunner's OOM logger (#4008)

lablup-octodog:
This is an auto-generated backport PR of #4008 to the 24.09 release.


jopemachine:
This backport is not merged. Is it intentional? @achimnol

PR #4037, opened 2025-03-26
fix(BA-1001): model service exceeding amount of session deleted

### LINKED ISSUE #3985: Fix bug where model service would terminate more sessions than intended
## Impact

The model service is terminating more sessions than intended at certain points. Users experience a situation where fewer sessions are maintained than desired. This can lead to service unavailability during scaling down.

## Expected behaviour

The model service should terminate only the intended number of sessions without dropping below the desired count.

## Actual behaviour

At specific times, the model service is terminating more sessions than intended. Although it eventually adjusts to the desired number, there are moments when fewer sessions are maintained.

## Steps to reproduce

- Monitor the model service during scaling down.
- Observe the number of sessions being terminated.
- Compare the terminated sessions with the intended count.

## Environment

N/A

## Workarounds

N/A

## Other information

- The issue needs to be fixed quickly to prevent service unavailability.
- Reference link: [Jonghyun Park: 20250312 model service replica update crash?](https://teams.microsoft.com/l/message/19:4d7932db9b2c47e8883acfc523c4d3f5@thread.tacv2/1741747268931?tenantId=13c6a44d-9b52-4b9e-aa34-0513ee7131f2&groupId=b1f3bcf4-facf-40d3-94ab-169abb4f0f52&parentMessageId=1741747268931&teamName=customers%20%26%20clients&channelName=42dot&createdTime=1741747268931)

kyujin-cho:
resolves #3985 (BA-1001). This PR updates manager's `scale_services()`  scheduler to exclude routings in `TERMINATING` status when calculating active routings, which in turn its count will be used as a "number of replica" definition.


src/ai/backend/manager/models/routing.py
```diff
@@ -34,6 +34,10 @@ class RouteStatus(Enum):
     PROVISIONING = "provisioning"
     FAILED_TO_START = "failed_to_start"
 
+    @classmethod
+    def active_route_statuses(cls) -> set["RouteStatus"]:
+        return {RouteStatus.HEALTHY, RouteStatus.UNHEALTHY, RouteStatus.PROVISIONING}
```
HyeockJinKim, line 39: Are you sure these are active statuses?
Also, it looks like "RouteStatus" can be used as a Self.

kyujin-cho: It will depend on the situation, but since the meaning of `active` here implies that the depending session is alive and user has no intent of terminating the session I think it is legit to call such statuses as `active`.

kyujin-cho: > Also, it looks like "RouteStatus" can be used as a Self.

Seems like you are talking about the `typing.Self` pattern. I will apply it.

HyeockJinKim (approved):

src/ai/backend/manager/models/routing.py
kyujin-cho: image
Looks like mypy does not like the idea...

PR #4048, opened 2025-03-27
feat: Add Action Tests for `Image`

jopemachine:
resolves #NNN (BA-MMM)

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/manager/models/image.py
```diff
@@ -301,7 +301,7 @@ async def rescan_images(
     # TODO: delete images removed from registry?
 
 
-type Resources = dict[SlotName, dict[str, Decimal]]
+type Resources = dict[SlotName, dict[str, Any]]
```
HyeockJinKim, line 303: Is there a reason you changed this? @jopemachine

src/ai/backend/manager/models/base.py
```diff
@@ -1355,6 +1356,10 @@ async def populate_fixture(
         async with engine.begin() as conn:
             # Apply typedecorator manually for required columns
             for col in table.columns:
+                if isinstance(col.type, sa.sql.sqltypes.DateTime):
+                    for row in rows:
+                        if col.name in row:
+                            row[col.name] = isoparse(row[col.name])
```
HyeockJinKim, line 1362: Wasn't there a datetime value acceptor before? Why did you add it?

src/ai/backend/manager/models/image.py
jopemachine: There was a type error in the code below, so I temporarily changed it to Any, but I think it would be better to use `int | str` or define an appropriate type.

https://github.com/lablup/backend.ai/pull/4048/files#diff-b22639c8c2dcbef7f7b7443c407831448d71c4140036cce9311c13cd59bbb48cR85-R89

src/ai/backend/manager/models/container_registry.py
```diff
@@ -154,19 +176,20 @@ async def get_known_container_registries(
...
+            type=data.type,
+            project=data.project,
+            username=data.username,
+            password=data.password,
+            ssl_verify=data.ssl_verify,
+            is_global=data.is_global,
+            extra=data.extra,
         )
+        instance.id = data.id
+        return instance
```
HyeockJinKim, line 192: Why don't you put the ID value at creation time?

tests/manager/services/container_registry/conftest.py
```diff
@@ -0,0 +1,12 @@
...
+from ai.backend.manager.services.container_registry.processors import ContainerRegistryProcessors
+from ai.backend.manager.services.container_registry.service import ContainerRegistryService
+
+
+@pytest.fixture
+def processors(extra_fixtures, database_fixture, database_engine):
+    container_registry_service = ContainerRegistryService(
+        db=database_engine,
+    )
+    return ContainerRegistryProcessors(container_registry_service, [])
```
HyeockJinKim, line 12: rename to `container_registry_processors`

src/ai/backend/manager/models/container_registry.py
jopemachine: Since the ID is injected from the database, the ID value passed through the constructor wouldn't be passed to the database, so I intentionally excluded it from the constructor.

tests/manager/services/container_registry/conftest.py
jopemachine: Since it's the processors defined in `container_registry/conftest.py`, wouldn't it be fine to just leave it as processors?

tests/manager/services/image/conftest.py
```diff
@@ -0,0 +1,11 @@
...
+
+from ai.backend.manager.services.image.processors import ImageProcessors
+from ai.backend.manager.services.image.service import ImageService
+
+
+@pytest.fixture
+def processors(extra_fixtures, database_fixture, database_engine, registry_ctx):
+    agent_registry, _, _, _, _, _, _ = registry_ctx
+    image_service = ImageService(db=database_engine, agent_registry=agent_registry)
+    return ImageProcessors(image_service, [])
```
HyeockJinKim, line 11: rename to `image_processors`

src/ai/backend/manager/models/base.py
jopemachine: In the previous code for `populate_fixture`, the logic for handling `DateTime` types was missing, so an error occurs when generating a fixture that includes `DateTime` columns like `ImageRow.created_at`.

tests/manager/services/container_registry/conftest.py
HyeockJinKim: In a situation where there may be multiple processors, it seems like a good idea to reduce the room for error.

jopemachine: I agree with you. But this file has only one processors.
I think it would be better to revise it later when there are multiple processors.

src/ai/backend/manager/models/image.py
HyeockJinKim: It's very strange that it's updated because of the test code, is that correct?

jopemachine: It seems that I added an incorrect type in the recent PR I submitted.

src/ai/backend/manager/models/base.py
```diff
@@ -1355,6 +1356,10 @@ async def populate_fixture(
...
                 if isinstance(col.type, EnumType):
```
HyeockJinKim: elif @jopemachine

PR #4056, opened 2025-03-28
feat(BA-1048): Add raise exception in action processor

### LINKED ISSUE #4053: Raise Exception in processor
## User story

- As a developer, when an exception occurs in the processor, I want to raise the exception so that it is not treated as a successful response.

## Context

The processor currently does not raise exceptions. As a result, errors are incorrectly handled as successful responses.

## Acceptance criteria

- Modify the code to ensure that exceptions are raised within the processor.
- Ensure that the system correctly identifies and handles exceptions.

## Other information

N/A

HyeockJinKim:
resolves #4053 (BA-1048)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/tests/manager/actions/test_processor.py
+++ b/tests/manager/actions/test_processor.py
@@ -1,0 +1,124 @@
+from dataclasses import dataclass
+from datetime import datetime
+from typing import Optional
+
+import pytest
+from tenacity import BaseAction
+
+from ai.backend.manager.actions.action import BaseActionResult, BaseActionResultMeta, ProcessResult
+from ai.backend.manager.actions.processor import ActionProcessor
+
+
+@dataclass
+class MockAction(BaseAction):
+    id: str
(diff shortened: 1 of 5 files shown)


tests/manager/actions/test_processor.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+async def mock_action_processor_func(action: MockAction) -> MockActionResult:
+    return MockActionResult(id=action.id)
+
+
+async def mock_exception_processor_func(action: MockAction) -> MockActionResult:
+    raise MockException("Mock exception")
+
+
```
seedspirit, line 82: Maybe we need to add `@pytest.mark.asyncio`

HyeockJinKim: Python 3.8 and later should be fine.

PR #4064, opened 2025-03-31
feat(BA-1034): Add abstract message queue

### LINKED ISSUE #4036: Add a redis event queue implementation
## Impact

The current implementation assumes all event queues use Redis. This may lead to issues when integrating with other message queue systems.

## Expected behaviour

The system should accept an abstract message queue as an argument and correctly pass the Redis message queue.

## Actual behaviour

The existing Redis message queue implementation has issues with XStream. It incorrectly acknowledges messages that are not fully processed and has problems with message loss when using `xread` and `xautoclaim`.

## Steps to reproduce

- Assume all event queues are using Redis.
- Implement an abstract message queue as an argument.
- Observe the handling of messages with the current Redis message queue implementation.

## Environment

N/A

## Workarounds

N/A

## Other information

- The current implementation has cases where the Redis message queue is misused with XStream.
- Issues include:
  - Removing code that acknowledges messages before processing is complete.
  - Fixing message loss issues when using `xread` and `xautoclaim`.

HyeockJinKim:
resolves #4036  (BA-1034)

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/common/events.py
```diff
@@ -1358,132 +1335,101 @@ async def dispatch_subscribers(
...
-                    )
-                    log.exception("EventDispatcher.consume(): unexpected-error")
-                except BaseException as e:
-                    self._metric_observer.observe_event_failure(
-                        event_type=event_type,
-                        duration=time.perf_counter() - start,
-                        exception=e,
-                    )
-                    raise
+        async for msg in self._msg_queue.consume_queue():  # type: ignore
```
jopemachine, line 1346: I think it would be good to add a type hint to the `msg` variable.

```suggestion
        async for msg in self._msg_queue.consume_queue():  # type: ignore
            msg = cast(MQMessage, msg)
```

src/ai/backend/common/events.py
```diff
@@ -1358,132 +1335,101 @@ async def dispatch_subscribers(
...
+            if self._closed:
+                return
+            event_type = "unknown"
+            start = time.perf_counter()
+            try:
+                decoded = msg.payload[b"name"].decode()
+                if decoded and isinstance(decoded, str):
```
jopemachine: Could `decoded` be `None` here?
What happens if decoded is an empty string `""`?

src/ai/backend/common/events.py
```diff
@@ -1358,132 +1335,101 @@ async def dispatch_subscribers(
...
+                    event_type = decoded
+                await self.dispatch_subscribers(
+                    decoded,
```
jopemachine: `event_name` or `decoded_event_name`?

src/ai/backend/agent/agent.py
```diff
@@ -683,25 +685,41 @@ async def __ainit__(self) -> None:
...
+        if self.local_config["agent"].get("use-experimental-redis-event-dispatcher"):
+            dispatcher_queue = HiRedisQueue(
+                stream_redis_config,
+                HiRedisMQArgs(
+                    stream_key="events",
+                    group_name=EVENT_DISPATCHER_CONSUMER_GROUP,
+                    node_id=node_id,
+                    db=REDIS_STREAM_DB,
+                ),
+            )
```
jopemachine:
```suggestion
        dispatcher_queue: AbstractMessageQueue
        if self.local_config["agent"].get("use-experimental-redis-event-dispatcher"):
            dispatcher_queue = HiRedisQueue(
                stream_redis_config,
                HiRedisMQArgs(
                    stream_key="events",
                    group_name=EVENT_DISPATCHER_CONSUMER_GROUP,
                    node_id=node_id,
                    db=REDIS_STREAM_DB,
                ),
            )
        else:
            dispatcher_queue = redis_mq

```

src/ai/backend/common/events.py
```diff
@@ -1159,76 +1153,59 @@ class EventDispatcher(aobject):
...
-                cancelled_tasks.append(self.subscriber_loop_task)
+            await self._consumer_taskgroup.shutdown()
+            await self._subscriber_taskgroup.shutdown()
+            if not self._consumer_loop_task.done():
+                self._consumer_loop_task.cancel()
+                cancelled_tasks.append(self._consumer_loop_task)
+            if not self._subscriber_loop_task.done():
+                self._subscriber_loop_task.cancel()
+                cancelled_tasks.append(self._subscriber_loop_task)
             await asyncio.gather(*cancelled_tasks, return_exceptions=True)
```
jopemachine, line 1206: Is it okay not to handle returned exceptions here?

src/ai/backend/common/events.py
HyeockJinKim: Should we be doing unnecessary casts because of mypy bug?

src/ai/backend/agent/agent.py
HyeockJinKim: This is an area of preference 😆

src/ai/backend/common/events.py
HyeockJinKim: I've left the existing logic in place.
It needs to be checked, but I don't want to deal with it in this PR.

src/ai/backend/common/events.py
HyeockJinKim: I left the existing logic untouched.
In general, the `EventDispatcher` will close when the server shuts down, so this shouldn't be a problem under normal circumstances.

tests/common/message_queue/test_redis_queue.py
jopemachine, line 1: Is the existing test for `ExperimentalEventDispatcher` (`hiredis`) no longer necessary? (test_hiredis_queue...)

src/ai/backend/common/message_queue/hiredis_queue.py
```diff
@@ -0,0 +1,439 @@
...
+from ai.backend.logging.utils import BraceStyleAdapter
+
+from ..types import RedisConfig
+from .queue import AbstractMessageQueue, MessageId, MQMessage
+
+log = BraceStyleAdapter(logging.getLogger(__spec__.name))
+
+_DEFAULT_AUTOCLAIM_IDLE_TIMEOUT = 1_000
+_DEFAULT_AUTOCLAIM_INTERVAL = 60_000
+_DEFAULT_AUTOCLAIM_COUNT = 64
```
jopemachine, line 21: `_DEFAULT_AUTOCLAIM_COUNT` seems not used.

src/ai/backend/common/message_queue/hiredis_queue.py
```diff
@@ -0,0 +1,278 @@
...
+                ],
+                command_timeout=(autoclaim_idle_timeout + 5_000) / 1000,
+            )
+            if reply is None or reply[0] == b"0-0":
+                log.debug("No messages to claim")
+                return autoclaim_start_id, False
+            for stream_msg in reply.values():
+                for msg_id, messages in stream_msg:
+                    payload = {}
+                    for i in range(0, len(messages), 2):
```
fregataa, line 170: How about using [itertools.pairwise()](https://docs.python.org/ko/3.13/library/itertools.html#itertools.pairwise)?

src/ai/backend/common/message_queue/hiredis_queue.py
```diff
@@ -0,0 +1,278 @@
...
+        log.debug("Reading messages from stream %s", self._stream_key)
+        while not self._closed:
+            try:
+                await self._read_messages()
+            except hiredis.HiredisError as e:
+                await self._failover_consumer(e)
+            except Exception as e:
+                log.error(f"Error while reading messages: {e}")
+
+    async def _read_messages(self):
```
fregataa:
```suggestion
    async def _read_messages(self) -> None:
```
mypy skips type checking if the return type is not defined. So let's add it

src/ai/backend/common/message_queue/hiredis_queue.py
```diff
@@ -0,0 +1,278 @@
...
+                    msg = MQMessage(msg_id, payload)
+                    await self._subscribe_queue.put(msg)
+                    last_msg_id = str(msg_id)
+        return last_msg_id
+
+    async def _failover_consumer(self, e: hiredis.HiredisError) -> None:
+        # If the group does not exist, create it
+        # and start the auto claim loop again
+        if not e.args[0].startswith("NOGROUP "):
+            log.error(f"Error while auto claiming messages: {e}")
```
fregataa:
```suggestion
            log.exception(f"Error while auto claiming messages: {e}")
```
(It is not a big suggestion) it seems to be an exception not an error

src/ai/backend/common/message_queue/hiredis_queue.py
```diff
@@ -0,0 +1,278 @@
...
+                ["XREAD", "BLOCK", "1000", "STREAMS", self._stream_key, last_msg_id],
+                command_timeout=5,
+            )
+            if reply is None:
+                log.debug("No messages to read")
+                return last_msg_id
+            for stream_msg in reply.values():
+                for msg_id, messages in stream_msg:
+                    payload = {}
+                    for i in range(0, len(messages), 2):
```
fregataa, line 268: How about using [itertools.pairwise()](https://docs.python.org/ko/3.13/library/itertools.html#itertools.pairwise)?

fregataa (commented):
It is good to detach the `XAUTOCLAIM` loop from the `READ` loop!
As I know, there are some components using `events_experimental` module. Such components should follow up this PR

src/ai/backend/common/message_queue/hiredis_queue.py
HyeockJinKim: If I do it with exception, I get a stack trace, which scares me with error

src/ai/backend/common/message_queue/redis_queue.py
```diff
@@ -218,7 +219,9 @@ async def _failover_consumer(self, e: redis.exceptions.ResponseError) -> None:
         # and start the auto claim loop again
         if "NOGROUP" in str(e):
             log.warning(
-                f"Consumer group {self._group_name} does not exist, creating it",
```
jopemachine: Is there a reason not to use f-strings?
I think f-string format looks more readable.

HyeockJinKim: By default, format string will always be formatted, but format in logs will not format If the loglevel is low.

src/ai/backend/common/message_queue/hiredis_queue.py
fregataa: My fault, pairwise() is not a proper option for this case.
We can use [more_itertools.batched()](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.batched) or [itertools.batched()](https://docs.python.org/ko/3.12/library/itertools.html#itertools.batched).
Since our Python version is 3.12.8 and built-in `itertools.batched()` does not support the `strict` option, it seems to good to use more_itertools in this case.

src/ai/backend/common/message_queue/hiredis_queue.py
fregataa: Same here, pairwise() is not a proper option for this case.
more_itertools.batched() or itertools.batched()?

fregataa (approved):
Except the comments, all look good

HyeockJinKim:
itertools.batched might be worth considering for a future version. @fregataa

PR #4068, opened 2025-04-01
fix(BA-1033): Setup source at producer creation

HyeockJinKim:
resolves #4034 (BA-1033)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/common/events.py
+++ b/src/ai/backend/common/events.py
@@ -1422,10 +1422,12 @@ class EventProducer:
         self,
         msg_queue: AbstractMessageQueue,
         *,
+        source: AgentId,
         log_events: bool = False,
     ) -> None:
         self._closed = False
         self._msg_queue = msg_queue
+        self._source_bytes = source.encode()
         self._log_events = log_events
 
     async def close(self) -> None:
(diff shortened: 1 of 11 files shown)


src/ai/backend/common/types.py
```diff
@@ -346,6 +346,8 @@ def check_typed_tuple(value: Tuple[Any, ...], types: Tuple[Type, ...]) -> Tuple:
 
 ResourceGroupID = NewType("ResourceGroupID", str)
 AgentId = NewType("AgentId", str)
+AGENTID_MANAGER = AgentId("manager")
+AGENTID_STORAGE = AgentId("storage")
 DeviceName = NewType("DeviceName", str)
```
jopemachine, line 351: `AgentId` is a type meant to hold the ID of an agent.
I don't think the manager and the storage proxy should be considered Agents, and using this kind of type is confusing.

src/ai/backend/common/events.py
```diff
@@ -1435,14 +1437,17 @@ async def close(self) -> None:
     async def produce_event(
         self,
         event: AbstractEvent,
-        *,
-        source: str = "manager",
+        source_override: Optional[AgentId] = None,
```
jopemachine, line 1440: It seems that the source should be able to include components other than just Agent.

jopemachine: How about defining and using a `NodeId` that includes an `AgentId`?

HyeockJinKim: Since the existing behavior is taking the agentId, I wanted to avoid modifying the event producer side as much as possible.

src/ai/backend/common/types.py
HyeockJinKim: This is a temporary addition, and like you said, I'd like to change it to accept values like sourceId.
This is a temporary field to at least ensure the behavior now.

PR #4104, opened 2025-04-07
feat(BA-1085): Add RequestID context & middleware

### LINKED ISSUE #4097: Pass requestID for monitoring
## User story

- As a developer, when handling requests, I want to pass the RequestID so that I can monitor each request effectively.
- As a developer, when producing events or sending external requests, I want to ensure the RequestID is wrapped and passed along so that tracking is consistent.

## Context

This issue involves passing a RequestID for each request. The RequestID will be generated or received in the middleware of the handler. It will be transmitted using context variables.

## Acceptance criteria

- The RequestID must be generated or received in the middleware.
- The RequestID should be passed using context variables.
- When producing events or sending external requests, the RequestID must be wrapped and included.

## Other information

N/A

HyeockJinKim:
resolves #4097 (BA-1085)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/common/contexts/request_id.py
+++ b/src/ai/backend/common/contexts/request_id.py
@@ -18,7 +18,7 @@ def current_request_id() -> Optional[str]:
 
 
 @contextmanager
-def with_request_id(request_id: Optional[str] = None) -> Iterator[str]:
+def with_request_id(request_id: Optional[str] = None) -> Iterator[None]:
     """
     context manager to set up the request ID.
     If request_id is not provided, generates a new UUID.
@@ -30,7 +30,7 @@ def with_request_id(request_id: Optional[str] = None) -> Iterator[str]:
 
     token = request_id_var.set(request_id)
     try:
-        yield token
+        yield
     finally:
         # Reset the context variable to its previous state
         request_id_var.reset(token)
--- a/src/ai/backend/common/contexts/request_id.py
+++ b/src/ai/backend/common/contexts/request_id.py
@@ -3,7 +3,7 @@ from contextlib import contextmanager
 from contextvars import ContextVar
 from typing import Iterator, Optional
 
-request_id_var: ContextVar[str] = ContextVar("request_id")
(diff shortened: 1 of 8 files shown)


tests/common/test_request_id.py
```diff
@@ -0,0 +1,39 @@
...
+
+def test_nested_request_id_contexts():
+    outer_id = str(uuid.uuid4())
+    inner_id = str(uuid.uuid4())
+
+    with with_request_id(outer_id):
+        assert current_request_id() == outer_id
+        with with_request_id(inner_id):
+            assert current_request_id() == inner_id
+        assert current_request_id() == outer_id
```
fregataa, line 38: Just a question. Is there any expected use case of such nested request id contextx?

fregataa (approved):
Looks good. I left a question comment

src/ai/backend/common/contexts/request_id.py
```diff
@@ -0,0 +1,36 @@
+import uuid
+from contextlib import contextmanager
+from contextvars import ContextVar
+from typing import Iterator, Optional
+
+request_id_var: ContextVar[str] = ContextVar("request_id")
```
jopemachine: Is there a reason why `request_id` is of type `str` instead of `UUID`?

tests/common/test_request_id.py
HyeockJinKim: No, it doesn't. This code is just to test that the with statement is clean and organized.

src/ai/backend/common/contexts/request_id.py
HyeockJinKim: ~Thanks. I need to replace it with `UUID`.~

HyeockJinKim: Oh, I left this part as `str` because of the Custom request ID.

HyeockJinKim: I just make `request_id_var` private.

PR #4129, opened 2025-04-10
fix(BA-1130): Cannot modify user info

### LINKED ISSUE #4128: Cannot modify user info
Cannot modify user info.
There are some bugs after action architecture refactor work.

fregataa:
resolves #4128 (BA-1130)

### DIFF
--- a/src/ai/backend/manager/services/group/types.py
+++ b/src/ai/backend/manager/services/group/types.py
@@ -88,8 +88,6 @@ class GroupModifier(PartialModifier):
     total_resource_slots: OptionalState[ResourceSlot] = field(
         default_factory=OptionalState[ResourceSlot].nop
     )
-    user_update_mode: OptionalState[str] = field(default_factory=OptionalState[str].nop)
-    user_uuids: OptionalState[list[str]] = field(default_factory=OptionalState[list[str]].nop)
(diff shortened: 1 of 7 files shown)


HyeockJinKim (approved):

src/ai/backend/manager/services/group/types.py
```diff
@@ -107,8 +107,6 @@ def fields_to_update(self) -> dict[str, Any]:
         self.is_active.update_dict(to_update, "is_active")
         self.domain_name.update_dict(to_update, "domain_name")
         self.total_resource_slots.update_dict(to_update, "total_resource_slots")
-        self.user_update_mode.update_dict(to_update, "user_update_mode")
-        self.user_uuids.update_dict(to_update, "user_uuids")
```
HyeockJinKim, line 111: If it's not a modifiable value, could you take it out of the modifier and put it in the action?

fregataa: It is not modifiable in groups table but it can update the association table `association_groups_users` which associates groups and users table. Should I consider the `...Modifier` types as DB table updaters?

fregataa: We agreed to set `...Modifier` types as DB table updaters.

PR #4167, opened 2025-04-14
feat(BA-1155): Add prometheus monitor

### LINKED ISSUE #4166: Add processor metrics
## User story

- As a developer, when the action architecture is changed, I want to collect additional Prometheus metrics from the processor layer so that I can monitor performance effectively.

## Context

The issue arises from a change to the action architecture. This change necessitates the collection of additional metrics from the processor layer using Prometheus.

## Acceptance criteria

- Additional Prometheus metrics must be defined and implemented for the processor layer.
- Metrics should be tested to ensure they are collected correctly.

## Other information

N/A

HyeockJinKim:
resolves #4166 (BA-1155)

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/manager/actions/monitors/audit_log.py
jopemachine, line 1: Why is this deleted?

src/ai/backend/manager/actions/monitors/smtp.py
jopemachine, line 1: Why is this deleted?

src/ai/backend/manager/actions/monitors/audit_log.py
HyeockJinKim: I didn't use it, so I removed it.

src/ai/backend/manager/actions/monitors/prometheus.py
```diff
@@ -0,0 +1,25 @@
+from typing import override
+
+from ai.backend.common.metrics.metric import ActionMetricObserver
+from ai.backend.manager.actions.action import BaseAction, BaseActionTriggerMeta, ProcessResult
+from ai.backend.manager.actions.monitors.monitor import ActionMonitor
+
+
+class PrometheusMonitor(ActionMonitor):
```
jopemachine, line 8: I think this should be `PrometheusReporter`, not Monitor

HyeockJinKim: The reporter only wants to apply tasks that are okay to be done in other layers, like message queue,
and the monitor wants to attach tasks that can be attached directly, so the Monitor implementation makes sense here.

jopemachine (approved):

PR #4176, opened 2025-04-15
feat(BA-1139): Validate SSH keypair when upload

### LINKED ISSUE #4142: Add keypair validation in keypair upload api
## Summary

There is no validation for keypairs during the upload process. This issue needs to be addressed to ensure that invalid keypairs are filtered out at the time of upload.

## Context

Currently, the keypair upload API does not perform any validation checks. This can lead to the acceptance of invalid keypairs, which may cause issues later in the process.

## Acceptance criteria

- Implement validation for keypairs during the upload process.
- Ensure that invalid keypairs are rejected at the time of upload.

## Other information

N/A

fregataa:
resolves #4142 (BA-1139)

### DIFF
--- a/src/ai/backend/manager/api/auth.py
+++ b/src/ai/backend/manager/api/auth.py
@@ -30,6 +30,7 @@ from ..models import keypair_resource_policies, keypairs, user_resource_policies
 from ..models.group import association_groups_users, groups
 from ..models.keypair import generate_keypair as _gen_keypair
 from ..models.keypair import generate_ssh_keypair as _gen_ssh_keypair
+from ..models.keypair import validate_ssh_keypair
 from ..models.user import (
     INACTIVE_USER_STATUSES,
     UserRole,
@@ -1129,6 +1130,10 @@ async def upload_ssh_keypair(request: web.Request, params: Any) -> web.Response:
     log_args = (domain_name, access_key)
     log.info(log_fmt, *log_args)
     root_ctx: RootContext = request.app["_root.context"]
+
+    is_valid, err_msg = validate_ssh_keypair(privkey, pubkey)
+    if not is_valid:
+        raise InvalidAPIParameters(err_msg)
     async with root_ctx.db.begin() as conn:
         data = {
             "ssh_public_key": pubkey,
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/models/keypair.py
```diff
@@ -148,6 +152,48 @@ def generate_ssh_keypair() -> Tuple[str, str]:
...
+        private_key_value.encode(),
+        password=None,  # No encryption as specified
+    )
+    public_key = crypto_serialization.load_ssh_public_key(public_key_value.encode())
+
+    # Check that both are RSA keys
+    if not isinstance(private_key, rsa.RSAPrivateKey) or not isinstance(
+        public_key, rsa.RSAPublicKey
+    ):
+        return False, "Keypair is not RSA"
```
HyeockJinKim: With so many algorithms like ECDSA and EDDSA, is it okay to just check RSA?

fregataa: Updated to cover ECDSA and ED25519 but not sure how many other algorithms to support

src/ai/backend/manager/api/auth.py
```diff
@@ -1129,6 +1130,10 @@ async def upload_ssh_keypair(request: web.Request, params: Any) -> web.Response:
     log_args = (domain_name, access_key)
     log.info(log_fmt, *log_args)
     root_ctx: RootContext = request.app["_root.context"]
+
+    is_valid, err_msg = validate_ssh_keypair(privkey, pubkey)
+    if not is_valid:
+        raise InvalidAPIParameters(err_msg)
```
HyeockJinKim, line 1136: Wouldn't it be better to throw an exception internally?

fregataa: Raising `InvalidAPIParameters` turns into returning 400 HTTP response because the `InvalidAPIParameters` inherits HTTPBadRequest exception. Should it return a bad-request response?

HyeockJinKim: Well... we'll clean up the exception in a later pick.

HyeockJinKim (approved):

PR #4220, opened 2025-04-18
feat: Add `service-config.toml` to `VFolderNode` GQL (WIP)

jopemachine:
resolves #NNN (BA-MMM)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/models/gql_models/vfolder.py
+++ b/src/ai/backend/manager/models/gql_models/vfolder.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 import uuid
 from collections.abc import Iterable, Mapping, Sequence
 from datetime import datetime
@@ -13,6 +14,7 @@ from typing import (
 import graphene
 import graphql
 import sqlalchemy as sa
+import tomli
 import trafaret as t
 from dateutil.parser import ParserError
 from dateutil.parser import parse as dtparse
@@ -117,6 +119,8 @@ class VirtualFolderNode(graphene.ObjectType):
         VFolderPermissionValueField,
         description=f"Added in 24.09.0. One of {[val.value for val in VFolderRBACPermission]}.",
     )
+
+    service_config = graphene.JSONString(description="Added in 25.7.0.")
 
     _queryfilter_fieldspec: Mapping[str, FieldSpecItem] = {
         "id": ("id", uuid.UUID),
@@ -255,6 +259,7 @@ class VirtualFolderNode(graphene.ObjectType):
                 graph_ctx.db, user["domain_name"], user["uuid"], user["role"]
             )
             permission_ctx = await get_permission_ctx(db_conn, client_ctx, scope_id, permission)
+            print("permission_ctx!", permission_ctx)
             cond = permission_ctx.query_condition
             if cond is None:
                 return None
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/models/vfolder.py
```diff
@@ -1647,6 +1649,23 @@ async def resolve_num_files(self, info: graphene.ResolveInfo) -> int:
...
+            await graph_ctx.processors.vfolder_file.fetch_service_config.wait_for_complete(
+                FetchServiceConfigAction(
+                    vfolder_uuid=self.id,
+                    quota_scope_id=self.quota_scope_id,
+                    host=self.host,
+                    unmanaged_path=self.unmanaged_path,
+                )
+            )
+        )
+        return json.dumps(action_result.result.model_dump())
```
HyeockJinKim: Does this look like a weird data structure? It looks like you're holding the DTO directly in the action result right now, which doesn't seem appropriate.

src/ai/backend/manager/services/vfolder/actions/file.py
```diff
@@ -10,6 +11,9 @@
 from ai.backend.common.types import ResultSet
 from ai.backend.manager.actions.action import BaseActionResult
 
+if TYPE_CHECKING:
+    from ai.backend.manager.data.vfolder.dto import ServiceConfig
+
```
HyeockJinKim: Make sure there is no dependency on DTOs within actions package.

src/ai/backend/manager/data/vfolder/dto.py
```diff
@@ -114,3 +116,14 @@ def to_field(self) -> VFolderItemField:
...
+
+class ServiceEnvironmentConfig(BaseModel):
+    image: str
+    architecture: str
+
+
+class ServiceConfig(BaseModel):
+    runtime_variant: RuntimeVariant
+    environment: ServiceEnvironmentConfig
+    resource_limits: dict[str, Any] = Field(default_factory=dict)
```
HyeockJinKim: I guess there should be a separation between the service config that goes into the dto and the internal use service config, but how are you thinking of proceeding with the part that uses the service config?

src/ai/backend/manager/models/gql_models/vfolder.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
```
HyeockJinKim: check `dump_json` in utils

src/ai/backend/manager/services/model_serving/services/model_serving.py
```diff
@@ -150,11 +154,60 @@ def __init__(
...
+                "vfid": str(vfid),
+                "relpath": f"./{filename}",
+            },
+        ) as (_, storage_resp):
+            while True:
+                chunk = await storage_resp.content.read(DEFAULT_CHUNK_SIZE)
+                if not chunk:
+                    break
+                chunks += chunk
+        return chunks
```
HyeockJinKim, line 183: I would like these actions to be separated into the client layer later.

src/ai/backend/manager/services/model_serving/services/model_serving.py
```diff
@@ -150,11 +154,60 @@ def __init__(
...
+                definition = action.creator.runtime_variant
+                if definition in service_definition:
+                    variant_def = ModelServiceDefinition.model_validate(
+                        service_definition[definition]
+                    )
+                    if variant_def.resource_slots:
+                        action.creator.config.resources = variant_def.resource_slots
+                    if variant_def.environment:
+                        action.creator.image = variant_def.environment.image
+                        action.creator.architecture = variant_def.environment.architecture
```
HyeockJinKim, line 207: Do you only override config.resources, image, and architecture?

jopemachine: Are there any other values that need to be overridden? If you let me know, I’ll make sure to reflect them.

HyeockJinKim: This part is something I don't know well right now, so I was asking as a question for confirmation. 😄

src/ai/backend/manager/services/model_serving/services/model_serving.py
```diff
@@ -150,11 +154,60 @@ def __init__(
...
         service_prepare_ctx = action.creator.model_service_prepare_ctx
-        async with self._db.begin_readonly_session() as session:
+
+        async with self._db.begin_readonly_session() as db_sess:
+            model_vfolder_row = await VFolderRow.get(db_sess, service_prepare_ctx.model_id)
+            chunks = await self._fetch_file_from_storage_proxy(
+                "service-definition.toml", model_vfolder_row
+            )
+
+            if chunks:
```
HyeockJinKim, line 194: This time, it was okay to add an if statement to work quickly,
but if we need to advance this part next time, please apply a structure that overrides service definitions, just like you did with the config loader. @jopemachine

PR #4002, opened 2025-03-17
refactor(BA-888): Add vfolder service and process

fregataa:
resolves #4083 (BA-888)

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version
- [x] Mention to the original issue
- [ ] API server-client counterparts (e.g., manager API -> client SDK)

PR #4223, opened 2025-04-18
fix(BA-1192): Resolve `KeyError` in Client SDK's vfolder `mkdir` command

### LINKED ISSUE #4221: KeyError raised from virtual folder `mkdir` command
## Motivation  

The `mkdir` command from the Python Client SDK raises a `KeyError` when attempting to create a virtual folder. This is caused by a mismatch between the expected and actual response schema — specifically, the absence of the `"results"` key in the response.

## Expected Behavior

The command should return a well-formed response containing the `"results"` key, even if it's an empty list or dict. If an error occurred, it should be reflected properly in the response code or error message — not through an unhandled exception.

## Actual Behavior

A `KeyError` is thrown when trying to access `reply["results"]`, even though the request may have succeeded. This results in a **false negative**, giving users the impression that the command failed.

## Impact  

This false-negative response can lead to confusion and harm the UX.

## Testing Scenarios  

- Execute `mkdir` once and check the response.

```
rapsealk@lablupui-MacBookPro backend.ai-dev % ./backend.ai vfolder mkdir new-pipeline-0-RLgfDi "task-0"
✘ KeyError: results
  *** Traceback ***
  File "/Users/rapsealk/Desktop/git/backend.ai-dev/src/ai/backend/client/cli/vfolder.py", line 488, in mkdir
      results = session.VFolder(name).mkdir(paths, parents=parents, exist_ok=exist_ok)
    File "/Users/rapsealk/Desktop/git/backend.ai-dev/src/ai/backend/client/func/base.py", line 34, in _method
      return _api_session.worker_thread.execute(coro)
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
    File "/Users/rapsealk/Desktop/git/backend.ai-dev/src/ai/backend/client/session.py", line 192, in execute
      raise result
    File "/Users/rapsealk/Desktop/git/backend.ai-dev/src/ai/backend/client/session.py", line 172, in run
      result = ctx.run(loop.run_until_complete, coro)  # type: ignore
    File "/Users/rapsealk/.pyenv/versions/3.13.3/lib/python3.13/asyncio/base_events.py", line 719, in run_until_complete
      return future.result()
             ~~~~~~~~~~~~~^^
    File "/Users/rapsealk/Desktop/git/backend.ai-dev/src/ai/backend/client/func/vfolder.py", line 583, in mkdir
      return await self._mkdir(path, parents, exist_ok)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/Users/rapsealk/Desktop/git/backend.ai-dev/src/ai/backend/client/func/vfolder.py", line 574, in _mkdir
      return reply["results"]  # TODO: ...KeyError: "results"
             ~~~~~^^^^^^^^^^^
```

rapsealk:
Follow-up of #4002 and resolves #4221 (BA-1192)

https://github.com/lablup/backend.ai/blob/aba0bffe0ae04fb6d1b387be384082db190f3cac/src/ai/backend/manager/api/vfolder.py#L1096

- [ ] Test case(s) to:

### DIFF
--- a/changes/.fix.md
+++ b/changes/.fix.md
@@ -1,0 +1,1 @@
+Fixed a `KeyError` in the Python Client SDK's vfolder `mkdir` command, caused by a missing "results" key in the API response
--- a/src/ai/backend/client/func/vfolder.py
+++ b/src/ai/backend/client/func/vfolder.py
@@ -570,8 +570,7 @@ class VFolderByName(BaseFunction):
             "exist_ok": exist_ok,
         })
         async with rqst.fetch() as resp:
-            # reply: {'success': [{'msg': None, 'item': 'task-1'}], 'failed': []}
-            reply = await resp.json()  # {'results': {'success': [], 'failed': [{'msg': "FileExistsError(17, 'File exists')", 'item': 'task-0'}]}}
+            reply = await resp.json()
             return reply
 
     @api_function
--- a/src/ai/backend/client/func/vfolder.py
+++ b/src/ai/backend/client/func/vfolder.py
@@ -570,8 +570,9 @@ class VFolderByName(BaseFunction):
             "exist_ok": exist_ok,
         })
         async with rqst.fetch() as resp:
-            reply = await resp.json()
-            return reply["results"]
+            # reply: {'success': [{'msg': None, 'item': 'task-1'}], 'failed': []}
+            reply = await resp.json()  # {'results': {'success': [], 'failed': [{'msg': "FileExistsError(17, 'File exists')", 'item': 'task-0'}]}}
+            return reply
 
     @api_function
     async def mkdir(


jopemachine:
I think it’s the API that needs to be updated, not the SDK. 
(Otherwise, it would be an API breaking change.)

rapsealk:
> I think it’s the API that needs to be updated, not the SDK. (Otherwise, it would be an API breaking change.)

@jopemachine It looks like most of the current test cases rely on mocks, but I’m not sure they’re sufficient. What do you think?

jopemachine:
> > I think it’s the API that needs to be updated, not the SDK. (Otherwise, it would be an API breaking change.)
> 
> @jopemachine It looks like most of the current test cases rely on mocks, but I’m not sure they’re sufficient. What do you think?

The test codes on the Client SDK side depend on REST APIs (and some GraphQL), and since those APIs are built using an Action-Service structure, I believe that as long as the Action test codes are well written going forward, the tests should be sufficient (Since SDK itself typically doesn't perform many things.)

rapsealk:
Superseded by #4227.

PR #4247, opened 2025-04-22
fix(BA-1217): Filter duplicate vfolder mounts

### LINKED ISSUE #4246: Cannot create SFTP session from auto-mount VFolders
Cannot create SFTP session from auto-mount VFolders due to “Duplicate mount point” error

fregataa:
resolves #4246 (BA-1217)


HyeockJinKim (approved):
The logic is a bit messy, but we'll stick with it for now.

PR #4232, opened 2025-04-21
fix(BA-1205): `ImageAliasData` type not found in `ImageAliasData.to_dataclass()`

jopemachine:
resolves #4233 (BA-1205).

**Checklist:** (if applicable)

- [x] Mention to the original issue

PR #4248, opened 2025-04-22
fix(BA-1205): `ImageAliasData` type not found in `ImageAliasData.to_dataclass()` (#4232)

lablup-octodog:
This is an auto-generated backport PR of #4232 to the 24.09 release.


jopemachine:
This doesn't need to be backported

PR #4272, opened 2025-04-25
fix(BA-1231): `BgTaskFailedError` is not propagated to the client

### LINKED ISSUE #4271: Commit quota exceeded error not passed to response
## Summary  

The error that occurs in the manager when the commit image quota is exceeded is not carried over to the response. Currently, a background task id is generated and returned.

## Steps to Reproduce  

1. Sets the maximum number of custom images in a user resource policy.
1. Commits more than the maximum number of custom images.Expected Behavior  

- When a commit that exceeds the maximum number of custom images occurs, an error should be passed as a response without creating a background task.

## Actual Behavior  

- The user cannot be notified of the error.

## Logs/Errors  

-

jopemachine:
resolves #4271 (BA-1231).

### DIFF
--- a/src/ai/backend/common/events/bgtask.py
+++ b/src/ai/backend/common/events/bgtask.py
@@ -1,3 +1,4 @@
+import logging
 import uuid
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
@@ -6,7 +7,6 @@ from typing import Optional, override
 from ai.backend.common.bgtask.types import BgtaskStatus
 from ai.backend.common.events.dispatcher import AbstractEvent, EventDomain
 from ai.backend.common.events.user_event.user_bgtask_event import (
-    UserBgtaskAlreadyDoneEvent,
     UserBgtaskCancelledEvent,
     UserBgtaskDoneEvent,
     UserBgtaskFailedEvent,
@@ -14,6 +14,9 @@ from ai.backend.common.events.user_event.user_bgtask_event import (
 )
 from ai.backend.common.events.user_event.user_event import UserEvent
 from ai.backend.common.exception import UnreachableError
+from ai.backend.common.logging import BraceStyleAdapter
+
+log = BraceStyleAdapter(logging.getLogger(__spec__.name))  # type: ignore
 
 
 @dataclass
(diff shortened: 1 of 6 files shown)


src/ai/backend/manager/services/session/service.py
```diff
@@ -479,22 +475,24 @@ async def _commit_and_upload(reporter: ProgressReporter) -> None:
...
                     reporter=reporter,
                 )
                 await reporter.update(increment=1, message="Completed")
-            except BackendError:
+                rescanned_image_ids = [image.id for image in rescan_result.images]
+                return DispatchResult.success(rescanned_image_ids)
+            except BackendError as e:
                 log.exception("CONVERT_SESSION_TO_IMAGE: exception")
-                raise
+                return DispatchResult.error(str(e))
```
HyeockJinKim: It doesn't seem like a good idea that we should always match raising an error here, but we'll get to that later.

src/ai/backend/common/bgtask.py
```diff
@@ -239,10 +239,17 @@ async def poll_bgtask_event(
...
             # It is an already finished task!
+            event_cls: type[BgtaskEvents]
+            if status == "failed":
+                event_cls = BgtaskFailedEvent
+            else:
+                event_cls = BgtaskDoneEvent
+
             yield (
-                BgtaskDoneEvent(task_id, message=task_info["msg"]),
+                event_cls(task_id, message=task_info["msg"]),
```
HyeockJinKim: Injecting a class is bad because it prevents the constructor from varying the arguments it receives.

src/ai/backend/common/events/bgtask.py
```diff
@@ -6,14 +7,16 @@
...
     UserBgtaskCancelledEvent,
     UserBgtaskDoneEvent,
     UserBgtaskFailedEvent,
     UserBgtaskUpdatedEvent,
 )
 from ai.backend.common.events.user_event.user_event import UserEvent
 from ai.backend.common.exception import UnreachableError
+from ai.backend.common.logging import BraceStyleAdapter
+
+log = BraceStyleAdapter(logging.getLogger(__spec__.name))  # type: ignore
```
HyeockJinKim: Why do you need type ignore??

src/ai/backend/common/exception.py
```diff
@@ -261,6 +261,17 @@ def default(cls) -> Self:
...
+    def with_detail(self, error_detail: ErrorDetail) -> "ErrorCode":
+        """
+        Returns a new error code with the specified error detail.
+        This is used when the error detail is not specified.
+        """
+        return ErrorCode(
+            domain=self.domain,
+            operation=self.operation,
+            error_detail=error_detail,
+        )
```
HyeockJinKim: The scenario where an error code generates another error code is too unnatural.

HyeockJinKim (approved):

PR #4274, opened 2025-04-25
fix(BA-1234): Check mount duplication by folder ID and subpath

rapsealk:
Follow-up to #4247 and resolves #NNN (BA-1234)

- [x] Test case(s) to:


fregataa (approved):

HyeockJinKim (approved):

rapsealk:
I will retouch the whole work, alongwith the changelog.

src/ai/backend/manager/models/vfolder.py
```diff
@@ -2776,3 +2776,10 @@ async def get_permission_ctx(
         # )
         # permission_ctx.apply_host_permission_ctx(host_permission_ctx)
     return permission_ctx
+
+
+def is_vfsubpath_duplicate(vfsubpath: PurePosixPath, vfmounts: Iterable[VFolderMount]) -> bool:
+    for mount in vfmounts:
+        if vfsubpath.is_relative_to(mount.vfsubpath):
```
rapsealk: Folder ID should be considered, as multiple folders with the same subpath may be incorrectly filtered, resulting in false positives.

HyeockJinKim (approved):

PR #4314, opened 2025-04-30
fix(BA-1276): Prevent invalid resource slot creation, and mutation

### LINKED ISSUE #4312: Prevent creating `ResourcePreset` without intrinsic resource
## Summary  

There is currently an issue where a ResourcePreset can be created without CPU or memory values.

Let's make these values required, and prevent creating `ResourcePreset` without intrinsic resource.

jopemachine:
resolves #4312 (BA-1276).

### DIFF
--- a/src/ai/backend/manager/services/resource_preset/service.py
+++ b/src/ai/backend/manager/services/resource_preset/service.py
@@ -81,7 +81,7 @@ class ResourcePresetService:
         name = action.creator.name
         creator = action.creator
 
-        if not creator.resource_slots.has_intrinsic_slot():
+        if not creator.resource_slots.has_intrinsic_slots():
             raise InvalidAPIParameters("ResourceSlot must have all intrinsic resource slots.")
 
         async def _create(db_session: AsyncSession) -> Optional[ResourcePresetRow]:
@@ -107,7 +107,7 @@ class ResourcePresetService:
             raise InvalidAPIParameters("One of (`id` or `name`) parameter should be not null")
 
         if resource_slots := modifier.resource_slots.optional_value():
-            if not resource_slots.has_intrinsic_slot():
+            if not resource_slots.has_intrinsic_slots():
                 raise InvalidAPIParameters("ResourceSlot must have all intrinsic resource slots.")
 
         async with self._db.begin_session() as db_sess:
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/services/resource_preset/service.py
```diff
@@ -81,13 +83,16 @@ async def create_preset(
...
         async def _create(db_session: AsyncSession) -> Optional[ResourcePresetRow]:
             return await ResourcePresetRow.create(creator, db_session=db_session)
 
         async with self._db.connect() as db_conn:
             preset_row = await execute_with_txn_retry(_create, self._db.begin_session, db_conn)
         if preset_row is None:
-            raise ValueError(
+            raise InvalidAPIParameters(
                 f"Duplicate resource preset name (name:{name}, scaling_group:{creator.scaling_group_name})"
             )
```
HyeockJinKim: It's hard to see this as an InvalidAPIParameter.

src/ai/backend/manager/services/resource_preset/service.py
```diff
@@ -101,7 +106,11 @@ async def modify_preset(
         modifier = action.modifier
 
         if preset_id is None and name is None:
-            raise ValueError("One of (`id` or `name`) parameter should be not null")
+            raise InvalidAPIParameters("One of (`id` or `name`) parameter should be not null")
+
+        if resource_slots := modifier.resource_slots.optional_value():
+            if not all(slot in resource_slots for slot in _INTRINSIC_RESOURCE_SLOTS):
+                raise InvalidAPIParameters("ResourceSlot must have all intrinsic resource slots.")
```
HyeockJinKim: Rather than checking this, it would be better to pass the value when the resource_slot is first created.

jopemachine: Would it be better to throw an exception in `to_creator`, and `to_modifier`? 🤔 
I think throwing an exception in these funcs might make it more bothersome to test than the current code.

HyeockJinKim: My intention was to put the validation logic in the dataclass definition.

src/ai/backend/manager/services/resource_preset/service.py
```diff
@@ -101,7 +104,11 @@ async def modify_preset(
         modifier = action.modifier
 
         if preset_id is None and name is None:
-            raise ValueError("One of (`id` or `name`) parameter should be not null")
+            raise InvalidAPIParameters("One of (`id` or `name`) parameter should be not null")
```
HyeockJinKim: `should not be`

HyeockJinKim (approved):

PR #4328, opened 2025-05-03
feat(BA-1292): Introduce LabelName enum

### LINKED ISSUE #4327: Replace Container label names from a hard-coded string to Enum
There are some hard-coded label names like `ai.backend.kernel-id` left on codebase.

Let’s replace them to `LabelName` Enum type

fregataa:
resolves #4327 (BA-1292)


src/ai/backend/agent/agent.py
```diff
@@ -270,7 +270,7 @@ def __init__(
         restarting: bool = False,
     ) -> None:
         self.image_labels = kernel_config["image"]["labels"]
-        self.kspec_version = int(self.image_labels.get("ai.backend.kernelspec", "1"))
+        self.kspec_version = int(self.image_labels.get(LabelName.KERNEL_SPEC.value, "1"))
```
HyeockJinKim: We already use LabelName as StrEnum. StrEnum is compatible with `str` even if you don't call value and just use the member value.

src/ai/backend/agent/docker/metadata/server.py
```diff
@@ -56,7 +57,11 @@ async def container_resolver_middleware(
         return web.Response(status=HTTPStatus.FORBIDDEN)
     async with closing_async(Docker()) as docker:
         containers = await docker.containers.list(
-            filters='{"label":["ai.backend.kernel-id"],"network":["bridge"],"status":["running"]}',
+            filters=(
+                '{"label":["'
+                + LabelName.KERNEL_ID.value
+                + '"],"network":["bridge"],"status":["running"]}'
```
HyeockJinKim: JSON value should be converted to JSON after creating the `dataclass` rather than making it as a JSON format string.

src/ai/backend/manager/registry.py
```diff
@@ -3622,7 +3622,7 @@ async def commit_session(
         session: SessionRow,
         new_image_ref: ImageRef,
         *,
-        extra_labels: dict[str, str] = {},
+        extra_labels: dict[str | LabelName, str] = {},
```
HyeockJinKim, line 3625: Can't you just take `LabelName` as a key?

HyeockJinKim (approved):

PR #4339, opened 2025-05-04
fix(BA-1302): Add TypeError handling in redis_helper

### LINKED ISSUE #4338: Handle TypeError in redis.py when client connection is closed.
## Impact

A TypeError may occur when the client connection is closed in the current version of [redis.py](http://redis.py). This impacts users by causing unexpected errors during operation.

## Expected behaviour

The system should handle the closure of the client connection without raising a TypeError.

## Actual behaviour

When the client connection is closed, a TypeError is raised in [redis.py](http://redis.py).

## Steps to reproduce

- Use the current version of [redis.py](http://redis.py).
- Close the client connection.
- Observe the TypeError that occurs.

## Environment

N/A

## Workarounds

N/A

## Other information

- A temporary handling for the TypeError needs to be added.

HyeockJinKim:
resolves #4338 (BA-1302)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/common/redis_helper.py
+++ b/src/ai/backend/common/redis_helper.py
@@ -189,6 +189,10 @@ async def execute(
     Note that when retried, the given function may be executed *multiple* times, so the caller
     should take care of side-effects of it.
     """
+
+    if not callable(func):
+        raise TypeError("The func must be a function or a coroutinefunction with no arguments.")
+
     redis_client = redis_obj.client
     service_name = service_name or redis_obj.service_name
     reconnect_poll_interval = redis_obj.redis_helper_config.get("reconnect_poll_timeout", 0.0)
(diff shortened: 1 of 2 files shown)


HyeockJinKim:
This is only a temporary measure and will require further action, such as upgrading redis.py. @achimnol

src/ai/backend/common/redis_helper.py
```diff
@@ -250,6 +249,7 @@ def show_retry_warning(e: Exception, warn_on_first_attempt: bool = True) -> None
             redis.exceptions.ReadOnlyError,
             redis.exceptions.ConnectionError,
             ConnectionResetError,
+            TypeError,  # The current version of redis.py raises a TypeError when the Connection is closed.
```
jopemachine: How about specifying the version clearly?

PR #4351, opened 2025-05-07
feat(BA-1263): Introduce Config Loaders, UnifiedConfig

### LINKED ISSUE #4303: Implement config loaders for each type
## Motivation  

- Let's add a config loader for each setting type.
  - Each config loader should have an interface for live reloading the corresponding settings.
  - Specifically, we will need watchers and tasks to monitor each configuration.

##

jopemachine:
resolves #4303 (BA-1263, BA-1264).


src/ai/backend/manager/services/vfolder/services/file.py
```diff
@@ -179,7 +185,9 @@ async def list_files(self, action: ListFilesAction) -> ListFilesActionResult:
         )
 
     async def rename_file(self, action: RenameFileAction) -> RenameFileActionResult:
-        allowed_vfolder_types = await self._shared_config.get_vfolder_types()
+        allowed_vfolder_types = (
+            await self._unified_config.legacy_etcd_config_loader.get_vfolder_types()
```
jopemachine, line 189: I know we should replace this, but I leave it as it is for now to minimize side effects.

PR #4354, opened 2025-05-08
feat(BA-1289): Allow anonymous users register TOTP key

### LINKED ISSUE #4324: Check TOTP status in manager `POST_AUTHORIZE` hook
To check the user's TOTP key registration status and enforce 2FA settings, you should implement this in the `POST_AUTHORIZE` hook handler of the manager. If a TOTP key is not registered and the `force-2FA` option is activated, the handler should return the token value in a bad request response.

(Note: It is challenging to allow the hook handler to return a redirect response (3xx) instead of a 400 error.)

fregataa:
resolves #4324 (BA-1289)

### 1. First login trial
Call `POST <WEBSERVER_ENDPOINT>/server/login`

Response JSON data is like below
```json
{
    "authenticated": false,
    "data": {
        "type": "https://api.backend.ai/probs/require-totp-registration",
        "title": "Two-Factor Authentication registration required.",
        "details": "You must register Two-Factor Authentication.",
        "two_factor_registration_token": "eyJh..."
    }
}
```

### 2. TOTP activate
Call `POST <WEBSERVER_ENDPOINT>/func/totp/anon` API with JSON data below
```json
{
    "registration_token": "eyJh..."
}
```

Response JSON data is like below
```json
{
    "totp_key": "5ES3...",
    "totp_uri": "..."
}
```

### 3. After Register TOTP key - Login retry
Call `POST <WEBSERVER_ENDPOINT>/server/login`

Response JSON data is like below
```json
{
    "authenticated": false,
    "data": {
        "type": "https://api.backend.ai/probs/require-totp-authentication",
        "title": "Two-Factor Authentication needed.",
        "details": "You must authenticate using Two-Factor Authentication."
    }
}
```


src/ai/backend/common/dto/manager/auth/field.py
```diff
@@ -0,0 +1,40 @@
...
+    def to_dict(self) -> dict[str, str]:
+        return self.model_dump(mode="json")
+
+
+class AuthResponse(BaseModel):
+    data: AuthSuccessResponseData | RequireTOTPRegistrationResponseData
+    type: AuthTokenTypes = AuthTokenTypes.KEYPAIR
+
+    @classmethod
+    def from_auth_response(cls, response: dict[str, Any]) -> "AuthResponse":
```
HyeockJinKim: Use `Self` in typing.

src/ai/backend/web/server.py
```diff
@@ -527,20 +543,32 @@ async def token_login_handler(request: web.Request) -> web.Response:
...
             public_return = {
-                "access_key": token.content["access_key"],
-                "role": token.content["role"],
-                "status": token.content.get("status"),
+                "access_key": token.access_key,
+                "role": token.role,
+                "status": token.status,
             }
             session["authenticated"] = True
             session["token"] = stored_token  # store full token
```
HyeockJinKim: The logic from the `stored_token` is the same as when we received the `AuthSuccessResponse`, which works fine for now, but the block separation doesn't look as clean.

fregataa: hmm, then how about raising an error and catch it?
Like this block https://github.com/lablup/backend.ai/blob/cfaec702deda14cb1dd77f42040797abcac09f21/src/ai/backend/web/server.py#L425-L435

ironAiken2 (approved):
LGTM

src/ai/backend/common/dto/manager/auth/field.py
```diff
@@ -0,0 +1,74 @@
...
+class RequireTwoFactorAuthResponse(AuthResponse):
+    type: TwoFactorType
+
+    def to_dict(self) -> dict[str, str]:
+        return self.model_dump(mode="json")
+
+
+AuthAuthResponseType = (
+    AuthSuccessResponse | RequireTwoFactorRegistrationResponse | RequireTwoFactorAuthResponse
+)
```
HyeockJinKim: This type doesn't seem necessary.

HyeockJinKim (approved):

HyeockJinKim (approved):

PR #4365, opened 2025-05-11
fix(BA-1331): Add `validation_alias` and `serialization_alias` to ManagerSharedConfig configs

### LINKED ISSUE #4367: Refactor the `alias` of `SharedConfig` into `validation_alias` and `serialization_alias`
Unfortunately, the naming conventions of the configuration fields in SharedConfig are not consistent.

In the Pydantic migration, many fields are aliased to maintain compatibility with existing configurations.

These aliases inconsistently use "_" and "-", and in some cases, there are no aliases at all, making mistakes hard to detect.

To make the aliases clearer, let's refactor them by using `validation_alias` and `serialization_alias` instead of plain aliases, and ensure they function correctly.

jopemachine:
resolves #4367 (BA-1331).

### DIFF
--- a/changes/4365.enhance.md
+++ b/changes/4365.enhance.md
@@ -1,0 +1,1 @@
+Refactor the `alias` of `ManagerSharedConfig` into `validation_alias` and `serialization_alias`.
--- a/src/ai/backend/manager/config/unified.py
+++ b/src/ai/backend/manager/config/unified.py
@@ -1266,7 +1266,8 @@ class SingleRedisConfig(BaseModel):
         Identifies which service to monitor for failover.
         """,
         examples=[None, "mymaster", "backend-ai"],
-        alias="service-name",
+        validation_alias=AliasChoices("service_name", "service-name"),
+        serialization_alias="service-name",
     )
     password: Optional[str] = Field(
         default=None,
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -116,7 +116,7 @@ disable_error_code = ["typeddict-unknown-key"]
 [tool.pydantic-mypy]
 init_forbid_extra = true
 init_typed = true
-warn_required_dynamic_aliases = true
+warn_required_dynamic_aliases = false
```
fregataa: Could you leave a comment of a reason why it changes?

fregataa (commented):
I checked there are some of serialization aliases with kebab cases and others with snake cases. Have we decided to use which cases we will use?

pyproject.toml
HyeockJinKim: It' config for using `AliasChoices` to accept both `-` and `_` concatenated names for backwards compatibility with existing configs.

HyeockJinKim (approved):

jopemachine:
> I checked there are some of serialization aliases with kebab cases and others with snake cases. Have we decided to use which cases we will use?

We can use both snake case and kebab case.
The recommended approach is to use the name specified as the `serialization_alias` (the name of the existing setting) for existing configs and to standardize on snake case for newly added configs. 

After this sprint, related announcements and documentation will be necessary. @HyeockJinKim

HyeockJinKim:
please rebase @jopemachine

jopemachine:
> please rebase @jopemachine

Resolved

PR #4386, opened 2025-05-13
feat(BA-1211): Add kernel last seen event and handler

### LINKED ISSUE #4240: Add Logic for Manager to Check 'Last Seen' Status
As a developer, I want to add logic for the manager to check the 'last seen' status of kernels, so that the system can trigger recovery actions if necessary.

fregataa:
resolves #4240 (BA-1211)


src/ai/backend/common/events/kernel.py
```diff
@@ -229,3 +229,24 @@ def user_event(self) -> Optional[UserEvent]:
     @classmethod
     def event_name(cls) -> str:
         return "do_sync_kernel_logs"
+
+
+@dataclass
+class KernelHeartbeatEvent(BaseKernelEvent):
+    timestamp: float
```
fregataa: 1. Remove timestamp here and set timestamp from event receivers
2. Add `container_id`

HyeockJinKim (approved):

PR #4392, opened 2025-05-14
fix(BA-1353): wrong location of event handling observer

### LINKED ISSUE #4391: Wrong location of event handling observer
metric observers report success/failure handling after creating a handler task which runs asynchronously

fregataa:
resolves #4391 (BA-1353)

### DIFF
--- a/src/ai/backend/common/events/dispatcher.py
+++ b/src/ai/backend/common/events/dispatcher.py
@@ -7,8 +7,10 @@ import secrets
 import time
 from abc import ABC, abstractmethod
 from collections import defaultdict
+from collections.abc import Sequence
 from typing import (
     Any,
+    Awaitable,
     Callable,
     Coroutine,
     Generic,
@@ -45,6 +47,11 @@ __all__ = (
 )
 
 log = BraceStyleAdapter(logging.getLogger(__spec__.name))
+
+
+class EventHandlerType(enum.StrEnum):
+    CONSUMER = enum.auto()
+    SUBSCRIBER = enum.auto()
 
 
 class EventDomain(enum.StrEnum):
@@ -127,6 +134,7 @@ class EventHandler(Generic[TContext, TEvent]):
     name: str
     context: TContext
     callback: EventCallback[TContext, TEvent]
+    handler_type: EventHandlerType
     coalescing_opts: Optional[CoalescingOptions]
     coalescing_state: CoalescingState
     args_matcher: Callable[[tuple], bool] | None
@@ -307,6 +315,7 @@ class EventDispatcher:
             name,
             context,
             callback,
+            EventHandlerType.CONSUMER,
             coalescing_opts,
             CoalescingState(),
             args_matcher,
@@ -348,6 +357,7 @@ class EventDispatcher:
             name,
             context,
             callback,
+            EventHandlerType.SUBSCRIBER,
             coalescing_opts,
             CoalescingState(),
             args_matcher,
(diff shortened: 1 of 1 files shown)


src/ai/backend/common/events/dispatcher.py
```diff
@@ -367,37 +377,69 @@ def unsubscribe(
...
+                    duration=time.perf_counter() - start,
+                )
+        except Exception as e:
+            self._metric_observer.observe_event_failure(
+                event_type=event_type,
+                duration=time.perf_counter() - start,
+                exception=e,
+            )
+            log.exception(f"EventDispatcher.{evh_type}(): unexpected-error, {repr(e)}")
+            raise
```
HyeockJinKim, line 451: Is it okay to raise it here?

src/ai/backend/common/events/dispatcher.py
```diff
@@ -420,71 +462,37 @@ async def _consume_loop(self) -> None:
         async for msg in self._msg_queue.consume_queue():  # type: ignore
             if self._closed:
                 return
-            event_type = "unknown"
-            start = time.perf_counter()
+            decoded_event_name = msg.payload[b"name"].decode()
+
+            async def done() -> None:
+                await self._msg_queue.done(msg.msg_id)
```
HyeockJinKim: Can you guarantee that it will only run once?

src/ai/backend/common/events/dispatcher.py
```diff
@@ -47,6 +49,11 @@
 log = BraceStyleAdapter(logging.getLogger(__spec__.name))
 
 
+class EventHandlerType(enum.StrEnum):
+    CONSUMER = enum.auto()
+    SUBSCRIBER = enum.auto()
```
HyeockJinKim: Make it private

HyeockJinKim: don't use enum.auto().

src/ai/backend/common/events/dispatcher.py
```diff
@@ -420,71 +469,45 @@ async def _consume_loop(self) -> None:
...
-            except BaseException as e:
-                self._metric_observer.observe_event_failure(
-                    event_type=event_type,
-                    duration=time.perf_counter() - start,
-                    exception=e,
-                )
-                raise
+            except Exception:
+                # Exception is already handled in handle() method.
+                pass
```
HyeockJinKim: Why add except block if you're just going to pass?

src/ai/backend/common/events/dispatcher.py
```diff
@@ -420,71 +469,45 @@ async def _consume_loop(self) -> None:
...
-            start = time.perf_counter()
+            decoded_event_name = msg.payload[b"name"].decode()
+
+            async def done() -> None:
+                # To ensure that all consumer handlers are called.
+                # Basically there should be only one consumer handler for one event.
+                handler_count = len(self._subscribers[decoded_event_name])
+                async with self._msg_id_consumed_count_lock:
+                    self._msg_id_consumed_count[msg.msg_id] += 1
+                    if self._msg_id_consumed_count[msg.msg_id] < handler_count:
```
HyeockJinKim: Since we're only doing synchronous logic, there's no need to lock it, right?

src/ai/backend/common/events/dispatcher.py
fregataa: I added this part because all execptions are handled in the `EventDispatcher._handle()` method. Updated to remove this exception skipping logic.

src/ai/backend/common/events/dispatcher.py
fregataa: It is re-raised because of this [test codes](https://github.com/lablup/backend.ai/blob/main/tests/common/test_events.py#L114).
It requires to catch all exceptions raised in the `EventDispatcher.handle()` method.

src/ai/backend/common/events/dispatcher.py
fregataa: the consumer loop creates async tasks so I think we need async lock

HyeockJinKim (approved):

PR #4459, opened 2025-05-21
fix(BA-1404): Calculate correct permissions for admins

### LINKED ISSUE #4458: Wrong permission calculation when admins query vfolders in Project scopes
`vfolder_nodes` GQL query has project_id parameter and the permission calculator malfunctions, meanwhile `vfolder_node` GQL query does not have scope parameter and it calculates permissions correctly

fregataa:
resolves #4458 (BA-1404)

### DIFF
--- a/src/ai/backend/manager/models/vfolder.py
+++ b/src/ai/backend/manager/models/vfolder.py
@@ -94,7 +94,6 @@ from .rbac import (
     AbstractPermissionContext,
     AbstractPermissionContextBuilder,
     DomainScope,
-    PredefinedRole,
     ProjectScope,
     ScopeType,
     StorageHost,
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/models/vfolder.py
```diff
@@ -2558,9 +2559,21 @@ async def build_ctx_in_domain_scope(
...
         permission_ctx = await self._build_at_project_scope_non_recursively(ctx, scope.project_id)
-        _user_perm_ctx = await self._build_at_user_scope_non_recursively(ctx, ctx.user_id)
-        permission_ctx.merge(_user_perm_ctx)
+        project_query = sa.select(AssocGroupUserRow).where(
+            AssocGroupUserRow.group_id == scope.project_id
+        )
+        user_perm_ctx = await self._build_at_user_scope_non_recursively(ctx, ctx.user_id)
+        permission_ctx.merge(user_perm_ctx)
+        roles = await get_predefined_roles_in_scope(ctx, scope, self.db_session)
+        if PredefinedRole.ADMIN in roles:
```
HyeockJinKim: Is there a reason to do the project query first?
Shouldn't you query the value just before using it?

HyeockJinKim (approved):

nowgnuesLee:
If that branch is merged, there are few things that need to be fixed in the existing webui screens. Previously, user folders were not visible in the list, but after this branch they are. I think the sharing modal needs to be fixed as well, and there's a bug when the admin accesses the user folder (which I need to figure out the cause of).

nowgnuesLee (approved):
LGTM

PR #4450, opened 2025-05-21
fix(BA-1379): Truncate generated session name in service creation to max allowed length

rapsealk:
resolves #4416 (BA-1379)

This pull request addresses an issue with the `Service.create()` method in the client SDK by ensuring that the default generated session name is truncated to comply with a maximum allowed length. It introduces a new constant for the session name length and updates relevant methods to enforce this constraint.

### Fixes and Improvements in Session Name Handling:

* **Truncate Default Generated Session Names**: Updated the `Service.create()` and `Service.try_start()` methods to truncate the default generated session name to the maximum allowed length using the new `SESSION_NAME_MAX_LENGTH` constant. (`src/ai/backend/client/func/service.py`, [[1]](diffhunk://#diff-5584a4e4cc558f070b181820354fe5ee11342eb8ea510815196561e11b2dba96L149-R150) [[2]](diffhunk://#diff-5584a4e4cc558f070b181820354fe5ee11342eb8ea510815196561e11b2dba96L272-R273)
* **New Constant for Session Name Length**: Added `SESSION_NAME_MAX_LENGTH` as a `Final[int]` constant with a value of 24, and updated the `SESSION_NAME_MATCHER` regex to use this constant. (`src/ai/backend/common/typed_validators.py`, [src/ai/backend/common/typed_validators.pyL170-R173](diffhunk://#diff-b2b56cb63b3568fdcbbcf5d41420f96228c379fbf098dee7c5cb8a2e26900a69L170-R173))

### Codebase Enhancements:

* **Improved Imports**: Added `SESSION_NAME_MAX_LENGTH` import to `service.py` and cleaned up import statements to include `Final` in `typed_validators.py`. (`src/ai/backend/client/func/service.py`, [[1]](diffhunk://#diff-5584a4e4cc558f070b181820354fe5ee11342eb8ea510815196561e11b2dba96R9); `src/ai/backend/common/typed_validators.py`, [[2]](diffhunk://#diff-b2b56cb63b3568fdcbbcf5d41420f96228c379fbf098dee7c5cb8a2e26900a69L6-R9)

### Documentation:

* **Changelog Update**: Documented the fix for session name truncation in the changelog file. (`changes/.fix.md`, [changes/.fix.mdR1](diffhunk://#diff-71043af5cf5bbdf08bac67d3f3011521153776c3bc7440557b0755ef4a10bfb1R1))

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version
- [x] Mention to the original issue
[...]

PR #4472, opened 2025-05-22
fix(BA-1379): Truncate generated session name in service creation to max allowed length (#4450)

lablup-octodog:
This is an auto-generated backport PR of #4450 to the 25.6 release.


HyeockJinKim:
Can you check your CI? @rapsealk

PR #4507, opened 2025-05-26
feat(BA-1443): `backend.ai mgr scheduler last-execution-time` command

### LINKED ISSUE #4506: Make manager CLI command to check each scheduler's last execution time
## Motivation  

- We need a CLI tool to check each scheduler’s last execution time

## Required Features

- Fetch each scheduler’s (`schedule`, `scale_services`, …) last execution information from `manager.{manager_id}.{scheduler_name`} hash map at Redis' `LIVE` database.
- Display the result on manager CLI (`backend.ai mgr`).

## Impact  

- System administrators can determine whether the scheduler is stopped working or not

## Testing Scenarios  

-

kyujin-cho:
resolves #4506 (BA-1443). This PR adds new command set to manager CLI implementation (not `client-py`).    
The command accepts `scheduler_name`, which refers target scheduler name to look up, and `--manager-id, which defines which manager node to fetch information of, being set as the local manager executing the command by default. When not specifying `scheduler_name` argument, the command will return status of every schedulers operated from the target manager node. On a complicated cluster with multiple active Backend.AI managers operating concurrently, Backend.AI keeps track of each manager node's last execution time of each scheduler locally, not on a global redis key namespace, hence requiring CLI users to fetch status of each manager individually.


src/ai/backend/manager/cli/scheduler.py
```diff
@@ -0,0 +1,140 @@
...
+@click.argument("scheduler_name", required=False)
+@click.pass_obj
+def last_execution_time(
+    cli_ctx: CLIContext,
+    scheduler_name: str | None,
+    *,
+    format: Literal["plain"] | Literal["csv"] | Literal["json"] = "plain",
+    manager_id: str | None = None,
+) -> None:
+    """Queries manager's scheduler execution footprint from Redis. When scheduler name is not specified, this command will return informations of all schedulers in store."""
```
HyeockJinKim, line 57: It seems like last_execution_time returns more information than the name.
Would footprint be a better name?

kyujin-cho: "footprint" looks good, perhaps we can count more generic  term like "information".

PR #4535, opened 2025-05-28
refactor(BA-1421): Apply service layer in auth apis

### LINKED ISSUE #4486: Apply audit logs for user login APIs
Apply audit log for user login and signup, signout, etc.

seedspirit:
resolves #4486 (BA-1421)

### DIFF
--- a/src/ai/backend/common/dto/manager/auth/field.py
+++ b/src/ai/backend/common/dto/manager/auth/field.py
@@ -10,13 +10,6 @@ class AuthTokenType(enum.StrEnum):
     KEYPAIR = "keypair"
     JWT = "jwt"
 
-    @classmethod
-    def from_str(cls, value: str) -> AuthTokenType:
-        try:
-            return cls(value)
-        except ValueError:
-            raise ValueError(f"Invalid AuthTokenType: {value}") from None
-
 
 class AuthResponseType(enum.StrEnum):
     SUCCESS = "success"
--- a/src/ai/backend/manager/api/auth.py
+++ b/src/ai/backend/manager/api/auth.py
@@ -663,7 +663,7 @@ async def authorize(request: web.Request, params: Any) -> web.StreamResponse:
     root_ctx: RootContext = request.app["_root.context"]
     action = AuthorizeAction(
         request=request,
-        type=AuthTokenType.from_str(params["type"]),
+        type=AuthTokenType(params["type"]),
         domain_name=params["domain"],
         email=params["username"],
         password=params["password"],


src/ai/backend/manager/models/user.py
```diff
@@ -333,42 +332,3 @@ def by_user_email(
...
-        return None
-    if row["password"] is None:
-        # user password is not set.
-        return None
-    try:
-        if _verify_password(password, row["password"]):
-            return row
-    except ValueError:
-        return None
-    return None
```
HyeockJinKim: Why did you move to the service layer?

src/ai/backend/manager/services/auth/actions/get_ssh_keypair.py
```diff
@@ -0,0 +1,28 @@
...
+from ai.backend.manager.services.auth.actions.base import AuthAction
+
+
+@dataclass
+class GetSSHKeypairAction(AuthAction):
+    access_key: str
+
+    @override
+    def entity_id(self) -> Optional[str]:
+        return None
```
HyeockJinKim: `return self.access_key`

src/ai/backend/manager/models/user.py
seedspirit: It only used in auth apis. Should I leave it as utility function?

HyeockJinKim: It’s not about where it’s used. When it comes to package separation, what’s important is the domain it belongs to.
Even if the current implementation isn’t ideal, moving it to the service layer would be a step in the wrong direction.

src/ai/backend/manager/services/auth/service.py
```diff
@@ -0,0 +1,616 @@
...
+
+        return AuthorizeActionResult(
+            stream_response=None,
+            authorization_result=AuthorizationResult(
+                access_key=main_keypair_row.access_key,
+                secret_key=main_keypair_row.secret_key,
+                role=user["role"],
+                status=user["status"],
+            ),
+        )
```
seedspirit: Is this structure okay? As we don't know what `hook_result.result` will return, I just included `web.StreamResponse` in `AuthorizeActionResult` @HyeockJinKim  @fregataa

HyeockJinKim: Web Response should not be used within internal layers.

src/ai/backend/manager/services/auth/actions/signout.py
```diff
@@ -0,0 +1,31 @@
...
+@dataclass
+class SignoutAction(AuthAction):
+    domain_name: str
+    requester_email: str
+    email: str
+    password: str
+
+    @override
+    def entity_id(self) -> Optional[str]:
+        return None
```
HyeockJinKim: requester_email

src/ai/backend/manager/services/auth/actions/signup.py
```diff
@@ -0,0 +1,52 @@
...
+    domain: str
+    email: str
+    password: str
+    username: Optional[str]
+    full_name: Optional[str]
+    description: Optional[str]
+
+    @override
+    def entity_id(self) -> Optional[str]:
+        return None
```
HyeockJinKim, line 23: email

src/ai/backend/manager/models/user.py
seedspirit: I see. I will rollback change. In between, I think it is better to move `_verify_password` or `check_credential` from `models` in the long term view

src/ai/backend/common/dto/manager/auth/field.py
```diff
@@ -10,6 +10,13 @@ class AuthTokenTypes(enum.Enum):
     KEYPAIR = "keypair"
     JWT = "jwt"
 
+    @classmethod
+    def from_str(cls, value: str) -> AuthTokenTypes:
+        try:
+            return cls(value)
+        except ValueError:
+            raise ValueError(f"Invalid AuthTokenType: {value}") from None
```
HyeockJinKim: Why not use `StrEnum`?

src/ai/backend/manager/api/auth.py
```diff
@@ -697,68 +658,32 @@ async def get_role(request: web.Request, params: Any) -> web.Response:
...
         response_type=AuthResponseType.SUCCESS,
-        access_key=main_keypair_row.access_key,
-        secret_key=main_keypair_row.secret_key,
-        role=user["role"],
-        status=user["status"],
+        access_key=auth_result.access_key,
+        secret_key=auth_result.secret_key,
+        role=auth_result.role,
+        status=auth_result.status,
     )
```
HyeockJinKim, line 685: This part of the behavior has changed.

PR #4541, opened 2025-05-29
fix: Broken `ListResourcePresetsAction`

jopemachine:
resolves #NNN (BA-MMM)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/services/resource_preset/service.py
+++ b/src/ai/backend/manager/services/resource_preset/service.py
@@ -182,6 +182,7 @@ class ResourcePresetService:
                 )
             query = query.where(query_condition)
             presets = []
+            await self._config_provider.legacy_etcd_config_loader.get_resource_slots()
             async for row in await db_session.stream_scalars(query):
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/services/resource_preset/service.py
```diff
@@ -182,6 +182,7 @@ async def list_presets(self, action: ListResourcePresetsAction) -> ListResourceP
                 )
             query = query.where(query_condition)
             presets = []
+            await self._config_provider.legacy_etcd_config_loader.get_resource_slots()
```
jopemachine, line 185: This function relies on side effects, making its behavior invisible from the outside.
(It is passed via contextvar.)
However, since this part was originally written this way in the existing code, I left it as is for now.

HyeockJinKim: Ah… this kind of code shouldn’t exist.
Fundamentally, there should be no code that accesses and updates a contextvar from another file.

PR #4554, opened 2025-05-29
fix(BA-1459): Broken `list_presets` API, and SDK (#4541)

lablup-octodog:
This is an auto-generated backport PR of #4541 to the 25.6 release.


jopemachine:
get_resource_slots is already called in 25.6, so I wil close this.

PR #4558, opened 2025-05-30
fix(BA-1465): Fix broken `Network` SDK implementations

### LINKED ISSUE #4528: Fix `Network.create` SDK implementation to successfully create network
## Summary  

- `Network.create` dose not create network though proper params are set

## Steps to Reproduce  

```python
from ai.backend.client.session import Session

def test():
    with Session() as session:
        result = (session
                  .Network.create(
                      project_id="2de2b969-1d04-48a6-af16-0bc8adb3c831",
                      name="test_network",
                      driver="bridge"
                  )
        )
        print(result)


if __name__ == "__main__":
    test()
```

## Expected Behavior  

- Network created and returned
- Information of created network persisted in DB row

## Actual Behavior    

- `None` returned
- No DB row are inserted

seedspirit:
resolves #4528 (BA-1465)


src/ai/backend/manager/models/network.py
```diff
@@ -359,14 +359,6 @@ async def mutate(
...
             raise
 
-        network_plugin = graph_ctx.network_plugin_ctx.plugins[_driver]
-        try:
-            network_info = await network_plugin.create_network()
-            network_name = network_info.network_id
-        except Exception:
-            log.exception(f"Failed to create the inter-container network (plugin: {_driver})")
-            raise
-
```
seedspirit, line 369: Removed this code block because it is duplicated

src/ai/backend/client/func/network.py
```diff
@@ -63,19 +64,27 @@ async def create(
...
                 "project_id": project_id,
                 "driver": driver,
             },
         )
-        return cls(network_id=UUID(data["create_network"]["network"]["row_id"]))
+
+        if not result["create_network"]["ok"]:
+            raise RuntimeError(
+                f"Failed to create network '{name}': {result['create_network']['msg']}"
+            )
```
HyeockJinKim, line 85: Have you checked the failure response?

HyeockJinKim (approved):

PR #4573, opened 2025-06-01
fix(BA-1479): Incorrect resource label loading

### LINKED ISSUE #4566: image required resources sync issue
## Impact

There are cases where session creation fails. This happens when the minimum resource requirements for an image and the label values do not match.

## Expected behaviour

The system should correctly create a session by using the larger value from the two provided values as a property.

## Actual behaviour

Currently, session creation is not possible when the minimum resource requirements and label values differ.

## Steps to reproduce

- N/A

## Environment

- N/A

## Workarounds

- N/A

## Other information

It is necessary to modify the system to pass the larger value of the two as a property.

jopemachine:
resolves #4566 (BA-1479)

### DIFF
--- a/src/ai/backend/manager/models/image.py
+++ b/src/ai/backend/manager/models/image.py
@@ -672,15 +672,22 @@ class ImageRow(Base):
 
     @property
     def resources(self) -> Resources:
-        custom_resources = self._resources or {}
-        label_resources = self.get_resources_from_labels()  # Set this as default
+        merged = dict(self._resources or {})
 
-        resources = dict(custom_resources)
-        for label_key, value in label_resources.items():
-            if label_key not in resources:
-                resources[label_key] = value
+        for label_key, label_spec in self.get_resources_from_labels().items():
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/models/image.py
```diff
@@ -673,14 +673,30 @@ def __repr__(self) -> str:
...
+
+        for label_key in {*custom_resources.keys(), *label_resources.keys()}:
+            custom_spec = custom_resources.get(label_key, {})
+            label_spec = label_resources.get(label_key, {})
+
+            merged_spec = dict(custom_spec)
+            for label_spec_key, value in label_spec.items():
+                merged_spec.setdefault(label_spec_key, value)
+
+            # TODO: Consider other slot types
```
jopemachine, line 688: In principle, we should consider other slots as well, but this PR is closer to a hotfix.
So, let's proceed with this PR for now, and work on improving the structure in a more appropriate direction in a follow-up PR.

HyeockJinKim (approved):

PR #4591, opened 2025-06-03
fix: Add explicit legacy format converting for config types

jopemachine:
resolves #NNN (BA-MMM)

WIP

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/common/config.py
+++ b/src/ai/backend/common/config.py
@@ -44,6 +44,23 @@ class BaseSchema(BaseModel):
     )
 
 
+class BaseConfigModel(BaseModel):
+    @staticmethod
+    def snake_to_kebab_case(string: str) -> str:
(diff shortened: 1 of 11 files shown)


src/ai/backend/common/config.py
```diff
@@ -44,6 +44,23 @@ class BaseSchema(BaseModel):
...
+    model_config = ConfigDict(
+        populate_by_name=True,
+        from_attributes=True,
+        use_enum_values=True,
+        extra="allow",
+        alias_generator=snake_to_kebab_case,
+    )
+
+    def to_legacy_format(self) -> Mapping[str, Any]:
+        return self.model_dump()
```
HyeockJinKim, line 61: Can you guarantee that this is the same as the value you received as input?
I'm not sure if it's easy to change it due to plugins.

PR #4594, opened 2025-06-04
fix: Include endpoint loading in route retrieval for delete_route function

HyeockJinKim:
Follow-up of #4590 (BA-1498)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/api/service.py
+++ b/src/ai/backend/manager/api/service.py
@@ -990,7 +990,7 @@ async def delete_route(request: web.Request) -> SuccessResponseModel:
     )
     async with root_ctx.db.begin_readonly_session() as db_sess:
         try:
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/api/service.py
```diff
@@ -990,7 +990,7 @@ async def delete_route(request: web.Request) -> SuccessResponseModel:
     )
     async with root_ctx.db.begin_readonly_session() as db_sess:
         try:
-            route = await RoutingRow.get(db_sess, route_id, load_session=True)
+            route = await RoutingRow.get(db_sess, route_id, load_session=True, load_endpoint=True)
```
rapsealk:
```suggestion
            route = await RoutingRow.get(db_sess, route_id, load_endpoint=True, load_session=True)
```

PR #4612, opened 2025-06-05
fix(BA-1511): `rpc_auth_manager_keypair` not found error in `generate-rpc-keypair`

### LINKED ISSUE #4611: Cannot `generate-rpc-keypair` when no file exists at the `rpc_auth_manager_keypair` path
## Summary  

An error occurs when executing `generate-rpc-keypair` if no file exists at the `rpc_auth_manager_keypair` path—this only happens when a custom path is specified in `manager.toml`, not when the default path is used.

jopemachine:
Resolves #4611 (BA-1511)

### DIFF
--- a/src/ai/backend/manager/config/unified.py
+++ b/src/ai/backend/manager/config/unified.py
@@ -488,7 +488,7 @@ class ManagerConfig(BaseModel):
         validation_alias=AliasChoices("internal-addr", "internal_addr"),
         serialization_alias="internal-addr",
     )
-    rpc_auth_manager_keypair: FilePath = Field(
+    rpc_auth_manager_keypair: Path = Field(
         default=Path("fixtures/manager/manager.key_secret"),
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/config/unified.py
```diff
@@ -839,6 +839,16 @@ def aiomonitor_terminal_ui_port(self) -> int:
...
 
+    @field_validator("rpc_auth_manager_keypair", mode="before")
+    @classmethod
+    def _parse_rpc_auth_manager_keypair(cls, v: str) -> str:
+        if not Path(v).exists():
+            log.warning(
+                f'RPC authentication keypair file does not exist: "{v}".',
+            )
+            return v
+        return v
```
HyeockJinKim: It doesn't seem necessary to return from the if block. @jopemachine

PR #4614, opened 2025-06-05
feat(BA-1508): Implement test specification management and execution framework

### LINKED ISSUE #4608: Make a skeleton code for tester
- BEP : [https://github.com/lablup/beps/pull/5](https://github.com/lablup/beps/pull/5) 
- We need to add a basic test structure for writing test scenarios in the current tester.

JIRA Issue: BA-1508

HyeockJinKim:
resolves #4608 (BA-1508)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/test/testcases/template.py
+++ b/src/ai/backend/test/testcases/template.py
@@ -1,0 +1,123 @@
+import asyncio
+from abc import ABC, abstractmethod
+
+from ai.backend.test.tester.exporter import TestExporter
+
+
+class TestCode(ABC):
+    @abstractmethod
+    async def test(self) -> None:
+        """
+        Run the test case. This method should be overridden by subclasses
+        to implement the specific test logic.
+        """
+        raise NotImplementedError("Subclasses must implement this method.")
+
+
+class NopTestCode(TestCode):
+    async def test(self) -> None:
+        """
+        A no-operation test code that does nothing.
+        This can be used as a placeholder for tests that do not require any action.
+        """
+        await asyncio.sleep(0)  # Simulate a no-op with an async sleep
+        return
+
+
+class TestTemplate(ABC):
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        """
+        Get the name of the test template.
+        This method should be overridden by subclasses to provide a specific name.
+        """
+        raise NotImplementedError("Subclasses must implement this method.")
+
+    @abstractmethod
+    async def run_test(self, exporter: TestExporter) -> None:
+        """
+        Run the test case. This method should be overridden by subclasses
+        to implement the specific test logic.
+        """
+        raise NotImplementedError("Subclasses must implement this method.")
+
+
+class BasicTestTemplate(TestTemplate):
+    _testcode: TestCode
+
+    def __init__(self, testcode: TestCode) -> None:
+        """
+        Initialize the basic template with a test code function.
+
+        :param testcode: The test code function to run.
+        """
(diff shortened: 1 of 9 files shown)


src/ai/backend/test/tester/runner.py
```diff
@@ -12,7 +12,7 @@ def __init__(self, spec: TestSpec, exporter: TestExporter) -> None:
 
     async def run(self) -> None:
         try:
-            await self._spec.run_test()
-            await self._exporter.export_done(self._spec)
+            await self._spec.template.run_test(self._exporter)
+            await self._exporter.export_done(self._spec.name)
         except BaseException as e:
-            await self._exporter.export_exception(self._spec, e)
+            await self._exporter.export_exception(self._spec.name, e)
```
jopemachine, line 18: Although CancelledError is caught by BaseException, it is captured as an empty string.
So I think it would be better to catch it as a separate exception and report an appropriate error message.

src/ai/backend/test/testcases/testcases.py
```diff
@@ -0,0 +1,76 @@
...
+
+class TestTag(enum.StrEnum):
+    # component based tags
+    MANAGER = "manager"
+    AGENT = "agent"
+
+    # Domain specific tags
+    VFOLDER = "vfolder"
+    IMAGE = "image"
+    SESSION = "session"
```
jopemachine, line 20: How about spliting `TestTag` into `ComponentTag` and `DomainTag`?
I think it would be good to make it mandatory to include at least one component tag and at least one domain tag.

HyeockJinKim: Wouldn’t managing the test code become too cumbersome?

src/ai/backend/test/tester/runner.py
HyeockJinKim: I’m planning to delegate how exceptions are handled and displayed to the **exporter**.
The specific way they are presented should be considered later within the exporter.

src/ai/backend/test/testcases/testcases.py
jopemachine: I think it's not such a hassle to add one component tag and at least one domain tag.
And even if tags are managed as a single type, I think it would still be desirable to attach one component tag and at least one domain tag.

HyeockJinKim: In actual usage, it’ll be used in a form like `run-test -t {tag}`, and since there won’t be many other use cases, I’m not sure if strict categorization is really necessary.

src/ai/backend/test/testcases/template.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class TestCode(ABC):
+    @abstractmethod
+    async def test(self) -> None:
+        """
+        Run the test case. This method should be overridden by subclasses
+        to implement the specific test logic.
+        """
+        raise NotImplementedError("Subclasses must implement this method.")
```
seedspirit, line 14: I noticed that currently TestCode, TestRunner, etc. are all written in asyncio. For the test code, is it necessary to write test code with `AsyncSession`, etc. (and if so, is it not necessary to enforce it structurally or give instruction in comment?)?

HyeockJinKim: Although there is no reason to only use `AsyncSession` just because the interface is an async function.
It's better to create a wrapper for components (agent, manager, ...) in the client sdk and request it separately if possible.

PR #4620, opened 2025-06-05
refactor(BA-1519): Differentiate consume and subscribe events

### LINKED ISSUE #4618: Separate events to be consumed and subscribed
We did not differentiate between events that should be consumed and those that should be subscribed to. Let's clarify this.

JIRA Issue: BA-1519

fregataa:
resolves #4618 (BA-1519)

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/common/events/dispatcher.py
```diff
@@ -632,7 +632,25 @@ async def close(self) -> None:
...
+        source_bytes = self._source_bytes
+        if source_override is not None:
+            source_bytes = source_override.encode()
+
+        raw_event = {
+            b"name": event.event_name().encode(),
+            b"source": source_bytes,
+            b"args": msgpack.packb(event.serialize()),
+        }
+        await self._msg_queue.send(raw_event)
```
HyeockJinKim, line 652: You'll probably have to transfer it some other way.

src/ai/backend/agent/kernel.py
```diff
@@ -1089,7 +1089,7 @@ async def read_output(self) -> None:
                                     else ModelServiceStatus.UNHEALTHY
                                 ),
                             )
-                            await self.event_producer.produce_event(event)
+                            await self.event_producer.anycast_event(event)
```
HyeockJinKim, line 1092: Anycast is good. 👍

src/ai/backend/agent/server.py
```diff
@@ -59,15 +59,18 @@
...
+from ai.backend.common.events.event_types.image import (
     ImagePullFailedEvent,
     ImagePullFinishedEvent,
     ImagePullStartedEvent,
 )
-from ai.backend.common.events.kernel import (
-    KernelLifecycleEventReason,
+from ai.backend.common.events.event_types.kernel.anycast import (
     KernelTerminatedEvent,
 )
```
HyeockJinKim: Some are separated into `anycast` and `broadcast`, and some are all included in the upper-level package, so please unify the package structure consistently.

src/ai/backend/common/distributed.py
```diff
@@ -9,7 +9,7 @@
 from ai.backend.logging import BraceStyleAdapter
 
 if TYPE_CHECKING:
-    from .events.dispatcher import AbstractEvent, EventProducer
+    from .events.dispatcher import AbstractAnycastEvent, EventProducer
```
HyeockJinKim: Isn't it unnecessary to be inside the `if TYPE_CHECKING` block? It seems like there shouldn't be any dependencies from `event.dispatcher`.

src/ai/backend/common/events/dispatcher.py
```diff
@@ -646,4 +648,24 @@ async def produce_event(
             b"source": source_bytes,
             b"args": msgpack.packb(event.serialize()),
         }
+        # TODO: impl anycast message queue
```
HyeockJinKim, line 651: Probably `anycast` messages will use the current message queue implementation as is, and it seems like we need to create a new broadcast method.
Since the places that use broadcast messages are doing subscribe operations that can almost be volatile, we are considering changing it to a pubsub-based approach.

src/ai/backend/manager/models/session.py
```diff
@@ -1582,9 +1588,14 @@ async def _post_status_transition(
...
             case SessionStatus.TERMINATED:
-                await self.event_producer.produce_event(
+                await self.event_producer.anycast_event(
                     SessionTerminatedEvent(session_row.id, session_row.main_kernel.status_info),
                 )
+                await self.event_producer.broadcast_event(
+                    SessionStartedBroadcastEvent(
+                        session_row.id, session_row.main_kernel.status_info
+                    ),
+                )
```
HyeockJinKim: Is it correct to send the SessionStartedBroadcastEvent? @fregataa

src/ai/backend/manager/registry.py
```diff
@@ -2446,13 +2467,20 @@ async def _decrease_concurrency_used(access_key: AccessKey, is_private: bool) ->
...
                                         reason,
                                     ),
                                 )
+                                await self.event_producer.broadcast_event(
+                                    SessionCancelledBroadcastEvent(
+                                        session_id,
+                                        target_session.creation_id,
+                                        reason,
+                                    ),
+                                )
```
HyeockJinKim: I think there was a function called `anycast_and_broadcast_event`, shouldn't we use that function?
It's not that the above function is good, but I want to easily check which parts need to be modified later.

It would be good to unify all parts where anycast and broadcast are used together with the above function, and then modify all parts using the above function in a batch later.

src/ai/backend/agent/agent.py
```diff
@@ -960,6 +985,14 @@ async def produce_error_event(
         pretty_tb = "".join(traceback.format_tb(tb)).strip()
         await self.produce_event(AgentErrorEvent(pretty_message, pretty_tb))
 
+    async def anycast_and_broadcast_event(
+        self,
+        anycast_event: AbstractAnycastEvent,
+        broadcast_event: AbstractBroadcastEvent,
+    ) -> None:
+        await self.produce_event(anycast_event)
+        await self.event_producer.broadcast_event(broadcast_event)
```
HyeockJinKim: Ah, this function was in the agent... Let's put it in common, and if both anycast and broadcast is needed, let's have them use this function from common.
And then, after the structure change, it seems good to organize the parts that use this function all at once.

src/ai/backend/agent/agent.py
```diff
@@ -78,40 +78,65 @@
...
+from ai.backend.common.events.event_types.kernel.broadcast import (
+    KernelPreparingEvent as KernelPreparingBroadcastEvent,
+)
+from ai.backend.common.events.event_types.kernel.broadcast import (
+    KernelPullingEvent as KernelPullingBroadcastEvent,
+)
+from ai.backend.common.events.event_types.kernel.broadcast import (
+    KernelStartedEvent as KernelStartedBroadcastEvent,
+)
+from ai.backend.common.events.event_types.kernel.broadcast import (
```
HyeockJinKim: Is there a reason why you should set alias?

fregataa: The same name events are already imported from `event_types.kernel.anycast` module

HyeockJinKim: The content written in the review was that from the beginning, the class name should be changed to the name currently set as the alias.
If classes with the same name are used with different intentions (anycast, broadcast), it seems easy to make mistakes.

src/ai/backend/common/events/event_types/agent/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+)
+
+__all__ = (
+    "AgentStartedEvent",
+    "AgentTerminatedEvent",
+    "AgentErrorEvent",
+    "AgentHeartbeatEvent",
+    "AgentImagesRemoveEvent",
+    "DoAgentResourceCheckEvent",
+)
```
HyeockJinKim: Do we really need to declare this in __init__.py?

src/ai/backend/common/events/types.py
```diff
@@ -73,3 +89,25 @@ def user_event(self) -> Optional[UserEvent]:
...
+
+class AbstractBroadcastEvent(AbstractEvent):
+    """
+    An event that should be broadcasted to all subscribers.
+    """
+
+    @classmethod
+    @override
+    def delivery_pattern(cls) -> DeliveryPattern:
+        return DeliveryPattern.BROADCAST
```
HyeockJinKim, line 113: 👍

src/ai/backend/manager/event_dispatcher/dispatch.py
```diff
@@ -37,17 +36,17 @@
...
     ModelServiceStatusEvent,
     RouteCreatedEvent,
 )
-from ai.backend.common.events.schedule import (
+from ai.backend.common.events.event_types.schedule import (
     DoCheckPrecondEvent,
     DoScaleEvent,
     DoScheduleEvent,
     DoStartSessionEvent,
 )
```
HyeockJinKim: How about putting all events into `{domain}.anycast`, `{domain}.broadcast` packages?

src/ai/backend/common/events/event_types/agent/__init__.py
fregataa: Since agent events are all anycast pattern, I'd like to make it easy to import

HyeockJinKim: I think it would improve readability if "anycast" and "broadcast" were explicitly visible in the import statements.

src/ai/backend/manager/api/events.py
```diff
@@ -211,7 +211,10 @@ async def push_background_task_events(
 async def enqueue_kernel_creation_status_update(
     app: web.Application,
     source: AgentId,
-    event: KernelPreparingEvent | KernelPullingEvent | KernelCreatingEvent | KernelStartedEvent,
+    event: KernelPreparingBroadcastEvent
+    | KernelPullingBroadcastEvent
+    | KernelCreatingBroadcastEvent
+    | KernelStartedBroadcastEvent,
 ) -> None:
```
HyeockJinKim, line 218: ??? Do we receive a broadcast event and then query DB?

HyeockJinKim:
It seems we need to modify the part that subscribes to the broadcast event separately.

PR #4626, opened 2025-06-07
fix(BA-1524): Config models validated only by alias

### LINKED ISSUE #4625: Config models only validated by alias
## Summary  

Currently, the Manager Unified Config is being validated only through aliases, which is causing compatibility issues with existing configurations.

Let's try to resolve this issue by using the `allow_population_by_field_name` setting or by passing `by_name` when calling `model_validate`.

jopemachine:
resolves #4625 (BA-1524)


HyeockJinKim (approved):

HyeockJinKim (approved):
Please ensure compatibility with existing operations afterwards.

PR #4651, opened 2025-06-10
refactor(BA-1542): Refactor test template with contextvars

### LINKED ISSUE #4648: Refactor `WrapperTestTemplate` with asynccontextmanager` and `contextvars`
## Motivation  

The current structure of `WrapperTestTemplate`'s separate setup and cleanup methods is not suitable for the statefulness required by the test code, and we want to refactor it based on `asynccontextmanager` and `contextvar`.

seedspirit:
resolves #4648 (BA-1542)

### DIFF
--- a/src/ai/backend/test/testcases/context.py
+++ b/src/ai/backend/test/testcases/context.py
@@ -1,6 +1,6 @@
 from contextlib import contextmanager
 from contextvars import ContextVar
-from typing import Generic, Iterator, Optional, TypeVar
+from typing import Generic, Iterator, Optional, TypeVar, final
 
 T = TypeVar("T")
 
@@ -21,6 +21,7 @@ class BaseTestContext(Generic[T]):
         cls._ctxvar = ContextVar[T](f"{cls.__name__}_ctxvar", default=None)
 
     @classmethod
+    @final
     def get_current(cls) -> T:
         """
         Get the current value from context
@@ -34,6 +35,7 @@ class BaseTestContext(Generic[T]):
         return res
 
     @classmethod
+    @final
     @contextmanager
     def with_current(cls, value: T) -> Iterator[None]:
         """Set the current value in the context"""


src/ai/backend/test/testcases/context.py
```diff
@@ -0,0 +1,31 @@
...
+        res = self._ctxvar.get()
+        if res is None:
+            raise RuntimeError(f"No value is set in {self.__class__.__name__}")
+        return res
+
+    def get_or_none(self) -> Optional[T]:
+        """Get the current value from context or None if not set"""
+        return self._ctxvar.get()
+
+    @asynccontextmanager
```
jopemachine: In our codebase, we import `asynccontextmanager` as `actxmgr` basically.

src/ai/backend/test/testcases/context.py
```diff
@@ -0,0 +1,31 @@
...
+    _ctxvar: ContextVar[Optional[T]]
+
+    def get_current(self) -> T:
+        """Get the current value from context"""
+        res = self._ctxvar.get()
+        if res is None:
+            raise RuntimeError(f"No value is set in {self.__class__.__name__}")
+        return res
+
+    def get_or_none(self) -> Optional[T]:
```
jopemachine: Is there specific reason why `get_or_none` is needed?
I thought test code always should throw exception if there are context variables that haven't been set up.

src/ai/backend/test/testcases/session/test_container_log_retriever.py
```diff
@@ -0,0 +1,42 @@
...
+    async def test(self) -> None:
+        test_id = test_run_context.get_test_id()
+        session = authentication_context.get_current()
+
+        try:
+            compute_session = await session.ComputeSession.get_or_create(
+                    image="cr.backend.ai/multiarch/python:3.13-ubuntu24.04",
+                    name=test_id,
+                )
+            if not compute_session:
```
jopemachine: Is there a case where compute_session can be `None`?
The return type of `get_or_create` is not optional.

src/ai/backend/test/testcases/context.py
```diff
@@ -0,0 +1,31 @@
+from contextlib import asynccontextmanager
+from contextvars import ContextVar
+from typing import Any, AsyncIterator, Generic, Optional, TypeVar
+
+T = TypeVar("T")
+
+
+class TestContextManager(Generic[T]):
```
jopemachine: It is not managing test contexts
How about renaming `BaseTestContext`?

src/ai/backend/test/testcases/root.py
```diff
@@ -0,0 +1,48 @@
...
+from typing import Any, AsyncIterator, Optional
+
+from ai.backend.test.testcases.context import TestContextManager
+from ai.backend.test.testcases.template import TestTemplate, WrapperTestTemplate
+
+
+class TestRunContext(TestContextManager):
+    """Context for the entire test run"""
+
+    _ctxvar: ContextVar[Optional[dict[str, Any]]]
```
jopemachine: Let's define a `TestMetadata` (or other name) type instead of using a dictionary.

src/ai/backend/test/testcases/root.py
```diff
@@ -0,0 +1,48 @@
...
+        self._ctxvar = ContextVar("test_run_context", default=None)
+
+    def get_current(self) -> Any:
+        return super().get_current()
+
+    def get_test_id(self) -> str:
+        return self.get_current()["test_id"]
+
+
+test_run_context = TestRunContext()
```
jopemachine: Instead of leaving test_run_context like this, how about making it a class variable?

src/ai/backend/test/testcases/session/test_container_log_retriever.py
```diff
@@ -0,0 +1,42 @@
...
+
+            info = await session.ComputeSession(name=test_id).get_info()
+
+            print(info)
+        finally:
+            await session.ComputeSession(name=test_id).destroy()
+
+
+class BasicContainerLogRetrieverTemplate(TestTemplate):
+    _testCode: TestContainerLogRetriever
```
jopemachine: why camel case?

src/ai/backend/test/testcases/utils.py
```diff
@@ -0,0 +1,52 @@
...
+from ai.backend.client.session import AsyncSession
+
+
+def get_test_temp_dir(test_id: str) -> Path:
+    temp_base = Path(tempfile.gettempdir()) / "backend_ai_tests"
+    temp_dir = temp_base / test_id
+    return temp_dir
+
+
+async def login(
```
jopemachine: The login test itself would still be useful.
But we can inject information such as the access key, secret key, and endpoint by passing `APIConfig` as an argument when creating the client session.
My opinion is that injecting APIConfig should be sufficient, as long as the behavior doesn't change when running on top of a login session.
I think managing and storing cookie data in a local path is cumbersome.

src/ai/backend/test/testcases/session/test_container_log_retriever.py
```diff
@@ -0,0 +1,42 @@
...
+from ai.backend.test.testcases.template import TestCode, TestTemplate
+from ai.backend.test.testcases.root import test_run_context
+from ai.backend.test.testcases.auth.login import authentication_context
+
+
+class TestContainerLogRetriever(TestCode):
+    @override
+    async def test(self) -> None:
+        test_id = test_run_context.get_test_id()
+        session = authentication_context.get_current()
```
jopemachine: I think a more descriptive name would be better than just "session."
It could be confusing with terms like client session or ComputeSession.

src/ai/backend/test/testcases/auth/login.py
```diff
@@ -0,0 +1,55 @@
...
+from typing import AsyncIterator, Optional, override
+
+from ai.backend.client.session import AsyncSession
+from ai.backend.test.testcases.context import TestContextManager
+from ai.backend.test.testcases.root import test_run_context
+from ai.backend.test.testcases.template import TestTemplate, WrapperTestTemplate
+from ai.backend.test.testcases.utils import login, logout
+
+
+class AuthenticationContext(TestContextManager):
```
jopemachine: Isn't `T` is `str`?
But get_current returns AsyncSession

src/ai/backend/test/testcases/session/test_container_log_retriever.py
jopemachine: How about managing the template as a separate file?

jopemachine (changes requested):

src/ai/backend/test/testcases/testcases.py
```diff
@@ -51,7 +57,22 @@ def default(cls) -> Self:
...
+            ),
+            "session": TestSpec(
+                name="session",
+                description="Test session management.",
+                tags={TestTag.SESSION, TestTag.MANAGER},
+                template=RootTestTemplate(
+                    template=AuthenticationWrapperTemplate(
+                        user_id="admin@lablup.com",
+                        password="wJalrXUt",
+                        otp=None,
```
jopemachine: I think you already know, but the user_id and password should be provided externally.

src/ai/backend/test/testcases/session/test_container_log_retriever.py
```diff
@@ -0,0 +1,42 @@
+from typing import override
+
+from ai.backend.test.testcases.template import TestCode, TestTemplate
+from ai.backend.test.testcases.root import test_run_context
+from ai.backend.test.testcases.auth.login import authentication_context
+
+
+class TestContainerLogRetriever(TestCode):
```
jopemachine: Overall, it seems that we need a convention for how to name things.

src/ai/backend/test/testcases/session/test_container_log_retriever.py
```diff
@@ -0,0 +1,42 @@
...
+    @override
+    async def test(self) -> None:
+        test_id = test_run_context.get_test_id()
+        session = authentication_context.get_current()
+
+        try:
+            compute_session = await session.ComputeSession.get_or_create(
+                    image="cr.backend.ai/multiarch/python:3.13-ubuntu24.04",
```
jopemachine: Let's declare these kinds of constants at the top of the code separately.

jopemachine (changes requested):

src/ai/backend/test/testcases/auth/login.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class AuthenticationWrapperTemplate(WrapperTestTemplate):
+    def __init__(
+        self, template: TestTemplate, user_id: str, password: str, otp: Optional[str] = None
+    ) -> None:
+        super().__init__(template)
+        self.user_id = user_id
+        self.password = password
+        self.otp = otp
```
HyeockJinKim: In test scenarios, for values that can be set up by preceding actions, please use contextvars rather than injecting them in the constructor. (e.g., In the case of testing a user signing up and then logging in, the value should be set up in the contextvar in the preceding signup wrapper.)

src/ai/backend/test/testcases/auth/login.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @property
+    def name(self) -> str:
+        return "auth-test"
```
HyeockJinKim: Please remove the postfix "test" as much as possible.
And the current behavior seems close to a login wrapper. Since "auth" is a very common name, please use names that better represent the implementation in the names.

src/ai/backend/test/testcases/context.py
```diff
@@ -0,0 +1,29 @@
...
+from contextvars import ContextVar
+from typing import Any, AsyncIterator, Generic, Optional, TypeVar
+
+T = TypeVar("T")
+
+
+class BaseTestContext(Generic[T]):
+    """Base class for test context managers"""
+
+    _ctxvar: ContextVar[Optional[T]]
```
HyeockJinKim: Using a `contextvar` with an instance variable defeats the purpose because this instance ultimately needs to be passed.
It should be a global variable or a class variable, but in this situation, it seems better to set it as a class variable since the class is being used as a kind of namespace to group things together.

However, if you create a `contextvar` in the base class, problems can occur if shadowing is not done in the subclasses, so it seems better to use `__init_subclass__` rather than declaring a class variable.

src/ai/backend/test/testcases/root.py
```diff
@@ -0,0 +1,44 @@
...
+
+    _ctxvar: ContextVar[Optional[dict[str, Any]]] = ContextVar("test_run_context", default=None)
+
+    @classmethod
+    def get_current(cls) -> Any:
+        return super().get_current()
+
+    @classmethod
+    def get_test_id(cls) -> str:
+        return cls.get_current()["test_id"]
```
HyeockJinKim: This part doesn't seem necessary.
And in this situation, **please don't specifically receive `dict` as a type.**
Just using contextvars already makes tracking difficult, but tracking dynamic fields with `dict` practically makes tracking impossible.

src/ai/backend/test/testcases/context.py
```diff
@@ -0,0 +1,48 @@
...
+            raise RuntimeError("Don't use BaseTestContext directly, subclass it instead")
+        res = cls._ctxvar.get()
+        if res is None:
+            raise RuntimeError(f"No value is set in {cls.__class__.__name__} tester context")
+        return res
+
+    @classmethod
+    @final
+    @contextmanager
+    def with_current(cls, value: T) -> Iterator[None]:
```
seedspirit, line 40: Just a question, why did you changed `BaseTestContext` methods as sync methods? Is it because `contextvars`' methods are all sync methods?

jopemachine (approved):

src/ai/backend/test/testcases/context.py
HyeockJinKim: Yes. ContextVar doesn't need async method.

HyeockJinKim (approved):

PR #4652, opened 2025-06-10
feat(BA-1541): Add main kernel id in session info response

### LINKED ISSUE #4647: Add main kernel id when quering sesssion information
## Motivation  

- When querying compute session information with the `/session/get_info` endpoint of the current session, we cannot get information about the kernel. To obtain this information, we need to perform an additional lookup on the kernel, so we add a lookup for the kernel's id to the response value of `/session/get_info`
- also need to add session name, ids

JIRA Issue: BA-1541

seedspirit:
resolves #4647 (BA-1541)

### DIFF
--- a/src/ai/backend/manager/services/session/service.py
+++ b/src/ai/backend/manager/services/session/service.py
@@ -1317,6 +1317,7 @@ class SessionService:
         resp["tag"] = sess.tag
 
         # Resource occupation
+        resp["kernelId"] = str(sess.main_kernel.id)
         resp["containerId"] = str(sess.main_kernel.container_id)
         resp["occupiedSlots"] = str(sess.main_kernel.occupied_slots)  # legacy
         resp["occupyingSlots"] = str(sess.occupying_slots)


src/ai/backend/manager/services/session/service.py
```diff
@@ -1317,6 +1317,7 @@ async def get_session_info(self, action: GetSessionInfoAction) -> GetSessionInfo
         resp["tag"] = sess.tag
 
         # Resource occupation
+        resp["kernelId"] = str(sess.main_kernel.id)
```
jopemachine, line 1320: Single session can have multiple kernels.
We need to change the kernel loading conditions and retrieve all kernels.

seedspirit: I had a similar thought, but api only returning `containerId` or `occupiedSlots` values for the `main_kernel`, so I matched it up with this. If I change it to return an id for all kernels, I'll need to change those parts to return values for all kernels as well

```
resp["lang"] = sess.main_kernel.image  # legacy
resp["image"] = sess.main_kernel.image
resp["architecture"] = sess.main_kernel.architecture
resp["registry"] = sess.main_kernel.registry
resp["containerId"] = str(sess.main_kernel.container_id)
resp["occupiedSlots"] = str(sess.main_kernel.occupied_slots)  # legacy
```

HyeockJinKim: It seems ambiguous to add the kernel ID to the response.
How about working on this anew after the tester work is finished later? @seedspirit

PR #4654, opened 2025-06-10
feat: Add session creation tests based on Tester

jopemachine:
resolves #NNN (BA-MMM)

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/test/testcases/session/create_single_node_multi_container_session.py
```diff
@@ -0,0 +1,63 @@
...
+            session_name = ComputeSessionContext.get_current()
+
+            EXPECTED_EVENTS = {
+                "session_enqueued",
+                "session_scheduled",
+                "kernel_preparing",
+                "kernel_creating",
+                "kernel_started",
+                "session_started",
+            }
```
HyeockJinKim: It seems like there may be timing issues, such as not receiving certain events, depending on the point in time when you subscribe to the event.

src/ai/backend/test/testcases/session/create_single_node_multi_container_session.py
```diff
@@ -0,0 +1,63 @@
...
+_IMAGE_NAME = "cr.backend.ai/stable/python:3.9-ubuntu20.04"
+_IMAGE_RESOURCES = {"cpu": 1, "mem": "512m"}
+_TEST_TIMEOUT = 10.0  # seconds
+
+
+# TODO: Refactor this using `parametrize` concept.
+class SingleNodeMultiContainerSessionCreation(TestCode):
+    async def test(self) -> None:
+        async with AsyncSession() as client_session:
+            session_name = ComputeSessionContext.get_current()
```
HyeockJinKim: It seems strange that session_name is already stored in the Context when testing the session creation scenario.

src/ai/backend/test/testcases/session/create_single_node_multi_container_session.py
```diff
@@ -0,0 +1,63 @@
...
+                "Session name should match the provided name"
+            )
+
+            try:
+                await listener_task
+                assert created_session.status == "RUNNING", "Session should be running"
+            except asyncio.TimeoutError as e:
+                raise asyncio.TimeoutError(
+                    f"Timed out after {_TEST_TIMEOUT}s; events received so far: {collected_events}"
+                ) from e
```
HyeockJinKim: Where is the timeout specified in the current code, as I don't see it?

src/ai/backend/test/testcases/session/create_single_node_multi_container_session.py
```diff
@@ -0,0 +1,63 @@
...
+                asyncio.wait_for(collect_events(), timeout=_TEST_TIMEOUT)
+            )
+
+            created_session = await client_session.ComputeSession.get_or_create(
+                _IMAGE_NAME,
+                name=session_name,
+                resources=_IMAGE_RESOURCES,
+                cluster_mode=ClusterMode.SINGLE_NODE,
+                cluster_size=3,
+            )
```
HyeockJinKim: Separately, but it seems that in this part, a session is created, and when the test ends, the session will not be cleaned up from backend.ai.

src/ai/backend/test/testcases/session/template.py
```diff
@@ -0,0 +1,27 @@
...
+    async def context(self) -> AsyncIterator[None]:
+        """
+        Context manager to set the compute session for the duration of the context.
+        """
+        session_name = f"test_session_{uuid.uuid4()}"
+        async with ComputeSessionContext.with_current(session_name):
+            yield
+
+            async with AsyncSession() as client_session:
+                await client_session.ComputeSession(session_name).destroy()
```
HyeockJinKim: It feels awkward that the template for creating a session name is separated from the template for creating a session.

HyeockJinKim (commented):
Please reflect the changes after the work from https://github.com/lablup/backend.ai/pull/4651 has been merged.

src/ai/backend/test/testcases/session/create_single_node_multi_container_session.py
jopemachine: `asyncio.wait_for` timed out after `_TEST_TIMEOUT` while expected events are not collected

```py
           listener_task = asyncio.create_task(
                asyncio.wait_for(collect_events(), timeout=_TEST_TIMEOUT)
            )
```

src/ai/backend/test/testcases/session/create_single_node_multi_container_session.py
jopemachine: It seems updating it so that the client can always correctly receive the event seems to be a separate issue.

But I agree with that it would be good if the test conditions could be relaxed.

src/ai/backend/test/testcases/session/create_single_node_multi_container_session.py
jopemachine: Could you please try again?
The session spawned in the test must be cleaned up in the `SessionSetupTemplateWrapper`.

src/ai/backend/test/testcases/session/create_single_node_multi_container_session.py
jopemachine: I passed the session name from the upper layer in order to clean up the session in the `SessionSetupTemplateWrapper`.

src/ai/backend/test/testcases/session/template.py
jopemachine: `SessionSetupTemplateWrapper` handle the creation of the session name and destroy,
while the actual creation of the session is performed in the test code,
since the purpose of the test code is to test the session creation process.

src/ai/backend/test/testcases/session/destroy_session.py
```diff
@@ -0,0 +1,52 @@
...
+                "kernel_terminating",
+                "kernel_terminated",
+            }
+
+            collected_events = set()
+
+            async def collect_events():
+                async with client_session.ComputeSession(session_name).listen_events() as events:
+                    async for event in events:
+                        data = json.loads(event.data)
```
HyeockJinKim: It looks like we should use a function that wraps orjson in common package.

src/ai/backend/test/testcases/session/destroy_session.py
```diff
@@ -0,0 +1,52 @@
+import asyncio
+import json
+
+from ai.backend.client.session import AsyncSession
+from ai.backend.test.templates.template import TestCode
+from ai.backend.test.testcases.session.context import ComputeSessionContext
+
+# Test environment configuration
+# TODO: Make these configurable loaderable by template wrapper
+_TEST_TIMEOUT = 30.0  # seconds
```
HyeockJinKim: I'm considering whether to receive timeout as a contextvar as well. (A method of using a default timeout if the contextvar is not set?)

src/ai/backend/test/testcases/session/destroy_session.py
```diff
@@ -0,0 +1,52 @@
...
+                async with client_session.ComputeSession(session_name).listen_events() as events:
+                    async for event in events:
+                        data = json.loads(event.data)
+                        assert data["reason"] == "user-requested", (
+                            "Session should be terminated by user request"
+                        )
+
+                        collected_events.add(event.event)
+                        if collected_events == EXPECTED_EVENTS:
+                            print("All expected events received.")
```
HyeockJinKim: Please remove the print after the work is done.

src/ai/backend/test/testcases/session/create_session.py
```diff
@@ -0,0 +1,73 @@
...
+_TEST_TIMEOUT = 30.0  # seconds
+
+
+@dataclass
+class CreateSessionArgs:
+    cluster_mode: ClusterMode
+    cluster_size: int = 1
+
+
+class CreateSession(TestCode):
```
HyeockJinKim: Ultimately, we need to clean up the session, shouldn't we implement it with a WrapperTemplate?
Sequence template does not guarantee session cleanup in case of an error in the middle.

configs/tester/tester.toml
```diff
@@ -12,3 +15,53 @@ password = "wJalrXUt"
...
+        },
+        "resources": {
+          "cpu": "2",
+          "mem": "4g"
+        }
+      }
+    }
+  }
+]
+"""
```
HyeockJinKim: Please separate the context object so that it doesn't become too large and get passed around.

src/ai/backend/test/contexts/config.py
```diff
@@ -0,0 +1,90 @@
...
+    @classmethod
+    def name(cls) -> ContextName:
+        return ContextName.LOGIN_CREDENTIAL
+
+
+class EndpointConfigContext(BaseTestContext[EndpointConfig]):
+    @classmethod
+    def name(cls) -> ContextName:
+        return ContextName.ENDPOINT
+
```
HyeockJinKim: It seems like these values are not only used in config, but can also be set in `parametrize`. Would it be okay to move them all to the config package?

HyeockJinKim: Please separate the packages focusing on "what domain they belong to" rather than "where they are currently used."
The places where they are used now can be changed, which could lead to an unnatural situation where you load the config package when using parametrize.

src/ai/backend/test/contexts/context.py
```diff
@@ -72,6 +80,19 @@ def current(cls) -> T:
...
+    def current_or_none(cls) -> Optional[T]:
+        """
+        Get the current value from context if it exists, otherwise return None.
+        """
+        if cls._ctxvar is None:
+            return None
+        res = cls._ctxvar.get()
+        if res is None:
+            return None
+        return res
```
HyeockJinKim: It seems like cases will become too complicated from the point of providing functions together with Optionals. I hope there are no branches in the test case itself as much as possible.
Wouldn't it be better to parametrize or separate the test cases?
I would actually recommend separating the test cases more. (Please make the test cases not complicated. I want to create **several simple tests to test various cases**, rather than a test case that can do multiple tasks.)

src/ai/backend/test/templates/auth/login.py
```diff
@@ -24,8 +24,8 @@ def name(self) -> str:
     async def _context(self) -> AsyncIterator[None]:
         test_id = TestIDContext.current()
         test_id_str = str(test_id)
-        credential = LoginCredentialContext.current()
-        endpoint = EndpointContext.current()
+        credential = LoginCredentialConfigContext.current()
+        endpoint = EndpointConfigContext.current()
```
HyeockJinKim: I want to consider cases that are not Config, so I would prefer not to use the name "Config".

src/ai/backend/test/templates/session/batch_session.py
```diff
@@ -0,0 +1,114 @@
...
+            created_session = await client_session.ComputeSession.get_or_create(
+                image_config.name,
+                architecture=image_config.architecture,
+                resources=session_config.resources,
+                type_="batch",
+                startup_command=batch_session_config.startup_command,
+                name=session_name,
+                cluster_mode=cluster_configs.cluster_mode,
+                cluster_size=cluster_configs.cluster_size,
+            )
```
HyeockJinKim: Please create each test case with the simplest possible flow.
Even if there is some duplicated code, simpler code is better than having a flag in the test code to make only parts of it behave differently.

src/ai/backend/test/tester/config.py
```diff
@@ -37,36 +43,122 @@ class LoginCredential(BaseModel):
...
     name: Optional[str] = Field(
         default=None,
         description="The Docker image to use for the test context.",
         examples=["cr.backend.ai/multiarch/python:3.13-ubuntu24.04"],
     )
+    architecture: Optional[str] = Field(
+        default=None,
+        description="The architecture of the session.",
+        examples=["x86_64"],
+    )
```
HyeockJinKim: Strictly speaking, this part should have the config and the internal dataclass separated.
That part is a bit cumbersome, so I haven't been attaching the Config postfix, but this makes me think about it a bit.

src/ai/backend/test/contexts/config.py
jopemachine: > It seems like these values are not only used in config, but can also be set in `parametrize`. Would it be okay to move them all to the config package?

I think setting it with parametrize can also be considered as a form of configuration.

src/ai/backend/test/templates/session/batch_session.py
jopemachine: Following the feedback, I separated the code, but there is quite a lot of duplicated code.
It would be nice if there were a way to consolidate the common code.

src/ai/backend/test/contexts/config.py
HyeockJinKim: In addition to parametrize, each value can also set values in the context from each template.

jopemachine: That’s the part I’m currently thinking about.
I wanted to separate the context configured directly in the config from the context set in the template.
Is there no need to separate these?
Wouldn't it become difficult to trace where the context was configured?

src/ai/backend/test/templates/session/utils.py
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    collected: set[str] = set()
+
+    async def collect_events() -> None:
+        async with client_session.ComputeSession(session_name).listen_events() as evs:
+            async for ev in evs:
+                collected.add(ev.event)
+                if ev.event == "session_cancelled":
+                    raise RuntimeError("Session creation was cancelled")
```
HyeockJinKim: Wouldn't it be better to raise an error when an event other than the expected event occurs, rather than adding exception handling for specific events like this?

HyeockJinKim: It seems like it would also be easier to generalize that way.

src/ai/backend/test/testcases/session/creation_failure_too_many_container.py
```diff
@@ -0,0 +1,39 @@
...
+            await client_session.ComputeSession.get_or_create(
+                image.name,
+                architecture=image.architecture,
+                name=session_name,
+                resources=session_config.resources,
+                cluster_mode=ClusterMode.SINGLE_NODE,
+                cluster_size=max_containers_per_session + 1,  # Exceeding the limit
+            )
+        except BackendAPIError as e:
+            assert e.data["error_code"] == "session_create_unavailable"
```
HyeockJinKim: Shouldn't we Raise an error if it doesn't fail?

src/ai/backend/test/templates/session/utils.py
```diff
@@ -0,0 +1,136 @@
...
+    collected_events: set[str] = set()
+
+    async def collect_events() -> None:
+        async with client_session.ComputeSession(session_name).listen_events() as events:
+            async for ev in events:
+                collected_events.add(ev.event)
+                if ev.event == "session_failure" or ev.event == "session_cancelled":
+                    raise RuntimeError(f"BatchSession failed with event: {ev.event}")
+                if collected_events == EXPECTED_EVENTS:
+                    break
```
HyeockJinKim: When thinking from a user scenario perspective, instead of subscribing to receive all events, wouldn't it be sufficient to subscribe to receive only the events of the desired final state?

src/ai/backend/test/templates/session/batch_session.py
```diff
@@ -0,0 +1,81 @@
...
+
+        listener = asyncio.create_task(
+            verify_session_events(
+                client_session,
+                session_name,
+                timeout,
+                "session_terminated",
+                {"session_failure", "session_cancelled"},
+            )
+        )
```
HyeockJinKim: Since it was created as an asyncio task, there is still a possibility that the compute session creation work actually finishes earlier and some events are missed.
It would be good to add a task like bgtask that caches the last event in Redis for a certain period and delivers it, so could you create an issue and link it to this comment? @jopemachine

src/ai/backend/test/templates/session/batch_session.py
```diff
@@ -0,0 +1,81 @@
...
+        assert created.created, "Session should be created successfully"
+        assert created.name == session_name, "Session name mismatch"
+
+        await listener
+        assert created.status in {"TERMINATING", "TERMINATED"}, (
+            f"Unexpected final status: {created.status}"
+        )
+        if created.id is None:
+            raise RuntimeError("Session ID is None after creation")
+        return created.id
```
HyeockJinKim: Is there a reason to verify after awaiting the listener when the created value is already in the received state?

src/ai/backend/test/templates/session/interactive_session.py
```diff
@@ -0,0 +1,104 @@
...
+        session_name = f"test_session_{str(test_id)}"
+        session_id = None
+        try:
+            session_id = await self._verify_session_creation(client_session, session_name)
+            with CreatedSessionIDContext.with_current(session_id):
+                yield
+        finally:
+            if session_id:
+                await self._verify_session_destruction(client_session, session_name)
+            pass
```
HyeockJinKim: There is unnecessary pass remaining.

src/ai/backend/test/templates/session/session_template.py
```diff
@@ -0,0 +1,232 @@
...
+        test_id = spec_meta.test_id
+        image = ImageContext.current()
+
+        if image.name is None:
+            raise ValueError("Image name is not set in ImageConfigContext.")
+        image_name = image.name.split(":")[0]
+        image_tag = image.name.split(":")[1]
+        session_config = SessionContext.current()
+
+        if session_config.resources is None:
```
HyeockJinKim: session_dep?

src/ai/backend/test/templates/session/session_template.py
```diff
@@ -0,0 +1,232 @@
...
+                            "mounts": {{}},
+                            "resources": {{
+                                "cpu": "{session_config.resources["cpu"]}",
+                                "mem": "{session_config.resources["mem"]}"
+                            }}
+                        }}
+                    }}
+                }}
+            ]
+        """)
```
HyeockJinKim: It's not a task we're doing immediately, but it would be good if these values were received from the client SDK in the form of dataclass or pydantic and then passed on. @jopemachine

src/ai/backend/test/templates/session/utils.py
```diff
@@ -0,0 +1,54 @@
...
+                        f"Session failed with event: {ev.event}, Expected event: {expected_event}"
+                    )
+
+    listener = asyncio.create_task(asyncio.wait_for(collect_events(), timeout))
+    try:
+        await listener
+    except asyncio.TimeoutError as e:
+        raise asyncio.TimeoutError(
+            f"Timeout after {timeout}s; no success event received, Actual events: {collected_events}"
+        ) from e
```
HyeockJinKim: Wouldn't using await here diminish the meaning of creating the asyncio.Task?

src/ai/backend/test/templates/session/session_template.py
```diff
@@ -0,0 +1,232 @@
...
+        listener = asyncio.create_task(
+            verify_session_events(
+                client_session,
+                session_name,
+                timeout,
+                "session_terminated",
+                {"session_failure", "session_cancelled"},
+                expected_termination_reason="user-requested",
+            )
+        )
```
HyeockJinKim: Thinking about it, instead of passing the timeout as an argument, it seems like it could be retrieved from the context internally in `verify_session_events`?

src/ai/backend/test/testcases/session/creation_failure_low_resources.py
```diff
@@ -0,0 +1,43 @@
...
+            await client_session.ComputeSession.get_or_create(
+                image.name,
+                architecture=image.architecture,
+                name=session_name,
+                # This failed due to the need for additional space for shared memory.
+                resources={"cpu": 1, "mem": min_mem},
+                cluster_mode=ClusterMode.SINGLE_NODE,
+                cluster_size=1,
+            )
+            assert False, "Expected BackendAPIError for low resources was not raised"
```
HyeockJinKim: raise `UnexpectedSuccess`.
Please use assert for value verification, and ensure that test failure cases other than value verification do not raise assertion errors.

src/ai/backend/test/tester/exporter.py
```diff
@@ -98,28 +98,30 @@ async def export_start(self) -> None:
     async def export_done(self) -> None:
         current = time.perf_counter()
         elapsed = current - self._started
-        print(f"\033[92m✓ '{self._spec_name()}' completed in {elapsed:.2f} seconds.")
+        print(f"\033[92m✓ '{self._spec_name()}' completed in {elapsed:.2f} seconds.\033[0m")
```
HyeockJinKim, line 101: Actually, it came up in the Copilot review last time... I was too lazy to apply it. I'm sorry.

src/ai/backend/test/templates/session/session_template.py
jopemachine: While it's possible to use Context everywhere, I think it's a better approach to pass it as a parameter.

src/ai/backend/test/testcases/session/execution_failure_wrong_command.py
```diff
@@ -0,0 +1,52 @@
...
+
+        async def collect_events():
+            async with client_session.ComputeSession(session_name).listen_events() as events:
+                async for event in events:
+                    if event.event == "session_failure":
+                        break
+                    if event.event == "session_cancelled":
+                        raise RuntimeError("BatchSession was cancelled")
+                    if event.event == "session_success":
+                        raise RuntimeError("BatchSession should not succeed with wrong command")
```
HyeockJinKim: We also use verify_session_event here.

src/ai/backend/test/testcases/session/execution_failure_wrong_command.py
```diff
@@ -0,0 +1,52 @@
...
+                type_="batch",
+                startup_command="some_wrong_command!@#123",
+                resources=session_dep.resources,
+                cluster_mode=ClusterMode.SINGLE_NODE,
+                cluster_size=1,
+            )
+
+            await listener_task
+        except BackendAPIError:
+            # TODO: Should `get_or_create` raise error when the command fails?
```
HyeockJinKim: I'm not sure if we really need to except exceptions. I think it might not convey information as well.

HyeockJinKim (approved):

HyeockJinKim (approved):

PR #4656, opened 2025-06-10
fix(BA-1543): Update container ports validation to prevent potential `IndexError`

rapsealk:
Follow-up of #1645 and resolves #4655 (BA-1543)

This pull request includes updates to improve type annotations and error handling in the codebase. The most important changes include updating container ports validation to prevent potential runtime errors.

### Type Annotation Improvements:
* Replaced `list[PortInfo] | None` with `Optional[list[PortInfo]]` in `_rollback_container_creation` to improve type clarity and ensure consistent handling of `None` values.
* Updated `str | None` to `Optional[str]` in `extract_image_command` to align with Python's standard type annotations.
* Changed `float | None | Sentinel` to `Optional[float] | Sentinel` in `push_image` and `float | None` to `Optional[float]` in `pull_image` for better readability. [[1]](diffhunk://#diff-4abe32fc0ac4d92071c86246bf16680a26548ebef7cbbb78e8d28514e5513f77L1678-R1678) [[2]](diffhunk://#diff-4abe32fc0ac4d92071c86246bf16680a26548ebef7cbbb78e8d28514e5513f77L1710-R1710)
* Replaced `ContainerNetworkInfo | None` with `Optional[ContainerNetworkInfo]` in `_rollback_container_creation` to maintain consistency in type annotations.

### Error Handling Enhancements:
* Updated container ports validation to catch omitted or empty list cases, preventing potential `IndexError` during runtime.

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version
- [x] Mention to the original issue

PR #4658, opened 2025-06-10
fix(BA-1543): Update container ports validation to prevent potential `IndexError` (#4656)

lablup-octodog:
This is an auto-generated backport PR of #4656 to the 24.09 release.


HyeockJinKim:
It seems it's not compatible with 24.09, please check it once. @rapsealk

PR #4661, opened 2025-06-11
feat(BA-1546): Add Tester configs template for config injection

### LINKED ISSUE #4660: Injection of APIConfig using toml config and a dedicated Wrapper template
## Outline

- APIConfig injection will be implemented on the Tester side by reading the toml file, not through the WrapperTemplate.
- To handle cases where different users' AK and SK need to be selected depending on the test case, multiple APIConfigs are included in the toml file.
- Tester should be implemented to optionally run all test cases for all users, or to run all test cases for a specific user only.

jopemachine:
resolves #4660 (BA-1546)


src/ai/backend/test/testcases/context.py
```diff
@@ -14,11 +15,31 @@ class BaseTestContext(Generic[T]):
...
+
+    @classmethod
+    def get_name(cls) -> str:
+        """
+        Get the name of the context
+        :return: name of the context
+        """
+        raise NotImplementedError(
+            f"{cls.__name__} must implement get_name method to return the context name"
+        )
```
HyeockJinKim: It would be good to use it as an abstractmethod, wouldn't it?

src/ai/backend/test/contexts/auth.py
```diff
@@ -0,0 +1,26 @@
...
+class LoginCredentialContext(BaseTestContext[LoginCredential]):
+    @classmethod
+    def get_name(cls) -> str:
+        return "login_credential"
+
+
+class LoginEndpointContext(BaseTestContext[Endpoint]):
+    @classmethod
+    def get_name(cls) -> str:
+        return "login_endpoint"
```
HyeockJinKim: get_name is good. 👍

src/ai/backend/test/templates/auth/keypair.py
```diff
@@ -0,0 +1,31 @@
...
+
+    @override
+    @actxmgr
+    async def context(self) -> AsyncIterator[None]:
+        keypair_ctx = KeypairContext.get_current()
+        endpoint_ctx = KeypairEndpointContext.get_current()
+
+        api_config = APIConfig(
+            endpoint=endpoint_ctx.endpoint,
+            endpoint_type="api",
```
HyeockJinKim, line 25: Is it okay if the endpoint_type is not injected?

src/ai/backend/test/contexts/auth.py
```diff
@@ -0,0 +1,26 @@
...
+class KeypairContext(BaseTestContext[KeyPair]):
+    @classmethod
+    def get_name(cls) -> str:
+        return "keypair"
+
+
+class KeypairEndpointContext(BaseTestContext[Endpoint]):
+    @classmethod
+    def get_name(cls) -> str:
+        return "keypair_endpoint"
```
HyeockJinKim: It doesn't seem necessary to have KeyPair as a prefix.

src/ai/backend/test/contexts/auth.py
HyeockJinKim: Here too, wouldn't it be better to use `EndpointContext` for everything, without deviding `KeypairEndpointContext` and `LoginEndpointContext`?

src/ai/backend/test/templates/auth/login.py
```diff
@@ -4,40 +4,43 @@
...
+    def __init__(self, template: TestTemplate) -> None:
         super().__init__(template)
-        self.user_id = user_id
-        self.password = password
-        self.otp = otp
 
     @property
     def name(self) -> str:
-        return "login"
+        return "auth_login"
```
HyeockJinKim: I think it's clear even without explicitly adding "auth".

src/ai/backend/test/testcases/auth/testspecs.py
```diff
@@ -0,0 +1,12 @@
+from ai.backend.test.templates.auth.login import LoginTemplate
+from ai.backend.test.templates.template import BasicTestTemplate, NopTestCode
+from ai.backend.test.testcases.testcases import TestSpec, TestTag
+
+AUTH_TEST_SPECS = {
+    "login": TestSpec(
+        name="login",
+        description="Test user login functionality.",
+        tags={TestTag.WEBSERVER, TestTag.AUTH},
+        template=LoginTemplate(template=BasicTestTemplate(NopTestCode())),
```
HyeockJinKim: It's more annoying than I thought to attach the basic template every time...
I will think about how to improve this part later.

src/ai/backend/test/testcases/context.py
```diff
@@ -14,11 +15,31 @@ class BaseTestContext(Generic[T]):
     """
 
     _ctxvar: Optional[ContextVar[Optional[T]]] = None
+    _used: Mapping[str, "BaseTestContext"] = {}
```
HyeockJinKim: Thinking about it, `Mapping` type is not mutable, but we are adding values to cls._used. Could you make the field a MutableMapping?

src/ai/backend/test/testcases/testcases.py
```diff
@@ -49,7 +49,11 @@ def __init__(self, specs: Mapping[str, TestSpec]) -> None:
 
     @classmethod
     def default(cls) -> Self:
+        # TODO: Resolve cyclic import
```
HyeockJinKim: The default function was initially a function to easily show the structure, so I think it's about time to get rid of it.
Could you separate the test case packages like you separated the auth?

src/ai/backend/test/tester/tester.py
```diff
@@ -17,13 +26,30 @@ def __init__(self, spec_manager: TestSpecManager, exporter: TestExporter) -> Non
...
-            runner = TestRunner(spec, self._exporter)
-            await runner.run()
+            with ExitStack() as stack:
+                for key, ctx in ctx_map.items():
+                    if config := getattr(tester_config.context, key, None):
+                        ctx_mgr = ctx.with_current(config)
+                        stack.enter_context(ctx_mgr)
+
+                runner = TestRunner(spec, self._exporter)
+                await runner.run()
```
HyeockJinKim: 👍

src/ai/backend/test/tester/tester.py
```diff
@@ -17,13 +26,30 @@ def __init__(self, spec_manager: TestSpecManager, exporter: TestExporter) -> Non
...
+            raw_content = await fp.read()
+            content = tomli.loads(raw_content)
+            config = TesterConfig.model_validate(content, by_alias=True)
+            return config
+
     async def _run_spec(self, spec: TestSpec) -> None:
         """
         Run a single test specification.
         """
+        tester_config = await self._load_tester_config("tester.toml")
```
HyeockJinKim: The config path should be injected in the constructor.

src/ai/backend/test/templates/auth/keypair.py
jopemachine: Yes, keypair authentication always use api mode

src/ai/backend/test/contexts/auth.py
jopemachine: I separated the two types because both the login endpoint and the keypair endpoint are needed simultaneously.
(Certain tests use API mode, while others use Session mode.)

src/ai/backend/test/contexts/auth.py
jopemachine: It's clear that the names need to be improved. I will update them tomorrow.

src/ai/backend/test/testcases/auth/testspecs.py
jopemachine: There is still room for improvement, but I have applied the middleware structure on a trial basis for now.

src/ai/backend/test/cli/__main__.py
```diff
@@ -116,7 +120,7 @@ def run_cli(
 @main.command()
 @click.pass_obj
 def get_all_specs(cli_ctx: CLIContext) -> None:
-    spec_manager = TestSpecManager.default()
+    spec_manager = TestSpecManager(specs=ROOT_TEST_SPECS)
```
HyeockJinKim: Let's make the TestSpecManager creation a private function, since it's always the same in the cli.

src/ai/backend/test/contexts/tester.py
```diff
@@ -4,4 +4,6 @@
 
 
 class TestIDContext(BaseTestContext[uuid.UUID]):
-    pass
+    @classmethod
+    def get_name(cls) -> str:
+        return "test_id"
```
HyeockJinKim: I don't particularly prefer putting a `get_` prefix on values for getters. How about leaving it as a property instead?

HyeockJinKim: Ah, because it's a classmethod, it probably won't work simultaneously... How about just changing the name to "name"?

jopemachine: To use `classproperty` in Python, we need additional code for it.
See https://stackoverflow.com/questions/128573/using-property-on-classmethods
If you have a preferred implementation, please let me know and I will update PR.

src/ai/backend/test/templates/auth/login.py
```diff
@@ -28,16 +24,22 @@ def name(self) -> str:
     async def context(self) -> AsyncIterator[None]:
         test_id = TestIDContext.get_current()
         test_id_str = str(test_id)
-        async with AsyncSession() as session:
+        credential_ctx = LoginCredentialContext.get_current()
+        endpoint_ctx = EndpointContext.get_current()
```
HyeockJinKim: This part seems to be an actual value rather than ctx, so the `_ctx` postfix doesn't seem necessary.

src/ai/backend/test/contexts/tester.py
jopemachine: > Ah, because it's a classmethod, it probably won't work simultaneously... How about just changing the name to "name"?

Ok, I will reflect them after you commit

src/ai/backend/test/templates/template.py
```diff
@@ -1,11 +1,42 @@
...
+) -> AsyncIterator[None]:
+    async with AsyncExitStack() as stack:
+        try:
+            stage_name = "undefined"
+            for w in wrapper_templates:
+                if isinstance(w, WrapperTestTemplate):
+                    wrapper = w
+                else:
+                    empty = BasicTestTemplate(NopTestCode())
+                    # TODO: Improve this type hinting
```
HyeockJinKim: It doesn't seem like a good idea to separate the functionality already provided by `WrapperTemplate` into another function and provide it in the same way. If it's the same functionality, it would be better to provide it in the same class in terms of future maintenance.

src/ai/backend/test/templates/template.py
```diff
@@ -48,33 +78,44 @@ async def run_test(self, exporter: TestExporter) -> None:
 @final
 class BasicTestTemplate(TestTemplate):
     _testcode: TestCode
+    _wrapper_templates: list["WrapperTestTemplateProtocol"]
 
-    def __init__(self, testcode: TestCode) -> None:
+    def __init__(
+        self, testcode: TestCode, wrapper_templates: list["WrapperTestTemplateProtocol"] = []
+    ) -> None:
```
HyeockJinKim: It seems better to use `abstractmethod` rather than trying to use the constructor, since the constructors may vary for each class.

src/ai/backend/test/cli/__main__.py
```diff
@@ -4,16 +4,20 @@
...
-from ai.backend.test.testcases.testcases import TestSpecManager
+from ai.backend.test.testcases.spec_manager import TestSpecManager
+from ai.backend.test.testcases.testspecs import ROOT_TEST_SPECS
 from ai.backend.test.tester.exporter import PrintExporter
 from ai.backend.test.tester.tester import Tester
 
 from .context import CLIContext
 from .utils import CommaSeparatedChoice, CustomUsageArgsCommand
 
+_CONFIG_PATH = Path("tester.toml")
```
HyeockJinKim: It seems good to receive a config path as a CLI option and use this value as the default.

src/ai/backend/test/templates/template.py
```diff
@@ -88,33 +129,46 @@ async def context(self) -> AsyncIterator[None]:
...
+        async with _apply_wrapper_templates(self._wrapper_templates, exporter):
             try:
-                await template.run_test(exporter)
+                for template in self._templates:
+                    try:
+                        await template.run_test(exporter)
+                        await exporter.export_stage_done(self.name)
+                    except BaseException as e:
+                        await exporter.export_stage_exception(self.name, e)
+                        raise
```
HyeockJinKim: It seems the implementation is wrong. The completion or failure status in each template should only be performed within each template and should only be called once.

HyeockJinKim (approved):

PR #4679, opened 2025-06-12
fix(BA-1560): Skip failure of code runner init

### LINKED ISSUE #4678: Fail to initialize code runner when agent restarts
Agents fail to initialize code runner if any kernel object hasn’t initialize code runner

JIRA Issue: BA-1560

fregataa:
resolves #4678 (BA-1560)

### DIFF
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -1841,7 +1841,15 @@ class AbstractAgent(
             pass
         for kernel_obj in self.kernel_registry.values():
             kernel_obj.agent_config = self.local_config
-            await kernel_obj.init(self.event_producer)
+            try:
+                await kernel_obj.init(self.event_producer)
+            except Exception as e:
+                log.error(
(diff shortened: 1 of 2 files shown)


src/ai/backend/agent/agent.py
```diff
@@ -1841,7 +1841,15 @@ async def scan_running_kernels(self) -> None:
...
-            await kernel_obj.init(self.event_producer)
+            try:
+                await kernel_obj.init(self.event_producer)
+            except Exception as e:
+                log.error(
+                    "Failed to initialize kernel {}: {}, skipping",
+                    kernel_obj.kernel_id,
+                    e,
+                )
+                continue
```
HyeockJinKim, line 1852: Is it okay to create code that ignores failed kernel and generates the rest?

fregataa: The kernel won't work as we wanted even before restarting the agent. It will be more clear to just kill the kernel but I am not sure it is a good way to handle it

PR #4680, opened 2025-06-12
feat(BA-1531): Add session rename tests in tester

### LINKED ISSUE #4635: Add Session test (Rename Session)

seedspirit:
resolves #4635 (BA-1531)


src/ai/backend/test/testcases/session/rename_failure_duplicated_name.py
```diff
@@ -0,0 +1,40 @@
+from typing import override
+
+from ai.backend.client.exceptions import BackendAPIError
+from ai.backend.test.contexts.client_session import ClientSessionContext
+from ai.backend.test.contexts.tester import TestIDContext
+from ai.backend.test.templates.template import TestCode
+
+# TODO: Change to injecting from an external config file
+_IMAGE_NAME = "cr.backend.ai/multiarch/python:3.13-ubuntu24.04"
```
HyeockJinKim: Use Image Context rather than const variable.

src/ai/backend/test/templates/session/batch_session.py
```diff
@@ -75,8 +77,10 @@ async def _context(self) -> AsyncIterator[None]:
         spec_meta = TestSpecMetaContext.current()
         test_id = spec_meta.test_id
         client_session = ClientSessionContext.current()
-        session_name = f"test_session_{str(test_id)}"
+        session_name = f"test_session_{test_id}_{int(time.time() * 1000000) % 1000000:06d}"
```
HyeockJinKim: Is it necessary to include a time value when the test_id is already a random value?

seedspirit: When creating two or more sessions based on a session template, we need to give them different names. Currently, it's based on the test id, so if we create a session with the template within the same TestSpec, it will create a session with the same name, which will cause an error. So we tried to bypass this by putting a postfix based on milliseconds

seedspirit: But as I chose making second session direct in TestCode, I've removed it

src/ai/backend/test/testcases/session/rename_failure_duplicated_name.py
```diff
@@ -0,0 +1,43 @@
...
+        )
+        try:
+            await client_session.ComputeSession(name=second_session_name).rename(session_name)
+            raise AssertionError(
+                "Renaming a session to a name that already exists should raise an error"
+            )
+        except BackendAPIError:
+            pass
+        finally:
+            await client_session.ComputeSession(name=second_session_name).destroy()
```
HyeockJinKim, line 43: It seems like we can just use the existing Session-related Wrapper?

seedspirit: As current teseter context structure memorize only one session meta info, we have to create second session manually on `TestCode`

HyeockJinKim: Let's do this task in the next sprint.

HyeockJinKim (approved):

PR #4690, opened 2025-06-13
fix(BA-1567): Fix Session Rename API to prevent duplicate names

### LINKED ISSUE #4689: Fix Session Rename API to prevent duplicate names
## Summary  

- Current `/session/rename` API is allowing duplicated name
- However modify session gql mutation does not allow duplicated name renaming
- As [backend.ai](http://backend.ai) manager system raise error when ‘too many session’ matched, we need to disallow `/session/rename` prevent duplicate name renaming

##

seedspirit:
resolves #4689 (BA-1567)

### DIFF
--- a/src/ai/backend/manager/services/session/service.py
+++ b/src/ai/backend/manager/services/session/service.py
@@ -1437,15 +1437,15 @@ class SessionService:
         new_name = action.new_name
 
         async with self._db.begin_session() as db_sess:
-            session_exists = await SessionRow.get_session(
+            session_with_new_name = await SessionRow.get_session(
                 db_sess,
                 new_name,
                 owner_access_key,
                 kernel_loading_strategy=KernelLoadingStrategy.NONE,
                 allow_stale=True,
             )
-            if session_exists:
-                raise InvalidAPIParameters("Session with new name already exists")
+            if session_with_new_name is not None:
+                raise InvalidAPIParameters(f"Session name of '{new_name}' already exists")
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/services/session/service.py
```diff
@@ -1437,6 +1437,19 @@ async def rename_session(self, action: RenameSessionAction) -> RenameSessionActi
...
+                    new_name,
+                    owner_access_key,
+                    kernel_loading_strategy=KernelLoadingStrategy.NONE,
+                )
+            except SessionNotFound:
+                pass
+            else:
+                raise InvalidAPIParameters(
+                    f"Duplicate session name. Session(id:{sess.id}) already has name({sess.name})"
+                )
```
HyeockJinKim, line 1452: I don't quite understand, can I know the history of why the name of the session isn't set to be unique? @fregataa 
If it doesn't have to be unique, I'm also curious as to why this task is necessary.

HyeockJinKim: Let's work in the direction of making it use the existing modify_session. It doesn't seem necessary to handle Rename differently on purpose.

seedspirit: `modify_session` requires `session_id`, but rename rest api doesn't require `session_id`. So I think it is better to fix rename api first, then change to `modify_session` when we upgrade api version

PR #4697, opened 2025-06-13
feat(BA-1561): Add session status history retriever test

### LINKED ISSUE #4682: Add Session test (Retrieve Status History)

seedspirit:
resolves #4682 (BA-1561)

### DIFF
--- a/src/ai/backend/test/testcases/session/testspecs.py
+++ b/src/ai/backend/test/testcases/session/testspecs.py
@@ -15,6 +15,7 @@ from ai.backend.test.templates.session.session_template import (
     InteractiveSessionFromTemplateTemplate,
     SessionTemplateTemplate,
 )
+from ai.backend.test.templates.template import BasicTestTemplate, NopTestCode
 from ai.backend.test.testcases.session.commit import (
     InteractiveSessionCommitSuccess,
     InteractiveSessionImagifySuccess,
@@ -45,14 +46,15 @@ from ai.backend.test.testcases.session.rename_failure_duplicated_name import (
     SessionRenameFailureDuplicatedName,
 )
 from ai.backend.test.testcases.session.session_rename import TestSessionRename
+from ai.backend.test.testcases.session.session_status_history_retriever import (
+    SessionStatusHistoryRetriever,
+)
 from ai.backend.test.testcases.spec_manager import TestSpec, TestTag
 from ai.backend.test.tester.dependency import (
     BootstrapScriptDep,
     ClusterDep,
     CodeExecutionDep,
 )
-
-from ...templates.template import BasicTestTemplate, NopTestCode
 
 BATCH_SESSION_TEST_SPECS = {
     "creation_batch_session_success": TestSpec(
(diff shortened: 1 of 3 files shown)


src/ai/backend/test/testcases/session/status_history_retriever.py
```diff
@@ -0,0 +1,30 @@
...
+        )
+
+        try:
+            result = await session.ComputeSession(name=test_name).get_status_history()
+            assert result["result"] != "", "Status history should not be empty"
+            # NOTE: SessionStatus enum is not available due to BUILD rules
+            # NOTE: src/ai/backend/test/BUILD[//src/ai/backend/**] -> src/ai/backend/manager/BUILD[!*] : DENY
+            # expected_status = [status.value for status in SessionStatus]
+            # for status_key, status_value in result["result"].items():
+            #     assert status_key in expected_status, f"Status {status_key} not in SessionStatus enum"
```
seedspirit: @jopemachine 
Currently, there is very little that can be validated by the `BUILD` rule - unless we move the `SessionStatus` to `common` package, I guess we can just check if it has a value, but is that enough?

seedspirit: or can I write SessionStatus values with raw string, not enum?

jopemachine: It is correct to move SessionStatus to the common package.
But if it takes some time, I think it wouldn't be a big problem to use a raw string for now. 
What do you think about it? @HyeockJinKim

src/ai/backend/test/testcases/session/status_history_retriever.py
```diff
@@ -0,0 +1,38 @@
...
+        ])
+
+        result = await session.ComputeSession(name=session_name).get_status_history()
+        assert result["result"] != "", "Status history should not be empty"
+        # NOTE: SessionStatus enum is not available due to BUILD rules
+        # NOTE: src/ai/backend/test/BUILD[//src/ai/backend/**] -> src/ai/backend/manager/BUILD[!*] : DENY
+        for status_key, status_value in result["result"].items():
+            assert status_key in expected_status, (
+                f"Status {status_key} is not a valid session status"
+            )
```
HyeockJinKim: What kind of test are you trying to do?

src/ai/backend/test/testcases/session/testspecs.py
```diff
@@ -265,9 +265,31 @@
...
+}
 
 SESSION_TEST_SPECS = {
-    **BATCH_SESSION_TEST_SPECS,
-    **INTERACTIVE_SESSION_TEST_SPECS,
-    **SESSION_TEMPLATE_TEST_SPECS,
+    # **BATCH_SESSION_TEST_SPECS,
+    # **INTERACTIVE_SESSION_TEST_SPECS,
+    # **SESSION_TEMPLATE_TEST_SPECS,
+    **SESSION_INFO_RETRIEVER_TEST_SPECS
```
HyeockJinKim: Is there a reason to ignore existing test cases?

seedspirit: I ignored other test cases to just test my testcode, and it was mistake not to revert it. I will revert it

src/ai/backend/test/testcases/session/status_history_retriever.py
seedspirit: I was trying to test the session status history retrieval API is working properly and second, none of the retrieved status histories have unexpected session history values

HyeockJinKim: Please create tests for this part like, "In both success and failure scenarios, a specific history must have existed." I don't think it's necessary to verify parts that are already checked by type checking with test code again.

seedspirit: As `get_status_history` only retrieve activate session history, I only defined test code with success scenario

Issue #4700, opened 2025-06-13
 Define Name field to TestSpec, and utilize them in Exporter


jopemachine:
## Motivation  

Currently, we can only identify the name of the stage where the error occurred, making it difficult to understand the overall context.

Let's add a name to `TestSpec` as well, and include this information during export.

JIRA Issue: BA-1573

PR #4701, opened 2025-06-13
feat(BA-1572, BA-1573): Improve Exporter to easily check error information when an error occurs

### LINKED ISSUE #4699: Define new `DefaultExporter` of Tester package
## Motivation  

Since it's difficult to catch errors occurring during the current development process,
let's define and use a new `DefaultExporter` that not only logs to the console but also writes a JSON file to a local path when an error occurs.

HyeockJinKim:
resolves #4699 #4700 (BA-1572, BA-1573)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/test/contexts/context.py
+++ b/src/ai/backend/test/contexts/context.py
@@ -14,12 +14,13 @@ class ContextName(enum.StrEnum):
     This enum provides a set of predefined context names that can be used to identify different contexts in the testing framework.
     """
 
-    TEST_ID = "test_id"
+    TEST_SPEC_META = "test_spec_meta"
     IMAGE = "image"
     CLIENT_SESSION = "client_session"
     ENDPOINT = "endpoint"
     KEYPAIR = "keypair"
     LOGIN_CREDENTIAL = "login_credential"
+    TEST_ERROR_OUTPUT_DIRECTORY = "test_error_output_directory"
 
 
 class BaseTestContext(Generic[T]):
--- a/src/ai/backend/test/tester/exporter.py
+++ b/src/ai/backend/test/tester/exporter.py
@@ -1,6 +1,12 @@
 import time
+import traceback
 from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from pathlib import Path
 from typing import Optional
+
+from ai.backend.common.json import pretty_json_str
+from ai.backend.test.contexts.tester import TestSpecMetaContext
 
 
 class TestExporter(ABC):
@@ -14,7 +20,7 @@ class TestExporter(ABC):
         raise NotImplementedError("Subclasses must implement this method.")
 
     @abstractmethod
-    async def export_start(self, spec_name: str) -> None:
+    async def export_start(self) -> None:
         """
         Export the start of a test run.
         This method can be used to log or notify the start of a test case.
@@ -22,7 +28,7 @@ class TestExporter(ABC):
         raise NotImplementedError("Subclasses must implement this method.")
 
     @abstractmethod
-    async def export_done(self, spec_name: str) -> None:
+    async def export_done(self) -> None:
         """
         Export the result of a test run.
         """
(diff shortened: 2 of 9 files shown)


src/ai/backend/test/tester/exporter.py
```diff
@@ -43,64 +49,118 @@ async def export_stage_exception(self, stage: str, exception: BaseException) ->
...
+        self._last_stage = stage
+        print(f"\033[92m✓ '{self._stage_name(stage)}' completed in {elapsed:.2f} seconds.")
 
     async def export_stage_exception(self, stage: str, exception: BaseException) -> None:
         current = time.perf_counter()
         elapsed = current - self._last_stage_done
         self._last_stage_done = current
         print(
-            f"Stage '{stage}' failed after {elapsed:.2f} seconds. {exception.__class__.__name__}: {exception}"
+            f"\033[91m❌ '{self._stage_name(stage)}' failed after {elapsed:.2f} seconds. {exception.__class__.__name__}"
```
jopemachine, line 115: How about defining a function like `_print_error` instead of putting special characters directly inside the print statement?

src/ai/backend/test/contexts/context.py
```diff
@@ -14,12 +14,13 @@ class ContextName(enum.StrEnum):
...
     """
 
-    TEST_ID = "test_id"
+    TEST_SPEC_META = "test_spec_meta"
     IMAGE = "image"
     CLIENT_SESSION = "client_session"
     ENDPOINT = "endpoint"
     KEYPAIR = "keypair"
     LOGIN_CREDENTIAL = "login_credential"
+    TEST_ERROR_OUTPUT_DIRECTORY = "test_error_output_directory"
```
jopemachine, line 23: How about placing this config under the `ExporterConfigContext`?

src/ai/backend/test/tester/exporter.py
```diff
@@ -43,64 +49,118 @@ async def export_stage_exception(self, stage: str, exception: BaseException) ->
...
+        spec_name = spec_meta.spec_name
         if not self._sub_name:
             return spec_name
-        return f"{spec_name} ({self._sub_name})"
+        return f"Test {spec_name} ({self._sub_name})"
+
+    def _stage_name(self, stage: str) -> str:
+        """
+        Helper method to format the stage name for export.
+        This can be overridden in subclasses for custom formatting.
```
jopemachine, line 140: If a method can be overridden, wouldn't it be better to make it public?

src/ai/backend/test/tester/exporter.py
```diff
@@ -43,64 +49,118 @@ async def export_stage_exception(self, stage: str, exception: BaseException) ->
...
+    _last_stage: Optional[str]
     _last_stage_done: float
+    _output_directory: Path
 
     def __init__(self, sub_name: Optional[str], started: float, last_stage_done: float) -> None:
         self._sub_name = sub_name
         self._started = started
+        self._last_stage = None
         self._last_stage_done = last_stage_done
+        self._output_directory = _DEFAULT_OUTPUT_DIRECTORY
```
jopemachine, line 84: It seems that `_output_directory` is always using the default value.

src/ai/backend/test/tester/exporter.py
HyeockJinKim: I don't assume overriding, and since it's not intended for external use or a behavior to be exposed externally, I think it's right to keep it private.

jopemachine: The comment insist that it could be overriden

src/ai/backend/test/tester/exporter.py
HyeockJinKim: Yes. I'm just leaving it as the default value and I'm not planning on considering the output yet.
The part I'm worried about right now is that this variable is only used in default exporter, so its injecting with context is difficult to handle in common.

PR #4704, opened 2025-06-14
feat(BA-1574): Add metric for sync container lifecycle task

### LINKED ISSUE #4703: Better observability for container lifecycle sync task
The `sync_container_lifecycles()` agent task operates on a regular schedule, synchronizing the kernel registry with the containers. It is important to monitor the task to ensure it is triggered and completed successfully.

fregataa:
resolves #4703 (BA-1574)

### DIFF
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -1355,7 +1355,7 @@ class AbstractAgent(
         ) as tg:
             while True:
                 ev = await self.container_lifecycle_queue.get()
-                log.info("process_lifecycle_events(): received lifecycle event: {}", str(ev))
+                log.info("received lifecycle event: {}", str(ev))
                 if isinstance(ev, Sentinel):
                     await self.save_last_registry(force=True)
                     return
@@ -1623,7 +1623,7 @@ class AbstractAgent(
             await self.produce_error_event()
         else:
             self._sync_container_lifecycle_observer.observe_container_lifecycle_success(
-                agent_id=self.id
+                agent_id=self.id, num_synced_kernels=len(terminated_kernels)
             )
 
     async def set_container_count(self, container_count: int) -> None:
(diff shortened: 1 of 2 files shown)


src/ai/backend/agent/agent.py
```diff
@@ -1345,6 +1347,7 @@ async def lifecycle_task_exception_handler(
         ) as tg:
             while True:
                 ev = await self.container_lifecycle_queue.get()
+                log.info("process_lifecycle_events(): received lifecycle event: {!r}", ev)
```
HyeockJinKim: It would be better to log only the type values instead of logging the event itself.

fregataa: I updated this to log `str(event)` because the event type has a string magic method.
https://github.com/lablup/backend.ai/blob/54d1e0562051cbdafd1fb90b5da57d1fd260f68f/src/ai/backend/agent/types.py#L109-L120

src/ai/backend/agent/agent.py
```diff
@@ -1353,6 +1355,7 @@ async def lifecycle_task_exception_handler(
         ) as tg:
             while True:
                 ev = await self.container_lifecycle_queue.get()
+                log.info("process_lifecycle_events(): received lifecycle event: {}", str(ev))
```
HyeockJinKim: I think the function name is populated by opentelemetry, so you don't have to write it down.

src/ai/backend/agent/agent.py
```diff
@@ -1599,13 +1606,25 @@ def _get_session_id(container: Container) -> SessionId | None:
...
+            )
         except Exception as e:
             log.exception(f"sync_container_lifecycles() failure, continuing (detail: {repr(e)})")
+            self._sync_container_lifecycle_observer.observe_container_lifecycle_failure(
+                agent_id=self.id, exception=e
+            )
             await self.produce_error_event()
+        else:
+            self._sync_container_lifecycle_observer.observe_container_lifecycle_success(
+                agent_id=self.id
```
HyeockJinKim: It seems difficult to collect meaningful metrics by only collecting failure and success metrics here. Shouldn't we actually increase the number of Kernels that are synced?

PR #4710, opened 2025-06-16
feat(BA-1552): Add purge container RPC

### LINKED ISSUE #4667: Add RPC to force delete dangling containers
Dangling containers do not match with agent’s kernel registry.

Should add a function to force-delete dangling containers that bypass kernel registry and clean up after container deletion

JIRA Issue: BA-1552

fregataa:
resolves #4667 (BA-1552)


src/ai/backend/agent/agent.py
```diff
@@ -1340,6 +1330,76 @@ async def _handle_clean_event(self, ev: ContainerLifecycleEvent) -> None:
...
+        # Exclude out-of-range ports, because when the agent restarts
+        # with a different port range, existing containers' host ports
+        # may not belong to the new port range.
+        restored_ports = [
+            *filter(
+                lambda p: port_range[0] <= p <= port_range[1],
+                host_ports,
+            )
+        ]
+        self.port_pool.update(restored_ports)
```
HyeockJinKim: Let's define it as a private method.

src/ai/backend/agent/agent.py
```diff
@@ -1340,6 +1330,76 @@ async def _handle_clean_event(self, ev: ContainerLifecycleEvent) -> None:
...
+                log.warning(f"filter_kernel_registry_by_containers(): removing {kid} from registry")
+                if kernel_obj.runner is not None:
+                    await kernel_obj.runner.close()
+                await kernel_obj.close()
+                host_ports = kernel_obj.get("host_ports")
+                if host_ports is not None:
+                    self.restore_ports(host_ports)
+            else:
+                new_registry[kid] = kernel_obj
+        self.kernel_registry = new_registry
```
HyeockJinKim: I hope there isn't logic that self-checks and deletes kernel IDs without receiving input for the kernel ID.

src/ai/backend/agent/agent.py
```diff
@@ -1342,6 +1333,132 @@ async def _handle_clean_event(self, ev: ContainerLifecycleEvent) -> None:
...
+        log.info(
+            "purging {} containers for kernels: {}",
+            len(containers_to_destroy),
+            ", ".join(map(str, kernel_ids)),
+        )
+
+        async def purge(kernel_container_pair: KernelIdContainerPair) -> KernelIdContainerPair:
+            await self._purge_container(kernel_container_pair)
+            await self._clean_kernel_object(kernel_container_pair[0])
+            return kernel_container_pair
```
HyeockJinKim: Don't use closure and make private method.

src/ai/backend/agent/agent.py
```diff
@@ -1342,6 +1333,132 @@ async def _handle_clean_event(self, ev: ContainerLifecycleEvent) -> None:
...
+
+        tasks = [purge(pair) for pair in containers_to_destroy]
+        results = await asyncio.gather(*tasks, return_exceptions=True)
+
+        purged_kernel_container_pairs: list[KernelIdContainerPair] = []
+        for result in results:
+            if isinstance(result, BaseException):
+                # _purge_container() and _clean_kernel_object() already logged the error
+                continue
```
HyeockJinKim: BaseException seems like it should be propagated upwards, and shouldn't only Exception be propagated?

src/ai/backend/agent/agent.py
```diff
@@ -1342,6 +1333,132 @@ async def _handle_clean_event(self, ev: ContainerLifecycleEvent) -> None:
...
+        try:
+            await self.clean_kernel(kernel_id, container.id, restarting=False)
+        except Exception as e:
+            log.exception(
+                "failed to clean kernel (kernel:{}, container:{}): {}",
+                kernel_id,
+                container.human_readable_id,
+                repr(e),
+            )
+            raise
```
HyeockJinKim: If it operates sequentially and raises an error, it seems like I only need to put one try block.

src/ai/backend/agent/agent.py
```diff
@@ -1342,6 +1333,132 @@ async def _handle_clean_event(self, ev: ContainerLifecycleEvent) -> None:
...
+            )
+        except Exception as e:
+            log.exception(
+                "failed to clean kernel object (kernel:{0}): {1}",
+                kernel_id,
+                repr(e),
+            )
+            raise
+        else:
+            log.info("cleaned kernel object (kernel:{})", kernel_id)
```
HyeockJinKim: Shouldn't deleting from the kernel registry be placed in a Finally block?

src/ai/backend/agent/agent.py
HyeockJinKim: And if an error occurs during purge_container execution, it seems that cleaning the kernel object won't even be executed.

PR #4717, opened 2025-06-16
feat(BA-1550): Agent heartbeat handler purges kernels

### LINKED ISSUE #4665: Handle dangling containers
Agents sometimes fail to clean up containers which leads to mismatch between DB and real containers. Let managers and agents resolve such dangling container issue

JIRA Issue: BA-1550

fregataa:
resolves #4665 (BA-1550)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/event_dispatcher/handlers/agent.py
+++ b/src/ai/backend/manager/event_dispatcher/handlers/agent.py
@@ -175,16 +175,28 @@ class AgentEventHandler:
         kernels_to_clean = self._filter_kernels_to_clean(event.active_kernels, kernel_should_alive)
 
         log.info(
-            "agent@{0} heartbeat: {1} dangling containers, {2} dangling kernel registries",
+            "agent@{0} heartbeat: Detected {1} dangling containers, {2} dangling kernel registries",
             event.agent_id,
             len(containers_to_purge),
             len(kernels_to_clean),
         )
-        await self._registry.purge_containers(
-            event.agent_id,
-            containers_to_purge,
-        )
-        await self._registry.drop_kernel_registry(
-            event.agent_id,
-            kernels_to_clean,
-        )
+        if containers_to_purge:
+            log.warning(
+                "agent@{0} heartbeat: Purging containers: {1}",
+                event.agent_id,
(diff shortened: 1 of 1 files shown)


src/ai/backend/agent/observer/heartbeat.py
```diff
@@ -33,34 +31,28 @@ def __init__(
...
+                container.status,
+            )
+            for kernel_id, container in containers
         ]
         await self._event_producer.anycast_event(
             AgentStatusHeartbeat(
                 self._agent.id,
                 container_data,
+                {
+                    kid: ContainerId(kernel_obj.container_id)
```
HyeockJinKim: Is the kernel ID used as the container ID value?

HyeockJinKim: It seems like we need to send the container ID value together with the kernel ID value.

src/ai/backend/common/events/event_types/agent/anycast.py
```diff
@@ -188,20 +186,21 @@ def from_dict(cls, data: dict[str, str]) -> Self:
...
             ContainerStatus(data["status"]),
-            data["kernel_lifecycle_status"],
         )
 
 
 @dataclass
 class AgentStatusHeartbeat(AgentOperationEvent):
     agent_id: AgentId
     containers: list[ContainerStatusData]
+    kernel_registry: dict[KernelId, Optional[ContainerId]]
```
HyeockJinKim: How about naming them `alive_containers` and `alive_kernels`?

src/ai/backend/manager/models/kernel.py
```diff
@@ -109,6 +117,13 @@ class KernelStatus(CIStrEnum):
     ERROR = "ERROR"
     CANCELLED = "CANCELLED"
 
+    @classmethod
+    def with_containers(cls) -> set[KernelStatus]:
+        return {
+            cls.CREATING,
+            cls.RUNNING,
+        }
```
HyeockJinKim: "with" gives the feeling that it might be used as a context manager. Please give it a different name.

PR #4719, opened 2025-06-17
fix(BA-1591): License config not included in PluginConfig

### LINKED ISSUE #4718: License config not recognized in Manager
## Summary  

After the Pydantic migration, a report came in that the activator is having trouble connecting to the license server from the manager process.

Investigate and resolve the cause of the problem.

jopemachine:
resolves #4718 (BA-1591)

### DIFF
--- a/src/ai/backend/manager/config/unified.py
+++ b/src/ai/backend/manager/config/unified.py
@@ -1829,6 +1829,23 @@ class ServiceDiscoveryConfig(BaseModel):
     )
 
 
+class LicenseConfig(BaseModel):
+    addr: HostPortPairModel = Field(
+        description="""
+        Address of the license server.
+        """,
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/config/unified.py
```diff
@@ -1829,6 +1829,23 @@ class ServiceDiscoveryConfig(BaseModel):
     )
 
 
+class LicenseConfig(BaseModel):
+    addr: HostPortPairModel = Field(
+        description="""
+        Address of the license server.
+        """,
+        examples=["license.example.com:443", "localhost:443"],
```
jopemachine, line 1837: I didn't know the default port number, so I used the value that Copilot auto-completed.
If there's a default value that's actually used, it would be good to include it.

HyeockJinKim (approved):

PR #4729, opened 2025-06-17
feat(BA-1594): Add session creation test using various options

### LINKED ISSUE #4728: Add Session test (Create Session with various options)
Let's investigate each of the options below and add tests.

- `batch_timeout`
- `bootstrap_script`
- `tag`
- `envs`
- `max_wait`

jopemachine:
resolves #4728 (BA-1594)

This PR adds tests for the options below.

- `batch_timeout`
- `bootstrap_script`
- `max_wait`


src/ai/backend/test/contexts/client_session.py
```diff
@@ -1,8 +1,14 @@
...
 class ClientSessionContext(BaseTestContext[AsyncSession]):
     @classmethod
     def name(cls) -> ContextName:
         return ContextName.CLIENT_SESSION
+
+
+class ClientSyncSessionContext(BaseTestContext[Session]):
+    @classmethod
+    def name(cls) -> ContextName:
+        return ContextName.CLIENT_SESSION
```
HyeockJinKim: Wouldn't it be okay to just receive both cases as BaseSession? (In that case, it seems like async and sync can both be unwrapped from the top.)

src/ai/backend/test/tester/dependency.py
```diff
@@ -87,6 +92,14 @@ class ClusterDep(BaseDependencyModel):
     )
 
 
+class BootstrapScriptDep(BaseDependencyModel):
+    bootstrap_script: Optional[str] = Field(
+        default=None,
+        description="The bootstrap script to run in the session.",
+        examples=["echo 'Bootstrapping...'"],
+    )
```
HyeockJinKim: I would appreciate it if you could provide a more detailed description of where it is being used.

src/ai/backend/test/testcases/session/creation_failure_schedule_timeout.py
```diff
@@ -0,0 +1,33 @@
...
+
+        resp: ComputeSession = client_session.ComputeSession.get_or_create(
+            image_dep.name,
+            architecture=image_dep.architecture,
+            # TODO: Investigate why max_wait does not function properly when session_name is provided.
+            # name=session_name,
+            enqueue_only=False,
+            max_wait=5,
+            resources={
+                "cpu": sys.maxsize  # Intentionally large value to trigger timeout
```
HyeockJinKim: Please use different values, as system-related sizes can yield unexpected results if the execution environment and target environment are different.

src/ai/backend/test/templates/session/utils.py
```diff
@@ -11,6 +11,7 @@ async def verify_session_events(
     failure_events: set[str],
     *,
     expected_termination_reason: Optional[str] = None,
+    expected_failure_reason: Optional[str] = None,
 ) -> None:
```
HyeockJinKim, line 15: I'm starting to think that the number of factors is increasing too much.

src/ai/backend/test/templates/session/interactive_session.py
```diff
@@ -24,17 +25,32 @@
...
+        params: dict[str, Any] = {
+            "image": image_dep.name,
+            "resources": session_dep.resources,
+            "type_": "interactive",
+            "name": session_name,
+            "cluster_mode": cluster_dep.cluster_mode,
+            "cluster_size": cluster_dep.cluster_size,
+        }
+        params.update(self._extra_session_params())
+        return params
```
HyeockJinKim, line 46: I'm thinking that currently the clientSDK receives arguments directly as parameters, but what if we pass this part as a v2-like form and pass a pydantic object?

src/ai/backend/test/testcases/session/creation_failure_command_timeout.py
```diff
@@ -0,0 +1,47 @@
...
+            name=session_name,
+            type_="batch",
+            startup_command="sleep 20",
+            batch_timeout=3,
+            resources=session_dep.resources,
+            cluster_mode=ClusterMode.SINGLE_NODE,
+            cluster_size=1,
+        )
+
+        await listener
```
HyeockJinKim: Isn't it possible that it might fail depending on the sse_dep.timeout value?

HyeockJinKim: It seems like the timeout value and the startup_command value here should be set as fixed values. However, sleep 20 seems too long, so how about setting it to around 10 seconds and setting the timeout to around 15~20 seconds?

PR #4740, opened 2025-06-18
fix(BA-1604): Close code runner properly

### LINKED ISSUE #4739: Close code runner properly
The `close()` method in the agent code runner is currently encountering issues, as indicated by the error log stating, “Task was destroyed but it is pending.” This method should terminate all associated tasks properly.

fregataa:
resolves #4739 (BA-1604)


src/ai/backend/agent/kernel.py
```diff
@@ -659,20 +666,30 @@ async def get_repl_in_addr(self) -> str:
     async def get_repl_out_addr(self) -> str:
         raise NotImplementedError
 
+    async def _close_task(self, task: Optional[asyncio.Task]) -> None:
```
rapsealk: I'd suggest to separate this method, as it can be used generally for other cases.

src/ai/backend/agent/kernel.py
```diff
@@ -681,6 +674,24 @@ async def close(self) -> None:
...
 
+    async def _create_tasks(self) -> None:
+        # close the previous task if any
+        await self._close_tasks()
+
+        loop = asyncio.get_running_loop()
+        self.status_task = loop.create_task(self.ping_status())
+        self.read_task = loop.create_task(self.read_output())
+        if self.exec_timeout > 0:
+            self.watchdog_task = loop.create_task(self.watchdog())
```
HyeockJinKim, line 685: If a task is not created, is it okay to put a None value to avoid problems?

src/ai/backend/agent/kernel.py
```diff
@@ -681,6 +674,24 @@ async def close(self) -> None:
         except Exception:
             log.exception("AbstractCodeRunner.close(): unexpected error")
 
+    async def _create_tasks(self) -> None:
+        # close the previous task if any
+        await self._close_tasks()
```
HyeockJinKim, line 679: Should this be called?

src/ai/backend/agent/kernel.py
```diff
@@ -585,14 +592,8 @@ def __init__(
...
     async def __ainit__(self) -> None:
-        loop = current_loop()
         await self._get_socket_pair()
-        self.status_task = loop.create_task(self.ping_status())
-        self.read_task = loop.create_task(self.read_output())
-        if self.exec_timeout > 0:
-            self.watchdog_task = loop.create_task(self.watchdog())
-        else:
-            self.watchdog_task = None
+        await self._create_tasks()
```
HyeockJinKim, line 596: Creating the task itself should probably be put in __init__.

src/ai/backend/common/asyncio.py
```diff
@@ -39,6 +40,22 @@ async def cancel_tasks(
...
+    if task is None:
+        return
+    being_canceled = task.cancel()
+    if not being_canceled:
+        # the task is already cancelled or done
+        return
+    await asyncio.sleep(0)  # yield to the event loop
+    if task.done():
+        return
+    await task
```
rapsealk: How about this approach? (ref: https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel)
```suggestion
    if task is None or task.done():
        return
    task.cancel()
    try:
        await task
    except asyncio.CancelledError:
        pass
    except Exception as e:
        log.error(...)
```

fregataa: It is a good example but using the return value of `cancel()` method can keep it simple

src/ai/backend/agent/kernel.py
fregataa: Duplicate tasks are created and assigned to the `..._task` attributes, but it is tricky to find where it is. If I just create a new task and assign them, the remaining tasks become dangling.
So I'd like to ensure that there is no dangling tasks

src/ai/backend/agent/kernel.py
fregataa: I left `_create_tasks()` as an awaitable to call `_close_tasks()` inside

HyeockJinKim (approved):

PR #4744, opened 2025-06-19
fix: Add EndpointWrapperTemplate (WIP)

jopemachine:
resolves #NNN (BA-MMM)

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/test/templates/model_service/endpoint.py
```diff
@@ -0,0 +1,173 @@
...
+            "resource_opts": {},
+            "domain_name": domain_dep.name,
+            "group_name": group_dep.name,
+            "scaling_group": scaling_group_dep.name,
+            "model_mount_destination": model_service_dep.model_mount_destination,
+            "model_definition_path": model_service_dep.model_definition_path,
+            "cluster_mode": cluster_dep.cluster_mode,
+            "cluster_size": cluster_dep.cluster_size,
+            # TODO: Make `envs` required.
+            "envs": {},
```
HyeockJinKim: Wouldn't it be worth putting even a suitable value into the `envs` to see if the Env is set?

src/ai/backend/test/templates/model_service/endpoint.py
```diff
@@ -0,0 +1,173 @@
...
+        }
+        params.update(self._extra_service_params())
+        return params
+
+    def _extra_service_params(self) -> dict[str, Any]:
+        return {}
```
HyeockJinKim: To induce inheritance, it would be good to declare it as abstractmethod here + raise NotImplemented.

src/ai/backend/test/tester/dependency.py
```diff
@@ -80,28 +98,25 @@ class SSEDep(BaseDependencyModel):
 class ClusterDep(BaseDependencyModel):
     # By default, testing is conducted for both single-node and multi-node setups through parametrization,
     # But we'd like to have left room for manually injecting values.
-    cluster_mode: Optional[str] = Field(
-        default=None,
+    cluster_mode: str = Field(
         description="The cluster mode for the session.",
-        examples=["single_node", "multi_node"],
```
HyeockJinKim: Let's define it as a StrEnum. Or, if ClusterMode is already in common, it seems like we could use the strenum in common.

src/ai/backend/test/templates/model_service/endpoint.py
jopemachine: I'll proceed with it as a follow-up task.

src/ai/backend/test/templates/model_service/endpoint.py
```diff
@@ -0,0 +1,165 @@
...
+                            ready_session_cnt += 1
+
+                    if ready_session_cnt >= model_service_dep.replicas:
+                        break
+
+                    await asyncio.sleep(1)
+
+            await asyncio.wait_for(
+                _wait_until_all_inference_sessions_ready(), timeout=_ENDPOINT_CREATION_TIMEOUT
+            )
```
HyeockJinKim: I think it's better not to use closures if possible.
Not only that, but the test case is getting too long now, and that's also a concern.

src/ai/backend/test/templates/model_service/endpoint.py
```diff
@@ -0,0 +1,178 @@
...
+                    f"Actual mounted vfolder: {session_info['vfolder_mounts']}, "
+                    f"session_id: {session_id}"
+                )
+                if session_info["status"] == "RUNNING":
+                    ready_session_cnt += 1
+
+            if ready_session_cnt >= replicas:
+                break
+
+            await asyncio.sleep(1)
```
HyeockJinKim, line 113: It seems like it's working this way because the current SSE model serving doesn't send events. Let's improve and change it later to deliver messages via SSE.

HyeockJinKim (approved):

PR #4750, opened 2025-06-19
fix(BA-1606): Fix session id based query available in SDK

### LINKED ISSUE #4747: Fix compute session sdk allow query based on session id
## Summary  

Currently, the API internals accept both the name and ID of the session, but the SDK only accepts the name. So we'll change this to allow the methods in the SDK to be used with the session ID as well

seedspirit:
resolves #4747 (BA-1606)

### DIFF
--- a/src/ai/backend/client/func/session.py
+++ b/src/ai/backend/client/func/session.py
@@ -578,6 +578,12 @@ class ComputeSession(BaseFunction):
                 identity_params["owner_access_key"] = self.owner_access_key
         return identity_params
 
+    def get_session_identifier(self) -> str:
+        if self.id is not None:
+            return str(self.id)
+        assert self.name is not None, "Either session name or id must be set."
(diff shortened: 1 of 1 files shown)


src/ai/backend/client/func/session.py
```diff
@@ -578,6 +578,17 @@ def get_session_identity_params(self) -> Mapping[str, str]:
...
+    def get_session_identifier(self) -> str:
+        """
+        Returns the session identifier, preferring the session ID if available,
+        otherwise falling back to the session name.
+        """
+        if self.id:
+            return str(self.id)
+        if self.name:
+            return self.name
+        raise ValueError("Session must have either an ID or a name.")
```
HyeockJinKim: How about setting it as a property?

seedspirit: refactored function as property

PR #4753, opened 2025-06-19
fix(BA-1609): Skip gathering metric of non-existent process

### LINKED ISSUE #4752: psutil warning log spamming
Spamming of warning log saying “psutil cannot found process XXXX” often happens. This issue can arise for several reasons:

1. It may occur naturally because Backend.AI Agent does not receive the process termination event via OS-guaranteed ways such as SIGCHLD or pidfd monitoring.
1. The Agent may attempt to gather utilization metrics for a process that has already terminated asynchronously as it scans the process list periodically based on the previous scanned result.

We need to simply skip extraneous error logs in this case, **_explicitly removing_** the stale entries in the Agent’s psutil scan results.

JIRA Issue: BA-1609

fregataa:
resolves #4752 (BA-1609)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -1171,13 +1171,11 @@ class AbstractAgent(
         if self.local_config["debug"]["log-stats"]:
             log.debug("collecting container statistics")
         try:
-            container_ids = []
-            containers = await self.enumerate_containers()
-            for kernel_id, container in containers:
-                kernel_obj = self.kernel_registry.get(kernel_id)
-                if kernel_obj is None or not kernel_obj.stats_enabled:
+            container_ids: list[ContainerId] = []
+            for kernel_obj in [*self.kernel_registry.values()]:
+                if not kernel_obj.stats_enabled or kernel_obj.container_id is None:
                     continue
-                container_ids.append(container.id)
+                container_ids.append(ContainerId(kernel_obj.container_id))
                 await self.stat_ctx.collect_container_stat(container_ids)
         except asyncio.CancelledError:
             pass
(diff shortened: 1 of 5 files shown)


src/ai/backend/agent/stats.py
```diff
@@ -643,26 +670,16 @@ async def collect_per_container_process_stat(
...
-                        log.debug(
-                            "collect_per_container_process_stat(): cannot found container {}", cid
-                        )
-                    else:
-                        for unused_pid in unused_pids:
-                            log.debug("removing pid_metric for {}: {}", cid, unused_pid)
-                            self.process_metrics[cid].pop(unused_pid, None)
-                    for pid in pids:
-                        pid_map[pid] = cid
-
```
fregataa, line 665: This part causes continuous queries to non-existent PIDs, resulting in log spamming

src/ai/backend/agent/agent.py
```diff
@@ -1171,15 +1171,11 @@ async def collect_container_stat(self, interval: float):
...
         try:
-            container_ids = []
-            async with self.registry_lock:
-                for kernel_id, kernel_obj in [*self.kernel_registry.items()]:
-                    if (
-                        not kernel_obj.stats_enabled
-                        or kernel_obj.state != KernelLifecycleStatus.RUNNING
-                    ):
-                        continue
-                    container_ids.append(kernel_obj["container_id"])
```
HyeockJinKim, line 1182: Why was registry_lock omitted?

fregataa: The lock is unnecessary because the collection task runs regularly and this task copy the items of kernel registry

HyeockJinKim (approved):
If we also consider that no-gil will be applied later, it seems like we need to acquire a lock when reading, even if it means narrowing the lock range, but it doesn't seem to have any impact for now, so I will merge it.

PR #4757, opened 2025-06-20
feat(BA-1450): Add installed field to image node query schema

### LINKED ISSUE #4514: Add `installed` field to ImageNode GQL type
`Image` GQL type has `installed` field which represents the image is installed in any agent(s). Let’s add the field to `ImageNode` type

JIRA Issue: BA-1450

fregataa:
resolves #4514 (BA-1450)

- [ ] Test case(s) to:
- [ ] Documentation

----
📚 Documentation preview 📚: https://sorna--4757.org.readthedocs.build/en/4757/

----
📚 Documentation preview 📚: https://sorna-ko--4757.org.readthedocs.build/ko/4757/


docs/manager/graphql-reference/schema.graphql
```diff
@@ -598,6 +598,9 @@ type ImageNode implements Node {
   Added in 25.3.0. One of ['read_attribute', 'update_attribute', 'create_container', 'forget_image'].
   """
   permissions: [ImagePermissionValueField]
+
+  """Added in 25.11.0. Indicates if the image is installed on any Agent."""
+  installed: Boolean
```
jopemachine, line 603: How about returning a list of AgentNode objects instead of exposing a boolean field like installed?
This would be useful on the frontend, not only to check whether the image is installed on one or more agents, but also to know exactly which agents the image is installed on.

jopemachine (changes requested):

docs/manager/graphql-reference/schema.graphql
jopemachine: If we should add `installed` in its current form, please add `installed` to the `queryfilter` as well.

src/ai/backend/manager/models/gql_models/image.py
```diff
@@ -464,6 +466,36 @@ class Meta:
         ImagePermissionValueField,
         description=f"Added in 25.3.0. One of {[val.value for val in ImagePermission]}.",
     )
+    installed = graphene.Boolean(
+        description="Added in 25.11.0. Indicates if the image is installed on any Agent."
+    )
+
+    @property
+    def _full_name(self) -> str:
+        return f"{self.registry}/{self.namespace}:{self.tag}"
```
fregataa: @jopemachine Is this correct to guess the full name of an image?

docs/manager/graphql-reference/schema.graphql
fregataa: The client part requires only "installed" field. If we want to add a field of a list of agents, then we should resolve it as another Agent GQL query or Agent Relay node.
I want to keep the work small and we should discuss how we should add the field.
And speaking of Image queries, while the `installed` field always returns boolean value `installed_agent` field returns null if the client does not have permission (superadmin or `hide-agent` config)

And the "installed" field is not able to filter. So I won't add it into query filter

src/ai/backend/manager/models/gql_models/image.py
jopemachine: To avoid such recurring these questions, I think it would be a good idea to add a canonical field like the one below.
@HyeockJinKim 

```py
    canonical = graphene.String()

    def resolve_canonical(self, info: graphene.ResolveInfo) -> str:
        join = functools.partial(join_non_empty, sep="/")
        return f"{join(self.registry, self.name)}:{self.tag}"
```

jopemachine: If we're hesitant to add `canonical` as a field due to the additional overhead it may cause, it could be also good idea to simply add it as a function or property instead.

jopemachine: For now, let's just change full_name to _canonical.

```py
    def _canonical(self, info: graphene.ResolveInfo) -> str:
        join = functools.partial(join_non_empty, sep="/")
        return f"{join(self.registry, self.name)}:{self.tag}"
```

HyeockJinKim: I think I've seen the function that creates canonical in several places, and it seems to be fragmenting?
How about creating a class that takes image registry, name, and tag, and providing canonical as a method?

jopemachine: > I think I've seen the function that creates canonical in several places, and it seems to be fragmenting? How about creating a class that takes image registry, name, and tag, and providing canonical as a method?

Although the code doesn't seem simple enough, `ImageRef` can serve that role.

```py
    def _canonical(
        self,
    ):
        return ImageRef.from_image_str(
            self.humanized_name,
            self.project,
            self.registry,
            architecture=self.architecture,
            is_local=self.is_local,
        ).canonical
```

jopemachine (approved):

nowgnuesLee (approved):
LGTM

jopemachine:
Please add missing news fragment @fregataa

src/ai/backend/manager/models/gql_models/image.py
```diff
@@ -464,6 +466,45 @@ class Meta:
...
+        result = cast(list[set[bytes]], result)
+        return [{AgentId(agent_id.decode()) for agent_id in agents} for agents in result]
+
+    async def resolve_installed(self, info: graphene.ResolveInfo) -> bool:
+        graph_ctx: GraphQueryContext = info.context
+        loader = graph_ctx.dataloader_manager.get_loader_by_func(
+            graph_ctx, self._batch_load_installed_agents
+        )
+        agent_ids = await loader.load(self._canonical)
+        print("agent_ids!", agent_ids)
```
fregataa: @jopemachine oops..

PR #4766, opened 2025-06-22
feat(BA-1617): Agent heartbeat handler query Kernel ids

### LINKED ISSUE #4765: Do not query agent records in agent status heartbeat handler
Do not query agent records in agent status heartbeat handler

Because agent id can change, and we do not want to clean living kernels whose agent id changed

fregataa:
resolves #4765 (BA-1617)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/event_dispatcher/handlers/agent.py
+++ b/src/ai/backend/manager/event_dispatcher/handlers/agent.py
@@ -24,7 +24,11 @@ from ai.backend.manager.errors.exceptions import InstanceNotFound
 from ai.backend.manager.registry import AgentRegistry
 
 from ...models.agent import AgentStatus, agents
-from ...models.kernel import ConditionMerger, KernelRow, KernelStatus, by_agent_id, by_status
+from ...models.kernel import (
+    ConditionMerger,
+    KernelRow,
+    by_kernel_ids,
+)
 from ...models.utils import (
     ExtendedAsyncSAEngine,
 )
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/event_dispatcher/handlers/agent.py
```diff
@@ -156,25 +160,32 @@ async def handle_agent_container_heartbeat(
...
-        agent_condition = by_agent_id(event.agent_id, ConditionMerger.AND)
+        # Do not query Agent id from the event because Agent id can change
+        # during the lifetime of the agent, e.g. when it is restarted.
+        all_kernel_ids: set[KernelId] = {k.kernel_id for k in event.active_kernels} | {
+            c.kernel_id for c in event.active_containers
+        }
+        kernel_condition = by_kernel_ids(
+            all_kernel_ids,
+            ConditionMerger.AND,
+        )
```
HyeockJinKim, line 171: I didn't quite see properly last week, but it seems a little strange to pass the conditions for how to do the merging, such as `AND` and `OR`, together.
It seems correct that the `by_kernel_ids` function only receives `kernel_ids`, and it seems more appropriate to handle whether to connect with AND or OR in another part.

HyeockJinKim: It seems like there are more chances to make mistakes if we embed `ConditionMerger` and append it from another location.

src/ai/backend/manager/event_dispatcher/handlers/agent.py
```diff
@@ -156,25 +160,32 @@ async def handle_agent_container_heartbeat(
...
-                status_condition,
-                agent_condition,
-            ],
+            [kernel_condition],
             db=self._db,
         )
-        kernel_should_alive: set[KernelId] = {kernel_row.id for kernel_row in kernel_rows}
+        kernel_should_alive: set[KernelId] = {
+            kernel_row.id for kernel_row in kernel_rows if kernel_row.status.have_container()
+        }
```
HyeockJinKim, line 178: Wouldn't it be reasonable to include everything as alive, as long as it's not in a terminated state? (ex: terminated or canceled)

HyeockJinKim (approved):

PR #4767, opened 2025-06-22
refactor(BA-1616): Apply simple model query patterns

### LINKED ISSUE #4764: Refactor and align DB query codes
`ai.backend.manager.models.group`, `ai.backend.manager.models.user`, `ai.backend.manager.models.kernel` all have different patterns of query functions.

Let’s align them

- Deprecate `ai.backend.manager.models.types.QueryCondition` and any `ConditionMerger` enum types
- Assume such conditions be merged with “and” condition only

fregataa:
resolves #4764 (BA-1616)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/models/session.py
+++ b/src/ai/backend/manager/models/session.py
@@ -12,7 +12,6 @@ from typing import (
     TYPE_CHECKING,
     Any,
     AsyncIterator,
-    Callable,
     Final,
     List,
     Optional,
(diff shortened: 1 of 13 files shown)


src/ai/backend/manager/models/session.py
```diff
@@ -1386,78 +1398,60 @@ class SessionQueryConditions:
...
-        new_cond = SessionRow.user_uuid == user_id
-        return sa.and_(cond, new_cond) if cond is not None else new_cond
+def by_user_id(user_id: UUID) -> QueryCondition:
+    def _by_user_id(
+        query_stmt: sa.sql.Select,
+    ) -> sa.sql.Select:
+        return query_stmt.where(SessionRow.user_uuid == user_id)
 
-    return _and_user_id
+    return _by_user_id
```
HyeockJinKim, line 1484: 👍

HyeockJinKim (commented):
It seems fine with the AND condition, but are we not particularly considering the OR case? (Due to reasons such as not using it now?)

fregataa:
> It seems fine with the AND condition, but are we not particularly considering the OR case? (Due to reasons such as not using it now?)

I think any condition that needs to combine with OR should be in a single query condition, rather than chain them

HyeockJinKim (approved):

PR #4782, opened 2025-06-23
feat(BA-1622): Implement `UserResourcePolicy` CRUD SDK functions

### LINKED ISSUE #4778: Implement `UserResourcePolicy` SDK
## Motivation  

- Currently, there is an SDK for `KeypairResourcePolicy`, but there is no SDK to create, modify, or delete `UserResourcePolicy`. Add the SDK to allow SDK control over information such as limiting the number of VFolders a user can have, or limiting the number of sessions that can be assigned to a Model Service.

seedspirit:
resolves #4778 (BA-1622)

### DIFF
--- a/src/ai/backend/client/func/user_resource_policy.py
+++ b/src/ai/backend/client/func/user_resource_policy.py
@@ -1,0 +1,178 @@
+from typing import Optional, Sequence
+
+from ..output.fields import user_resource_policy_fields
+from ..output.types import FieldSpec
+from ..session import api_session
+from ..utils import dedent as _d
+from .base import BaseFunction, api_function
+
+_default_list_fields = (
+    user_resource_policy_fields["name"],
+    user_resource_policy_fields["max_vfolder_count"],
+    user_resource_policy_fields["max_quota_scope_size"],
+    user_resource_policy_fields["max_session_count_per_model_session"],
+    user_resource_policy_fields["max_customized_image_count"],
+)
+
+_default_detail_fields = (
+    user_resource_policy_fields["name"],
+    user_resource_policy_fields["max_vfolder_count"],
+    user_resource_policy_fields["max_quota_scope_size"],
+    user_resource_policy_fields["max_session_count_per_model_session"],
+    user_resource_policy_fields["max_customized_image_count"],
+)
+
+
+class UserResourcePolicy(BaseFunction):
+    """
+    Provides interactions with user resource policy.
+    """
+
+    def __init__(self, name: Optional[str]) -> None:
+        super().__init__()
+        self.name = name
+
+    @api_function
+    @classmethod
+    async def create(
+        cls,
+        name: str,
+        *,
+        max_vfolder_count: int,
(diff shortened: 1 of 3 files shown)


src/ai/backend/client/func/user_resource_policy.py
```diff
@@ -0,0 +1,178 @@
...
+    user_resource_policy_fields["max_customized_image_count"],
+)
+
+_default_detail_fields = (
+    user_resource_policy_fields["name"],
+    user_resource_policy_fields["max_vfolder_count"],
+    user_resource_policy_fields["max_quota_scope_size"],
+    user_resource_policy_fields["max_session_count_per_model_session"],
+    user_resource_policy_fields["max_customized_image_count"],
+)
```
HyeockJinKim, line 24: Let's create new client behaviors with pydantic.

src/ai/backend/client/func/user_resource_policy.py
```diff
@@ -0,0 +1,178 @@
...
+    @classmethod
+    async def create(
+        cls,
+        name: str,
+        *,
+        max_vfolder_count: int,
+        max_quota_scope_size: int,
+        max_session_count_per_model_session: int,
+        max_customized_image_count: int,
+    ) -> dict:
```
jopemachine, line 48: I know that the current SDK return types are all in the form of `dict`, but it would be better if the new SDKs use Pydantic types for their return values.

src/ai/backend/client/func/user_resource_policy.py
```diff
@@ -0,0 +1,178 @@
+from typing import Optional, Sequence
```
jopemachine: Please import `Sequence` from `collections.abc` (Refer to our code convention)

src/ai/backend/client/func/user_resource_policy.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+class UserResourcePolicy(BaseFunction):
+    """
+    Provides interactions with user resource policy.
+    """
+
+    def __init__(self, name: Optional[str]) -> None:
+        super().__init__()
+        self.name = name
```
jopemachine:
```suggestion
        self._name = name
```

src/ai/backend/client/func/user_resource_policy.py
```diff
@@ -0,0 +1,178 @@
...
+
+        result = await api_session.get().Admin._query(q, variables)
+        return result["modify_user_resource_policy"]
+
+    @api_function
+    async def delete(self) -> dict:
+        """
+        Delete an existing user resource policy.
+
+        :param name: Name of the user resource policy to delete.
```
jopemachine: This does not receive any params

src/ai/backend/client/func/user_resource_policy.py
seedspirit: Discussed with @HyeockJinKim that defining the request response for the current client using pydantic will be applied at once after other discussions

PR #4799, opened 2025-06-24
fix(BA-1608): Exclude None values when creating model service

### LINKED ISSUE #4751: Make `Service.create` SDK method's `envs` argument required
## Summary  

The current `Service.create` SDK method has the `envs` parameter marked as optional, so it can be omitted. However, an error actually occurs if the `envs` parameter is not provided.

Let's make the `envs` parameter required or assign it a default value.

seedspirit:
resolves #4751 (BA-1608)

### DIFF
--- a/src/ai/backend/client/func/service.py
+++ b/src/ai/backend/client/func/service.py
@@ -189,8 +189,9 @@ class Service(BaseFunction):
         }
         if model_version:
             model_config["model_version"] = model_version
+        model_config = {k: v for k, v in model_config.items() if v is not None}
         rqst = Request("POST", "/services")
-        rqst.set_json({
+        rqst_body = {
             "name": service_name,
             "replicas": initial_session_count,
             "image": image,
@@ -206,7 +207,9 @@ class Service(BaseFunction):
             "open_to_public": expose_to_public,
             "runtime_variant": runtime_variant,
             "config": model_config,
-        })
+        }
+        rqst_body = {k: v for k, v in rqst_body.items() if v is not None}
+        rqst.set_json(rqst_body)
         async with rqst.fetch() as resp:
             body = await resp.json()
             return {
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/api/service.py
```diff
@@ -303,12 +303,12 @@ class ServiceConfigModel(LegacyBaseRequestModel):
...
         ),
-        default={},
+        default_factory=dict,
     )
 
-    environ: Optional[dict[str, str]] = Field(
+    environ: dict[str, str] = Field(
         description="Environment variables to be set inside the inference session",
-        default=None,
+        default_factory=dict,
```
HyeockJinKim: Shouldn't we receive it as none from the API and convert it when passing it to the internal service logic? I don't think we necessarily need to change the API layer.

src/ai/backend/manager/services/model_serving/types.py
```diff
@@ -78,7 +78,7 @@ def to_dict(self) -> dict[str, Any]:
             "model_definition_path": self.model_definition_path,
             "model_mount_destination": self.model_mount_destination,
             "extra_mounts": {key: value.to_dict() for key, value in self.extra_mounts.items()},
-            "environ": self.environ,
+            "environ": self.environ if self.environ is not None else {},
```
seedspirit, line 81: @HyeockJinKim Moved data conversion logic inside dataclass as there are not proper space in `service.py`. Is it okay or is it more clear to put logic explicitly in `service.py`?

HyeockJinKim: This part seems to have no problem to me because the data conversion logic, rather than service logic, is in types.py.

src/ai/backend/client/func/service.py
```diff
@@ -206,7 +207,9 @@ async def create(
             "open_to_public": expose_to_public,
             "runtime_variant": runtime_variant,
             "config": model_config,
-        })
+        }
+        rqst_body = {k: v for k, v in rqst_body.items() if v is not None}
```
HyeockJinKim, line 211: Why is None verification necessary?

PR #4809, opened 2025-06-24
fix(BA-1638): Fix `DeleteEndpointAutoScalingRuleNode` mutation to accept UUID

### LINKED ISSUE #4808: Fix delete endpoint auto scaling rule to accept UUID format
## Summary  

Currently, the `modify_endpoint_auto_scaling_rule_node` mutation accepts both UUID and global ID formats due to fallback logic, while `delete_endpoint_auto_scaling_rule_node` only accepts global ID format and returns 400 error for UUID input. Both operations should consistently handle both ID formats for better API usability.

The delete mutation needs to implement the same fallback mechanism as the update mutation to accept raw UUID when global ID resolution fails.

seedspirit:
resolves #4808 (BA-1638)

### DIFF
--- a/src/ai/backend/manager/models/gql_models/endpoint.py
+++ b/src/ai/backend/manager/models/gql_models/endpoint.py
@@ -533,6 +533,8 @@ class DeleteEndpointAutoScalingRuleNode(graphene.Mutation):
         id: str,
     ) -> Self:
         _, rule_id = AsyncNode.resolve_global_id(info, id)
+        if rule_id is None:
+            rule_id = id
         try:
             _rule_id = RuleId(UUID(rule_id))
         except ValueError:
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/models/gql_models/endpoint.py
```diff
@@ -533,6 +533,8 @@ async def mutate(
         id: str,
     ) -> Self:
         _, rule_id = AsyncNode.resolve_global_id(info, id)
+        if rule_id is None:
+            rule_id = id
```
HyeockJinKim, line 537: Is it okay to consider global_id and rule_id as the same?

seedspirit: The two are technically different, and I think this is more of a workaround to help clients work with `rule_id` (or `row_id`) instead of having to convert to global id every time. 

I think we should align create and update mutations either delete fallback logic or add delete mutation a fallback logic for consistent API usage on the client

PR #4811, opened 2025-06-24
feat(BA-1559): Add auto scaling rule test

### LINKED ISSUE #4676: Add Model Serivce test (Create autoscaling rules)
## Motivation  

- Check autoscaling rule successfully created and replica size are affected
  - scale up rule
  - scale down rule

## 

JIRA Issue: BA-1559

seedspirit:
resolves #4676 (BA-1559)

### DIFF
--- a/src/ai/backend/test/testcases/model_service/scale_by_auto_scaling_rule.py
+++ b/src/ai/backend/test/testcases/model_service/scale_by_auto_scaling_rule.py
@@ -20,7 +20,7 @@ from ai.backend.test.contexts.model_service import (
 from ai.backend.test.contexts.session import ClusterContext
 from ai.backend.test.templates.model_service.utils import ensure_inference_sessions_ready
 from ai.backend.test.templates.template import TestCode
-from ai.backend.test.utils.exceptions import DependencyNotSet, UnexpectedFailure
+from ai.backend.test.utils.exceptions import DependencyNotSet
 
 _METRIC_COLLECTION_TIMEOUT = 120
 _SCALE_TIMEOUT = 60
@@ -122,7 +122,7 @@ class ScaleInByCPUAutoScalingRule(_CPUAutoScalingRuleTestBase):
             raise DependencyNotSet("AutoScalingRuleContext.min_replicas must be set")
 
         if current_replica_count <= min_replicas:
-            raise UnexpectedFailure(
+            raise AssertionError(
                 f"Current replicas must be greater than min_replicas(current: {current_replica_count} / min_replicas: {min_replicas}). Check test configuration or TestSpec"
             )
 
@@ -179,7 +179,7 @@ class ScaleOutByCPUAutoScalingRule(_CPUAutoScalingRuleTestBase):
             raise DependencyNotSet("AutoScalingRuleContext.max_replicas must be set")
 
         if current_replica_count >= max_replicas:
-            raise UnexpectedFailure(
+            raise AssertionError(
                 f"Current replicas must be less than max_replicas(current: {current_replica_count} / max_replicas: {max_replicas}). Check test configuration or TestSpec"
             )
 
--- a/src/ai/backend/test/utils/exceptions.py
+++ b/src/ai/backend/test/utils/exceptions.py
@@ -10,10 +10,3 @@ class UnexpectedSuccess(Exception):
 
     def __init__(self, message="Test succeeded unexpectedly."):
         super().__init__(message)
-
-
-class UnexpectedFailure(Exception):
-    """Raised when a test marked as expected to succeed actually fails."""
-
-    def __init__(self, message="Test failed unexpectedly."):
-        super().__init__(message)


configs/tester/tester.toml
```diff
@@ -51,3 +51,13 @@ group = "default"
 usage_mode = "general"
 permission = "rw"
 cloneable = "False"
+
```
jopemachine: How about moving this under the model service config?

```
[context.model-service.auto-scaling-rule]
```

src/ai/backend/test/templates/vfolder/utils.py
```diff
@@ -0,0 +1,15 @@
+from uuid import UUID
+
+from ai.backend.client.session import AsyncSession
+
+
+async def get_vfolder_id_by_name(
```
jopemachine: How about creating this method to SDK?

```python
await client_session.VFolder(name=vfolder_name).get_id()
```

src/ai/backend/test/testcases/model_service/scale_by_auto_scaling_rule.py
```diff
@@ -0,0 +1,70 @@
...
+        result = await client_session.Service(service_id).info()
+        assert result["replicas"] == max_replicas, (
+            f"Expected replicas count: {max_replicas}, actual: {result['replicas']}"
+        )
+        assert result["desired_session_count"] == max_replicas, (
+            f"Expected desired session count: {max_replicas}, "
+            f"actual: {result['desired_session_count']}"
+        )
+
+    async def _make_cpu_intensive_task_in_service(
```
jopemachine: Function name seems not appropriate because any value can be passed into code.
Either avoid taking code as a parameter or generalize the function name.

I think the former is probably more in line with your intention.

src/ai/backend/test/tester/dependency.py
```diff
@@ -179,6 +179,43 @@ class VFolderDep(BaseDependencyModel):
...
+        description="The name of the metric to monitor for auto-scaling.",
+        examples=["cpu_util", "mem"],
+    )
+    threshold: str = Field(
+        description="The threshold value for triggering auto-scaling.",
+        examples=["0.8", "0.5"],
+    )
+    comparator: str = Field(
+        description="The comparator to use for the threshold.",
+        examples=["GREATER_THAN", "LESS_THAN"],
```
jopemachine: Use `AutoScalingMetricComparator` in `common` module

src/ai/backend/test/templates/model_service/auto_scaling_rule.py
```diff
@@ -0,0 +1,43 @@
...
+                comparator=auto_scaling_rule_dep.comparator,  # type: ignore
+                step_size=auto_scaling_rule_dep.step_size,
+                cooldown_seconds=auto_scaling_rule_dep.cooldown_seconds,
+                min_replicas=auto_scaling_rule_dep.min_replicas,
+                max_replicas=auto_scaling_rule_dep.max_replicas,
+            )
+            with CreatedAutoScalingRuleIDContext.with_current(result.rule_id):
+                yield
+        finally:
+            global_rule_id = to_global_id("delete_endpoint_auto_scaling_rule_node", result.rule_id)
```
jopemachine: `delete_endpoint_auto_scaling_rule_node` is mutation name, not node name

src/ai/backend/test/testcases/model_service/scale_by_auto_scaling_rule.py
```diff
@@ -0,0 +1,70 @@
...
+            client_session, model_service_dep.model_vfolder_name
+        )
+
+        await self._make_cpu_intensive_task_in_service(
+            client_session,
+            service_id,
+            code="while True: pass",
+        )
+
+        await asyncio.sleep(5)  # Give some time for making cpu-intensive environment
```
jopemachine: If possible, please avoid including logic that waits for an arbitrary amount of time.
No matter what arbitrary time is set, it could be too long or too short depending on the situation.
Use polling for now since our model service API lacks SSE API.

src/ai/backend/test/testcases/model_service/scale_by_auto_scaling_rule.py
```diff
@@ -0,0 +1,70 @@
...
+)
+from ai.backend.test.templates.model_service.utils import wait_until_all_inference_sessions_ready
+from ai.backend.test.templates.template import TestCode
+from ai.backend.test.templates.vfolder.utils import get_vfolder_id_by_name
+from ai.backend.test.utils.exceptions import DependencyNotSet
+
+_SCALE_TIMEOUT = 30
+
+
+class ScaleByAutoScalingRules(TestCode):
```
jopemachine: I don't like that this test depends on the `replicas` of model-service.
The test assumes that the "initial" replica count is less than 3, but this could not be true depending on the configuration.

src/ai/backend/test/testcases/model_service/testspecs.py
```diff
@@ -94,6 +106,77 @@
...
                     cluster_size=1,
                 ),
+                ClusterDep(
+                    cluster_mode=ClusterMode.SINGLE_NODE,
+                    cluster_size=3,
+                ),
+                ClusterDep(
+                    cluster_mode=ClusterMode.MULTI_NODE,
+                    cluster_size=3,
+                ),
```
jopemachine, line 422: I think there are too many repeated injections of the same `ClusterDep`.
How about grouing them into a single constant?

src/ai/backend/test/testcases/model_service/testspecs.py
```diff
@@ -94,6 +106,77 @@
...
+            ]
+        },
+    ),
+    "creation_auto_scaling_rule_success": TestSpec(
```
jopemachine: Since `scale_by_auto_scaling_rule_success` already performs the creation test, I think it alone is sufficient.

src/ai/backend/test/contexts/context.py
```diff
@@ -40,8 +40,10 @@ class ContextName(enum.StrEnum):
 
     CREATED_SESSION_META = "created_session_meta"
     CREATED_SESSION_TEMPLATE_ID = "created_session_template_id"
-    CREATED_MODEL_SERVICE_ENDPOINT = "created_model_service_endpoint"
     CREATED_MODEL_SERVICE_TOKEN = "created_model_service_token"
+    CREATED_MODEL_SERVICE_ENDPOINT = "created_model_service_endpoint"
+    AUTO_SCALING_RULE = "auto_scaling_rule"
```
jopemachine: Let's group all the `CREATED_` constants together below.

jopemachine (changes requested):
.

src/ai/backend/test/testcases/model_service/scale_by_auto_scaling_rule.py
seedspirit: I think it can be solved by validating model service config. How about implementing validation logic like this?

```python
@model_validator(mode="after")
def check_replicas(self) -> "ModelServiceDep":
    if self.auto_scaling_rule is None:
        return self

    auto_scaling = self.auto_scaling_rule

    if auto_scaling.max_replicas is not None and self.replicas > auto_scaling.max_replicas:
        raise ValueError(
            f"replicas ({self.replicas}) cannot exceed max_replicas ({auto_scaling.max_replicas})"
        )

    if auto_scaling.min_replicas is not None and self.replicas < auto_scaling.min_replicas:
        raise ValueError(
            f"replicas ({self.replicas}) cannot be less than min_replicas ({auto_scaling.min_replicas})"
        )

    return self
```

seedspirit: Solved this issue by explicitly inject auto scaling config in `TestSpec`

```python
"single_node_single_container_scale_by_cpu_util_auto_scaling_rule_success": TestSpec(
        name="single_node_single_container_scale_by_cpu_util_auto_scaling_rule_success",
        description=textwrap.dedent("""\
        """),
        tags={},
        template=BasicTestTemplate(ScaleOutByCPUAutoScalingRule()).with_wrappers(
            KeypairAuthTemplate, EndpointTemplate, AutoScalingRuleTemplate
        ),
        parametrizes={
            ContextName.AUTO_SCALING_RULE: [
                AutoScalingRuleDep(
                    metric_source=AutoScalingMetricSource.KERNEL,
                    metric_name="cpu_util",
                    threshold="0.01",
                    comparator=AutoScalingMetricComparator.GREATER_THAN_OR_EQUAL,
                    step_size=1,
                    cooldown_seconds=10,
                    min_replicas=1,
                    max_replicas=3,
                ),
            ],
        },
    ),
```

src/ai/backend/test/testcases/model_service/testspecs.py
```diff
@@ -291,6 +508,95 @@
...
+            2. Scale the service using the auto-scaling rule.
+            3. Verify that the service is scaled in successfully.
+            4. Clean up the auto-scaling rule after verification.
+        """),
+        tags={
+            TestTag.MANAGER,
+            TestTag.AGENT,
+            TestTag.MODEL_SERVICE,
+            TestTag.SESSION,
+            TestTag.LONG_RUNNING,
```
jopemachine, line 725: Please append `REQUIRED_MULTI_NODE_MULTI_CONTAINER_CONFIGURATION` tag wherever multi node configuration required

src/ai/backend/test/templates/model_service/auto_scaling_rule.py
```diff
@@ -0,0 +1,45 @@
...
+                step_size=auto_scaling_rule_dep.step_size,
+                cooldown_seconds=auto_scaling_rule_dep.cooldown_seconds,
+                min_replicas=auto_scaling_rule_dep.min_replicas,
+                max_replicas=auto_scaling_rule_dep.max_replicas,
+            )
+            with CreatedAutoScalingRuleIDContext.with_current(result.rule_id):
+                yield
+        finally:
+            global_rule_id = to_global_id(_AUTO_SCALING_RULE_NODE_NAME, result.rule_id)
+            await client_session.ServiceAutoScalingRule(global_rule_id).delete()  # type: ignore
```
jopemachine: Why do we need type ignore? I think we can create and pass id as UUID

src/ai/backend/client/func/service_auto_scaling_rule.py
```diff
@@ -173,6 +173,10 @@ async def update(
             """
         )
         inputs: dict[str, Any] = {}
+        if isinstance(metric_source, AutoScalingMetricSource):
+            metric_source = metric_source.name  # type: ignore
```
jopemachine, line 177: I think type: ignore is being overused.
You can use cast, or rasing exception, and using `type: ignore` should be the last resort.

configs/tester/tester.toml
```diff
@@ -36,6 +36,19 @@ group-name = "model-store"
 model-mount-destination = "/models"
 model-definition-path = "./model-definition.yaml"
 
+[context.model-service-scale-replicas]
+replicas = 2
```
jopemachine: It's confusing because there are two replicas defined in the configuration file.
Let's change one side to initial_replica to make it explicit.

src/ai/backend/test/templates/model_service/scale_replicas.py
```diff
@@ -0,0 +1,52 @@
...
+                    vfolder_id,
+                ),
+                timeout=_SCALE_TIMEOUT,
+            )
+            with ScaledModelServiceReplicaCountContext.with_current(model_service_scale.replicas):
+                yield
+        finally:
+            await client_session.Service(model_service_meta.service_id).scale(
+                to=model_service_dep.replicas
+            )
```
jopemachine, line 63: I think we should call scale and wait until the change actually takes effect.

src/ai/backend/test/utils/exceptions.py
```diff
@@ -10,3 +10,10 @@ class UnexpectedSuccess(Exception):
 
     def __init__(self, message="Test succeeded unexpectedly."):
         super().__init__(message)
+
+
+class UnexpectedFailure(Exception):
```
jopemachine: I don't see how this is different from just using an `AssertionError`.

In the case of `UnexpectedSuccess`, it needs to be specially marked, but in the opposite case, I think it's sufficient to either refine the exception handling or simply throw an AssertionException.

src/ai/backend/test/tester/dependency.py
```diff
@@ -132,6 +137,59 @@ class SessionImagifyDep(BaseDependencyModel):
...
+    def check_replicas(self) -> "ModelServiceAutoScalingRuleDep":
+        if self.max_replicas is not None and self.min_replicas is not None:
+            if self.max_replicas < self.min_replicas:
+                raise ValueError(
+                    f"max_replicas ({self.max_replicas}) must be >= min_replicas ({self.min_replicas})"
+                )
+        return self
+
+
+class ModelServiceScaleReplicasDep(BaseDependencyModel):
```
jopemachine, line 193: How about `ModelServiceScaleDesiredStateDep` to make it explict?
(or renaming `replicas` to `desired_replicas`)

src/ai/backend/client/func/service_auto_scaling_rule.py
seedspirit: I think neither cast nor exception is appropriate in this situation as `metric_source.name` is a name for string enum. For your notice, AutoScalingMetricSource is string enum and it composed like `KERNEL = 'kernel'`. But as graphql client only accept with Enum's name I've applied this logic.

src/ai/backend/test/utils/exceptions.py
seedspirit: I think we need to specially mark cases where the value is set differently than the intent of the test. In the case of an `AssertionError`, since it means fail in the test scenario, shouldn't we mark differently the situation where the test itself is broken by setting a different config than the test author intended?

jopemachine: `Assert` means that a specific condition should be met.
`AssertionError` means that the condition has been violated, and as a result, some tests fail.

While AssertionError is indeed used in places outside of test code, there's no need to create a separate type like UnexpectedFailure just for that. On the contrary, I think our code should be modified to throw the appropriate error instead of throwing an AssertionError.

seedspirit: Removed `UnexpectedFailure` and changed to `AssertionError`

HyeockJinKim:
I will postpone the tester work until after the client SDK modification.

PR #4831, opened 2025-06-25
fix(BA-1664): Name session test using `test_id` in failure cases

### LINKED ISSUE #4830: Inject correct random session name to SessionCreation Failure test cases
## Summary  

Don’t use fixed session name since they’re going to try to reuse current session if they exist.

jopemachine:
resolves #4830 (BA-1664)


src/ai/backend/test/testcases/session/creation_failure_low_resources.py
```diff
@@ -11,7 +12,9 @@ class SessionCreationFailureLowResources(TestCode):
     async def test(self) -> None:
         client_session = ClientSessionContext.current()
         image_dep = ImageContext.current()
-        session_name = "test-session-creation-failure"
+        spec_meta = TestSpecMetaContext.current()
+        test_id = spec_meta.test_id
+        session_name = f"test_session_{str(test_id)}"
```
HyeockJinKim: In the case of failures, how about distinguishing session_name by giving it a prefix like test_failure?

PR #4851, opened 2025-06-26
fix: compute plugin's `cleanup()` method not being called

kyujin-cho:
This PR resolves Backend.AI agent not cleaning up each `ComputePlugin`'s context properly. `ComputePlugin` interface is designed to declare its cleanup sequences by a method named `cleanup()`. We found out that Backend.AI was omitting the required `cleanup()` calls to each initialized Compute Plugins.


src/ai/backend/agent/agent.py
```diff
@@ -933,6 +933,12 @@ async def shutdown(self, stop_signal: signal.Signals) -> None:
         """
         await cancel_tasks(self._ongoing_exec_batch_tasks)
 
+        for _, computer in self.computers.items():
+            try:
+                await computer.instance.cleanup()
+            except Exception as e:
+                log.error("Failed to clean up computer instance: {}", e)
```
HyeockJinKim: You need to use log.exception to show the traceback.

kyujin-cho: Oopsie Claude Code screwed up the code convention

HyeockJinKim (approved):

PR #4875, opened 2025-06-27
fix(BA-1703): Prevent handling auto scaling rules for deleted model services

### LINKED ISSUE #4874: Prevent auto scaling rule creation in deleted model service
## Summary  

We can create an auto scaling rule regardless of the current life cycle of the model service. However, creating an auto scaling rule for a deleted model service is not allowed in user scenarios, and should be prevented

seedspirit:
resolves #4874 (BA-1703)

### DIFF
--- a/src/ai/backend/manager/services/model_serving/services/auto_scaling.py
+++ b/src/ai/backend/manager/services/model_serving/services/auto_scaling.py
@@ -82,10 +82,7 @@ class AutoScalingService:
         async with self._db.begin_session(commit_on_end=True) as db_session:
             try:
                 row = await EndpointRow.get(db_session, action.endpoint_id)
-                if row.lifecycle_stage in (
-                    EndpointLifecycle.DESTROYING,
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/services/model_serving/services/auto_scaling.py
```diff
@@ -168,6 +176,8 @@ async def delete_endpoint_auto_scaling_rule(
                 row = await EndpointAutoScalingRuleRow.get(
                     db_session, action.id, load_endpoint=True
                 )
+                if row.endpoint_row.lifecycle_stage in EndpointLifecycle.inactive_states():
+                    raise EndpointAutoScalingRuleNotFound
```
HyeockJinKim: Should the inactive status affect the deletion?

seedspirit: Excluded validation for delete operations

HyeockJinKim (approved):

PR #4907, opened 2025-07-01
feat: make endpoint GQL query filterable by `created_user`

kyujin-cho:
This PR updates the `Endpoint` GQL class' queryfilter map definition to allow filtering query results by created user's UUID.


src/ai/backend/manager/models/gql_models/endpoint.py
```diff
@@ -647,6 +647,7 @@ class Meta:
         "url": ("endpoints_url", None),
         "lifecycle_stage": (EnumFieldItem("endpoints_lifecycle_stage", EndpointLifecycle), None),
         "open_to_public": ("endpoints_open_to_public", None),
+        "created_user": ("users_id", None),
```
jopemachine: Since there's a `created_user_email`, I think it could cause some confusion.
How about `created_user_id` instead?

kyujin-cho: Good point.

HyeockJinKim (approved):

PR #4909, opened 2025-07-01
feat: Allow passing arguments to `WrapperTestTemplate`

jopemachine:
resolves #NNN (BA-MMM)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/test/testcases/session/testspecs.py
+++ b/src/ai/backend/test/testcases/session/testspecs.py
@@ -105,7 +105,7 @@ BATCH_SESSION_TEST_SPECS = {
         """),
         tags={TestTag.MANAGER, TestTag.AGENT, TestTag.SESSION},
         template=BasicTestTemplate(BatchSessionCreationFailureWrongCommand()).with_wrappers(
-            KeypairAuthTemplate
+            KeypairAuthTemplate(abc="def")
         ),
     ),
     "creation_batch_session_failure_command_timeout": TestSpec(


src/ai/backend/test/templates/auth/keypair.py
```diff
@@ -1,16 +1,20 @@
...
 from ai.backend.test.contexts.client_session import ClientSessionContext
 from ai.backend.test.templates.template import (
+    TestTemplate,
     WrapperTestTemplate,
 )
 
 
 class KeypairAuthTemplate(WrapperTestTemplate):
+    def __init__(self, template: Optional[TestTemplate] = None, *args, **kwargs) -> None:
+        super().__init__(template, *args, **kwargs)
```
HyeockJinKim, line 16: I would like to avoid using args and kwargs unless it is an unavoidable situation.

src/ai/backend/test/templates/template.py
```diff
@@ -36,15 +42,22 @@ async def test(self) -> None:
 
 class TestTemplate(ABC):
     @final
-    def with_wrappers(self, *wrappers: WrapperTestTemplateProtocol) -> "TestTemplate":
+    def with_wrappers(
+        self, *wrappers: WrapperTestTemplateProtocol | type[WrapperTestTemplateProtocol]
```
HyeockJinKim, line 46: Please avoid using type unions where a specific interface should be used in a fixed manner as much as possible.

src/ai/backend/test/templates/template.py
```diff
@@ -36,15 +42,22 @@ async def test(self) -> None:
...
         """
         current = self
         for wrapper in wrappers[::-1]:
-            current = wrapper.wrap(current)
+            if isinstance(wrapper, type):
+                current = wrapper.wrap(current)
+            elif isinstance(wrapper, WrapperTestTemplate):
+                current = wrapper.wrap_by_instance(current)
+            else:
+                current = wrapper.wrap(current)
```
HyeockJinKim, line 60: I hope you haven't used branching the type internally.

HyeockJinKim (changes requested):

jopemachine:
With this approach, we can't preserve the order of the context. We could potentially enforce the order by removing the cases where a type is passed as an argument and converting everything to instances, but that probably wouldn't be a good idea.

So, I will close this PR for now.

PR #4913, opened 2025-07-01
fix(BA-1735): Resolve empty unmanaged_path as null

### LINKED ISSUE #4911: Resolve empty "unmanaged_path" value as null
unmanaged_path field of vfolder schema is nullable and the value should be null if the vfolder is not “unmanaged vfolder”. But it used to be saved with empty string value ““, it makes difficult for clients to distinguish whether the vfolder is unmanaged or no

fregataa:
resolves #4911 (BA-1735)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/models/gql_models/vfolder.py
+++ b/src/ai/backend/manager/models/gql_models/vfolder.py
@@ -201,7 +201,7 @@ class VirtualFolderNode(graphene.ObjectType):
             group=row.group_row.id if row.group_row else None,
             group_name=row.group_row.name if row.group_row else None,
             creator=row.creator,
-            unmanaged_path=row.unmanaged_path,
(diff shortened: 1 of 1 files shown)


ironAiken2 (changes requested):
The stored value still has "" (empty string), so it looks like it needs to be fixed so that empty strings are filtered out when filtering `'status == "READY" & unmanaged_path != null'`.

HyeockJinKim (approved):

HyeockJinKim:
Let's add one alembic migration code.

fregataa:
@sungchul2  I added migration script to set `NULL` to replace empty string values. Please run the alembic upgrade and review again!

ironAiken2 (approved):
LGTM

PR #4973, opened 2025-07-04
fix(BA-1768): Remove the incorrect warning logs that `BgtaskUpdated` is not supported

### LINKED ISSUE #4974: Remove the incorrect warning logs that `BgtaskUpdated` is not supported
## Summary  

Remove the incorrect warning logs that `BgtaskUpdated` is not supported

jopemachine:
resolves #4974 (BA-1768)

### DIFF
--- a/src/ai/backend/manager/services/session/service.py
+++ b/src/ai/backend/manager/services/session/service.py
@@ -23,6 +23,7 @@ from ai.backend.common.bgtask.types import BgtaskStatus
 from ai.backend.common.docker import DEFAULT_KERNEL_FEATURE, ImageRef, KernelFeatures, LabelName
 from ai.backend.common.events.event_types.bgtask.broadcast import (
     BaseBgtaskDoneEvent,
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/services/session/service.py
```diff
@@ -504,6 +507,8 @@ async def _commit_and_upload(
                 try:
                     cache_id = EventCacheDomain.BGTASK.cache_id(bgtask_id)
                     async for event in propagator.receive(cache_id):
+                        if isinstance(event, BgtaskUpdatedEvent):
+                            continue
                         if not isinstance(event, BaseBgtaskDoneEvent):
                             log.warning("unexpected event: {}", event)
```
HyeockJinKim: Since it semantically receives all BgtaskEvents, check for BgtaskEvent (in this case, if an event other than bgtask is input, it should be a warning log), and then skip Updated in the match case.

HyeockJinKim (approved):

PR #4996, opened 2025-07-07
fix(BA-1763): Modify configs to avoid resource creation timeouts

### LINKED ISSUE #4966: Investigate the cause of the race conditions
## Summary

Let's investigate and resolve the cause of specific tests failing or hanging intermittently during execution.

seedspirit:
resolves #4966 (BA-1763)
When running tests that create multiple multi-container sessions, there were cases where the session creation would pend due to exhausting the computing resources of the test environment and lack of time creating multi-kernel routes. This can be avoided by setting the memory value on session creation to minimum (320mb), which is set in the test default config, so we want to modify the config value and timeouts

- Changed session resource memory size to 320mb(minimum resource to commit session image)
- Doubled timeout limit waiting model service creation

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/configs/tester/tester.toml
+++ b/configs/tester/tester.toml
@@ -41,7 +41,7 @@ name = "cr.backend.ai/multiarch/python:3.13-ubuntu24.04"
 architecture = "x86_64"
 
 [context.session]
-resources = { cpu = 1, mem = "512mb" }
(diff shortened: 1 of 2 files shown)


jopemachine:
We need to find a way to run the tester regardless of the host's specifications, but for now, let's minimize the required resources.

PR #5006, opened 2025-07-07
feat(BA-1781): Add group purge test

### LINKED ISSUE #5000: Add purge group test in tester
## Motivation  

To ensure the proper behavior of the purge group feature, add tests related to the feature to tester

##

seedspirit:
resolves #5000 (BA-1781)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/changes/5006.feature.md
+++ b/changes/5006.feature.md
@@ -1,0 +1,1 @@
+Add tests for both successful and failed purge group operations  


src/ai/backend/test/templates/group/group.py
```diff
@@ -0,0 +1,55 @@
...
+
+    async def _retry_purge_group(self, client_session, group_id: str) -> None:
+        for _ in range(3):
+            try:
+                await client_session.Group.purge(group_id)
+                return
+            except BackendAPIError as e:
+                if e.status == 400:
+                    # Wait for other resources(ex. Model Service) to be cleaned up
+                    await asyncio.sleep(10)
```
HyeockJinKim, line 55: Is there a reason for repeating it 3 times?

seedspirit: I set the timeout to match the timeout of the other tests, which is around 30 seconds. Since I can't figure out which sessions or endpoints are active in the group based on the current group id, there doesn't seem to be a way to trigger a purge other than setting an arbitrary time.

HyeockJinKim (approved):

PR #5035, opened 2025-07-09
feat(BA-1794): Apply ValkeyStatClient 

### LINKED ISSUE #5015: Apply valkeyClient pattern to redis_stat
Currently, when using Redis, a Redis leak is occurring.

The code using Redis is separated, making it difficult to easily check how it is being used.

Due to the above reasons, the client layer will be separated like redis_stream.

JIRA Issue: BA-1794

HyeockJinKim:
resolves #5015 (BA-1794)

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/manager/api/auth.py
```diff
@@ -484,20 +481,20 @@ async def _query_cred(access_key):
...
+            num_queries_key = f"kp:{access_key}:num_queries"
+            await root_ctx.valkey_stat_client.incr(num_queries_key)
+            await root_ctx.valkey_stat_client.expire(
+                num_queries_key, 86400 * 30
+            )  # retention: 1 month
+            last_call_time_key = f"kp:{access_key}:last_call_time"
+            await root_ctx.valkey_stat_client.set(last_call_time_key, str(now).encode())
+            await root_ctx.valkey_stat_client.expire(
+                last_call_time_key, 86400 * 30
+            )  # retention: 1 month
```
fregataa: Could this part raise performance issue?

src/ai/backend/manager/api/auth.py
```diff
@@ -518,20 +515,20 @@ async def _pipe_builder(r: Redis) -> RedisPipeline:
...
```
fregataa: Same here

src/ai/backend/manager/models/health.py
```diff
@@ -132,9 +134,15 @@ async def get_redis_object_info_list(root_ctx: RootContext) -> list[RedisObjectC
         err_msg = None
         num_connections = None
         try:
-            pool = cast(ConnectionPool, info.client.connection_pool)
-            num_connections = cast(int, pool._created_connections)  # type: ignore[attr-defined]
-            max_connections = pool.max_connections
+            # ValkeyStatClient doesn't have connection_pool attribute
+            if hasattr(info, "execute"):  # ValkeyStatClient
+                # For ValkeyStatClient, we can't get connection pool info
```
fregataa:
```suggestion
            if isinstance(info, ValkeyStatClient):
                # For ValkeyStatClient, we can't get connection pool info
```
Is there any limitation of using `isinstance()` here?

src/ai/backend/manager/scheduler/dispatcher.py
```diff
@@ -293,16 +293,9 @@ def __init__(
...
-            self.config_provider.config.redis.model_dump()
-        )
         # redis_live will be initialized in __ainit__
         self.redis_live = cast(ValkeyLiveClient, None)
-        self.redis_stat = redis_helper.get_redis_object(
-            redis_profile_target.profile_target(RedisRole.STATISTICS),
-            name="stat",
-            db=REDIS_STATISTICS_DB,
-        )
+        self.valkey_stat_client = cast(ValkeyStatClient, None)
```
fregataa: I know this is because `ValkeyStatClient` requires asynchronous task to initialize
Let's go with this for now but we should change such codes

fregataa (commented):
Looks okay in general
but let's not using `Union` type

src/ai/backend/manager/api/auth.py
HyeockJinKim: If it's on the same network, there aren't really any performance issues, but it bothers me that it's not grouped into a method, so I'm changing this to operate in batches in the PR that unifies it. 
Of course, the Redis RTT accumulating every time isn't good.
fyi. This is the existing implementation, though

src/ai/backend/manager/models/health.py
HyeockJinKim: I am removing all unnecessary type checks in the final PR.

PR #5145, opened 2025-07-15
fix(BA-1870): Relax stats serialization (#5143)

lablup-octodog:
This is an auto-generated backport PR of #5143 to the 25.6 release.


fregataa:
Closed this PR because related codes are not backported

PR #5134, opened 2025-07-14
Offload health check capability to Redis with updated API specs

kyujin-cho:
## Summary

- Refactor model service health check architecture to offload status tracking to Redis
- Update API specifications for endpoint route management and health checking
- Implement separate model entity handling for health check configuration
- Add support for reading health check information from model-definition.yaml

## Breaking Changes

**Important**: This PR breaks health check capability on OSS AppProxy. Future work will restore support, but for now disable the health check feature in `model-definition.yaml` to use model service on Open Source Backend.AI.

### API Specifications
- Updated model serving event types with new `ModelServiceStatusEventArgs` base class
- Added `EndpointRouteListUpdatedEvent` for endpoint route synchronization
- Refactored anycast/broadcast event structures for better separation of concerns

### Health Check System
- Added `ModelServiceHelper` class replacing `ModelServicePredicateChecker`
- Implemented health check configuration reading from model-definition.yaml
- Added Redis-based route connection info storage for AppProxy integration
- Separated model definition validation into discrete functions

### Route Management
- Updated endpoint route generation to use Redis for connection info storage
- Modified session lifecycle handling for route status updates
- Added `generate_redis_route_info` method for serializable connection data
- Implemented `notify_endpoint_route_update_to_appproxy` for real-time updates

### Database Changes
- Enhanced endpoint model with Redis route info generation
- Updated routing status management in session callbacks
- Improved error handling and retry logic for route operations

## Technical Details

The health check system now reads configuration from either:
1. Runtime variant profiles for predefined endpoints
2. model-definition.yaml for custom runtime variants

Route connection information is stored in Redis with the key pattern:
`endpoint.{endpoint_id}.route_connection_info`

## Testing

Existing tests should continue to pass. New functionality requires model service integration testing with Redis and AppProxy components.

PR #5230, opened 2025-07-21
feat: sync up-to-date health check information with AppProxy

kyujin-cho:
Follow-up PR of #5134. This PR updates session status transition logic to generate up-to-date health check information upon model service's target session status updates.

### DIFF
--- a/src/ai/backend/common/clients/valkey_client/valkey_live/client.py
+++ b/src/ai/backend/common/clients/valkey_client/valkey_live/client.py
@@ -483,22 +483,24 @@ class ValkeyLiveClient:
         connection_info: dict[str, Any],
         health_check_config: HealthCheckConfig | None,
     ) -> None:
-        await self._client.client.set(
+        pipe = self._create_batch()
+        pipe.set(
             f"endpoint.{endpoint_id}.route_connection_info",
             json.dumps(connection_info),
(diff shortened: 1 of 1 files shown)


src/ai/backend/common/clients/valkey_client/valkey_live/client.py
```diff
@@ -473,6 +476,30 @@ async def hgetall_str(self, key: str) -> dict[str, str]:
...
+            f"endpoint.{endpoint_id}.health_check_enabled",
+            "true" if health_check_config is not None else "false",
+            expiry=ExpirySet(ExpiryType.SEC, 3600),
+        )
+        if health_check_config:
+            await self._client.client.set(
+                f"endpoint.{endpoint_id}.health_check_config",
+                health_check_config.model_dump_json(),
+                expiry=ExpirySet(ExpiryType.SEC, 3600),
+            )
```
HyeockJinKim: Use batch.

kyujin-cho: This block was originally written in pipeline but I just unwrapped it because there seemed to be no alternative option available for valkey... But I guess batch would fit. Thanks for the heads up.

HyeockJinKim (approved):

PR #5237, opened 2025-07-21
feat(BA-1915): Add `mount_ids`, `mount_id_map` fields to session creation config

### LINKED ISSUE #5209: Update create_from_params API to support ID-based vfolder lookup
## Context

Currently, when creating a session using the `create_from_params` API, if mounts and mount_map are requested based on IDs, the vfolder cannot be found. This issue needs to be addressed.

## Acceptance criteria

- The create_from_params API should be able to find vfolders based on IDs.
- Ensure that the API works correctly with both ID-based and name-based vfolder lookups.

jopemachine:
resolves #5209 (BA-1915)


jopemachine:
Direction change: Instead of modifying the mounts to include an ID, we decided to create a separate new field called mount_ids.

src/ai/backend/manager/registry.py
```diff
@@ -704,8 +717,9 @@ async def create_cluster(
                 "architecture": template["spec"]["kernel"].get("architecture", DEFAULT_IMAGE_ARCH),
                 "cluster_role": node["cluster_role"],
                 "creation_config": {
-                    "mount": mounts,
-                    "mount_map": mount_map,
+                    # "mount": mounts,  # Should we keep it for legacy?
+                    "mounts": mounts,
```
jopemachine: This is a separate bug fix.
This bug fix will likely need to be backported, so I'll handle it in a separate PR and just comment it out here for now.

src/ai/backend/manager/registry.py
```diff
@@ -405,29 +405,31 @@ async def create_session(
...
-                    raise InvalidAPIParameters(str(alias_name) + " is reserved for internal path.")
-                if alias_name in original_folders:
-                    raise InvalidAPIParameters(
-                        "Alias name cannot be set to an existing folder name: " + str(alias_name)
-                    )
+        # Check work directory and reserved name directory.
+        original_folders = combined_mount_map.keys()
+        alias_folders = combined_mount_map.values()
+        if len(alias_folders) != len(set(alias_folders)):
+            raise InvalidAPIParameters("Duplicate alias folder name exists.")
```
HyeockJinKim, line 417: A situation where UUID and name point to the same vfolder is expected, but we will correct this next time.

src/ai/backend/manager/registry.py
```diff
@@ -704,8 +706,8 @@ async def create_cluster(
                 "architecture": template["spec"]["kernel"].get("architecture", DEFAULT_IMAGE_ARCH),
                 "cluster_role": node["cluster_role"],
                 "creation_config": {
+                    # TODO: Rename this to `mounts`
                     "mount": mounts,
-                    "mount_map": mount_map,
```
HyeockJinKim, line 708: ?

src/ai/backend/manager/api/session.py
```diff
@@ -302,6 +302,54 @@ def check_and_return(self, value: Any) -> object:
...
+        UndefChecker | t.Null | t.String
+    ),
+    t.Key("resources", default=undefined): UndefChecker | t.Null | t.Mapping(t.String, t.Any),
+    tx.AliasedKey(["resource_opts", "resourceOpts"], default=undefined): (
+        UndefChecker | t.Null | resource_opts_iv
+    ),
+    tx.AliasedKey(["attach_network", "attachNetwork"], default=undefined): (
+        UndefChecker | t.Null | tx.UUID
+    ),
+})
```
HyeockJinKim: It seems like it is not needed.

PR #5244, opened 2025-07-22
feat(BA-1937): Implement ActionValidator

### LINKED ISSUE #5241: Add validator in processor architecture
Add to processor operation to be able to receive validator interface.

HyeockJinKim:
resolves #5241 (BA-1937)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/changes/5244.feature.md
+++ b/changes/5244.feature.md
@@ -1,0 +1,1 @@
+Implement ActionValidator


src/ai/backend/common/data/user/types.py
```diff
@@ -0,0 +1,12 @@
...
+from uuid import UUID
+
+
+@dataclass(frozen=True)
+class UserData:
+    user_id: UUID
+    is_authorized: bool
+    is_admin: bool
+    is_superadmin: bool
+    role: str
```
fregataa, line 11: How about `UserRole` enum here?

fregataa (commented):
Should we consider a case when user role changes?

src/ai/backend/common/data/user/types.py
HyeockJinKim: good.
It doesn't seem necessary to consider the situation where the User Role is changed in this type.

fregataa (approved):

src/ai/backend/manager/actions/processor.py
```diff
@@ -71,11 +84,10 @@ async def _run(self, action: TAction) -> TActionResult:
...
+                try:
+                    await monitor.done(action, process_result)
+                except Exception as e:
+                    log.warning("Error in monitor done method: {}", e)
 
     async def wait_for_complete(self, action: TAction) -> TActionResult:
         return await self._run(action)
-
-    async def fire_and_forget(self, action: TAction) -> None:
-        asyncio.create_task(self.wait_for_complete(action))
```
fregataa, line 81: why this is removed?

PR #5268, opened 2025-07-24
chore: Change the owner of the tempo and grafana volumes

Yaminyam:
Unlike other containers, grafana does not change the owner of a volume directly when it runs.
Need to change the owner directly to grant permission to use the volume.


rapsealk:
Follow-up of #3570.

PR #5286, opened 2025-07-25
feat(BA-1855, BA-1857): Add `storages` API to storage proxy

### LINKED ISSUE #5119: Add REST APIs for Artifact Registry to the storage proxy
## Motivation   

- Let's add the necessary REST APIs to use the storage proxy as an Artifact Registry for external images, model data, and similar assets.
- For more details, refer to BEP-1012.

jopemachine:
resolves #5119 (BA-1855, BA-1857)


docker-compose.halfstack-main.yml
```diff
@@ -164,5 +164,25 @@ services:
...
+    networks:
+      - half
+    ports:
+      - "9000:9000"    # API port
+      - "9001:9001"    # Console port
+    environment:
+      - MINIO_ROOT_USER=minioadmin
+      - MINIO_ROOT_PASSWORD=minioadmin
+    volumes:
+      - "./volumes/${DATADIR_PREFIX:-.}/minio-data:/data:rw"
```
HyeockJinKim: use name volume.

scripts/install-dev.sh
```diff
@@ -879,7 +880,65 @@ configure_backendai() {
...
+        show_warning "Failed to generate MinIO credentials, using default credentials"
+        MINIO_ACCESS_KEY="minioadmin"
+        MINIO_SECRET_KEY="minioadmin"
+      fi
+    fi
+  else
+    show_warning "MinIO container not found, using default credentials"
+    MINIO_ACCESS_KEY="minioadmin"
+    MINIO_SECRET_KEY="minioadmin"
+  fi
```
HyeockJinKim: If minio-related setup is needed, please separate the script and call it.

scripts/install-dev.sh
```diff
@@ -970,6 +1029,12 @@ configure_backendai() {
   # add LOCAL_STORAGE_VOLUME vfs volume
   echo "\n[volume.${LOCAL_STORAGE_VOLUME}]\nbackend = \"vfs\"\npath = \"${ROOT_PATH}/${VFOLDER_REL_PATH}\"" >> ./storage-proxy.toml
 
+  # Replace placeholder values in the [[storages]] section with actual MinIO configuration
+  sed_inplace 's/bucket = "enter_bucket_name"/bucket = "backendai-storage"/' ./storage-proxy.toml
+  sed_inplace 's#endpoint = "http://minio.example.com:9000"#endpoint = "http://127.0.0.1:9000"#' ./storage-proxy.toml
+  sed_inplace "s/access-key = \"<minio-access-key>\"/access-key = \"${MINIO_ACCESS_KEY}\"/" ./storage-proxy.toml
+  sed_inplace "s/secret-key = \"<minio-secret-key>\"/secret-key = \"${MINIO_SECRET_KEY}\"/" ./storage-proxy.toml
```
HyeockJinKim: Let's manage these fixed values with .env.

src/ai/backend/storage/api/storages.py
```diff
@@ -0,0 +1,398 @@
...
+        ),
+        t.Key("bucket"): t.String(),
+        t.Key("key"): t.String(),
+        t.Key("expiration", optional=True): t.Int(gt=0, lte=604800),
+        t.Key("content_type", optional=True): t.String(),
+        t.Key("min_size", optional=True): t.Int(gte=0),
+        t.Key("max_size", optional=True): t.Int(gt=0),
+        t.Key("filename", optional=True): t.String(),
+    },
+).allow_extra("*")  # allow JWT-intrinsic keys
```
HyeockJinKim: With pydantic, isn't there no need to specifically use s3 token data iv?
How about reusing the `@pydantic_api_handler` that I worked on before?

src/ai/backend/storage/api/storages.py
```diff
@@ -0,0 +1,398 @@
...
+            )
+
+        s3_client = S3Client(
+            bucket_name=token_data.bucket,
+            endpoint_url=storage_config.endpoint,
+            region_name=storage_config.region,
+            aws_access_key_id=storage_config.access_key,
+            aws_secret_access_key=storage_config.secret_key,
+        )
+
```
HyeockJinKim: Instead of implementing in the handler, please group the basic behavior into the service layer. However, group the part that uses the s3 client into the client layer.

src/ai/backend/storage/api/storages.py
```diff
@@ -0,0 +1,398 @@
...
+def create_app(ctx: RootContext) -> web.Application:
+    app = web.Application()
+    app["ctx"] = ctx
+    app["prefix"] = "storages"
+
+    # Route definitions following the pattern from manager.py
+    app.router.add_route("POST", "/s3/upload", stream_upload)
+    app.router.add_route("POST", "/s3/presigned-upload-url", presigned_upload_url)
+    app.router.add_route("GET", "/s3/presigned-download-url", presigned_download_url)
```
HyeockJinKim: How about the following URL?
```
/v1/storages/s3/files/upload
/v1/storages/s3/files/download
/v1/storages/s3/files/presigned/upload
/v1/storages/s3/files/presigned/download
```

src/ai/backend/storage/api/storages.py
```diff
@@ -0,0 +1,398 @@
...
+    app.router.add_route("GET", "/s3", get_object_info)
+    app.router.add_route("DELETE", "/s3", delete_object)
```
HyeockJinKim: It seems like additional information is needed to attach it only as an object.
It seems good to include bucket information in the URL path parameter overall.
`/v1/storages/s3/buckets/:bucket_name/files/upload`?

jopemachine: Changed to `/v1/storages/s3/:storage_name/buckets/:bucket_name` format.

configs/storage-proxy/halfstack.toml
```diff
@@ -106,3 +106,11 @@ path = "vfolder/local/volume1"
...
 log-level = "INFO"
 endpoint = "http://127.0.0.1:4317"
+
+[[storages]]
+name = "minio-storage"
+bucket = "enter_bucket_name"
+endpoint = "http://minio.example.com:9000"
+region = "us-east-1"
+access-key = "<minio-access-key>"
+secret-key = "<minio-secret-key>"
```
HyeockJinKim: Shouldn't it point to Minio on halfk stack?

scripts/configure-minio.sh
HyeockJinKim, line 1: Is this script necessary even when setting up with Docker Compose?

src/ai/backend/common/api_handlers.py
HyeockJinKim, line 1: This task should have had a separate PR.

src/ai/backend/storage/api/v1/storages.py
```diff
@@ -0,0 +1,323 @@
...
+from ...services.storages import StoragesService
+from ...utils import log_client_api_entry
+
+if TYPE_CHECKING:
+    from ...context import RootContext
+
+log = BraceStyleAdapter(logging.getLogger(__spec__.name))
+
+_JWT_MAX_TTL = 300
+_JWT_LEEWAY = 30
```
HyeockJinKim: It seems like the TTL is too short for pre-signed downloading and uploading.

src/ai/backend/storage/api/v1/storages.py
```diff
@@ -0,0 +1,323 @@
...
+        storages_service = StoragesService(config_ctx.storages)
+        download_stream = storages_service.stream_download(storage_name, bucket_name, filepath)
+
+        return APIStreamResponse(
+            body=download_stream,
+            status=HTTPStatus.OK,
+            headers={
+                "Content-Type": "application/octet-stream",
+            },
+        )
```
HyeockJinKim: I was thinking of only providing direct upload and download in the manager API.
It seems good to only use pre-signed URLs in the client API.

src/ai/backend/storage/api/v1/storages.py
```diff
@@ -0,0 +1,323 @@
...
+        bucket_name = path.parsed.bucket_name
+
+        await log_client_api_entry(log, "delete_file", req)
+        storages_service = StoragesService(config_ctx.storages)
+        response = await storages_service.delete_file(storage_name, bucket_name, filepath)
+
+        return APIResponse.build(
+            status_code=HTTPStatus.OK,
+            response_model=response,
+        )
```
HyeockJinKim: It seems like the manager API and client API are not separated yet, please check this once.

configs/storage-proxy/halfstack.toml
jopemachine: When running `install-dev.sh`, the `mc` command is used to create a bucket and issue the access key, and secret key.
And then the issued keys are then overwritten into the relevant file using the `sed` command.

scripts/configure-minio.sh
jopemachine: https://github.com/lablup/backend.ai/pull/5286#discussion_r2259069437
Even when configured as halfstack, we still need to create a bucket and issue the keys.

src/ai/backend/storage/api/v1/storages.py
jopemachine: The `storages` API was designed with only the client-facing API in mind.
I thought a separate manager-facing storages API wouldn't be necessary for now because the registries import service internally handles storage saving as well.

configs/storage-proxy/halfstack.toml
HyeockJinKim: Configs to be dropped after the manager integration work:
- bucket
- access-key
- secret-key

HyeockJinKim: It looks like the entire configs needs to go into the fixture.

src/ai/backend/storage/api/v1/storages.py
HyeockJinKim: Please create the TTL for pre-signed in the config.
and set default TTL as 1-hour.

src/ai/backend/storage/config/unified.py
```diff
@@ -563,6 +563,54 @@ class StorageProxyConfig(BaseModel):
     )
 
 
+# TODO: Remove this after migrating this to database
+class ObjectStorageConfig(BaseModel):
```
jopemachine, line 567: This type was arbitrarily added for use as a type hint and is scheduled to be removed in a follow-up PR.

PR #5287, opened 2025-07-26
fix(BA-1983): Use persistent aiohttp.ClientSession per route in App Proxy circuits

achimnol:
resolves #xxx (BA-1983)

- [x] Test case(s)

### DIFF
--- a/src/ai/backend/appproxy/worker/proxy/backend/http.py
+++ b/src/ai/backend/appproxy/worker/proxy/backend/http.py
@@ -3,7 +3,7 @@ import logging
 import random
 import time
 from contextlib import asynccontextmanager
-from typing import AsyncIterator, Final
+from typing import Any, AsyncIterator, Final, override
 
 import aiohttp
 import aiotools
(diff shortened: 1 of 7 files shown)


src/ai/backend/appproxy/worker/proxy/backend/http.py
```diff
@@ -27,6 +27,49 @@ class HTTPBackend(AbstractBackend):
...
+        )
+        self.client_opts: dict[str, Any] = {
+            "auto_decompress": False,
+        }
+        self.client_sessions = {
+            route.route_id: aiohttp.ClientSession(
+                base_url=f"http://{route.kernel_host}:{route.kernel_port}",
+                timeout=self.client_timeout,
+                **self.client_opts,
+            )
```
HyeockJinKim: It seems good to make the client pool limit setting configurable.

src/ai/backend/appproxy/worker/proxy/backend/http.py
```diff
@@ -27,6 +27,49 @@ class HTTPBackend(AbstractBackend):
...
+    async def update_routes(self, routes: list[RouteInfo]) -> None:
+        # Take a diff of new/old routes to create-or-destroy client sessions
+        async with self.route_lock:
+            old_route_ids = {route.route_id for route in self.routes}
+            new_route_ids = {route.route_id for route in routes}
+            await super().update_routes(routes)
+            new_routes_map = {route.route_id: route for route in self.routes}
+            async with asyncio.TaskGroup() as tg:
+                for route_id in old_route_ids - new_route_ids:
+                    tg.create_task(self.client_sessions.pop(route_id).close())
```
HyeockJinKim: If we cut it off immediately at the update time, wouldn't some responses from the existing connection potentially be lost?

src/ai/backend/appproxy/worker/proxy/backend/http.py
```diff
@@ -57,60 +100,35 @@ async def request_http(
...
-        ) as session:
-            log.debug("connecting to {}{}", base_url, request.rel_url)
-            async with session.ws_connect(request.rel_url, protocols=protocols) as ws:
-                log.debug("connected")
-                yield ws
+        client_session = self.client_sessions[route.route_id]
+        log.debug("connecting to {}{}", client_session._base_url, request.rel_url)
+        async with client_session.ws_connect(request.rel_url, protocols=protocols) as ws:
+            log.debug("connected")
+            yield ws
```
HyeockJinKim: Shouldn't we need the ws connection count metric?

src/ai/backend/appproxy/worker/proxy/backend/http.py
HyeockJinKim: Updating the routes themselves entirely seems like it wouldn't be free from race conditions.
I think a direction that improves by receiving additions, removals.
Guaranteeing at least minimal cleanup, such as expire, if needed.

achimnol:
Going to reimplement using `ai.backend.common.client.http`.

src/ai/backend/appproxy/worker/proxy/backend/http.py
achimnol: This will be divided to another PR.

PR #5289, opened 2025-07-27
fix(BA-1976): Make AsyncEtcd an explicit async context manager

achimnol:
resolves #xxx (BA-1976)

But.. this still does not resolve the issue. T_T

See also: lablup/etcd-client-py#14


achimnol:
Confirmed successful installation in our internal dev-test farm, which had issues without `asyncio.sleep(0.1)` before this PR.

HyeockJinKim (approved):

PR #5343, opened 2025-07-31
fix: Broken Agent, Webserver in HA dev environment

jopemachine:
resolves #NNN (BA-MMM)

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/common/utils.py
```diff
@@ -478,3 +478,15 @@ def deep_merge(*args: Mapping[str, Any]) -> Mapping[str, Any]:
             else:
                 merged[k] = vb
     return merged
+
+
+def addr_to_hostport_pair(addr: str) -> tuple[str, int]:
```
jopemachine, line 483: If there’s already an existing utility function, it would be better to use that instead.

src/ai/backend/agent/agent.py
```diff
@@ -802,9 +801,12 @@ async def __ainit__(self) -> None:
         self.registry_lock = asyncio.Lock()
         self.container_lifecycle_queue = asyncio.Queue()
 
-        redis_profile_target: RedisProfileTarget = RedisProfileTarget.from_dict(
-            self.local_config.model_dump(by_alias=True)["redis"]
-        )
+        if self.local_config.redis is None:
+            raise ConfigurationError({
+                "AbstractAgent.__ainit__": "Redis runtime configuration is not set."
+            })
```
HyeockJinKim, line 807: Isn't it unnecessary to check if there was a redis config in the agent?

src/ai/backend/agent/agent.py
```diff
@@ -817,13 +819,13 @@ async def __ainit__(self) -> None:
...
-        self.redis_stream_client = await ValkeyStreamClient.create(
-            redis_profile_target.profile_target(RedisRole.STREAM),
+        self.valkey_stream_client = await ValkeyStreamClient.create(
+            redis_profile_target.profile_target(RedisRole.STREAM).to_valkey_target(),
             human_readable_name="event_producer.stream",
             db_id=REDIS_STREAM_DB,
         )
         self.valkey_stat_client = await ValkeyStatClient.create(
-            redis_profile_target.profile_target(RedisRole.STATISTICS),
+            redis_profile_target.profile_target(RedisRole.STATISTICS).to_valkey_target(),
```
HyeockJinKim, line 828: How about making it `profile_valkey_target` instead of using `profile_target(...).to_valkey_target`?

src/ai/backend/agent/config/unified.py
```diff
@@ -940,6 +941,14 @@ class AgentUnifiedConfig(BaseModel):
...
         """),
     )
+    redis: Optional[RedisConfig] = Field(
+        default=None,
+        description=textwrap.dedent("""
+        Redis configuration.
+        This field is injected at runtime based on etcd configuration.
+        It is not intended to be set in the other way.
+        """),
+    )
```
HyeockJinKim, line 951: Redis config is not Optional, so please remove Optional.

src/ai/backend/agent/agent.py
```diff
@@ -802,9 +801,12 @@ async def __ainit__(self) -> None:
...
 
-        redis_profile_target: RedisProfileTarget = RedisProfileTarget.from_dict(
-            self.local_config.model_dump(by_alias=True)["redis"]
-        )
+        if self.local_config.redis is None:
+            raise ConfigurationError({
+                "AbstractAgent.__ainit__": "Redis runtime configuration is not set."
+            })
+
+        redis_profile_target = self.local_config.redis.to_redis_profile_target()
```
HyeockJinKim, line 809: Wouldn't it be okay to use to_valkey_profile_target here?

src/ai/backend/common/clients/valkey_client/client.py
```diff
@@ -38,57 +37,46 @@
...
-        return DelimiterSeperatedList(_HostPortPair).check_and_return(redis_target.sentinel)
-    elif isinstance(redis_target.sentinel, list):
-        return redis_target.sentinel
-    return None
-
-
 @dataclass
 class ValkeySentinelTarget:
-    sentinel_addresses: Iterable[HostPortPair]
+    sentinel_addresses: list[str]
```
HyeockJinKim: Please use it in a form like `Iterable[str]`, rather than `list[str]`, at caller side.

src/ai/backend/agent/agent.py
jopemachine: We need to pass `stream_redis_target` (`RedisTarget`) to `_make_message_queue`

src/ai/backend/agent/agent.py
jopemachine: Although we inject `self.local_config.redis` in the line right above, `read_agent_config`, but wouldn't it be better to double-check just in case?

src/ai/backend/agent/agent.py
jopemachine: I think it's less confusing to only allow conversion of the edge type.

src/ai/backend/agent/config/unified.py
jopemachine: TODO: We need to separate the Bootstrap config in follow-up work, just like we did in the manager.

HyeockJinKim (approved):

PR #5404, opened 2025-08-05
feat(BA-2029): Enhance Scheduler allocation logic and add comprehensive tests

### LINKED ISSUE #5352: Integrate Scheduler components
As a backend engineer, I want to integrate the Scheduler component to handle session creation and termination, so that all orchestrator features can be reliably managed on top of it.

JIRA Issue: BA-2029

HyeockJinKim:
resolves #5352 (BA-2029)

- [ ] Test case(s) to:
- [ ] Documentation


tests/manager/sokovan/scheduler/test_scheduler.py
```diff
@@ -0,0 +1,572 @@
...
+    SessionWorkload,
+)
+
+
+def create_session_workload(
+    session_id: Optional[SessionId] = None,
+    cluster_mode: ClusterMode = ClusterMode.SINGLE_NODE,
+    kernels: Optional[list[KernelWorkload]] = None,
+    designated_agent: Optional[AgentId] = None,
+    **kwargs,
```
fregataa, line 44: how about specifying signatures rather than using kwargs?

tests/manager/sokovan/scheduler/test_scheduler.py
```diff
@@ -0,0 +1,572 @@
...
+            try:
+                # Capture the state of agents at each call
+                agent_states = []
+                for agent in agents:
+                    available = agent.available_slots - agent.occupied_slots
+                    # Copy occupied_slots properly
+                    if hasattr(agent.occupied_slots, "copy"):
+                        occupied_copy = agent.occupied_slots.copy()
+                    else:
+                        occupied_copy = ResourceSlot(dict(agent.occupied_slots))
```
fregataa, line 155: `occupied_slots` always should be `ai.backend.common.types.ResourceSlot` type

tests/manager/sokovan/scheduler/test_scheduler.py
```diff
@@ -0,0 +1,572 @@
...
+                        "occupied": occupied_copy,
+                        "container_count": agent.container_count,
+                    })
+                # Copy requested_slots properly
+                if hasattr(resource_req.requested_slots, "copy"):
+                    req_slots_copy = resource_req.requested_slots.copy()
+                else:
+                    # It's a ResourceSlot object, create a new one from its data
+                    req_slots_copy = ResourceSlot(dict(resource_req.requested_slots))
+
```
fregataa, line 169: same here

tests/manager/sokovan/scheduler/test_scheduler.py
```diff
@@ -0,0 +1,572 @@
...
+        # Verify designated agent was selected
+        assert result is not None
+        assert result.kernel_allocations[0].agent_id == designated_agent
+
+        # Verify designated_agent was passed to selector
+        selector_calls = (
+            scheduler._agent_selector.select_agent_for_resource_requirements.call_args_list
+        )
+        assert len(selector_calls) == 1
+        assert selector_calls[0][0][4] == designated_agent  # 5th argument
```
fregataa: This looks too much detailed

fregataa (commented):
Just a comment about naming. How about name the field that represents the number of kernels alive in an agent to `kernel_count` rather than `container_count`? It is not a big thing but agents can have containers which is not BackendAI kernels.

tests/manager/sokovan/scheduler/test_scheduler.py
HyeockJinKim: I will leave it as is since it is test code.

PR #5461, opened 2025-08-11
fix(BA-2105): Session concurrency not updated when creating new session

jopemachine:
resolves #NNN (BA-2105)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/common/clients/valkey_client/valkey_stat/client.py
+++ b/src/ai/backend/common/clients/valkey_client/valkey_stat/client.py
@@ -687,6 +687,34 @@ class ValkeyStatClient:
         await self._client.client.exec(batch, raise_on_error=True)
 
     @valkey_decorator()
+    async def increment_keypair_concurrencies(
+        self, concurrency_to_increment: Mapping[str, int], is_private: bool = False
+    ) -> None:
+        """
+        Increment keypair concurrency counters.
+
+        :param access_key: The access key to increment concurrency for.
+        :param is_private: Whether this is for SFTP concurrency (True) or regular concurrency (False).
+        :return: The new value after increment.
+        """
+        if not concurrency_to_increment:
+            return
+
+        batch = self._create_batch()
+        op_count = 0
+
+        for access_key, delta in concurrency_to_increment.items():
+            # Skip no-op changes
+            if not delta:
+                continue
+            key = self._get_keypair_concurrency_key(access_key, is_private)
+            batch.incrby(key, int(delta))
+            op_count += 1
+
+        if op_count > 0:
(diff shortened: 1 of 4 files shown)


src/ai/backend/common/clients/valkey_client/valkey_stat/client.py
```diff
@@ -686,6 +686,34 @@ async def increment_keypair_query_count(
...
+        if not concurrency_to_increment:
+            return
+
+        batch = self._create_batch()
+        op_count = 0
+
+        for access_key, delta in concurrency_to_increment.items():
+            # Skip no-op changes
+            if not delta:
+                continue
```
HyeockJinKim: Is this kind of case possible?

src/ai/backend/common/clients/valkey_client/valkey_stat/client.py
```diff
@@ -686,6 +686,34 @@ async def increment_keypair_query_count(
...
+
+        :param access_key: The access key to increment concurrency for.
+        :param is_private: Whether this is for SFTP concurrency (True) or regular concurrency (False).
+        :return: The new value after increment.
+        """
+        if not concurrency_to_increment:
+            return
+
+        batch = self._create_batch()
+        op_count = 0
```
HyeockJinKim: It seems better to check the size of concurrency_to_increment rather than having op_count.

src/ai/backend/manager/repositories/schedule/repository.py
```diff
@@ -1207,6 +1214,12 @@ async def allocate_sessions(self, allocation_batch: AllocationBatch) -> None:
             for failure in allocation_batch.failures:
                 await self._update_session_failure_status(db_session, failure)
 
+            # Update concurrency statistics
+            await self._valkey_stat.increment_keypair_concurrencies(concurrency_to_increment)
+            await self._valkey_stat.increment_keypair_concurrencies(
+                sftp_concurrency_to_increment, is_private=True
+            )
```
HyeockJinKim: I don't think you should use the same function?
And rather than calling functions separately, please receive it as one function.

src/ai/backend/manager/sokovan/scheduler/types.py
```diff
@@ -281,6 +281,8 @@ class SessionAllocation:
     kernel_allocations: list[KernelAllocation]
     # List of agent allocations for this session
     agent_allocations: list[AgentAllocation]
+    # Keypair associated with the session
+    access_key: AccessKey
```
HyeockJinKim, line 285: Is it necessary to pass the access_key when SessionWorkload is being passed?

HyeockJinKim (approved):

PR #5493, opened 2025-08-13
fix(BA-1816): Handle clone task through events

### LINKED ISSUE #5045: Potential state stuck of VFolder in `cloning` status
The status of VFolder stucks in `cloning` status if the Manager side error occurs.

The cloning task should be fully handle in Storage-proxy side, not in Manager

JIRA Issue: BA-1816

fregataa:
resolves #5045 (BA-1816)

- [ ] Test case(s) to:
- [ ] Documentation


HyeockJinKim (changes requested):
Isn't it the same that it falls into an infinite cloning state when server operation fails or redis event produce fails?

fregataa:
> Isn't it the same that it falls into an infinite cloning state when server operation fails or redis event produce fails?

It is true and we will resolve such failure in retry feature

src/ai/backend/manager/event_dispatcher/handlers/vfolder.py
```diff
@@ -41,3 +43,26 @@ async def handle_vfolder_deletion_failure(
...
         )
+
+    async def handle_vfolder_clone_success(
+        self,
+        context: None,
+        source: AgentId,
+        event: VFolderCloneSuccessEvent,
+    ) -> None:
+        await update_vfolder_status(
+            self._db, [event.vfid.folder_id], VFolderOperationStatus.READY, do_log=True
```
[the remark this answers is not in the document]
HyeockJinKim: Is it not necessary to set dst_vfid to READY?

src/ai/backend/storage/api/manager.py
```diff
@@ -469,6 +474,39 @@ async def _delete_vfolder(
...
+                str(e),
+            )
+        )
+    else:
+        await event_producer.anycast_event(
+            VFolderCloneSuccessEvent(
+                src_vfid,
+                dst_vfid,
+            )
+        )
```
HyeockJinKim, line 507: Is there any action for when the event producer fails?

fregataa: Manager will change the status of source VFolder to `ready`. But it is the same how the Manager `VFolderCloneSuccessEvent` handles

HyeockJinKim (approved):

PR #5521, opened 2025-08-15
refactor: Document affinity-aware allocation logic and improve error messages

achimnol:
A follow-up to BA-2100


src/ai/backend/agent/affinity_map.py
```diff
@@ -113,26 +113,25 @@ def get_distance_ordered_neighbors(
         If source_devices is None, it will return the first largest connected component from the
```
rapsealk: This docstring needs to be updated as `src_devices` is not `Optional`.

src/ai/backend/agent/affinity_map.py
```diff
@@ -113,26 +113,25 @@ def get_distance_ordered_neighbors(
...
-                        else:
-                            secondary_set.add(u)
-                primary_sets.append(primary_set)
-            for primary_set in primary_sets:
-                secondary_set -= primary_set
-            return [*(list(primary_set) for primary_set in primary_sets)], list(secondary_set)
-        else:
-            components = self.get_device_clusters_with_lowest_distance(device_name)
-            return components, []
+        if next(iter(src_devices)).device_name == device_name:
```
rapsealk, line 127: This will raise `StopIteration` when `src_devices` is empty. Is this intended?

achimnol: When `src_devices` is empty, this method is NOT called at all after this PR's refactoring. I'll clean up descriptions.

PR #5533, opened 2025-08-16
feat(BA-2141): Modify value reading to avoid cache-based scheduling

### LINKED ISSUE #5532: Implemented changes to prevent calculations using cached values.
Modifying cases where calculations are based on cached values in the scheduler.

JIRA Issue: BA-2141

HyeockJinKim:
resolves #5532 (BA-2141)

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/manager/registry.py
```diff
@@ -922,11 +994,37 @@ async def enqueue_session(
...
+                internal_data=internal_data,
+                starts_at=starts_at,
+                batch_timeout=batch_timeout,
+                agent_list=agent_list,
+                dependency_sessions=dependency_sessions,
+                callback_url=callback_url,
+                route_id=route_id,
+                sudo_session_enabled=sudo_session_enabled,
+                network=network,
+            )
```
rapsealk, line 1022: The new implementation no longer supports string `session_name` in `dependency_sessions`, unlike the legacy version.

PR #5550, opened 2025-08-17
fix: Update logging import in App Proxy for compatibility with BA-2135

achimnol:
This is a follow-up fix of #5523 (BA-2135) for App Proxy.

`ai.backend.appproxy.common.logging_utils` were referring to `ai.backend.logging.BraceMessage` directly, so it became broken after #5523 due to missing `user_kwargs`.

Since the core logging now have the contextual request ID support on its own, let's update all AppProxy codebase to use the core logging directly.


achimnol:
> Backport to 25.6 is failed. Please backport manually.

There is nothing to backport here because the (deprecated and removed) `ai.backend.wsproxy` already used the `ai.backend.logging` module.

PR #5564, opened 2025-08-19
feat(BA-57): Allow specifying multiple etcd endpoint addresses

### LINKED ISSUE #3094: Set `AsyncEtcd` to receive more than one entrypoint value
### Main idea

Currently, only one etcd entrypoint can be received.In our system, in case of configuring etcd cluster, we should be able to set multiple etcd entrypoints.

### History

When requesting grpc to etcd cluster, if one etcd server is terminated abnormally, etcd request timeout occurs.This issue is a solution to register all servers in the entire etcd cluster, but it is also necessary to check the issue of etcd grpc request timeout.

### Alternative ideas

_No response_

### Anything else?

_No response_

JIRA Issue: BA-57

achimnol:
resolves #3094 (BA-57)

### DIFF
--- a/src/ai/backend/common/config.py
+++ b/src/ai/backend/common/config.py
@@ -54,7 +54,7 @@ class BaseConfigModel(BaseModel):
 etcd_config_iv = t.Dict({
     t.Key("etcd"): t.Dict({
         t.Key("namespace"): t.String,
-        t.Key("addr", ("127.0.0.1", 2379)): tx.HostPortPair | t.List(tx.HostPortPair),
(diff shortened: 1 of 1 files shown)


fregataa (commented):
How about adding a new list type field to the config schema rather than allowing List?

achimnol:
> How about adding a new list type field to the config schema rather than allowing List?

When possible, I'd like to avoid making extraneous composite type that can be expressed using combination of existing collection/container types.

HyeockJinKim (approved):

PR #5588, opened 2025-08-21
deps: Update all-smi binaries to v0.8.0

lablup-octodog:
This PR updates the all-smi binaries (musl variant) for both aarch64 and x86_64 architectures to version v0.8.0.

Source: https://github.com/inureyes/all-smi/releases/tag/v0.8.0

This update was triggered by:
- Event: workflow_dispatch
- Manual workflow dispatch


achimnol:
TODO: Update import-all-smi workflow to handle the manapage. Move the manpage into the kernel-binary directory and mount inside the container.

PR #5633, opened 2025-08-25
fix: convert image string to ImageRow object on model service creation

lizable:
resolves #5633 (BA-MMM)

The fix addresses an issue in src/ai/backend/manager/sokovan/deployment/deployment_controller.py:517 where model service creation was failing due to improper image handling:

  - Added image resolution logic: Uses ImageRow.resolve() to convert image string to proper ImageRef object
  - Fixed type mismatch: kernel_config.image_ref now receives the correctly typed ImageRef instead of a raw string
  - Maintained consistency: Follows the same image resolution pattern used in regular session creation
  - Preserved JSON serialization: Kept endpoint_resource_opts.image_ref as string for proper serialization

  The change ensures model services can be created successfully by providing the kernel configuration with the expected ImageRef object type while maintaining backward compatibility for stored endpoint resource options.

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -26,26 +26,45 @@ Consult `src/ai/backend/{package}/README.md` for package-specific descriptions.
 
 * Maintain high code quality throughout development
 * Eliminate duplication ruthlessly
-* Express intent clearly through naming and structure
-* Make dependencies explicit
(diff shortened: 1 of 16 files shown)


HyeockJinKim:
I will close it as it has been modified.

PR #5747, opened 2025-09-05
feat(BA-2262): Add `Artifact`, `ArtifactRegistry` REST API

### LINKED ISSUE #5746: Add ArtifactRegistry REST API
## Motivation

When used by the client, the REST API may be preferred.

But we also can use GQL, so it’s not an urgent.

JIRA Issue: BA-2262

jopemachine:
resolves #5746 (BA-2262)

- [ ] Test case(s) to:
- [ ] Documentation

----
📚 Documentation preview 📚: https://sorna--5747.org.readthedocs.build/en/5747/

----
📚 Documentation preview 📚: https://sorna-ko--5747.org.readthedocs.build/ko/5747/

### DIFF
--- a/src/ai/backend/manager/repositories/artifact/repository.py
+++ b/src/ai/backend/manager/repositories/artifact/repository.py
@@ -715,7 +715,7 @@ class ArtifactRepository:
             return artifact_revision_id
 
     @repository_decorator()
-    async def get_artifact_revision_readme(self, artifact_revision_id: uuid.UUID) -> Optional[str]:
+    async def get_artifact_revision_readme(self, artifact_revision_id: uuid.UUID) -> str:
         async with self._db.begin_session() as db_sess:
             result = await db_sess.execute(
                 sa.select(ArtifactRevisionRow.readme).where(


src/ai/backend/manager/dto/request.py
```diff
@@ -22,3 +24,41 @@ class SearchArtifactsReq(BaseRequestModel):
...
+
+class ImportArtifactsReq(BaseRequestModel):
+    artifact_revision_ids: list[uuid.UUID] = Field(
+        description="List of artifact revision IDs to import."
+    )
+
+
+class UpdateArtifactReq(BaseRequestModel):
+    artifact_id: uuid.UUID = Field(description="The artifact ID to update.")
+    modifier: ArtifactModifier = Field(description="Fields to modify in the artifact.")
```
HyeockJinKim: It seems good to create a separate field because the modifier is intended for internal processing objects. Please make it have fields similar to gql.

src/ai/backend/manager/repositories/artifact/repository.py
```diff
@@ -697,6 +701,34 @@ async def update_artifact_revision_bytesize(
...
+                sa.select(ArtifactRevisionRow.readme).where(
+                    ArtifactRevisionRow.id == artifact_revision_id
+                )
+            )
+            readme = result.scalar_one_or_none()
+            if readme is None:
+                raise ArtifactRevisionNotFoundError(
+                    f"Artifact revision with ID {artifact_revision_id} not found"
+                )
+            return readme
```
HyeockJinKim: The return value doesn't seem like it's Optional, but is it correct that it returns an Optional?

PR #5813, opened 2025-09-11
feat(BA-2320): Introduce error code to storage-proxy exceptions

### LINKED ISSUE #5812: Introduce error code to storage-proxy exceptions
## Motivation

Since the storage proxy also runs the bgtask,
in order to return an error code when an exception occurs in the bgtask,
we first need to introduce error codes into the storage proxy.

JIRA Issue: BA-2320

jopemachine:
resolves #5812 (BA-2320)

### DIFF
--- a/src/ai/backend/storage/volumes/abc.py
+++ b/src/ai/backend/storage/volumes/abc.py
@@ -255,7 +255,7 @@ class AbstractVolume(metaclass=ABCMeta):
             raise VFolderNotFoundError(f"VFolder not found or not a directory: {vfid}")
         target_path = (vfpath / relpath).resolve()
         if not target_path.is_relative_to(vfpath):
-            raise InvalidSubpathError(f"Invalid subpath '{relpath}' for VFolder {vfid}")
+            raise InvalidSubpathError(vfid, relpath)
         return target_path
 
     @final


src/ai/backend/storage/exception.py
```diff
@@ -1,305 +1,406 @@
...
+    @classmethod
+    def error_code(cls) -> ErrorCode:
+        return ErrorCode(
+            domain=ErrorDomain.STORAGE_PROXY,
+            operation=ErrorOperation.ACCESS,
+            error_detail=ErrorDetail.INTERNAL_ERROR,
+        )
+
+
+class NotEmptyError(BackendAIError, web.HTTPConflict):
```
HyeockJinKim: The class name does not seem clear.

HyeockJinKim: While you're organizing things this time, please organize the names as well.

src/ai/backend/storage/volumes/pool.py
```diff
@@ -148,15 +148,15 @@ def list_volumes(self) -> Mapping[str, VolumeInfo]:
...
+            raise InvalidVolumeError(str(volume_id))
         return self._volumes[volume_id].info()
 
     @actxmgr
     async def get_volume(self, volume_id: VolumeID) -> AsyncIterator[AbstractVolume]:
         try:
             yield self._volumes[volume_id]
         except KeyError:
-            raise InvalidVolumeError(volume_id)
+            raise InvalidVolumeError(str(volume_id))
```
HyeockJinKim: It seems good to view the case where the same Exception is repeated as a signal that a private method has not been separated for the same logic.
How about separating the exception handling when pulling out from the volume into a private method?

src/ai/backend/storage/volumes/dellemc/__init__.py
```diff
@@ -114,7 +114,7 @@ async def unset_quota(
         quota_id = await self._get_quota_id(qspath)
         if len([p for p in qspath.iterdir() if p.is_dir()]) > 0:
             # Check if any directory exists in the quota scope path
-            raise NotEmptyError(quota_scope_id)
+            raise QuotaDirectoryNotEmptyError(str(quota_scope_id))
```
HyeockJinKim: Add meaningful error message

PR #5841, opened 2025-09-12
feat(BA-2325): Apply client pool to synchronous API session

### LINKED ISSUE #5820: Add connection pooling support to synchronous API session
## Motivation

- 

## Required Features

- 

## Impact

- 

## Testing Scenarios

- 

JIRA Issue: BA-2325

rapsealk:
Follow-up of {#5223, #5253} and resolves #5820 (BA-2325)


src/ai/backend/client/session.py
```diff
@@ -513,12 +517,12 @@ def __init__(
         super().__init__(config=config, proxy_mode=proxy_mode)
         if aiohttp_session is not None:
             self.aiohttp_session = aiohttp_session
-            self._aiohttp_session_injected = True
+            self._owns_session = False
```
HyeockJinKim: I'm not sure if the name "own_session" is appropriate.

rapsealk: Renamed it as `_has_aiohttp_session_ownership`.

HyeockJinKim (commented):
I didn't really understand why such a major change was necessary.

rapsealk:
Closing after investigating actual usage of the sync `Session` (`_SyncWorkerThread`) across the SDK and its downstream consumers. The `SyncClientPool` added here has no consumer, and the underlying goal is better served on the async path.

**Findings:**
- The sync `Session` (`client/session.py`, backed by `_SyncWorkerThread`) is **CLI-only** — one session per short-lived invocation, nothing to pool. Every long-running server (manager, web, storage, agent, appproxy) is async and already uses the async `ClientPool`.
- Downstream synchronous services don't construct the sync `Session` either: they reach the Manager through the async v2 registry (`BackendAIClientRegistry.create`) wrapped in `async_to_sync`, or via `requests`. The sync `Session` is reached only through the CLI.
- So `SyncClientPool` pools a path that is never long-lived or high-QPS.

**The real inefficiency is on the async path:** the v2 client does `create()` → one request → `close()` per call (`base_client.py:86,98`), a fresh TCP + TLS handshake each time. The fix is reusing the aiohttp session across v2-registry calls (a session-injection seam + a per-loop pool) — a different design, so there is little to salvage here. Tracked in BA-2325 (#5820).

**Notes:**
- The `SyncWorkerThread` / `Sentinel` extraction into `common/` is reusable but currently unconsumed — not merging it standalone.
- If a long-lived *synchronous* embedder of the core `Session` ever lands on the roadmap, this becomes relevant — but the loop-affinity gap must be tested first: nothing here wires a pooled session through `Session(aiohttp_session=...)` into a real request, and aiohttp sessions are not safe across event loops.

PR #5854, opened 2025-09-12
doc: Remove wrong idle checker usage example

jopemachine:
resolves #NNN (BA-MMM)

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/manager/config/unified.py
```diff
@@ -1506,9 +1506,6 @@ class IdleCheckerConfig(BaseModel):
         """,
         examples=[
             {
-                "timeout": {
-                    "threshold": "10m",
-                },
```
HyeockJinKim: Shouldn't we replace it with an example for `network_timeout`?

PR #5887, opened 2025-09-16
feat(BA-2261): Add VFS Storage for Artifact Registry

### LINKED ISSUE #5743: Integrate VFS as Artifact Registry Storage
## Motivation

JIRA Issue: BA-2261

jopemachine:
resolves #5743 (BA-2261)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/storage/services/storages/vfs_storage.py
+++ b/src/ai/backend/storage/services/storages/vfs_storage.py
@@ -5,7 +5,7 @@ from ai.backend.common.dto.storage.response import (
 )
 from ai.backend.common.types import StreamReader
 from ai.backend.logging.utils import BraceStyleAdapter
-from ai.backend.storage.exception import StorageNotFoundError
+from ai.backend.storage.exception import StorageNotFoundError, StorageTypeInvalidError
 from ai.backend.storage.storages.storage_pool import StoragePool
 from ai.backend.storage.storages.vfs_storage import VFSStorage
 
@@ -133,7 +133,7 @@ class VFSStorageService:
         try:
             storage = self._storage_pool.get_storage(storage_name)
             if not isinstance(storage, VFSStorage):
-                raise StorageNotFoundError(f"Storage '{storage_name}' is not a VFS storage")
+                raise StorageTypeInvalidError(f"Storage '{storage_name}' is not a VFS storage")
             return storage
         except KeyError:
             raise StorageNotFoundError(f"No VFS storage configuration found for: {storage_name}")
--- a/src/ai/backend/storage/storages/object_storage.py
+++ b/src/ai/backend/storage/storages/object_storage.py
@@ -17,10 +17,10 @@ from ai.backend.storage.exception import (
     FileStreamDownloadError,
     FileStreamUploadError,
     ObjectInfoFetchError,
+    ObjectStorageObjectDeletionError,
     PresignedDownloadURLGenerationError,
     PresignedUploadURLGenerationError,
     StorageBucketFileNotFoundError,
-    StorageBucketNotFoundError,
 )
 from ai.backend.storage.storages.base import AbstractStorage
 
@@ -204,7 +204,7 @@ class ObjectStorage(AbstractStorage):
             s3_client = self._get_s3_client()
             await s3_client.delete_object(prefix)
         except Exception as e:
-            raise StorageBucketNotFoundError(f"Delete object failed: {str(e)}") from e
+            raise ObjectStorageObjectDeletionError(f"Delete object failed: {str(e)}") from e
 
     def _get_s3_client(self) -> S3Client:
         return S3Client(
--- a/src/ai/backend/storage/exception.py
+++ b/src/ai/backend/storage/exception.py
@@ -263,7 +263,20 @@ class StorageNotFoundError(BackendAIError, web.HTTPNotFound):
         )
 
 
-class StorageBucketNotFoundError(BackendAIError, web.HTTPNotFound):
+class StorageTypeInvalidError(BackendAIError, web.HTTPBadRequest):
+    error_type = "https://api.backend.ai/probs/storage/object-type-invalid"
+    error_title = "Storage Config Invalid Type"
+
+    @classmethod
+    def error_code(cls) -> ErrorCode:
+        return ErrorCode(
+            domain=ErrorDomain.STORAGE_PROXY,
+            operation=ErrorOperation.READ,
+            error_detail=ErrorDetail.BAD_REQUEST,
+        )
+
+
+class ObjectStorageBucketNotFoundError(BackendAIError, web.HTTPNotFound):
     error_type = "https://api.backend.ai/probs/storage/bucket/object-not-found"
     error_title = "Storage Bucket Not Found"
 
@@ -403,4 +416,17 @@ class NotImplementedAPI(BackendAIError, web.HTTPBadRequest):
             domain=ErrorDomain.STORAGE_PROXY,
             operation=ErrorOperation.GENERIC,
             error_detail=ErrorDetail.NOT_IMPLEMENTED,
+        )
+
+
+class ObjectStorageObjectDeletionError(BackendAIError, web.HTTPBadRequest):
+    error_type = "https://api.backend.ai/probs/storage/object/deletion/failed"
+    error_title = "Object Deletion Failed"
+
+    @classmethod
+    def error_code(cls) -> ErrorCode:
+        return ErrorCode(
+            domain=ErrorDomain.OBJECT_STORAGE,
+            operation=ErrorOperation.HARD_DELETE,
+            error_detail=ErrorDetail.INTERNAL_ERROR,
         )
--- a/src/ai/backend/storage/services/artifacts/reservoir.py
+++ b/src/ai/backend/storage/services/artifacts/reservoir.py
@@ -20,8 +20,8 @@ from ai.backend.storage.config.unified import (
 from ai.backend.storage.exception import (
     ArtifactRevisionEmptyError,
     ArtifactStorageEmptyError,
+    ObjectStorageBucketNotFoundError,
     ReservoirStorageConfigInvalidError,
-    StorageBucketNotFoundError,
     StorageNotFoundError,
 )
 from ai.backend.storage.storages.object_storage import ObjectStorage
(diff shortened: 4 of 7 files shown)


src/ai/backend/common/dto/storage/request.py
```diff
@@ -229,3 +229,52 @@ class DeleteObjectReq(BaseRequestModel):
...
+    This is used to specify the target filepath and options for the file upload.
+    """
+
+    filepath: str = Field(description="The file path within VFS storage to upload the file to.")
+    overwrite: bool = Field(
+        default=False, description="Whether to overwrite the file if it already exists."
+    )
+    content_type: Optional[str] = Field(
+        default=None, description="MIME type of the file being uploaded (optional for VFS)."
+    )
```
HyeockJinKim, line 252: Isn't it unnecessary to inject it as a header?

src/ai/backend/storage/api/v1/storages/vfs.py
```diff
@@ -0,0 +1,219 @@
...
+        Download a file from VFS storage using streaming.
+        Streams file content directly to the client without loading into memory.
+        """
+        req = body.parsed
+        filepath = req.filepath
+        storage_name = path.parsed.storage_name
+
+        await log_client_api_entry(log, "download_file", req)
+        vfs_service = VFSStorageService(self._storage_pool)
+        download_stream = vfs_service.stream_download(storage_name, filepath)
```
HyeockJinKim: It doesn't really look good to create the service in the middle.

src/ai/backend/storage/config/unified.py
```diff
@@ -609,6 +609,64 @@ class PresignedDownloadConfig(BaseModel):
...
+        """,
+        examples=["local-models", "shared-datasets"],
+    )
+    type: Literal["vfs"] = Field(
+        description="""
+        Storage type identifier for VFS storage.
+        Must be 'vfs' for VFS storage configurations.
+        """,
+        default="vfs",
+    )
```
HyeockJinKim: Make a StrEnum type for storage type.

src/ai/backend/storage/config/unified.py
```diff
@@ -880,6 +938,15 @@ class StorageProxyUnifiedConfig(BaseModel):
...
     )
+    vfs_storages: list[VFSStorageConfig] = Field(
+        default_factory=list,
+        description="""
+        List of VFS storage configurations.
+        Each configuration defines a filesystem-based storage backend.
+        """,
+        validation_alias=AliasChoices("vfs-storages", "vfs_storages"),
+        serialization_alias="vfs-storages",
+    )
```
HyeockJinKim: Should we set it to a dict instead of a list to ensure the uniqueness of the name?

src/ai/backend/common/dto/storage/request.py
jopemachine: Yes, this should be removed

src/ai/backend/storage/services/artifacts/huggingface.py
```diff
@@ -118,6 +120,149 @@ def log_progress(offset: int, final: bool = False) -> None:
...
+                        e.__class__.__name__,
+                    )
+                    await asyncio.sleep(backoff)
+                    backoff = min(backoff * 2, 30.0)
+
+                    # Refresh metadata before retry
+                    await _probe_head()
+                    continue
+        finally:
+            await session.close()
```
HyeockJinKim: A very large method has appeared, so please break down the private method well according to role.

src/ai/backend/storage/storages/object_storage.py
```diff
@@ -54,7 +55,7 @@ def __init__(self, cfg: ObjectStorageConfig) -> None:
     async def stream_upload(
         self,
         filepath: str,
-        data_stream: AsyncIterable[bytes],
+        data_stream: FileStream,
         content_type: Optional[str] = None,
```
HyeockJinKim: It seems like FileStream is usually used with a content type, so it would be good if the FileStream's method returned the content type.

src/ai/backend/storage/stream.py
```diff
@@ -0,0 +1,81 @@
...
+
+    async def __anext__(self) -> bytes:
+        if self._generator is None:
+            self._generator = self.read()
+
+        try:
+            return await self._generator.__anext__()
+        except StopAsyncIteration:
+            self._generator = None
+            raise
```
HyeockJinKim: Let's not have implementations or fields in ABC.
And there doesn't seem to be a need to leave room for reading in multiple places.

src/ai/backend/common/types.py
```diff
@@ -1894,3 +1895,8 @@ class SessionExecutionStatus(enum.StrEnum):
     FINISHED = "finished"
     CANCELED = "canceled"
     TIMEOUT = "timeout"
+
+
+class StreamReader(Protocol):
+    def read(self) -> AsyncIterator[bytes]: ...
+    def content_type(self) -> Optional[str]: ...
```
HyeockJinKim: For this current purpose, it is correct to inherit ABC. Protocol is not intended for inheritance, but is for situations where there is duck typing in multiple classes, so the usage is different.

src/ai/backend/storage/api/v1/storages/vfs_storage.py
```diff
@@ -0,0 +1,192 @@
...
+            content_type = meta.content_type or "application/octet-stream"
+        except Exception:
+            content_type = "application/octet-stream"
+
+        return APIStreamResponse(
+            body=file_stream,
+            status=HTTPStatus.OK,
+            headers={
+                "Content-Type": content_type,
+            },
```
HyeockJinKim, line 117: Please define meaningful constant values separately as Final.

src/ai/backend/storage/api/v1/storages/vfs_storage.py
```diff
@@ -0,0 +1,192 @@
...
+        Configured aiohttp Application for VFS storage API
+    """
+    app = web.Application()
+    app["ctx"] = ctx
+    app["prefix"] = "v1/storages/vfs"
+
+    vfs_service = VFSStorageService(ctx.storage_pool)
+    api_handler = VFSStorageAPIHandler(
+        vfs_service=vfs_service,
+    )
```
HyeockJinKim, line 186: The creation location is inappropriate, but I'll move on from this for now.

src/ai/backend/storage/client/s3.py
```diff
@@ -1,14 +1,68 @@
...
+    def __init__(
+        self,
+        s3_key: str,
+        bucket_name: str,
+        chunk_size: int,
+        endpoint_url: str,
+        region_name: Optional[str],
+        aws_access_key_id: Optional[str],
+        aws_secret_access_key: Optional[str],
+        content_type: Optional[str],
```
HyeockJinKim: How about managing related values together, such as a dataclass for the s3 target and a dataclass for the secret?

src/ai/backend/storage/config/unified.py
HyeockJinKim: Is `type` needed in `vfsStorage`?

src/ai/backend/storage/services/artifacts/huggingface.py
```diff
@@ -119,6 +128,168 @@ def log_progress(offset: int, final: bool = False) -> None:
...
+        """
+        Probe metadata via HEAD.
+        - May set total size, etag, and Accept-Ranges support.
+        - Any failure is swallowed (best effort).
+        """
+        total: Optional[int] = None
+        etag: Optional[str] = None
+        accept_ranges: bool = False
+
+        headers_base = {"Accept-Encoding": "identity"}
```
HyeockJinKim: If you use the same const value, it seems worth defining it as Final in global.

src/ai/backend/storage/services/artifacts/huggingface.py
```diff
@@ -119,6 +128,168 @@ def log_progress(offset: int, final: bool = False) -> None:
...
+                etag = etag or new_etag
+                accept_ranges = "bytes" in (resp.headers.get("Accept-Ranges", "")).lower()
+        except Exception:
+            # HEAD is best-effort; ignore failures.
+            pass
+
+        return _ProbeHeadInfo(total=total, etag=etag, accept_ranges=accept_ranges)
+
+    @override
+    async def read(self) -> AsyncIterator[bytes]:
```
HyeockJinKim, line 178: The function line looks too long; please refactor it appropriately later.

src/ai/backend/storage/config/unified.py
jopemachine: It's useless in the current config structure.
We need to decide on the storage config structure first

src/ai/backend/storage/config/unified.py
```diff
@@ -866,19 +1026,42 @@ class StorageProxyUnifiedConfig(BaseModel):
...
+    artifact_registries: list[ArtifactRegistryConfig] = Field(
         default_factory=list,
         description="""
         List of artifact registry configurations.
         Each configuration defines how to connect to and use an artifact registry service.
         """,
+        validation_alias=AliasChoices("artifact-registries", "artifact_registries"),
+        serialization_alias="artifact-registries",
     )
```
HyeockJinKim: I remember we talked about managing it with a dict.

src/ai/backend/storage/services/storages/vfs_storage.py
```diff
@@ -0,0 +1,139 @@
...
+            VFSStorage instance
+
+        Raises:
+            StorageNotFoundError: If storage is not found or not a VFS storage
+        """
+        try:
+            storage = self._storage_pool.get_storage(storage_name)
+            if not isinstance(storage, VFSStorage):
+                raise StorageNotFoundError(f"Storage '{storage_name}' is not a VFS storage")
+            return storage
```
HyeockJinKim, line 137: This seems more like a bad request than NotFound.

src/ai/backend/storage/storages/base.py
```diff
@@ -1,31 +1,154 @@
...
+        Args:
+            config: Unified storage proxy configuration
+
+        Returns:
+            StoragePool instance with all configured storages
+        """
+        storages: dict[str, AbstractStorage] = {}
+
+        # Add Legacy Object Storage instances
+        from ai.backend.storage.storages.object_storage import ObjectStorage
```
HyeockJinKim: Is it really necessary to import from within the internal?

HyeockJinKim: If this was done due to import cycle issues, it would be appropriate to either separate the storage pool with base.py or separate the from_config implementation. I personally think the latter looks better.

PR #6004, opened 2025-09-22
feat(BA-2478): Add RBAC Global scope

### LINKED ISSUE #6003: Add Global Scope to RBAC Implementation
**Description**
Add Global scope functionality to the existing RBAC implementation to enable easy migration of current super admin and monitor roles into the RBAC system.

**Implementation Options**
There are several approaches to represent Global scope in the RBAC database schema:

- Store a constant string value only in the scope type field, leaving scope ID empty
- Store constant string values in both scope type and scope ID fields, allowing queries on either field
- Other approaches

**Recommended Approach**
Global scope should be implemented by specifying a constant string only in the scope type field, since conceptually, global scope is inherently a type without a specific ID. This approach better reflects the semantic nature of global permissions that apply system-wide rather than to specific resources.

JIRA Issue: BA-2478

fregataa:
resolves #6003 (BA-2478)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/changes/6006.feature.md
+++ b/changes/6006.feature.md
@@ -1,0 +1,1 @@
+Add superadmin and monitor role fixtures and migrate existing admin and monitor data to RBAC DB
--- a/fixtures/manager/example-roles.json
+++ b/fixtures/manager/example-roles.json
@@ -1,5 +1,7 @@
 {
     "roles": [
+        {"id": "942d7613-27f1-49c5-944a-5c79045346a1", "name": "role_superadmin", "description": "", "source": "system", "status": "active", "created_at": "2025-09-12 09:51:58.265582+00", "updated_at": null, "deleted_at": null},
+        {"id": "02d084f3-49e1-4fcb-bc62-e9d06637e60d", "name": "role_monitor", "description": "", "source": "system", "status": "active", "created_at": "2025-09-12 09:51:58.265582+00", "updated_at": null, "deleted_at": null},
         {"id": "31f74f6d-04f7-418b-b375-ab0c39fa3058", "name": "role_user_user2", "description": "", "source": "system", "status": "active", "created_at": "2025-09-12 09:51:58.265582+00", "updated_at": null, "deleted_at": null},
         {"id": "5e96b4ee-2296-495b-90b4-0fc845d13c67", "name": "role_user_monitor", "description": "", "source": "system", "status": "active", "created_at": "2025-09-12 09:51:58.265582+00", "updated_at": null, "deleted_at": null},
         {"id": "cb0768f1-3b35-415d-beef-8916a5c3c88b", "name": "role_user_domain-admin", "description": "", "source": "system", "status": "active", "created_at": "2025-09-12 09:51:58.265582+00", "updated_at": null, "deleted_at": null},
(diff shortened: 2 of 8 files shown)


src/ai/backend/manager/data/permission/id.py
```diff
@@ -17,12 +20,18 @@ def from_str(cls, val: str) -> Self:
     def to_str(self) -> str:
         return f"{self.scope_type}:{self.scope_id}"
 
+    def is_global(self) -> bool:
+        return self.scope_type == ScopeType.GLOBAL
```
HyeockJinKim: I’m not sure if this is really meaningful.

src/ai/backend/manager/data/permission/role.py
```diff
@@ -74,11 +76,41 @@ class RoleDataWithPermissions:
 
 
 @dataclass
-class PermissionCheckInput:
+class MergedPermissionData:
+    scope_permissions: Mapping[ScopeId, set[OperationType]]
+    global_permissions: set[OperationType]
+    object_permissions: Mapping[ObjectId, set[OperationType]]
```
HyeockJinKim: It seems like it would be good to separate it. However, I think we need to consider how to separate it a little bit, so I'll take a look and suggest some ideas.

src/ai/backend/manager/data/permission/id.py
```diff
@@ -9,6 +9,9 @@ class ScopeId:
     scope_type: ScopeType
     scope_id: str
 
+    def __hash__(self) -> int:
+        return hash((self.scope_type, self.scope_id))
```
HyeockJinKim: How about setting the frozen attribute of the dataclass?

HyeockJinKim: It seems better to make values immutable rather than making mutable values hashable.

src/ai/backend/manager/data/permission/role.py
```diff
@@ -94,6 +97,28 @@ class SingleEntityPermissionCheckInput:
...
+    scope_permissions: set[OperationType]
+    global_permissions: Optional[set[OperationType]]
+
+
+@dataclass
+class ObjectPermissionSet:
+    object_id: ObjectId
+    object_permissions: set[OperationType]
+    mapped_scopes: dict[ScopeId, set[OperationType]]
+    global_permissions: Optional[set[OperationType]]
```
HyeockJinKim: I think it would be good to keep most of the values frozen.

src/ai/backend/manager/repositories/permission_controller/db_source.py
```diff
@@ -142,3 +146,271 @@ async def get_entity_mapped_scopes(
...
+                )
+            )
+            .where(
+                sa.and_(
+                    RoleRow.status == RoleStatus.ACTIVE,
+                    UserRoleRow.user_id == user_id,
+                    sa.or_(
+                        PermissionGroupRow.scope_type == ScopeType.GLOBAL,
+                        PermissionGroupRow.scope_id == scope_id.scope_id,
+                    ),
```
HyeockJinKim: Do we need to check the scope type?

HyeockJinKim: I think we should also receive the value for the operation.

src/ai/backend/manager/repositories/permission_controller/db_source.py
```diff
@@ -142,3 +146,271 @@ async def get_entity_mapped_scopes(
...
+            for role in role_rows:
+                for pg in role.permission_group_rows:
+                    if pg.parsed_scope_id() != scope_id:
+                        continue
+                    if pg.scope_type == ScopeType.GLOBAL:
+                        global_permissions = {perm.operation for perm in pg.permission_rows}
+                    else:
+                        scope_permissions |= {perm.operation for perm in pg.permission_rows}
+
+        return ScopePermissionSet(scope_id, scope_permissions, global_permissions)
```
HyeockJinKim: There doesn't seem to be a need to separate global permission specifically. What's actually needed is whether or not there's permission for the operation.

src/ai/backend/manager/repositories/permission_controller/db_source.py
```diff
@@ -142,3 +146,271 @@ async def get_entity_mapped_scopes(
...
+
+    async def check_scope_permission_exist(
+        self,
+        user_id: uuid.UUID,
+        scope_id: ScopeId,
+        operation: OperationType,
+    ) -> bool:
```
HyeockJinKim, line 153: If we are returning the permission set anyway, it doesn't seem necessary to have a separate check function.

src/ai/backend/manager/repositories/permission_controller/db_source.py
```diff
@@ -142,3 +146,271 @@ async def get_entity_mapped_scopes(
...
+        return (
+            sa.select(RoleRow)
+            .select_from(
+                sa.join(UserRoleRow, RoleRow.id == UserRoleRow.role_id)
+                .join(PermissionGroupRow, RoleRow.id == PermissionGroupRow.role_id)
+                .join(
+                    AssociationScopesEntitiesRow,
+                    PermissionGroupRow.scope_id == AssociationScopesEntitiesRow.scope_id,
+                )
+                .join(ObjectPermissionRow, RoleRow.id == ObjectPermissionRow.role_id)
```
HyeockJinKim, line 197: Now I'm starting to get confused about whether we decided to remove object permissions or not. How were we proceeding?

src/ai/backend/manager/repositories/permission_controller/repository.py
```diff
@@ -97,13 +102,34 @@ async def check_permission_of_entity(self, data: SingleEntityPermissionCheckInpu
...
-        for role in role_rows:
-            for permission_group in role.permission_group_rows:
-                if permission_group.parsed_scope_id() != target_scope_id:
-                    continue
-                for permission in permission_group.permission_rows:
-                    if permission.operation == data.operation:
-                        return True
+        return await self._db_source.check_scope_permission_exist(
+            data.user_id, data.target_scope_id, data.operation
+        )
```
HyeockJinKim, line 105: Due to the situation where a considerable amount of computation is performed in the DB, a cache layer is essential. As a follow-up task, please also apply a cache source.

src/ai/backend/manager/data/permission/types.py
```diff
@@ -118,3 +118,8 @@ class ScopeType(enum.StrEnum):
     DOMAIN = "domain"
     PROJECT = "project"
     USER = "user"
+
+    GLOBAL = "global"  # Global scope
```
HyeockJinKim: Please remove line breaks.

src/ai/backend/manager/repositories/permission_controller/db_source.py
```diff
@@ -142,3 +146,271 @@ async def get_entity_mapped_scopes(
...
+                ),
+                contains_eager(RoleRow.object_permission_rows),
+            )
+        )
+
+    async def get_object_permissions(
+        self,
+        user_id: uuid.UUID,
+        object_id: ObjectId,
+    ) -> ObjectPermissionSet:
```
HyeockJinKim: You missed resilience decorator.

fregataa: This is a db source module, not a repository module

HyeockJinKim (approved):

PR #6006, opened 2025-09-23
feat(BA-2477): Add superadmin and monitor roles

### LINKED ISSUE #6002: Support system-managed monitor role
Currently, there are two types of RBAC Role Sources: SYSTEM and CUSTOM. The SYSTEM source role automatically adds entity permissions when a new entity type is introduced. For example, the user role created when a user is generated is a SYSTEM role.

The SYSTEM role adds all permissions of the entity upon its addition.

We aim to create a monitor role that is either default or system-managed, with read-only permissions.

We should extend the existing Role Source to manage the system-sourced monitor role.

JIRA Issue: BA-2477

fregataa:
resolves #6002 (BA-2477)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/models/alembic/versions/09206ac04fd3_create_global_scope_roles.py
+++ b/src/ai/backend/manager/models/alembic/versions/09206ac04fd3_create_global_scope_roles.py
@@ -8,7 +8,6 @@ Create Date: 2025-09-22 19:44:13.067238
 
 import enum
 import uuid
-from collections.abc import Iterable
 from typing import Any
 
 import sqlalchemy as sa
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/models/alembic/versions/09206ac04fd3_create_global_scope_roles.py
```diff
@@ -0,0 +1,237 @@
...
+            mapper_registry.metadata,
+            IDColumn("uuid"),
+            sa.Column("username", sa.String(length=64), unique=True),
+            sa.Column("role", EnumValueType(UserRole), default=UserRole.USER),
+            extend_existing=True,
+        )
+        return users_table
+
+
+class RoleCreator:
```
HyeockJinKim, line 78: Let's move RoleCreator to migration_utils.

PR #6014, opened 2025-09-23
fix(BA-2489): Change RBAC association table constraint

### LINKED ISSUE #6012: Change `uq_scope_id_entity_id` of association_scopes_entities table
The `association_scopes_entities` table has a `uq_scope_id_entity_id` constraint that indexes (scope_id, entity_id).

The domain scope uses string scope IDs like “default,” which can lead to unique constraint violations. This issue may also arise during migration.

To address this, perform the following actions:

1. Update migration scripts to permit duplicate records in `association_scopes_entities`.
1. Change the `uq_scope_id_entity_id` constraint to index (scope_type, scope_id, entity_id) after ensuring unique records.

JIRA Issue: BA-2489

fregataa:
resolves #6012 (BA-2489)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/repositories/permission_controller/role_manager.py
+++ b/src/ai/backend/manager/repositories/permission_controller/role_manager.py
@@ -135,8 +135,8 @@ class RoleManager:
                 sa.insert(AssociationScopesEntitiesRow).values(creator.fields_to_store())
             )
         except IntegrityError:
-            log.warning(
-                "entity and scope mapping already exists: %s, %s. Skipping.",
+            log.exception(
+                "entity and scope mapping already exists: {}, {}. Skipping.",
                 entity_id.to_str(),
                 scope_id.to_str(),
             )


src/ai/backend/manager/repositories/permission_controller/role_manager.py
```diff
@@ -125,9 +130,16 @@ async def map_entity_to_scope(
...
-        await db_session.execute(
-            sa.insert(AssociationScopesEntitiesRow).values(creator.fields_to_store())
-        )
+        try:
+            await db_session.execute(
+                sa.insert(AssociationScopesEntitiesRow).values(creator.fields_to_store())
+            )
+        except IntegrityError:
+            log.warning(
+                "entity and scope mapping already exists: %s, %s. Skipping.",
```
HyeockJinKim: You should have used {} instead of %s.

HyeockJinKim: And please also check if it would be appropriate to log this part as an exception log, as it seems like a situation that should not occur if possible.

PR #6139, opened 2025-10-01
fix(BA-2596): Clean container utilization metric when destroy kernel

### LINKED ISSUE #6138: Container utilization metric does not been cleaned
We save the container utilization metric to Prometheus with labels such as `kernel_id` and `user_id`.

We must clear Prometheus data for the specified `kernel_id` when the kernel is destroyed. Currently, this step is missing, causing users to see the total utilization of destroyed kernels.

JIRA Issue: BA-2596

fregataa:
resolves #6138 (BA-2596)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/agent/metrics/metric.py
+++ b/src/ai/backend/agent/metrics/metric.py
@@ -1,4 +1,8 @@
+import asyncio
 import enum
+import functools
+import uuid
+from collections.abc import Iterable
 from typing import Optional, Self
 
 from prometheus_client import Counter, Gauge, Histogram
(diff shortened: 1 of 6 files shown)


src/ai/backend/agent/metrics/metric.py
```diff
@@ -116,6 +127,40 @@ def observe_container_metric(
...
+                            session_id or UNDEFINED,
+                            owner_user_id or UNDEFINED,
+                            project_id or UNDEFINED,
+                            value_type,
+                        )
+                    except KeyError:
+                        continue
+
+        def callback(task: asyncio.Task) -> None:
+            self._removal_tasks.pop(kernel_id, None)
```
HyeockJinKim, line 158: Is it really necessary to manage it all the way with a callback?

PR #6227, opened 2025-10-10
fix(BA-2658): handle quota scope not found

### LINKED ISSUE #6209: GQL `quota_scope` query does not respond proper data
GQL `quota_scope` query raises a VFolder-not-found error instead of returning the correct data.

JIRA Issue: BA-2658

fregataa:
resolves #6209 (BA-2658)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/common/exception.py
+++ b/src/ai/backend/common/exception.py
@@ -297,18 +297,18 @@ class ErrorCode:
 
         :param code_str: The error code string to parse.
         :return: An ErrorCode instance.
-        :raises ValueError: If the code_str is not in the correct format.
+        :raises InvalidErrorCode: If the code_str is not in the correct format.
         """
         parts = code_str.split("_")
         if len(parts) != 3:
-            raise ValueError(f"Invalid error code format: {code_str}")
+            raise InvalidErrorCode(f"Invalid error code format: {code_str}")
         domain_str, operation_str, error_detail_str = parts
         try:
             domain = ErrorDomain(domain_str)
             operation = ErrorOperation(operation_str)
             error_detail = ErrorDetail(error_detail_str)
         except ValueError as e:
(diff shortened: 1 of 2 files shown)


src/ai/backend/common/exception.py
```diff
@@ -289,6 +290,27 @@ def default(cls) -> Self:
...
+        if len(parts) != 3:
+            raise ValueError(f"Invalid error code format: {code_str}")
+        domain_str, operation_str, error_detail_str = parts
+        try:
+            domain = ErrorDomain(domain_str)
+            operation = ErrorOperation(operation_str)
+            error_detail = ErrorDetail(error_detail_str)
+        except ValueError as e:
+            raise ValueError(f"Invalid error code value: {e}") from e
+        return cls(domain=domain, operation=operation, error_detail=error_detail)
```
HyeockJinKim, line 312: 👍

HyeockJinKim: It seems better to raise an exception inheriting from BackendAIError instead of ValueError.

src/ai/backend/manager/clients/storage_proxy/base.py
```diff
@@ -35,6 +37,59 @@ def __init__(self, client_session: aiohttp.ClientSession, args: StorageProxyClie
...
+                    ),
+                )
+            case HTTPStatus.INTERNAL_SERVER_ERROR:
+                raise VFolderOperationFailed(
+                    extra_msg="Internal server error from storage proxy",
+                )
+            case _:
+                raise VFolderOperationFailed(
+                    extra_msg=f"Unexpected error {status_code} from storage proxy",
+                )
```
HyeockJinKim, line 63: Let's temporarily use the above exception and make it so that the storage proxy can pass the error code.

PR #6285, opened 2025-10-19
fix(BA-2720): Only create XFS lockfile when XFS storage backend is configured

### LINKED ISSUE #6282: XFS Storage backend lockfile
XFS Storage backend lockfile is located at `/tmp/backendai-xfs-file-lock`. It always attempts to create this file, even if XFS is not configured. This may result in a permission denied error if the Storage Proxy cannot access the `/tmp` directory.

JIRA Issue: BA-2720

fregataa:
resolves #6282 (BA-2720)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/storage/volumes/xfs/__init__.py
+++ b/src/ai/backend/storage/volumes/xfs/__init__.py
@@ -17,6 +17,7 @@ import aiofiles.os
 
 from ai.backend.common.etcd import AsyncEtcd
 from ai.backend.common.events.dispatcher import EventDispatcher, EventProducer
+from ai.backend.common.exception import InvalidConfigError
 from ai.backend.common.lock import FileLock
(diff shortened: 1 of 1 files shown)


src/ai/backend/storage/volumes/xfs/__init__.py
```diff
@@ -281,6 +287,39 @@ class XfsVolume(BaseVolume):
...
+        except OSError as e:
+            log.exception(
+                "Failed to create XFS backend lock file at {}: (Error: {})",
+                self._lock_path,
+                e,
+            )
+            raise InvalidConfigError(
+                f"Cannot create XFS backend lock file at {self._lock_path}"
+            ) from e
+
```
HyeockJinKim, line 322: It is generally discouraged to do work that might cause an error in the constructor.

PR #6311, opened 2025-10-21
feat(BA-2750): Add support for array in config generator 

### LINKED ISSUE #6310: Add support for array of tables syntax in config sample generator
## Motivation

- Agents config now need to express an array of complex config object values
- This is best done by using array of tables syntax `[[array.name]]` syntax
- The config generator does not currently support this syntax

## Required Features

- Ensure config generator can handle the array of tables syntax

## Impact

- The sample config generated by this script will necessarily change due to this change
- As this is only user-facing data, the impact on actual runtime should remain minimal

## Testing Scenarios

- Test by generating AgentUnifiedConfig with table values

JIRA Issue: BA-2750

hhoikoo:
resolves #6310 (BA-2750)

This change adds array of tables syntax `[[table]]` of TOML in the sample config generator. 

As the downstream change of adding multiple agents in agent runtime server will need the config to be able to express array of tables, having the sample generator be able to handle this would be beneficial.

- [ ] Test case(s) to:
- [ ] Documentation


configs/agent/sample.toml
```diff
@@ -175,9 +183,52 @@
...
+  [logging.graylog]
+    # Graylog hostname.
+    host = "127.0.0.1"
+    # Graylog server port number.
+    port = 8000
+    # Log level.
+    level = "INFO"
+    # The custom source identifier. If not specified, fqdn will be used instead.
+    ## localname = "..."
+    # The fuly qualified domain name of the source.
```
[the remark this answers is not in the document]
hhoikoo: Unrelated to this change

src/ai/backend/common/configs/sample_generator.py
```diff
@@ -135,6 +143,8 @@ def _dump_toml_scalar(
             .strip()
             .split(" = ", 1)[1]
         )
+    if isinstance(value, (list, tuple, set, frozenset)):
+        value = sorted(value)
```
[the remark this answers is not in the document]
hhoikoo: Not a realistic issue for config generation

src/ai/backend/common/configs/sample_generator.py
```diff
@@ -223,10 +233,9 @@ def _process_property(
 
         # Add description as comment if available
         description = field_info.get("description") or prop_schema.get("description")
+        is_runtime_field = False
         if description:
-            if "This field is injected at runtime" in description:
-                # Skip runtimme-generated fields.
-                return []
+            is_runtime_field = _is_runtime_injected_field(description)
```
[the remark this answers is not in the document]
hhoikoo: This hurts readability more as if description: block has more code than just setting this field

configs/agent/sample.toml
```diff
@@ -197,10 +248,11 @@
   # page/hugepage size (e.g., 2 MiB).
   memory-align-size = "16M"
   # Resource allocation order
-  allocation-order = [ "cuda", "rocm", "tpu", "cpu", "mem",]
+  allocation-order = [ "cpu", "cuda", "mem", "rocm", "tpu",]
```
HyeockJinKim: Can you guarantee that the order will be the same as the order written in the example or default?
The order of the list is important, so it shouldn't be changed. @hhoikoo

HyeockJinKim (changes requested):

configs/agent/sample.toml
hhoikoo: Good point - I changed the code such that only set values (whose ordering don't matter) are sorted, not list values.

src/ai/backend/common/configs/sample_generator.py
```diff
@@ -49,6 +49,13 @@ class _InlineTable(dict, InlineTableDict):
     pass
 
 
+def _is_runtime_injected_field(description: str) -> bool:
+    return (
+        "This field is injected at runtime" in description
+        or "not intended to be set in the configuration file" in description
+    )
```
HyeockJinKim: What is this?

src/ai/backend/common/configs/sample_generator.py
```diff
@@ -122,7 +129,8 @@ def _dump_toml_scalar(
             case "BinarySize":
                 value = f"{BinarySize(value):s}".upper()
             case "HostPortPair":
-                value = {"host": value.host, "port": value.port}
+                if isinstance(value, HostPortPair):
+                    value = {"host": value.host, "port": value.port}
```
HyeockJinKim: Does this need to be checked?

src/ai/backend/common/configs/sample_generator.py
```diff
@@ -290,21 +299,28 @@ def _process_property(
         print(
             f"{indent_str}{prop_name} ({optional=}, {union=}, {default=}, {example=}, {annotation=})",
         )
+
+        # Runtime fields should always be commented out
+        if is_runtime_field:
+            line += "## "
+
```
HyeockJinKim: Please review these sub-codes, as they are considered somewhat of a black box, and let me know.

HyeockJinKim (changes requested):

src/ai/backend/common/configs/sample_generator.py
hhoikoo: It's a convenience function for checking whether the field is runtime injected i.e. should not be included in the final config output, but the condition check was bloated and I used it wrong. Removed this function and replaced it with something simpler

src/ai/backend/common/configs/sample_generator.py
hhoikoo: Removed all confusing black box codes (to the best of my abilities)

src/ai/backend/common/configs/sample_generator.py
hhoikoo: The isinstance check was useless, but some changes were necessary

HyeockJinKim (approved):

PR #6315, opened 2025-10-21
feat(BA-2752): Add multi agents in agent server config

### LINKED ISSUE #6313: Add support for multiple agents in agent server config
## Motivation

- To support having multiple agents defined in the agent runtime server, the user should be able to define multiple agents in the agent server configuration.

## Required Features

- Add support for defining multiple agents in the agents config TOML file
- Without breaking support for existing single-agent configurations
- Implement this by using a model of defining default configurations at root level, and each agent definitions are an override.

## Impact

- This change should not break existing configurations.

## Testing Scenarios

- Add unit tests that test validation rules for agent configs and checking whether agent overrides work as intended
-  Run sample config generator to ensure definition is well formed

JIRA Issue: BA-2752

hhoikoo:
resolves #6313 (BA-2752)

This change adds config changes that allow the user to specify multiple agents in an agent runtime server. It works by keeping the existing agent configurations as default and add multiple agents as overrides.

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/agent/config/unified.py
+++ b/src/ai/backend/agent/config/unified.py
@@ -11,6 +11,7 @@ import logging
 import os
 import sys
 import textwrap
+from decimal import Decimal
 from pathlib import Path
 from typing import (
     Any,
@@ -49,6 +50,8 @@ from ai.backend.common.types import (
     BinarySizeField,
     ResourceGroupType,
     ServiceDiscoveryType,
+    SlotName,
+    SlotNameField,
 )
 from ai.backend.logging import BraceStyleAdapter
 from ai.backend.logging.config import LoggingConfig
@@ -75,6 +78,12 @@ class ScratchType(enum.StrEnum):
     HOSTFILE = "hostfile"
     MEMORY = "memory"
     K8S_NFS = "k8s-nfs"
+
+
+class ResourceAllocationMode(enum.StrEnum):
+    SHARED = "shared"
+    AUTO_SPLIT = "auto-split"
+    MANUAL = "manual"
 
 
 class AgentConfigValidationContext(BaseConfigValidationContext):
@@ -504,6 +513,34 @@ class CommonAgentConfig(BaseConfigSchema):
         validation_alias=AliasChoices("metadata-server-port", "metadata_server_port"),
         serialization_alias="metadata-server-port",
     )
+    allow_compute_plugins: Optional[set[str]] = Field(
+        default=None,
+        description="Allowed compute plugins",
+        examples=[{"ai.backend.activator.agent", "ai.backend.accelerator.cuda_open"}],
+        validation_alias=AliasChoices("allow-compute-plugins", "allow_compute_plugins"),
+        serialization_alias="allow-compute-plugins",
+    )
+    block_compute_plugins: Optional[set[str]] = Field(
+        default=None,
+        description="Blocked compute plugins",
+        examples=[{"ai.backend.accelerator.mock"}],
+        validation_alias=AliasChoices("block-compute-plugins", "block_compute_plugins"),
+        serialization_alias="block-compute-plugins",
+    )
+    allow_network_plugins: Optional[set[str]] = Field(
+        default=None,
+        description="Allowed network plugins",
+        examples=[{"ai.backend.manager.network.overlay"}],
+        validation_alias=AliasChoices("allow-network-plugins", "allow_network_plugins"),
+        serialization_alias="allow-network-plugins",
+    )
+    block_network_plugins: Optional[set[str]] = Field(
+        default=None,
+        description="Blocked network plugins",
+        examples=[{"ai.backend.manager.network.overlay"}],
+        validation_alias=AliasChoices("block-network-plugins", "block_network_plugins"),
+        serialization_alias="block-network-plugins",
+    )
     image_commit_path: AutoDirectoryPath = Field(
         default=AutoDirectoryPath("./tmp/backend.ai/commit"),
         description="Path for image commit",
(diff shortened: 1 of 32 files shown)


src/ai/backend/agent/config/unified.py
```diff
@@ -1065,29 +1096,194 @@ class AgentUnifiedConfig(BaseConfigSchema):
...
+        def validate(config: AgentSpecificConfig) -> None:
+            if config.agent.backend == AgentBackend.KUBERNETES:
+                is_scratch_k8s_nfs = config.container.scratch_type == ScratchType.K8S_NFS
+                is_nfs_address_missing = config.container.scratch_nfs_address is None
+                is_nfs_options_missing = config.container.scratch_nfs_options is None
+
+                if is_scratch_k8s_nfs and (is_nfs_address_missing or is_nfs_options_missing):
+                    raise ValueError(
+                        "scratch-nfs-address and scratch-nfs-options are required for k8s-nfs"
+                    )
```
HyeockJinKim: It would have been better to create a method in AgentSpecificConfig than this.

src/ai/backend/agent/config/unified.py
```diff
@@ -994,14 +1034,8 @@ class DockerExtraConfig(BaseConfigSchema):
     )
 
 
-class AgentUnifiedConfig(BaseConfigSchema):
+class AgentGlobalConfig(BaseConfigSchema):
```
HyeockJinKim, line 1135: It would be good if docstrings were added overall, since several Agent-related configs have been created.

src/ai/backend/agent/config/unified.py
```diff
@@ -1065,29 +1096,194 @@ class AgentUnifiedConfig(BaseConfigSchema):
...
+    @property
+    def agent_default(self) -> OverridableAgentConfig:
+        return self.agent
+
+    @property
+    def agent_configs(self) -> Sequence[AgentUnifiedConfig]:
+        agent_configs = self._for_each_agent(lambda x: x)
+        if not agent_configs:
+            raise ValueError("There must be at least one agent config")
+        return agent_configs
```
HyeockJinKim, line 1334: 👍

src/ai/backend/agent/config/unified.py
```diff
@@ -1065,29 +1096,194 @@ class AgentUnifiedConfig(BaseConfigSchema):
...
+        agents = [agent.construct_unified_config(default=self) for agent in self.agents]
+        if not agents:
+            agents.append(self)
+
+        results = []
+        for agent in agents:
+            result = func(agent)
+            if result is not None:
+                results.append(result)
+        return results
```
HyeockJinKim: Seems like excessive complexity.

hhoikoo: This is actually necessary because the configs in self.agents (if multi-agent mode is used) only contains the overrides for each agent and not the defaults. This function provides a way to deal with the config of individual agents as if it's a config of an agent in single agent mode i.e. the config contains both default values and overrides.

I honestly don't see a better way of abstracting this behavior away. We could give up on having defaults and overrides in multi agent mode, and instead require that all agent specific configs in multi agent mode should explicitly write all config values, which I think would get annoying quite easily.

src/ai/backend/agent/config/unified.py
hhoikoo: no unfortunately this has to live here and not in AgentSpecificConfig. As mentioned below in the comment for _for_each_agent(), whenever we deal with AgentSpecificConfig, we always have to normalize the AgentSpecificConfig so that the default values are also available. This is not possible at config validation stage, even with "after" mode validators as the config value as given will not be able to resolve the default values.

src/ai/backend/agent/config/unified.py
hhoikoo: Good point - done

src/ai/backend/agent/config/unified.py
HyeockJinKim: If it can be used as `def validate(config: AgentSpecificConfig) -> None:`, can't it also be provided as a method in `AgentSpecificConfig`?

HyeockJinKim: Is there any reason why accessing the value via config.check_...() wouldn’t work, given that config.container.scratch_nfs_options is accessible?

src/ai/backend/agent/config/unified.py
HyeockJinKim: I was reviewing the code, and I think we probably don’t need to create a for_each function to iterate over internal fields like this.
It would be cleaner to provide a dedicated method for the specific operations below, rather than running a function over a generic for_each loop:
```python
agents = [agent.construct_unified_config(default=self) for agent in self.agents]
if not agents:
    agents.append(self)
```

HyeockJinKim (commented):
Please check the review content.

src/ai/backend/agent/config/unified.py
hhoikoo: oh yeah that makes sense, will change it that way

src/ai/backend/agent/config/unified.py
hhoikoo: Discussed offline. Done

src/ai/backend/agent/config/unified.py
```diff
@@ -1065,29 +1128,190 @@ class AgentUnifiedConfig(BaseConfigSchema):
...
-                    "scratch-nfs-address and scratch-nfs-options are required for k8s-nfs"
-                )
-        return self
+    @property
+    def agent_default(self) -> OverridableAgentConfig:
+        return self.agent
+
+    @property
+    def agent_configs(self) -> Sequence[AgentUnifiedConfig]:
+        agent_configs = [agent.construct_unified_config(default=self) for agent in self.agents]
```
HyeockJinKim: When the operation seems to have some load, please remove the property.
This action seems to be called every time, so if the value is fixed, it would be good to set something like lru_cache.

hhoikoo: I can't use caching unfortunately as in our code the config object gets modified after it is initialized.
I did make this a regular function to make the intent clearer as you suggested

src/ai/backend/agent/config/unified.py
```diff
@@ -1065,29 +1128,189 @@ class AgentUnifiedConfig(BaseConfigSchema):
...
+    )
+
+    def validate_agent_specific_config(self) -> None:
+        match self.agent.backend:
+            case AgentBackend.KUBERNETES:
+                self.container.validate_kubernetes_nfs()
+            case AgentBackend.DOCKER:
+                DockerExtraConfig.model_validate(self.container.model_dump())
+            case AgentBackend.DUMMY:
+                pass
```
HyeockJinKim, line 1241: 👍

HyeockJinKim (approved):

PR #6320, opened 2025-10-22
feat(BA-2753): Spawn multiple agents and route RPC appropriately

### LINKED ISSUE #6314: Update Agent server RPC functions to include agent ID for agent runtime with multiple agents
## Motivation

- Enable [Backend.AI](http://Backend.AI)  agent server to support multiple agents within a single physical agent process, now called ‘Agent Runtime’, allowing different scheduling units and backend configurations while maintaining backward compatibility with existing single-agent deployments.

## Required Features

- Agent: create multiple agents based on the given config
- Agent: modify RPC functions to take in agent ID from the client
- Manager: Add agent ID on call sites of the agent RPC functions

## Impact

- Should work the same for existing single-agent configurations

## Testing Scenarios

- Add unit tests
- Run tests on local instance
- If possible, run multiple agent servers and test if agent runtime routing works properly

JIRA Issue: BA-2753

hhoikoo:
resolves #6314 (BA-2753)

This PR adds support for actually spawning multiple agents within the same agent server and adding agent_id field for all appropriate RPC calls in the agent server, then ensuring that the manager sends that info such that the agent server can correctly route the RPC calls to the correct agent.

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/agent/config/unified.py
+++ b/src/ai/backend/agent/config/unified.py
@@ -11,6 +11,7 @@ import logging
 import os
 import sys
 import textwrap
+from decimal import Decimal
 from pathlib import Path
 from typing import (
     Any,
@@ -49,6 +50,8 @@ from ai.backend.common.types import (
     BinarySizeField,
     ResourceGroupType,
     ServiceDiscoveryType,
+    SlotName,
+    SlotNameField,
 )
 from ai.backend.logging import BraceStyleAdapter
 from ai.backend.logging.config import LoggingConfig
@@ -75,6 +78,12 @@ class ScratchType(enum.StrEnum):
     HOSTFILE = "hostfile"
     MEMORY = "memory"
     K8S_NFS = "k8s-nfs"
+
+
+class ResourceAllocationMode(enum.StrEnum):
+    SHARED = "shared"
+    AUTO_SPLIT = "auto-split"
+    MANUAL = "manual"
 
 
 class AgentConfigValidationContext(BaseConfigValidationContext):
@@ -504,6 +513,34 @@ class CommonAgentConfig(BaseConfigSchema):
         validation_alias=AliasChoices("metadata-server-port", "metadata_server_port"),
         serialization_alias="metadata-server-port",
     )
+    allow_compute_plugins: Optional[set[str]] = Field(
+        default=None,
+        description="Allowed compute plugins",
+        examples=[{"ai.backend.activator.agent", "ai.backend.accelerator.cuda_open"}],
+        validation_alias=AliasChoices("allow-compute-plugins", "allow_compute_plugins"),
+        serialization_alias="allow-compute-plugins",
+    )
+    block_compute_plugins: Optional[set[str]] = Field(
+        default=None,
+        description="Blocked compute plugins",
+        examples=[{"ai.backend.accelerator.mock"}],
+        validation_alias=AliasChoices("block-compute-plugins", "block_compute_plugins"),
+        serialization_alias="block-compute-plugins",
+    )
+    allow_network_plugins: Optional[set[str]] = Field(
+        default=None,
+        description="Allowed network plugins",
+        examples=[{"ai.backend.manager.network.overlay"}],
+        validation_alias=AliasChoices("allow-network-plugins", "allow_network_plugins"),
+        serialization_alias="allow-network-plugins",
+    )
+    block_network_plugins: Optional[set[str]] = Field(
+        default=None,
+        description="Blocked network plugins",
+        examples=[{"ai.backend.manager.network.overlay"}],
+        validation_alias=AliasChoices("block-network-plugins", "block_network_plugins"),
+        serialization_alias="block-network-plugins",
+    )
     image_commit_path: AutoDirectoryPath = Field(
         default=AutoDirectoryPath("./tmp/backend.ai/commit"),
         description="Path for image commit",
@@ -595,34 +632,6 @@ class OverridableAgentConfig(BaseConfigSchema):
         examples=[item.value for item in ResourceGroupType],
         validation_alias=AliasChoices("scaling-group-type", "scaling_group_type"),
         serialization_alias="scaling-group-type",
-    )
-    allow_compute_plugins: Optional[set[str]] = Field(
-        default=None,
-        description="Allowed compute plugins",
-        examples=[{"ai.backend.activator.agent", "ai.backend.accelerator.cuda_open"}],
-        validation_alias=AliasChoices("allow-compute-plugins", "allow_compute_plugins"),
-        serialization_alias="allow-compute-plugins",
-    )
-    block_compute_plugins: Optional[set[str]] = Field(
-        default=None,
-        description="Blocked compute plugins",
-        examples=[{"ai.backend.accelerator.mock"}],
-        validation_alias=AliasChoices("block-compute-plugins", "block_compute_plugins"),
-        serialization_alias="block-compute-plugins",
-    )
-    allow_network_plugins: Optional[set[str]] = Field(
-        default=None,
-        description="Allowed network plugins",
-        examples=[{"ai.backend.manager.network.overlay"}],
-        validation_alias=AliasChoices("allow-network-plugins", "allow_network_plugins"),
-        serialization_alias="allow-network-plugins",
-    )
-    block_network_plugins: Optional[set[str]] = Field(
-        default=None,
-        description="Blocked network plugins",
-        examples=[{"ai.backend.manager.network.overlay"}],
-        validation_alias=AliasChoices("block-network-plugins", "block_network_plugins"),
-        serialization_alias="block-network-plugins",
     )
     force_terminate_abusing_containers: bool = Field(
         default=False,
@@ -867,6 +876,48 @@ class ContainerConfig(CommonContainerConfig, OverridableContainerConfig):
     pass
 
 
+class ResourceAllocationConfig(BaseConfigSchema):
+    cpu: int = Field(
+        description=textwrap.dedent("""
+        Hard CPU allocation for this agent (e.g., 8 cores).
+        Only used in MANUAL allocation mode.
+        All agents must specify this value when allocation-mode is MANUAL.
+        """),
+        examples=[8, 16],
+    )
+    mem: BinarySizeField = Field(
+        description=textwrap.dedent("""
+        Hard memory allocation for this agent (e.g., "32G").
+        Only used in MANUAL allocation mode.
+        All agents must specify this value when allocation-mode is MANUAL.
+        """),
+        examples=["32G", "64G"],
+    )
+    devices: Mapping[SlotNameField, Decimal] = Field(
+        default_factory=dict,
+        description=textwrap.dedent("""
+        Device-specific per-slot resource allocations.
+        Only used in MANUAL allocation mode.
+        """),
+        examples=[{"cuda.mem": "0.3", "cuda.shares": "0.5"}],
+    )
+
+    model_config = ConfigDict(
+        extra="allow",
+        arbitrary_types_allowed=True,
+    )
+
+    @model_validator(mode="after")
+    def validate_values_are_positive(self) -> Self:
+        if self.cpu is not None and self.cpu < 0:
+            raise ValueError(f"Allocated cpu must not be a negative value, but given {self.cpu}")
+        if self.mem is not None and self.mem < 0:
+            raise ValueError(f"Allocated mem must not be a negative value, but given {self.mem}")
+        if any(value < 0 for value in self.devices.values()):
+            raise ValueError("All allocated device resource values must not be a negative value")
+        return self
+
+
 class ResourceConfig(BaseConfigSchema):
     reserved_cpu: int = Field(
         default=1,
@@ -898,6 +949,25 @@ class ResourceConfig(BaseConfigSchema):
         examples=["8G", "16G"],
         validation_alias=AliasChoices("reserved-disk", "reserved_disk"),
         serialization_alias="reserved-disk",
+    )
+    allocation_mode: ResourceAllocationMode = Field(
+        default=ResourceAllocationMode.SHARED,
+        description=textwrap.dedent("""
+        Resource allocation mode for multi-agent scenarios.
+        - `shared`: All agents share the full resource pool (default, backward compatible).
+        - `auto-split`: Automatically divide resources equally (1/N) among all agents.
+        - `manual`: Manually specify per-agent resource allocations via config.
+        """),
+        examples=[item.value for item in ResourceAllocationMode],
+        validation_alias=AliasChoices("allocation-mode", "allocation_mode"),
+        serialization_alias="allocation-mode",
+    )
+    allocations: Optional[ResourceAllocationConfig] = Field(
+        default=None,
+        description=textwrap.dedent("""
+        Resource allocations.
+        Only used in MANUAL allocation mode.
+        """),
     )
     memory_align_size: BinarySizeField = Field(
         default=BinarySize.finite_from_str("16M"),
@@ -1183,11 +1253,11 @@ class AgentOverrideConfig(BaseConfigSchema):
         Only override fields if necessary.
         """),
     )
-    container: OverridableContainerConfig | None = Field(
+    container: Optional[OverridableContainerConfig] = Field(
         default=None,
         description="Container config overrides for the individual agent",
     )
-    resource: ResourceConfig | None = Field(
+    resource: Optional[ResourceAllocationConfig] = Field(
         default=None,
         description="Resource config overrides for the individual agent",
     )
@@ -1209,10 +1279,19 @@ class AgentOverrideConfig(BaseConfigSchema):
                 update=container_override_fields
             )
         if self.resource is not None:
-            resource_override_fields = self.resource.model_dump(
-                include=self.resource.model_fields_set
+            default_allocations = default.resource.allocations
+            override_allocations = self.resource
+            if default_allocations is None:
+                merged_allocations = override_allocations
+            else:
+                merged_allocations = default_allocations.model_copy(
+                    update=override_allocations.model_dump(
+                        include=override_allocations.model_fields_set
+                    )
+                )
+            agent_updates["resource"] = default.resource.model_copy(
+                update={"allocations": merged_allocations}
             )
-            agent_updates["resource"] = default.resource.model_copy(update=resource_override_fields)
         return default.model_copy(update=agent_updates)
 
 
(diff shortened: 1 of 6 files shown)


src/ai/backend/manager/registry.py
```diff
@@ -2168,7 +2172,7 @@ async def _query() -> ResourceSlot:
     async def update_scaling_group(self, id, scaling_group) -> None:
         agent = await self.get_instance(id, agents.c.addr)
         agent_client = self._get_agent_client(agent["id"])
-        await agent_client.update_scaling_group(scaling_group)
+        await agent_client.update_scaling_group(scaling_group, agent_client.agent_id)
```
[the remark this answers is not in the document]
hhoikoo: uh no?

src/ai/backend/manager/clients/agent/client.py
```diff
@@ -375,21 +397,22 @@ async def push_image(
...
+            return await rpc.call.push_image(image_ref, registry, agent_id=agent_id)
 
     # Scaling group management
     @agent_client_resilience.apply()
-    async def update_scaling_group(self, scaling_group: str) -> None:
+    async def update_scaling_group(self, scaling_group: str, agent_id: AgentId) -> None:
         """Update scaling group on the agent."""
         async with self._with_connection() as rpc:
-            await rpc.call.update_scaling_group(scaling_group)
+            await rpc.call.update_scaling_group(scaling_group, agent_id)
```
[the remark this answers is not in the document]
hhoikoo: nah

src/ai/backend/manager/clients/agent/client.py
```diff
@@ -375,21 +397,22 @@ async def push_image(
...
 
     # Local configuration management
     @agent_client_resilience.apply()
-    async def get_local_config(self) -> Mapping[str, str]:
+    async def get_local_config(self, agent_id: AgentId) -> Mapping[str, str]:
         """Get local configuration from the agent."""
         async with self._with_connection() as rpc:
-            return await rpc.call.get_local_config()
+            return await rpc.call.get_local_config(agent_id)
```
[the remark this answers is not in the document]
hhoikoo: nah

src/ai/backend/agent/server.py
```diff
@@ -1461,10 +1670,8 @@ def main(
         raise click.Abort()
 
     if not is_invoked_subcommand:
-        server_config.agent.pid_file.write_text(str(os.getpid()))
-        image_commit_path = server_config.agent.image_commit_path
-        image_commit_path.mkdir(parents=True, exist_ok=True)
-        ipc_base_path = server_config.agent.ipc_base_path
+        server_config.agent_common.pid_file.write_text(str(os.getpid()))
```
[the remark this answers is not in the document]
hhoikoo: The fact that linter passes means that this comment is wrong, as all fields are mutually exclusive between the common and overridable models

src/ai/backend/agent/config/unified.py
```diff
@@ -687,6 +687,46 @@ class OverridableAgentConfig(BaseConfigSchema):
...
+    ) -> OverridableAgentConfig:
+        if self.id is None and default_agent.id is None:
+            return self
+
+        agent_id = self.id if self.id is not None else f"{default_agent.id}-{agent_idx}"
+
+        if "ipc_base_path" in self.model_fields_set:
+            ipc_base_path = self.ipc_base_path
+        else:
+            ipc_base_path = default_agent.ipc_base_path / agent_id
```
HyeockJinKim: I'm not sure if string-based operations are needed.

src/ai/backend/manager/clients/agent/client.py
```diff
@@ -177,45 +191,54 @@ async def restart_kernel(
...
-            return await rpc.call.sync_kernel_registry(kernel_tuples)
+            return await rpc.call.sync_kernel_registry(kernel_tuples, agent_id=agent_id)
 
     @agent_client_resilience.apply()
-    async def drop_kernel_registry(self, kernel_id_list: list[KernelId]) -> None:
+    async def drop_kernel_registry(self, kernel_id_list: list[KernelId], agent_id: AgentId) -> None:
         """Drop kernel registry entries on the agent."""
         async with self._with_connection() as rpc:
-            await rpc.call.drop_kernel_registry(kernel_id_list)
+            await rpc.call.drop_kernel_registry(kernel_id_list, agent_id=agent_id)
```
HyeockJinKim: I think it would be good if we didn't receive the agent_id every time in the agent client's method, but instead put it as a field in the agent client and have the agent client fill it. (Because I don't want to change the arguments of the method)

hhoikoo: I agree - in fact AgentClient already stores the agent_id associated with the client object, so this is entirely unnecessary. This was straight up bad code generation from Claude Code. Reverted this mistake

src/ai/backend/agent/config/unified.py
hhoikoo: In fact this whole section was not needed after all. Removed this code

src/ai/backend/agent/docker/agent.py
```diff
@@ -1324,6 +1324,9 @@ def __init__(
             agent_public_key=agent_public_key,
         )
         self.checked_invalid_images = set()
+        # MetadataServer must be shared across all instances of DockerAgent.
+        # metadata_server is initialized by AgentRPCServer and assigned via assign_metadata_server()
+        self.metadata_server = None
```
HyeockJinKim: Couldn’t we make the metadata_server set up before the setup phase, instead of performing setup after the constructor?
Making something `optional` into `required` is a good direction, but doing the opposite increases internal complexity — so we should make sure it’s really necessary before applying it.

src/ai/backend/agent/server.py
```diff
@@ -226,7 +243,10 @@ async def _inner(self_: AgentRPCServer, request: RPCMessage) -> Any:
                 raise
             except Exception:
                 log.exception("unexpected error")
-                await self_.error_monitor.capture_exception()
+                error_monitor = self_.error_monitors.get(
+                    agent_id, self_.error_monitors[self_._default_agent_id]
+                )
+                await error_monitor.capture_exception()
```
HyeockJinKim: For the error monitor, it seems better not to distinguish by agent_id.
Defining behaviors per agent ID in the common layer (the RPC server) doesn’t seem ideal.
If needed, we could modify it — but in that case, it should be done by stacking separate layers per agent instead.

src/ai/backend/agent/server.py
```diff
@@ -1027,90 +1221,122 @@ async def get_local_config(self) -> Mapping[str, Any]:
...
+    async def upload_file(
+        self,
+        kernel_id: str,
+        filename: str,
+        filedata: bytes,
+        agent_id: AgentId | None = None,
+    ):
         log.info("rpc::upload_file(k:{0}, fn:{1})", kernel_id, filename)
-        await self.agent.accept_file(KernelId(UUID(kernel_id)), filename, filedata)
+        await self.get_agent(agent_id).accept_file(KernelId(UUID(kernel_id)), filename, filedata)
```
HyeockJinKim: I understand the intention to reduce code lines, but please try to keep one function call per line whenever possible.

src/ai/backend/agent/etcd.py
```diff
@@ -0,0 +1,58 @@
...
+from ai.backend.common.data.config.types import EtcdConfigData
+from ai.backend.common.etcd import AsyncEtcd, ConfigScopes
+from ai.backend.common.types import AgentId
+
+
+class EtcdClientRegistry:
+    _etcd_config: EtcdConfigData
+    _etcd_clients: dict[AgentId, AsyncEtcd]
+    _global_etcd: AsyncEtcd
+
```
HyeockJinKim: It seems like some changes in this PR are going in a different direction from its original purpose — could you clarify what those are?

HyeockJinKim (changes requested):

src/ai/backend/agent/server.py
hhoikoo: Good point - done. I was worried that passing in a local config that does not contain individual agent's data would be a problem, but it actually seems fine as the agent specific configs are never accessed in the plugins for error and stats mointors

src/ai/backend/agent/docker/agent.py
hhoikoo: Modified how metadata server is constructed to ensure metadata server in docker agent is required

src/ai/backend/agent/etcd.py
hhoikoo: I'll split out this PR into multiple PRs

src/ai/backend/agent/server.py
hhoikoo: Done

src/ai/backend/agent/docker/metadata/server.py
```diff
@@ -132,6 +167,9 @@ async def __ainit__(self):
         self.app.router.add_route("GET", "/", list_versions)
         self.app.router.add_route("GET", "/{version}", self.list_available_apps)
 
+    def register_agent(self, agent: AbstractAgent) -> None:
+        self.kernel_registry.register_agent(agent)
```
HyeockJinKim: The kernel must not be aware of the existence of the agent.

src/ai/backend/agent/server.py
```diff
@@ -310,15 +315,20 @@ class AgentRPCServer(aobject):
     rpc_auth_agent_secret_key: Optional[SecretKey]
 
     loop: asyncio.AbstractEventLoop
-    agent: AbstractAgent
     etcd_client_registry: EtcdClientRegistry
+    agents: Mapping[AgentId, AbstractAgent]
+    _default_agent_id: AgentId
```
HyeockJinKim: It will be convenient to manage these by bundling them together.

src/ai/backend/agent/server.py
```diff
@@ -591,26 +638,49 @@ async def __aexit__(self, *exc_info) -> None:
...
+    def get_agent(self, agent_id: AgentId | None) -> AbstractAgent:
+        if agent_id is None:
+            agent_id = self._default_agent_id
+        if agent_id not in self.agents:
+            raise ValueError(
+                f"Agent '{agent_id}' not found in this runtime. "
+                f"Available agents: {', '.join(self.agents.keys())}"
+            )
+        return self.agents[agent_id]
+
```
HyeockJinKim: Please make it so that the agent is managed separately from the server.

src/ai/backend/agent/server.py
```diff
@@ -591,26 +638,49 @@ async def __aexit__(self, *exc_info) -> None:
...
+                    break
+        else:
+            data["agent"]["scaling-group"] = scaling_group  # type: ignore[index]
         shutil.copy(cfg_src_path, f"{cfg_src_path}.bak")
         with open(cfg_src_path, "w") as f:
             tomlkit.dump(data, f)
 
-        self.agent.update_scaling_group(scaling_group)
+        agent.update_scaling_group(scaling_group)
         log.info("rpc::update_scaling_group()")
```
HyeockJinKim: Let's handle the method separately for when agent_id is specified and when it is not specified.

src/ai/backend/agent/server.py
```diff
@@ -591,26 +638,49 @@ async def __aexit__(self, *exc_info) -> None:
...
         await self.error_monitor.cleanup()
 
+    def get_agent(self, agent_id: AgentId | None) -> AbstractAgent:
+        if agent_id is None:
+            agent_id = self._default_agent_id
+        if agent_id not in self.agents:
+            raise ValueError(
+                f"Agent '{agent_id}' not found in this runtime. "
+                f"Available agents: {', '.join(self.agents.keys())}"
+            )
```
HyeockJinKim: Please define and create it as our BackendAIError.

src/ai/backend/agent/server.py
hhoikoo: Done

src/ai/backend/agent/server.py
hhoikoo: Done

src/ai/backend/agent/server.py
hhoikoo: Done

src/ai/backend/agent/server.py
hhoikoo: Done

src/ai/backend/agent/docker/metadata/server.py
hhoikoo: Removed use of AggregateKernelRegistry, and instead made the AgentRuntime handle kernel registry

src/ai/backend/agent/docker/agent.py
```diff
@@ -1369,6 +1370,7 @@ def __init__(
             kernel_registry=kernel_registry,
         )
         self.checked_invalid_images = set()
+        self.metadata_server = metadata_server
```
HyeockJinKim: Is there a need to pass it as a field if it doesn't operate within the docker package?

src/ai/backend/agent/runtime.py
```diff
@@ -9,16 +10,38 @@
...
-    agent: AbstractAgent
+    agents: dict[AgentId, AbstractAgent]
     kernel_registry: KernelRegistry
     etcd: AsyncEtcd
-    etcd_view: AgentEtcdClientView
+    etcd_views: dict[AgentId, AgentEtcdClientView]
+    metadata_server: "MetadataServer | None"
 
+    _default_agent_id: AgentId
     _stop_signal: signal.Signals
```
HyeockJinKim: Please make all fields private.

src/ai/backend/agent/runtime.py
```diff
@@ -9,16 +10,38 @@
...
 
 class AgentRuntime(aobject):
     local_config: AgentUnifiedConfig
-    agent: AbstractAgent
+    agents: dict[AgentId, AbstractAgent]
     kernel_registry: KernelRegistry
     etcd: AsyncEtcd
-    etcd_view: AgentEtcdClientView
+    etcd_views: dict[AgentId, AgentEtcdClientView]
+    metadata_server: "MetadataServer | None"
```
HyeockJinKim: There is no reason to express the type with "".

src/ai/backend/agent/runtime.py
```diff
@@ -9,16 +10,38 @@
...
+
+
+class AgentIdNotFoundError(BackendAIError):
+    @classmethod
+    def error_code(cls) -> ErrorCode:
+        return ErrorCode(
+            domain=ErrorDomain.AGENT,
+            operation=ErrorOperation.ACCESS,
+            error_detail=ErrorDetail.NOT_FOUND,
+        )
```
HyeockJinKim: Please write the exception in a separate file.

src/ai/backend/agent/runtime.py
```diff
@@ -30,27 +53,59 @@ def __init__(
...
+            await agent.shutdown(self._stop_signal)
+        if self.metadata_server is not None:
+            await self.metadata_server.cleanup()
+
+    def get_agents(self) -> list[AbstractAgent]:
+        return list(self.agents.values())
+
+    def get_agent(self, agent_id: Optional[AgentId]) -> AbstractAgent:
+        if agent_id is None:
+            agent_id = self._default_agent_id
```
HyeockJinKim: If we handle it with `_default_agent` instead of `_default_agent_id`, I think we can early return and not have to worry about whether there will be an impact on the code afterwards.

src/ai/backend/agent/runtime.py
```diff
@@ -30,27 +53,59 @@ def __init__(
...
+
+    def get_etcd(self, agent_id: Optional[AgentId]) -> AgentEtcdClientView:
+        if agent_id is None:
+            agent_id = self._default_agent_id
+        if agent_id not in self.agents:
+            raise AgentIdNotFoundError(
+                f"Agent '{agent_id}' not found in this runtime. "
+                f"Available agents: {', '.join(self.agents.keys())}"
+            )
+        return self.etcd_views[agent_id]
```
HyeockJinKim: Instead of repeating the same logic, we call get_agent(agent_id).id() to use the id value.

tests/agent/test_agent.py
```diff
@@ -98,3 +104,164 @@ async def test_read_agent_config_container_2valid(arpcs_no_ainit, mocker):
...
+[container]
+scratch-type = "hostdir"
+
+[resource]
+
+[etcd]
+namespace = "test"
+addr = { host = "127.0.0.1", port = 2379 }
+"""
+    )
```
HyeockJinKim: Please separate the necessary parts for setup into fixtures, and group the test codes based on TestClass.

src/ai/backend/agent/docker/agent.py
hhoikoo: oh yeah good point - removed it

src/ai/backend/agent/runtime.py
hhoikoo: Done

src/ai/backend/agent/runtime.py
hhoikoo: It was needed because I needed to import MetadataServer in TYPE_CHECKING if-guard (as MetadataServer is a Docker agent-specific class), but doing that meant that the execution failed as the symbol MetadataServer was needed even at runtime because of this line
I just removed this declaration here entirely as it's not needed and also not part of the core functionality of AgentRuntime

src/ai/backend/agent/runtime.py
hhoikoo: Good point - done

src/ai/backend/agent/runtime.py
hhoikoo: There was actually a bug here (the dictionary being checked should not have been self.agents, but rather self.etcd_views) and if that's different, it doesn't really make sense to get access to etcd through agents, especially since now the point at which etcd is initialized is different to when agents are initialized.
Also agent_id is now non-optional as there are no use cases where default agent's etcd is needed.

src/ai/backend/agent/runtime.py
hhoikoo: Done. I followed the style of manager/ module

tests/agent/test_agent.py
hhoikoo: Done

src/ai/backend/agent/server.py
```diff
@@ -337,9 +338,7 @@ async def __ainit__(self) -> None:
             self.rpc_auth_agent_secret_key = None
             auth_handler = None
 
-        self.runtime = await AgentRuntime.new(
-            self.local_config,
-            self.etcd,
+        await self.runtime.create_agents(
             self.stats_monitor,
             self.error_monitor,
             self.rpc_auth_agent_public_key,
```
HyeockJinKim, line 340: I'm thinking it might be better to create and generate it with classmethod rather than this method.
It's good for all fields to be prepared at the constructor point.

hhoikoo: Done

PR #6498, opened 2025-10-31
feat(BA-2851): Add resource isolation options for multi-agent

### LINKED ISSUE #6432: Add resource partitioning for agents within the same agent runtime
## Motivation

- In the current multi-agent setup, if multiple agents are spawned within the same agent runtime, all agents see its own universe of resources, which currently is equivalent to all resource available in the hardware.
- This would cause issues as full interleaving of resources across multiple agents would mean that  all agents can potentially over-allocate resources for kernels. This would result in kernels crashing from out of resources error when in each agent’s view there are enough resources.
- This may be a desired behavior for certain use cases e.g. stress testing with thousands of agents on a single agent runtime, but in most cases resource segregation is needed.

## Required Features

- Let the user configure per-agent resource limits
- Add some additional config options
  - Full interleaving: each agent assumes it has access to the full hardware
  - Equal distribution: each agent takes exactly 1/N of all resources (for convenience)

## Impact

- For single agent, regardless of what mode is used, the agent will always have access to the full hardware (excluding reserved resources for the agent server itself)

## Testing Scenarios

- Unit tests for ensuring that partitioning works as intended, testing cases where all agents combined would over-allocate (for modes where that is not an intended behavior)
- Regression testing to ensure single-agent behavior is not changed
- Manual testing to see if partitioning actually works

JIRA Issue: BA-2851

hhoikoo:
resolves #6432 (BA-2851)

* SHARED: allows all agents to see full resources (useful for stress testing). This is the same behavior as before.
* AUTO_SPLIT: automatically divides resources equally among agents.
* MANUAL: lets users specify exact per-agent allocations for all resources.
    
- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/agent/resources.py
+++ b/src/ai/backend/agent/resources.py
@@ -584,9 +584,9 @@ class ResourcePartitioner(aobject):
 
     def _calculate_available_total_slots(self) -> SlotsMap:
         reserved_resources = {
-            SlotName("cpu"): Decimal(self.local_config.resource_common.reserved_cpu),
-            SlotName("mem"): Decimal(self.local_config.resource_common.reserved_mem),
-            SlotName("disk"): Decimal(self.local_config.resource_common.reserved_disk),
+            SlotName("cpu"): Decimal(self.local_config.resource.reserved_cpu),
+            SlotName("mem"): Decimal(self.local_config.resource.reserved_mem),
+            SlotName("disk"): Decimal(self.local_config.resource.reserved_disk),
         }
 
         available_slots: dict[SlotName, Decimal] = {}
@@ -671,19 +671,18 @@ class ResourcePartitioner(aobject):
                         f"Unrecognized AbstractAllocMap type {alloc_map_type}"
                     )
             case ResourceAllocationMode.MANUAL:
+                assert resource_config.allocations is not None
                 match slot_name:
                     case SlotName(device_name="cpu"):
-                        assert resource_config.allocated_cpu is not None
-                        return Decimal(resource_config.allocated_cpu)
+                        return Decimal(resource_config.allocations.cpu)
                     case SlotName(device_name="mem"):
-                        assert resource_config.allocated_mem is not None
-                        return Decimal(resource_config.allocated_mem)
+                        return Decimal(resource_config.allocations.mem)
                     case slot_name:
-                        if slot_name not in resource_config.allocated_devices:
+                        if slot_name not in resource_config.allocations.devices:
                             raise ValueError(
-                                f"{slot_name=} not found in config {resource_config.allocated_devices!r}"
+                                f"{slot_name=} not found in config {resource_config.allocations.devices!r}"
                             )
-                        return resource_config.allocated_devices[slot_name]
+                        return resource_config.allocations.devices[slot_name]
 
     def _calculate_reserved_slots(self, device_slots: SlotsMap) -> SlotsMap:
         reserved_slots: dict[SlotName, Decimal] = {}
@@ -693,7 +692,7 @@ class ResourcePartitioner(aobject):
         return reserved_slots
 
     def _calculate_resource_scaling_factor(self, allocated_slots: SlotsMap) -> SlotsMap:
-        match self.local_config.resource_common.allocation_mode:
+        match self.local_config.resource.allocation_mode:
             case ResourceAllocationMode.SHARED:
                 return defaultdict(lambda: Decimal(1.0))
             case ResourceAllocationMode.AUTO_SPLIT:
@@ -748,7 +747,7 @@ class ResourcePartitioner(aobject):
                 return await dummy_scan(compute_device_types)
 
     def _ensure_slots_are_not_overallocated(self) -> None:
-        if self.local_config.resource_common.allocation_mode != ResourceAllocationMode.MANUAL:
+        if self.local_config.resource.allocation_mode != ResourceAllocationMode.MANUAL:
             return
 
         allocated_slots: dict[SlotName, Decimal] = defaultdict(lambda: Decimal("0"))
--- a/tests/agent/test_resource_allocation.py
+++ b/tests/agent/test_resource_allocation.py
@@ -76,12 +76,16 @@ def create_test_config(
     }
 
     # Add manual allocation fields if provided
-    if allocated_cpu is not None:
-        base_config["resource"]["allocated_cpu"] = allocated_cpu
-    if allocated_mem is not None:
-        base_config["resource"]["allocated_mem"] = BinarySize.finite_from_str(allocated_mem)
-    if allocated_devices is not None:
-        base_config["resource"]["allocated_devices"] = allocated_devices
+    if allocated_cpu is not None or allocated_mem is not None or allocated_devices is not None:
+        base_config["resource"]["allocations"] = {}
+        if allocated_cpu is not None:
+            base_config["resource"]["allocations"]["cpu"] = allocated_cpu
+        if allocated_mem is not None:
+            base_config["resource"]["allocations"]["mem"] = BinarySize.finite_from_str(
+                allocated_mem
+            )
+        if allocated_devices is not None:
+            base_config["resource"]["allocations"]["devices"] = allocated_devices
 
     # Add multiple agents if requested
     if num_agents > 1:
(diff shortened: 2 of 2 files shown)


src/ai/backend/agent/resources.py
```diff
@@ -306,7 +311,7 @@ async def extra_info(self) -> Mapping[str, str]:
         Return extra information related to this plugin,
         such as the underlying driver version and feature flags.
         """
-        return {}
+        raise NotImplementedError
```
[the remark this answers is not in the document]
hhoikoo: It's an abstractmethod

src/ai/backend/agent/resources.py
```diff
@@ -356,7 +361,7 @@ async def get_hooks(self, distro: str, arch: str) -> Sequence[Path]:
                            "alpine3.8"
         :param str arch: The target CPU architecture such as "amd64"
         """
-        return []
+        raise NotImplementedError
```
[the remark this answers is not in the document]
hhoikoo: It's an abstractmethod

src/ai/backend/agent/resources.py
```diff
@@ -369,14 +374,14 @@ async def generate_docker_args(
         docker container create API as a dictionary, referring the given allocation
         map.  The agent will merge it with its own options.
         """
-        return {}
+        raise NotImplementedError
```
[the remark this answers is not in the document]
hhoikoo: It's an abstractmethod

src/ai/backend/agent/resources.py
```diff
@@ -441,7 +446,136 @@ def get_additional_allowed_syscalls(self) -> list[str]:
...
+            )
+            for device_slot in alloc_map.device_slots.values()
+        }
+
+    def _calculate_device_slot(
+        self,
+        slot_info: DeviceSlotInfo,
+        total_slot: Decimal,
+        alloc_map_type: Type[AbstractAllocMap],
+    ) -> Decimal:
```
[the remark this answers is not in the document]
hhoikoo: uh no? do you not see that every case is covered? with either a return or a raise

src/ai/backend/agent/resources.py
```diff
@@ -441,7 +446,136 @@ def get_additional_allowed_syscalls(self) -> list[str]:
...
+        total_slots: Mapping[SlotName, Decimal],
+    ) -> Mapping[SlotName, Decimal]:
+        reserved_slots: dict[SlotName, Decimal] = {}
+        for device in computers.values():
+            device_slots = self._calculate_device_slots(device.alloc_map, total_slots)
+            self._update_alloc_map(device.alloc_map, device_slots)
+
+            device_reserved_slots = self._calculate_reserved_slots(device_slots, total_slots)
+            reserved_slots = {**reserved_slots, **device_reserved_slots}
+        return reserved_slots
```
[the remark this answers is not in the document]
hhoikoo: nah that's intentional. ignore

tests/agent/test_resource_allocation.py
```diff
@@ -0,0 +1,444 @@
...
+        }
+
+        total_slots = ResourcePartitioner.calculate_total_slots(computers, base_resource_config)
+        splitter = ResourcePartitioner(base_resource_config, num_agents=2, agent_idx=0)
+        reserved_slots = splitter.restrict_computer_resources(computers, total_slots)
+
+        assert cpu_alloc_map.device_slots[DeviceId("cpu0")].amount == Decimal("16")
+        assert cpu_alloc_map.device_slots[DeviceId("cpu1")].amount == Decimal("16")
+        assert cuda_alloc_map.device_slots[DeviceId("cuda0")].amount == Decimal("2.0")
+        assert cuda_alloc_map.device_slots[DeviceId("cuda1")].amount == Decimal("2.0")
```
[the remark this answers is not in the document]
hhoikoo: tests fixed

src/ai/backend/agent/agent.py
```diff
@@ -3541,7 +3560,7 @@ async def restart_kernel__load_config(
         """
         Restore the cluster config from a previous launch of the kernel.
         """
-        pass
+        raise NotImplementedError
```
HyeockJinKim: Changing the default behavior from pass could be a bit risky.
There might be cases that rely on the current default behavior, so we should verify those before making any changes.

src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
```diff
@@ -2039,6 +2044,10 @@ async def update_kernel_status_terminated(
                 )
             )
             result = await db_sess.execute(stmt)
+
+            if result.rowcount > 0 and agent_id is not None:
+                await recalc_agent_resource_occupancy(db_sess, agent_id)
+
```
HyeockJinKim: recalc_agent_resource_occupancy involves heavy computation, so it shouldn’t be modified to be called every time like this.

src/ai/backend/agent/agent.py
hhoikoo: Reverted the changes

src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
hhoikoo: Removed. This is not necessary for this change to work

src/ai/backend/agent/stats.py
```diff
@@ -555,6 +574,35 @@ async def collect_node_stat(self) -> None:
...
+                ]
+                if len(slot_names_with_same_prefix_as_metric_key) == 1:
+                    slot_name = slot_names_with_same_prefix_as_metric_key[0]
+                    log.info(f"Found slot name {slot_name} with same prefix as {metric_key}")
+                    return self.agent.resource_partitioner.get_resource_scaling_factor(slot_name)
+                else:
+                    raise ValueError(
+                        f"Plugin defines more than 1 device slots {slot_names}, "
+                        f"and matching with metric key {metric_key} yielded no results."
+                    )
```
HyeockJinKim, line 605: Exposing the resource partitioner externally doesn’t seem ideal. 
Unlike a simple dataclass that only holds values, for classes that perform specific actions, it’s better not to expose internal fields externally. Doing so creates coupling with external code, which makes future modifications difficult.

src/ai/backend/agent/agent.py
```diff
@@ -941,6 +949,12 @@ async def __ainit__(self) -> None:
             self.computers[name] = ComputerContext(computer, devices, alloc_map)
             metadatas.append(computer.get_metadata())
 
+        self.total_slots = self.resource_partitioner.calculate_total_slots(
+            self.computers, self.local_config.resource_common
+        )
+        self.reserved_slots = self.resource_partitioner.restrict_computer_resources(
+            self.computers, self.total_slots
+        )
```
HyeockJinKim: In the current implementation, the resource partitioner only operates during the agent’s initial injection, so it would be better to think through the structure first before making changes here.

src/ai/backend/agent/docker/agent.py
```diff
@@ -1323,6 +1325,7 @@ def __init__(
             error_monitor=error_monitor,
             skip_initial_scan=skip_initial_scan,
             agent_public_key=agent_public_key,
+            resource_partitioner=resource_partitioner,
```
HyeockJinKim: I wonder if an object like the resource partitioner, which is meant for specific operations on a resource, really needs to be part of the agent.
It would make more sense for it to exist under an object that manages resources as a whole. Ideally, the agent shouldn’t contain the actual implementations but should instead orchestrate operations by combining various methods.

src/ai/backend/agent/stats.py
hhoikoo: Changed the design completely to respond to this feedback

src/ai/backend/agent/agent.py
hhoikoo: Changed the design completely to respond to this feedback

src/ai/backend/agent/docker/agent.py
hhoikoo: Changed the design completely to respond to this feedback

src/ai/backend/common/types.py
```diff
@@ -324,6 +324,7 @@ def check_typed_tuple(value: Tuple[Any, ...], types: Tuple[Type, ...]) -> Tuple:
 
 class SlotName(UserString):
     __slots__ = ("_parsed", "_device_name", "_major_type", "_minor_type")
+    __match_args__ = ("device_name", "major_type", "minor_type")
```
HyeockJinKim, line 327: Why was this added?

src/ai/backend/agent/resources.py
```diff
@@ -444,6 +469,303 @@ def get_additional_allowed_syscalls(self) -> list[str]:
...
+
+                return await docker_scan(compute_device_types)
+            case AgentBackend.KUBERNETES:
+                from .kubernetes.resources import scan_available_resources as kubernetes_scan
+
+                return await kubernetes_scan(compute_device_types)
+            case AgentBackend.DUMMY:
+                from .dummy.resources import scan_available_resources as dummy_scan
+
+                return await dummy_scan(compute_device_types)
```
HyeockJinKim, line 747: This change, in terms of extensibility, rather seems like a regressive implementation and doesn't look good.

HyeockJinKim (changes requested):

src/ai/backend/common/types.py
hhoikoo: It's not necessary but I wanted to use pattern matching here (https://github.com/lablup/backend.ai/pull/6498/files#diff-a4da2a344d73525736025bcd638112245de4a7225d6293d21a7e1e5152224ec8R675) and this class needs this added for match statement to work

hhoikoo:
Will create a new PR

PR #6627, opened 2025-11-05
fix(BA-2948): Adjust reservoir download API client timeout and add proper connection termination handling

### LINKED ISSUE #6628: Adjust reservoir download API client timeout and add proper connection termination handling
JIRA Issue: BA-2948

jopemachine:
resolves #6628 (BA-2948)

### DIFF
--- a/src/ai/backend/storage/config/unified.py
+++ b/src/ai/backend/storage/config/unified.py
@@ -913,7 +913,7 @@ class LegacyReservoirConfig(ReservoirConfig):
 
 class ReservoirClientConfig(BaseConfigSchema):
     timeout_total: Optional[float] = Field(
-        default=None,
+        default=300.0,
         description="""
         Total timeout for the entire request (in seconds).
         None means no timeout.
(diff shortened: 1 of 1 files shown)


src/ai/backend/storage/client/manager.py
```diff
@@ -81,7 +82,8 @@ async def _request_stream(self, method: str, rel_url: str, **kwargs) -> AsyncIte
         headers = self._build_header(method=method, rel_url=rel_url)
         url = yarl.URL(self._endpoint) / rel_url.lstrip("/")
 
-        async with aiohttp.ClientSession() as session:
+        timeout = aiohttp.ClientTimeout(total=None, connect=None, sock_connect=None, sock_read=None)
+        async with aiohttp.ClientSession(timeout=timeout) as session:
```
HyeockJinKim: Please change this so that the client session is created initially and then reused. You can see the pattern of clients requesting from the manager to the storage proxy.

src/ai/backend/storage/config/unified.py
```diff
@@ -911,6 +911,49 @@ class LegacyReservoirConfig(ReservoirConfig):
...
+        default=None,
+        description="""
+        Timeout for connecting to a peer for a new connection (in seconds).
+        None means no timeout.
+        """,
+        examples=[60.0],
+        validation_alias=AliasChoices("timeout-sock-connect", "timeout_sock_connect"),
+        serialization_alias="timeout-sock-connect",
+    )
+    timeout_sock_read: Optional[float] = Field(
```
HyeockJinKim, line 945: If the default value is entered as None, the default setting may become an infinite wait, so please set it to an appropriate value.

PR #6686, opened 2025-11-10
fix(BA-2997): Handle redirect parameter in app proxy auth endpoint

yomybaby:
## Summary

Fixed an issue where the `redirect` query parameter in `/v2/proxy/auth` requests was being ignored for HTTP protocol apps. The app proxy now correctly handles the redirect path by incorporating it into the URL generation logic.

## Problem

When accessing apps through `/v2/proxy/auth` with a redirect parameter (e.g., `?redirect=/notebooks/08_dimensionality_reduction.ipynb`), the parameter was:
- ✅ Received in coordinator as query parameter
- ✅ Encoded into JWT token body  
- ✅ Decoded in worker setup handler
- ❌ **NOT APPENDED** to the final proxy URL

## Solution

Refactored the URL generation to handle redirect paths within the `generate_proxy_url` function:
- Added `redirect_path` optional parameter to `generate_proxy_url()`
- Function now generates complete URLs including redirect paths
- Ensures redirect paths start with `/` for proper URL construction
- Simplified HTTP protocol handler to use the enhanced function

## Changes

- Modified `generate_proxy_url()` to accept and process redirect paths
- Updated HTTP protocol case to pass redirect parameter to URL generator
- Added explicit `redirect_path=None` for TCP protocol for clarity
- Created changelog entry in `changes/2997.fix.md`

## Test plan

- [x] Create a session with Jupyter notebook app
- [x] Access `/v2/proxy/auth?app=jupyter&protocol=http&token={token}&session_id={session_id}&redirect=/notebooks/08_dimensionality_reduction.ipynb`
- [x] Verify it redirects to `http://127.0.0.1:10200/notebooks/08_dimensionality_reduction.ipynb` (complete path)
- [x] Verify that without redirect parameter, it redirects to base URL only
- [x] Test with various redirect paths (with/without leading slash)

## Example

**Before:** `?redirect=/notebooks/test.ipynb` → Redirects to `http://127.0.0.1:10200/` (base URL only)  
**After:** `?redirect=/notebooks/test.ipynb` → Redirects to `http://127.0.0.1:10200/notebooks/test.ipynb` (full path)

### DIFF
--- a/src/ai/backend/appproxy/worker/api/setup.py
+++ b/src/ai/backend/appproxy/worker/api/setup.py
@@ -42,16 +42,27 @@ def generate_proxy_url(
     | TraefikWildcardDomainConfig,
     protocol: str,
     circuit: Circuit,
+    redirect_path: str | None = None,
 ) -> str:
+    # Generate base URL based on config type
     match config:
         case PortProxyConfig():
-            return f"{protocol}://{config.advertised_host or config.bind_host}:{circuit.port}"
+            base_url = f"{protocol}://{config.advertised_host or config.bind_host}:{circuit.port}"
         case TraefikPortProxyConfig():
-            return f"{protocol}://{config.advertised_host}:{circuit.port}"
+            base_url = f"{protocol}://{config.advertised_host}:{circuit.port}"
         case WildcardDomainConfig():
-            return f"{protocol}://{circuit.subdomain}{config.domain}:{config.advertised_port or config.bind_addr.port}"
+            base_url = f"{protocol}://{circuit.subdomain}{config.domain}:{config.advertised_port or config.bind_addr.port}"
         case TraefikWildcardDomainConfig():
-            return f"{protocol}://{circuit.subdomain}{config.domain}:{config.advertised_port}"
+            base_url = f"{protocol}://{circuit.subdomain}{config.domain}:{config.advertised_port}"
+
+    # Append redirect path if provided
+    if redirect_path:
+        # Ensure redirect path starts with /
+        if not redirect_path.startswith("/"):
+            redirect_path = f"/{redirect_path}"
+        return f"{base_url}{redirect_path}"
+
+    return base_url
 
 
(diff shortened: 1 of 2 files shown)


src/ai/backend/appproxy/worker/api/setup.py
```diff
@@ -148,9 +148,16 @@ async def setup(
...
-                generate_proxy_url(port_config, protocol, circuit), headers=cors_headers
-            )
+            base_url = generate_proxy_url(port_config, protocol, circuit)
+
+            # Handle redirect parameter from JWT body
+            redirect_path = jwt_body.get("redirect", "")
+            if redirect_path and not redirect_path.startswith("/"):
+                redirect_path = f"/{redirect_path}"
+
+            full_url = f"{base_url}{redirect_path}" if redirect_path else base_url
```
HyeockJinKim: It seems right to handle even this logic within generate_proxy_url.

src/ai/backend/appproxy/worker/api/setup.py
```diff
@@ -42,16 +42,27 @@ def generate_proxy_url(
...
         case TraefikWildcardDomainConfig():
-            return f"{protocol}://{circuit.subdomain}{config.domain}:{config.advertised_port}"
+            base_url = f"{protocol}://{circuit.subdomain}{config.domain}:{config.advertised_port}"
+
+    # Append redirect path if provided
+    if redirect_path:
+        # Ensure redirect path starts with /
+        if not redirect_path.startswith("/"):
+            redirect_path = f"/{redirect_path}"
+        return f"{base_url}{redirect_path}"
```
rapsealk, line 63: I suggest using [`yarl.URL`](https://yarl.aio-libs.org/en/latest/api/#yarl.URL).

HyeockJinKim: For now, the type is being provided as str, so I will work up to this point in this PR.

PR #6702, opened 2025-11-10
fix(BA-3008): Model Card resolver returns original error

### LINKED ISSUE #6701: Return original error message rather than "Unknown error"
The error message from the Model Card resolver is always "Unknown error," making debugging difficult. Let the resolver returns the original error message

JIRA Issue: BA-3008

fregataa:
resolves #6701 (BA-3008)

- [ ] Test case(s) to:
- [ ] Documentation

----
📚 Documentation preview 📚: https://sorna--6702.org.readthedocs.build/en/6702/

----
📚 Documentation preview 📚: https://sorna-ko--6702.org.readthedocs.build/ko/6702/

### DIFF
--- a/src/ai/backend/manager/models/gql_models/vfolder.py
+++ b/src/ai/backend/manager/models/gql_models/vfolder.py
@@ -570,7 +570,7 @@ class ModelCard(graphene.ObjectType):
     ) -> Optional[Self]:
         try:
             return await cls.parse_row(graph_ctx, vfolder_row)
-        except ModelCardParseError as e:
+        except Exception as e:
             log.exception(
                 "Failed to parse model card from vfolder (id: {}, error: {})",
                 vfolder_row.id,
@@ -585,7 +585,7 @@ class ModelCard(graphene.ObjectType):
                     row_id=vfolder_row.id,
                     name=vfolder_row.name,
                     author=vfolder_row.creator or "",
-                    error_msg=e.extra_msg,
+                    error_msg=str(e),
                 )
             else:
                 return None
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/models/gql_models/vfolder.py
```diff
@@ -573,7 +573,7 @@ async def from_row(cls, graph_ctx: GraphQueryContext, vfolder_row: VFolderRow) -
             if isinstance(e, ModelCardProcessError):
                 error_msg = e.msg
             else:
-                error_msg = "Unknown error"
+                error_msg = repr(e)
```
HyeockJinKim: It needs to be checked if this is simply applied by sending the error as is. In the case of BackendAIError, the exception message can be passed, but if an internal query is being passed, it might not be appropriate.

src/ai/backend/manager/models/gql_models/vfolder.py
```diff
@@ -622,29 +619,36 @@ async def parse_row(cls, graph_ctx: GraphQueryContext, vfolder_row: VFolderRow)
...
                 model_definition_dict = yaml.load(model_definition_yaml)
             except YAMLError as e:
-                raise ModelCardProcessError(
-                    f"Invalid YAML syntax (data:{model_definition_yaml}, detail:{str(e)})"
+                log.exception(
+                    "Failed to parse model definition YAML from vfolder (id: {}, filename: {}, error: {})",
+                    vfolder_row.id,
+                    model_definition_filename,
+                    repr(e),
                 )
```
HyeockJinKim: There doesn't seem to be a reason to log here, since we are already except-ing and logging at a higher level.

src/ai/backend/manager/models/gql_models/vfolder.py
```diff
@@ -566,14 +565,17 @@ def parse_model(
...
-                error_msg = e.msg
-            else:
-                error_msg = "Unknown error"
+        except ModelCardParseError as e:
+            log.exception(
+                "Failed to parse model card from vfolder (id: {}, error: {})",
+                vfolder_row.id,
+                repr(e),
+            )
             if (
```
HyeockJinKim: Isn't the problem that an unknown error appears when an error other than ModelCardParseError occurs?
It seems that partial failure will not be possible if other Exceptions occur.

PR #6721, opened 2025-11-11
feat(BA-3023): Add unified multi agent etcd client handling

### LINKED ISSUE #6720: Introduce Etcd managing class for handling etcd clients for multiple agents
## Motivation

- 

## Required Features

- 

## Impact

- 

## Testing Scenarios

- 

JIRA Issue: BA-3023

hhoikoo:
resolves #6720 (BA-3023)

This change introduces EtcdClientRegistry, which is a class that can contain etcd clients for multiple agents, each with its own prefix information at scaling group and individual node level.

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/changes/6320.feature.md
+++ b/changes/6320.feature.md
@@ -1,0 +1,1 @@
+Update Agent server RPC functions to include agent ID for agent runtime with multiple agents
--- a/changes/6724.feature.md
+++ b/changes/6724.feature.md
@@ -1,0 +1,1 @@
+Add custom resource allocation in agent server config
--- a/changes/6730.enhance.md
+++ b/changes/6730.enhance.md
@@ -1,0 +1,1 @@
+Move kernel registry ownership to agent runtime
--- a/configs/agent/sample.toml
+++ b/configs/agent/sample.toml
@@ -16,14 +16,6 @@
   scaling-group = "default"
   # Scaling group type
   scaling-group-type = "compute"
-  # Allowed compute plugins
-  ## allow-compute-plugins = [ "ai.backend.accelerator.cuda_open", "ai.backend.activator.agent",]
-  # Blocked compute plugins
-  ## block-compute-plugins = [ "ai.backend.accelerator.mock",]
-  # Allowed network plugins
-  ## allow-network-plugins = [ "ai.backend.manager.network.overlay",]
-  # Blocked network plugins
-  ## block-network-plugins = [ "ai.backend.manager.network.overlay",]
   # Whether to force terminate abusing containers
   force-terminate-abusing-containers = false
   # Kernel creation concurrency
@@ -81,6 +73,14 @@
   metadata-server-bind-host = "0.0.0.0"
   # Metadata server port
   metadata-server-port = 40128
+  # Allowed compute plugins
+  ## allow-compute-plugins = [ "ai.backend.activator.agent", "ai.backend.accelerator.cuda_open",]
+  # Blocked compute plugins
+  ## block-compute-plugins = [ "ai.backend.accelerator.mock",]
+  # Allowed network plugins
+  ## allow-network-plugins = [ "ai.backend.manager.network.overlay",]
+  # Blocked network plugins
+  ## block-network-plugins = [ "ai.backend.manager.network.overlay",]
   # Path for image commit
   image-commit-path = "tmp/backend.ai/commit"
   # Path for abuse reports
@@ -156,6 +156,12 @@
   # Currently this value is unused. In future releases, it may be used to preserve
   # the minimum disk space from the scratch disk allocation via loopback files.
   reserved-disk = "8G"
+  # Resource allocation mode for multi-agent scenarios.
+  # - `shared`: All agents share the full resource pool (default, backward
+  # compatible).
+  # - `auto-split`: Automatically divide resources equally (1/N) among all agents.
+  # - `manual`: Manually specify per-agent resource allocations via config.
+  allocation-mode = "shared"
   # The alignment of the reported main memory size to absorb tiny deviations from
   # per-node firmware/hardware settings. Recommended to be multiple of the
   # page/hugepage size (e.g., 2 MiB).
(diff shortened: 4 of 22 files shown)


src/ai/backend/agent/etcd.py
```diff
@@ -0,0 +1,58 @@
...
+            self._etcd_clients[agent_id] = self._create_client(
+                agent_id, agent_config.agent.scaling_group
+            )
+
+    def _create_client(self, agent_id: AgentId | None, scaling_group: str | None) -> AsyncEtcd:
+        scope_prefix_map = {ConfigScopes.GLOBAL: ""}
+        if scaling_group is not None:
+            scope_prefix_map[ConfigScopes.SGROUP] = f"sgroup/{scaling_group}"
+        if agent_id is not None:
+            scope_prefix_map[ConfigScopes.NODE] = f"nodes/agents/{agent_id}"
```
HyeockJinKim: Are these values optional? It seems like they shouldn't be optional values, could you let me know if it's the same in the existing code?

src/ai/backend/agent/etcd.py
```diff
@@ -0,0 +1,58 @@
...
+        self._etcd_clients = {}
+        self._global_etcd = self._create_client(agent_id=None, scaling_group=None)
+
+    async def close(self) -> None:
+        for etcd in self._etcd_clients.values():
+            await etcd.close()
+        await self._global_etcd.close()
+
+    def get_client(self, agent_id: AgentId) -> AsyncEtcd:
+        return self._etcd_clients[agent_id]
```
HyeockJinKim: It seems like you've separated the etcd client for each multi-agent. Is that really necessary?
It seems like reading values with distinction, rather than taking a different etcd client approach, is the correct approach.

hhoikoo: Discussed offline. Changed approach

src/ai/backend/agent/etcd.py
hhoikoo: Discussed offline. Changed approach

tests/agent/test_etcd.py
```diff
@@ -0,0 +1,501 @@
...
+from ai.backend.common import config
+from ai.backend.common.etcd import AsyncEtcd, ConfigScopes, Event, GetPrefixValue
+from ai.backend.common.types import QueueSentinel
+
+
+@pytest.fixture
+def mock_etcd() -> AsyncMock:
+    """Create a mock AsyncEtcd instance."""
+    mock = AsyncMock(spec=AsyncEtcd)
+    return mock
```
HyeockJinKim: Let's check this part by launching the etcd container.

src/ai/backend/agent/etcd.py
```diff
@@ -0,0 +1,229 @@
...
+    ) -> Mapping[ConfigScopes, str]:
+        if scope_prefix_map is None:
+            scope_prefix_map = {}
+
+        agent_config = self._config.agent
+        return {
+            **scope_prefix_map,
+            ConfigScopes.SGROUP: f"sgroup/{agent_config.scaling_group}",
+            ConfigScopes.NODE: f"nodes/agents/{agent_config.id}",
+        }
```
HyeockJinKim: It seems right to use ChainMap like an internal implementation.

src/ai/backend/agent/etcd.py
```diff
@@ -0,0 +1,229 @@
...
+    ) -> None:
+        self._etcd = etcd
+        self._config = config
+
+    def _augment_scope_prefix_map(
+        self,
+        scope_prefix_map: Optional[Mapping[ConfigScopes, str]],
+    ) -> Mapping[ConfigScopes, str]:
+        if scope_prefix_map is None:
+            scope_prefix_map = {}
```
HyeockJinKim: It seems like AgentEtcdClientView should have a scope prefix map, even if it's not entered. Passing the scope_prefix_map every time doesn't seem like a good idea.

src/ai/backend/agent/etcd.py
```diff
@@ -0,0 +1,229 @@
...
+            error_detail=ErrorDetail.NOT_FOUND,
+        )
+
+
+# AgentEtcdClientView inherits from AsyncEtcd, but really it's just composing an AsyncEtcd instance
+# and acting as an adaptor. Inheritance is made only to make the type checker happy, and the places
+# that use AsyncEtcd really should not take the concrete implementation type AsyncEtcd, but rather
+# the interface type AbstractKVStore. In the current codebase, manually modifying places that
+# currently take in an AsyncEtcd instance to instead take in AbstractKVStore would be too invasive.
+class AgentEtcdClientView(AsyncEtcd, AbstractKVStore):
```
HyeockJinKim: AsyncEtcd inheritance doesn't seem appropriate either.

src/ai/backend/agent/etcd.py
hhoikoo: I agree. Done

src/ai/backend/agent/etcd.py
hhoikoo: Done

src/ai/backend/agent/etcd.py
hhoikoo: Removed

tests/agent/test_etcd.py
hhoikoo: Done

src/ai/backend/agent/etcd.py
```diff
@@ -0,0 +1,235 @@
...
+        self,
+        override: Optional[Mapping[ConfigScopes, str]],
+    ) -> Mapping[ConfigScopes, str]:
+        """
+        This stub ensures immutable usage of the ChainMap because ChainMap does *not*
+        have the immutable version in typeshed.
+        (ref: https://github.com/python/typeshed/issues/6042)
+        """
+        return ChainMap(
+            cast(MutableMapping, override) or {}, cast(MutableMapping, self._agent_scope_prefix_map)
```
HyeockJinKim: If override is None, then only self._agent_scope_prefix_map is returned.

src/ai/backend/agent/etcd.py
```diff
@@ -0,0 +1,235 @@
...
+    def _agent_scope_prefix_map(self) -> Mapping[ConfigScopes, str]:
+        """
+        This is kept as a @property method instead of a simple variable, because this way any
+        updates that are made to the config object (e.g. scaling group) is correctly applied as the
+        scope prefix mapping is recalculated every time.
+        """
+        return {
+            ConfigScopes.SGROUP: f"sgroup/{self._config.agent.scaling_group}",
+            ConfigScopes.NODE: f"nodes/agents/{self._config.agent.id}",
+        }
```
HyeockJinKim, line 40: It doesn't seem like it will be changed, so it seems appropriate to use cached_property.

HyeockJinKim: ~And rather than forcibly type casting the override, you'd input the type of Optional MutableMapping.~
Ah, this will break the existing etcd interface.

src/ai/backend/agent/etcd.py
hhoikoo: Good point - done

src/ai/backend/agent/etcd.py
hhoikoo: No I cannot cache this unfortunately. Scaling group on an agent can change after initialization e.g. through the control panel, and the etcd client must reflect this information.

HyeockJinKim (approved):

PR #6722, opened 2025-11-11
refactor(BA-3022): Make `error_code` method in `BackendAIError` as instance method

### LINKED ISSUE #6719: Change the error code method from a class method to an instance method
## Motivation

`BackendAIError` provides an `error_code` method to supply an error code for each error. Whilst the original intent was for each class to have its own unique error code, instances have arisen where the error code itself is injected, such as for bypassing errors. Therefore, we intend to change the error code to an instance method.

JIRA Issue: BA-3022

seedspirit:
resolves #6719 (BA-3022)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/tests/common/test_exception.py
+++ b/tests/common/test_exception.py
@@ -1,13 +1,6 @@
 import pytest
-from aiohttp import web
 
 from ai.backend.common.exception import (
-    BackendAIError,
-    ErrorCode,
-    ErrorDetail,
-    ErrorDomain,
-    ErrorOperation,
-    InvalidAPIParameters,
     UserNotFound,
 )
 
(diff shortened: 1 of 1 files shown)


tests/common/test_exception.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @pytest.mark.asyncio
+    async def test_multiple_instances_have_same_error_code(self) -> None:
+        """Test that different instances of the same error class return the same error code."""
+        error1 = InvalidAPIParameters(extra_msg="Missing parameter: foo")
+        error2 = InvalidAPIParameters(extra_msg="Missing parameter: bar")
+
+        # Both instances should return the same error code
+        assert error1.error_code() == error2.error_code()
+        assert str(error1.error_code()) == str(error2.error_code())
```
HyeockJinKim: This test doesn't seem necessary. It's ambiguous to call it an implementation requirement.

tests/common/test_exception.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @pytest.mark.asyncio
+    async def test_custom_error_subclass(self) -> None:
```
HyeockJinKim: Please do not test the code written within the test code. It is a meaningless test code in the current code.

HyeockJinKim (changes requested):

PR #6724, opened 2025-11-11
feat(BA-3024): Add custom resource alloc for agents in config

### LINKED ISSUE #6723: Add support for custom resource allocation in agent server config
## Motivation

- 

## Required Features

- 

## Impact

- 

## Testing Scenarios

- 

JIRA Issue: BA-3024

hhoikoo:
resolves #6723 (BA-3024)

SHARED mode allows all agents to see full resources (useful for stress testing). This is the same behavior as before. AUTO_SPLIT automatically divides resources equally among agents. MANUAL mode lets users specify exact per-agent allocations for all resources.

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/configs/agent/sample.toml
+++ b/configs/agent/sample.toml
@@ -16,14 +16,6 @@
   scaling-group = "default"
   # Scaling group type
   scaling-group-type = "compute"
-  # Allowed compute plugins
-  ## allow-compute-plugins = [ "ai.backend.accelerator.cuda_open", "ai.backend.activator.agent",]
-  # Blocked compute plugins
-  ## block-compute-plugins = [ "ai.backend.accelerator.mock",]
-  # Allowed network plugins
-  ## allow-network-plugins = [ "ai.backend.manager.network.overlay",]
-  # Blocked network plugins
-  ## block-network-plugins = [ "ai.backend.manager.network.overlay",]
   # Whether to force terminate abusing containers
   force-terminate-abusing-containers = false
   # Kernel creation concurrency
@@ -81,6 +73,14 @@
   metadata-server-bind-host = "0.0.0.0"
   # Metadata server port
   metadata-server-port = 40128
+  # Allowed compute plugins
+  ## allow-compute-plugins = [ "ai.backend.activator.agent", "ai.backend.accelerator.cuda_open",]
+  # Blocked compute plugins
+  ## block-compute-plugins = [ "ai.backend.accelerator.mock",]
+  # Allowed network plugins
+  ## allow-network-plugins = [ "ai.backend.manager.network.overlay",]
+  # Blocked network plugins
+  ## block-network-plugins = [ "ai.backend.manager.network.overlay",]
   # Path for image commit
   image-commit-path = "tmp/backend.ai/commit"
   # Path for abuse reports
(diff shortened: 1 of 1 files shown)


src/ai/backend/agent/config/unified.py
```diff
@@ -1312,6 +1394,43 @@ def validate(config: AgentSpecificConfig) -> None:
...
+            ]):
+                raise ValueError(
+                    "On MANUAL mode, config must specify cpu and mem resource allocations"
+                )
+
+        match self.resource.allocation_mode:
+            case ResourceAllocationMode.SHARED | ResourceAllocationMode.AUTO_SPLIT:
+                self._for_each_agent(validate_manual_resource_not_specified)
+            case ResourceAllocationMode.MANUAL:
+                self._for_each_agent(validate_mandatory_manual_resource_specified)
```
HyeockJinKim: As mentioned in the previous PR, using a closure here makes the function too large, so please handle this via a relevant class’s instance method instead.

Also, iterating over a list created by a specific action isn’t very clear—please update this for better readability as suggested before.

hhoikoo: done

configs/agent/sample.toml
```diff
@@ -409,24 +427,15 @@
...
-    # the minimum disk space from the scratch disk allocation via loopback files.
-    reserved-disk = 8589934592
-    # The alignment of the reported main memory size to absorb tiny deviations from
-    # per-node firmware/hardware settings. Recommended to be multiple of the
-    # page/hugepage size (e.g., 2 MiB).
-    memory-align-size = 16777216
-    # Resource allocation order
-    allocation-order = [ "cuda", "rocm", "tpu", "cpu", "mem",]
-    # Affinity policy
-    affinity-policy = 1
```
HyeockJinKim, line 432: It seems like the existing sample.toml should not be omitted.

src/ai/backend/agent/config/unified.py
```diff
@@ -1314,3 +1396,36 @@ def _validate_agent_configs(self) -> Self:
...
+                            "On non-MANUAL mode, config must not specify manual resource allocations"
+                        )
+
+            case ResourceAllocationMode.MANUAL:
+                for config in agent_configs:
+                    resource = config.resource
+                    if any([resource.allocated_cpu is None, resource.allocated_mem is None]):
+                        raise ValueError(
+                            "On MANUAL mode, config must specify cpu and mem resource allocations"
+                        )
```
HyeockJinKim, line 1415: It seems more appropriate to specify this at the type level. Allowing a type to be optional when it shouldn’t be, and then validating it at runtime, feels inconsistent.

configs/agent/sample.toml
```diff
@@ -156,6 +164,12 @@
   # Currently this value is unused. In future releases, it may be used to preserve
   # the minimum disk space from the scratch disk allocation via loopback files.
   reserved-disk = "8G"
```
hhoikoo, line 158: Here

configs/agent/sample.toml
hhoikoo: No these are not from existing config. The type of (sub) agent's resource used to be the full ResourceConfig, but we changed the type to OverridableResourceConfig, meaning that these fields no longer had to exist. These fields for the global-level resources are still available above

src/ai/backend/agent/config/unified.py
hhoikoo: I don't see a way of guaranteeing this at type-level without giving up on having a separate allocation mode enum, which in our current setup would be needed to specify shared or auto-split mode.

hhoikoo: Nevermind discussed offline and there is an elegant way of refactoring this. Done

src/ai/backend/agent/config/unified.py
```diff
@@ -867,6 +876,48 @@ class ContainerConfig(CommonContainerConfig, OverridableContainerConfig):
...
+
+    @model_validator(mode="after")
+    def validate_values_are_positive(self) -> Self:
+        if self.cpu is not None and self.cpu < 0:
+            raise ValueError(f"Allocated cpu must not be a negative value, but given {self.cpu}")
+        if self.mem is not None and self.mem < 0:
+            raise ValueError(f"Allocated mem must not be a negative value, but given {self.mem}")
+        if any(value < 0 for value in self.devices.values()):
+            raise ValueError("All allocated device resource values must not be a negative value")
+        return self
```
fregataa: It would be nice to let which value is negative

PR #6728, opened 2025-11-11
feat(BA-3026): Extract agent common resources to AgentRuntime

### LINKED ISSUE #6727: Extract common agent state management out to AgentRuntime
## Motivation

- 

## Required Features

- 

## Impact

- 

## Testing Scenarios

- 

JIRA Issue: BA-3026

hhoikoo:
resolves #6727 (BA-3026)

This change introduces AgentRuntime, which contains common resources shared by all agents. Currently the class is minimal but future multi agent changes will make more use of the class.

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/agent/server.py
+++ b/src/ai/backend/agent/server.py
@@ -48,6 +48,7 @@ from pydantic import ValidationError
 from setproctitle import setproctitle
 from zmq.auth.certs import load_certificate
 
+from ai.backend.agent.agent import AbstractAgent
 from ai.backend.agent.metrics.metric import RPCMetricObserver
 from ai.backend.agent.monitor import AgentErrorPluginContext, AgentStatsPluginContext
 from ai.backend.agent.resources import scan_gpu_alloc_map
@@ -97,6 +98,7 @@ from ai.backend.common.service_discovery.service_discovery import (
     ServiceMetadata,
 )
 from ai.backend.common.types import (
+    AgentId,
     ClusterInfo,
     CommitStatus,
     ContainerId,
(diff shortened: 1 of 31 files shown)


src/ai/backend/agent/server.py
```diff
@@ -377,14 +332,12 @@ async def __ainit__(self) -> None:
...
+        self.runtime = await AgentRuntime.new(
             self.local_config,
-            stats_monitor=self.stats_monitor,
-            error_monitor=self.error_monitor,
-            agent_public_key=self.rpc_auth_agent_public_key,
+            self.etcd,
+            self.stats_monitor,
+            self.error_monitor,
+            self.rpc_auth_agent_public_key,
         )
```
HyeockJinKim, line 388: It seems like we shouldn't change this right now. There were cases of injecting external code like plugins in the parts using importlib in the past, so we need to check this part as well.

src/ai/backend/agent/server.py
```diff
@@ -577,13 +533,14 @@ async def __aexit__(self, *exc_info) -> None:
...
-        await self.agent.shutdown(self._stop_signal)
+        await self.runtime.__aexit__(*exc_info)
         await self.stats_monitor.cleanup()
         await self.error_monitor.cleanup()
 
     @collect_error
     async def update_status(self, status):
-        await self.etcd.put("", status, scope=ConfigScopes.NODE)
+        etcd = self.runtime.get_etcd()
+        await etcd.put("", status, scope=ConfigScopes.NODE)
```
HyeockJinKim: If you want to view the runtime as a kind of service layer, it seems right to bypass the update request itself, rather than handling it by returning etcd.

src/ai/backend/agent/server.py
hhoikoo: reverted to using importlib

src/ai/backend/agent/server.py
hhoikoo: Moved the handling to runtime

PR #6763, opened 2025-11-14
refactor(BA-3048): Use resources functions directly in AbstractAgent

### LINKED ISSUE #6761: Extract resource functions out of agent implementations
## Motivation

- 

## Required Features

- 

## Impact

- 

## Testing Scenarios

- 

JIRA Issue: BA-3048

hhoikoo:
resolves #6761 (BA-3048)

This change replaces the abstractmethod implementation of load_resources and scan_available_resoruces methods, and replaces it by having the implementations in AbstractAgent directly import the agent backend resources module and call the functions.

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/changes/6766.feature.md
+++ b/changes/6766.feature.md
@@ -1,0 +1,1 @@
+Move ownership of resources from agent to a separate component in agent runtime
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -29,17 +29,19 @@ from collections.abc import (
 from contextlib import contextmanager
 from dataclasses import dataclass
 from decimal import Decimal
-from functools import cached_property
 from io import SEEK_END, BytesIO
+from itertools import chain
 from pathlib import Path
 from types import TracebackType
 from typing import (
     TYPE_CHECKING,
     Any,
+    Concatenate,
     Final,
     Generic,
     Literal,
     Optional,
+    ParamSpec,
     TypeVar,
     cast,
 )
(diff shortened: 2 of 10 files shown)


src/ai/backend/agent/dummy/__init__.py
```diff
@@ -1,8 +1,14 @@
...
 from .agent import DummyAgent
+from .resources import DummyResourceDiscovery
 
 
 def get_agent_cls() -> Type[AbstractAgent]:
     return DummyAgent
+
+
+def get_resource_discovery() -> AbstractResourceDiscovery:
+    return DummyResourceDiscovery()
```
HyeockJinKim: Hmm... I didn't really want to add even this to the import lib...

hhoikoo: I also dislike this approach, but I don't see a way of implementing this without using either importlib magic (like we do here now) or some match statement logic for selecting the correct ResourceDiscovery class. I think we don't have a dependency injection framework set up in this repo, so those two are the only options I think? Could you please tell me what way would be the best, or if there is an approach that I haven't thought of?

HyeockJinKim: It seems good to perform importlib only once and read based on the abc that was imported.
In the current case, logic that guarantees when one of the two imports is not provided needs to be included, but it doesn't seem appropriate.

hhoikoo: ah that sounds like a good approach. I've made a new ABC that encompasses both agent_cls and resource discovery + also there's an importlib usage in `server.py:prepare_krunner_volumes()`

hhoikoo: ah that sounds like a good approach. I've made a new ABC that encompasses both agent_cls and resource discovery + also there's an importlib usage in `server.py:prepare_krunner_volumes()`

PR #6770, opened 2025-11-14
feat(BA-2851): Add resource isolation options for multi-agent setup

### LINKED ISSUE #6432: Add resource partitioning for agents within the same agent runtime

hhoikoo:
resolves #6432 (BA-2851)

* SHARED: allows all agents to see full resources (useful for stress testing). This is the same behavior as before.
* AUTO_SPLIT: automatically divides resources equally among agents.
* MANUAL: lets users specify exact per-agent allocations for all resources.

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/agent/resources.py
+++ b/src/ai/backend/agent/resources.py
@@ -512,7 +512,7 @@ class ResourceAllocator(aobject):
             res = await self._calculate_agent_partition(agent_idx, agent_config)
             agent_computer, reserved_slots, resource_scaling_factor = res
 
-            agent_id = AgentId(agent_config.agent.id)
+            agent_id = AgentId(agent_config.agent.defaulted_id)
             agent_computers[agent_id] = agent_computer
             agent_reserved_slots[agent_id] = reserved_slots
             agent_resource_scaling_factor[agent_id] = resource_scaling_factor


src/ai/backend/agent/stats.py
```diff
@@ -449,28 +463,34 @@ async def collect_node_stat(self) -> None:
...
+                    log.error("collect_node_stat(): gather_node_measures() error", exc_info=res)
                     continue
+                slot_names, result = res
                 for node_measure in result:
                     metric_key = node_measure.key
                     # update node metric
+                    slot_scale_factor = self._extract_slot_measurement_scale_factor(
+                        resource_scaling_factors, slot_names, metric_key
+                    )
+                    per_node = node_measure.per_node.apply_scale_factor(slot_scale_factor)
```
HyeockJinKim, line 477: Isn't it okay to call `_extract_slot_measurement_scale_factor` outside the for loop?

PR #6800, opened 2025-11-17
fix(BA-2921): Disallow dot('.') usage in model service name

### LINKED ISSUE #6582: Add validation to prevent period (.) in model service names
## User Story

As a platform administrator, I want to prevent users from using period (`.`) characters in model service names so that HTTPS certificate validation works correctly.

## Background

When service names contain period (`.`) characters, it can cause HTTPS certificate validation issues. The frontend already has this validation, but the backend needs to enforce this restriction as well to prevent bypassing the validation through direct API calls.

## Acceptance Criteria

- [ ] Add validation rule to reject service names containing `.` (period) character
- [ ] Return appropriate error message when validation fails (e.g., "Service name cannot contain period (.) character")
- [ ] Apply validation to model service name creation and update endpoints
- [ ] Ensure existing services with `.` in names continue to work (backward compatibility)
- [ ] Add unit tests for the validation logic

## Technical Notes

- The validation should be implemented in the service name validation layer
- Consider using a regex pattern or simple string validation
- Error message should be clear and actionable for users

## Impact

This prevents potential HTTPS certificate issues and maintains consistency with frontend validation.

JIRA Issue: BA-2921

seedspirit:
resolves #6582 (BA-2921)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/docs/manager/rest-reference/openapi.json
+++ b/docs/manager/rest-reference/openapi.json
@@ -942,6 +942,9 @@
         "properties": {
           "name": {
             "description": "Name of the service",
+            "maxLength": 24,
+            "minLength": 4,
+            "pattern": "^\\w[\\w-]*\\w$",
             "title": "Name",
             "type": "string"
           },


src/ai/backend/common/typed_validators.py
```diff
@@ -170,18 +171,22 @@ class _NaiveTimeDurationPydanticAnnotation(_TimeDurationPydanticAnnotation):
...
+    r"^(?=.{4,%d}$)\w[\w-]*\w$" % (SESSION_NAME_MAX_LENGTH,), re.ASCII
 )
 
 
-def session_name_validator(s: str) -> str:
+def _session_name_validator(s: str) -> str:
     if SESSION_NAME_MATCHER.search(s):
         return s
     else:
-        raise AssertionError(f"String did not match {SESSION_NAME_MATCHER}")
```
HyeockJinKim, line 181: Isn't this a session name validator?

src/ai/backend/manager/api/service.py
```diff
@@ -478,6 +480,20 @@ class NewServiceRequestModel(LegacyBaseRequestModel):
...
+        service_name_matcher = re.compile(
+            r"^(?=.{4,%d}$)\w[\w-]*\w$" % (tv.SESSION_NAME_MAX_LENGTH,), re.ASCII
+        )
+        if not service_name_matcher.search(name):
+            raise InvalidAPIParameters(
+                f"Invalid service name format."
+                f"Service name must be 4-{tv.SESSION_NAME_MAX_LENGTH} characters long, and can only contain alphanumeric characters, underscores, and hyphens."
+                f"It must start and end with an alphanumeric character."
+            )
+        return name
```
HyeockJinKim: Let's check the length limit with the Field `ge` and make it only check the pattern so we can simply see the pattern matching.
In that case, it seems unnecessary to define a field_validator since it is already possible with just the pattern and ge in Field.

PR #6808, opened 2025-11-18
refactor(BA-3046): Return new `KnownContainerRegistry` dataclass instead of a dict when fetching Known Registries

### LINKED ISSUE #6758: Separate tasks requiring a project from those not requiring one in Container Registry-related operations
## Motivation

Currently, operations related to Container Registry simultaneously exist where the project is a mandatory parameter and where it is not. For instance, creating a Container Registry requires the project as a mandatory parameter, whereas operations like Rescan will rescan the registry with the same registry name if the project is not explicitly specified. This lack of clear separation in the business logic within the code is causing confusion. To prevent this, code separation based on the presence or absence of the project is required.

JIRA Issue: BA-3046

seedspirit:
resolves #6758 (BA-3046)

- [ ] Test case(s) to:
- [ ] Documentation

----
📚 Documentation preview 📚: https://sorna--6808.org.readthedocs.build/en/6808/

----
📚 Documentation preview 📚: https://sorna-ko--6808.org.readthedocs.build/ko/6808/

### DIFF
--- a/src/ai/backend/manager/data/container_registry/types.py
+++ b/src/ai/backend/manager/data/container_registry/types.py
@@ -38,3 +38,12 @@ class PerProjectContainerRegistryInfo:
     ssl_verify: bool
     is_global: bool
     extra: dict[str, Any]
+
+
+@dataclass
+class KnownContainerRegistry:
+    """Lightweight representation of a registered container registry for discovery purposes."""
+
+    project: str
+    registry_name: str
+    url: str
--- a/changes/6808.enhance.md
+++ b/changes/6808.enhance.md
@@ -1,0 +1,1 @@
+Return a dataclass instead of a dict when fetching Known Container Registries
--- a/src/ai/backend/manager/api/resource.py
+++ b/src/ai/backend/manager/api/resource.py
@@ -372,7 +372,11 @@ async def get_container_registries(request: web.Request) -> web.Response:
         )
     )
 
-    return web.json_response(result.registries, status=HTTPStatus.OK)
+    registries_dict = {
+        f"{registry.project}/{registry.registry_name}": registry.url
+        for registry in result.registries
+    }
+    return web.json_response(registries_dict, status=HTTPStatus.OK)
 
 
 def create_app(
--- a/src/ai/backend/manager/models/container_registry/row.py
+++ b/src/ai/backend/manager/models/container_registry/row.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 import re
 import uuid
-from collections.abc import Mapping, MutableMapping, Sequence
+from collections.abc import Sequence
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Self, cast
 from urllib.parse import urlparse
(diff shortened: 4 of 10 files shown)


src/ai/backend/manager/models/container_registry.py
```diff
@@ -235,12 +235,16 @@ async def get_known_container_registries(
...
+        query_stmt = (
+            sa.select(ContainerRegistryRow)
+            .options(
+                load_only(
+                    ContainerRegistryRow.project,
+                    ContainerRegistryRow.registry_name,
+                    ContainerRegistryRow.url,
+                )
             )
+            .order_by(ContainerRegistryRow.registry_name, ContainerRegistryRow.project)
```
HyeockJinKim: It is strange to perform "order by" without receiving any requests.

src/ai/backend/manager/data/container_registry/types.py
```diff
@@ -73,3 +73,10 @@ def fields_to_update(self) -> dict[str, Any]:
         self.ssl_verify.update_dict(to_update, "ssl_verify")
         self.extra.update_dict(to_update, "extra")
         return to_update
+
+
+@dataclass
+class ContainerRegistryLocationInfo:
+    project: Optional[str]
+    registry_name: str
+    url: str
```
fregataa, line 49:
```suggestion
    url: yarl.URL
```
Let's use URL type here

tests/manager/repositories/container_registry/test_container_registry_repository.py
```diff
@@ -919,3 +922,227 @@ async def test_modify_registry_remove_nonexistent_allowed_groups(
...
+
+        async with (
+            self.create_test_registry(database_engine, registry1_name, project1_name),
+            self.create_test_registry(database_engine, registry2_name, project2_name),
+        ):
+            # When
+            result = await repository.get_known_registries(querier=None)
+
+            # Then - Should return all known registries
+            assert len(result) >= 2
```
fregataa: Please check the length of results with `==`, not `>=`

src/ai/backend/manager/data/container_registry/types.py
```diff
@@ -73,3 +73,10 @@ def fields_to_update(self) -> dict[str, Any]:
         self.ssl_verify.update_dict(to_update, "ssl_verify")
         self.extra.update_dict(to_update, "extra")
         return to_update
+
+
+@dataclass
+class ContainerRegistryLocationInfo:
```
jopemachine: I think this is only used in legacy code
```suggestion
class LegacyContainerRegistryInfo:
```

src/ai/backend/manager/data/container_registry/types.py
```diff
@@ -38,3 +38,12 @@ class PerProjectContainerRegistryInfo:
     ssl_verify: bool
     is_global: bool
     extra: dict[str, Any]
+
+
+@dataclass
+class KnownContainerRegistry:
```
jopemachine, line 44: Do we really need to introduce this type?
Could we refactor the code based on ContainerRegistryData and the search logic instead?
(Of course, I know that this is legacy, but I don’t see a reason why the internal logic couldn’t be changed to be search-based.)

fregataa: If we already have ContainerRegistryData, let's use it rather than introducing a new type

seedspirit: As you mentioned, it seems there are many logics that can be unified using various filters (e.g., when the project is None, when it is not None, etc.) based on the single search logic. How about closing this issue but adding the search logic and modifying it so that diverse queries can be performed through filters? @jopemachine @fregataa  @HyeockJinKim

seedspirit:
Since it has been discussed to standardize using the search action, I am closing this PR.

PR #6813, opened 2025-11-18
feat(BA-3076): Ensure normal URLs are called even if the protocol is included in the host of `HostPortPair`

### LINKED ISSUE #6812: Prevent protocol inclusion in host fields and provide protocol-aware advertise URL
## Motivation

Currently, `connection_info.host` fields can receive protocol prefixes (http://, https://), causing errors when constructing URLs. The system needs to:

1. Strip protocols from host fields to store only hostname/IP
1. Provide advertise URLs with correct protocol based on TLS configuration

So we need to implemented protocol handling for appproxy coordinator

JIRA Issue: BA-3076

seedspirit:
resolves #6812 (BA-3076)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/tests/appproxy/coordinator/dependencies/bootstrap/test_config.py
+++ b/tests/appproxy/coordinator/dependencies/bootstrap/test_config.py
@@ -1,14 +1,178 @@
 from __future__ import annotations
 
+import os
+from dataclasses import dataclass
 from pathlib import Path
+from typing import Any
 from unittest.mock import MagicMock, patch
 
 import pytest
 
+from ai.backend.appproxy.common.config import HostPortPair
+from ai.backend.appproxy.coordinator.config import ProxyCoordinatorConfig
 from ai.backend.appproxy.coordinator.dependencies.bootstrap.config import (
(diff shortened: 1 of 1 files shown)


src/ai/backend/appproxy/coordinator/config.py
```diff
@@ -287,6 +287,19 @@ class ProxyCoordinatorConfig(BaseSchema):
...
+        """
+        Generate the full advertise URL with protocol.
+        Uses advertised_addr if set, otherwise falls back to bind_addr.
+        Protocol is determined by tls_advertised or tls_listen flags.
+        """
+        connection_info = (
+            self.advertised_addr if self.advertised_addr is not None else self.bind_addr
+        )
+        protocol = "https" if (self.tls_advertised or self.tls_listen) else "http"
+        return f"{protocol}://{connection_info.clean_host}:{connection_info.port}"
```
HyeockJinKim: Please add test code to match the scenarios being generated. (This would be good to set up with parametrize.)

PR #6903, opened 2025-11-24
feat(BA-2947): Insert new scopes to role when adding object permissions

### LINKED ISSUE #6619: Insert permission groups when any invite action occurs
Insert permission groups during any invite action to ensure the user can list the invited entity.

JIRA Issue: BA-2947

fregataa:
resolves #6619 (BA-2947)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/repositories/permission_controller/role_manager.py
+++ b/src/ai/backend/manager/repositories/permission_controller/role_manager.py
@@ -176,12 +176,15 @@ class RoleManager:
         """
 
         role = await self._query_system_role_by_user(db_session, user_id)
+        role_id = role.id
         entity_associated_scopes = await self._query_entity_associated_scopes(db_session, entity_id)
         await self._add_permission_groups_to_role_if_not_exist(
             db_session, role, entity_associated_scopes
         )
-        await self._add_object_permissions_to_role(db_session, role.id, entity_id, operations)
-        updated_role = await self._query_role_by_id(db_session, role.id)
+        await self._add_object_permissions_to_role(db_session, role_id, entity_id, operations)
+        await db_session.commit()
+        db_session.expire(role)
+        updated_role = await self._query_role_by_id(db_session, role_id)
         return updated_role
 
     async def _query_system_role_by_user(
@@ -218,7 +221,7 @@ class RoleManager:
         role_row = cast(Optional[RoleRow], role_row)
         if role_row is None:
             raise RoleNotFound(f"Role with id {role_id} not found")
-        return role_row
+        return role_row.to_data_with_permissions()
 
     async def _query_entity_associated_scopes(
         self,
@@ -248,6 +251,8 @@ class RoleManager:
         for permission_group in role_row.permission_group_rows:
             scope_id = permission_group.parsed_scope_id()
             scope_ids_to_add.discard(scope_id)
+        if not scope_ids_to_add:
+            return
         creators = [
             PermissionGroupCreator(
                 role_id=role_row.id,
--- a/tests/manager/repositories/rbac/BUILD
+++ b/tests/manager/repositories/rbac/BUILD
@@ -1,0 +1,6 @@
+python_tests(
+    name="tests",
+    dependencies=[
+        "src/ai/backend/manager:src",
+    ],
+)
--- a/tests/manager/repositories/rbac/test_role_manager.py
+++ b/tests/manager/repositories/rbac/test_role_manager.py
@@ -1,0 +1,203 @@
+"""Tests for RoleManager functionality."""
+
+from __future__ import annotations
+
+import uuid
+from collections.abc import AsyncGenerator, Iterable, Mapping
+from dataclasses import dataclass
+
+import pytest
+import sqlalchemy as sa
+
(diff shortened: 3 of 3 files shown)


src/ai/backend/manager/repositories/permission_controller/role_manager.py
```diff
@@ -163,13 +165,113 @@ async def add_object_permission_to_user_role(
...
 
+    async def _query_system_role_by_user(
+        self,
+        db_session: SASession,
+        user_id: uuid.UUID,
+    ) -> RoleRow:
+        role_row = await db_session.scalar(
+            sa.select(RoleRow)
+            .select_from(sa.join(RoleRow, UserRoleRow, RoleRow.id == UserRoleRow.role_id))
+            .where(sa.and_(UserRoleRow.user_id == user_id, RoleRow.source == RoleSource.SYSTEM))
```
HyeockJinKim, line 198: The basic assumption seems to be that only one system role is registered per user, but in the case of an admin role, even though it is a system role, object permissions shouldn't be registered; can that be distinguished?

HyeockJinKim: This is a question about whether the basic system role for users and a role for admins are distinguished. If it's difficult, it seems like we might have to immutably put something like a main role in the user. (Or put in a type called main role).

src/ai/backend/manager/repositories/permission_controller/role_manager.py
```diff
@@ -163,13 +165,113 @@ async def add_object_permission_to_user_role(
...
+        db_session: SASession,
+        role_id: uuid.UUID,
+        entity_id: ObjectId,
+        operations: Collection[OperationType],
+    ) -> None:
+        if not operations:
+            log.warning(
+                "no operations provided to add object permissions to role {}, skipping.", role_id
+            )
+            return
```
HyeockJinKim: This is a log that should not actually occur in properly written code, so if this log occurs, it should be fixed immediately.

fregataa: Should an error be raised here?

src/ai/backend/manager/repositories/permission_controller/role_manager.py
fregataa: It is assumed that all scopes has only one system sourced role, no matter whether the role is user's one or project's one

src/ai/backend/manager/repositories/permission_controller/role_manager.py
HyeockJinKim: The context was that errors don't need to be raised, but it's normal operation for errors to never occur, so logs generally shouldn't occur.

src/ai/backend/manager/repositories/permission_controller/role_manager.py
```diff
@@ -163,13 +165,113 @@ async def add_object_permission_to_user_role(
...
+        )
+        role_row = cast(Optional[RoleRow], role_row)
+        if role_row is None:
+            raise RoleNotFound(f"Role with id {role_id} not found")
+        return role_row
+
+    async def _query_entity_associated_scopes(
+        self,
+        db_session: SASession,
+        entity_id: ObjectId,
```
jopemachine, line 229:
```suggestion
        object_id: ObjectId,
```

jopemachine: The code uses the variable `entity_id` for `ObjectId` in several places - is there a specific reason for this?

Since `ObjectId` is a higher-level concept that includes `entity_id`, it seems better to call it `object_id`.

src/ai/backend/manager/repositories/permission_controller/role_manager.py
```diff
@@ -163,13 +165,113 @@ async def add_object_permission_to_user_role(
...
+                    AssociationScopesEntitiesRow.entity_id == entity_id.entity_id,
+                )
+            )
+        )
+        scope_entity_rows = cast(list[AssociationScopesEntitiesRow], raw_scope_entity_rows.all())
+        return [
+            ScopeId(scope_type=row.scope_type, scope_id=row.scope_id) for row in scope_entity_rows
+        ]
+
+    async def _add_permission_groups_to_role_if_not_exist(
```
jopemachine, line 244: Why not just `_add_permission_groups_to_role`?

src/ai/backend/manager/repositories/permission_controller/role_manager.py
```diff
@@ -163,13 +165,110 @@ async def add_object_permission_to_user_role(
...
+                scope_id=scope_id,
+            )
+            for scope_id in scope_ids_to_add
+        ]
+
+        rows = [PermissionGroupRow.from_input(creator) for creator in creators]
+        db_session.add_all(rows)
+        await db_session.flush()
 
+    async def _add_object_permissions_to_role(
```
jopemachine, line 268: IMHO, `_add_object_permissions_to_role` and `add_object_permission_to_user_role` have too similar names but differ significantly in their level of abstraction and what they do.

fregataa:
Closed this PR since it is outdated

Issue #6889, opened 2025-11-24
Update test code covering user resource policy with source based pattern


seedspirit:
## Motivation

Update test code covering user resource policy with source based pattern

JIRA Issue: BA-3118

PR #6907, opened 2025-11-24
refactor(BA-3117, BA-3118): Introduce source-based structure in user resource policy

### LINKED ISSUE #6888: Introduce DB Source in `UserResourcePolicy` repository
## Motivation

Currently, [http://Backend.AI](http://Backend.AI) 's Repository Layer adopts a Source-Based Structure design. Access to the database is managed via `db_source`, while cache access is handled by `cache_source`, and so forth. By separating and managing these sources, the Repository layer is solely responsible for combining and managing each source.

However, the current `UserResourcePolicy` Repository accesses database within the Repository class itself. We plan to separate the existing database access functionality into a DB Source layer to enable flexible addition of other sources in the future.

## Current Behavior

The current `UserResourcePolicy` Repository accesses the database directly within the Repository class. This results in the database access functionality being mixed with other features of the Repository, leading to a lack of flexibility when adding new sources.

## Expected Behavior

The `UserResourcePolicy` Repository should access the database through a DB Source layer. This will allow for the database access functionality to be managed separately, making it easier to add other sources in the future. Each source will be managed independently, and the Repository will be responsible for integrating these sources for use.

JIRA Issue: BA-3117

seedspirit:
resolves #6888, #6889 (BA-3117, BA-3118)

Apply the DB Source to the Repository within the User resource policy. The DB Source communicates with the actual database, fetching data that meets specified conditions and returning DTOs or Entities to the repository. This enables the Repository layer to combine various Sources without needing to know the detailed implementation of the DB, thereby returning data suitable for the Service layer.

This PR includes the introduction of the DB Source and the resulting modifications to the test code.
As the Repository's current role is not significant, tests communicating with the actual database are conducted within the Repository tests.
The existing integration tests overlap with the newly written tests and are less aligned with Backend.AI's test code writing patterns, so we intend to remove them.

- [x] Test case(s) to:
- [ ] Documentation

----
📚 Documentation preview 📚: https://sorna--6907.org.readthedocs.build/en/6907/

----
📚 Documentation preview 📚: https://sorna-ko--6907.org.readthedocs.build/ko/6907/

### DIFF
--- a/src/ai/backend/manager/repositories/user_resource_policy/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/user_resource_policy/db_source/db_source.py
@@ -105,8 +105,17 @@ class UserResourcePolicyDBSource:
     async def delete(self, name: str) -> UserResourcePolicyData:
         """Deletes a user resource policy."""
         async with self._db.begin_session() as db_sess:
-            query = sa.select(UserResourcePolicyRow).where(UserResourcePolicyRow.name == name)
-            row: Optional[UserResourcePolicyRow] = await db_sess.scalar(query)
+            delete_stmt = (
+                sa.delete(UserResourcePolicyRow)
+                .where(UserResourcePolicyRow.name == name)
+                .returning(UserResourcePolicyRow)
+            )
+            query_stms = (
+                sa.select(UserResourcePolicyRow)
+                .from_statement(delete_stmt)
+                .execution_options(populate_existing=True)
+            )
+            row: Optional[UserResourcePolicyRow] = await db_sess.scalar(query_stms)
             if row is None:
                 raise UserResourcePolicyNotFound(
                     f"User resource policy with name {name} not found."


src/ai/backend/manager/repositories/user_resource_policy/db_source/db_source.py
```diff
@@ -0,0 +1,115 @@
...
+        ),
+    ]
+)
+
+
+class UserResourcePolicyDBSource:
+    """
+    Database source for user resource policy operations.
+    Handles all database operations for user resource policies.
+    """
```
HyeockJinKim, line 43: Why is serializable applied here? It seems like applying Read Committed this time would be worthwhile, don't you think?

src/ai/backend/manager/repositories/user_resource_policy/db_source/db_source.py
```diff
@@ -0,0 +1,115 @@
...
+                .where(UserResourcePolicyRow.name == name)
+                .values(**fields)
+                .returning(UserResourcePolicyRow)
+            )
+            query_stmt = (
+                sa.select(UserResourcePolicyRow)
+                .from_statement(update_stmt)
+                .execution_options(populate_existing=True)
+            )
+            updated_row: Optional[UserResourcePolicyRow] = await db_sess.scalar(query_stmt)
```
HyeockJinKim, line 97: The query feels unnecessarily complicated.
It seems like using the returning value of the update would work, but is an additional select necessary?

src/ai/backend/manager/repositories/user_resource_policy/db_source/db_source.py
```diff
@@ -0,0 +1,115 @@
...
+    async def delete(self, name: str) -> UserResourcePolicyData:
+        """Deletes a user resource policy."""
+        async with self._db.begin_session() as db_sess:
+            query = sa.select(UserResourcePolicyRow).where(UserResourcePolicyRow.name == name)
+            row: Optional[UserResourcePolicyRow] = await db_sess.scalar(query)
+            if row is None:
+                raise UserResourcePolicyNotFound(
+                    f"User resource policy with name {name} not found."
+                )
+            await db_sess.delete(row)
```
HyeockJinKim: Could you check if there is a returning from delete?

src/ai/backend/manager/repositories/user_resource_policy/db_source/db_source.py
seedspirit: we need to use another select in SQLAlchemy 1.4
https://docs.sqlalchemy.org/en/14/orm/persistence_techniques.html#using-insert-update-and-on-conflict-i-e-upsert-to-return-orm-objects

We can exclude extra select after upgrading SQLAlchemy to 2.0
https://docs.sqlalchemy.org/en/20/orm/queryguide/dml.html#using-returning-with-update-delete-and-custom-where-criteria

src/ai/backend/manager/repositories/user_resource_policy/db_source/db_source.py
seedspirit: yes there is returning in delete, but we have to use extra select pattern for that

```python
        async with self._db.begin_session() as db_sess:
            delete_stmt = (
                sa.delete(UserResourcePolicyRow)
                .where(UserResourcePolicyRow.name == name)
                .returning(UserResourcePolicyRow)
            )
            query_stms = (
                sa.select(UserResourcePolicyRow)
                .from_statement(delete_stmt)
                .execution_options(populate_existing=True)
            )
            row: Optional[UserResourcePolicyRow] = await db_sess.scalar(query_stms)
```

src/ai/backend/manager/repositories/user_resource_policy/db_source/db_source.py
HyeockJinKim: It seems like applying it in 1.4 would be limited... Let's see after it goes over 2.0 for now.

src/ai/backend/manager/repositories/user_resource_policy/db_source/db_source.py
```diff
@@ -0,0 +1,124 @@
...
+        """Retrieves a user resource policy by name."""
+        async with self._db.begin_readonly_session() as db_sess:
+            query = sa.select(UserResourcePolicyRow).where(UserResourcePolicyRow.name == name)
+            row = await db_sess.scalar(query)
+            if row is None:
+                raise UserResourcePolicyNotFound(
+                    f"User resource policy with name {name} not found."
+                )
+            return row.to_dataclass()
+
```
HyeockJinKim, line 70: It seems like this could also be integrated into the querier.

tests/manager/repositories/user_resource_policy/test_user_resource_policy_repository.py
```diff
@@ -1,265 +1,256 @@
...
+
+    @pytest.fixture
+    async def sample_policy(
+        self, database_engine: ExtendedAsyncSAEngine
+    ) -> AsyncGenerator[UserResourcePolicyData, None]:
+        """Create a sample policy in the database for testing"""
+        policy_name = "test-policy-sample"
+        async with database_engine.begin_session() as db_sess:
+            policy_row = UserResourcePolicyRow(
+                name=policy_name,
```
HyeockJinKim, line 43: The name value should use a random value and should not use an arbitrary constant value in the test.

tests/manager/repositories/user_resource_policy/test_user_resource_policy_repository.py
```diff
@@ -1,265 +1,256 @@
...
+                UserResourcePolicyCreator(
+                    name=policy_name,
+                    max_vfolder_count=max_vfolder_count,
+                    max_quota_scope_size=max_quota_scope_size,
+                    max_session_count_per_model_session=max_session_count_per_model_session,
+                    max_customized_image_count=max_customized_image_count,
+                    max_vfolder_size=None,  # Deprecated
+                )
+            )
+            retrieved = await repository.get_by_name(created.name)
```
HyeockJinKim, line 241: It seems better to write tests for the case where you've already created it with a fixture, and tests for the case where you haven't, rather than creating and removing it here.

HyeockJinKim (approved):

PR #6913, opened 2025-11-25
fix(BA-3134): PureStorage client incorrectly using context variables

### LINKED ISSUE #6912: PurityClient ContextVar token causes ValueError in concurrent server requests
## Problem Description

When using PurityClient in a server environment with concurrent requests, a `ValueError` occurs:

```
ValueError: <Token var=<ContextVar name='auth_token' at 0x7f944d5c4680> at 0x7f944c185540> was created in a different Context
```

This happens because `PurityClient` stores the ContextVar token (`_auth_token_cvtoken`) as an instance variable. When `__aenter__` is called in one async context and `__aexit__` is called in a different context (common in aiohttp server handlers), `ContextVar.reset()` fails.

## Root Cause

In `src/ai/backend/storage/volumes/purestorage/purity.py`:

```python
class PurityClient:
    _auth_token_cvtoken: Token  # Stored as instance variable - problematic!
    
    async def __aenter__(self) -> PurityClient:
        # Token created in context A
        self._auth_token_cvtoken = self.auth_token.set(auth_token)
        return self
        
    async def __aexit__(self, *exc_info) -> None:
        # May be called in context B - causes ValueError
        self.auth_token.reset(self._auth_token_cvtoken)
```

## Proposed Solution

Separate the stateful session management from the client:

1. Create a new `PuritySession` class that acts as the async context manager
1. `PuritySession` manages its own auth token lifecycle within a single async context
1. `PurityClient` becomes a factory/holder for configuration, creating sessions on demand
1. Each API call creates its own session scope, ensuring token creation and reset happen in the same context

Example structure:

```python
class PuritySession:
    """Manages a single authenticated session within one async context."""
    def __init__(self, client: PurityClient) -> None:
        self._client = client
        self._auth_token: str | None = None
        
    async def __aenter__(self) -> PuritySession:
        # Login and store token locally (not in ContextVar)
        self._auth_token = await self._login()
        return self
        
    async def __aexit__(self, *exc_info) -> None:
        # Clean up - no ContextVar reset needed
        self._auth_token = None

class PurityClient:
    """Configuration holder and session factory."""
    def create_session(self) -> PuritySession:
        return PuritySession(self)
```

## Affected Component

- Storage Proxy
- Pure Storage volume driver

## Reproduction Steps

1. Run storage-proxy with Pure Storage backend
1. Send concurrent requests that trigger `get_fs_usage` or similar API calls
1. Observe the ValueError in logs

JIRA Issue: BA-3134

fregataa:
resolves #6912 (BA-3134)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/storage/volumes/purestorage/purity.py
+++ b/src/ai/backend/storage/volumes/purestorage/purity.py
@@ -36,16 +36,21 @@ class PurityClient:
 
     async def __aenter__(self) -> PurityClient:
         await self._lock.acquire()
-        async with self._session.post(
-            self.endpoint / "api" / "login",
-            headers={"api-token": self.api_token},
-            ssl=False,
-            raise_for_status=True,
-        ) as resp:
-            auth_token = resp.headers["x-auth-token"]
-            self._auth_token = auth_token
-            _ = await resp.json()
-        return self
+        try:
+            async with self._session.post(
+                self.endpoint / "api" / "login",
+                headers={"api-token": self.api_token},
+                ssl=False,
+                raise_for_status=True,
+            ) as resp:
+                auth_token = resp.headers["x-auth-token"]
+                self._auth_token = auth_token
+                _ = await resp.json()
+            return self
+        except Exception:
+            self._auth_token = None
+            self._lock.release()
+            raise
 
     async def __aexit__(self, *exc_info) -> None:
         self._auth_token = None


src/ai/backend/storage/volumes/purestorage/purity.py
```diff
@@ -26,39 +27,42 @@ def __init__(
...
+        await self._lock.acquire()
         async with self._session.post(
             self.endpoint / "api" / "login",
             headers={"api-token": self.api_token},
             ssl=False,
             raise_for_status=True,
         ) as resp:
             auth_token = resp.headers["x-auth-token"]
-            self._auth_token_cvtoken = self.auth_token.set(auth_token)
+            self._auth_token = auth_token
```
HyeockJinKim: It seems like if an exception occurs in `__aenter__`, the lock will not be released.

HyeockJinKim (approved):

PR #6987, opened 2025-11-27
feat(BA-3162): Exclude agent heartbeats from audit logs

### LINKED ISSUE #6966: Remove agent heartbeat handling observation from audit log
## Motivation

Remove agent heartbeat observations from audit log as it is infra layer behavior and occurs too often

JIRA Issue: BA-3162

seedspirit:
resolves #6966 (BA-3162)

## Summary

- Add a monitor-filtered action processor factory that resolves the monitor list at processor construction time.
- Use the factory for agent heartbeat processing so routine heartbeat events do not emit audit log records.
- Add unit coverage for excluding `AuditLogMonitor` from a constructed action processor.

## Tests

- `pants check src/ai/backend/manager/actions/processor/factory.py src/ai/backend/manager/services/agent/processors.py tests/unit/manager/actions/test_processor.py`
- `pants lint src/ai/backend/manager/actions/processor/factory.py src/ai/backend/manager/services/agent/processors.py tests/unit/manager/actions/test_processor.py`
- `pants test tests/unit/manager/actions/test_processor.py`

- [x] Test case(s) to:
- [ ] Documentation


HyeockJinKim (commented):
Shouldn't we be looking at the part where we're applying action monitors, rather than removing the spec?

HyeockJinKim (changes requested):
The implementation is wrong.

HyeockJinKim (changes requested):
This seems to be the wrong approach.

fregataa (commented):
1. Using negative phrasing like "SKIP_AUDIT_LOG" in enum names can lead to readability issues, especially when used in conditional statements. But in this case it does not harm readability that much
2. All monitors will be skipped during Heartbeat action handling. Is there any monitors that Heartbeat should be done?

HyeockJinKim:
I believe that the method is wrong, and it can be applied by excluding the monitor at the time of registration, but adding an additional flag field rather increases complexity. This is a problem that needs to be handled differently during registration, not by adding tags or the like and applying them differently after registration.

fregataa (commented):
In my opinion, the action should determine whether it should be excluded by a certain rather than the processor. How do you think? @HyeockJinKim

src/ai/backend/manager/actions/monitors/audit_log.py
```diff
@@ -61,4 +73,6 @@ async def prepare(self, action: BaseAction, meta: BaseActionTriggerMeta) -> None
 
     @override
     async def done(self, action: BaseAction, result: ProcessResult) -> None:
+        if action.spec() in self._excluded_action_specs:
+            return
```
HyeockJinKim: We should exclude it at the setup stage.

HyeockJinKim (changes requested):

src/ai/backend/manager/actions/processor/factory.py
```diff
@@ -0,0 +1,34 @@
...
+                continue
+            monitors.append(monitor)
+        return monitors
+
+    def build[TAction: BaseAction, TActionResult: BaseActionResult](
+        self,
+        func: Callable[[TAction], Awaitable[TActionResult]],
+    ) -> ActionProcessor[TAction, TActionResult]:
+        monitors = self._build_monitors()
+        return ActionProcessor(func, monitors)
```
HyeockJinKim: It doesn't look like this task will have the right level of abstraction, so let's just implement it in `processors` for now and add the abstraction later.

HyeockJinKim (changes requested):

HyeockJinKim (approved):

PR #6997, opened 2025-11-28
feat(BA-3184): Add scratch utils and scratch data schema

### LINKED ISSUE #6999: Add scratch data schema
Add scratch data schema to serialize and validate data stored in scratch directories

JIRA Issue: BA-3184

fregataa:
resolves #6999 (BA-3184)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/agent/kernel_registry/pickle/creator.py
+++ b/src/ai/backend/agent/kernel_registry/pickle/creator.py
@@ -1,10 +1,16 @@
+from __future__ import annotations
+
 from dataclasses import dataclass
 from pathlib import Path
+from typing import TYPE_CHECKING, Self
 
 from ai.backend.common.types import AgentId
 
 from ..loader.pickle import PickleBasedKernelRegistryLoader
 from ..writer.pickle import PickleBasedKernelRegistryWriter
+
+if TYPE_CHECKING:
+    from ai.backend.agent.agent import AgentClass
 
 
 @dataclass
@@ -12,6 +18,8 @@ class PickleBasedKernelRegistryCreatorArgs:
     scratch_root: Path
     ipc_base_path: Path
     var_base_path: Path
+
+    agent_class: AgentClass
     agent_id: AgentId
     local_instance_id: str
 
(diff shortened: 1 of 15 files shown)


src/ai/backend/agent/scratch/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    async def get_kernel_resource_spec(self) -> KernelResourceSpec:
+        filepath = self._resource_file_path()
+        raw_data = filepath.read_text()
+        return KernelResourceSpec.read_from_string(raw_data)
+
+    async def save_json_recovery_data(self, data: KernelRecoveryDataSchema) -> None:
+        filepath = self._json_recovery_file_path()
+        filepath.parent.mkdir(parents=True, exist_ok=True)
+        with open(filepath, "w") as f:
+            json.dump(data.model_dump(mode="json"), f)
```
jopemachine: I think we'd better use `aiofiles` here instead of performing synchronous operations

src/ai/backend/agent/scratch/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    await loop.run_in_executor(None, shutil.rmtree, str(scratch_dir))
+
+
+class ScratchUtils:
+    @staticmethod
+    def get_scratch_kernel_config_dir(scratch_root: Path, kernel_id: KernelId) -> Path:
+        return scratch_root / str(kernel_id) / "config"
+
+    @staticmethod
+    def list_kernel_id_and_config_path(scratch_root: Path) -> list[tuple[KernelId, Path]]:
```
jopemachine: This is a just suggestion, 
How about changing this method to 
`def scan_kernel_configs(scratch_root: Path) -> dict[KernelId, Path]:` for better readability?

fregataa: This function is not used anymore so its removed

src/ai/backend/agent/scratch/types.py
```diff
@@ -0,0 +1,94 @@
...
+
+from ..kernel import KernelLifecycleStatus, KernelOwnershipData
+from ..kernel_registry.types import KernelRecoveryData
+from ..proxy import DomainSocketPathPair
+from ..resources import KernelResourceSpec
+from ..types import AgentBackend
+
+
+class KernelRecoveryDataSchema(BaseModel):
+    """
```
jopemachine, line 16: It seems the `KernelRecoveryDataSchema` type name is inappropriate.

jopemachine: I think if you want to name it `KernelRecoveryDataSchema`,
one option is to have `KernelRecoveryData` inherit from `KernelRecoveryDataSchema`.

src/ai/backend/agent/kernel_registry/recovery/recovery.py
```diff
@@ -36,8 +46,8 @@ def __init__(
 
     @classmethod
     def create(cls, args: KernelRegistryRecoveryArgs) -> Self:
-        registry_file_name = f"kernel_registry.{args.agent_id}.dat"
-        fallback_registry_file_name = f"kernel_registry.{args.local_instance_id}.dat"
+        registry_file_name = f"last_registry.{args.agent_id}.dat"
+        fallback_registry_file_name = f"last_registry.{args.local_instance_id}.dat"
```
HyeockJinKim: Is there no problem even if the name changes?

src/ai/backend/agent/kernel_registry/types.py
```diff
@@ -60,3 +58,8 @@ class KernelRecoveryData(BaseModel):
     )
     resource_spec: KernelResourceSpec = Field(description="Resource specifications for the kernel")
     environ: Mapping[str, str] = Field(description="Environment variables for the kernel")
+
+    kernel_host: str = Field(
+        description="Host where the kernel is running. Could be an IP address or hostname.",
+        examples=["127.0.0.1"],
```
HyeockJinKim: How did you previously infer the host information of the kernel?

fregataa: I am not sure how this field is used

src/ai/backend/agent/kernel_registry/recovery/recovery.py
fregataa: The original name is `last_registry`

src/ai/backend/agent/docker/kernel.py
```diff
@@ -44,6 +44,8 @@
 DEFAULT_CHUNK_SIZE: Final = 256 * 1024  # 256 KiB
 DEFAULT_INFLIGHT_CHUNKS: Final = 8
 
+KERNEL_HOST = "127.0.0.1"
```
HyeockJinKim: Is there a reason to declare the kernel host as global?

src/ai/backend/agent/kernel_registry/types.py
HyeockJinKim: Does the kernel host value seem to not be used at all; is this field necessary?

src/ai/backend/agent/docker/kernel.py
jopemachine: It’s probably because the other global variables are marked as private. How about making all of them private in this PR as well?

src/ai/backend/agent/scratch/types.py
jopemachine: Most fields of the two types overlap. Is there a more suitable approach than inheritance? Is there any particular reason why inheritance wouldn’t be appropriate? @HyeockJinKim

src/ai/backend/agent/kernel_registry/recovery/recovery.py
jopemachine: What does ‘original name’ mean? Why was it changed from the previous name?

src/ai/backend/agent/scratch/utils.py
```diff
@@ -0,0 +1,108 @@
...
+            text = await file.read()
+        raw_data = load_json(text)
+        return KernelRecoveryScratchData.model_validate(raw_data)
+
+    async def get_kernel_environ(self) -> Mapping[str, str]:
+        result: dict[str, str] = {}
+        filepath = self._environ_file_path()
+        raw_data = filepath.read_text()
+        for line in raw_data.splitlines():
+            k, _, v = line.partition("=")
```
jopemachine, line 97: Don't we need to consider cases where whitespace is attached to the key or value?

fregataa: https://github.com/lablup/backend.ai/blob/b012e5f020401623500ed482e6eb4c1217b436c4/src/ai/backend/agent/docker/agent.py#L829-L840
we save the environ values without any whitespace

src/ai/backend/agent/kernel_registry/recovery/recovery.py
fregataa: https://github.com/lablup/backend.ai/blob/b012e5f020401623500ed482e6eb4c1217b436c4/src/ai/backend/agent/agent.py#L3801-L3828
it has never changed and he asked the question because he thought the filename is `kernel_registry.dat`

src/ai/backend/agent/kernel_registry/pickle/creator.py
```diff
@@ -0,0 +1,43 @@
...
+@dataclass
+class PickleBasedKernelRegistryCreatorArgs:
+    scratch_root: Path
+    ipc_base_path: Path
+    var_base_path: Path
+    agent_id: AgentId
+    local_instance_id: str
+
+
+class PickleBasedLoaderWriterCreator:
```
jopemachine, line 27: How about splitting the class into two separate ones — `PickleBasedLoaderCreator` and `PickleBasedWriterCreator`?

src/ai/backend/agent/kernel_registry/recovery/kubernetes_recovery.py
```diff
@@ -0,0 +1,46 @@
...
+from typing import TYPE_CHECKING, Self
+
+from ai.backend.common.types import AgentId
+from ai.backend.logging import BraceStyleAdapter
+
+from ..loader.pickle import PickleBasedKernelRegistryLoader
+from ..writer.pickle import PickleBasedKernelRegistryWriter
+from .base_recovery import BaseKernelRegistryRecovery
+
+if TYPE_CHECKING:
```
jopemachine, line 15: Empty TYPE_CHECKING

PR #7049, opened 2025-12-02
refactor(BA-3213): Replace `AgentDataExtended` to `AgentData`

### LINKED ISSUE #7048: Consolidate `AgentDataExtended` to `AgentData`
## Motivation

Currently, two types of data classes are used to represent `AgentRow`: `AgentData` and `AgentDataExtended`. This leads to confusion about when to use `AgentData` and when to use `AgentDataExtended`.

Upon detailed examination of the current codebase, it appears that the additional fields present in `AgentDataExtended` compared to `AgentData` are not being utilized. This suggests that `AgentDataExtended` will effectively become unnecessary once the following PR is merged.

The reason `AgentDataExtended` existed was due to the `running_kernel_occupied_slots` function, which utilizes the Agent's kernel relationship. However, as it is being removed in the PR below, it appears the `kernels` field within AgentExtendedData is no longer used separately, and `known_slot_types` also does not seem to be used independently.

[https://github.com/lablup/backend.ai/pull/6257/files](https://github.com/lablup/backend.ai/pull/6257/files) 

Therefore, we plan to consolidate by eliminating `AgentDataExtended` and using `AgentData` instead.

JIRA Issue: BA-3213

seedspirit:
resolves #7048 (BA-3213)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/changes/7049.enhance.md
+++ b/changes/7049.enhance.md
@@ -1,1 +1,1 @@
-Since `AgendDataExtended` no longer uses the fields it additionally extended from `AgentData`, replace instances using `AgendDataExtended` with `AgentData`.
+Since `AgentDataExtended` no longer uses the fields it additionally extended from `AgentData`, replace instances using `AgentDataExtended` with `AgentData`.


HyeockJinKim:
Please create an issue to write test code as a follow-up issue.

PR #7057, opened 2025-12-02
feat(BA-3209): Add `artifact_storages` common table

### LINKED ISSUE #7040: Create common table across the artifact storage tables
## Motivation

- If the tables have an inheritance structure, let’s add a common base table so that a single UUID can point to a single record across them.
- This can be useful in scenarios such as creating a cache or a stateful view for `artifact_storage`.

JIRA Issue: BA-3209

jopemachine:
resolves #7040 (BA-3209)

----
📚 Documentation preview 📚: https://sorna--7057.org.readthedocs.build/en/7057/

----
📚 Documentation preview 📚: https://sorna-ko--7057.org.readthedocs.build/ko/7057/


src/ai/backend/manager/models/vfs_storage/row.py
```diff
@@ -43,23 +49,27 @@ class VFSStorageRow(Base):
         primaryjoin=_get_vfs_storage_association_artifact_join_cond,
         overlaps="association_artifacts_storages_rows,object_storage_row",
     )
+    meta = relationship(
```
seedspirit: (Not necessarily required) How about changing the name to `descriptor` or `identity` instead of `meta`? It seems there's potential for confusion with SQLAlchemy's `MetaData`.

jopemachine:
I will request PR review after release 26.2.0

src/ai/backend/common/data/storage/exceptions.py
```diff
@@ -0,0 +1,21 @@
...
+class ArtifactStorageNotFoundError(BackendAIError, web.HTTPNotFound):
+    error_type = "https://api.backend.ai/probs/artifact-storage-not-found"
+    error_title = "Artifact Storage Not Found"
+
+    def error_code(self) -> ErrorCode:
+        return ErrorCode(
+            domain=ErrorDomain.ARTIFACT_STORAGE,
+            operation=ErrorOperation.READ,
+            error_detail=ErrorDetail.NOT_FOUND,
+        )
```
seedspirit, line 21: Wouldn't the exception in the common package be better than here? or we need to decide convention where to put exception classess @HyeockJinKim @fregataa

src/ai/backend/manager/models/alembic/versions/35dfab3b0662_add_artifact_storages_common_table.py
```diff
@@ -0,0 +1,174 @@
...
+
+def upgrade() -> None:
+    op.create_table(
+        "artifact_storages",
+        sa.Column("id", GUID(), server_default=sa.text("uuid_generate_v4()"), nullable=False),
+        sa.Column("name", sa.String(), nullable=False),
+        sa.Column("storage_id", GUID(), nullable=False),
+        sa.Column("type", sa.String(), nullable=False),
+        sa.PrimaryKeyConstraint("id", name=op.f("pk_artifact_storages")),
+        sa.UniqueConstraint("name", name=op.f("uq_artifact_storages_name")),
```
seedspirit, line 92: Is it acceptable to apply a global unique constraint to the artifact storage name? Could there be cases where the storage type differs but the names are identical? (As this is a policy matter, please verify this for me.)

src/ai/backend/manager/models/artifact_storages.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __str__(self) -> str:
+        return f"ArtifactStorageRow(id={self.id}, storage_id={self.storage_id}, type={self.type}, name={self.name})"
+
+    def __repr__(self) -> str:
+        return self.__str__()
+
+    def to_dataclass(self) -> ArtifactStorageData:
+        from ai.backend.common.types import ArtifactStorageId, ConcreteStorageId
+        from ai.backend.manager.data.artifact_storages.types import ArtifactStorageData
```
seedspirit: Is dynamic import necessary in this situation?

src/ai/backend/common/types.py
```diff
@@ -306,6 +306,10 @@ def check_typed_tuple(value: tuple[Any, ...], types: tuple[type, ...]) -> tuple[
 RuleId = NewType("RuleId", UUID)
 SessionId = NewType("SessionId", UUID)
 KernelId = NewType("KernelId", UUID)
+# ID of the `artifact_storages` common table (storage metadata).
+ArtifactStorageId = NewType("ArtifactStorageId", UUID)
+# ID of the concrete storage table (`object_storages`, `vfs_storages`, etc.).
+ConcreteStorageId = NewType("ConcreteStorageId", UUID)
```
seedspirit: It seems there's room for confusion about whether this is an Artifact Storage implementation or another storage implementation, so it would be better to give it a more specific name.

src/ai/backend/manager/models/alembic/versions/35dfab3b0662_add_artifact_storages_common_table.py
jopemachine: I think since artifact storage management is an admin-only feature, it doesn’t look awkward so far.
If, in the future, artifact storage management becomes available to regular users as well, it would make more sense to adjust the constraints at that point.

src/ai/backend/common/data/storage/exceptions.py
jopemachine: I placed it under manager because it’s only used for DB access, but considering a potential future introduction of a stateful layer, it seems better to put it under common.

src/ai/backend/common/types.py
jopemachine: I wiil rename it `ConcreteArtifactStorageId`

src/ai/backend/manager/models/artifact_storages.py
jopemachine: I will move those types to common, then we don't need them there

src/ai/backend/manager/models/artifact_storages.py
```diff
@@ -0,0 +1,81 @@
...
+        uselist=False,
+        viewonly=True,
+    )
+    vfs_storages: Mapped[VFSStorageRow | None] = relationship(
+        "VFSStorageRow",
+        back_populates="meta",
+        primaryjoin=_get_vfs_storage_join_condition,
+        uselist=False,
+        viewonly=True,
+    )
```
fregataa: Is it possible that both of them are None?

src/ai/backend/manager/models/artifact_storages.py
```diff
@@ -0,0 +1,81 @@
...
+def _get_vfs_storage_join_condition() -> sa.ColumnElement[bool]:
+    from ai.backend.manager.models.vfs_storage import VFSStorageRow
+
+    return VFSStorageRow.id == foreign(ArtifactStorageRow.storage_id)
+
+
+class ArtifactStorageRow(Base):  # type: ignore[misc]
+    """
+    Common information of all artifact storage records.
+    """
```
fregataa, line 27: Seems this row type contains a generic metadata of storage rows. Then how about using SQLAlchemy Table inheritance? https://docs.sqlalchemy.org/en/21/orm/inheritance.html

fregataa (commented):
I think that using SQLAlchemy Table inheritance would be great here.

src/ai/backend/manager/data/artifact_storages/types.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @override
+    def build_row(self) -> ArtifactStorageRow:
+        if self._storage_id is None:
+            raise InternalServerError("storage_id must be set before building row")
+        return ArtifactStorageRow(
+            name=self._name,
+            storage_id=self._storage_id,
+            type=self._storage_type,
+        )
```
fregataa: Why such check is needed?

src/ai/backend/manager/repositories/vfs_storage/db_source/db_source.py
```diff
@@ -28,42 +33,87 @@ async def get_by_name(self, storage_name: str) -> VFSStorageData:
...
         """
         async with self._db.begin_session() as db_session:
             creator_result = await execute_creator(db_session, creator)
-            return creator_result.row.to_dataclass()
+            new_row = creator_result.row
+
+            # Set the storage_id on the meta creator spec and create ArtifactStorageRow
+            meta_spec = meta_creator.spec
+            if not isinstance(meta_spec, ArtifactStorageCreatorSpec):
+                raise InternalServerError("meta_creator.spec must be ArtifactStorageCreatorSpec")
```
fregataa: Let's avoid the need of such check

fregataa (changes requested):

src/ai/backend/manager/repositories/vfs_storage/db_source/db_source.py
```diff
@@ -50,24 +50,41 @@ async def get_by_id(self, storage_id: uuid.UUID) -> VFSStorageData:
...
 
-    async def update(self, updater: Updater[VFSStorageRow]) -> VFSStorageData:
+    async def update(
+        self,
+        updater: Updater[VFSStorageRow],
+    ) -> VFSStorageData:
         """
         Update an existing VFS storage configuration in the database.
+        Uses plain UPDATE + SELECT instead of execute_updater's RETURNING+from_statement,
+        which is incompatible with JTI (RETURNING only includes child table columns).
```
fregataa, line 66: is it not allowed to update VFSStorageRow?

fregataa (commented):
Are there storage insert/update unit tests? It would be good to have them.
Plus, how do you think of [Single Table Inheritance](https://docs.sqlalchemy.org/en/21/orm/inheritance.html#single-table-inheritance)? Update and insert execution got complicated because of joined table inheritance, so I guess single table one makes it easy

jopemachine:
> Are there storage insert/update unit tests? It would be good to have them. Plus, how do you think of [Single Table Inheritance](https://docs.sqlalchemy.org/en/21/orm/inheritance.html#single-table-inheritance)? Update and insert execution got complicated because of joined table inheritance, so I guess single table one makes it easy

I believe STI would not good for extensibility because it will introduce many nullable columns.

jopemachine:
Since this PR has gone stale, I'll close it. This issue should be addressed when we redesign the storage proxy after the agent refactoring is complete.

PR #7094, opened 2025-12-03
feat(BA-3239): make overlay network's linux interface name consistent

### LINKED ISSUE #7096: make overlay network's linux interface name consistent
## Motivation

- Docker 28 introduced new feature to designate linux interface name for docker networks. It would be good if we can make use of such feature to make overlay network’s interface name consistent across different kernels.

## Required Features

- Update agent’s `OverlayNetworkPlugin` to provide fixed `com.docker.network.endpoint.ifname` label for overlay network

## Impact

- 

## Testing Scenarios

- Check if kernels in multi-container session on Docker 28 or newer all have same interface name for overlay network
- Check if agents with Docker 27 or under is not affected by the update

JIRA Issue: BA-3239

kyujin-cho:
Resolves #7096 (BA-3239). This PR updates agent to set linux interface name of the overlay network attached to kernel as `baimulti0`. Since this feature relies on libnet's `com.docker.network.endpoint.ifname` netlabel, which introduced on libnet 28.0, the interface remapping would also work only on agents backed by Docker 28 or later. Docker version check guard for the feature was skipped since docker engine seems to silently ignore unknown network labels upon request (see screenshots below).
image
image

### DIFF
--- a/changes/7094.feature.md
+++ b/changes/7094.feature.md
@@ -1,1 +1,1 @@
-Make overlay network's linux interface name consistent (applicable for [Docker 28+])(https://github.com/moby/moby/commit/6c3797923dcb082370a09f9381511da10120bd7b) only)
+Make overlay network's linux interface name consistent (applicable for [Docker 28+](https://github.com/moby/moby/commit/6c3797923dcb082370a09f9381511da10120bd7b) only)


achimnol (commented):
Wouldn't we need to make this new option gracefully omitted for Docker version 27 or older?

PR #7101, opened 2025-12-03
refactor(BA-3237): Explicitly raise error when Group operation fails

### LINKED ISSUE #7081: Explicitly raise error rather than catch error in GQL Resolvers in `Group`
## Motivation

Currently, for some APIs, even in exceptional situations, errors are not raised; instead, error conditions are communicated solely via string-based messages. This makes it difficult for clients to accurately determine the cause of failure. Furthermore, the messages themselves, such as 'failed', provide insufficient detail to understand why an exception occurred during the API call.

```python
    async def mutate(
        cls,
        root,
        info: graphene.ResolveInfo,
        name: str,
        props: GroupInput,
    ) -> CreateGroup:

        action = props.to_action(name)
        res = await graph_ctx.processors.group.create_group.wait_for_complete(action)

        return cls(
            ok=res.success,
            msg="success" if res.success else "failed",
            group=Group.from_dto(res.data) if res.success else None,
        )
```

To improve this, we plan to explicitly raise errors on the backend and provide clearer reasons through error messages.

JIRA Issue: BA-3237

seedspirit:
resolves #7081 (BA-3237)

- [x] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/common/exception.py
+++ b/src/ai/backend/common/exception.py
@@ -729,42 +729,6 @@ class DomainNotFound(BackendAIError, web.HTTPNotFound):
         )
 
 
-class GroupCreationError(BackendAIError, web.HTTPBadRequest):
-    error_type = "https://api.backend.ai/probs/group-creation-failed"
-    error_title = "Project Creation Failed"
-
-    def error_code(self) -> ErrorCode:
-        return ErrorCode(
-            domain=ErrorDomain.GROUP,
-            operation=ErrorOperation.CREATE,
-            error_detail=ErrorDetail.BAD_REQUEST,
-        )
-
-
-class GroupModificationError(BackendAIError, web.HTTPBadRequest):
-    error_type = "https://api.backend.ai/probs/group-modification-failed"
-    error_title = "Project Modification Failed"
-
-    def error_code(self) -> ErrorCode:
-        return ErrorCode(
-            domain=ErrorDomain.GROUP,
-            operation=ErrorOperation.UPDATE,
-            error_detail=ErrorDetail.BAD_REQUEST,
-        )
-
-
-class GroupDeletionError(BackendAIError, web.HTTPBadRequest):
-    error_type = "https://api.backend.ai/probs/group-deletion-failed"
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/models/gql_models/group.py
```diff
@@ -658,11 +660,10 @@ async def mutate(
 
         action = props.to_action(name)
         res = await graph_ctx.processors.group.create_group.wait_for_complete(action)
-
         return cls(
-            ok=res.success,
-            msg="success" if res.success else "failed",
-            group=Group.from_dto(res.data) if res.success else None,
+            ok=True,
+            msg="Project created successfully",
```
HyeockJinKim: The message needs to be aligned the same as before, right?

src/ai/backend/manager/repositories/group/repository.py
```diff
@@ -153,8 +192,8 @@ async def mark_inactive(self, group_id: uuid.UUID) -> bool:
                 .where(groups.c.id == group_id)
             )
             if result.rowcount > 0:
-                return True
-            raise GroupNotFound()
```
HyeockJinKim, line 157: It seems like "not found" is correct.

src/ai/backend/manager/repositories/group/repository.py
```diff
@@ -73,11 +78,44 @@ async def _get_group_by_id(self, session: SASession, group_id: uuid.UUID) -> Opt
...
+            policy_exists = await db_session.scalar(
+                sa.select(
+                    sa.exists().where(keypair_resource_policies.c.name == creator.resource_policy)
+                )
+            )
+            if not policy_exists:
+                raise InvalidAPIParameters(
+                    f"Cannot create group: Resource policy '{creator.resource_policy}' does not exist"
+                )
+
```
HyeockJinKim, line 100: The query separation doesn't look very good.
However, I will defer this to be worked on during the later sqlalchemy2.0 work or when cleaning up the repository layer.

PR #7120, opened 2025-12-04
feat(BA-3252): Add explicit `timeout` configuration argument to `StorageProxyHTTPClient`

### LINKED ISSUE #7121: Add client timeout configuration to `StorageProxyHTTPClient`
## Motivation

[Why is this Epic important? Clearly describe the context and reasons for creating this Epic.]

- 

## Objective

[What is the main goal of this Epic? Describe the desired outcome]

- 

## Expected Sub Issue

[Please list the subtasks you expect.]

- 

JIRA Issue: BA-3253

jopemachine:
resolves #7121 (BA-3252)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/clients/storage_proxy/base.py
+++ b/src/ai/backend/manager/clients/storage_proxy/base.py
@@ -135,7 +135,7 @@ class StorageProxyHTTPClient:
         *,
         body: Mapping[str, Any] | None = None,
         params: Mapping[str, Any] | None = None,
-        timeout: ClientTimeout,
+        timeout: ClientTimeout = DEFAULT_TIMEOUT,
     ) -> AsyncIterator[aiohttp.ClientResponse]:
         """
         Make an HTTP request using the session client.
@@ -178,7 +178,7 @@ class StorageProxyHTTPClient:
         *,
         body: Mapping[str, Any] | None = None,
         params: Mapping[str, Any] | None = None,
-        timeout: ClientTimeout,
+        timeout: ClientTimeout = DEFAULT_TIMEOUT,
     ) -> Optional[Mapping[str, Any]]:
         """
         Make an HTTP request using the session client.
@@ -206,7 +206,7 @@ class StorageProxyHTTPClient:
         *,
         body: Mapping[str, Any] | None = None,
         params: Mapping[str, Any] | None = None,
-        timeout: ClientTimeout,
+        timeout: ClientTimeout = DEFAULT_TIMEOUT,
     ) -> Mapping[str, Any]:
         """
         Make an HTTP request and return the response as a dictionary.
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/clients/storage_proxy/base.py
```diff
@@ -159,6 +164,10 @@ async def request_stream_response(
             raise StorageProxyConnectionError(
                 extra_msg="Failed to connect to storage proxy",
             ) from e
+        except asyncio.TimeoutError as e:
+            raise StorageProxyConnectionError(
+                extra_msg="Request to storage proxy timed out",
+            ) from e
```
HyeockJinKim, line 168: Does aiohttp client session raise asyncio.TimeoutError?

src/ai/backend/manager/clients/storage_proxy/base.py
```diff
@@ -131,13 +133,15 @@ async def request_stream_response(
         *,
         body: Mapping[str, Any] | None = None,
         params: Mapping[str, Any] | None = None,
+        timeout: Optional[ClientTimeout] = None,
```
HyeockJinKim: Instead of relying on aiohttp's internal implementation, how about we make our basic timeout settings non-optional to explicitly set them?

src/ai/backend/manager/clients/storage_proxy/base.py
```diff
@@ -159,6 +164,10 @@ async def request_stream_response(
             raise StorageProxyConnectionError(
                 extra_msg="Failed to connect to storage proxy",
             ) from e
+        except asyncio.TimeoutError as e:
+            raise StorageProxyConnectionError(
```
seedspirit: How about define new error type rather than using same `StorageProxyConnectionError` and distinguish it from other connection error?

src/ai/backend/manager/clients/storage_proxy/base.py
jopemachine: https://docs.aiohttp.org/en/stable/client_reference.html
The document says
`To catch all timeouts, including the total timeout, use asyncio.TimeoutError.`

src/ai/backend/manager/clients/storage_proxy/manager_facing_client.py
```diff
@@ -80,7 +80,7 @@ async def get_volumes(self) -> Mapping[str, Any]:
 
         :return: Response containing volume information
         """
-        return await self._client.request_with_response("GET", "volumes")
+        return await self._client.request_with_response("GET", "volumes", timeout=DEFAULT_TIMEOUT)
```
HyeockJinKim: If it's the default timeout, we provide it in request_with_response.

jopemachine: https://github.com/lablup/backend.ai/pull/7120#discussion_r2587433160
I think you suggested that we should explicitly add a timeout on the client side

jopemachine: Since we’re trying to let each request specify its own timeout, providing a default value would be less natural.
Because a follow-up PR will change it so that each request can configure its own timeout anyway.

jopemachine: Resolved

HyeockJinKim (approved):

PR #7125, opened 2025-12-04
feat(BA-3256): Add RBAC service and repository

### LINKED ISSUE #7124: Implement RBAC service and repository layers before API handlers
Complete the service and repository layer implementation for the RBAC before proceeding with API handler development.

JIRA Issue: BA-3256

fregataa:
resolves #7124 (BA-3256)

- [ ] Test case(s) to:
- [ ] Documentation

----
📚 Documentation preview 📚: https://sorna--7125.org.readthedocs.build/en/7125/

----
📚 Documentation preview 📚: https://sorna-ko--7125.org.readthedocs.build/ko/7125/

### DIFF
--- a/changes/7301.misc.md
+++ b/changes/7301.misc.md
@@ -1,0 +1,1 @@
+Relocate RBAC type definitions for better dependency management
--- a/src/ai/backend/common/api_handlers.py
+++ b/src/ai/backend/common/api_handlers.py
@@ -1,3 +1,4 @@
+import enum
 import functools
 import inspect
 import json
@@ -35,6 +36,15 @@ from .exception import (
     MiddlewareParamParsingFailed,
     ParameterNotParsedError,
 )
+
+
+class Sentinel(enum.Enum):
+    """A sentinel value to represent an unset value."""
+
+    TOKEN = enum.auto()
+
+
+SENTINEL = Sentinel.TOKEN
 
 
 class BaseRequestModel(BaseModel):
--- a/src/ai/backend/common/data/permission/types.py
+++ b/src/ai/backend/common/data/permission/types.py
@@ -1,0 +1,142 @@
+"""Common permission types shared between manager and client."""
+
+from __future__ import annotations
+
+import enum
+
+
+class PermissionStatus(enum.StrEnum):
+    ACTIVE = "active"
+    # 'inactive' status is used when the permission is temporarily disabled
+    INACTIVE = "inactive"
+    # 'deleted' status is used when the permission is permanently removed
+    DELETED = "deleted"
+
+
+class RoleStatus(enum.StrEnum):
+    """Status of a role."""
+
+    ACTIVE = "active"
+    # 'inactive' status is used when the role is temporarily disabled
+    INACTIVE = "inactive"
+    # 'deleted' status is used when the role is permanently removed
+    DELETED = "deleted"
+
+
+class RoleSource(enum.StrEnum):
+    """Definition source of the role."""
+
+    SYSTEM = "system"  # System-defined role, e.g., default roles
+    CUSTOM = "custom"  # Custom role defined
+
+
+class OperationType(enum.StrEnum):
+    CREATE = "create"
+    READ = "read"
+    UPDATE = "update"
+    SOFT_DELETE = "soft-delete"
+    HARD_DELETE = "hard-delete"
+    GRANT_ALL = "grant:all"  # Allow user to grant all permissions, including grant of grant
+    GRANT_READ = "grant:read"
+    GRANT_UPDATE = "grant:update"
+    GRANT_SOFT_DELETE = "grant:soft-delete"
+    GRANT_HARD_DELETE = "grant:hard-delete"
+
+    @classmethod
+    def owner_operations(cls) -> set[OperationType]:
+        """
+        Returns a set of operations that are considered owner operations.
+        Owner operations are those that allow full control over an entity.
+        """
+        return {op for op in cls}
+
+    @classmethod
+    def admin_operations(cls) -> set[OperationType]:
+        """
+        Returns a set of operations that are considered admin operations.
+        Admin operations are those that allow management of entities, including creation and deletion.
+        """
+        return {op for op in cls}
+
+    @classmethod
+    def member_operations(cls) -> set[OperationType]:
+        """
+        Returns a set of operations that are considered member operations.
+        Member operations are those that allow read access.
+        """
+        return {
+            cls.READ,
+        }
+
+
+class EntityType(enum.StrEnum):
+    USER = "user"
+    PROJECT = "project"
+    DOMAIN = "domain"
+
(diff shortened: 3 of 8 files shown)


src/ai/backend/manager/data/permission/role.py
```diff
@@ -123,3 +135,13 @@ class UserRoleAssignmentData:
     user_id: uuid.UUID
     role_id: uuid.UUID
     granted_by: Optional[uuid.UUID] = None
+
+
+@dataclass(frozen=True)
+class RoleListResult:
```
jopemachine, line 145: How about moving this to `types.py` like `notification`?
It will be difficult to locate if the type is in a different place each time

src/ai/backend/manager/data/permission/role.py
```diff
@@ -70,14 +70,26 @@ class RoleData:
...
+
+    user_id: uuid.UUID
+    username: str
+    email: str
+    granted_by: Optional[uuid.UUID]
+    granted_at: datetime
+
+
+@dataclass(frozen=True)
+class RoleDetailData:
```
jopemachine, line 67: It would be helpful if the comments explained the differences between RoleDetailData and RoleData.

src/ai/backend/manager/data/permission/role.py
```diff
@@ -70,14 +70,26 @@ class RoleData:
 
 
 @dataclass(frozen=True)
-class RoleDataWithPermissions:
+class AssignedUserInfo:
```
jopemachine: If there’s no particular reason not to, how about standardizing them with the Data postfix?

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
jopemachine, line 1: In other directories, we have a dedicated db_source directory with a db_source.py inside it.
It seems we should keep this structure consistent across the project. What do you think? @HyeockJinKim

src/ai/backend/manager/repositories/permission_controller/options.py
```diff
@@ -0,0 +1,105 @@
...
+        return inner
+
+
+class RoleOrders:
+    """Query orders for roles."""
+
+    @staticmethod
+    def name(ascending: bool = True) -> QueryOrder:
+        if ascending:
+            return RoleRow.name.asc()  # type: ignore[attr-defined]
```
jopemachine, line 89: Is `# type: ignore[attr-defined]` should be necessory?

src/ai/backend/manager/services/permission_contoller/actions/get_role_detail.py
```diff
@@ -0,0 +1,30 @@
...
+    role_id: UUID
+
+    @override
+    def entity_id(self) -> Optional[str]:
+        return str(self.role_id)
+
+    @override
+    @classmethod
+    def operation_type(cls) -> str:
+        return "get_detail"
```
jopemachine, line 21:
```suggestion
        return "get_role_detail"
```

jopemachine: Never mind — I initially thought the entity was permission_controller.
However, to avoid confusion, it might be better to rename the directory to role, just like in the other directories.
Not in this PR, but maybe later.

src/ai/backend/manager/data/permission/role.py
fregataa: A `role.py` file already exists and it has type schemas that other type schemas refer. so it will make confusion if we move this role specific definition to types.py

src/ai/backend/manager/repositories/permission_controller/options.py
fregataa: yes, type checks fails without it

src/ai/backend/manager/data/permission/role.py
```diff
@@ -70,14 +75,31 @@ class RoleData:
...
+    """
+
     id: uuid.UUID
     name: str
     source: RoleSource
     status: RoleStatus
 
     permission_groups: list[PermissionGroupExtendedData]
     object_permissions: list[ObjectPermissionData]
+    assigned_users: list[AssignedUserData]
```
HyeockJinKim: Let's provide this separately with a connection.

HyeockJinKim: If the value is of list type and can theoretically increase infinitely, please extract it via connection. (Not queried together)

src/ai/backend/manager/repositories/permission_controller/options.py
jopemachine: How about removing the `str` type annotation of `RoleRow.name`?
It seems that these annotations are causing unnecessary `# type: ignore` comments to appear repeatedly. How about removing the annotations, as we did in `NotificationChannelRow`?

src/ai/backend/manager/data/permission/role.py
```diff
@@ -70,7 +75,23 @@ class RoleData:
 
 
 @dataclass(frozen=True)
-class RoleDataWithPermissions:
+class AssignedUserData:
+    """Information about a user assigned to a role."""
+
+    user_id: uuid.UUID
+    username: str
+    email: str
```
HyeockJinKim: I doubt if any information other than user id will be needed.

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
```diff
@@ -390,17 +389,7 @@ async def search_users_assigned_to_role(
...
-                .select_from(
-                    sa.join(
-                        UserRow,
-                        UserRoleRow,
-                        UserRoleRow.user_id == UserRow.uuid,
-                    )
-                )
-                .where(UserRoleRow.role_id == role_id)
-            )
+            query = sa.select(UserRoleRow).where(UserRoleRow.role_id == role_id)
```
HyeockJinKim: I was trying not to care about `where`...

src/ai/backend/manager/services/permission_contoller/service.py
```diff
@@ -117,7 +117,6 @@ async def search_users_assigned_to_role(
     ) -> SearchUsersAssignedToRoleActionResult:
         """Search users assigned to a specific role with pagination and filtering."""
         result = await self._repository.search_users_assigned_to_role(
-            role_id=action.role_id,
             querier=action.querier,
```
HyeockJinKim: Does the existing Querier include the role_id condition?

HyeockJinKim: Please remove the role id from SearchUsersAssignedToRoleAction as well.

fregataa: SearchUsersAssignedToRoleAction uses role_id as an entity id.
Shouldn't we need at least a single id attribute for entity_id() ?

PR #7160, opened 2025-12-05
fix(BA-3276): Service try API allows resource value with float type

### LINKED ISSUE #7159: Invalid input failure when call Service try API
An invalid input failure exception occurs when trying to call the Service try API. It fails because the value of cuda.shares in config.resources is input as a float type value.

JIRA Issue: BA-3276

fregataa:
resolves #7159 (BA-3276)

- [ ] Test case(s) to:
- [ ] Documentation


fregataa:
Closed this PR because we decided to handle float values through string

PR #7170, opened 2025-12-08
chore: Update SBOM workflow to use cyclonedx-bom via pipx

Yaminyam:
Replaces the CycloneDX GitHub Action with direct installation of cyclonedx-bom using pipx and updates the SBOM generation step to output bom.json instead of bom.xml. Adjusts the artifact upload path accordingly.


HyeockJinKim (approved):
I approved this PR, but I'll merge it after 25.18.0 release

PR #7175, opened 2025-12-08
chore: Upgrade aiohttp to 3.13.0

Yaminyam:
Bump versions for several dependencies in requirements.txt, including aiohttp, aiodataloader, aioitertools, anyio, asyncpg, blessed, certifi, deprecated, fsspec, google-auth, googleapis-common-protos, graphql-core, and greenlet to address upstream updates and maintain compatibility.

Upgrading the version due to security vulnerabilities in the existing version.
https://nvd.nist.gov/vuln/detail/CVE-2025-53643

### DIFF
--- a/changes/7223.fix.md
+++ b/changes/7223.fix.md
@@ -1,0 +1,1 @@
+Add missing lock IDs to handlers with short cycle preventing race conditions caused by concurrent execution between short and long cycles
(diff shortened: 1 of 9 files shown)


HyeockJinKim:
It seems like the test is failing because the version was updated. Please check it once.

PR #7182, opened 2025-12-08
feat(BA-3281): Add `health_check_config` to Endpoint table

jopemachine:
resolves (BA-3281)

----
📚 Documentation preview 📚: https://sorna--7182.org.readthedocs.build/en/7182/

----
📚 Documentation preview 📚: https://sorna-ko--7182.org.readthedocs.build/ko/7182/


jopemachine:
Let’s make the model definition usable regardless of the runtime variant so we can keep this issue as simple as possible without requiring any WebUI changes. Later, when adding the configuration to the DB, we should make not only the health_check_config but the entire model definition overridable.

PR #7204, opened 2025-12-10
feat(BA-3281): Add model definition override feature with deep merge support

HyeockJinKim:
Add ModelDefinitionGeneratorRegistry.generate_model_definition() method that:
- Generates model definitions programmatically for non-CUSTOM variants
- Optionally fetches and deep merges override definitions from storage
- Falls back to generated definition on fetch failure

Changes:
- Add DeploymentConfig with enable_model_definition_override flag
- Add _deep_merge_dicts helper for recursive dict merging
- Update DeploymentExecutor to use new registry method
- Add comprehensive parametrized tests for all variants and scenarios

resolves #NNN (BA-MMM)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/tests/manager/sokovan/deployment/definition_generator/test_registry.py
+++ b/tests/manager/sokovan/deployment/definition_generator/test_registry.py
@@ -79,7 +79,7 @@ def create_override_dict(expected: VariantExpectation) -> dict[str, Any]:
         "port": expected.port,
     }
     if expected.health_check_path:
-        service["health_check"] = {
+        service["health-check"] = {
             "path": expected.health_check_path,
             "interval": 30.0,
             "initial-delay": 600.0,


src/ai/backend/manager/sokovan/deployment/definition_generator/registry.py
```diff
@@ -26,16 +33,40 @@
...
+log = BraceStyleAdapter(logging.getLogger(__spec__.name))
+
 
 @dataclass
 class RegistryArgs:
     deployment_repository: DeploymentRepository
+    enable_model_definition_override: bool = False
+
+
+def _deep_merge_dicts(base: dict[str, Any], override: dict[str, Any]) -> dict[str, Any]:
```
jopemachine: Why not using `deep_merge`?
https://github.com/lablup/backend.ai/blob/775b74394f47adfbba1b0099025d8bb576d5e77f/src/ai/backend/common/utils.py#L503

HyeockJinKim: I forgot this. 😄

src/ai/backend/manager/sokovan/deployment/definition_generator/registry.py
```diff
@@ -52,3 +83,76 @@ def get(self, runtime_variant: RuntimeVariant) -> ModelDefinitionGenerator:
...
+        # For other variants, generate first
+        generated_definition = await generator.generate_model_definition(model_revision)
+
+        # Check if override is enabled and path exists
+        if not self._enable_model_definition_override:
+            return generated_definition
+
+        model_definition_path = model_revision.mounts.model_definition_path
+        if not model_definition_path:
+            return generated_definition
```
seedspirit, line 101: In the Non Custom runtime variant, if there is no model definition path specified, shouldn't we use the default model definition path ("/models/model-definition.yaml", "/models/model-definition.yml") and allow the user to input their own model definition?

jopemachine: For now, only the default value will be available

HyeockJinKim: Since we are operating without webui support, we will only operate based on the default values for now.

src/ai/backend/manager/sokovan/deployment/definition_generator/registry.py
```diff
@@ -26,16 +34,26 @@
...
+if TYPE_CHECKING:
+    from ai.backend.manager.data.deployment.types import ModelRevisionSpec
+
+log = BraceStyleAdapter(logging.getLogger(__spec__.name))
+
 
 @dataclass
 class RegistryArgs:
     deployment_repository: DeploymentRepository
+    enable_model_definition_override: bool = False
```
jopemachine, line 46: I think we don't need to include this default in dataclass

PR #7219, opened 2025-12-11
fix(BA-3301): Implement client-IP filtering in Appproxy

jopemachine:
resolves (BA-3301)


HyeockJinKim (commented):
I think this implementation should be done at the L4 layer rather than this implementation.
Let's get to work on the new epic. @jopemachine

PR #7222, opened 2025-12-11
fix(BA-3322): remove unnecessary shmem pre-deduction from Memory cgroup

### LINKED ISSUE #7221: Remove unnecessary shmem pre-deduction from Docker Memory cgroup
## Problem

The Docker agent incorrectly pre-deducts shmem from Memory and MemorySwap cgroup limits, causing users to receive less memory than requested.

## Root Cause

In `src/ai/backend/agent/docker/agent.py:1157-1161`:

```python
self.computer_docker_args["HostConfig"]["MemorySwap"] -= shmem
self.computer_docker_args["HostConfig"]["Memory"] -= shmem
```

These lines are unnecessary because:

- shm (tmpfs) and app memory **share** the Memory cgroup space
- shm is already limited by ShmSize parameter
- Pre-deduction reduces cgroup size, resulting in less available memory

## Impact

When user requests 4GB memory + 1GB shmem:

- **Current (buggy):** cgroup = 3GB, so only 2GB available when shm is fully used
- **Expected:** cgroup = 4GB, so 3GB available when shm is fully used

## Solution

Remove the Memory/MemorySwap subtraction, only set ShmSize.

## References

- [Linux Kernel cgroup v1 Memory](https://docs.kernel.org/admin-guide/cgroup-v1/memory.html) - §2.3 first touch principle, §8.2 tmpfs/shm as file pages
- [Linux Kernel cgroup v2](https://docs.kernel.org/admin-guide/cgroup-v2.html) - shmem in memory.stat

JIRA Issue: BA-3322

HyeockJinKim:
Remove `Memory -= shmem` and `MemorySwap -= shmem` as they unnecessarily reduce the available memory. Docker's shm (tmpfs) and app memory share the Memory cgroup space, so pre-deducting shmem reduces the cgroup size itself, causing users to receive less memory than requested.

- shm is already limited by ShmSize parameter (tmpfs max size)
- shm + app together are limited by Memory cgroup
- Pre-deduction was double-counting the shmem allocation

Also add Shared Memory documentation to Agent README explaining:
- Docker memory architecture (shm/app share cgroup)
- OOM conditions (SIGKILL vs SIGBUS)
- Configuration options

resolves #7221 (BA-3322)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/changes/7222.fix.md
+++ b/changes/7222.fix.md
@@ -1,0 +1,1 @@
+remove unnecessary shmem pre-deduction from Memory cgroup


src/ai/backend/agent/README.md
```diff
@@ -224,6 +224,28 @@ VFolders are mounted to containers at runtime:
 - Detect OOM (Out-of-Memory) conditions
 - Enforce memory limits via cgroups
 
+### Shared Memory (shmem)
+Containers can request shared memory (`/dev/shm`) for inter-process communication.
+
+**Docker Memory Architecture**:
+- shm (tmpfs) and app memory share the Memory cgroup space
+- shm has an additional ShmSize limit (tmpfs maximum size)
+- Effective shm limit = `min(ShmSize, Memory cgroup available space)`
```
jopemachine, line 233: It seems like we need descriptions for variables such as shm and shmSize

HyeockJinKim: shm, shmsize would likely be understood as shared memory and shared memory size.

changes/7222.fix.md
```diff
@@ -0,0 +1 @@
+remove unnecessary shmem pre-deduction from Memory cgroup
```
jopemachine, line 1: Ideally it would be good also to explain why this change was introduced.
And I think maybe it was generated by claude, but it would be nice if the capitalization also followed news fragment conventions

PR #7233, opened 2025-12-11
fix(BA-3324): Add missing `raise_for_status` for huggingface downloader

### LINKED ISSUE #7228: Raise error when the huggingface repository is gated
## Summary

When downloading a HuggingFace model, instead of raising an error, the error message is being saved as a file.

JIRA Issue: BA-3324

jopemachine:
resolves #7228 (BA-3324)

### DIFF
--- a/tests/storage-proxy/artifacts/test_service.py
+++ b/tests/storage-proxy/artifacts/test_service.py
@@ -7,6 +7,7 @@ from unittest.mock import AsyncMock, MagicMock, Mock, patch
 
 import pytest
 from aiohttp import ClientError, ClientResponseError
+from aioresponses import aioresponses
 
 from ai.backend.common.bgtask.bgtask import BackgroundTaskManager, ProgressReporter
 from ai.backend.common.data.storage.registries.types import (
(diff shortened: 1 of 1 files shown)


src/ai/backend/storage/services/artifacts/huggingface.py
```diff
@@ -168,6 +168,8 @@ async def _probe_head(self) -> _ProbeHeadInfo:
         async with self._session.head(
             self._url, headers=headers_base, allow_redirects=True
         ) as resp:
+            resp.raise_for_status()
+
```
HyeockJinKim: If you always intended to raise, pass True to raise_for_status when creating the client session.

src/ai/backend/storage/services/artifacts/huggingface.py
```diff
@@ -193,19 +195,19 @@ async def read(self) -> AsyncIterator[bytes]:
...
 
         self._download_complete = False
         self._progress_task = None
 
-        # Probe head first - if this fails, we can't proceed
-        probe_info = await self._probe_head()
-        total = probe_info.total
-        etag = probe_info.etag
-        accept_ranges = probe_info.accept_ranges
-
```
HyeockJinKim, line 207: Is there a reason this action went inside the try? Wouldn't it be better to simply raise here?

jopemachine: You're right, this doesn't need to be included in the try statement since redis progress task is not created.

PR #7238, opened 2025-12-12
fix(BA-3329): `check_presets` API returning inflated remaining resources

### LINKED ISSUE #7237: `check_presets` returns inflated remaining resources by including non-ALIVE agents
## Summary

  The `check_presets` API returns inflated `remaining` resource values for scaling groups because it does not filter agents by `ALIVE` status. This causes the `remaining` values to potentially exceed `agent_stats.capacity`, leading to inconsistent resource reporting between the two APIs.

## Steps to Reproduce

1. Have a cluster with multiple agents in a scaling group
1. Cause one or more agents to become `LOST` or `TERMINATED` (e.g., agent heartbeat timeout)
1. Note that `schedulable` remains `true` and `available_slots` is not cleared when agent status changes
1. Query `agent_stats` to get `totalResource.capacity`
1. Query `check_presets` and sum up `scaling_groups[*].remaining` values
1. Compare the two values

## Expected Behavior

- `check_presets.scaling_groups.remaining` should only include resources from `ALIVE` agents
- The sum of `remaining` across all scaling groups should never exceed `agent_stats.capacity`
- Both APIs should use consistent agent filtering criteria

## Actual Behavior

- `check_presets` includes `LOST`/`TERMINATED` agents in its calculation because it only checks `available_slots IS NOT NULL AND schedulable == true`, without verifying `status == ALIVE`
- This causes `remaining` to be overestimated, potentially exceeding the actual cluster capacity
- Users may see available resources that don't actually exist, leading to failed session creation attempts

## Logs/Errors

  **Code comparison:**

  `agent_stats` query filter (`scheduler/db_source/db_source.py:3497-3498`):

```python
AgentRow.status == AgentStatus.ALIVE, AgentRow.schedulable == sa.true()

check_presets query filter (resource_preset/db_source/db_source.py:352-358):
AgentRow.scaling_group.in_(sgroup_names),
AgentRow.available_slots.isnot(None),
AgentRow.schedulable == sa.true(),
# Missing: AgentRow.status == AgentStatus.ALIVE

Root cause: When an agent transitions to LOST or TERMINATED status, the AgentStatusModifier only updates status, status_changed, and lost_at fields - it does not set schedulable = false or clear available_slots.
```

JIRA Issue: BA-3329

jopemachine:
resolves #7237 (BA-3329)

### DIFF
--- a/tests/manager/repositories/resource_preset/test_check_presets.py
+++ b/tests/manager/repositories/resource_preset/test_check_presets.py
@@ -8,10 +8,10 @@ Addresses BA-3054: Wrong parse of inference metrics.
 """
 
 import uuid
-from collections.abc import AsyncGenerator, Callable, Coroutine
+from collections.abc import AsyncGenerator
 from datetime import datetime, timedelta
 from decimal import Decimal
-from typing import Any, Optional
+from typing import Optional
 from unittest.mock import AsyncMock, MagicMock
 
 import pytest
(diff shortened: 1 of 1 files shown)


tests/manager/repositories/resource_preset/test_check_presets.py
```diff
@@ -334,42 +335,72 @@ async def test_keypair_access_key(
...
         test_scaling_group_name: str,
-    ) -> AsyncGenerator[AgentId, None]:
-        """Create test agent and return agent ID"""
-        agent_id = AgentId(f"agent-{uuid.uuid4().hex[:8]}")
+    ) -> Callable[..., Coroutine[Any, Any, AgentId]]:
+        """Factory fixture to create agents with specified status and resources."""
+        # Assume agents length exceeds 255
+        _addr_counter = [0]
+
+        async def _create_agent(
```
HyeockJinKim: Create an _create_agent method.
Instead, it looks good to see the agent creation itself as a situation, and create a fixture that makes an agent.

tests/manager/repositories/resource_preset/test_check_presets.py
```diff
@@ -899,3 +930,43 @@ async def test_ignores_cached_occupied_slots_in_agent_row(
...
+        """
+        # Create ALIVE agents (should be counted)
+        await create_agent()  # 16 CPU, 32GB
+        await create_agent()  # 16 CPU, 32GB
+
+        # Create non-ALIVE agents (should be excluded)
+        non_alive_slots = ResourceSlot({"cpu": Decimal("100"), "mem": Decimal("204800")})
+        await create_agent(status=AgentStatus.LOST, available_slots=non_alive_slots)
+        await create_agent(status=AgentStatus.TERMINATED, available_slots=non_alive_slots)
+        await create_agent(status=AgentStatus.RESTARTING, available_slots=non_alive_slots)
```
HyeockJinKim: This `given` logic in `given - then` pattern should be in fixture.

tests/manager/repositories/resource_preset/test_check_presets.py
```diff
@@ -333,43 +334,101 @@ async def test_keypair_access_key(
...
+        """Private factory method to create agents with specified status and resources."""
+        _addr_counter = [0]
+
+        async def _create(
+            *,
+            status: AgentStatus = AgentStatus.ALIVE,
+            available_slots: Optional[ResourceSlot] = None,
+            occupied_slots: Optional[ResourceSlot] = None,
+            schedulable: bool = True,
+        ) -> AgentId:
```
HyeockJinKim: Can't you just call it without the factory?

HyeockJinKim (approved):

PR #7254, opened 2025-12-13
feat(BA-3330): Add deployment handler to reconcile desired replicas and routes

### LINKED ISSUE #7246: Handle scaling for ready deployments with mismatched replica count
**Problem**

Currently, the autoscaling rule updates a deployment's desired replica count and transitions it to the `scaling` state. The scheduler then processes deployments in the `scaling` state by comparing their desired replica count with the number of routes.

However, there's a gap in this logic: if a deployment is in the `ready` state and its desired replica count doesn't match its route count, no scaling action is triggered.

**Solution**

Add a handler that transitions deployments from `ready` to `scaling` state when their desired replica count differs from their number of routes. This ensures the scheduler can properly process these deployments for scaling.

JIRA Issue: BA-3330

fregataa:
resolves #7246 (BA-3330)
### Summary
Implemented `ReconcileDeploymentHandler` to handle autoscaling for ready deployments with mismatched replica counts.

### Changes
- Added `ReconcileDeploymentHandler` that transitions ready deployments to scaling state when their desired replica count differs from their route count

### Problem
Previously, deployments in the ready state with replica-route mismatches were not being scaled because no state transition was triggered.

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/defs.py
+++ b/src/ai/backend/manager/defs.py
@@ -113,7 +113,6 @@ class LockID(enum.IntEnum):
     LOCKID_DEPLOYMENT_CHECK_PENDING = 226  # For operations checking PENDING sessions
     LOCKID_DEPLOYMENT_CHECK_REPLICA = 227  # For operations checking REPLICA sessions
     LOCKID_DEPLOYMENT_DESTROYING = 228  # For operations destroying deployments
-    LOCKID_DEPLOYMENT_RECONCILE = 229  # For operations reconciling deployments
     # Sokovan target status locks (prevent concurrent operations on same status)
     LOCKID_SOKOVAN_TARGET_PENDING = 230  # For operations targeting PENDING sessions
     LOCKID_SOKOVAN_TARGET_PREPARING = 231  # For operations targeting PREPARING/PULLING sessions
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/sokovan/deployment/executor.py
```diff
@@ -146,6 +146,31 @@ async def check_pending_deployments(
...
+        for deployment in deployments:
+            routes = route_map[deployment.id]
+            if len(routes) != deployment.replica_spec.target_replica_count:
+                log.info(
+                    "Deployment {} has mismatched active routes: expected {}, found {}",
+                    deployment.id,
+                    deployment.replica_spec.target_replica_count,
+                    len(routes),
+                )
+                successes.append(deployment)
```
HyeockJinKim: It might be a little awkward that the replica count and intent count are different and it still shows as success, so how about putting it in errors?

src/ai/backend/manager/defs.py
```diff
@@ -113,6 +113,7 @@ class LockID(enum.IntEnum):
     LOCKID_DEPLOYMENT_CHECK_PENDING = 226  # For operations checking PENDING sessions
     LOCKID_DEPLOYMENT_CHECK_REPLICA = 227  # For operations checking REPLICA sessions
     LOCKID_DEPLOYMENT_DESTROYING = 228  # For operations destroying deployments
+    LOCKID_DEPLOYMENT_RECONCILE = 229  # For operations reconciling deployments
```
HyeockJinKim: When not using a short interval, there seems to be no need to have a lock id. (Because only one cycle runs)

src/ai/backend/manager/sokovan/deployment/handlers/reconcile.py
```diff
@@ -0,0 +1,70 @@
+"""Handler for checking and managing deployment replicas."""
```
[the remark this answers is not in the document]
HyeockJinKim: Please reflect this too.

PR #7302, opened 2025-12-15
chore: Refactor Dockerfile locations and enhance SBOM workflow

Yaminyam:
Moved and renamed Dockerfiles to the docker/ directory for better organization. Updated build scripts to use new Dockerfile paths. Enhanced the GitHub Actions SBOM workflow to generate SBOMs for both Python dependencies and all container images using a matrix strategy.


HyeockJinKim (commented):
Is it okay to move the dockerfile? I need to check because I don't know how it was used.

Yaminyam:
> Is it okay to move the dockerfile? I need to check because I don't know how it was used.

We are not currently building images in the pipeline using these Dockerfiles.
All build scripts in use have had their paths updated accordingly.

PR #7329, opened 2025-12-16
chore: Run transitive tests only for merge_group events in CI

Yaminyam:
Updates the CI workflow to use the --changed-dependents=transitive option with pants test only when the GitHub event is merge_group, reducing the test scope for other events.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -323,27 +323,34 @@ jobs:
         cache-lmdb-store: 'true'
     - name: Test
       timeout-minutes: 15
+      env:
+        IS_RELEASE: ${{ needs.optimize-ci.outputs.release }}
       run: |
         # configure redis sentinel cluster hostnames for testing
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -338,7 +338,12 @@ jobs:
         else
           BASE_REF="HEAD~1"
         fi
-        pants test --changed-since=$BASE_REF --changed-dependents=transitive -- -v
+        # Use --changed-dependents=transitive only for merge_group to run broader tests
+        DEPENDENTS_OPT=""
+        if [ "$GITHUB_EVENT_NAME" == "merge_group" ]; then
+          DEPENDENTS_OPT="--changed-dependents=transitive"
+        fi
+        pants test --changed-since=$BASE_REF $DEPENDENTS_OPT -- -v
```
HyeockJinKim: This seems like it will be a problem because it only works directly in a release situation.

PR #7363, opened 2025-12-16
feat(BA-3379): Add RBAC API handlers

### LINKED ISSUE #7309: Implement Role update REST API
Implement REST API endpoint for updating existing roles in the RBAC system.

## Endpoint

- `PATCH /roles/{role_id`}

## Requirements

- Allow updating role name, description, and permissions
- Validate updated role data
- Handle non-existent role errors
- Return updated role information

JIRA Issue: BA-3379

fregataa:
resolves #7309 (BA-3379)

- [ ] Test case(s) to:
- [ ] Documentation

----
📚 Documentation preview 📚: https://sorna--7363.org.readthedocs.build/en/7363/

----
📚 Documentation preview 📚: https://sorna-ko--7363.org.readthedocs.build/ko/7363/

### DIFF
--- a/docs/manager/rest-reference/openapi.json
+++ b/docs/manager/rest-reference/openapi.json
@@ -10229,6 +10229,211 @@
         ],
         "description": "Validate a notification rule by rendering its template with test data.\n\n**Preconditions:**\n* User privilege required.\n"
       }
+    },
+    "/v2.0/admin/rbac/roles": {
+      "post": {
+        "operationId": "v2.0.create_role",
+        "tags": [
+          "v2.0"
+        ],
+        "responses": {
+          "200": {
+            "description": "Successful response"
+          }
+        },
+        "security": [
+          {
+            "TokenAuth": []
+          }
+        ],
+        "parameters": [],
+        "description": "Create a new role.\n\n**Preconditions:**\n* User privilege required.\n"
+      }
+    },
+    "/v2.0/admin/rbac/roles/search": {
+      "post": {
+        "operationId": "v2.0.search_roles",
+        "tags": [
+          "v2.0"
+        ],
+        "responses": {
+          "200": {
+            "description": "Successful response"
+          }
+        },
+        "security": [
+          {
+            "TokenAuth": []
+          }
+        ],
+        "parameters": [],
+        "description": "Search roles with filters, orders, and pagination.\n\n**Preconditions:**\n* User privilege required.\n"
+      }
+    },
+    "/v2.0/admin/rbac/roles/{role_id}": {
+      "get": {
+        "operationId": "v2.0.get_role",
+        "tags": [
+          "v2.0"
+        ],
+        "responses": {
+          "200": {
+            "description": "Successful response"
+          }
+        },
+        "security": [
+          {
+            "TokenAuth": []
+          }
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/api/rbac/handler.py
```diff
@@ -0,0 +1,352 @@
...
+        return APIResponse.build(status_code=HTTPStatus.OK, response_model=resp)
+
+
+def create_app(
+    default_cors_options: CORSOptions,
+) -> Tuple[web.Application, Iterable[WebMiddleware]]:
+    """Create aiohttp application for RBAC API endpoints."""
+    app = web.Application()
+    app["api_versions"] = (4, 5)
+    app["prefix"] = "rbac"
```
HyeockJinKim: If it’s an API for super admin, how about taking the prefix rule a little differently?

.github/workflows/backport.yml
```diff
@@ -109,7 +109,7 @@ jobs:
           result="[]"
           for branch in "${target_branches[@]}"; do
             if [ -n "$branch" ]; then
```
jopemachine, line 111: I think this is not an intended change.

src/ai/backend/manager/api/rbac/handler.py
fregataa: Then how about `admin-rbac`?

HyeockJinKim: The intention was to have a separate admin prefix and handle things there.
For example, requests that are delivered to /v2/admin/... can all be considered as admin api.

src/ai/backend/manager/api/rbac/handler.py
```diff
@@ -0,0 +1,345 @@
...
+    async def search_roles(
+        self,
+        body: BodyParam[SearchRolesRequest],
+        processors_ctx: ProcessorsCtx,
+    ) -> APIResponse:
+        """Search roles with filters, orders, and pagination."""
+        processors = processors_ctx.processors
+        me = current_user()
+        if me is None or not me.is_superadmin:
+            raise web.HTTPForbidden(reason="Only superadmin can search roles.")
```
jopemachine: Let's define an exception type

src/ai/backend/manager/api/rbac/handler.py
```diff
@@ -0,0 +1,345 @@
...
+        processors = processors_ctx.processors
+        me = current_user()
+        if me is None or not me.is_superadmin:
+            raise NotEnoughPermission("Only superadmin can delete roles.")
+
+        # Create purger
+        purger: Purger[RoleRow] = Purger(row_class=RoleRow, pk_value=path.parsed.role_id)
+
+        # Call service action
+        await processors.permission_controller.delete_role.wait_for_complete(
```
jopemachine, line 204: How about passing updater instead of purger in case of soft-delete?

src/ai/backend/manager/api/rbac/handler.py
fregataa: I tried to set the prefix to "v2/admin/rbac/" but aiohttp does not register the app with the prefix. It should be registered as separate subapps.

PR #7397, opened 2025-12-17
fix(BA-3442): Remove `read-write` validation for `read-only` model vfolder

jopemachine:
resolves (BA-3442)

### DIFF
--- a/tests/manager/repositories/vfolder/test_vfolder_repository.py
+++ b/tests/manager/repositories/vfolder/test_vfolder_repository.py
@@ -6,7 +6,6 @@ Tests the repository layer with real database operations.
 from __future__ import annotations
 
 import uuid
-from dataclasses import dataclass, field
 from typing import AsyncGenerator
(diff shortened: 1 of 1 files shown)


tests/manager/repositories/vfolder/test_vfolder_repository.py
```diff
@@ -0,0 +1,324 @@
...
+            for policy_name in cleanup_tracker.user_policy_names:
+                await db_sess.execute(
+                    sa.delete(UserResourcePolicyRow).where(
+                        UserResourcePolicyRow.name == policy_name
+                    )
+                )
+
+            # 5. Domain (referenced by User and Group)
+            for domain_name in cleanup_tracker.domain_names:
+                await db_sess.execute(sa.delete(DomainRow).where(DomainRow.name == domain_name))
```
HyeockJinKim: Since it only persists for the fixture lifecycle anyway, let's just delete everything rather than bothering to check before deleting.

PR #7462, opened 2025-12-18
fix(BA-3448): Allow duplicated session names across users

### LINKED ISSUE #7412: Allow duplicate session names across different users
## Background

Currently, the `ix_sessions_unique_name_nonterminal` unique index on the sessions table enforces global uniqueness of session names across all users for non-terminal sessions (status NOT IN 'ERROR', 'TERMINATED', 'CANCELLED'). This constraint is defined in `src/ai/backend/manager/models/session.py` and prevents different users from using the same session name, even though their sessions are completely separate.

 

## Problem

Users cannot create sessions with the same name as another user's active session, which limits naming flexibility and can cause confusion when multiple users want to use common session names like "dev", "test", "training", or "experiment-1". This global uniqueness constraint is unnecessarily restrictive since session names only need to be unique within each user's scope.

  

## Required Features

1. Modify unique index to include user_uuid
   - Change `ix_sessions_unique_name_nonterminal` from `(name)` to `(name, user_uuid)`
   - Rename index to `ix_sessions_unique_name_per_user_nonterminal` for clarity
   - Keep the PostgreSQL WHERE clause: `status NOT IN ('ERROR', 'TERMINATED', 'CANCELLED')`
1. Create Alembic migration
   - Drop existing `ix_sessions_unique_name_nonterminal` index
   - Create new `ix_sessions_unique_name_per_user_nonterminal` index with composite key
1. Verify query logic
   - Ensure session lookup queries use both name and user_uuid where appropriate
   - Check GraphQL resolvers, API handlers, and service layer methods

## Impact

- **Positive**: Users gain naming flexibility - can use any session name without conflicts from other users
- **Positive**: More intuitive behavior - session names naturally scoped to user ownership
- **Positive**: Backward compatible - existing sessions remain valid after migration
- **Neutral**: No API changes required - session name uniqueness already enforced per user in practice
- **Risk**: Low - Index change is safe, only affects constraint validation

  

## Testing Scenarios

1. **Basic duplicate names across users**
   - User A creates session "my-session" → Success
   - User B creates session "my-session" → Success (previously would fail)
   - User A tries to create another "my-session" → Fail with duplicate error
1. **Session status transitions with duplicate names**
   - User A has "test" session (RUNNING)
   - User B has "test" session (RUNNING)
   - User A terminates their "test" → Only User A's session terminated
   - User B's "test" session unaffected
1. **Session lifecycle with duplicate names**
   - User A creates "experiment-1" → Success
   - User B creates "experiment-1" → Success
   - User A cancels "experiment-1"
   - User A creates "experiment-1" again → Success (old one is CANCELLED)
1. **Query correctness**
   - List sessions by user → Returns correct sessions per user
   - Get session by name + user → Returns correct session
   - Filter sessions by name → Returns all sessions with that name across users

 

## Files to Modify

- `src/ai/backend/manager/models/session.py` - Update Index definition
- `src/ai/backend/manager/models/alembic/versions/` - Add new migration file
- Potentially affected query logic in repositories and services

JIRA Issue: BA-3448

seedspirit:
resolves #7412 (BA-3448)

- [x] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/tests/manager/models/test_session.py
+++ b/tests/manager/models/test_session.py
@@ -46,17 +46,18 @@ class TestSessionUniqueNamePerUser:
         self, database_engine: ExtendedAsyncSAEngine
     ) -> AsyncGenerator[DomainRow, None]:
         """Create test domain."""
-        domain = DomainRow(name="test-domain")
+        domain = DomainRow(name=f"test-{uuid.uuid4()}")
 
-        async with database_engine.begin_session() as db_sess:
-            db_sess.add(domain)
-            await db_sess.flush()
+        try:
+            async with database_engine.begin_session() as db_sess:
+                db_sess.add(domain)
+                await db_sess.flush()
 
-        yield domain
-
-        # Clean up domain data
-        async with database_engine.begin() as conn:
-            await conn.execute(sa.delete(DomainRow).where(DomainRow.name == domain.name))
+            yield domain
+        finally:
+            # Clean up domain data
+            async with database_engine.begin() as conn:
+                await conn.execute(sa.delete(DomainRow).where(DomainRow.name == domain.name))
 
     @pytest.fixture
     async def user_policy(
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/models/alembic/versions/a4289ef5f0cd_allow_duplicated_session_names_across_.py
```diff
@@ -0,0 +1,50 @@
...
+        "ix_sessions_unique_name_nonterminal",
+        table_name="sessions",
+        postgresql_where="((status)::text <> ALL ((ARRAY['ERROR'::character varying, 'TERMINATED'::character varying, 'CANCELLED'::character varying])::text[]))",
+    )
+    op.create_index(
+        "ix_sessions_unique_name_per_user_nonterminal",
+        "sessions",
+        ["name", "user_uuid"],
+        unique=True,
+        postgresql_where=sa.text("status NOT IN ('ERROR', 'TERMINATED', 'CANCELLED')"),
```
fregataa, line 31: The condition statement looks different from the original one. Could you check to ensure that it works as the same?

tests/manager/models/test_session.py
```diff
@@ -0,0 +1,334 @@
...
+
+        async with database_engine.begin_session() as db_sess:
+            db_sess.add(domain)
+            await db_sess.flush()
+
+        yield domain
+
+        # Clean up domain data
+        async with database_engine.begin() as conn:
+            await conn.execute(sa.delete(DomainRow).where(DomainRow.name == domain.name))
```
fregataa: Does this need to be handled in a finally block?

src/ai/backend/manager/models/alembic/versions/a4289ef5f0cd_allow_duplicated_session_names_across_.py
HyeockJinKim: Is it not a problem since it's making names unique per user, and the problem is ensuring the uniqueness of the name for the entire service right now?

tests/manager/models/test_session.py
HyeockJinKim: When tidying up in a fixture, it's better to tidy up with finally.

HyeockJinKim (approved):

PR #7468, opened 2025-12-18
fix(BA-2767): Misuses `strawberry` Dataloader of Model Artifact service

jopemachine:
resolves (BA-2767)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/changes/7468.fix.md
+++ b/changes/7468.fix.md
@@ -1,0 +1,1 @@
+Fix misuses strawberry Dataloader of Model Artifact service


src/ai/backend/manager/api/gql/artifact/resolver.py
```diff
@@ -169,17 +189,28 @@ async def artifacts(
...
     edges = []
     for artifact_data in action_result.data:
-        registry_meta = await registry_meta_loader.load(artifact_data.registry_id)
-        source_registry_meta = await registry_meta_loader.load(artifact_data.source_registry_id)
+        registry_data = await registry_loader.load(artifact_data.registry_id)
+        source_registry_data = await registry_loader.load(artifact_data.source_registry_id)
+        if registry_data is None:
+            raise ArtifactRegistryNotFoundError(f"Registry {artifact_data.registry_id} not found")
+        if source_registry_data is None:
+            raise ArtifactRegistryNotFoundError(
```
seedspirit, line 203: Wouldn't it be better to throw errors at the service layer or repository layer rather than at the handler level?

jopemachine: I agree, but this is not the goal of this PR.

HyeockJinKim (approved):

PR #7632, opened 2025-12-30
feat(BA-3561): Add pre-validation for user modification to prevent raw IntegrityError

jopemachine:
resolves BA-3561

  ## Summary

  - Add pre-validation checks in `update_user_validated` to prevent database IntegrityError from being exposed to users
  - Introduce `UserModificationConflict` and `UserModificationBadRequest` error classes for user-friendly error messages during user updates

  ## Problem

  When modifying a user with values that violate database constraints (unique or foreign key), the system exposed raw PostgreSQL `IntegrityError` messages like:

  (sqlalchemy.dialects.postgresql.asyncpg.IntegrityError) <class 'asyncpg.exceptions.UniqueViolationError'>:
  duplicate key value violates unique constraint "uq_users_username"
  DETAIL: Key (username)=(user) already exists.

  This error message is unfriendly and leaks internal database details.

  ## Solution

  Added pre-validation checks before executing the update query:

  | Field | Constraint | Error Class |
  |-------|-----------|-------------|
  | `username` | Unique | `UserModificationConflict` |
  | `domain_name` | FK → domains | `UserModificationBadRequest` |
  | `resource_policy` | FK → user_resource_policies | `UserModificationBadRequest` |

  ### New Error Classes

  - `UserModificationConflict`: For unique constraint violations (operation: UPDATE, detail: CONFLICT)
  - `UserModificationBadRequest`: For foreign key reference errors (operation: UPDATE, detail: BAD_REQUEST)

  ### New Helper Methods

  - `_check_username_exists_for_other_user`: Checks if username is taken by another user
  - `_check_resource_policy_exists`: Checks if resource policy exists

### DIFF
--- a/changes/7635.fix.md
+++ b/changes/7635.fix.md
@@ -1,1 +1,0 @@
-Fix `unified_config` incorrectly assigned as tuple in `ServiceConfigNode.load`


jopemachine:
The existing user repository code is incorrect.
However, fixing everything in this PR would unnecessarily increase its size.
Let’s address those issues in a follow-up PR.
Test code also will be added in that PR.

src/ai/backend/manager/repositories/user/repository.py
```diff
@@ -195,6 +197,33 @@ async def update_user_validated(
...
+                    raise UserModificationBadRequest(f"Domain '{new_domain_name}' does not exist.")
+
+            # Check if new resource_policy exists
+            new_resource_policy = updater_spec.resource_policy.optional_value()
+            if new_resource_policy and new_resource_policy != current_user.resource_policy:
+                policy_exists = await self._check_resource_policy_exists(conn, new_resource_policy)
+                if not policy_exists:
+                    raise UserModificationBadRequest(
+                        f"Resource policy '{new_resource_policy}' does not exist."
+                    )
```
HyeockJinKim, line 225: Here, it seems appropriate to parse SQLAlchemy's IntegrityError since a timing issue during the inspection could result in an integrity error.

HyeockJinKim (approved):
It seems better to parse the error based on IntegrityError, but for now, it doesn't appear to be a problem so I'll merge it.

PR #7653, opened 2025-12-31
test(BA-3589): add tests for manager's auth api handler

### LINKED ISSUE #7624: Add comprehensive tests to track authorization states (pre-modification baseline)
## Motivation

- Before modifying the authorization system, we must establish a baseline. Without it:
- Changes could break permissions silently
- We cannot verify what actually changed
- Debugging failures becomes difficult
- Authorization changes must be auditable for security
- This task creates a "golden snapshot" ensuring safe, trackable modifications.

## Objective

- Captures baseline authorization states for all roles
- Tracks authorization state transitions
- Validates behavior before/after modifications
- Enables safe refactoring with regression detection
- Provides audit trails for compliance

## Details

- Reverage /test files to effectively perform unit test.
- Analyze coverage of pre-defined tests.

## Impact

- Benefits overall coverage of test.
- Build components for test automation.

JIRA Issue: BA-3589

kwonkwonn:
resolves #7624 (BA-3589)

Increase test coverage before migrating api/auth.py to follow pydantic model.
This Pr specifically contains 'validate_ip' and 'validate decorator' which will be used before and after migration of auth.py 

- [v] Mention to the original issue
- [v] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/tests/unit/manager/api/auth/test_handlers.py
+++ b/tests/unit/manager/api/auth/test_handlers.py
@@ -1,0 +1,656 @@
+"""
+Tests for auth.py API handlers.
+
+TODO: Currently auth decorators (auth_required) are bypassed
+      by mocking request.get(). This should be refactored to use proper middleware
+      integration for more realistic testing.
+"""
+
+from __future__ import annotations
+
+import json
+import uuid
+from datetime import UTC, datetime
+from http import HTTPStatus
+from typing import Any, cast
+from unittest.mock import AsyncMock, MagicMock, PropertyMock
+
+import pytest
+from aiohttp import web
+
+from ai.backend.manager.api.auth import (
+    authorize,
+    generate_ssh_keypair,
+    get_role,
+    get_ssh_keypair,
+    signout,
+    signup,
+    update_full_name,
+    update_password,
+    update_password_no_auth,
+    upload_ssh_keypair,
+)
+from ai.backend.manager.data.auth.types import AuthorizationResult, SSHKeypair
+from ai.backend.manager.errors.auth import AuthorizationFailed
+from ai.backend.manager.models.user import UserRole, UserStatus
+from ai.backend.manager.services.auth.actions.authorize import AuthorizeActionResult
+from ai.backend.manager.services.auth.actions.generate_ssh_keypair import (
+    GenerateSSHKeypairActionResult,
+)
+from ai.backend.manager.services.auth.actions.get_role import GetRoleActionResult
+from ai.backend.manager.services.auth.actions.get_ssh_keypair import GetSSHKeypairActionResult
+from ai.backend.manager.services.auth.actions.signout import SignoutActionResult
+from ai.backend.manager.services.auth.actions.signup import SignupActionResult
+from ai.backend.manager.services.auth.actions.update_full_name import UpdateFullNameActionResult
+from ai.backend.manager.services.auth.actions.update_password import UpdatePasswordActionResult
+from ai.backend.manager.services.auth.actions.update_password_no_auth import (
+    UpdatePasswordNoAuthActionResult,
+)
+from ai.backend.manager.services.auth.actions.upload_ssh_keypair import UploadSSHKeypairActionResult
+
+# ---------------------------------------------------------------------------
+# Fixtures
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture
+def mock_root_ctx() -> MagicMock:
+    """RootContext mock with processors."""
+    return MagicMock()
+
+
+@pytest.fixture
+def unauthorized_request(mock_root_ctx: MagicMock) -> MagicMock:
+    """Mock request for unauthorized user."""
+    req = MagicMock(spec=web.Request)
+    req.app = {"_root.context": mock_root_ctx}
+    req.get = lambda k, default=None: {
+        "is_authorized": False,
+        "is_superadmin": False,
+    }.get(k, default)
+    return req
+
+
+@pytest.fixture
+def authorized_request(mock_root_ctx: MagicMock) -> MagicMock:
+    """Mock request for authorized user with POST body support."""
+    req = MagicMock(spec=web.Request)
+    req.app = {"_root.context": mock_root_ctx}
+    req.get = lambda k, default=None: {
+        "is_authorized": True,
+        "is_superadmin": False,
+        "is_admin": False,
+    }.get(k, default)
+    type(req).can_read_body = PropertyMock(return_value=True)
+    req.method = "POST"
+    req.content_type = "application/json"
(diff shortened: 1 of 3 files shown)


tests/component/manager/api/test_auth.py
```diff
@@ -192,3 +206,184 @@ async def do_authorize(hash_type, api_version):
...
+    request.headers = {"X-Forwarded-For": "10.0.0.50"}
+    request.remote = None
+
+    # Empty allowlist allows all IPs
+    user: dict[str, Any] = {"allowed_client_ip": None}
+    validate_ip(request, user)
+
+    # IP within CIDR range is allowed
+    user = {"allowed_client_ip": [ReadableCIDR("10.0.0.0/24", is_network=True)]}
+    validate_ip(request, user)
```
HyeockJinKim: Using parametrization makes it easy to view multiple cases at once, so please use parametrization.

tests/component/manager/api/test_auth.py
```diff
@@ -192,3 +206,184 @@ async def do_authorize(hash_type, api_version):
...
+
+        @auth_required
+        async def handler(request: web.Request) -> web.Response:
+            return web.Response(text="ok")
+
+        request = MagicMock(spec=web.Request)
+        request.get = MagicMock(side_effect=lambda k, d=None: {"is_authorized": True}.get(k, d))
+
+        response = await handler(request)
+        assert response.text == "ok"
```
HyeockJinKim: Using `is_authorized` is closer to an implementation detail. In test code, it’s preferable to verify desired scenarios rather than inspect implementation details (e.g., if auth middleware isn't applied, the auth-required handler doesn’t run, an authorization error occurs from the auth middleware if authentication fails, etc.).

HyeockJinKim:
And the timeline-check part in CI got stuck, please check how it applies to other PRs and changes/directories. (Refer to https://github.com/lablup/backend.ai?tab=contributing-ov-file#4-write-changelog-fragment-towncrier.)
You can see which items are available by checking the `pyproject.toml` file.

HyeockJinKim (changes requested):
I'm concerned that there are too many utility functions in this work. Ideally, we should keep test-related objects with simpler mocking objects, rather than multiple implementations. (Considering using aiohttp's Fixture as well.)

kwonkwonn:
Thanks for the review!

I thought the same thing while writing the code, but at the same time, I thought it will be easier to test other `manager/api` handlers which most of them follow the `root_ctx: RootContext = request.app["_root.context"]` pattern with this class.

If I minimize utility functions but keep `APITestAppBuilder`, would that be acceptable?

kwonkwonn:
Or I can fix it to look something more like `tests/unit/manager/api/storage_namespace/` which will make each test file simpler.

kwonkwonn:
This commit implements minimalized mocking utility for test auth.py,  referenced vfolder test scenarios, i think we can move `create_auth_app` outside to esier referencing from other http testing.

tests/unit/manager/api/auth/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+from typing import TYPE_CHECKING, Any
+from unittest.mock import MagicMock
+
+import pytest
+from aiohttp import web
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+    RouteDefinition = tuple[str, str, Callable[..., Any]]
```
fregataa: This part is not needed. You can just import and define them without TYPE_CHECKING indentation

tests/unit/manager/api/auth/test_handlers.py
```diff
@@ -0,0 +1,593 @@
...
+)
+from ai.backend.manager.services.auth.actions.upload_ssh_keypair import UploadSSHKeypairActionResult
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+    from aiohttp.test_utils import TestClient
+
+    RouteDefinition = tuple[str, str, Callable[..., object]]
+    CreateAuthApp = Callable[[list[RouteDefinition]], web.Application]
```
fregataa: same here

tests/unit/manager/api/auth/test_handlers.py
```diff
@@ -0,0 +1,593 @@
...
+                secret_key="SECRET",
+                role=UserRole.USER,
+                status=UserStatus.ACTIVE,
+            ),
+        )
+
+    @pytest.mark.asyncio
+    async def test_valid_keypair_input(
+        self,
+        aiohttp_client,
```
fregataa: it would be nice to have a type annotation here

fregataa (commented):
Did we decide to implement fixture in a separate conftest file? @HyeockJinKim

kwonkwonn:
> Could I create a sample test and share it? @kwonkwonn

yes, If it's not bothering you so much, i will be very appreciated!
@HyeockJinKim

tests/unit/manager/api/auth/test_handlers.py
kwonkwonn: I will remove TYPE_CHECKING and add type annotations!
thanks for review!

tests/unit/manager/api/auth/test_handlers.py
```diff
@@ -0,0 +1,635 @@
...
+    @pytest.mark.asyncio
+    async def test_calls_processor(
+        self,
+        authorized_request: MagicMock,
+        mock_root_ctx: MagicMock,
+    ) -> None:
+        """Verify processor is called with correct params."""
+        user_uuid = uuid.uuid4()
+        authorized_request.text = AsyncMock(
+            return_value=json.dumps({"email": "user@example.com", "full_name": "New Name"})
```
fregataa: How about using a variable for "full_name" rather than using a hard coded value? Because this should be tested by comparing

kwonkwonn: modified "full_name" to be compared as variable. 
thank you!

fregataa: What I mean is that we need to check *all* values that are compared for testing and change them to variables, not just the `full_name` case.

For example:
```python
# OK
name = "hello"
result = func(name)
assert result.name == name

# NOT
result = func("hello")
assert result.name == "hello"
```

kwonkwonn: Oh i misunderstand that.
I'll fix it to be in variable form

kwonkwonn: For any value which is subject of comparions, I move it out from function scope and move it aside to class fields.
which can increase flexability for future test scenarios.

thank you for concise review!

fregataa:
Please rebase to main

HyeockJinKim (approved):

PR #7655, opened 2025-12-31
feat(BA-3490): Implement ScalingGroup Keypair Association Actions

### LINKED ISSUE #7490: Implement KeyPair Association Actions
## Objective

Implement all 4 KeyPair association mutations.

## Mutations

- AssociateScalingGroupWithKeyPair
- AssociateScalingGroupsWithKeyPair
- DisassociateScalingGroupWithKeyPair
- DisassociateScalingGroupsWithKeyPair

## Implementation Details

- `actions/keypair_association.py`: 4 action definitions (~80 lines)
- `service.py`: Add 4 methods (~32 lines)
- `processors.py`: Add 4 processors (~20 lines)
- `repository.py`: KeyPair association methods (~40 lines)
- `db_source/db_source.py`: KeyPair DB operations (~45 lines)
- GraphQL mutations modification (~25 lines)
- Unit tests: (~80 lines)

## Estimated Lines

~322 lines

## Dependencies

- Story 5: ModifyScalingGroup Action (can run parallel with Story 6, 8)

## Acceptance Criteria

- [ ] All 4 KeyPair association actions implemented
- [ ] Single and batch operations working
- [ ] All GraphQL mutations use new processor pattern
- [ ] Unit tests cover all operations

JIRA Issue: BA-3490

jopemachine:
resolves #7490 (BA-3490)

- [x] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/repositories/scaling_group/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/scaling_group/db_source/db_source.py
@@ -194,19 +194,19 @@ class ScalingGroupDBSource:
             result = await session.scalar(query)
             return (result or 0) > 0
 
-    async def associate_scaling_group_with_keypair(
+    async def associate_scaling_group_with_keypairs(
         self,
-        creator: Creator[ScalingGroupForKeypairsRow],
+        bulk_creator: BulkCreator[ScalingGroupForKeypairsRow],
     ) -> None:
-        """Associates a single scaling group with a keypair."""
+        """Associates a scaling group with multiple keypairs."""
         async with self._db.begin_session() as session:
-            await execute_creator(session, creator)
+            await execute_bulk_creator(session, bulk_creator)
 
-    async def disassociate_scaling_group_with_keypair(
+    async def disassociate_scaling_group_with_keypairs(
         self,
         purger: BatchPurger[ScalingGroupForKeypairsRow],
     ) -> None:
-        """Disassociates a single scaling group from a keypair."""
+        """Disassociates a scaling group from multiple keypairs."""
         async with self._db.begin_session() as session:
             await execute_batch_purger(session, purger)
 
--- a/src/ai/backend/manager/repositories/scaling_group/repository.py
+++ b/src/ai/backend/manager/repositories/scaling_group/repository.py
@@ -119,19 +119,19 @@ class ScalingGroupRepository:
             domain=domain,
         )
 
-    async def associate_scaling_group_with_keypair(
+    async def associate_scaling_group_with_keypairs(
         self,
-        creator: Creator[ScalingGroupForKeypairsRow],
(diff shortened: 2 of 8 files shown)


src/ai/backend/manager/repositories/scaling_group/db_source/db_source.py
```diff
@@ -189,3 +193,37 @@ async def check_scaling_group_domain_association_exists(
...
+            query = sa.select(
+                sa.exists().where(
+                    sa.and_(
+                        ScalingGroupForKeypairsRow.scaling_group == scaling_group_name,
+                        ScalingGroupForKeypairsRow.access_key == access_key,
+                    )
+                )
+            )
+            result = await session.execute(query)
+            return result.scalar() or False
```
seedspirit, line 229: It appears to be used only in tests and not in the production API. If it's only used in test code, wouldn't it be acceptable to define it as a utility function within the test file?

src/ai/backend/manager/repositories/scaling_group/purgers.py
```diff
@@ -38,3 +59,17 @@ def create_scaling_group_for_domain_purger(
...
+    access_key: AccessKey,
+) -> BatchPurger[ScalingGroupForKeypairsRow]:
+    """Create a BatchPurger for disassociating a scaling group from a keypair."""
+    return BatchPurger(
+        spec=ScalingGroupForKeypairsPurgerSpec(
+            scaling_group=scaling_group,
+            access_key=access_key,
+        ),
+        batch_size=1,  # We expect only one row to be deleted
+    )
```
seedspirit, line 75: Is there a reason we should wrap the purger creation in a generic function? Since the logic seems very simple, it might be better to create a `BatchPurger` on the fly instead of calling a function.

HyeockJinKim: I don’t think BatchPurger is necessary. Are there any combinations of scaling groups and access keys exceeding one?

src/ai/backend/manager/repositories/scaling_group/db_source/db_source.py
jopemachine: Although it is currently used only in tests, I think it will be needed in the future, since service-level logic is required to check whether the association exists.

src/ai/backend/manager/repositories/scaling_group/purgers.py
jopemachine: I used BatchPurger because Purger could not be used.
Since there is no PurgerSpec type, using Purger would require providing the association’s ID. @HyeockJinKim

jopemachine: We decided it's ok since there is db unique constraint

PR #7658, opened 2025-12-31
fix(BA-3610): `modify_keypair_resource_policy` ignores empty JSON for `total_resource_slots`

### LINKED ISSUE #7657: `modify_keypair_resource_policy` ignores empty JSON for `total_resource_slots`
### Summary

  When calling `modify_keypair_resource_policy` mutation with an empty JSON object (`{`}) for `total_resource_slots`, the database value is not updated. The empty value should be stored as `{`} in the database, but it is silently ignored.

### Steps to Reproduce

1. Create a keypair resource policy with some `total_resource_slots` value (e.g., `{"cpu": "1", "mem": "1g"`})
1. Call `modify_keypair_resource_policy` mutation with `total_resource_slots: "{}"`
1. Query the keypair resource policy to check the `total_resource_slots` value

### Expected Behavior

  The `total_resource_slots` field should be updated to `{`} (empty object) in the database.

### Actual Behavior

  The `total_resource_slots` field retains its previous value. The empty JSON input is ignored and no update occurs.

### Root Cause

  In `ModifyKeyPairResourcePolicyInput.to_modifier()` method (`src/ai/backend/manager/models/resource_policy.py`), the code uses a truthy check:

```python
total_resource_slots = (
    ResourceSlot.from_user_input(self.total_resource_slots, None)
    if self.total_resource_slots  # Bug: empty dict {} is falsy in Python
    else Undefined
)

Since an empty dictionary {} evaluates to False in Python, it is incorrectly treated as Undefined, causing the field to be skipped during the update.

Fix

Change the truthy check to an explicit comparison with Undefined:

total_resource_slots = (
    ResourceSlot.from_user_input(self.total_resource_slots, None)
    if self.total_resource_slots is not Undefined
    else Undefined
)
```

JIRA Issue: BA-3610

jopemachine:
resolves #7657 (BA-3610)

- [x] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/tests/unit/manager/api/keypair_resource_policy/test_types.py
+++ b/tests/unit/manager/api/keypair_resource_policy/test_types.py
@@ -2,22 +2,21 @@ from __future__ import annotations
 
 from unittest.mock import Mock
 
+import pytest
 from graphql import Undefined
 
 from ai.backend.manager.models.resource_policy import ModifyKeyPairResourcePolicyInput
+from ai.backend.manager.repositories.keypair_resource_policy.updaters import (
+    KeyPairResourcePolicyUpdaterSpec,
+)
 from ai.backend.manager.types import _TriStateEnum
 
 
 class TestModifyKeyPairResourcePolicyInputType:
-    def test_empty_total_resource_slots_should_be_updated(self) -> None:
-        """
-        Regression test: empty dict {} for total_resource_slots should result in UPDATE state.
-
-        Previously, empty dict was treated as falsy and converted to Undefined,
-        causing the field to be skipped during updates.
-        """
+    @pytest.fixture
+    def mock_modify_input(self) -> Mock:
         mock_input = Mock(spec=ModifyKeyPairResourcePolicyInput)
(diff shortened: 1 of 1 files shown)


tests/unit/manager/api/keypair_resource_policy/test_types.py
```diff
@@ -0,0 +1,45 @@
...
+        mock_input.max_concurrent_sftp_sessions = Undefined
+        mock_input.max_containers_per_session = Undefined
+        mock_input.idle_timeout = Undefined
+        mock_input.allowed_vfolder_hosts = Undefined
+        mock_input.max_vfolder_count = Undefined
+        mock_input.max_vfolder_size = Undefined
+        mock_input.max_quota_scope_size = Undefined
+        mock_input.max_pending_session_count = Undefined
+        mock_input.max_pending_session_resource_slots = Undefined
+        return mock_input
```
HyeockJinKim, line 32: Could you explain why you used a Mock instead of the ModifyKeyPairResourcePolicyInput object?

PR #7805, opened 2026-01-06
fix(BA-3780): Fix cursor pagination by adding missing ID cursor condition when no `order_expr` provided

seedspirit:
resolves (BA-3780)
### Problem
  When cursor-based pagination was used without an explicit `order` parameter, the cursor condition was not being applied to the SQL WHERE clause. This caused pagination to return the same first page of results repeatedly, instead of advancing through the result set.

  **Root Cause**: The cursor condition logic was inside a loop over `ordering_item_list`:
  ```python
  for col, direction in ordering_item_list:
      # Add cursor condition for this ordering column
      cursor_conditions.append(subq_to_condition(col, subq, direction))
```

When no explicit ordering was provided, `ordering_item_list` was empty, so the loop never executed and no cursor conditions were added.

### Solution

  Added a fallback id-based cursor condition that is applied only when no explicit ordering is provided:

```python
  # Add id-based cursor WHERE condition ONLY when no explicit ordering is provided.
  # This is CRITICAL for pagination to work when no explicit order_expr is provided.
  # When ordering_item_list is not empty, the id condition is already embedded
  # in the ordering cursor conditions above (via condition_when_same_with_subq).
  if not ordering_item_list:
      match pagination_order:
          case ConnectionPaginationOrder.FORWARD | None:
              cursor_conditions.append(id_column > cursor_row_id)
          case ConnectionPaginationOrder.BACKWARD:
              cursor_conditions.append(id_column < cursor_row_id)

  for cond in cursor_conditions:
      stmt = stmt.where(cond)
```

Why the if not `ordering_item_list` check is necessary:
  - When explicit ordering is provided (e.g., `order="-created_at"`), the id condition is already included in `subq_to_condition()` via `condition_when_same_with_subq`
  - Without this check, the id condition would be added twice, creating an overly restrictive WHERE clause that filters out valid results

---

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/manager/api/gql_legacy/base.py
```diff
@@ -928,6 +929,140 @@ def validate_connection_args(
...
+        """
+        match pagination_order:
+            case ConnectionPaginationOrder.FORWARD | None:
+                if direction == OrderDirection.ASC:
+                    cond = column_to_be_compared > subquery
+                else:
+                    cond = column_to_be_compared < subquery
+
+                # Comparing ID field - The direction of inequality sign is not affected by `direction` argument here
+                # because the ordering direction of ID field is always determined by `pagination_order` only.
```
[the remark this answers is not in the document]
seedspirit: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/lablup/backend.ai/pull/7805#discussion_r2664396656)

src/ai/backend/manager/api/gql_legacy/base.py
```diff
@@ -939,81 +1074,40 @@ def _build_sql_stmt_from_connection_args(
...
-            # Default ordering by id column
-            id_ordering_item = OrderingItem(id_column, OrderDirection.DESC)
-            set_ordering = lambda col, direction: (
-                col.desc() if direction == OrderDirection.ASC else col.asc()
-            )
-    # id column should be applied last
-    for col, direction in [*ordering_item_list, id_ordering_item]:
-        stmt = stmt.order_by(set_ordering(col, direction))
+    # Apply ORDER BY for cursor-based pagination
+    stmt = _apply_ordering(stmt, id_column, ordering_item_list, pagination_order)
```
fregataa, line 1086: This is the main change?

src/ai/backend/manager/api/gql_legacy/base.py
```diff
@@ -928,6 +929,140 @@ def validate_connection_args(
...
+    # in the ordering cursor conditions above (via condition_when_same_with_subq).
+    if not ordering_item_list:
+        match pagination_order:
+            case ConnectionPaginationOrder.FORWARD | None:
+                cursor_conditions.append(id_column > cursor_row_id)
+            case ConnectionPaginationOrder.BACKWARD:
+                cursor_conditions.append(id_column < cursor_row_id)
+
+    for cond in cursor_conditions:
+        stmt = stmt.where(cond)
```
seedspirit, line 1061: @fregataa  @HyeockJinKim  Main change is here. Going forward, I'll separate refactoring and bug fixes. Mixing them together makes it hard to find main change

fregataa: Okay. The legacy code is hard to understand but we should keep fix tasks as small as possible. Such fix tasks with refactoring makes it hard to review

PR #7865, opened 2026-01-08
refactor(BA-3590): Implement Pydantic migration for `manager/api/auth.py`

### LINKED ISSUE #7625: Restructure auth.py into modular directory structure (auth/adapter, auth/handler, auth/models, auth/init.py)
## Motivation

- Split auth.py from one single directory based pot, helps following maintainer easy to understand the logics.

## Required Features

- Assure largely reffered functions to work safely after dividing.
- Needs pre-splitting before __init__ , handler, adaptper pattern implemented.

## Impact

- Helps developer better understand code structure.
- Benefits preventioning Cycliclic dependency before restructuring other files because mostly reffered functions divided first.

## Testing Scenarios

- import error check
- lint error check

JIRA Issue: BA-3590

kwonkwonn:
resolves #7625 (BA-3590)

This PR would consists of sequence of commits following.
- define DTO for req/res (can easily validate with manager/api/auth.py's res, req)
- define Data class and expected actions behavior and form.

Real Migration will be taken from next pr. 

Once this process 
- write test code(refering `tests/unit/manager/api/test_resource`)
- Define DTO, DataClass, action validator.
- Define Handler and Adapter. 
succesfully done, It can be easily tested and automatized. 

- [ ] Test case(s) to:
- [ ] Documentation

----
📚 Documentation preview 📚: https://sorna--7865.org.readthedocs.build/en/7865/

----
📚 Documentation preview 📚: https://sorna-ko--7865.org.readthedocs.build/ko/7865/

### DIFF
--- a/docs/manager/rest-reference/openapi.json
+++ b/docs/manager/rest-reference/openapi.json
@@ -1323,6 +1323,195 @@
         "title": "ScanArtifactModelsReq",
         "type": "object"
       },
+      "SignupRequest": {
+        "description": "Request to sign up a new user. Created with INACTIVE status; requires separate activation.",
+        "properties": {
+          "domain": {
+            "description": "Domain (tenant) to create the user in",
+            "title": "Domain",
+            "type": "string"
+          },
+          "email": {
+            "description": "Email for the account; must be globally unique, also used as username if not specified",
+            "title": "Email",
+            "type": "string"
+          },
+          "password": {
+            "description": "Initial password; validated via VERIFY_PASSWORD_FORMAT hook",
+            "title": "Password",
+            "type": "string"
+          },
+          "username": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "default": null,
+            "description": "Display username; defaults to email if omitted",
+            "title": "Username"
+          },
+          "full_name": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "default": null,
+            "description": "User's full name for display; defaults to empty string",
+            "title": "Full Name"
+          },
+          "description": {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "default": null,
+            "description": "Free-text account description; defaults to empty string",
+            "title": "Description"
+          }
+        },
+        "required": [
+          "domain",
+          "email",
+          "password"
+        ],
+        "title": "SignupRequest",
+        "type": "object"
+      },
+      "SignoutRequest": {
+        "description": "Request to deactivate a user account (sets INACTIVE, does not delete).",
+        "properties": {
+          "email": {
+            "description": "Email of the account to deactivate; must match the requester's own email",
+            "title": "Email",
+            "type": "string"
+          },
+          "password": {
+            "description": "Current password for identity confirmation",
+            "title": "Password",
+            "type": "string"
+          }
+        },
+        "required": [
+          "email",
+          "password"
+        ],
+        "title": "SignoutRequest",
+        "type": "object"
+      },
+      "UpdatePasswordNoAuthRequest": {
+        "description": "Request to update an expired password without API auth.\nOnly available when max_password_age is configured.",
+        "properties": {
+          "domain": {
+            "description": "Domain (tenant) the user belongs to",
+            "title": "Domain",
+            "type": "string"
+          },
+          "username": {
+            "description": "Login identifier (email or username)",
+            "title": "Username",
+            "type": "string"
+          },
+          "current_password": {
+            "description": "Current (expired) password for verification",
+            "title": "Current Password",
+            "type": "string"
+          },
+          "new_password": {
+            "description": "New password; must differ from current password",
+            "title": "New Password",
+            "type": "string"
+          }
+        },
+        "required": [
+          "domain",
+          "username",
+          "current_password",
+          "new_password"
+        ],
+        "title": "UpdatePasswordNoAuthRequest",
+        "type": "object"
+      },
+      "UpdatePasswordRequest": {
+        "description": "Request to update the authenticated user's password with confirmation.",
+        "properties": {
+          "old_password": {
+            "description": "Current password for identity verification",
+            "title": "Old Password",
+            "type": "string"
+          },
+          "new_password": {
+            "description": "Desired new password; validated via VERIFY_PASSWORD_FORMAT hook",
+            "title": "New Password",
+            "type": "string"
+          },
+          "new_password2": {
+            "description": "New password confirmation; must match new_password exactly",
+            "title": "New Password2",
+            "type": "string"
+          }
+        },
+        "required": [
+          "old_password",
+          "new_password",
+          "new_password2"
+        ],
+        "title": "UpdatePasswordRequest",
+        "type": "object"
+      },
+      "UpdateFullNameRequest": {
+        "description": "Request to update the authenticated user's display full name.",
+        "properties": {
+          "email": {
+            "description": "Email of the user to update",
+            "title": "Email",
+            "type": "string"
+          },
+          "full_name": {
+            "description": "New full name to set",
+            "title": "Full Name",
+            "type": "string"
+          }
+        },
+        "required": [
+          "email",
+          "full_name"
+        ],
+        "title": "UpdateFullNameRequest",
+        "type": "object"
+      },
+      "UploadSSHKeypairRequest": {
+        "description": "Request to upload an SSH keypair. The pair is validated for consistency.",
+        "properties": {
+          "pubkey": {
+            "description": "SSH public key in standard format (e.g., 'ssh-rsa AAAA...')",
+            "title": "Pubkey",
+            "type": "string"
+          },
+          "privkey": {
+            "description": "SSH private key in PEM format",
+            "title": "Privkey",
+            "type": "string"
+          }
+        },
+        "required": [
+          "pubkey",
+          "privkey"
+        ],
+        "title": "UploadSSHKeypairRequest",
+        "type": "object"
+      },
       "VFolderMountPermission": {
         "description": "Permissions for a virtual folder given to a specific access key.\nRW_DELETE includes READ_WRITE and READ_WRITE includes READ_ONLY.",
         "enum": [
@@ -10559,7 +10748,7 @@
     },
     "/auth": {
       "get": {
-        "operationId": "auth.test",
+        "operationId": "auth.verify_auth",
         "tags": [
           "auth"
         ],
@@ -10576,17 +10765,19 @@
         "parameters": [
           {
             "name": "echo",
-            "schema": {
-              "type": "string"
-            },
+            "in": "query",
             "required": true,
-            "in": "query"
+            "schema": {
+              "description": "Arbitrary string echoed back to confirm auth is working",
+              "title": "Echo",
+              "type": "string"
+            }
           }
         ],
         "description": "\n**Preconditions:**\n* User privilege required.\n"
       },
       "post": {
-        "operationId": "auth.test.2",
+        "operationId": "auth.verify_auth.2",
         "tags": [
           "auth"
         ],
@@ -10600,31 +10791,24 @@
             "TokenAuth": []
           }
         ],
-        "requestBody": {
-          "content": {
-            "application/json": {
-              "schema": {
-                "type": "object",
-                "properties": {
-                  "echo": {
-                    "type": "string"
-                  }
-                },
-                "required": [
-                  "echo"
-                ]
-              },
-              "examples": {}
+        "parameters": [
+          {
+            "name": "echo",
+            "in": "query",
+            "required": true,
+            "schema": {
+              "description": "Arbitrary string echoed back to confirm auth is working",
+              "title": "Echo",
+              "type": "string"
             }
           }
-        },
-        "parameters": [],
+        ],
         "description": "\n**Preconditions:**\n* User privilege required.\n"
       }
     },
     "/auth/test": {
       "get": {
-        "operationId": "auth.test.3",
+        "operationId": "auth.verify_auth.3",
         "tags": [
           "auth"
         ],
@@ -10641,17 +10825,19 @@
         "parameters": [
           {
             "name": "echo",
-            "schema": {
-              "type": "string"
-            },
+            "in": "query",
             "required": true,
-            "in": "query"
+            "schema": {
+              "description": "Arbitrary string echoed back to confirm auth is working",
+              "title": "Echo",
+              "type": "string"
+            }
           }
         ],
         "description": "\n**Preconditions:**\n* User privilege required.\n"
       },
       "post": {
-        "operationId": "auth.test.4",
+        "operationId": "auth.verify_auth.4",
         "tags": [
           "auth"
         ],
@@ -10665,25 +10851,18 @@
             "TokenAuth": []
           }
         ],
-        "requestBody": {
-          "content": {
-            "application/json": {
-              "schema": {
-                "type": "object",
-                "properties": {
-                  "echo": {
-                    "type": "string"
-                  }
-                },
-                "required": [
-                  "echo"
-                ]
-              },
-              "examples": {}
+        "parameters": [
+          {
+            "name": "echo",
+            "in": "query",
+            "required": true,
+            "schema": {
+              "description": "Arbitrary string echoed back to confirm auth is working",
+              "title": "Echo",
+              "type": "string"
             }
           }
-        },
-        "parameters": [],
+        ],
         "description": "\n**Preconditions:**\n* User privilege required.\n"
       }
     },
@@ -10696,40 +10875,6 @@
         "responses": {
           "200": {
             "description": "Successful response"
-          }
-        },
-        "requestBody": {
-          "content": {
-            "application/json": {
-              "schema": {
-                "type": "object",
-                "properties": {
-                  "type": {
-                    "type": "string",
-                    "enum": [
-                      "keypair",
-                      "jwt"
-                    ]
-                  },
-                  "domain": {
-                    "type": "string"
-                  },
-                  "username": {
-                    "type": "string"
-                  },
-                  "password": {
-                    "type": "string"
-                  }
-                },
-                "required": [
-                  "type",
-                  "domain",
-                  "username",
-                  "password"
-                ]
-              },
-              "examples": {}
-            }
           }
         },
         "parameters": [],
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/repositories/auth/repository.py
```diff
@@ -100,3 +100,19 @@ async def get_user_row_by_uuid(self, user_uuid: UUID) -> UserRow:
...
+        Used by authentication middleware to populate request context.
+
+        Args:
+            access_key: The access key to look up.
+
+        Returns:
+            CredentialData containing user, keypair, and resource policies,
+            or None if not found or inactive.
+        """
+        return await self._db_source.fetch_credential_by_access_key(access_key)
```
fregataa: How is this function used? I cannot find

src/ai/backend/common/dto/manager/auth/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+    "AuthorizeResponse",
+    "GetRoleResponse",
+    "SignupResponse",
+    "SignoutResponse",
+    "UpdateFullNameResponse",
+    "UpdatePasswordResponse",
+    "UpdatePasswordNoAuthResponse",
+    "GetSSHKeypairResponse",
+    "SSHKeypairResponse",
+)
```
fregataa: Let's not export all modules in `__init__.py`.
Ref: https://github.com/lablup/backend.ai/pull/7702

fregataa (changes requested):

src/ai/backend/manager/repositories/auth/repository.py
kwonkwonn: In order to use this function 
I should change some structures in api/auth.py, but since test for auth.py is not being passed i postponed to modify auth.py for now.

src/ai/backend/common/dto/manager/auth/__init__.py
kwonkwonn: I will modify to export only needed modules.
Thank you!

src/ai/backend/manager/repositories/auth/repository.py
kwonkwonn: Or I can modify a little bit to use that function as it is only draft pr.
If you think it seems alright.

src/ai/backend/common/dto/manager/auth/__init__.py
kwonkwonn: removed re-exporting from __init__py. 
Please refer my second commit. ( as 'authadapter` imports each modules from __init__.py, i had to modify from second commit to ensure no conflicts to happen for each commit.)

kwonkwonn:
- Add pydantic DTOs for auth API request/response validation
- Split `auth.py` into `auth/` package structure with handler and adapter
- Add compatibility tests verifying trafaret and pydantic parse same JSON

## Design Decisions

### 1. No `@api_handler` decorator for now

The `authorize` endpoint returns either `web.Response` or `StreamResponse` depending on auth type (keypair vs jwt). The `@api_handler` decorator expects a single return type, so we kept the manual request parsing approach for now.

### 2. Trafaret handlers preserved

Both trafaret (legacy) and pydantic handlers coexist in `auth/__init__.py`. The `create_app` currently routes to the new `AuthAPIHandler`. As file name confliction happens, I've decided to rm for this commmit, but we can restore and do gradual migration if needed.

### 3. Test strategy changed

Previous `test_handlers.py` mocked `request.text()` to test trafaret's `check_params()`. With pydantic, validation happens during DTO construction which is already tested in `test_dto_compatibility.py`. The new `test_adapter.py` focuses on ActionResult → Response DTO conversion instead. 

## Test plan

- [x] `test_dto_compatibility.py` - Gold format tests (trafaret & pydantic parse same JSON)
- [x] `test_adapter.py` - ActionResult to Response DTO conversion

src/ai/backend/common/dto/manager/auth/request.py
```diff
@@ -0,0 +1,106 @@
...
+    password: str = Field(description="Password")
+
+    @model_validator(mode="after")
+    def validate_email_or_username(self) -> SignoutRequest:
+        """Ensure either email or username is provided, prefer email."""
+        if self.email is None and self.username is None:
+            raise ValueError("Either email or username must be provided")
+        # If only username is provided, use it as email
+        if self.email is None and self.username is not None:
+            object.__setattr__(self, "email", self.username)
```
fregataa: How about adding [`compute_field`](https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.computed_field) rather than setting value directly?

src/ai/backend/common/dto/manager/auth/types.py
```diff
@@ -57,12 +72,12 @@ def to_dict(self) -> dict[str, str]:
 
 def parse_auth_response(data: dict[str, Any]) -> AuthResponse:
     raw_response_type = data.get("response_type")
-    respones_type = (
+    response_type = (
```
fregataa, line 84: 👍

src/ai/backend/manager/api/auth/__init__.py
```diff
@@ -436,6 +438,8 @@ async def _query_cred_by_access_key(
     Returns:
         Tuple of (user_row, keypair_row) or (None, None) if not found
     """
+    import sqlalchemy as sa
+
```
fregataa: Please leave import statements as it was

src/ai/backend/manager/api/auth/adapter.py
```diff
@@ -0,0 +1,103 @@
...
+    """Adapter for converting auth ActionResult to Response DTO."""
+
+    def convert_authorize_result(self, result: AuthorizeActionResult) -> AuthorizeResponse:
+        """Convert AuthorizeActionResult to AuthorizeResponse.
+
+        Note: Caller must check result.stream_response before calling this method.
+        This method only handles the keypair authentication case.
+        """
+        if result.authorization_result is None:
+            raise ValueError("authorization_result is required for keypair authentication")
```
fregataa: Please raise a customized exception here.
ref: https://github.com/lablup/backend.ai/blob/ad0d17e720655cd7dba6e7be909a427ed532186a/src/ai/backend/manager/errors/api.py#L18-L29

src/ai/backend/manager/api/auth/handler.py
```diff
@@ -0,0 +1,362 @@
...
+    @api_handler
+    async def test(
+        self,
+        user_ctx: UserContext,
+    ) -> APIResponse:
+        """Test authentication."""
+        log.info("AUTH.TEST(ak:{})", user_ctx.access_key)
+        # Note: echo parameter handling requires QueryParam, but test endpoint
+        # is simple enough to just return authorized status
+        from ai.backend.common.api_handlers import BaseResponseModel
```
fregataa: Place import statements on the top of this file

src/ai/backend/manager/api/auth/handler.py
```diff
@@ -0,0 +1,362 @@
...
+        return APIResponse.build(
+            status_code=HTTPStatus.OK, response_model=TestResponse(authorized="yes")
+        )
+
+    # Authorization endpoints
+
+    async def authorize(
+        self,
+        request: web.Request,
+    ) -> web.StreamResponse:
```
fregataa: No `api_handler` applied here?

fregataa (changes requested):

src/ai/backend/manager/api/auth/handler.py
kwonkwonn: For authorize function, type of return variable have two options.
- web.Response for usual authentication.
- for requests like 2fa, it returns web.StreamResponse.

because return var can be varied, i couldn't just wrap with usual @api_handler or @stream_api_handler.

src/ai/backend/common/dto/manager/auth/request.py
fregataa: Please check this part

src/ai/backend/manager/api/auth/handler.py
fregataa: web.Response is a subtype of web.StreamResponse. Please try applying `@api_handler`

fregataa (changes requested):
First, we have to update Open API spec generator because it lose all spec information of changed API handlers

kwonkwonn:
For decorating authrize with @api_handler, 

from @api_handler
```  
 @functools.wraps(handler)
    async def wrapped(first_arg: Any, request: web.Request) -> web.Response:
        kwargs: dict[str, _ParamType] = {}
        for name, param_instance_or_class in signature_parser_map.items():
            param_instance = await _serialize_parameter(request, param_instance_or_class)
            kwargs[name] = param_instance
        response = await handler(first_arg, **kwargs)
        return _parse_response(response)
```
in _parse_response, 
```
def _parse_response(response: APIResponse) -> web.Response:
    return web.json_response(
        data=response.to_json,
        status=response.status_code,
    )
``` 
checks .to_json and status_code attribute, so when 2fa return streamresponse, i think it will be broken.

if i've overlooked on something please let me know!

kwonkwonn:
For open api, unified `APIResponse` doesn't let open api to refer the response model.
current 
```
@dataclass
class APIResponse:
    _status_code: int
    _data: Optional[BaseResponseModel]

    @classmethod
    def build(cls, status_code: int, response_model: BaseResponseModel) -> Self:
        return cls(_status_code=status_code, _data=response_model)

    @classmethod
    def no_content(cls, status_code: int) -> Self:
        return cls(_status_code=status_code, _data=None)

    @property
    def to_json(self) -> Optional[JSONDict]:
        return self._data.model_dump(mode="json") if self._data else None

    @property
    def status_code(self) -> int:
        return self._status_code
```

type generic
```
@dataclass
class APIResponse(Generic[TResponseModel]):
    _data: Optional[TResponseModel]
```


But it require us to look in every occurences of APIRespone.

src/ai/backend/common/dto/manager/auth/request.py
kwonkwonn: I have small concern as compute_fields can be not represented openapi fetcher and 
goal of compute field is compute with other fields in the class.

How do you think about it

fregataa:
> For decorating authrize with @api_handler,
> 
> checks .to_json and status_code attribute, so when 2fa return streamresponse, i think it will be broken.
> 
> if i've overlooked on something please let me know!

Then how about `stream_api_handler()` instead? This should not be broken

src/ai/backend/common/dto/manager/auth/request.py
fregataa: I suggest to add a computed field for easy access of email or username field from the handler. The computed field should not be accessed by API schema checker (=open API fetcher).
```python
email: str = Field(description="email. It is nullable and will be replaced by username if it is null.")
username: str = Field(description="username. It will replace email if email is null")
...

@computed_field
def email_or_username(self) -> str:
    # The handler should access this field rather than email or username
    return self.email if self.email is not None else self.username
```

HyeockJinKim: Are the username and email the same value?

kwonkwonn: Email and Username were accecpting value as alias at the first time.  
`tx.AliasedKey(["email", "username"])`

What i assume is email and username interchangebly acts as user identifier in authorization cluster.

kwonkwonn:
I've been working on @api_streamresponse, and what i've found was action result here,

```
        result: APIStreamResponse = await handler(first_arg, **kwargs)
```
@stream_api_handler expects wrapped handler to emit `APIStreamResponse` and transform it into web.streamResponse.
So usual `APIResponse` can't be guaranteed to be transformed unlike web.streamResponse <-> web.Response is possible.

src/ai/backend/common/dto/manager/auth/request.py
fregataa: Then we should use validation_alias rather than assign values.
Here is an example. https://github.com/lablup/backend.ai/blob/f1a00abfbbe7ae2b0f9a2add7efbbd19f36e264f/src/ai/backend/manager/config/unified.py#L319-L326

docs/manager/rest-reference/openapi.json
HyeockJinKim, line 1: Why has this file been changed so much?

kwonkwonn:
- updated to use AliasChoices for two interchangable inputs.

**follow up issues**:
 #8149 
- decorator strategy for `authorize`.

 #8150 
- Return type clarify for openApi automated builder

Thank you!

kwonkwonn:
## Summary
- Migrate auth API handlers from monolithic `auth.py` to modular package structure
- Add pydantic DTOs for request/response validation
- Preserve legacy handlers in `api/legacy/` for regression testing
- Add regression test using both legacy and new handler. let me know if this approach seems valid!

## Notes

### Deferred items (out of scope)
- `current_user` vs `user_context` field doesn't contain what is needed from user_context - kept as-is due to broader impact
- `SuccessResponse` unification - touches too many files, separate PR
- Circular import workaround: `create_app` wrapper in `__init__.py` defers `handler.py` import
  - Root cause: `repositories.session.repository` imports `api.session.find_dependency_sessions`
  - TODO: Move `find_dependency_sessions` to repository layer to remove wrapper
- OpenAPI addition - refer https://github.com/lablup/backend.ai/issues/8150

docs/manager/rest-reference/openapi.json
kwonkwonn: I think this file should automaticaly built with openapi and ocotodo's macro per PR .

And as https://github.com/lablup/backend.ai/pull/8245, which contains logics for detecting `@api_handler` wrapped requests, i believe this would contain request schema well from next time.

kwonkwonn: I will clean micro updates and old reviews and re-request for this pr. 

thanks for review!

src/ai/backend/manager/api/auth/decorators.py
```diff
@@ -0,0 +1,124 @@
...
+    async def wrapped(
+        self: Any, request: web.Request, *args: Any, **kwargs: Any
+    ) -> web.StreamResponse:
+        if request.get("is_authorized", False) and request.get("is_superadmin", False):
+            return await method(self, request, *args, **kwargs)
+        raise AuthorizationFailed("Unauthorized access")
+
+    set_handler_attr(wrapped, "auth_required", True)
+    set_handler_attr(wrapped, "auth_scope", "superadmin")
+    return wrapped
```
fregataa: Doesn't this function already exist? I remember that I reviewed this code

kwonkwonn: As i moved utility functions in `auth.py` into seperated files `auth/*`,  I think git doesn't think it is a mere transition of the place.
though what i did was copy and paste into seperated files.

docs/manager/rest-reference/openapi.json
```diff
@@ -8521,42 +8700,8 @@
...
-                "type": "object",
-                "properties": {
-                  "type": {
-                    "type": "string",
-                    "enum": [
-                      "keypair",
-                      "jwt"
-                    ]
-                  },
-                  "domain": {
```
kwonkwonn: `@api_handler` for authorize needs to be implemented.
https://github.com/lablup/backend.ai/issues/8149

docs/manager/rest-reference/openapi.json
```diff
@@ -8488,26 +8674,19 @@
             "TokenAuth": []
           }
         ],
-        "requestBody": {
-          "content": {
-            "application/json": {
-              "schema": {
-                "type": "object",
```
kwonkwonn: As origin `test` handler recevies both request and query parmeter while it was used in both `get` and `post`. 
some parameters were lost while migration.

src/ai/backend/manager/api/auth/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+    sign_request,
+    validate_ip,
+)
+
+
+def create_app(
+    default_cors_options: Any,
+) -> tuple[web.Application, list[Any]]:
+    """Create aiohttp application for auth API endpoints.
+
```
kwonkwonn: session handler cause cyclic imports when i tried to lift import state up,
This should be refactored after session handler's migration done.

src/ai/backend/manager/api/auth/decorators.py
```diff
@@ -0,0 +1,124 @@
...
+
+__all__ = (
+    "auth_required",
+    "auth_required_for_method",
+    "admin_required",
+    "admin_required_for_method",
+    "superadmin_required",
+    "superadmin_required_for_method",
+)
+
```
kwonkwonn: https://github.com/lablup/backend.ai/blob/c1be008a3f59602a8e5326768dd6c4c862eaef26/src/ai/backend/manager/api/auth.py#L776 

This seems like new implementation, but mere copy and pase of the existing decorators.
sorry for inconvenience, but i didn't wanted to put ~700 lines in one single file... 
@fregataa

src/ai/backend/manager/api/auth/middleware.py
```diff
@@ -0,0 +1,388 @@
...
+from ai.backend.logging import BraceStyleAdapter
+from ai.backend.logging.utils import with_log_context_fields
+from ai.backend.manager.api.utils import get_handler_attr
+from ai.backend.manager.errors.auth import AuthorizationFailed, InvalidAuthParameters
+from ai.backend.manager.errors.common import RejectedByHook
+from ai.backend.manager.models.keypair import keypairs
+from ai.backend.manager.models.resource_policy import (
+    keypair_resource_policies,
+    user_resource_policies,
+)
```
kwonkwonn: https://github.com/lablup/backend.ai/blob/c1be008a3f59602a8e5326768dd6c4c862eaef26/src/ai/backend/manager/api/auth.py#L720 
<-- same in here

tests/unit/manager/api/auth/test_regression.py
```diff
@@ -0,0 +1,433 @@
...
+    app.router.add_post("/auth/signout", legacy_auth.signout)
+    app.router.add_post("/auth/update-password-no-auth", legacy_auth.update_password_no_auth)
+    app.router.add_post("/auth/update-password", legacy_auth.update_password)
+    app.router.add_post("/auth/update-full-name", legacy_auth.update_full_name)
+    app.router.add_get("/auth/ssh-keypair", legacy_auth.get_ssh_keypair)
+    app.router.add_patch("/auth/ssh-keypair", legacy_auth.generate_ssh_keypair)
+    app.router.add_post("/auth/ssh-keypair", legacy_auth.upload_ssh_keypair)
+
+    return app
+
```
HyeockJinKim, line 132: It is questionable whether route registration must be identical in unit tests.

src/ai/backend/manager/dto/context.py
```diff
@@ -13,7 +15,9 @@
 from ai.backend.manager.config.unified import ExportConfig
 from ai.backend.manager.models.storage import StorageSessionManager
 from ai.backend.manager.repositories.export import ExportRepository
-from ai.backend.manager.services.processors import Processors
+
+if TYPE_CHECKING:
+    from ai.backend.manager.services.processors import Processors
```
kwonkwonn, line 21: even I've modified two different pr to resolve cyclic import error,
It happens again here.

once we can seperate middleware and decoraters from auth handlers, we can remove TYPE_CHECKING

HyeockJinKim:
All migration tasks have been applied; I will now close this.

HyeockJinKim:
All migration tasks have been applied; I will now close this.

HyeockJinKim:
All migration tasks have been applied; I will now close this.

PR #7871, opened 2026-01-08
docs(BA-3549): Clarify host and X-Forwarded-URL header configuration in Apollo Router `router.yaml`

### LINKED ISSUE #7578: Document environment-specific configuration for host and X-Forwarded-URL headers in Apollo Router
## Background

The `headers` section in Apollo Router's `router.yaml` configuration file contains insert values that are included in request headers sent to the Manager. The Manager uses these header values to generate credentials and validate requests.

## Problem

In production deployments, the values in `router.yaml` must be changed according to the host and port configuration. However, these configuration requirements were not documented, causing significant time spent during deployment troubleshooting.

  The following headers must be configured according to the environment:

- `host`: The actual hostname/IP and port of the Manager
- `X-Forwarded-URL`: The external URL that clients use to access the router (must use host port when Docker Compose port mapping is used)

## Solution

 Add comments to `router.yaml` to clearly document the purpose of each configuration value and how to set them for different environments.

## Changes

- `host` header: Added note that it must be changed to the actual Manager address
- `X-Forwarded-URL` header:
  - Must match the external endpoint that clients use to access the router
  - Must use the host port when using Docker Compose port mapping (not the container port)
  - Trailing slash is required

JIRA Issue: BA-3549

seedspirit:
resolves #7578 (BA-3549)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/configs/graphql/router.yaml
+++ b/configs/graphql/router.yaml
@@ -12,7 +12,7 @@ headers:
       # NOTE: Headers required for authentication when communicating with Manager
       - insert:
           name: "host"
-         # NOTE: localhost must be changed to the IP or hostname of the Manager
+         # NOTE: Set this to the address where the Manager is reachable from this component
           value: "localhost:8091"
       - insert:
           name: "X-Forwarded-URL"


configs/graphql/router.yaml
```diff
@@ -14,4 +14,6 @@ headers:
           value: "localhost:8091"
       - insert:
           name: "X-Forwarded-URL"
-          value: "http://127.0.0.1:4000/"
+          # NOTE: When the host port and container port differ in Docker Compose, you must use the host port
```
fregataa: Does this mean "the client from outside of docker network must use the host port" or "the client in the same docker network also must use host port"?

configs/graphql/router.yaml
```diff
@@ -9,9 +9,14 @@ headers:
     request:
       - propagate:
           matching: ".*"
+      # NOTE: Headers required for authentication when communicating with Manager
       - insert:
           name: "host"
+         # NOTE: localhost must be changed to the IP or hostname of the Manager
           value: "localhost:8091"
```
HyeockJinKim: That phrasing "localhost mustbe" doesn't sound right.
We should use `Set this to the address where the Manager is reachable from this component`.

HyeockJinKim (approved):

PR #7936, opened 2026-01-12
refactor(BA-3832): prework to refactor auth.py from direct database access to follow dir pattern

kwonkwonn:
resolves (BA-3832)// issue includes multiple pr. 

## this commit address
- Moved previous `_query_cred_by_access_key` functions with new function do the same thing in `directory`.
- new class CredentialData to migrate Dict based argument passing to class based verification.

- [x] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/changes/7936.enhance.md
+++ b/changes/7936.enhance.md
@@ -1,0 +1,1 @@
+Migrate direct Database query from api/auth.py to follow directory pattern and safer validation.
--- a/tests/unit/manager/data/auth/BUILD
+++ b/tests/unit/manager/data/auth/BUILD
@@ -1,0 +1,6 @@
+python_tests(
+    name="tests",
+    dependencies=[
+        "src/ai/backend/manager:src",
+    ],
+)
--- a/tests/unit/manager/data/auth/test_credential_data.py
+++ b/tests/unit/manager/data/auth/test_credential_data.py
@@ -1,0 +1,196 @@
+"""Tests for CredentialData dict conversion matching legacy Core query output."""
+
+from __future__ import annotations
+
+import uuid
+from dataclasses import replace
+from datetime import UTC, datetime
+
+import pytest
+
+from ai.backend.common.types import (
+    DefaultForUnspecified,
+    ResourceSlot,
+    VFolderHostPermissionMap,
+)
+from ai.backend.manager.data.auth.types import (
+    CredentialData,
+    KeypairDataForCredential,
+    KeypairResourcePolicyDataForCredential,
+    UserDataForCredential,
+    UserResourcePolicyDataForCredential,
+)
+from ai.backend.manager.data.user.types import UserRole, UserStatus
+
+
+@pytest.fixture
+def user_rp() -> UserResourcePolicyDataForCredential:
+    return UserResourcePolicyDataForCredential(
+        name="user-policy",
+        created_at=datetime.now(UTC),
+        max_vfolder_count=10,
+        max_quota_scope_size=-1,
+        max_session_count_per_model_session=5,
+        max_customized_image_count=3,
+    )
+
+
+@pytest.fixture
+def keypair_rp() -> KeypairResourcePolicyDataForCredential:
+    return KeypairResourcePolicyDataForCredential(
+        name="keypair-policy",
+        created_at=datetime.now(UTC),
+        default_for_unspecified=DefaultForUnspecified.LIMITED,
+        total_resource_slots=ResourceSlot({}),
+        max_session_lifetime=0,
+        max_concurrent_sessions=10,
+        max_pending_session_count=None,
+        max_pending_session_resource_slots=None,
+        max_concurrent_sftp_sessions=2,
+        max_containers_per_session=5,
+        idle_timeout=3600,
+        allowed_vfolder_hosts=VFolderHostPermissionMap({}),
+    )
+
+
+@pytest.fixture
+def user_data(user_rp: UserResourcePolicyDataForCredential) -> UserDataForCredential:
+    return UserDataForCredential(
+        uuid=uuid.uuid4(),
+        username="testuser",
+        email="test@example.com",
+        need_password_change=False,
+        password_changed_at=datetime.now(UTC),
+        full_name="Test User",
+        status=UserStatus.ACTIVE,
+        status_info="OK",
+        modified_at=datetime.now(UTC),
+        domain_name="default",
+        role=UserRole.USER,
+        allowed_client_ip=None,
+        totp_key=None,
+        totp_activated=False,
+        totp_activated_at=None,
+        resource_policy_name="user-policy",
+        sudo_session_enabled=False,
+        main_access_key="AKTEST123",
+        integration_id=None,
+        container_uid=None,
+        container_main_gid=None,
+        container_gids=None,
+        resource_policy=user_rp,
+    )
+
+
+@pytest.fixture
+def keypair_data(
+    user_data: UserDataForCredential,
+    keypair_rp: KeypairResourcePolicyDataForCredential,
+) -> KeypairDataForCredential:
+    return KeypairDataForCredential(
+        user_id="test@example.com",
+        access_key="AKTEST123",
+        secret_key="secret_key_value",
+        is_active=True,
+        is_admin=False,
+        created_at=datetime.now(UTC),
+        modified_at=datetime.now(UTC),
+        last_used=None,
+        rate_limit=1000,
+        num_queries=0,
+        ssh_public_key=None,
+        ssh_private_key=None,
+        user=user_data.uuid,
+        resource_policy_name="keypair-policy",
+        dotfiles=b"\x90",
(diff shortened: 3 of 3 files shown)


src/ai/backend/manager/data/auth/types.py
```diff
@@ -1,9 +1,16 @@
...
-from typing import Optional
+from typing import Any, Optional
 
-from ai.backend.manager.models.user import UserRole, UserStatus
+from ai.backend.manager.models.keypair import KeyPairRow
+from ai.backend.manager.models.resource_policy import (
+    KeyPairResourcePolicyRow,
+    UserResourcePolicyRow,
+)
+from ai.backend.manager.models.user import UserRole, UserRow, UserStatus
```
fregataa: data modules should not depend on any model modules

src/ai/backend/manager/api/auth.py
```diff
@@ -557,20 +479,19 @@ async def _authenticate_via_jwt(
         if not access_key:
             raise AuthorizationFailed("Access key not found in JWT token")
 
-        # 2. Query keypair and user from database to get secret_key
-        user_row, keypair_row = await execute_with_retry(
-            functools.partial(_query_cred_by_access_key, root_ctx, access_key)
-        )
+        # 2. Query credential from database via repository
+        auth_repo = root_ctx.repositories.auth.repository
+        credential = await auth_repo.get_credential_by_access_key(access_key)
```
fregataa: API handlers should access to auth service(`root_ctx.processors.auth`) rather than using repository directly

src/ai/backend/manager/data/auth/types.py
```diff
@@ -47,3 +54,73 @@ class UserData:
...
+        self,
+        user_row: UserRow,
+        keypair_row: KeyPairRow,
+        user_resource_policy_row: UserResourcePolicyRow,
+        keypair_resource_policy_row: KeyPairResourcePolicyRow,
+    ) -> None:
+        self._user_row = user_row
+        self._keypair_row = keypair_row
+        self._user_rp_row = user_resource_policy_row
+        self._keypair_rp_row = keypair_resource_policy_row
```
fregataa: It is recommended to use `dataclass` in such case
```suggestion

@dataclass
class CredentialData:
    """
    Aggregated credential data for authentication.

    Contains ORM Row objects directly to ensure exact field compatibility
    with the legacy request["user"] and request["keypair"] dict format.
    Used by authentication middleware to populate request context.
    """
    user_row: UserRow
    keypair_row: KeyPairRow
    user_resource_policy_row: UserResourcePolicyRow
    keypair_resource_policy_row: KeyPairResourcePolicyRow
```

src/ai/backend/manager/data/auth/types.py
```diff
@@ -47,3 +54,73 @@ class UserData:
...
+    def is_admin(self) -> bool:
+        """Check if the keypair has admin privileges."""
+        return self._keypair_row.is_admin or False
+
+    @property
+    def is_superadmin(self) -> bool:
+        """Check if the user has superadmin role."""
+        return self._user_row.role == UserRole.SUPERADMIN
+
+    def to_keypair_dict(self) -> dict[str, Any]:
```
fregataa: How about defining a new typed-dict using `typing.TypedDict`?

src/ai/backend/manager/data/auth/types.py
```diff
@@ -47,3 +54,73 @@ class UserData:
...
+            for col in KeyPairRow.__table__.c
+            if col.name != "secret_key"
+        }
+        result["resource_policy"] = {
+            col.name: getattr(self._keypair_rp_row, col.name)
+            for col in KeyPairResourcePolicyRow.__table__.c
+        }
+        return result
+
+    def to_user_dict(self) -> dict[str, Any]:
```
fregataa: Same here

fregataa (changes requested):
I left comments. Please ask me any question if you need help

src/ai/backend/manager/api/auth.py
kwonkwonn: I've been thinking about best approach here, as get_credential_by_access_key here is more like pure orm logic,
i would move it into repository dir and make other services utilize it.(as refering this function only happens in `api/auth`. 

if i'm missing something please let me know!

kwonkwonn:
As I investigate patterns in the `api` layer, I've identified 
three distinct function types that are currently intermingled. 
Here's my refactoring plan to separate concerns:

1. Direct Access Pattern
   - API handler following the convention: 
     api → action → processors → service → directory → db_source
example: https://github.com/lablup/backend.ai/blob/67de09534f7fe4f9ef5bb7f5a6e64b37eb6f6aff/src/ai/backend/manager/api/stream.py#L69
 
2. Pure ORM Operations
   - Database access logic that should be isolated into 
     a separate directory
example: https://github.com/lablup/backend.ai/blob/67de09534f7fe4f9ef5bb7f5a6e64b37eb6f6aff/src/ai/backend/manager/api/auth.py#L429
   
3. Mixed Type (ORM + Utility/Middleware)
   - This is where I'm uncertain. I want to organize these 
     in a separate directory, but it's challenging to determine:
     • Which logic is ORM vs. utility
     • Which APIs actually need which utilities
     • How to prevent circular dependencies
 utility func example:
https://github.com/lablup/backend.ai/blob/67de09534f7fe4f9ef5bb7f5a6e64b37eb6f6aff/src/ai/backend/manager/api/auth.py#L584
orm+ utility: https://github.com/lablup/backend.ai/blob/67de09534f7fe4f9ef5bb7f5a6e64b37eb6f6aff/src/ai/backend/manager/api/logs.py#L211


I think best approach for now is resolve first two types and plan for third one with some time.

src/ai/backend/manager/api/auth.py
fregataa: yes, the best approach is adding a new service function.

fregataa:
The goal of the original issue is refactoring Auth API handlers. I didn't get why such separate steps are required.
Do you want to refactor all API handlers by suggested plan?

> As I investigate patterns in the `api` layer, I've identified three distinct function types that are currently intermingled. Here's my refactoring plan to separate concerns:
> 
> 1. Direct Access Pattern
>    
>    * API handler following the convention:
>      api → action → processors → service → directory → db_source
>      example: https://github.com/lablup/backend.ai/blob/67de09534f7fe4f9ef5bb7f5a6e64b37eb6f6aff/src/ai/backend/manager/api/stream.py#L69
> 2. Pure ORM Operations
>    
>    * Database access logic that should be isolated into
>      a separate directory
>      example: https://github.com/lablup/backend.ai/blob/67de09534f7fe4f9ef5bb7f5a6e64b37eb6f6aff/src/ai/backend/manager/api/auth.py#L429
> 3. Mixed Type (ORM + Utility/Middleware)
>    
>    * This is where I'm uncertain. I want to organize these
>      in a separate directory, but it's challenging to determine:
>      • Which logic is ORM vs. utility
>      • Which APIs actually need which utilities
>      • How to prevent circular dependencies
>      utility func example:
>      https://github.com/lablup/backend.ai/blob/67de09534f7fe4f9ef5bb7f5a6e64b37eb6f6aff/src/ai/backend/manager/api/auth.py#L584
>      
>      orm+ utility: https://github.com/lablup/backend.ai/blob/67de09534f7fe4f9ef5bb7f5a6e64b37eb6f6aff/src/ai/backend/manager/api/logs.py#L211
> 
> I think best approach for now is resolve first two types and plan for third one with some time.

kwonkwonn:
Maybe i was overthinking about it. you are right. 
I will revise commits and resubmit as you reviewed shortly.

otherwise, I'm happy to refactor the other API layers too, if it's worth it.

src/ai/backend/manager/data/auth/types.py
```diff
@@ -1,9 +1,293 @@
...
+    ResourceSlot,
+    VFolderHostPermissionMap,
+)
+from ai.backend.manager.data.user.types import UserRole, UserStatus
+
+# TypedDict definitions for legacy request dict format compatibility
+
+
+class KeypairResourcePolicyDict(TypedDict):
+    """Legacy format for request['keypair']['resource_policy']."""
```
jopemachine, line 20: If it’s a legacy type, how about adding a Legacy prefix to the type names?

src/ai/backend/manager/repositories/auth/db_source/db_source.py
```diff
@@ -291,3 +299,119 @@ async def fetch_current_time(self) -> datetime:
             result = await db_conn.scalar(sa.select(sa.func.now()))
             assert result is not None
             return result
+
+    @auth_db_source_resilience.apply()
+    async def fetch_credential_by_access_key(self, access_key: str) -> Optional[CredentialData]:
```
jopemachine, line 304: Unless there is a clear need otherwise, let’s make it return the required CredentialData, and throw an error if it’s missing.

PR #8010, opened 2026-01-13
docs: Add BEP-1027 Entity Querier for RBAC

fregataa:
resolves #NNN (BA-MMM)

## Summary
This PR introduces a design proposal for the Entity Querier abstraction in the RBAC system.

 The Entity Querier provides a clean way to enrich entity search results with human-readable names. When querying entities within a scope (e.g., users in a project), the current API only returns entity IDs. This proposal defines an extensible pattern to include entity names by joining with the appropriate entity tables.

  Key aspects of the proposal:
  - Defines an abstract interface for entity-specific query logic
  - Provides a base implementation for simple table lookups
  - Ensures all entity types have querier implementations through static type checking and unit tests
  - Maintains clean layer separation by placing the abstraction in the Repository layer

- [ ] Test case(s) to:
- [ ] Documentation


fregataa:
Closed this PR because this is overengineering and can cause heavy stress to db

PR #8048, opened 2026-01-15
feat(BA-3698): apply RBAC Creator pattern to Model Deployment

### LINKED ISSUE #7728: Apply RBAC Creator pattern to Model Deployment
Apply RBAC Creator pattern to Model Deployment

JIRA Issue: BA-3698

fregataa:
resolves #7728 (BA-3698)

- Add RBAC scope association when creating endpoints
- Create USER scope association for endpoint ownership
- Add _LegacyEndpointCreatorSpec wrapper for legacy creator pattern
- Update both create_endpoint() and create_endpoint_legacy() methods

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/tests/unit/manager/repositories/deployment/test_deployment_repository.py
+++ b/tests/unit/manager/repositories/deployment/test_deployment_repository.py
@@ -52,6 +52,7 @@ from ai.backend.manager.models.image import ImageRow
 from ai.backend.manager.models.kernel import KernelRow, KernelStatus
 from ai.backend.manager.models.keypair import KeyPairRow
 from ai.backend.manager.models.rbac_models import UserRoleRow
+from ai.backend.manager.models.rbac_models.entity_field import EntityFieldRow
 from ai.backend.manager.models.resource_policy import (
     KeyPairResourcePolicyRow,
     ProjectResourcePolicyRow,
@@ -1187,6 +1188,7 @@ class TestDeploymentRevisionOperations:
                 VFolderRow,
                 ImageRow,
                 EndpointRow,
+                EntityFieldRow,  # DeploymentRevisionRow relationship dependency
                 DeploymentRevisionRow,
             ],
         ):


src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -140,6 +154,19 @@ class EndpointWithRoutesRawData:
     route_rows: list[RoutingRow]
 
 
+class _LegacyEndpointCreatorSpec(CreatorSpec[EndpointRow]):
```
jopemachine: This type is newly introduced, so could you clarify why it is labeled as Legacy?

fregataa: Because `DeploymentCreatorSpec` already exists. And this new spec supports legacy codes

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -225,9 +251,16 @@ async def create_endpoint(
...
-            db_sess.add(endpoint)
-            await db_sess.flush()
+
+            # Create endpoint with RBAC scope association
+            rbac_creator: RBACEntityCreator[EndpointRow] = RBACEntityCreator(
+                spec=spec,
+                scope_type=ScopeType.USER,
+                scope_id=str(spec.metadata.created_user_id),
+                entity_type=EntityType.MODEL_DEPLOYMENT,
+            )
```
HyeockJinKim: It looks better to create the all Creator including RBACEntityCreator at the point of creation rather than generating them from the Repository.
We should avoid creating RBAC Creators unintentionally in the API or service layers. @fregataa

HyeockJinKim (changes requested):

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -140,6 +153,19 @@ class EndpointWithRoutesRawData:
...
+
+    Wraps a pre-built EndpointRow for use with RBACEntityCreator.
+    """
+
+    def __init__(self, *, _row: EndpointRow) -> None:
+        self._row = _row
+
+    def build_row(self) -> EndpointRow:
+        return self._row
+
```
HyeockJinKim: I don’t think this is right. Why not define new CreatorSpec for EndpointRow?

fregataa: ~~There will be too many changes. I will refactor this in another task~~
I updated this part

src/ai/backend/manager/sokovan/deployment/deployment_controller.py
```diff
@@ -109,10 +112,19 @@ async def create_deployment(
         await self._scheduling_controller.validate_session_spec(
             SessionValidationSpec.from_revision(model_revision=model_revision)
         )
+        image_id = await self._deployment_repository.get_image_id(model_revision.image_identifier)
```
jopemachine, line 115: How about injecting `image_repository` in this instead of defining `get_image_id` in the `deployment_repository`?

fregataa: Since model deployment and image are related to each other, we do not need to separate repositories

changes/8048.feature.md
```diff
@@ -0,0 +1 @@
+Apply RBAC Creator pattern to Model Deployment, automatically creating USER scope associations when endpoints are created for proper ownership tracking and permission management.
```
jopemachine, line 1: Nit: 
```suggestion
Apply RBAC Creator pattern to Model Deployment, automatically creating USER scope associations when endpoints are created for proper ownership tracking and permission management
```

PR #8070, opened 2026-01-15
fix(BA-3913): Maintain status_history JSONB field updates

HyeockJinKim:
Add status_history updates to ensure backward compatibility with existing code that relies on this field:

- SessionStatusBatchUpdaterSpec.build_values(): Add status_history
  merge for all batch status updates via Coordinator
- mark_session_cancelled(): Add status_history for both SessionRow and KernelRow when cancelling sessions

This ensures status_history is updated in parallel with the new scheduling_history system.

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

resolves #NNN (BA-MMM)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/changes/8070.fix.md
+++ b/changes/8070.fix.md
@@ -1,0 +1,1 @@
+Maintain status_history JSONB field updates


src/ai/backend/manager/repositories/scheduler/updaters.py
```diff
@@ -30,11 +31,18 @@ def row_class(self) -> type[SessionRow]:
 
     @override
     def build_values(self) -> dict[str, Any]:
-        values: dict[str, Any] = {"status": self.to_status}
+        now = datetime.now(tzutc())
```
fregataa, line 34: no more tzutc. Let's use datetime.UTC instead

PR #8092, opened 2026-01-16
refactor(BA-3878): replace RequesterCtx with UserData in model_serving module

MinJuTur:
## Summary
- Replace `RequesterCtx` with `UserData` from `ai.backend.common.data.user.types` across the model_serving module
- Add deprecation notice to `RequesterCtx` class
- Update all action classes, services, repositories, and API handlers to use `UserData`
- Update all related unit tests

## Related Issue
- Jira: BA-3878

## Changes
- **Types**: Added deprecated docstring to `RequesterCtx` class
- **Utils**: Updated `verify_user_access_scopes` and `validate_endpoint_access` functions
- **Actions**: Updated 14 action classes to use `user_data: UserData` instead of `requester_ctx: RequesterCtx`
- **Services**: Updated `model_serving.py` and `auto_scaling.py`
- **Repository**: Updated `repository.py` with proper `UserRole` conversion
- **API**: Updated `service.py` and `endpoint.py` to create `UserData` instances
- **Tests**: Updated 11 test files to use `UserData`

## Test plan
- [x] All existing unit tests pass
- [x] Type checking passes
- [x] Linting passes

### DIFF
--- a/tests/unit/manager/services/model_serving/actions/test_create_auto_scaling_rule.py
+++ b/tests/unit/manager/services/model_serving/actions/test_create_auto_scaling_rule.py
@@ -55,153 +55,103 @@ def mock_create_auto_scaling_rule(mocker, mock_repositories):
 
 class TestCreateEndpointAutoScalingRule:
     @pytest.mark.parametrize(
-        ("scenario", "user_data"),
+        "scenario",
         [
-            (
-                ScenarioBase.success(
-                    "CPU based scaling",
-                    CreateEndpointAutoScalingRuleAction(
-                        endpoint_id=EndpointId(uuid.UUID("11111111-1111-1111-1111-111111111111")),
-                        creator=EndpointAutoScalingRuleCreator(
-                            metric_source=AutoScalingMetricSource.KERNEL,
-                            metric_name="cpu_utilization",
-                            threshold="70.0",
-                            comparator=AutoScalingMetricComparator.GREATER_THAN,
-                            step_size=1,
-                            cooldown_seconds=300,
-                            min_replicas=2,
-                            max_replicas=10,
-                        ),
-                    ),
-                    CreateEndpointAutoScalingRuleActionResult(
-                        success=True,
-                        data=MagicMock(id=uuid.UUID("22222222-2222-2222-2222-222222222222")),
+            ScenarioBase.success(
+                "CPU based scaling",
+                CreateEndpointAutoScalingRuleAction(
+                    endpoint_id=EndpointId(uuid.UUID("11111111-1111-1111-1111-111111111111")),
+                    creator=EndpointAutoScalingRuleCreator(
+                        metric_source=AutoScalingMetricSource.KERNEL,
+                        metric_name="cpu_utilization",
+                        threshold="70.0",
+                        comparator=AutoScalingMetricComparator.GREATER_THAN,
+                        step_size=1,
+                        cooldown_seconds=300,
+                        min_replicas=2,
+                        max_replicas=10,
                     ),
                 ),
-                UserData(
-                    user_id=uuid.UUID("00000000-0000-0000-0000-000000000001"),
-                    is_authorized=True,
-                    is_admin=False,
-                    is_superadmin=False,
-                    role=UserRole.USER.value,
-                    domain_name="default",
+                CreateEndpointAutoScalingRuleActionResult(
+                    success=True,
+                    data=MagicMock(id=uuid.UUID("22222222-2222-2222-2222-222222222222")),
                 ),
             ),
-            (
-                ScenarioBase.success(
-                    "Request count based scaling",
-                    CreateEndpointAutoScalingRuleAction(
-                        endpoint_id=EndpointId(uuid.UUID("33333333-3333-3333-3333-333333333333")),
-                        creator=EndpointAutoScalingRuleCreator(
-                            metric_source=AutoScalingMetricSource.KERNEL,
-                            metric_name="requests_per_second",
-                            threshold="100.0",
-                            comparator=AutoScalingMetricComparator.GREATER_THAN,
-                            step_size=2,
-                            cooldown_seconds=600,
-                            min_replicas=1,
-                            max_replicas=20,
-                        ),
-                    ),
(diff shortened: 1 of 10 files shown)


src/ai/backend/manager/api/gql_legacy/endpoint.py
```diff
@@ -450,10 +448,12 @@ async def mutate(
...
-                is_authorized=None,
+            user_data=UserData(
                 user_id=info.context.user["uuid"],
-                user_role=info.context.user["role"],
+                is_authorized=True,
+                is_admin=info.context.user["role"] == UserRole.ADMIN,
+                is_superadmin=info.context.user["role"] == UserRole.SUPERADMIN,
+                role=info.context.user["role"].value,
                 domain_name=info.context.user["domain_name"],
             ),
```
HyeockJinKim: I don’t think this actions(endpoint related) needs to have access to user information; 
could you please check the `current_user()` using contextVar?
Also, could you review the direction before starting the issue work?

jopemachine: I mentioned this once before, but I would appreciate it if you could share the direction you’re planning to take before starting the work. @MinJuTur

src/ai/backend/manager/repositories/model_serving/repository.py
```diff
@@ -731,15 +731,16 @@ async def _do_mutate() -> MutationResult:
                         load_routes=True,
                         load_image=True,
                     )
-                    match action.requester_ctx.user_role:
+                    user_role = UserRole(action.user_data.role)
```
jopemachine: I think `action.user_data.role` already should be `UserRole` insteadof `str`

src/ai/backend/manager/data/model_serving/types.py
```diff
@@ -238,6 +238,11 @@ class CompactServiceInfo:
 
 @dataclass
 class RequesterCtx:
+    """
+    .. deprecated::
```
jopemachine: Other comments aren’t written that way either, so I don’t think we need to insist on using the Sphinx style. What do you think about comment style? @HyeockJinKim

jopemachine (approved):

src/ai/backend/manager/repositories/model_serving/repository.py
MinJuTur: > I think `action.user_data.role` already should be `UserRole`

`UserData.role` is currently defined as `str` type in `ai.backend.common.data.user.types.py`, so the conversion to `UserRole` is necessary here. Are you suggesting that `UserData.role` should be changed from `str` to `UserRole`? @jopemachine

jopemachine: > Are you suggesting that UserData.role should be changed from str to UserRole?

Yes, I think it should be updated

MinJuTur: > Yes, I think it should be updated

Changing `UserData.role` from `str` to `UserRole` would affect the common package and require changes across multiple modules. I'll create a separate issue/PR for this refactoring to keep the scope of this PR focused.

tests/unit/manager/services/model_serving/test_auto_scaling_rule_validated_service.py
```diff
@@ -145,9 +146,9 @@ async def test_search_auto_scaling_rules_with_pagination(
         )
         action = SearchAutoScalingRulesAction(
             querier=querier,
-            user_data=sample_user_data,
         )
-        result = await auto_scaling_service.search_auto_scaling_rules(action)
+        with with_user(sample_user_data):
```
jopemachine: Could we avoid calling `with_user` in every test and instead apply it as shared logic through a fixture?

jopemachine:
Let's define fixture like this in `conftest.py`

```python
SUPERADMIN_USER_DATA = UserData(
    user_id=uuid.UUID("00000000-0000-0000-0000-000000000002"),
    is_authorized=True,
    is_admin=False,
    is_superadmin=True,
    role=UserRole.SUPERADMIN.value,
    domain_name="default",
)
```

And inject like this

```python
class TestCreateEndpointAutoScalingRuleAsSuperadmin:
    @pytest.fixture
    def user_data(self) -> UserData:
        """Override user_data fixture for superadmin tests."""
        return SUPERADMIN_USER_DATA
```

I’d like to remove conftest.py entirely, but that would significantly expand the scope of this PR, so I’ll leave it as is for now.

jopemachine:
FYI, please use parametrize only when the test outcome actually changes depending on the parameter values.
Otherwise, it just introduces unnecessary complexity without providing any benefit.

MinJuTur:
> FYI, please use parametrize only when the test outcome actually changes depending on the parameter values. Otherwise, it just introduces unnecessary complexity without providing any benefit.

Thanks for the feedback. Should I remove the redundant parametrized cases (e.g., USER and SUPERADMIN with same expected result) in this PR, or is this just a note for future reference?

jopemachine:
> Thanks for the feedback. Should I remove the redundant parametrized cases (e.g., USER and SUPERADMIN with same expected result) in this PR, or is this just a note for future reference?

I think since the result is the same anyway, it seems sufficient to keep just one.

MinJuTur:
I've created Jira issues for the items you mentioned earlier.

> Changing `UserData.role` from `str` to `UserRole` would affect the common package and require changes across multiple modules. I'll create a separate issue/PR for this refactoring to keep the scope of this PR focused.

https://lablup.atlassian.net/browse/BA-3943?atlOrigin=eyJpIjoiZjIxZjhkZjIwYjZhNDM4ODk1NTdhOTNiNDljZWE0ODEiLCJwIjoiaiJ9


> I’d like to remove conftest.py entirely, but that would significantly expand the scope of this PR, so I’ll leave it as is for now.

https://lablup.atlassian.net/browse/BA-3946?atlOrigin=eyJpIjoiOWY4ZmVhOWQ2OTM3NDNhZmIxMjMzYTQxZWNiNTE4NDkiLCJwIjoiaiJ9

jopemachine (approved):

PR #8121, opened 2026-01-19
fix(BA-3924): Return `UNHEALTHY` for endpoints with no active routes instead of `READY`

jopemachine:
resolves (BA-3924)

- [x] Test case(s) to:
- [ ] Documentation


HyeockJinKim (changes requested):
If endpoint's status is ready, it mean endpoint can receive traffic.
But applying this change would allow user scenarios that request traffic without any routing.

HyeockJinKim (approved):

PR #8180, opened 2026-01-21
refactor(BA-3758): add Purger pattern to User repository

MinJuTur:
## Summary

- Apply the purger pattern to `UserRepository.purge_user()` method for consistent and maintainable user data deletion
- Add `purgers.py` with 5 `PurgerSpec` classes and factory functions for user-related data deletion:
  - `UserErrorLogPurgerSpec` / `create_user_error_log_purger`
  - `UserKeyPairPurgerSpec` / `create_user_keypair_purger`
  - `UserVFolderPermissionPurgerSpec` / `create_user_vfolder_permission_purger`
  - `UserGroupAssociationPurgerSpec` / `create_user_group_association_purger`
  - `UserPurgerSpec` / `create_user_purger`
- Refactor `purge_user()` to use `execute_batch_purger()` with factory functions
- Remove legacy `_delete_*` private methods from repository
- Add comprehensive test coverage for all purgers (unit tests + integration tests)

## Jira

- [BA-3758](https://lablup.atlassian.net/browse/BA-3758)

## Test plan

- [x] All existing tests pass
- [x] New unit tests for PurgerSpec classes verify correct SQL generation
- [x] New unit tests for factory functions verify correct BatchPurger creation
- [x] New integration tests verify actual data deletion with real database
- [x] Type checking passes (`pants check`)
- [x] Linting passes (`pants lint`)

[BA-3758]: https://lablup.atlassian.net/browse/BA-3758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/unit/manager/repositories/user/test_user_purgers.py
+++ b/tests/unit/manager/repositories/user/test_user_purgers.py
@@ -20,21 +20,21 @@ from ai.backend.manager.data.vfolder.types import (
     VFolderOperationStatus,
     VFolderOwnershipType,
 )
-from ai.backend.manager.models.agent import AgentRow
+from ai.backend.manager.models.agent import AgentRow  # noqa: F401
 from ai.backend.manager.models.domain import DomainRow
 from ai.backend.manager.models.error_logs import ErrorLogRow
 from ai.backend.manager.models.group import AssocGroupUserRow, GroupRow, ProjectType
 from ai.backend.manager.models.hasher.types import PasswordInfo
-from ai.backend.manager.models.image import ImageRow
-from ai.backend.manager.models.kernel import KernelRow
+from ai.backend.manager.models.image import ImageRow  # noqa: F401
+from ai.backend.manager.models.kernel import KernelRow  # noqa: F401
 from ai.backend.manager.models.keypair import KeyPairRow
 from ai.backend.manager.models.resource_policy import (
     KeyPairResourcePolicyRow,
     ProjectResourcePolicyRow,
     UserResourcePolicyRow,
 )
-from ai.backend.manager.models.scaling_group import ScalingGroupRow
-from ai.backend.manager.models.session import SessionRow
+from ai.backend.manager.models.scaling_group import ScalingGroupRow  # noqa: F401
+from ai.backend.manager.models.session import SessionRow  # noqa: F401
 from ai.backend.manager.models.user import UserRole, UserRow, UserStatus
 from ai.backend.manager.models.vfolder.row import VFolderPermissionRow, VFolderRow
 from ai.backend.manager.repositories.base.purger import BatchPurger, execute_batch_purger
@@ -54,10 +54,6 @@ from ai.backend.testutils.db import with_tables
 
 if TYPE_CHECKING:
     from ai.backend.manager.models.utils import ExtendedAsyncSAEngine
-
-# Force mapper initialization by importing Row classes with relationships
-# This ensures all related mappers are initialized before tests run
-_ = (AgentRow, KernelRow, SessionRow, ImageRow, ScalingGroupRow)
 
 
 class TestUserPurgerSpecs:


src/ai/backend/manager/repositories/user/purgers.py
```diff
@@ -0,0 +1,107 @@
...
+from ai.backend.manager.models.error_logs import ErrorLogRow
+from ai.backend.manager.models.group import AssocGroupUserRow
+from ai.backend.manager.models.keypair import KeyPairRow
+from ai.backend.manager.models.user import UserRow
+from ai.backend.manager.models.vfolder import VFolderPermissionRow
+from ai.backend.manager.repositories.base.purger import BatchPurger, BatchPurgerSpec
+
+
+@dataclass
+class UserErrorLogPurgerSpec(BatchPurgerSpec[ErrorLogRow]):
```
jopemachine:
```suggestion
class UserErrorLogBatchPurgerSpec(BatchPurgerSpec[ErrorLogRow]):
```

jopemachine: Same to other class names

tests/unit/manager/repositories/user/test_user_purgers.py
```diff
@@ -0,0 +1,592 @@
...
+    create_user_vfolder_permission_purger,
+)
+from ai.backend.testutils.db import with_tables
+
+if TYPE_CHECKING:
+    from ai.backend.manager.models.utils import ExtendedAsyncSAEngine
+
+# Force mapper initialization by importing Row classes with relationships
+# This ensures all related mappers are initialized before tests run
+_ = (AgentRow, KernelRow, SessionRow, ImageRow, ScalingGroupRow)
```
jopemachine: Could you explain this?

tests/unit/manager/repositories/user/test_user_purgers.py
```diff
@@ -0,0 +1,592 @@
...
+
+
+def create_test_password_info(password: str = "test_password") -> PasswordInfo:
```
jopemachine: What about moving this into TestUserPurgerSpecs's fixture?
and default value seems meaningless

tests/unit/manager/repositories/user/test_user_purgers.py
```diff
@@ -0,0 +1,592 @@
...
+                    is_active=True,
+                    domain_name=sample_domain,
+                    total_resource_slots={},
+                    allowed_vfolder_hosts={},
+                    integration_id=None,
+                    resource_policy=project_resource_policy,
+                    type=ProjectType.GENERAL,
+                )
+                session.add(group)
+                await session.flush()
```
jopemachine: Do not insert db row in purge test.
Make fixture

tests/unit/manager/repositories/user/test_user_purgers.py
```diff
@@ -0,0 +1,592 @@
...
+        # Verify permissions are deleted
+        async with db_with_cleanup.begin_session() as session:
+            count = await session.scalar(
+                sa.select(sa.func.count())
+                .select_from(VFolderPermissionRow)
+                .where(VFolderPermissionRow.user == user_uuid)
+            )
+            assert count == 0
+
+    @pytest.mark.asyncio
```
jopemachine, line 568: We don't need this

tests/unit/manager/repositories/user/test_user_purgers.py
```diff
@@ -0,0 +1,592 @@
...
+            assert result.deleted_count == 1
+
+        # Verify user is deleted
+        async with db_with_cleanup.begin_session() as session:
+            count = await session.scalar(
+                sa.select(sa.func.count()).select_from(UserRow).where(UserRow.uuid == user_uuid)
+            )
+            assert count == 0
+
+    @pytest.mark.asyncio
```
jopemachine: Don't make this kind of test.
create_user_xxx is just wrapper of BatchPurger constructor.
This test `execute_batch_purger`, which is out of scope from this test

tests/unit/manager/repositories/user/test_user_purgers.py
MinJuTur: > Could you explain this?

KeyPairRow has string-based relationships to these Row classes. Importing them triggers SQLAlchemy mapper registration, which is required before queries can resolve the relationships. Without this, tests fail with "mapper not found" errors.

tests/unit/manager/repositories/user/test_user_purgers.py
MinJuTur: > What about moving this into TestUserPurgerSpecs's fixture?

I moved it to `TestUserPurgersIntegration` as a fixture since that's the only class using it. Did you mean `TestUserPurgersIntegration` instead of `TestUserPurgerSpecs`?

jopemachine:
The test I thought was unnecessary is `test_purge_nonexistent_user_data`. The others are not.
Is there a reason you think the other tests are unnecessary as well?

.claude/tasks/consolidate-admin-repository.md
jopemachine, line 1: I'd recommend to add it to the global .gitignore.

MinJuTur:
> The test I thought was unnecessary is `test_purge_nonexistent_user_data`. The others are not. Is there a reason you think the other tests are unnecessary as well?

 I misinterpreted your original comment and thought all integration tests were out of scope since they all use execute_batch_purger(). I've now restored the other integration tests and only removed test_purge_nonexistent_user_data as you intended.

.gitignore
```diff
@@ -160,3 +160,4 @@ backup.*.json
 /DIST-INFO
 /INSTALL-INFO
 .serena/
+.claude/tasks/
```
jopemachine: I meant your personal global gitignore file.
Please remove this.
We might add claude to .gitignore in the future, but that’s not what we’re trying to do right now.

src/ai/backend/manager/repositories/user/repository.py
```diff
@@ -751,22 +758,14 @@ async def _get_user_uuid_by_email(self, conn: SAConnection, email: str) -> Optio
...
-    async def _delete_error_logs(self, conn: SAConnection, user_uuid: UUID) -> int:
-        """Delete user's error logs."""
-        result = await conn.execute(sa.delete(error_logs).where(error_logs.c.user == user_uuid))
-        return result.rowcount
-
-    async def _delete_keypairs(self, conn: SAConnection, user_uuid: UUID) -> int:
-        """Delete user's keypairs."""
-        result = await conn.execute(sa.delete(keypairs).where(keypairs.c.user == user_uuid))
-        return result.rowcount
+    async def _get_user_uuid_by_email_with_session(
```
jopemachine: Function name seems inappropriate

src/ai/backend/manager/repositories/user/repository.py
```diff
@@ -751,22 +758,14 @@ async def _get_user_uuid_by_email(self, conn: SAConnection, email: str) -> Optio
...
+        self, session: SASession, email: str
+    ) -> Optional[UUID]:
```
jopemachine: Let's return UUID, and raise exception in the method

tests/unit/manager/repositories/user/test_user_purgers.py
jopemachine: It seems that the tests still pass even after removing that line. Could you please double-check? @MinJuTur

src/ai/backend/manager/repositories/user/repository.py
MinJuTur: > Function name seems inappropriate

This function uses SASession, while `_get_user_uuid_by_email` on line 755 uses SAConnection. I added the `_with_session` suffix to distinguish between them.
How about renaming line 755 to `_get_user_uuid_by_email_with_conn` to match the `_with_conn` / `_with_session` pattern already used in this file (e.g., `_get_user_by_email_with_conn`)?  @jopemachine

MinJuTur: I just noticed that `_get_user_uuid_by_email` on line 755 is not used anywhere. Only `_get_user_uuid_by_email_with_session` is called (at line 653).
How about deleting the unused function on line 755 and renaming `_get_user_uuid_by_email_with_session` to `_get_user_uuid_by_email`?

tests/unit/manager/repositories/user/test_user_purgers.py
MinJuTur: > It seems that the tests still pass even after removing that line. Could you please double-check?

The tests pass when only removing the `_ = (...)` line because the imports themselves still remain. However, when running `pants fix` (which includes `ruff check --fix`), ruff removes these imports as "unused" since they're not directly referenced anywhere else in the code.
After `pants fix` removes the imports, the tests fail because the SQLAlchemy mapper for KeyPairRow's relationships is not initialized. @jopemachine

jopemachine: Then you can attach `# noqa: F401` to that imports

jopemachine: https://github.com/lablup/backend.ai/blob/94b0b7fb0486a06f9f14bbca9da27df0283a16ba/tests/unit/manager/conftest.py#L18-L22
Please refer to the existing codebase.
You can use # noqa: F401 only once and group all unused imports together.

MinJuTur: > Please refer to the existing codebase.
> You can use # noqa: F401 only once and group all unused imports together.

These imports come from different modules, so they can't be combined into a single import statement. Should I create a re-export module in ai.backend.testutils to group them together? @jopemachine

jopemachine: Never mind, this seems intended behavior
https://github.com/lablup/backend.ai/blob/ce4d41d51e7254cd6816a33a5c3be36f519f2ac6/src/ai/backend/manager/models/__init__.py#L1-L9

jopemachine:
I will merge this after release

src/ai/backend/manager/repositories/user/purgers.py
```diff
@@ -0,0 +1,107 @@
...
+        spec=UserGroupAssociationBatchPurgerSpec(user_uuid=user_uuid),
+    )
+
+
+def create_user_purger(user_uuid: UUID) -> BatchPurger[UserRow]:
+    """Create a BatchPurger for deleting a user."""
+    return BatchPurger(
+        spec=UserBatchPurgerSpec(user_uuid=user_uuid),
+        batch_size=1,  # We expect only one row to be deleted
+    )
```
HyeockJinKim, line 107: Is there really any reason to create a separate creation method?

MinJuTur: > Is there really any reason to create a separate creation method?

I followed the existing pattern from `scaling_group/purgers.py`, which also provides both `*PurgerSpec` classes and `create_*_purger()` factory functions.
But you're right—there's no real benefit to having separate factory functions.

PR #8201, opened 2026-01-21
fix: Return empty GPFS performance metric rather than raising exception

fregataa:
resolves #NNN (BA-MMM)

## Summary
  - Fixed unhandled GPFSNoMetricError exception in GPFS volume performance metric collection
  - When GPFS performance metric data is unavailable, return an empty FSPerfMetric with zero values instead of raising an exception

## Problem
  The GPFSNoMetricError exception was raised when GPFS API returned no performance metric data (due to KeyError or IndexError), but this exception was not caught or handled anywhere in the codebase. This caused unexpected crashes when metrics were temporarily unavailable.

##  Solution
  Instead of raising an unhandled exception, return a valid FSPerfMetric object with all values set to zero. This allows the system to gracefully handle cases where metric data is unavailable without disrupting normal operations.

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/storage/volumes/gpfs/exceptions.py
+++ b/src/ai/backend/storage/volumes/gpfs/exceptions.py
@@ -37,10 +37,6 @@ class GPFSInternalError(GPFSAPIError):
     pass
 
 
-class GPFSNoMetricError(GPFSError):
-    pass
-
-
 class GPFSJobFailedError(GPFSError):
     pass
 


src/ai/backend/storage/volumes/gpfs/__init__.py
```diff
@@ -256,7 +255,14 @@ async def get_performance_metric(self) -> FSPerfMetric:
             metrics = await self.api_client.get_metric(query)
             latest_metric = metrics["performanceData"]["rows"][-1]["values"]
         except (KeyError, IndexError):
-            raise GPFSNoMetricError
```
jopemachine, line 259: `GPFSNoMetricError` is not used anywhere, so we can remove it

jopemachine:
Is there a reason for changing the behavior that previously raised an exception?
I think it would be helpful if the rationale were explained in the PR or the related issue.

HyeockJinKim (changes requested):
It looks like return an empty value instead of raising an error might be riskier.
What's the reason for this work?

fregataa:
Closed this PR because the issue is not aligned

PR #8219, opened 2026-01-22
refactor(BA-4011): Remove TODOs in `manager/api/resource.py`

### LINKED ISSUE #8218: migrate legacy ToDos in `resource.py`
## Motivation

- There are some legacy TODOs in `manager/api/resource.py` which can be tackled now 

## Objective

- remove TODO which can be handled right now

## Details

1. Add comments for Duplicated function
1. remove unnecessary validation logic

## Impact

- 

JIRA Issue: BA-4011

kwonkwonn:
resolves #8218 (BA-4011)

Remove Legacy TODOs before migrating `resource.py`

- for validator, removed try-catch logic which is only for catching errors for not implemented scailing param validation.
which is now replaced with dict type enforcement.
- As `agent_watch` function has no relation with `resource.py` and is only reffered from `vfolder` moved into vfolder with added description for future plan.

- [ ] Test case(s) to:
- [ ] Documentation

----
📚 Documentation preview 📚: https://sorna--8219.org.readthedocs.build/en/8219/

----
📚 Documentation preview 📚: https://sorna-ko--8219.org.readthedocs.build/ko/8219/

### DIFF
--- a/docs/manager/rest-reference/openapi.json
+++ b/docs/manager/rest-reference/openapi.json
@@ -7877,8 +7877,44 @@
       }
     },
     "/resource/check-presets": {
-      "post": {
+      "get": {
         "operationId": "resource.check_presets",
+        "tags": [
+          "resource"
+        ],
+        "responses": {
+          "200": {
+            "description": "Successful response"
+          }
+        },
+        "security": [
+          {
+            "TokenAuth": []
+          }
+        ],
+        "parameters": [
+          {
+            "name": "scaling_group",
+            "schema": {
+              "type": "string"
+            },
+            "required": true,
+            "in": "query"
+          },
+          {
+            "name": "group",
+            "schema": {
+              "type": "string",
+              "default": "default"
+            },
+            "required": true,
+            "in": "query"
+          }
+        ],
+        "description": "\nReturns the list of all resource presets in the specified scaling group,\nwith additional information including allocatability of each preset,\namount of total remaining resources, and the current keypair resource limits.\n\n\n**Preconditions:**\n* User privilege required.\n* Manager status required: one of FROZEN, RUNNING\n"
+      },
+      "post": {
+        "operationId": "resource.check_presets_legacy",
         "tags": [
           "resource"
         ],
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/api/resource.py
```diff
@@ -81,7 +79,7 @@ async def list_presets(request: web.Request) -> web.Response:
 @auth_required
 @check_api_params(
     t.Dict({
-        t.Key("scaling_group", default=None): t.Null | t.String,
+        t.Key("scaling_group"): t.String,
```
HyeockJinKim, line 144: Modifying the existing schema doesn’t look ideal.
Although there’s a desire to enforce stricter value constraints, it’s not something we can change in the current API.
Introducing a new API for this purpose would be a better option.

kwonkwonn: Understood
I will look for better way to handle it, while retainig a exsiting api handlers.

Another small question here:
https://github.com/lablup/backend.ai/blob/681d882f71a4ee5a82adc63b1354d0f936ada09d/src/ai/backend/manager/api/resource.py#L388 
is being handled as "POST", although it doesn't effects current state, 
best practice i think would be "GET" with parameter querier. 

Current plan is adding new handler with "GET" with new function which we can use while transition.

docs/manager/rest-reference/openapi.json
```diff
@@ -7877,7 +7877,7 @@
       }
     },
     "/resource/check-presets": {
-      "post": {
+      "get": {
```
fregataa, line 7880: Could you check why the open api schema generator changed this?

kwonkwonn: https://github.com/lablup/backend.ai/pull/8219/changes/c2296765eb1dd3b5cad54b41b74e1e0d202634bf#diff-51adf4e1db91582bd045aea8a34e9565383e2c437c8cf7aa9eab64aebbdf9b3cR415-R416
```
    - cors.add(add_route("POST", "/check-presets", check_presets))
    + cors.add(add_route("GET", "/check-presets", check_presets))
    + cors.add(add_route("POST", "/check-presets", check_presets_legacy))
``` 


I guess that modifcation of OpenAPI would occured from here. 
With intention of  soft migration I've added Get handler for legacy Post handler which doesn't make sense As it doesn't modify any state of program.

but i still need to investigate why "post" were deleted instead of coexistencing.

kwonkwonn:
that make sense.

I'll just add some comments about discussions we've made and leave it for now.
So we can refer it next time when we can clean this stuff.

PR #8353, opened 2026-01-27
refactor(BA-3587): Define DTO types for auth handler's soft migration

### LINKED ISSUE #7626: Implement DTO model and actions for pydantic migration
## Motivation

**[Why is this Story important?]**

 The legacy `Trafaret` validation library has become a bottleneck, degrading code readability and type assurance. By migrating to **Pydantic**, we aim to leverage modern type hinting and more robust data validation. 

## Required Features

**[Clearly describe the features or functionality needed for this Story.]**

- **Pydantic Model Implementation:** Replace all `Trafaret` schemas with Pydantic `BaseModel` for stricter type safety and better IDE support.
- **Layered Architecture (Init/Adapter/Handler):**
  - **Handlers:** Manage request/response flow and entry points.
  - **Adapters:** Handle interactions with external systems or databases.
  - **__init__:** Orchestrate the public API of the module.

## Impact

**[Specify which components or services will be affected by this Story.]**

- **Validation Layer:** Complete replacement of validation logic across the auth module.
- **System Architecture:** Improved modularity, making it easier to swap out external dependencies (Adapters) without touching business logic (Handlers).
- **Maintainability:** Reduced cognitive load for developers through clear separation of concerns.

## Testing Scenarios

**[Outline test cases or scenarios for validating the implementation.]**

- **Serialization Audit:** Ensure that the `json()` output or `dict()` conversion matches the expected format of downstream services.
- **Strict Type Checking:** Run `mypy` to ensure that data flowing between Handlers and Adapters maintains type integrity via Pydantic models.
- **Linting & Style:** Execute `Ruff` or `Flake8` to ensure the new multi-file structure follows clean code standards.
- **Handler-to-Adapter Contract:** Mock the Adapter layer to verify the Handler correctly processes business logic.

JIRA Issue: BA-3591

kwonkwonn:
resolves #7626 (BA-3587)

fine grained PR from 
https://github.com/lablup/backend.ai/pull/7865/commits 

implements DTO and actions before using it as api_handler's arguments.

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/manager/api/legacy/auth.py
```diff
@@ -0,0 +1,1174 @@
+"""
+Legacy auth handlers (function-based).
```
kwonkwonn: This is nothing but copy-and-pasted version of old auth.py for regression test.

src/ai/backend/common/dto/manager/auth/response.py
```diff
@@ -0,0 +1,85 @@
...
+class UpdatePasswordResponse(BaseResponseModel):
+    """Response for update password."""
+
+    error_msg: Optional[str] = Field(default=None, description="Error message if failed")
+
+
+class UpdatePasswordNoAuthResponse(BaseResponseModel):
+    """Response for update password without auth."""
+
+    password_changed_at: datetime = Field(description="Timestamp when password was changed")
```
fregataa: The original API handler returns string typed `password_changed_at`. Is this serialize the value to a iso format string?

kwonkwonn:
```
def to_json(
    value: Any,
    *,
    indent: int | None = None,
    include: _IncEx | None = None,
    exclude: _IncEx | None = None,
    # Note: In Pydantic 2.11, the default value of `by_alias` on `SchemaSerializer` was changed from `True` to `None`,
    # to be consistent with the Pydantic "dump" methods. However, the default of `True` was kept here for
    # backwards compatibility. In Pydantic V3, `by_alias` is expected to default to `True` everywhere:
    by_alias: bool = True,
    exclude_none: bool = False,
    round_trip: bool = False,
    timedelta_mode: Literal['iso8601', 'float'] = 'iso8601',
    bytes_mode: Literal['utf8', 'base64', 'hex'] = 'utf8',
    inf_nan_mode: Literal['null', 'constants', 'strings'] = 'constants',
    serialize_unknown: bool = False,
    fallback: Callable[[Any], Any] | None = None,
    serialize_as_any: bool = False,
    context: Any | None = None,
) -> bytes:
```

From pydantic core, i guess it will detect the type(iso 8601) and serialize into json string, 
but there can be some slight difference like
`2024-01-28T12:00:00+00:00` with `2024-01-28T12:00:00Z`. 

I guess it will be better to explicitely transfer with method like isoformat(). 
as both of the share same class `datetime`. 


Thanks for letting me know!!

kwonkwonn: Modifiied response field as string, 
And it will be converted from API adapter in following PR(to adapter migration)
(e.g.)
```        self, result: UpdatePasswordNoAuthActionResult
    ) -> UpdatePasswordNoAuthResponse:
        """Convert UpdatePasswordNoAuthActionResult to UpdatePasswordNoAuthResponse."""
        return UpdatePasswordNoAuthResponse(
            password_changed_at=result.password_changed_at.isoformat(), // example
        )
```   
thank you

src/ai/backend/manager/api/auth.py
HyeockJinKim, line 1: Have you checked if moving the directory might cause issues? Also, it doesn't seem necessary to change the existing handlers this time. I'd like to clarify the current work direction, and I think declaring it deprecated is too hasty a decision. To mark it deprecated, we need to have a new alternative API already in place to guide users.

kwonkwonn:
```
def _init_subapp(
    pkg_name: str,
    root_app: web.Application,
    subapp: web.Application,
    global_middlewares: Iterable[Middleware],
) -> None:
    subapp.on_response_prepare.append(on_prepare)

    async def _set_root_ctx(subapp: web.Application):
        # Allow subapp's access to the root app properties.
        # These are the public APIs exposed to plugins as well.
        subapp["_root.context"] = root_app["_root.context"]
        subapp["_root_app"] = root_app

    # We must copy the public interface prior to all user-defined startup signal handlers.
    subapp.on_startup.insert(0, _set_root_ctx)
    if "prefix" not in subapp:
        subapp["prefix"] = pkg_name.split(".")[-1].replace("_", "-")
    prefix = subapp["prefix"]
    root_app.add_subapp("/" + prefix, subapp)
    root_app.middlewares.extend(global_middlewares)
```

when i had a investigate before migrating, 
I thought it could be alright if we retain sub-servers name the same. 

But i understand the concern, and it seems subapp would detect ./auth.py faster than /auth/__init__.py
it will be much more safer to migrate into legacy with handler/adapter pattern fully assured.

thanks you!

src/ai/backend/manager/api/auth_migrate/auth.py
```diff
@@ -0,0 +1,1174 @@
+"""
+Legacy auth handlers (function-based).
+
+DEPRECATED: This module is preserved for regression testing only.
+Use ai.backend.manager.api.auth (class-based handlers) for production.
+"""
+
+from __future__ import annotations
+
```
kwonkwonn: Temporary file for migrating into pydantic class from following PR.

src/ai/backend/manager/api/auth_migrate/auth.py
```diff
@@ -0,0 +1,1171 @@
...
+    "YAKST": 10 * 3600,
+    "YAKT": 9 * 3600,
+    "YAPT": 10 * 3600,
+    "YEKST": 6 * 3600,
+    "YEKT": 5 * 3600,
+    "Z": 0 * 3600,
+}
+whois_timezone_info: Final[Mapping[str, int]] = {k: int(v) for k, v in _whois_timezone_info.items()}
+
+
```
kwonkwonn: As there are following PR for defining types for credential information, 
I've just marked it as tuple.

src/ai/backend/manager/api/auth_migrate/auth.py
HyeockJinKim: Is this auth_migrate directory really necessary? It seems like all we need is to add the handler method. (If verified, replace the handler method)

HyeockJinKim (changes requested):

src/ai/backend/manager/api/auth_migrate/auth.py
kwonkwonn: It that mean putting handler into same `auth.py` or adding new file like `auth_handler`?

If we are going to seperate utils and methods function into somewhere else It seems ok, but i was worried about having 1000~ lines in single `auth.py`.

kwonkwonn: But as we decided to remain auth.py the same, auth_migrate seem unecessary for this pr anyway.

i will remove it and re-post pr shortly.
thank you

src/ai/backend/common/dto/manager/auth/request.py
```diff
@@ -0,0 +1,105 @@
...
+    type: AuthTokenType = Field(description="Authentication type (keypair or jwt)")
+    domain: str = Field(description="Domain name")
+    username: str = Field(description="Username or email")
+    password: str = Field(description="Password")
+    stoken: Optional[str] = Field(
+        default=None,
+        description="Session token",
+        validation_alias=AliasChoices("stoken", "sToken"),
+    )
+
```
HyeockJinKim: Could you add more explanation to the descriptions? It's unclear what purpose these values serve.

PR #8406, opened 2026-01-29
fix(BA-4136): Prevent kernel status regression from TERMINATED to TERMINATING

### LINKED ISSUE #8405: Kernel status regresses from TERMINATED to TERMINATING in registry.destroy_session()
## Summary

  The kernel-level match statement in registry.destroy_session() lacks explicit cases for KernelStatus.TERMINATED and KernelStatus.CANCELLED, causing already-terminal kernels to fall into case _: and regress to TERMINATING via a raw SQL UPDATE with no status guard.

### Root Cause:

  In src/ai/backend/manager/registry.py, the destroy_session() method (~line 2597) handles kernel statuses as follows:

- PENDING | PULLING → CANCELLED
- SCHEDULED | PREPARING | PREPARED | CREATING | TERMINATING | ERROR → TERMINATED (forced only)
- case _: → TERMINATING (raw SQL UPDATE with no status guard in WHERE clause)

  There is no explicit case for TERMINATED or CANCELLED. Kernels already in these terminal states fall into case _:, which unconditionally sets them back to TERMINATING.

### Reproduction Scenario (Livelock):

1. An inference session (model serving) triggers ModelServiceStatusBroadcastEvent
1. model_serving.py calls registry.destroy_session(forced=True)
1. The kernel is already TERMINATED → falls into case _: → regressed to TERMINATING
1. Since the kernel is back in TERMINATING, the session cannot transition to TERMINATED
1. The event loop in model serving waits indefinitely for SessionTerminatedBroadcastEvent
1. Session is permanently stuck

### Impact:

- KERNEL_STATUS_TRANSITION_MAP defines TERMINATED: set() (no valid outgoing transitions), but raw SQL bypasses this check entirely
- The sokovan scheduler's safe state transition guards (terminatable_statuses() WHERE clauses) are defeated by this legacy code path
- Inference sessions using model serving can become permanently stuck with no recovery

JIRA Issue: BA-4136

fregataa:
resolves #8405 (BA-4136)
## Summary

  - `registry.destroy_session()` lacked explicit match cases for `KernelStatus.TERMINATED` and `KernelStatus.CANCELLED`, causing already-terminal kernels to fall into `case _:` and regress to TERMINATING via raw SQL with no status guard
  - This created a livelock in model serving: `destroy_session(forced=True)` repeatedly regressed TERMINATED kernels back to TERMINATING, preventing the session from ever reaching TERMINATED state and permanently stalling the event loop

  ## Changes

  - Added explicit `case KernelStatus.TERMINATED | KernelStatus.CANCELLED:` that logs and skips kernels already in terminal states, preventing the `case _:` fallthrough

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/changes/8406.fix.md
+++ b/changes/8406.fix.md
@@ -1,0 +1,1 @@
+Prevent kernel status regression from TERMINATED to TERMINATING in `destroy_session()` which caused a livelock in model serving sessions


jopemachine (approved):

jopemachine:
Since this is a bug-fix PR, could we add test cases that cover this fix?

HyeockJinKim:
This section is code to be removed, so I'll skip testing for now. However, please verify how far backporting is needed for the milestone and apply it once.

PR #8455, opened 2026-01-30
refactor(BA-4080): add DB Source Pattern in User repository

MinJuTur:
## Summary
- Extract all DB session-opening logic from `UserRepository` into `UserDBSource`, following the established `db_source` pattern used across other repository domains
- `UserRepository` now delegates pure DB methods to `UserDBSource` while retaining resilience decorators and orchestration logic
- Mixed methods (`delete_vfolders`, `delete_keypairs_with_valkey`, `_get_time_binned_monthly_stats`) are restructured so that DB access is handled by `UserDBSource` with no behavioral changes

## Related Issue
- Jira: BA-4080

## Test plan
- [x] `pants check` passes on all modified files
- [x] `pants lint` passes on all modified files
- [x] Existing unit tests pass (`test_user_repository.py`)

### DIFF
--- a/src/ai/backend/manager/repositories/user/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/user/db_source/db_source.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 from collections.abc import Sequence
 from datetime import datetime, timedelta
-from typing import Any, Optional, cast
+from typing import Any, cast
 from uuid import UUID, uuid4
 
 import aiotools
@@ -118,7 +118,7 @@ class UserDBSource:
             return UserData.from_row(user_row)
 
     async def create_user_validated(
-        self, creator: Creator[UserRow], group_ids: Optional[list[str]]
+        self, creator: Creator[UserRow], group_ids: list[str] | None
     ) -> UserCreateResultData:
         """
         Create a new user with default keypair and group associations.
@@ -358,7 +358,7 @@ class UserDBSource:
 
     async def get_kernel_rows_for_monthly_stats(
         self,
-        user_uuid: Optional[UUID],
+        user_uuid: UUID | None,
     ) -> Sequence[Row[Any]]:
         """Fetch kernel rows for time-binned monthly stats."""
         now = datetime.now(tzutc())
@@ -405,7 +405,7 @@ class UserDBSource:
                     )
                 ),
             )
-            rows = cast(list[VFolderRow], result.fetchall())
+            rows = result.fetchall()
             for vf in rows:
                 target_vfs.append(
                     VFolderDeletionInfo(VFolderID.from_row(vf), vf.host, vf.unmanaged_path)
@@ -472,7 +472,6 @@ class UserDBSource:
             sa.or_(UserRow.email == email, UserRow.username == username)
         )
         result = await session.scalar(query)
-        result = cast(Optional[UUID], result)
         return result is not None
 
     async def _check_username_exists_for_other_user(
@@ -483,7 +482,6 @@ class UserDBSource:
             sa.and_(UserRow.username == username, UserRow.email != exclude_email)
         )
         result = await conn.scalar(query)
-        result = cast(Optional[UUID], result)
         return result is not None
 
     async def _get_user_by_email(self, session: SASession, email: str) -> UserRow:
@@ -506,7 +504,7 @@ class UserDBSource:
         res = result.first()
         if res is None:
             raise UserNotFound(f"User with email {email} not found.")
-        return res
+        return cast(UserRow, res)
 
     async def _add_user_to_groups(
         self, db_session: SASession, user_uuid: UUID, domain_name: str, group_ids: list[str]
@@ -655,7 +653,7 @@ class UserDBSource:
         row = result.first()
         if not row:
             raise UserNotFound()
-        return row.uuid
+        return cast(UUID, row.uuid)
 
     async def _user_vfolder_mounted_to_active_kernels(
         self,
--- a/src/ai/backend/manager/repositories/user/repository.py
+++ b/src/ai/backend/manager/repositories/user/repository.py
@@ -4,7 +4,7 @@ import logging
 from collections.abc import Mapping
 from datetime import datetime, timedelta
 from decimal import Decimal
-from typing import Any, Optional
+from typing import Any
 from uuid import UUID
 
 import msgpack
@@ -161,7 +161,7 @@ class UserRepository:
     async def get_admin_time_binned_monthly_stats(
         self,
         valkey_stat_client: ValkeyStatClient,
-    ) -> list[dict]:
+    ) -> list[dict[str, Any]]:
         """Get time-binned monthly statistics for all users."""
         return await self._get_time_binned_monthly_stats(None, valkey_stat_client)
 


src/ai/backend/manager/repositories/user/db_source/db_source.py
```diff
@@ -0,0 +1,798 @@
...
+
+        storage_ptask_group = aiotools.PersistentTaskGroup()
+        try:
+            await initiate_vfolder_deletion(
+                self._db,
+                target_vfs,
+                storage_manager,
+                storage_ptask_group,
+            )
+        except VFolderOperationFailed:
```
MinJuTur: This `try/except` block catches `VFolderOperationFailed` only to re-raise it, which is functionally identical to not having the `try/except` at all. This was carried over from the original code as-is. It may have been intentional to explicitly document that this exception can occur, but it could also be simplified by removing the `try/except` block entirely.

src/ai/backend/manager/repositories/user/db_source/db_source.py
```diff
@@ -0,0 +1,798 @@
...
+
+    async def _get_user_by_email_with_conn(self, conn: AsyncConnection, email: str) -> UserRow:
+        """Private method to get user by email using connection."""
+        result = await conn.execute(sa.select(users).where(users.c.email == email))
+        res = result.first()
+        if res is None:
+            raise UserNotFound(f"User with email {email} not found.")
+        return res
+
+    def _validate_user_access(self, _user_row: UserRow, _requester_uuid: Optional[UUID]) -> bool:
```
MinJuTur: `_validate_user_access` is dead code — it always returns `True` and is not called anywhere in production code. It appears to be a placeholder for future ownership validation logic. This was also carried over from the original code.

src/ai/backend/manager/repositories/user/db_source/db_source.py
```diff
@@ -0,0 +1,798 @@
...
+                    & (kernels.c.status.in_(RESOURCE_USAGE_KERNEL_STATUSES)),
+                )
+                .order_by(sa.asc(kernels.c.created_at))
+            )
+            if user_uuid is not None:
+                query = query.where(kernels.c.user_uuid == user_uuid)
+            result = await conn.execute(query)
+            return result.fetchall()
+
+    async def delete_vfolders(
```
MinJuTur, line 387: This method takes `StorageSessionManager` as a parameter because the original `_delete_vfolders` performed both DB operations and `initiate_vfolder_deletion` within a single flow. To avoid changing the runtime behavior, the entire logic was moved to `db_source` as-is rather than splitting it.

src/ai/backend/manager/repositories/user/db_source/db_source.py
```diff
@@ -0,0 +1,798 @@
...
+                target_vfs,
+                storage_manager,
+                storage_ptask_group,
+            )
+        except VFolderOperationFailed:
+            raise
+
+        return len(target_vfs)
+
+    async def delete_keypairs_with_valkey(
```
MinJuTur, line 424: Similarly, this method takes `ValkeyStatClient` as a parameter because the original `_delete_keypairs_with_valkey` performed Valkey cleanup and DB deletion within a single transaction. Splitting them would change the transaction boundary — if Valkey cleanup fails, the DB transaction should roll back, which requires both operations to be in the same `begin()` scope.

src/ai/backend/manager/repositories/user/repository.py
```diff
@@ -533,39 +157,47 @@ async def get_user_time_binned_monthly_stats(
...
-                    & (kernels.c.status.in_(RESOURCE_USAGE_KERNEL_STATUSES)),
-                )
-                .order_by(sa.asc(kernels.c.created_at))
-            )
-            if user_uuid is not None:
-                query = query.where(kernels.c.user_uuid == user_uuid)
-            result = await conn.execute(query)
-            rows = result.fetchall()
+        # DB query via db_source
+        rows = await self._db_source.get_kernel_rows_for_monthly_stats(user_uuid)
```
MinJuTur, line 200: Only the DB query part was extracted to `db_source`

tests/unit/manager/repositories/user/test_user_repository.py
```diff
@@ -510,7 +510,7 @@ def test_validate_user_access_owner(
             role=UserRole.USER,
             resource_policy="default",
         )
-        result = user_repository._validate_user_access(user_row, user_uuid)
+        result = user_repository._db_source._validate_user_access(user_row, user_uuid)
```
jopemachine: Don't access private field `_db_source` (and private method)

src/ai/backend/manager/repositories/user/db_source/db_source.py
jopemachine: Let's remove this

src/ai/backend/manager/repositories/user/db_source/db_source.py
jopemachine: Let's remove this

tests/unit/manager/repositories/user/test_user_repository.py
MinJuTur: > Don't access private field `_db_source` (and private method)

`_validate_user_access` is dead code — it always returns `True` and is not called anywhere in production code. It was a placeholder carried over from the original `UserRepository`. How about removing both `_validate_user_access` and these 3 test cases? @jopemachine

src/ai/backend/manager/repositories/user/db_source/db_source.py
```diff
@@ -0,0 +1,798 @@
...
+                sa.delete(keypairs).where(keypairs.c.user == user_uuid),
+            )
+            return result.rowcount
+
+    async def _check_domain_exists(
+        self, session: SASession | AsyncConnection, domain_name: str
+    ) -> bool:
+        query = sa.select(DomainRow.name).where(DomainRow.name == domain_name)
+        result = await session.scalar(query)
+        result = cast(Optional[str], result)
```
jopemachine: I think we don't need this casting anymore

tests/unit/manager/repositories/user/test_user_repository.py
jopemachine: Let's remove them

src/ai/backend/manager/repositories/user/db_source/db_source.py
```diff
@@ -0,0 +1,788 @@
...
+        return result is not None
+
+    async def _check_user_exists_with_email_or_username(
+        self, session: SASession, *, email: str, username: str
+    ) -> bool:
+        query = sa.select(UserRow.uuid).where(
+            sa.or_(UserRow.email == email, UserRow.username == username)
+        )
+        result = await session.scalar(query)
+        result = cast(Optional[UUID], result)
```
jopemachine: Please check all `cast` are required. I think we don't need these casts

src/ai/backend/manager/repositories/user/db_source/db_source.py
```diff
@@ -0,0 +1,786 @@
...
+    async def purge_user(self, email: str) -> None:
+        """Completely purge user and all associated data."""
+        async with self._db.begin_session() as session:
+            user_uuid = await self._get_user_uuid_by_email(session, email)
+
+            # Delete all user data in proper order using purger pattern
+            await execute_batch_purger(session, create_user_error_log_purger(user_uuid))
+            await execute_batch_purger(session, create_user_keypair_purger(user_uuid))
+            await execute_batch_purger(session, create_user_vfolder_permission_purger(user_uuid))
+            await execute_batch_purger(session, create_user_group_association_purger(user_uuid))
```
MinJuTur, line 301: Resolved merge conflict with `purge_user` from main, integrated into db_source pattern

jopemachine:
Overall, there seems to be a lot of code that needs improvement, but it doesn’t seem appropriate to modify the existing code here.

PR #8472, opened 2026-01-30
fix(BA-4167): Use `joinedload` instead of selectinload in `AgentDBSource.get_by_id`

### LINKED ISSUE #8468: AgentDBSource.get_by_id() raises a MissingGreenlet error
## Description                                                                                                         

`AgentDBSource.get_by_id()` raises a `MissingGreenlet` error when accessing the kernels relationship despite using selectinload.                                                                                                       

  

## Root Cause                                                                                                          

scalar() returns a single value immediately, but selectinload requires a separate SELECT ... WHERE id IN (...) query to be executed after the main query. When using scalar(), this second query is not properly triggered. As a result, when self.kernels is accessed later, lazy loading is attempted, which is not supported in async contexts

JIRA Issue: BA-4167

seedspirit:
resolves #8468 (BA-4167)

- [x] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/tests/unit/manager/repositories/agent/test_repository.py
+++ b/tests/unit/manager/repositories/agent/test_repository.py
@@ -62,7 +62,6 @@ from ai.backend.manager.models.session import SessionRow
 from ai.backend.manager.models.user import UserRow
 from ai.backend.manager.models.utils import ExtendedAsyncSAEngine
 from ai.backend.manager.models.vfolder import VFolderRow
-from ai.backend.manager.repositories.agent.db_source.db_source import AgentDBSource
 from ai.backend.manager.repositories.agent.repository import AgentRepository
 from ai.backend.manager.repositories.base.pagination import OffsetPagination
 from ai.backend.manager.repositories.base.querier import BatchQuerier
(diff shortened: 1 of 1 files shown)


tests/unit/manager/repositories/agent/test_repository.py
```diff
@@ -904,3 +907,21 @@ async def test_search_agents_validates_actual_occupied_slots(
...
+        self,
+        db_source: AgentDBSource,
+        agent_with_kernels: KernelFilteringTestCase,
+    ) -> None:
+        # Act - This should not raise MissingGreenlet error
+        agent_data = await db_source.get_by_id(agent_with_kernels.agent_id)
+
+        # Assert - Verify actual_occupied_slots is computed correctly
+        actual_cpu = agent_data.actual_occupied_slots.get("cpu", 0)
+        assert Decimal(str(actual_cpu)) == agent_with_kernels.expected_actual_occupied_cpu
```
HyeockJinKim, line 931: We need to test the repository, but you are only testing the database source.

seedspirit: Modified it to test the repository. Since `test_search_agents_validates_actual_occupied_slots` was also testing DBSource, changed it to test the repository as well.

PR #8501, opened 2026-02-02
docs: Add BEP-1044 for fixing multi agent device allocation

hhoikoo:

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/proposals/BEP-1044-multi-agent-device-split.md
+++ b/proposals/BEP-1044-multi-agent-device-split.md
@@ -14,91 +14,136 @@ Implemented-Version:
 - JIRA: BA-4143, BA-4144, BA-4145, BA-4146
 - GitHub: #8433, #8440, #8447, #8463
 
+## Background: Multi-Agent Resource Allocation Modes
+
+Backend.AI supports running multiple agent processes on a single host, each managing a subset of the machine's compute resources. This is useful for isolating workloads, running different container runtimes, or partitioning resources between different user groups.
+
+The system provides three **allocation modes** that determine how resources are divided among agents:
+
+### SHARED Mode
+
+All agents see and can allocate from the full set of available resources. There is no partitioning - agents compete for resources on a first-come-first-served basis. This is the simplest mode and is suitable when agents handle non-overlapping workloads or when resource contention is managed at a higher level.
+
+### AUTO_SPLIT Mode
+
+The system automatically divides resources among agents. Each agent receives an exclusive subset of devices (for discrete resources like CPU cores and GPUs) or a proportional share (for shared resources like memory). This mode requires no manual configuration beyond specifying the number of agents.
+
+### MANUAL Mode
+
+Administrators explicitly configure which resources each agent can use. This provides fine-grained control over resource assignment, allowing specific devices to be assigned to specific agents based on workload requirements, hardware topology, or organizational policies.
+
 ## Motivation
 
-As part of the Accelerator Interface v2 redesign ([BEP-1016](BEP-1016-accelerator-interface-v2.md)), we are moving toward device-centric resource management. The current slot-based allocation model is an abstraction mismatch that complicates multi-agent resource partitioning.
+As part of the Accelerator Interface v2 redesign ([BEP-1016](BEP-1016-accelerator-interface-v2.md)), we are moving toward device-centric resource management.
 
-The previous multi-agent resource allocation used **slot-based** configuration where devices were specified by slot names with decimal fractional values:
+The previous multi-agent resource allocation was incorrectly implemented using a **slot-based** configuration where devices were specified by slot names with decimal fractional values:
 
 ```toml
 [agent.resource.allocations]
 devices = { "cuda.mem" = 0.3, "cuda.shares" = 0.5 }
 ```
 
-This approach has problems:
+This approach has fundamental problems:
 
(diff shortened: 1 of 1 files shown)


proposals/BEP-1043-multi-agent-device-split.md
```diff
@@ -0,0 +1,242 @@
...
+Target-Version:
+Implemented-Version:
+---
+
+# Multi-Agent Device Split: Slot-Based to Device-Based Allocation
+
+## Related Issues
+
+- JIRA: BA-XXXX
+- GitHub: #XXXX
```
HyeockJinKim: Please add it together.

proposals/BEP-1043-multi-agent-device-split.md
```diff
@@ -0,0 +1,242 @@
...
+```python
+def _calculate_device_slot(self, slot_name, agent_idx, agent_config, alloc_map_type):
+    match agent_config.resource.allocation_mode:
+        case ResourceAllocationMode.SHARED:
+            return self._calculate_device_slot_shared(slot_name)
+        case ResourceAllocationMode.AUTO_SPLIT:
+            return self._calculate_device_slot_auto_split(slot_name, alloc_map_type, agent_idx)
+        case ResourceAllocationMode.MANUAL:
+            return self._calculate_device_slot_manual(slot_name, agent_config)
+```
```
HyeockJinKim: Could you briefly explain what each existing mode does? It's difficult to understand without prior knowledge of the current implementation.

proposals/BEP-1044-multi-agent-device-split.md
```diff
@@ -0,0 +1,242 @@
...
+```toml
+[agent.resource.allocations]
+devices = { "cuda.mem" = 0.3, "cuda.shares" = 0.5 }
+```
+
+This approach has problems:
+
+1. **No device exclusivity**: Slot-based allocation divides abstract quantities (e.g., "0.5 of cuda.shares") but cannot enforce that agents use mutually exclusive physical devices. For CPU cores and accelerators, multiple agents must not share the same device - slot amounts alone cannot express or enforce this.
+
+2. **Abstraction mismatch**: The actual resource being partitioned is a set of physical devices (cuda0, cuda1, etc.), not slot amounts. Slot names like `cuda.mem` and `cuda.shares` are derived properties - we should configure at the device level and derive slots from that, not the other way around.
```
HyeockJinKim, line 50: That's technically correct, but I suspect this implementation arose because of an existing use case. Do you happen to know anything about that?

proposals/BEP-1043-multi-agent-device-split.md
```diff
@@ -0,0 +1,242 @@
...
+class DevicePartitioner(Protocol):
+    """Protocol for generating device assignments in AUTO_SPLIT mode."""
+    device_name: DeviceName
+
+    def generate_assignments(
+        self,
+        devices: Sequence[AbstractComputeDevice],
+        agent_ids: Sequence[AgentId],
+    ) -> Mapping[AgentId, Sequence[DeviceId]]: ...
+```
```
HyeockJinKim: I don't have a view of the whole picture right now, so it's hard to understand. This BEP appears as a list of implementations rather than feeling organically connected. Could you add context about this implementations?

proposals/BEP-1044-multi-agent-device-split.md
```diff
@@ -0,0 +1,242 @@
...
+
+```toml
+# Before (slot-based)
+[agent.resource.allocations]
+devices = { "cuda.mem" = 0.5, "cuda.shares" = 0.5 }
+
+# After (device-based)
+[agent.resource.allocations.devices]
+cuda = ["cuda0", "cuda1"]
+```
```
HyeockJinKim, line 235: Users may still request the existing (user-friendly) allocation of 0.5, etc. In that case, please specify how the settings should be configured.

proposals/BEP-1043-multi-agent-device-split.md
hhoikoo: Good point - forgot to add this. Done

proposals/BEP-1044-multi-agent-device-split.md
hhoikoo: No there is no good reason why this was done. This was simply a wrong implementation of the multi agent feature, which was done because of my misunderstanding on how the multi agent resource splitting was supposed to be implemented. There is no need for backwards compatibility here, especially considering how no production (or even dev system) is using the multi agent feature

proposals/BEP-1044-multi-agent-device-split.md
hhoikoo: As mentioned above, no current production system uses the multi agent feature. I don't think we need to be backwards compatible here, but I could be wrong. Could you please tell me?

proposals/BEP-1043-multi-agent-device-split.md
hhoikoo: I rewrote quite a bit of the document to address this feedback

proposals/BEP-1043-multi-agent-device-split.md
hhoikoo: I rewrote quite a bit of the document to address this feedback

PR #8511, opened 2026-02-02
fix(BA-4191): Resolve all images when commiting sessions

### LINKED ISSUE #8512: Cannot commit image from session if the base image deleted
Users cannot commit images from a session if the base image of the session has been deleted

JIRA Issue: BA-4191

fregataa:
resolves #8512 (BA-4191)

## Summary
  - Fix session commit failing when the base image has been deleted from the registry
  - SessionRepository.resolve_image() now accepts an alive_only parameter; when set to False, it resolves images with both ALIVE and DELETED statuses
  - The session commit flow passes alive_only=False so that sessions based on deleted images can still be committed successfully
  - Add unit tests for ImageRow.resolve() covering identifier/alias/ref resolution, status filtering, candidate priority, and error cases

##  Problem
  When a user attempted to commit a session whose base image had already been marked as DELETED, the operation failed because ImageRow.resolve() only looked for ALIVE images by default. This made it impossible to commit any session running on a deleted image.

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/repositories/session/repository.py
+++ b/src/ai/backend/manager/repositories/session/repository.py
@@ -231,9 +231,26 @@ class SessionRepository:
     async def resolve_image(
         self,
         image_identifiers: list[ImageAlias | ImageRef | ImageIdentifier],
+        alive_only: bool = True,
     ) -> ImageRow:
+        """Resolve an image from the given identifiers.
+
+        When ``alive_only`` is True (default), only images with the ALIVE status
+        are considered.  Set it to False to also include DELETED images, which is
+        useful when the caller needs to reference images that are no longer active
+        (e.g., committing a session whose base image has been deleted).
+        """
         async with self._db.begin_readonly_session() as db_sess:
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/repositories/session/repository.py
```diff
@@ -231,9 +231,26 @@ async def get_container_registry(
...
+            if alive_only:
+                return await ImageRow.resolve(db_sess, image_identifiers)
+            return await ImageRow.resolve(
+                db_sess,
+                image_identifiers,
+                filter_by_statuses=[
+                    ImageStatus.ALIVE,
+                    ImageStatus.DELETED,
+                ],
+            )
```
HyeockJinKim, line 253: This option doesn't seem to fit well, but for now, we'll proceed based on the Fix standard. Subsequently, it should be able to operate based on search.

HyeockJinKim (approved):

PR #8514, opened 2026-02-02
chore: Add service.instance.id to OpenTelemetry

Yaminyam:
Introduce an optional service_instance_id on OpenTelemetrySpec and include it as the service.instance.id attribute when present. Propagate service_instance_id from various servers (agent, appproxy coordinator/worker, manager, storage) using meta.display_name, and set a hostname-based instance id for the web server (webserver-{hostname}). This makes telemetry resources identify individual service instances for easier tracing and debugging.

### DIFF
--- a/changes/8502.enhance.md
+++ b/changes/8502.enhance.md
@@ -1,0 +1,1 @@
+Introduce the `db_source` pattern to `GroupRepository`, extracting all DB session-opening logic into `GroupDBSource` for improved separation of concerns
--- a/changes/8513.enhance.md
+++ b/changes/8513.enhance.md
@@ -1,0 +1,1 @@
+migrate unnecessary service related function from api to service 
--- a/changes/8525.enhance.md
+++ b/changes/8525.enhance.md
@@ -1,0 +1,1 @@
+Apply the Purger pattern to `DomainRepository` for domain-related deletions, improving code consistency and maintainability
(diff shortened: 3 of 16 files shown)


src/ai/backend/logging/otel.py
```diff
@@ -23,13 +23,17 @@ class OpenTelemetrySpec:
...
     def to_resource(self) -> Resource:
-        return Resource.create({
+        attributes = {
             "service.name": self.service_name,
             "service.id": str(self.service_id),
             "service.version": self.service_version,
-        })
+        }
+        if self.service_instance_id:
+            attributes["service.instance.id"] = self.service_instance_id
```
HyeockJinKim: It always seems to get added, so there's no reason to make it optional.

src/ai/backend/logging/otel.py
```diff
@@ -18,18 +18,26 @@
...
     log_level: str
     endpoint: str
+    service_instance_name: str
 
     def to_resource(self) -> Resource:
-        return Resource.create({
+        # Generate stable UUID from instance name using UUID v5
+        # as recommended by OpenTelemetry specification
+        # Official OTEL namespace UUID for service.instance.id generation
+        otel_namespace_uuid = uuid.UUID("4d63009a-8d0f-11ee-aad7-4c796ed8e320")
```
HyeockJinKim: Instead of inserting an arbitrary ID, wouldn't it be better to keep the received service_id as-is rather than clearing it? This value is generated at each service startup point, and it would be better to use that value as the ID.

HyeockJinKim (changes requested):

PR #8525, opened 2026-02-02
refactor(BA-3764): add Purger pattern to Domain repository

MinJuTur:
## Summary
- Introduce the `BatchPurger` pattern in `DomainRepository.purge_domain()` for domain-related deletions
- Add `DomainKernelBatchPurgerSpec` and `DomainBatchPurgerSpec` for batch deletion of kernels and domains
- Migrate `purge_domain()` and related helper methods from `SAConnection` (Core) to `SASession` (ORM)
- Add integration tests for domain purger specs

## Related Issue
- Jira: BA-3764

## Test plan
- [x] `pants lint src/ai/backend/manager/repositories/domain::`
- [x] `pants check src/ai/backend/manager/repositories/domain::`
- [x] `pants test tests/unit/manager/repositories/domain::`

[BA-3764]: https://lablup.atlassian.net/browse/BA-3764?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/unit/manager/repositories/domain/test_domain_purgers.py
+++ b/tests/unit/manager/repositories/domain/test_domain_purgers.py
@@ -12,7 +12,7 @@ from typing import TYPE_CHECKING
 import pytest
 import sqlalchemy as sa
 
-from ai.backend.common.types import ResourceSlot
+from ai.backend.common.types import ResourceSlot, VFolderHostPermissionMap
 from ai.backend.manager.data.auth.hash import PasswordHashAlgorithm
 from ai.backend.manager.data.kernel.types import KernelStatus
 from ai.backend.manager.models.agent import AgentRow
@@ -75,8 +75,8 @@ class TestDomainPurgersIntegration:
                 name=domain_name,
                 description=f"Test domain {domain_name}",
                 is_active=True,
-                total_resource_slots={},
-                allowed_vfolder_hosts={},
+                total_resource_slots=ResourceSlot({}),
+                allowed_vfolder_hosts=VFolderHostPermissionMap({}),
                 allowed_docker_registries=[],
                 dotfiles=b"",
                 integration_id=None,
@@ -164,8 +164,8 @@ class TestDomainPurgersIntegration:
                 description="Test group for integration tests",
                 is_active=True,
                 domain_name=sample_domain,
-                total_resource_slots={},
-                allowed_vfolder_hosts={},
+                total_resource_slots=ResourceSlot({}),
+                allowed_vfolder_hosts=VFolderHostPermissionMap({}),
                 dotfiles=b"\x90",
                 resource_policy=project_resource_policy,
                 type=ProjectType.GENERAL,


src/ai/backend/manager/repositories/domain/repository.py
MinJuTur, line 114: `soft_delete_domain()` still uses `SAConnection`. Only `purge_domain()` and its helper methods were migrated to `SASession` in this PR.

tests/unit/manager/repositories/domain/test_domain_purgers.py
```diff
@@ -0,0 +1,299 @@
...
+    DomainBatchPurgerSpec,
+    DomainKernelBatchPurgerSpec,
+)
+from ai.backend.testutils.db import with_tables
+
+if TYPE_CHECKING:
+    from ai.backend.manager.models.utils import ExtendedAsyncSAEngine
+
+
+class TestDomainPurgersIntegration:
```
MinJuTur, line 43: The initial plan included `keypair_resource_policy`, `sample_keypair`, and `sample_scaling_group` fixtures, but they turned out to be unnecessary for these purger spec tests. So I excluded them.

src/ai/backend/manager/repositories/domain/purgers.py
```diff
@@ -0,0 +1,32 @@
...
+
+import sqlalchemy as sa
+
+from ai.backend.manager.models.domain import DomainRow
+from ai.backend.manager.models.kernel.row import KernelRow
+from ai.backend.manager.repositories.base.purger import BatchPurgerSpec
+
+
+@dataclass
+class DomainKernelBatchPurgerSpec(BatchPurgerSpec[KernelRow]):
```
jopemachine, line 14: We previously introduced SessionByIdBatchPurgerSpec.
Then this type should be named KernelByDomainBatchPurgerSpec.
Type naming seems inconsistent.

Also, shouldn’t this purger go under the kernel side?

tests/unit/manager/repositories/domain/test_domain_purgers.py
```diff
@@ -0,0 +1,297 @@
...
+from typing import TYPE_CHECKING
+
+import pytest
+import sqlalchemy as sa
+
+from ai.backend.common.types import ResourceSlot
+from ai.backend.manager.data.auth.hash import PasswordHashAlgorithm
+from ai.backend.manager.data.kernel.types import KernelStatus
+
+# Import Row classes to ensure SQLAlchemy mapper initialization
```
jopemachine: Please remove this

src/ai/backend/manager/repositories/domain/purgers.py
MinJuTur: > We previously introduced SessionByIdBatchPurgerSpec.
> Then this type should be named KernelByDomainBatchPurgerSpec.
> Type naming seems inconsistent.

Regarding the naming: the existing codebase predominantly uses the `{OwnerDomain}{Target}BatchPurgerSpec` pattern — e.g., `GroupKernelBatchPurgerSpec`, `GroupSessionBatchPurgerSpec`, `UserKeyPairBatchPurgerSpec`, `UserErrorLogBatchPurgerSpec`, etc. `SessionByIdsBatchPurgerSpec` appears to be the only exception. So `DomainKernelBatchPurgerSpec` follows the majority convention.


> Also, shouldn’t this purger go under the kernel side?

Regarding the location: currently all purger specs are co-located with the repository that triggers the purge (e.g., `GroupKernelBatchPurgerSpec` lives in `group/purgers.py`). Should we move them to the target side instead?

jopemachine: > Regarding the naming: the existing codebase predominantly uses the {OwnerDomain}{Target}BatchPurgerSpec pattern — e.g., GroupKernelBatchPurgerSpec, GroupSessionBatchPurgerSpec, UserKeyPairBatchPurgerSpec, UserErrorLogBatchPurgerSpec, etc. SessionByIdsBatchPurgerSpec appears to be the only exception. So DomainKernelBatchPurgerSpec follows the majority convention.

If SessionByIdsBatchPurgerSpec is the only exception, then its name probably shouldn’t have been chosen that way in the first place. That said, once a clear convention is defined, we can rename everything consistently at that time.

> Regarding the location: currently all purger specs are co-located with the repository that triggers the purge (e.g., GroupKernelBatchPurgerSpec lives in group/purgers.py). Should we move them to the target side instead?

This should likewise be kept as it was originally implemented.
Once a convention is defined, we can make a batch change if needed.

MinJuTur: Understood. I'll keep the current naming and location as-is for now, and we can align them in a batch once the convention is formally defined.

tests/unit/manager/repositories/domain/test_domain_purgers.py
```diff
@@ -0,0 +1,295 @@
...
+        # Verify kernels are deleted
+        async with db_with_cleanup.begin_session() as session:
+            count = await session.scalar(
+                sa.select(sa.func.count())
+                .select_from(KernelRow)
+                .where(KernelRow.domain_name == domain_name)
+            )
+            assert count == 0
+
+    @pytest.mark.asyncio
```
jopemachine: Useless decorator

tests/unit/manager/repositories/domain/test_domain_purgers.py
```diff
@@ -0,0 +1,295 @@
...
+    DomainBatchPurgerSpec,
+    DomainKernelBatchPurgerSpec,
+)
+from ai.backend.testutils.db import with_tables
+
+if TYPE_CHECKING:
+    from ai.backend.manager.models.utils import ExtendedAsyncSAEngine
+
+
+class TestDomainPurgersIntegration:
```
jopemachine, line 43: Please rename this `TestDomainPurgers`.

MinJuTur: > Please rename this `TestDomainPurgers`.

Other purger test files also use the `Integration` suffix (e.g., `TestGroupPurgersIntegration`, `TestUserPurgersIntegration`, `TestVFolderPurgersIntegration`). Should I drop it only in this file, or keep it consistent with the existing tests?

jopemachine: Tests for repositories written against a real database are all integration tests.
The term “integration” is clearly redundant.
But since this naming has already been used in the existing code, I’m fine with leaving it as is for now.

PR #8534, opened 2026-02-03
refactor(BA-4079): add DB Source Pattern in Session repository

MinJuTur:
## Summary
- Introduce the `db_source` pattern to `SessionRepository`, extracting all DB session-opening logic into `SessionDBSource` for improved separation of concerns
- Move `rescan_images` method from `SessionRepository` to `ImageRepository` as it is image-related functionality
- Remove dead code: `get_session_with_eager_loading`, `find_dependent_sessions`
- Update `CommitSessionHandler` to use `ImageRepository` for `rescan_images`

## Related Issue
- Jira: BA-4079

## Test plan
- [x] Verify existing session repository tests pass
- [x] Verify commit session functionality works correctly with ImageRepository

### DIFF
--- a/tests/unit/manager/repositories/session/test_session_repository.py
+++ b/tests/unit/manager/repositories/session/test_session_repository.py
@@ -84,14 +84,6 @@ class TestSessionRepository:
         return SessionRepository(db_with_cleanup)
 
     @pytest.fixture
-    def default_querier(self) -> BatchQuerier:
-        return BatchQuerier(
-            pagination=OffsetPagination(limit=10, offset=0),
-            conditions=[],
-            orders=[],
-        )
-
-    @pytest.fixture
     async def session_with_kernel(self, db_with_cleanup: ExtendedAsyncSAEngine) -> SessionTestData:
         """Create a session with kernel for testing search operations."""
         domain_name = "test-domain"
@@ -312,11 +304,15 @@ class TestSessionRepository:
     async def test_search_kernels(
         self,
         repository: SessionRepository,
-        default_querier: BatchQuerier,
         session_with_kernel: SessionTestData,
     ) -> None:
         """Test search_kernels returns kernel info when kernels exist"""
-        result = await repository.search_kernels(default_querier)
+        querier = BatchQuerier(
+            pagination=OffsetPagination(limit=10, offset=0),
+            conditions=[],
+            orders=[],
+        )
+        result = await repository.search_kernels(querier)
 
         assert result.total_count == 1
         assert len(result.items) == 1
@@ -331,10 +327,14 @@ class TestSessionRepository:
     async def test_search_kernels_empty_result(
         self,
         repository: SessionRepository,
-        default_querier: BatchQuerier,
     ) -> None:
         """Test search_kernels returns empty result when no kernels exist"""
-        result = await repository.search_kernels(default_querier)
+        querier = BatchQuerier(
+            pagination=OffsetPagination(limit=10, offset=0),
+            conditions=[],
+            orders=[],
+        )
+        result = await repository.search_kernels(querier)
 
         assert result.total_count == 0
         assert len(result.items) == 0
@@ -349,11 +349,15 @@ class TestSessionRepository:
     async def test_search_sessions(
         self,
         repository: SessionRepository,
-        default_querier: BatchQuerier,
         session_with_kernel: SessionTestData,
     ) -> None:
         """Test search returns session data when sessions exist"""
-        result = await repository.search(querier=default_querier)
+        querier = BatchQuerier(
+            pagination=OffsetPagination(limit=10, offset=0),
+            conditions=[],
+            orders=[],
+        )
+        result = await repository.search(querier=querier)
 
         assert result.total_count == 1
         assert len(result.items) == 1
@@ -372,10 +376,14 @@ class TestSessionRepository:
     async def test_search_sessions_empty_result(
         self,
         repository: SessionRepository,
-        default_querier: BatchQuerier,
     ) -> None:
         """Test search returns empty result when no sessions exist"""
-        result = await repository.search(querier=default_querier)
+        querier = BatchQuerier(
+            pagination=OffsetPagination(limit=10, offset=0),
+            conditions=[],
+            orders=[],
+        )
+        result = await repository.search(querier=querier)
 
         assert result.total_count == 0
         assert len(result.items) == 0


src/ai/backend/manager/repositories/image/repository.py
```diff
@@ -314,3 +320,12 @@ async def search_images(self, querier: BatchQuerier) -> ImageListResult:
...
         return await self._db_source.search_images(querier)
+
+    @image_repository_resilience.apply()
+    async def rescan_images(
+        self,
+        registry_or_image: str | None = None,
+        project: str | None = None,
+        reporter: ProgressReporter | None = None,
+    ) -> RescanImagesResult:
+        return await self._db_source.rescan_images(registry_or_image, project, reporter)
```
MinJuTur, line 331: Moved `rescan_images` from `SessionRepository` to `ImageRepository` as requested. While it is called from the session commit workflow, the method itself has no knowledge of sessions and belongs to the image domain.

src/ai/backend/manager/repositories/session/db_source/db_source.py
```diff
@@ -0,0 +1,552 @@
...
+from ai.backend.manager.utils import query_userinfo
+
+
+class SessionDBSource:
+    _db: ExtendedAsyncSAEngine
+
+    def __init__(self, db: ExtendedAsyncSAEngine) -> None:
+        self._db = db
+
+    async def get_session_owner(self, session_id: str | SessionId) -> UserData | None:
```
jopemachine: Return type seems not optional

tests/unit/manager/repositories/session/test_session_repository.py
```diff
@@ -133,7 +133,7 @@ async def test_search_sessions(
             mock_result.has_previous_page = False
 
             mp.setattr(
-                "ai.backend.manager.repositories.session.repository.execute_batch_querier",
+                "ai.backend.manager.repositories.session.db_source.db_source.execute_batch_querier",
```
HyeockJinKim: In the repository layer, tests must be written based on database connections to verify that values actually function. Could you please make the necessary modifications?

tests/unit/manager/repositories/session/test_session_repository.py
```diff
@@ -45,185 +42,15 @@
 from ai.backend.manager.repositories.session.repository import SessionRepository
 from ai.backend.testutils.db import with_tables
 
-if TYPE_CHECKING:
-    pass
 
-
-class TestSessionRepositorySearch:
```
MinJuTur, line 52: Removed mock-based tests as they don't verify actual SQL query execution. Replaced with real database integration tests below.

tests/unit/manager/repositories/session/test_session_repository.py
```diff
@@ -45,185 +42,15 @@
...
-            assert result.has_next_page is True
-            assert result.has_previous_page is True
+@dataclass
+class SessionTestData:
+    domain_name: str
+    user_id: uuid.UUID
+    group_id: uuid.UUID
+    session_id: SessionId
+    kernel_id: KernelId
+    access_key: AccessKey
```
MinJuTur, line 53: Introduced a dataclass for type-safe fixture return values instead of using a plain dict.

tests/unit/manager/repositories/session/test_session_repository.py
```diff
@@ -257,10 +84,16 @@ def repository(self, db_with_cleanup: ExtendedAsyncSAEngine) -> SessionRepositor
...
-    async def setup_test_data(
-        self, db_with_cleanup: ExtendedAsyncSAEngine
-    ) -> dict[str, uuid.UUID | str]:
-        """Setup test domain, user, group, keypair, session, and kernel"""
+    def default_querier(self) -> BatchQuerier:
+        return BatchQuerier(
+            pagination=OffsetPagination(limit=10, offset=0),
+            conditions=[],
+            orders=[],
+        )
```
MinJuTur: Extracted common BatchQuerier setup into a reusable fixture.

tests/unit/manager/repositories/session/test_session_repository.py
```diff
@@ -257,10 +84,16 @@ def repository(self, db_with_cleanup: ExtendedAsyncSAEngine) -> SessionRepositor
...
+
+    @pytest.fixture
+    async def session_with_kernel(self, db_with_cleanup: ExtendedAsyncSAEngine) -> SessionTestData:
```
MinJuTur, line 87: Renamed from `setup_test_data` to `session_with_kernel` for clarity about what the fixture provides.

tests/unit/manager/repositories/session/test_session_repository.py
```diff
@@ -45,185 +42,15 @@
...
-                pagination=OffsetPagination(limit=10, offset=10),
-                conditions=[],
-                orders=[],
-            )
-
-            result = await session_repository.search(querier=querier)
-
-            assert result.total_count == 25
-            assert result.has_next_page is True
-            assert result.has_previous_page is True
```
MinJuTur, line 226: Removed `test_search_sessions_with_pagination` and `test_search_kernels_with_pagination` as pagination behavior is already covered in `test_pagination_integration.py`. These tests were essentially testing the internal `execute_batch_querier` function rather than repository-specific logic.

src/ai/backend/manager/repositories/session/db_source/db_source.py
```diff
@@ -504,7 +504,7 @@ async def search(
             SessionListResult with items, total count, and pagination info
         """
         async with self._db.begin_readonly_session() as db_sess:
-            query = sa.select(SessionRow)
+            query = sa.select(SessionRow).options(selectinload(SessionRow.kernels))
```
MinJuTur, line 507: Added eager loading for `SessionRow.kernels` to fix lazy loading error. Without this, `SessionRow.to_dataclass()` fails when accessing `main_kernel` outside the database session context.

tests/unit/manager/repositories/session/test_session_repository.py
jopemachine: I think this hide the test input from each tests.
Extracting fixtures to avoid code duplication is good, but I think hiding the test inputs actually makes the tests worse rather than better.

tests/unit/manager/repositories/session/test_session_repository.py
```diff
@@ -496,46 +320,72 @@ async def test_search_kernels_returns_results(
...
         self,
         repository: SessionRepository,
-        setup_test_data: dict[str, uuid.UUID | str],
+        session_with_kernel: SessionTestData,
     ) -> None:
-        """Test search_kernels respects pagination parameters"""
-        # Test with offset that should return no results
+        """Test search returns session data when sessions exist"""
         querier = BatchQuerier(
-            pagination=OffsetPagination(limit=10, offset=10),
```
jopemachine, line 531: Please read existing test code, this test is been removed now

src/ai/backend/manager/repositories/session/db_source/db_source.py
jopemachine: If we do it this way, a separate query will be executed every time.
Isn't there a better approach?

tests/unit/manager/repositories/session/test_session_repository.py
MinJuTur: > Please read existing test code, this test is been removed now

I think there's a misunderstanding. The `BatchQuerier` code in `test_search_sessions()` is not from the removed `test_search_kernels_with_pagination()`. This code was originally in `test_search_sessions()`.

src/ai/backend/manager/repositories/session/db_source/db_source.py
MinJuTur: > If we do it this way, a separate query will be executed every time. Isn't there a better approach?

The alternative is `joinedload`, which executes a single JOIN query. However, for one-to-many relationships (Session → Kernels), `joinedload` causes data duplication - each session row is repeated for every kernel in the JOIN result, increasing network transfer size. `selectinload` executes 2 queries but without data duplication.

tests/unit/manager/repositories/session/test_session_repository.py
jopemachine: Original test's offset is 10.

MinJuTur:
```
@pytest.mark.asyncio
    async def test_search_sessions(
        self,
        session_repository: SessionRepository,
        mock_db: MagicMock,
        sample_session_data: SessionData,
    ) -> None:
        """Test searching sessions with querier"""
        mock_row = MagicMock()
        mock_row.SessionRow.to_dataclass.return_value = sample_session_data

        mock_db_sess = MagicMock()
        mock_db_sess.__aenter__ = AsyncMock(return_value=mock_db_sess)
        mock_db_sess.__aexit__ = AsyncMock(return_value=None)
        mock_db.begin_readonly_session.return_value = mock_db_sess

        # Mock execute_batch_querier result
        with pytest.MonkeyPatch.context() as mp:
            mock_result = MagicMock()
            mock_result.rows = [mock_row]
            mock_result.total_count = 1
            mock_result.has_next_page = False
            mock_result.has_previous_page = False

            mp.setattr(
                "ai.backend.manager.repositories.session.repository.execute_batch_querier",
                AsyncMock(return_value=mock_result),
            )

            querier = BatchQuerier(
                pagination=OffsetPagination(limit=10, offset=0),
                conditions=[],
                orders=[],
            )

            result = await session_repository.search(querier=querier)

            assert result.items == [sample_session_data]
            assert result.total_count == 1
            assert result.has_next_page is False
            assert result.has_previous_page is False
```
The original `test_search_sessions` used `offset=0`. 
I have removed `test_search_kernels_with_pagination`.

jopemachine: https://github.com/lablup/backend.ai/pull/8534/files#diff-f447694819401b565fede97d9ff55397908417f9dcff6c5d4769d14a3524c33cL217

jopemachine: It looks like we need to add search fixtures and corresponding test code for multiple entities. 

Testing search functionality with only the session_with_kernel fixture is not sufficient.

Previously, we were mocking the return values of repository methods, so this was overlooked, but now we need to test these parts with real behavior instead.

HyeockJinKim (approved):

HyeockJinKim (approved):

PR #8581, opened 2026-02-04
refactor(BA-4270): introduce TargetScope

### LINKED ISSUE #8580: Add TargetScope type for validation
## Motivation

  The existing SearchScope serves two distinct roles:

1. Query filtering — converting scope parameters into SQL conditions (to_condition)
1. Entity validation — verifying that referenced entities exist before query execution (existence_checks)

  However, it lacks the ability to express which scope a permission check should target and what prerequisite scopes must independently pass. This information is essential for RBAC permission validation, where a single operation may require authorization against multiple scopes (e.g., accessing a project-level resource requires both domain-level and project-level permission checks).

  Additionally, SearchScope is tied to query operations via to_condition(). Non-query operations (e.g., entity creation, mutation, deletion) also need permission scope definitions but have no use for query conditions. There is currently no base abstraction to support this.

## Approach

  Extract a TargetScope base class from SearchScope that captures the permission-related interface:

- target — the primary scope against which permissions are checked
- prerequisite_scopes — additional scopes that must each independently satisfy the permission check (AND logic)
- existence_checks — entity existence validation (moved up from SearchScope)

SearchScope continues to extend TargetScope with to_condition() for query-specific use cases. ScopeValidationArgs and EntityValidationArgs are introduced as standardized input structures for the permission validation layer that will consume TargetScope.

JIRA Issue: BA-4270

fregataa:
resolves #8580 (BA-4270)

## Summary
 - Introduce TargetScope as a new abstract base class that defines the permission check interface (target, prerequisite_scopes, existence_checks)
  - SearchScope now extends TargetScope, inheriting the permission-related properties while keeping its own to_condition() for query filtering
  - Add ScopeValidationArgs and EntityValidationArgs dataclasses to support upcoming RBAC permission validation workflows
  - Implement target and prerequisite_scopes in all existing SearchScope subclasses across fair share, group, resource usage history, and user repositories

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/manager/repositories/base/types.py
```diff
@@ -60,10 +57,51 @@ def existence_checks(self) -> Sequence[ExistenceCheck[Any]]:
...
+    @abstractmethod
+    def target(self) -> ScopeId:
+        """The primary scope for permission checks."""
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def prerequisite_scopes(self) -> Collection[ScopeId]:
+        """Scopes that must each independently satisfy the permission check (AND logic)."""
+        raise NotImplementedError
```
HyeockJinKim, line 70: Wouldn't it be better to just combine this part into existence_checks?

fregataa: existence_checks looks up real table (users, groups, domains) and these are used to look up RBAC table. And existence_checks raises not-found while they raise forbidden

fregataa:
Closed this PR since the idea Target Scope is not planned anymore

PR #8615, opened 2026-02-05
fix(BA-4281): Prevent unintended user group deletion on role change

jopemachine:
resolves (BA-4281)

- [x] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/tests/unit/manager/repositories/user/test_user_repository.py
+++ b/tests/unit/manager/repositories/user/test_user_repository.py
@@ -272,20 +272,16 @@ class TestUserRepository:
         return str(group_id)
 
     @pytest.fixture
-    async def sample_user_with_group(
+    def sample_user_creator(
         self,
-        user_repository: UserRepository,
         sample_domain: str,
         user_resource_policy: str,
-        default_keypair_resource_policy: str,
-        sample_group_id: str,
-    ) -> UserWithGroup:
-        """Create a test user with group association."""
-        password_info = create_test_password_info("test_password")
+    ) -> Creator[UserRow]:
+        """Create a Creator for a test user."""
         spec = UserCreatorSpec(
             username=f"testuser-{uuid.uuid4().hex[:8]}",
             email=f"test-{uuid.uuid4().hex[:8]}@example.com",
-            password=password_info,
+            password=create_test_password_info("test_password"),
             need_password_change=False,
             full_name="Test User",
             description="Test Description",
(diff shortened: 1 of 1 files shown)


tests/unit/manager/repositories/user/test_user_repository.py
```diff
@@ -446,6 +496,67 @@ async def test_update_user_validated_success(
...
+        """Verify that sample_user_with_group fixture creates user with group association."""
+        async with db_with_cleanup.begin_session() as session:
+            groups = await session.scalars(
+                sa.select(AssocGroupUserRow).where(
+                    AssocGroupUserRow.user_id == sample_user_with_group.user_uuid
+                )
+            )
+            group_list = list(groups)
+            assert len(group_list) == 1
+            assert str(group_list[0].group_id) == sample_user_with_group.group_id
```
seedspirit: It seems like a test that verifies fixtures rather than business logic, but I'm not sure if it's absolutely necessary.

tests/unit/manager/repositories/user/test_user_repository.py
```diff
@@ -446,6 +496,67 @@ async def test_update_user_validated_success(
...
+
+    @pytest.mark.asyncio
+    async def test_update_user_role_preserves_group_associations(
+        self,
+        db_with_cleanup: ExtendedAsyncSAEngine,
+        user_repository: UserRepository,
+        sample_user_with_group: UserWithGroup,
```
seedspirit, line 533: `test_update_user_role_preserves_group_associations` tests how role changes affect group associations. However, the current test makes it difficult to determine the user fixture's existing role. Therefore, it would be better to use the fixture's indirect method or rename the fixture to something like `sample_user_with_user_role`.

tests/unit/manager/repositories/user/test_user_repository.py
jopemachine: That’s a good point.
test_sample_user_with_group_has_group_association is somewhat ambiguous since it’s essentially testing the sample_user_with_group fixture.

It would be better to have tests for user_repository.create_user_validated separate from the fixture.

tests/unit/manager/repositories/user/test_user_repository.py
jopemachine: I don’t agree.

Since all users have a user_role, the name `sample_user_with_user_role` would not be appropriate.
The purpose of this regression test is to ensure that the group association is not removed.
Therefore, the name `sample_user_with_group` is not awkward.

PR #8629, opened 2026-02-06
feat(BA-4131): Implement resource_group field resolver in KernelV2GQL

jopemachine:
## Summary
- Implement the `resource_group` node reference field resolver in `KernelV2GQL`
- Uses the existing `resource_group_loader` dataloader to efficiently batch-load resource groups by name
- Returns `ResourceGroupGQL` for the kernel's assigned scaling group, or `null` if not assigned

## Test plan
- [ ] Verify `resource_group` field resolves correctly in GraphQL queries for kernels
- [ ] Verify `null` is returned when `resource_group_name` is not set
- [ ] Verify batch loading works correctly when querying multiple kernels


jopemachine:
Consolidated into #8633

PR #8630, opened 2026-02-06
feat(BA-4131): Implement domain field resolver in KernelV2GQL

jopemachine:
## Summary
- Implement the `domain` node reference field resolver in `KernelV2GQL`
- Uses `GetDomainAction` via `processors.domain.get_domain` to fetch the domain by name
- Returns `DomainV2GQL` for the kernel's associated domain, or `null` if domain_name is not set

## Test plan
- [ ] Verify `domain` field resolves correctly in GraphQL queries for kernels
- [ ] Verify `null` is returned when `domain_name` is not set
- [ ] Verify domain data matches the kernel's `user_info.domain_name`


jopemachine:
Consolidated into #8633

PR #8631, opened 2026-02-06
feat(BA-4131): Implement project field resolver in KernelV2GQL

jopemachine:
## Summary
- Implement the `project` node reference field resolver in `KernelV2GQL`
- Uses `GetProjectAction` via `processors.group.get_project` to fetch the project by UUID
- Returns `ProjectV2GQL` for the kernel's associated project/group, or `null` if group_id is not set

## Test plan
- [ ] Verify `project` field resolves correctly in GraphQL queries for kernels
- [ ] Verify `null` is returned when `group_id` is not set
- [ ] Verify project data matches the kernel's `user_info.group_id`


jopemachine:
Consolidated into #8633

PR #8632, opened 2026-02-06
feat(BA-4131): Implement user field resolver in KernelV2GQL

jopemachine:
## Summary
- Implement the `user` node reference field resolver in `KernelV2GQL`
- Uses `GetUserAction` via `processors.user.get_user` to fetch the user by UUID
- Returns `UserV2GQL` for the kernel's owner user, or `null` if user_id is not set

## Test plan
- [ ] Verify `user` field resolves correctly in GraphQL queries for kernels
- [ ] Verify `null` is returned when `user_id` is not set
- [ ] Verify user data matches the kernel's `user_info.user_id`


jopemachine:
Consolidated into #8633

PR #8647, opened 2026-02-06
fix(BA-4288): Return 409 instead of 500 for duplicate deployment name

### LINKED ISSUE #8622: Duplicate endpoint name causes 500 Internal Server Error
## Summary

When creating an endpoint with a duplicate name in create_endpoint_legacy, a database integrity error occurs and returns a 500 Internal Server Error. This is a UX issue as users are not provided with an appropriate error message (409 Conflict).

## Steps to Reproduce

1. Create an endpoint in a specific project/domain (e.g., name="my-deployment")
1. Attempt to create another endpoint with the same name in the same project/domain (name="my-deployment")
1. A 500 Internal Server Error is returned

## Expected Behavior

- When attempting to create an endpoint with a duplicate name, a 409 Conflict response should be returned
- The error message should clearly indicate the name is duplicated (e.g., "Endpoint with name 'my-deployment' already exists in this project")
- Users should be able to immediately understand the cause and retry with a different name

## Actual Behavior

- A 500 Internal Server Error is returned due to database unique constraint violation
- Users cannot identify the cause of the error and may assume it is a server failure

## Root Cause

The create_endpoint_legacy method (src/ai/backend/manager/repositories/deployment/db_source/db_source.py:269) lacks name duplication validation before endpoint creation. When execute_rbac_entity_creator is executed, an IntegrityError is raised by the DB's unique constraint, and this exception is not properly handled, resulting in a 500 error.

## Proposed Solution

1. Add _check_endpoint_name_exists private method

- Pre-validate endpoint name duplication within the same project/domain

1. Call duplication check before endpoint creation in create_endpoint_legacy
1. Raise EndpointNameAlreadyExists exception when duplicate is found (mapped to 409 Conflict)

## Affected Files

- src/ai/backend/manager/repositories/deployment/db_source/db_source.py
- src/ai/backend/manager/errors/deployment.py (add new exception class)

## Test Cases

- Verify 409 Conflict response when creating endpoint with duplicate name
- Verify successful creation when using the same name in different projects
- Verify error message includes the duplicated name

JIRA Issue: BA-4288

seedspirit:
resolves #8622 (BA-4288)

- [x] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/tests/unit/manager/repositories/deployment/test_deployment_repository.py
+++ b/tests/unit/manager/repositories/deployment/test_deployment_repository.py
@@ -36,6 +36,7 @@ from ai.backend.manager.data.deployment.types import (
     RouteTrafficStatus,
 )
 from ai.backend.manager.data.image.types import ImageType
+from ai.backend.manager.data.permission.id import ScopeId
 from ai.backend.manager.errors.deployment import DeploymentRevisionNotFound
 from ai.backend.manager.errors.service import AutoScalingPolicyNotFound, DeploymentPolicyNotFound
 from ai.backend.manager.models.agent import AgentRow, AgentStatus
@@ -3129,9 +3130,8 @@ class TestDeploymentRepositoryDuplicateName:
         )
         return RBACEntityCreator(
             spec=spec,
-            scope_type=ScopeType.PROJECT,
-            scope_id=str(group.id),
             entity_type=EntityType.MODEL_DEPLOYMENT,
+            scope_ref=ScopeId(scope_type=ScopeType.PROJECT, scope_id=str(group.id)),
         )
 
     @pytest.mark.asyncio


src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -240,6 +241,9 @@ async def create_endpoint(
         spec = cast(DeploymentCreatorSpec, creator.spec)
         async with self._begin_session_read_committed() as db_sess:
             await self._check_group_exists(db_sess, spec.metadata.domain, spec.metadata.project_id)
+            await self._check_endpoint_name_exists(
+                db_sess, spec.metadata.domain, spec.metadata.project_id, spec.metadata.name
+            )
```
HyeockJinKim, line 246: It should be triggered as an Integrity Error rather than a pre-check. I'll let this one slide for now.

PR #8745, opened 2026-02-11
fix: use IF EXISTS for object_permissions drops in f41bbe0c0f12

fregataa:
resolves #NNN (BA-MMM)
object_permissions.permission_group_id is added by f8a9b3c2d1e0 which is on a separate migration branch and may not be applied.

- [ ] Test case(s) to:
- [ ] Documentation


HyeockJinKim:
It seems like it wasn't applied to the models side, which caused an issue. Please be careful next time.

PR #8766, opened 2026-02-11
fix(BA-4369): Rename `User` GQL types to follow V2 naming convention

jopemachine:
Resolve BA-4369.

## Summary
- Rename all user GQL type classes and their external strawberry `name=` parameters to follow `UserV2` naming convention
- Primarily modifies files within the `user/` GQL directory, plus minimal cross-module TYPE_CHECKING import updates in `domain_v2/types/node.py` and `project_v2/types/node.py`
- Examples: `UserFilterGQL` -> `UserV2FilterGQL`, `UserStatusEnumGQL` -> `UserV2StatusEnumGQL`, `UserBasicInfoGQL` -> `UserV2BasicInfoGQL`

## Test plan
- [x] All pre-commit hooks pass (ruff, mypy, BUILD)
- [x] Unit tests pass (`test_filters.py` updated)
- [ ] Confirm all renamed types are consistent across types/, fetcher/, resolver/

----
📚 Documentation preview 📚: https://sorna--8766.org.readthedocs.build/en/8766/

----
📚 Documentation preview 📚: https://sorna-ko--8766.org.readthedocs.build/ko/8766/

### DIFF
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -11558,6 +11558,18 @@ enum UserFairShareOrderField
   USER_EMAIL @join__enumValue(graph: STRAWBERRY)
 }
 
+input UserFairShareScope
+  @join__type(graph: STRAWBERRY)
+{
+  """Resource group to filter fair shares by."""
+  resourceGroupName: String!
+
+  """
+  Project ID that the user belongs to (required for user-level fair shares).
+  """
+  projectId: UUID!
+}
+
 """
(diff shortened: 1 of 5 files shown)


jopemachine:
Personal opinion on the naming convention:


There’s no real need to include V2 in scope names.


It would be better not to add V2 to mutation names either. Mutations don’t really have types per data structure; it makes more sense for the mutation itself to carry the version rather than encoding it in the name.

jopemachine:
For the User, Project, and Domain types, we’ve decided not to modify them since they are already being used on the client side.

PR #8771, opened 2026-02-11
feat(BA-3953): Implement bulk_update_users mutation (Strawberry GQL)

MinJuTur:
## Summary
- Implement `admin_bulk_update_users` Strawberry GraphQL mutation for updating multiple users in bulk
- Each user update is executed within a savepoint, enabling partial failure support (successful updates persist even when others fail)

## Related Issue
- Jira: BA-3953

## Changes by Layer
- **Base Repository**: Add `BulkUpdaterError` dataclass (mirrors `BulkCreatorError`)
- **Data Types**: Add `BulkUserUpdateResultData` with success/failure tracking
- **Actions**: Add `BulkModifyUserAction`, `BulkModifyUserActionResult`, `UserUpdateSpec`
- **DB Source**: Add `bulk_update_users_validated()` with savepoint-based partial failure, `_update_single_user_validated()` helper, `_get_user_by_uuid_with_conn()` lookup
- **Repository**: Add `bulk_update_users_validated()` with resilience decorator
- **Service**: Add `bulk_modify_users()` method
- **Processors**: Register `bulk_modify_users` processor
- **GQL Types**: Add `BulkUpdateUserItemInputGQL`, `BulkUpdateUserInputGQL`, `BulkUpdateUserErrorGQL`, `BulkUpdateUsersPayloadGQL`
- **Resolver**: Add `admin_bulk_update_users` mutation with full field mapping from `UpdateUserInputGQL` to `UserUpdaterSpec`
- **Schema**: Register mutation in `Mutation` class
- **Tests**: Add success and partial failure tests for `bulk_update_users_validated`

## Test plan
- [x] `pants lint --changed-since=HEAD~1` passes
- [x] `pants check --changed-since=HEAD~1` passes (mypy)
- [x] `pants test --changed-since=HEAD~1 --changed-dependents=transitive` passes
- [x] Unit test: bulk update 3 users successfully
- [x] Unit test: partial failure (1 success + 1 non-existent UUID failure)

----
📚 Documentation preview 📚: https://sorna--8771.org.readthedocs.build/en/8771/

----
📚 Documentation preview 📚: https://sorna-ko--8771.org.readthedocs.build/ko/8771/

### DIFF
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -1449,7 +1449,7 @@ input BulkCreateUserV2Input
   users: [CreateUserV2Input!]!
 }
 
-"""Added in 26.2.0. Error information for a failed user in bulk update."""
+"""Added in 26.3.0. Error information for a failed user in bulk update."""
 type BulkUpdateUserError
   @join__type(graph: STRAWBERRY)
 {
@@ -1460,7 +1460,7 @@ type BulkUpdateUserError
   message: String!
 }
 
-"""Added in 26.2.0. Payload for bulk user update mutation."""
+"""Added in 26.3.0. Payload for bulk user update mutation."""
 type BulkUpdateUsersV2Payload
   @join__type(graph: STRAWBERRY)
 {
@@ -1472,7 +1472,7 @@ type BulkUpdateUsersV2Payload
 }
 
 """
-Added in 26.2.0. Input for bulk updating multiple users. Each user has individual update specifications.
+Added in 26.3.0. Input for bulk updating multiple users. Each user has individual update specifications.
 """
 input BulkUpdateUserV2Input
   @join__type(graph: STRAWBERRY)
@@ -1482,7 +1482,7 @@ input BulkUpdateUserV2Input
 }
 
 """
-Added in 26.2.0. Input for a single user update within a bulk operation. Pairs a user ID with the fields to update.
+Added in 26.3.0. Input for a single user update within a bulk operation. Pairs a user ID with the fields to update.
 """
 input BulkUpdateUserV2ItemInput
   @join__type(graph: STRAWBERRY)
@@ -7423,12 +7423,12 @@ type Mutation
   adminBulkCreateUsers(input: BulkCreateUserV2Input!): BulkCreateUsersV2Payload! @join__field(graph: STRAWBERRY)
 
   """
-  Added in 26.2.0. Update multiple users in bulk (admin only). Requires superadmin privileges. Each user has individual update specifications.
+  Added in 26.3.0. Update multiple users in bulk (admin only). Requires superadmin privileges. Each user has individual update specifications.
   """
-  adminBulkUpdateUsers(input: BulkUpdateUserV2Input!): BulkUpdateUsersV2Payload! @join__field(graph: STRAWBERRY)
+  adminBulkUpdateUsersV2(input: BulkUpdateUserV2Input!): BulkUpdateUsersV2Payload! @join__field(graph: STRAWBERRY)
 
   """
-  Added in 26.2.0. Update a user's information (admin only). Requires superadmin privileges. Only provided fields will be updated.
+  Added in 26.3.0. Update a user's information (admin only). Requires superadmin privileges. Only provided fields will be updated.
   """
   adminUpdateUser(userId: UUID!, input: UpdateUserV2Input!): UpdateUserV2Payload! @join__field(graph: STRAWBERRY)
 
@@ -11682,7 +11682,7 @@ type UpdateRouteTrafficStatusPayload
 }
 
 """
-Added in 26.2.0. Input for updating user information. All fields are optional - only provided fields will be updated.
+Added in 26.3.0. Input for updating user information. All fields are optional - only provided fields will be updated.
 """
 input UpdateUserV2Input
   @join__type(graph: STRAWBERRY)
@@ -11736,7 +11736,7 @@ input UpdateUserV2Input
   containerGids: [Int!]
 }
 
-"""Added in 26.2.0. Payload for user update mutation."""
+"""Added in 26.3.0. Payload for user update mutation."""
 type UpdateUserV2Payload
   @join__type(graph: STRAWBERRY)
 {


src/ai/backend/manager/repositories/user/db_source/db_source.py
```diff
@@ -401,6 +403,118 @@ async def update_user_validated(
...
+        """Update a single user with full validation.
+
+        Extracted from update_user_validated to be reused by both
+        single and bulk update operations.
+        """
+        to_update = updater_spec.build_values()
+
+        # Get current user data for validation (by UUID)
+        current_user = await self._get_user_by_uuid_with_conn(conn, user_id)
+
```
MinJuTur, line 449: `_update_single_user_validated` largely duplicates the validation/update logic of the existing `update_user_validated` — the only difference is that it uses `_get_user_by_uuid_with_conn` instead of `_get_user_by_email_with_conn` for user lookup. 
To reduce this duplication, we could extract the logic after user lookup into a common helper like `_apply_user_update(conn, current_user, updater_spec)`. What do you think?

src/ai/backend/manager/api/gql/user/types/payloads.py
```diff
@@ -66,6 +66,32 @@ class UpdateUserPayloadGQL:
...
 
+@strawberry.type(
+    name="BulkUpdateUserError",
+    description="Added in 26.2.0. Error information for a failed user in bulk update.",
+)
+class BulkUpdateUserErrorGQL:
+    """Error information for a single user that failed during bulk update."""
+
+    user_id: UUID = strawberry.field(description="UUID of the user that failed to update.")
+    message: str = strawberry.field(description="Error message describing the failure.")
```
MinJuTur: Unlike `BulkCreateUserErrorGQL` which uses `index`/`username`/`email` to identify failed items (since newly created users don't have UUIDs yet), `BulkUpdateUserErrorGQL` uses `user_id` directly since update targets are already identified by UUID in the input.

src/ai/backend/manager/api/gql/user/resolver/mutation.py
```diff
@@ -170,6 +180,99 @@ async def admin_update_user(
...
+
+        # Convert status enum if provided
+        status_state: OptionalState[UserStatus] = OptionalState.nop()
+        if user_input.status is not None and user_input.status is not UNSET:
+            status_state = OptionalState.update(UserStatus(user_input.status.value))
+
+        # Convert role enum if provided
+        role_state: OptionalState[UserRole] = OptionalState.nop()
+        if user_input.role is not None and user_input.role is not UNSET:
+            role_state = OptionalState.update(UserRole(user_input.role.value))
```
MinJuTur: `status` and `role` use the extract-to-variable pattern (like `password` and `group_ids`) because they require enum conversion before passing to `OptionalState`. However, this means explicit `null` for these fields is silently treated as `nop()`, while other `OptionalState` fields (e.g. `username`, `full_name`) raise `ValueError` on explicit `null` via `OptionalState.from_graphql(None)`. This inconsistency is minor since clients shouldn't send explicit `null` for NOT NULL fields, but could be unified later by adding early validation that rejects explicit `null` for non-nullable fields before reaching the conversion logic.

src/ai/backend/manager/api/gql/user/resolver/mutation.py
```diff
@@ -170,6 +180,99 @@ async def admin_update_user(
...
+    """Update multiple users in bulk with individual specifications.
+
+    Args:
+        info: Strawberry GraphQL context.
+        input: Bulk user update input with individual specs.
+
+    Returns:
+        BulkUpdateUsersPayloadGQL with updated users and failures.
+    """
+    ctx = info.context
```
jopemachine, line 206: I think we should call `check_admin_only` at first

src/ai/backend/manager/api/gql/user/types/inputs.py
```diff
@@ -103,71 +104,100 @@ class UpdateUserInputGQL:
...
 
 
+@strawberry.input(
+    name="BulkUpdateUserV2ItemInput",
+    description=(
+        "Added in 26.2.0. Input for a single user update within a bulk operation. "
+        "Pairs a user ID with the fields to update."
+    ),
+)
+class BulkUpdateUserItemInputGQL:
```
jopemachine: Shouldn’t we also add V2 to the Python types?

jopemachine: And personally, I’m not sure whether we really need to add V2 to the mutation, and mutation types as well.

src/ai/backend/manager/api/gql/user/resolver/mutation.py
jopemachine: When value conversion is required, the logic tends to become overly messy.
I think we'd better to add `map` like method like this to simplify the code.

```py
  group_ids_state = OptionalState.from_graphql(user_input.group_ids).map(
      lambda gids: [str(gid) for gid in gids]
  )
  status_state = OptionalState.from_graphql(user_input.status).map(
      lambda s: UserStatus(s.value)
  )
  role_state = OptionalState.from_graphql(user_input.role).map(
      lambda r: UserRole(r.value)
  )
```

What do you think? @HyeockJinKim

src/ai/backend/manager/repositories/user/db_source/db_source.py
```diff
@@ -401,6 +403,118 @@ async def update_user_validated(
...
+        Args:
+            items: List of UserUpdateSpec for each user to update.
+        """
+        if not items:
+            return BulkUserUpdateResultData(successes=[], failures=[])
+
+        successes: list[UserData] = []
+        failures: list[BulkUpdaterError[UserRow]] = []
+
+        async with self._db.begin() as conn:
```
jopemachine, line 418: If possible, new code should be written using the ORM.

Is there a particular reason you implemented this using the Connection and Core API style?

src/ai/backend/manager/api/gql/user/types/inputs.py
MinJuTur: The `V2` in GraphQL schema names is needed to distinguish them from the legacy V1 types that clients see. Python class names don't include `V2` because the `GQL` suffix already distinguishes them from legacy graphene types in our codebase. Do you think we should unify them by adding `V2` to the Python class names as well (e.g. `BulkUpdateUserItemInputV2GQL`)? If so, it would affect all user types, so I'd prefer to handle it in a separate PR.

src/ai/backend/manager/repositories/user/db_source/db_source.py
MinJuTur: > Is there a particular reason you implemented this using the Connection and Core API style?

`_update_single_user_validated` reuses the same validation and update logic as the existing `update_user_validated`, which also uses the Core API style. Converting to ORM would require refactoring `update_user_validated` as well.

src/ai/backend/manager/api/gql/user/types/inputs.py
jopemachine: We decided to add 'V2' to all entities that involve V2, in order to prevent confusion (e.g., BulkUpdateUserV2ItemInputGQL) You don't need to rename existing types in this PR, but please follow the convention in this PR.

jopemachine:
Please update milestone to 26.3 @MinJuTur

src/ai/backend/manager/api/gql/user/types/inputs.py
jopemachine, line 87: This is not updated

docs/manager/graphql-reference/v2-schema.graphql
```diff
@@ -930,6 +930,43 @@ type BulkCreateUsersV2Payload {
   failed: [BulkCreateUserError!]!
 }
 
+"""Added in 26.2.0. Error information for a failed user in bulk update."""
+type BulkUpdateUserError {
```
jopemachine, line 934: Please update this, and also check other types

src/ai/backend/manager/api/gql/user/resolver/mutation.py
```diff
@@ -170,6 +182,82 @@ async def admin_update_user(
     raise NotImplementedError("admin_update_user is not yet implemented")
 
 
+@strawberry.mutation(
+    description=(
+        "Added in 26.2.0. Update multiple users in bulk (admin only). "
```
jopemachine: I mean to update this too

src/ai/backend/manager/api/gql/user/types/inputs.py
MinJuTur: > You don't need to rename existing types in this PR, but please follow the convention in this PR.

I've added V2 to all types introduced in this PR.
However, your recent comments also point to types from previous PRs. Should I rename those in this PR as well?
@jopemachine

src/ai/backend/manager/api/gql/user/resolver/mutation.py
MinJuTur: > I mean to update this too

Could you clarify what you'd like me to update here?

jopemachine: Added in 26.3.0.

src/ai/backend/manager/api/gql/user/resolver/mutation.py
```diff
@@ -170,6 +182,82 @@ async def admin_update_user(
...
+        "Requires superadmin privileges. "
+        "Each user has individual update specifications."
+    )
+)  # type: ignore[misc]
+async def admin_bulk_update_users(
```
jopemachine: admin_bulk_update_users_v2

src/ai/backend/manager/api/gql/user/types/inputs.py
jopemachine: No, but please do it as the very next follow-up task once this PR is merged.

PR #8780, opened 2026-02-12
fix(BA-4374): Modify legacy endpoint status resolver to ignore terminated route rows

jopemachine:
resolves BA-4374.

- [x] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/changes/8780.fix.md
+++ b/changes/8780.fix.md
@@ -1,0 +1,1 @@
+Modify legacy endpoint status resolver to ignore terminated route rows


src/ai/backend/manager/api/gql_legacy/endpoint.py
```diff
@@ -886,12 +886,16 @@ async def resolve_status(self, info: graphene.ResolveInfo) -> str:
...
             case _:
+                active_route_status_names = {s.name for s in RouteStatus.active_route_statuses()}
+                active_routings = [
+                    r for r in self.routings if r.status in active_route_status_names
+                ]
                 healthy_count = sum(
-                    1 for r in self.routings if r.status == RouteStatus.HEALTHY.name
+                    1 for r in active_routings if r.status == RouteStatus.HEALTHY.name
                 )
                 if healthy_count == 0:
```
HyeockJinKim, line 896: I'm concerned that the number of routes could become excessive, and I'm wondering if there might be a better approach.

PR #8783, opened 2026-02-12
feat(BA-3128): Generate dropbear SSH host key on agent side to prevent user deletion

### LINKED ISSUE #6904: Prevent dropbear rsa host key deletion by user making /tmp/dropbear readonly
## Motivation

Users are currently encountering situations where SSH access becomes impossible after deleting `/tmp/dropbear/dropbear_rsa_host_key` within the kernel. To prevent this, code modifications are required to grant read-only permissions to `/tmp/dropbear/dropbear_rsa_host_key` from the kernel creation stage onwards.

## Required Features

The dropbear key required for SSH connections is currently generated via the `init_sshd_service` function in `intrinsic.py`.

```python
async def init_sshd_service(child_env):
    ...
    proc = await asyncio.create_subprocess_exec(
        *[
            "/opt/kernel/dropbearmulti",
            "dropbearkey",
            "-t",
            "rsa",
            "-s",
            "2048",
            "-f",
            "/tmp/dropbear/dropbear_rsa_host_key",
        ],
        stdout=asyncio.subprocess.PIPE,
        stderr=asyncio.subprocess.PIPE,
        env=child_env,
    )
    stdout, stderr = await proc.communicate()
    if proc.returncode != 0:
        raise RuntimeError(f"sshd init error: {stderr.decode('utf8')}")
    ...
```

Read-only permissions for the subsequently generated `dropbear_rsa_host_key` have not been granted. Therefore, we intend to add logic to adjust permissions for the generated key.

```python
proc = await asyncio.create_subprocess_exec(
        *[
            "/opt/kernel/dropbearmulti",
            "dropbearkey",
            "-t",
            "rsa",
            "-s",
            "2048",
            "-f",
            "/tmp/dropbear/dropbear_rsa_host_key",
        ],
        stdout=asyncio.subprocess.PIPE,
        stderr=asyncio.subprocess.PIPE,
        env=child_env,
    )
    stdout, stderr = await proc.communicate()
    if proc.returncode != 0:
        raise RuntimeError(f"sshd init error: {stderr.decode('utf8')}")

    # Add permission settings
    Path("/tmp/dropbear/dropbear_rsa_host_key").chmod(0o400)
```

## 

## Testing Scenarios

Need to test `dropbear_rsa_host_key` operations trial(any operation except read) would fail

```python
async def test_ssh_host_key_deletion_protection_non_root():
    """Verify that non-root users cannot delete the SSH host key"""
    await init_sshd_service({})
    
    host_key = Path("/tmp/dropbear/dropbear_rsa_host_key")
    if os.geteuid() != 0:
        # Attempt to delete (should fail)
        with pytest.raises(PermissionError):
            host_key.unlink()
        
        # Attempt to modify (should fail)
        with pytest.raises(PermissionError):
            host_key.write_text("malicious content")
        
        # Attempt to move (should fail)
        with pytest.raises(PermissionError):
            host_key.rename("/tmp/test_key")
    else:
        pytest.skip("Test requires non-root user")
```

JIRA Issue: BA-3128

seedspirit:
resolves #6904 (BA-3128)

### Why use `dropbearkey` from krunner instead of Python cryptography?
  - Dropbear may not read OpenSSH format keys for host key (`-r` flag)
  - Using `dropbearkey` guarantees native Dropbear format compatibility
  - krunner binaries are already available on the Agent (same binaries mounted into
  containers)

### Compatibility Matrix

  | Agent | Kernel | Behavior |
  |-------|--------|----------|
  | New | New | Uses agent-generated key (`/home/config/ssh/`) |
  | Old | New | Falls back to container-generated key (`/tmp/dropbear/`) |
  | New | Old | Old kernel ignores agent key, generates its own |

### Cross-kernel Communication
  - No impact: `StrictHostKeyChecking no` skips host key verification between kernels
  - Cluster keypair (`id_cluster`) path remains unchanged

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/agent/docker/agent.py
+++ b/src/ai/backend/agent/docker/agent.py
@@ -732,6 +732,7 @@ class DockerKernelCreationContext(AbstractKernelCreationContext[DockerKernel]):
                     raise FileNotFoundError(
                         f"dropbearmulti binary not found at {dropbearmulti_path}"
                     )
+                # If the host key already exists, we assume it's valid and skip generation.
                 if not host_key_path.is_file():
                     try:
                         subprocess_run(
@@ -761,6 +762,7 @@ class DockerKernelCreationContext(AbstractKernelCreationContext[DockerKernel]):
                     host_key_path.chmod(0o600)
                 paths_to_chown.append(host_key_path)
 
+                # Write provided SSH keypair for cluster access if exists
                 if sshkey is not None:
                     cluster_priv_key_path = ssh_dir / "id_cluster"
                     cluster_pub_key_path = ssh_dir / "id_cluster.pub"


src/ai/backend/agent/docker/agent.py
```diff
@@ -722,43 +724,79 @@ async def apply_network(self, cluster_info: ClusterInfo) -> None:
     @override
     async def prepare_ssh(self, cluster_info: ClusterInfo) -> None:
         sshkey = cluster_info["ssh_keypair"]
-        if sshkey is None:
-            return
```
fregataa, line 717: Is it okay to remove this branch?

src/ai/backend/agent/docker/agent.py
```diff
@@ -722,43 +724,79 @@ async def apply_network(self, cluster_info: ClusterInfo) -> None:
...
+                host_key_path = ssh_dir / "dropbear_rsa_host_key"
+                arch = get_arch_name()
+                dropbearmulti_path = self.resolve_krunner_filepath(
+                    f"runner/dropbearmulti.{arch}.bin"
+                )
+                if not dropbearmulti_path.exists():
+                    raise FileNotFoundError(
+                        f"dropbearmulti binary not found at {dropbearmulti_path}"
+                    )
+                if not host_key_path.is_file():
```
fregataa, line 736: is there any chance that the file does not exist?

src/ai/backend/agent/docker/agent.py
seedspirit: The logic remains unchanged: if the `cluster_info keypair` is absent, the creation of id_cluster is still omitted.

1) dropbear is common, so it is created regardless of cluster info presence
2) id_cluster-related keypairs are created only when `if sshkey is not None:` (line 773)

PR #8804, opened 2026-02-13
fix(BA-4387): Use folder_id attribute(UUID) for extra mounts in model service dry run

### LINKED ISSUE #8803: dry_run model service fails with AttributeError: 'VFolderID' object has no attribute 'partition'
## Summary

The /services/_/try endpoint fails when there are auto-mounted vfolders (dot-prefixed vfolders like .ssh, .config) due to type mismatch between VFolderID and uuid.UUID.

## 

## Steps to Reproduce

1. Ensure the user has access to dot-prefixed vfolders (auto-mounted vfolders)
1. Call the /services/_/try endpoint with a valid model service configuration
1. The request fails with AttributeError

## Expected Behavior

The dry-run should complete successfully, validating the model service configuration.

## Actual Behavior

AttributeError: 'VFolderID' object has no attribute 'partition'

## Root Cause

In model_serving.py, the dry_run method incorrectly uses m.vfid (type: VFolderID) instead of m.vfid.folder_id (type: uuid.UUID) when building the mounts list:

dry_run method (INCORRECT)

```python
mounts = [ model_vfolder_id, *[m.vfid for m in service_prepare_ctx.extra_mounts], # VFolderID objects ]
```

This differs from the create method which correctly uses .folder_id:

create method (CORRECT)

```python
creation_config["mounts"] = [ model_vfolder_id, *[m.vfid.folder_id for m in service_prepare_ctx.extra_mounts], # uuid.UUID ]
```

The downstream prepare_vfolder_mounts function only accepts `str | uuid.UUID`, causing the error when VFolderID is passed.

JIRA Issue: BA-4387

seedspirit:
resolves #8803 (BA-4387)

- [x] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/changes/8804.fix.md
+++ b/changes/8804.fix.md
@@ -1,0 +1,1 @@
+Fix model service dry run failing when extra mounts are present by correctly extracting `folder_id` from `VFolderID` objects. Previously, the full `VFolderID` object was passed instead of the UUID, causing type mismatches in mount configuration.


tests/unit/manager/services/model_serving/actions/test_dry_run_model_service.py
```diff
@@ -835,3 +846,325 @@ async def test_service_definition_overrides_applied(
...
+                owner_uuid=uuid.UUID("00000000-0000-0000-0000-000000000001"),
+                owner_role=UserRole.USER,
+                group_id=uuid.UUID("00000000-0000-0000-0000-000000000002"),
+                resource_policy={},
+                scaling_group="default",
+                extra_mounts=[extra_mount],
+            ),
+        )
+
+    @pytest.mark.asyncio
```
fregataa, line 1125: I personally don't prefer this decorator because it makes hard to distinguish whether it is a fixture or not at the first glance (tests work without this decorator)

src/ai/backend/manager/services/model_serving/services/model_serving.py
```diff
@@ -487,17 +486,17 @@ async def dry_run(self, action: DryRunModelServiceAction) -> DryRunModelServiceA
...
-        mount_map: dict[uuid.UUID | VFolderID, str] = {
+        mount_map: dict[uuid.UUID, str] = {
             model_vfolder_id: action.config.model_mount_destination,
         }
         for m in service_prepare_ctx.extra_mounts:
-            mount_map[m.vfid] = str(m.kernel_path)
-        mount_options = {
-            m.vfid: {"permission": m.mount_perm} for m in service_prepare_ctx.extra_mounts
+            mount_map[m.vfid.folder_id] = str(m.kernel_path)
+        mount_options: dict[uuid.UUID, dict[str, Any]] = {
```
fregataa, line 498: Seems it can be a strict type rather than Any

PR #8922, opened 2026-02-14
feat(BA-4458): Migrate RBACGranter from object_permissions to permissions table

### LINKED ISSUE #8910: Migrate RBACGranter from object_permissions to permissions table
## Goal

Migrate `RBACGranter` to insert into the `permissions` table using the entity-as-scope pattern, replacing `ObjectPermissionRow` inserts.

## Details

- Replace `ObjectPermissionRow` insert with `permissions` table insert where scope is the granted entity (entity-as-scope pattern)
- Update `RBACGranter` data class if needed (field naming may change)
- Update unit tests
- Depends on `permissions` table having `role_id`, `scope_type`, `scope_id` columns (BEP-1048 Phase 1 schema)

## Files

- `src/ai/backend/manager/repositories/base/rbac/granter.py`
- `tests/unit/manager/repositories/base/rbac/test_granter.py`

JIRA Issue: BA-4458

fregataa:
resolves #8910 (BA-4458)

## Summary

Migrate RBACGranter from `object_permissions` to `permissions` table using the entity-as-scope pattern.

### RBACGranter Migration
- **Before**: Used `object_permissions` table with `granted_entity_scope_id` parameter
- **After**: Uses `permissions` table with entity-as-scope pattern
  - `granted_entity_scope_type`: The scope_type for the entity (e.g., `ScopeType.VFOLDER`)
  - `target_scope_id`: The scope to associate the entity with (e.g., invitee's User scope)

### Two-Phase Operation
The new RBACGranter performs two operations:

1. **Visibility** (ref edge): Insert into `association_scopes_entities` with `RelationType.REF`
   - Uses `ON CONFLICT DO NOTHING` to safely handle repeated grants for the same (scope_type, scope_id, entity_id) triple
2. **Access Control**: Insert entity-scope permissions into `permissions` table

### Updated Callsites
- VFolder repository methods updated to use new RBACGranter signature
- Example: When user B is invited to a VFolder, creates ref edge linking the VFolder to user B's User scope, plus entity-scope permissions

### DIFF
--- a/src/ai/backend/common/data/permission/types.py
+++ b/src/ai/backend/common/data/permission/types.py
@@ -325,6 +325,7 @@ class ScopeType(enum.StrEnum):
         except ValueError as e:
             raise RBACTypeConversionError(f"{self!r} has no corresponding RBACElementType") from e
 
+
 GLOBAL_SCOPE_ID = "global"
 
 


jopemachine (approved):

src/ai/backend/manager/repositories/base/rbac/granter.py
```diff
@@ -99,8 +74,36 @@ async def execute_rbac_granter(
     role_ids = granter.target_role_ids
     entity_id = granter.granted_entity_id
 
-    if not role_ids:
+    if not role_ids or not granter.operations:
```
seedspirit, line 77: Non-critical suggestion: Making it a property inside the granter object and giving it an appropriate name would likely improve expressiveness.

src/ai/backend/manager/repositories/base/rbac/granter.py
```diff
@@ -99,8 +74,36 @@ async def execute_rbac_granter(
...
+            entity_id=entity_id.entity_id,
+            relation_type=RelationType.REF,
+        )
+        .on_conflict_do_nothing(constraint="uq_scope_id_entity_id")
+    )
+    await db_sess.execute(ref_edge_stmt)
+
+    # 2. Insert entity-scope permissions (access control)
+    perms = [
+        PermissionRow(
```
seedspirit, line 98: I think `PermissionRow` also need to require a unique constraint as `AssociationScopesEntitiesRow` and handle conflict

fregataa: good point. I already registered it as an another issue

PR #9030, opened 2026-02-16
feat(BA-4522): Add RBACElementType unified enum

### LINKED ISSUE #9022: Define new unified RBAC enum replacing ScopeType and RBAC EntityType
Create a new unified RBAC enum in common/data/permission/types.py that merges ScopeType and RBAC-relevant EntityType values. Only BEP-1048 entities should be included (GLOBAL, DOMAIN, PROJECT, USER, STORAGE_HOST, RESOURCE_GROUP, CONTAINER_REGISTRY, SESSION, VFOLDER, IMAGE, ARTIFACT, ROLE, etc.). The existing EntityType remains untouched for the Action system.

JIRA Issue: BA-4522

fregataa:
resolves #9022 (BA-4522)

## Summary
- Add `RBACElementType` enum that unifies `ScopeType` and RBAC-relevant `EntityType` values into a single type for the BEP-1048 entity relationship model.

## Context
Currently `EntityType` is shared between the Action type system (~200 sub-entity values) and the RBAC system (~30 entities), and `ScopeType` exists as a separate enum with partial overlap. `RBACElementType` replaces both for RBAC purposes, where each node can serve as both scope (parent) and entity (child) depending on the edge direction.

### DIFF
--- a/changes/9067.fix.md
+++ b/changes/9067.fix.md
@@ -1,0 +1,1 @@
+Fix UndefinedObjectError on vfolder queries after upgrade by adding alembic migration to rename the `vfolderpermission` PostgreSQL enum type to `vfoldermountpermission`.
--- a/src/ai/backend/manager/models/alembic/versions/03ff6767b2e4_rename_vfolderpermission_type.py
+++ b/src/ai/backend/manager/models/alembic/versions/03ff6767b2e4_rename_vfolderpermission_type.py
@@ -1,0 +1,45 @@
+"""Rename vfolderpermission PG enum type to vfoldermountpermission
+
+Revision ID: 03ff6767b2e4
+Revises: 12b02ba4d44b
+Create Date: 2026-02-18 00:00:00.000000
+
+"""
+
+from alembic import op
+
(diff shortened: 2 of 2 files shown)


src/ai/backend/common/data/permission/types.py
```diff
@@ -308,6 +308,55 @@ class ScopeType(enum.StrEnum):
...
+class RBACNodeType(enum.StrEnum):
+    """Unified RBAC node type.
+
+    Each value represents a node in the RBAC entity relationship graph.
+    The same node can serve as both scope (parent) and entity (child)
+    depending on the edge direction.
+
+    This enum replaces the separate ``ScopeType`` and RBAC-subset of
+    ``EntityType`` with a single unified type.
+    """
```
HyeockJinKim: Is there a reason to provide NodeType in the data?

fregataa: This is not a GraphQL node type - it represents a participant in the RBAC permission hierarchy (scope or entity). I've updated the name to `RBACElementType` to make this clearer.

PR #9156, opened 2026-02-20
feat(BA-4615): Implement `SessionV2` GQL query filtering, ordering, and dataloader

### LINKED ISSUE #9155: Add SessionV2 GQL types implementation
## Motivation

- 

## Required Features

- 

## Impact

- 

## Testing Scenarios

- 

JIRA Issue: BA-4615

jopemachine:
resolves #9155 (BA-4615)

- [ ] Test case(s) to:
- [ ] Documentation

----
📚 Documentation preview 📚: https://sorna--9156.org.readthedocs.build/en/9156/

----
📚 Documentation preview 📚: https://sorna-ko--9156.org.readthedocs.build/ko/9156/

### DIFF
--- a/src/ai/backend/manager/api/gql/session/types.py
+++ b/src/ai/backend/manager/api/gql/session/types.py
@@ -476,16 +476,8 @@ class SessionV2GQL(Node):
     @classmethod
     def from_data(cls, data: SessionData) -> Self:
         """Create SessionV2GQL from SessionData dataclass."""
-        requested_slots = (
-            ResourceSlotGQL.from_resource_slot(data.requested_slots)
-            if data.requested_slots
-            else ResourceSlotGQL(entries=[])
-        )
-        occupying_slots = (
-            ResourceSlotGQL.from_resource_slot(data.occupying_slots)
-            if data.occupying_slots
-            else None
-        )
+        requested_slots = ResourceSlotGQL.from_resource_slot(data.requested_slots)
+        occupying_slots = ResourceSlotGQL.from_resource_slot(data.occupying_slots)
 
         environ_gql: EnvironmentVariablesGQL | None = None
         if data.environ:


jopemachine:
## PR #9156 Security and Performance Review

**Reviewed by**: AI Code Reviewer
**PR**: feat(BA-4615): Implement `SessionV2` GQL query filtering, ordering, and dataloader
**Files analyzed**: 14 changed files (+410/-56)

---

### Finding 1: MEDIUM - Status filter `to_internal()` mapping loses PULLING sessions when filtering by PREPARING

**File**: `/src/ai/backend/manager/api/gql/session/types.py` (lines 86-108)
**Category**: Correctness

The `from_internal()` method correctly maps both `SessionStatus.PREPARING` and `SessionStatus.PULLING` to `SessionV2StatusGQL.PREPARING` (line 66-67). However, the reverse `to_internal()` method only maps `PREPARING` back to `SessionStatus.PREPARING` (line 93-94), not to both statuses.

This means when a user filters by `status: { in: [PREPARING] }`, the generated SQL condition `SessionRow.status.in_([SessionStatus.PREPARING])` will **miss sessions that are in `PULLING` status**, even though those sessions would be displayed as `PREPARING` to the user.

The same issue exists for the `CANCELLED` catch-all: `from_internal` maps `RESTARTING`, `RUNNING_DEGRADED`, and `ERROR` to `CANCELLED`, but `to_internal` for `CANCELLED` only maps back to `SessionStatus.CANCELLED`. Filtering by `status: { in: [CANCELLED] }` will miss `ERROR`, `RESTARTING`, and `RUNNING_DEGRADED` sessions.

**Recommendation**: The `build_condition()` method in `SessionV2StatusFilterGQL` should expand each GQL status to all its corresponding internal statuses. For instance, `PREPARING` should expand to `[SessionStatus.PREPARING, SessionStatus.PULLING]`, and `CANCELLED` should expand to `[SessionStatus.CANCELLED, SessionStatus.ERROR, SessionStatus.RESTARTING, SessionStatus.RUNNING_DEGRADED]`.

---

### Finding 2: MEDIUM - SQL LIKE wildcard characters in user input are not escaped

**File**: `/src/ai/backend/manager/repositories/scheduler/options.py` (lines 55-260)
**Category**: Security / Data Integrity

The `by_name_contains`, `by_name_starts_with`, `by_name_ends_with` methods (and their equivalents for `access_key`, `domain_name`, `scaling_group`) use f-string interpolation to build LIKE patterns:

```python
condition = SessionRow.name.like(f"%{spec.value}%")
```

If `spec.value` contains SQL LIKE wildcard characters (`%`, `_`), they will be interpreted as wildcards rather than literal characters. For example, a user searching for the name `test_session` would match `test1session`, `testXsession`, etc.

While this is not a SQL injection vulnerability (SQLAlchemy parameterizes the value), it is a data integrity issue that can return unintended results.

**Note**: This is a pre-existing pattern in the codebase (also used in kernel conditions), not introduced by this PR. However, since this PR adds new methods using the same pattern, it is worth noting.

**Recommendation**: Escape LIKE wildcards before interpolation. SQLAlchemy supports this natively via the `escape` parameter on `.like()` and `.ilike()` methods.

---

### Finding 3: MEDIUM - `SessionV2StatusFilterGQL.build_condition()` silently ignores `not_in` when `in_` is also provided

**File**: `/src/ai/backend/manager/api/gql/session/types.py` (lines 133-138)
**Category**: Correctness

```python
def build_condition(self) -> QueryCondition | None:
    if self.in_:
        return SessionConditions.by_status_in([s.to_internal() for s in self.in_])
    if self.not_in:
        return SessionConditions.by_status_not_in([s.to_internal() for s in self.not_in])
    return None
```

If a client provides both `in` and `notIn` fields simultaneously, only the `in` filter is applied and `notIn` is silently ignored. This differs from the `KernelV2StatusFilterGQL` which uses a factory pattern with `KernelStatusInMatchSpec` that has an explicit `negated` flag.

While it is reasonable to enforce only one filter at a time, silently ignoring the second creates confusing behavior.

**Recommendation**: Either raise an `InvalidGraphQLParameters` error when both `in_` and `not_in` are provided, or document this behavior explicitly. The same pattern should ideally be applied consistently with the kernel status filter.

---

### Finding 4: MEDIUM - Sensitive data (`environ`, `callback_url`, `bootstrap_script`, `access_key`) exposed through admin-only API without consideration of field-level access control

**File**: `/src/ai/backend/manager/api/gql/session/types.py` (lines 413-477)
**Category**: Security / Data Exposure

The `from_data()` method exposes the following potentially sensitive fields through the GQL API:

1. **`environ`** (line 427-433): Environment variables may contain secrets such as API keys, database passwords, and tokens.
2. **`access_key`** (line 444): The access key used to create the session.
3. **`callback_url`** (line 470): Could contain internal URLs or tokens.
4. **`bootstrap_script`** (line 468): Could contain sensitive configuration.

Currently, this endpoint is admin-only (`check_admin_only()`), which provides a base level of protection. However, if a non-admin session list query is added later (which is common for user-scoped session lists), these fields would need careful access control.

**Recommendation**: Consider whether `environ` entries should be redacted or filtered (e.g., only expose variable names without values for non-admin users). Add a comment/TODO noting that field-level access control must be applied if a non-admin resolver is added. This is consistent with the `hide_agents` pattern seen in `KernelV2GQL.from_kernel_info()`.

---

### Finding 5: LOW - Cursor-based pagination assumes `created_at` ordering but cursor condition only compares `created_at`

**File**: `/src/ai/backend/manager/repositories/scheduler/options.py` (lines 465-492)
**Category**: Correctness / Edge Case

The cursor pagination conditions (`by_cursor_forward` and `by_cursor_backward`) compare `created_at` using a subquery:

```python
def inner() -> sa.sql.expression.ColumnElement[bool]:
    subquery = (
        sa.select(SessionRow.created_at).where(SessionRow.id == cursor_id).scalar_subquery()
    )
    return SessionRow.created_at < subquery
```

If two sessions have exactly the same `created_at` timestamp, cursor pagination may skip or duplicate entries. The tiebreaker order (`SessionRow.id.asc()`) in the `PaginationSpec` helps with ordering, but the cursor **condition** does not include the tiebreaker column. This means:
- Session A and Session B have `created_at = T`
- If the cursor points at Session A, `created_at < T` would exclude **both** A and B from the next page

**Note**: This is the same pattern used by the kernel cursor pagination, so it is a pre-existing architectural issue.

**Recommendation**: For truly robust cursor pagination, consider a composite cursor that includes both `created_at` and `id`, with a compound condition like `(created_at < cursor_created_at) OR (created_at = cursor_created_at AND id > cursor_id)`.

---

### Finding 6: LOW - `resolve_nodes` catches `load_many` errors as `None` values, losing error information from the dataloader

**File**: `/src/ai/backend/manager/api/gql/session/types.py` (lines 400-410)
**Category**: Code Quality

```python
results = await info.context.data_loaders.session_loader.load_many([
    SessionId(UUID(nid)) for nid in node_ids
])
return [cls.from_data(data) if data is not None else None for data in results]
```

Strawberry's `DataLoader.load_many()` can return `Exception` objects in the result list for individual items that failed to load. The current code does not check for `isinstance(data, Exception)` and would pass an `Exception` object to `from_data()`, causing an `AttributeError` at runtime rather than a descriptive error.

**Note**: This follows the same pattern as `KernelV2GQL.resolve_nodes`, so it is consistent with existing code.

**Recommendation**: Add an explicit check for `isinstance(data, Exception)` in the list comprehension to either raise the error or return `None` explicitly.

---

### Finding 7: LOW - `target_resource_groups` and `images` resolvers raise `NotImplementedError` at runtime

**File**: `/src/ai/backend/manager/api/gql/session/types.py` (lines 382-395)
**Category**: Code Quality

Three field resolvers raise `NotImplementedError`:
- `target_resource_groups()` (line 383)
- `images()` (line 389)
- `kernels()` (line 395)

These fields are exposed in the GQL schema and clients can query them, but they will receive a runtime error. This is acceptable for an incremental implementation, but should be tracked.

**Recommendation**: These are noted as not-yet-implemented. Ensure follow-up issues exist for completing these resolvers before the 26.3.0 release. Consider using `@strawberry.field(deprecation_reason="Not yet implemented")` or documenting the limitation in the field descriptions.

---

### Finding 8: LOW - Inconsistent filter pattern: `name` uses plain `str` instead of `StringFilter`

**File**: `/src/ai/backend/manager/api/gql/session/types.py` (line 147)
**Category**: Consistency

The `name` field in `SessionV2FilterGQL` is typed as `str | None`, which only supports exact match filtering. Other V2 GQL types in the codebase (e.g., `NotificationRuleFilterGQL`, `ResourceGroupFilterGQL`, `ArtifactFilterGQL`, `ImageV2FilterGQL`) use `StringFilter | None` for string fields, which supports `contains`, `startsWith`, `endsWith`, `equals`, and their case-insensitive/negated variants.

Despite `SessionConditions` already providing methods for `by_name_contains`, `by_name_starts_with`, and `by_name_ends_with`, only `by_name_equals` is wired up.

**Recommendation**: Consider using `StringFilter` for the `name` field to be consistent with other V2 GQL types and to provide richer query capabilities.

---

### Summary

| # | Severity | Category | Description |
|---|----------|----------|-------------|
| 1 | MEDIUM | Correctness | Status filter `to_internal()` loses `PULLING`/`ERROR`/`RESTARTING`/`RUNNING_DEGRADED` sessions |
| 2 | MEDIUM | Security | LIKE wildcard chars not escaped (pre-existing pattern) |
| 3 | MEDIUM | Correctness | `in_` + `not_in` both set: `not_in` silently ignored |
| 4 | MEDIUM | Security | Sensitive data in `environ`/`access_key` exposed without field-level control |
| 5 | LOW | Correctness | Cursor pagination may skip entries with identical `created_at` (pre-existing) |
| 6 | LOW | Code Quality | `resolve_nodes` does not handle `Exception` from `load_many` |
| 7 | LOW | Code Quality | Three resolvers raise `NotImplementedError` at runtime |
| 8 | LOW | Consistency | `name` uses `str` instead of `StringFilter` unlike other V2 types |

No CRITICAL or HIGH severity issues were found. The implementation follows the established codebase patterns (adapter-based pagination, dataloader registration, condition/order factory pattern) correctly. The most impactful finding is #1 (status filter correctness), which could cause user-visible bugs where filtering by `PREPARING` or `CANCELLED` misses sessions in mapped-to-but-not-mapped-from statuses.

src/ai/backend/manager/api/gql/session/types.py
```diff
@@ -325,12 +468,77 @@ async def resolve_nodes(  # type: ignore[override]  # Strawberry Node uses Await
...
+        requested_slots = (
+            ResourceSlotGQL.from_resource_slot(data.requested_slots)
+            if data.requested_slots
+            else ResourceSlotGQL(entries=[])
+        )
+        occupying_slots = (
+            ResourceSlotGQL.from_resource_slot(data.occupying_slots)
+            if data.occupying_slots
+            else None
+        )
```
seedspirit: Looking at the rows for `requested_slots` and `occupying_slots`, since they are nullable false, it appears that handling for None values needs to be modified.

PR #9240, opened 2026-02-23
fix(BA-4391): Add error handling and prevent FD leak in Agent's `_report_all_kernel_commit_status_map`

### LINKED ISSUE #9239: Fix iterator/generator FD leak in _report_all_kernel_commit_status_map
## What

Fix file descriptor leaks in `_report_all_kernel_commit_status_map()` (`agent.py:1209-1238`).

## Why

`Path.iterdir()` and `Path.glob()` generators are not explicitly closed on exception, leaking 1-2 FDs per invocation (called every 7 seconds).

## How

- Wrap `iterdir()` and `glob()` calls with proper resource management (e.g., `list()` conversion or explicit generator closing)
- Add `base_commit_path.exists()` check before iteration
- Add try/except around the `run_in_executor` call to prevent unhandled exceptions from propagating

JIRA Issue: BA-4391

seedspirit:
resolves #9239 (BA-4391)

- [x] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import enum
+import errno
 import logging
 import pickle
 import re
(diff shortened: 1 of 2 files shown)


seedspirit:
> Is it possible to test any unclosed file descriptor?

@fregataa Since intentionally causing an FD leak carries risks, I additionally wrote a mock test that triggers an error where errno is EMFILE.

src/ai/backend/agent/agent.py
```diff
@@ -1224,19 +1225,33 @@ async def _report_all_kernel_commit_status_map(self, interval: float) -> None:
...
 
-        # Update kernel commit statuses using ValkeyStatClient
-        await self.valkey_stat_client.update_kernel_commit_statuses(
-            list(commit_kernels),
-            COMMIT_STATUS_EXPIRE,
-        )
+            log.exception("error while scanning kernel commit statuses")
+            return
+        except Exception:
+            log.exception("unexpected error in commit status reporting")
```
fregataa, line 1253: Let's log the detail of original error

PR #9256, opened 2026-02-23
fix(BA-4653): Remove unnecessary locks from StatContext

fregataa:
resolves BA-4653

## Summary
- Remove the unnecessary `asyncio.Lock` from `StatContext` entirely
- Self-concurrency is already prevented by `TimerDelayPolicy.CANCEL` which cancels the previous invocation before starting a new one
- Each collect method (`collect_node_stat`, `collect_container_stat`, `collect_per_container_process_stat`) operates on independent data structures, so cross-method locking is also unnecessary

## Test plan
- [ ] Verify agent stat collection runs without errors
- [ ] Confirm node, container, and process stat collection proceeds without blocking each other

### DIFF
--- a/changes/9256.fix.md
+++ b/changes/9256.fix.md
@@ -1,1 +1,1 @@
-Split the single shared `asyncio.Lock` in `StatContext` into three per-scope locks to allow concurrent node, container, and process stat collection
+Remove the unnecessary `asyncio.Lock` from `StatContext` as self-concurrency is already prevented by `TimerDelayPolicy.CANCEL` and each collect method operates on independent data structures


src/ai/backend/agent/stats.py
```diff
@@ -361,7 +361,9 @@ def __init__(
         self.kernel_metrics = {}
         self.process_metrics = {}
 
-        self._lock = asyncio.Lock()
+        self._node_lock = asyncio.Lock()
+        self._container_lock = asyncio.Lock()
+        self._process_lock = asyncio.Lock()
```
seedspirit: In the PR below, `TimerDelayPolicy.CANCEL` is applied, so if any callback is still running, it cancels and restarts it. This already prevents the same method from overlapping with itself during execution. 

https://github.com/lablup/backend.ai/pull/9257

Since these three do not share any overlapping fields, it appears there is no longer a need to acquire a lock.

PR #9262, opened 2026-02-23
feat(BA-4655): Add integrity_error_checks to RBAC entity creators

fregataa:
## Summary
- Add `integrity_error_checks` support to `execute_rbac_entity_creator()`, `execute_rbac_bulk_entity_creator()`, and `execute_rbac_entity_creators()` in `entity_creator.py`
- Wraps `flush()` calls with `try/except sa.exc.IntegrityError` handling, mirroring the existing pattern from the base `creator.py` module
- Imports `parse_integrity_error` and `_match_integrity_error` from the base integrity module

resolves BA-4655

## Test plan
- [x] Verify that `CreatorSpec.integrity_error_checks` are properly matched when an `IntegrityError` occurs during RBAC entity creation
- [x] Verify that unmatched integrity errors still raise `RepositoryIntegrityError` as fallback
- [x] Verify no regression in normal (no-error) entity creation flows

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/ai/backend/manager/repositories/base/rbac/entity_creator.py
+++ b/src/ai/backend/manager/repositories/base/rbac/entity_creator.py
@@ -19,7 +19,7 @@ from ai.backend.manager.models.rbac_models.association_scopes_entities import (
 )
 from ai.backend.manager.repositories.base.creator import CreatorSpec
 from ai.backend.manager.repositories.base.integrity import (
-    _match_integrity_error,
+    match_integrity_error,
     parse_integrity_error,
 )
 
@@ -99,7 +99,7 @@ async def execute_rbac_entity_creator[TRow: Base](
         await db_sess.flush()
     except sa.exc.IntegrityError as e:
         parsed = parse_integrity_error(e)
-        _match_integrity_error(parsed, spec.integrity_error_checks)
+        match_integrity_error(parsed, spec.integrity_error_checks)
 
     # 3. Extract RBAC info and insert associations for all scopes
     instance_state = inspect(row)
@@ -188,7 +188,7 @@ async def execute_rbac_bulk_entity_creator[TRow: Base](
         parsed = parse_integrity_error(e)
         # Use first spec's checks (all specs share the same CreatorSpec subclass)
         checks = creator.specs[0].integrity_error_checks
-        _match_integrity_error(parsed, checks)
+        match_integrity_error(parsed, checks)
 
     entity_type = creator.element_type.to_entity_type()
     associations = [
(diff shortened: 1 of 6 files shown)


src/ai/backend/manager/repositories/base/rbac/entity_creator.py
```diff
@@ -17,6 +18,10 @@
     AssociationScopesEntitiesRow,
 )
 from ai.backend.manager.repositories.base.creator import CreatorSpec
+from ai.backend.manager.repositories.base.integrity import (
+    _match_integrity_error,
```
jopemachine: It doesn’t appear to be a private function, so let’s remove the leading underscore.

src/ai/backend/manager/repositories/base/rbac/entity_creator.py
```diff
@@ -173,7 +182,13 @@ async def execute_rbac_bulk_entity_creator[TRow: Base](
         )
 
     # 2. Flush to get DB-generated IDs and insert associations
-    await db_sess.flush()
+    try:
+        await db_sess.flush()
+    except sa.exc.IntegrityError as e:
+        parsed = parse_integrity_error(e)
+        # Use first spec's checks (all specs share the same CreatorSpec subclass)
+        checks = creator.specs[0].integrity_error_checks
```
jopemachine, line 190: Shouldn't the fact that all specs share the same CreatorSpec also be reflected in the code?
(for example, by validating it and logging an error if it doesn't hold)

src/ai/backend/manager/repositories/base/rbac/entity_creator.py
fregataa: Done — renamed `_match_integrity_error` → `match_integrity_error`

src/ai/backend/manager/repositories/base/rbac/entity_creator.py
fregataa: This is the same pattern used in the base `execute_bulk_creator()` (creator.py:198-199):

```python
# Use first spec's checks (all specs share the same CreatorSpec subclass)
checks = bulk_creator.specs[0].integrity_error_checks
```

`integrity_error_checks` maps table constraints to domain errors, so it is determined by the Row type, not per-instance data. Since `BulkCreator[TRow]` / `RBACBulkEntityCreator[TRow]` already constrains all specs to the same Row type, they naturally use the same `CreatorSpec` subclass.

Also, in the bulk flush case, SQLAlchemy raises a single `IntegrityError` without identifying which row caused it, so per-spec matching is not feasible anyway. For cases that need per-spec error handling, `execute_bulk_creator_partial` already exists (individual flush + per-spec checks).

Adding a runtime validation here would be inconsistent with the base creator and adds complexity without practical benefit. I think keeping the comment-documented assumption is sufficient.

PR #9295, opened 2026-02-24
fix(BA-4665): alias MAIN_KERNEL_JOIN to avoid duplicate unaliased table join

HyeockJinKim:
## Summary
- When exporting sessions to CSV with fields requiring both `MAIN_KERNEL_JOIN` (e.g. `main_kernel_image`) and `KERNEL_JOIN` (e.g. `kernel_agent`), the query joined the `kernels` table twice without aliasing, causing PostgreSQL to return empty data rows
- Introduce a SQLAlchemy alias (`main_kernel`) for `MAIN_KERNEL_JOIN` and update the four `main_kernel_*` field column references to use the aliased table
- Widen `ExportFieldDef.column` type from `InstrumentedAttribute` to `sa.ColumnElement` to support aliased column references

## Test plan
- [x] Add regression test verifying aliased + unaliased JOINs to the same table produce valid SQL
- [x] Add regression test using real `SESSION_REPORT` definition with both `main_kernel_image` and `kernel_agent` fields
- [x] All existing adapter tests pass
- [x] `pants fmt`, `fix`, `lint` pass

Resolves BA-4665

### DIFF
--- a/changes/9295.fix.md
+++ b/changes/9295.fix.md
@@ -1,0 +1,1 @@
+Fix session CSV export returning empty data rows when both main_kernel and kernel fields are requested by aliasing the MAIN_KERNEL_JOIN table


HyeockJinKim:
Closing in favor of a simpler fix: remove main_kernel_* fields entirely and add missing kernel_architecture/registry/tag to kernel_* fields instead of aliasing.

PR #9338, opened 2026-02-25
feat(BA-4692): Add deployment policy create/update service logic and REST API

### LINKED ISSUE #9333: Implement REST API for create_deployment_policy / update_deployment_policy
## Objective

Implement REST API endpoints for create_deployment_policy and update_deployment_policy operations. These REST endpoints will expose the service-layer methods implemented in BA-4632 via the REST API interface.

## Background

BA-4632 implements the service layer and GraphQL mutations for create_deployment_policy / update_deployment_policy. This issue covers the corresponding REST API endpoints, following the project's API layer conventions.

## Acceptance Criteria

- REST API endpoint for create_deployment_policy is implemented
- REST API endpoint for update_deployment_policy is implemented
- Proper request/response schema validation
- Unit tests cover the new REST endpoints

## Related

- BEP-1049: Zero-downtime Deployment Strategy Architecture
- BA-4632: Add create_deployment_policy, update_deployment_policy to service layer
- Epic: BA-3419

JIRA Issue: BA-4692

jopemachine:
Resolves #9333 (BA-4692).

## Summary
- Add action/action-result pairs (`CreateDeploymentPolicyAction`, `UpdateDeploymentPolicyAction`) with service layer methods and processor wiring
- Add REST API endpoints: `POST /{deployment_id}/policy`, `PATCH /{deployment_id}/policy`, `GET /{deployment_id}/policy`
- Add DTOs (`CreateDeploymentPolicyRequest`, `UpdateDeploymentPolicyRequest`, `DeploymentPolicyDTO`) and adapter
- Add unit tests for create and update deployment policy service operations

## Test plan
- [x] Unit tests for `TestCreateDeploymentPolicy` (rolling + blue_green strategies)
- [x] Unit tests for `TestUpdateDeploymentPolicy` (full update + partial update)
- [ ] `pants lint --changed-since=HEAD~1`
- [ ] `pants test tests/unit/manager/services/deployment/test_deployment_service.py`

----
📚 Documentation preview 📚: https://sorna--9338.org.readthedocs.build/en/9338/

----
📚 Documentation preview 📚: https://sorna-ko--9338.org.readthedocs.build/ko/9338/

### DIFF
--- a/src/ai/backend/manager/api/rest/deployment/adapter.py
+++ b/src/ai/backend/manager/api/rest/deployment/adapter.py
@@ -52,6 +52,7 @@ from ai.backend.common.dto.manager.deployment.types import (
 from ai.backend.common.types import ClusterMode, RuntimeVariant
 from ai.backend.manager.api.adapter import BaseFilterAdapter
 from ai.backend.manager.data.deployment.creator import (
+    DeploymentPolicyConfig,
     DeploymentPolicyCreator,
     ModelRevisionCreator,
     NewDeploymentCreator,
@@ -482,8 +483,8 @@ class CreateDeploymentAdapter:
     def _build_policy_config(
         self,
         strategy_input: Any,  # DeploymentStrategyInput
-    ) -> DeploymentPolicyCreator:
-        """Build DeploymentPolicyCreator from strategy input."""
+    ) -> DeploymentPolicyConfig:
+        """Build DeploymentPolicyConfig from strategy input."""
         strategy = DeploymentStrategy(strategy_input.type)
 
         strategy_spec: RollingUpdateSpec | BlueGreenSpec
@@ -505,7 +506,7 @@ class CreateDeploymentAdapter:
                         promote_delay_seconds=strategy_input.blue_green.promote_delay_seconds,
                     )
 
-        return DeploymentPolicyCreator(
+        return DeploymentPolicyConfig(
             strategy=strategy,
             strategy_spec=strategy_spec,
             rollback_on_failure=strategy_input.rollback_on_failure,


src/ai/backend/manager/api/deployment/handler.py
```diff
@@ -409,6 +427,87 @@ async def update_route_traffic_status(
...
+            spec=updater_spec,
+            pk_value=policy_result.data.id,
+        )
+
+        action_result = await deployment_processors.update_deployment_policy.wait_for_complete(
+            UpdateDeploymentPolicyAction(
+                policy_id=policy_result.data.id,
+                updater=updater,
+            )
+        )
```
seedspirit: Wouldn't it be possible to pass the `deployment_id` to `UpdateDeploymentPolicyAction` and have it look up the policy in the database where the deployment id matches the id passed in the path param? Is there a specific reason why two round trips are absolutely necessary?

src/ai/backend/manager/services/deployment/actions/deployment_policy/create_deployment_policy.py
```diff
@@ -0,0 +1,41 @@
...
+from ai.backend.manager.services.deployment.actions.deployment_policy.base import (
+    DeploymentPolicyBaseAction,
+)
+
+
+@dataclass
+class CreateDeploymentPolicyAction(DeploymentPolicyBaseAction):
+    """Action to create a deployment policy for an endpoint."""
+
+    endpoint_id: UUID
```
seedspirit: How about
```suggestion
    deployment_id: UUID
```

src/ai/backend/manager/services/deployment/service.py
```diff
@@ -474,6 +482,45 @@ async def search_deployment_policies(
...
+            CreateDeploymentPolicyActionResult: Result containing the created policy data
+        """
+        config = action.policy_config
+        spec = DeploymentPolicyCreatorSpec(
+            endpoint_id=action.endpoint_id,
+            strategy=config.strategy,
+            strategy_spec=config.strategy_spec,
+            rollback_on_failure=config.rollback_on_failure,
+        )
+        creator: Creator[DeploymentPolicyRow] = Creator(spec=spec)
```
seedspirit: I'm asking because I don't quite remember. Is the reason for generating specs in the Service Layer due to dependencies with the repository layer? (I was wondering if it could be done in the API Layer if it's just simple assembly code.)

seedspirit: Updator operates at the API layer through an adapter. Shouldn't Creator also consistently operate at the API layer?

seedspirit (changes requested):

src/ai/backend/manager/api/deployment/handler.py
jopemachine: Since UpdateDeploymentPolicyAction modifies the DeploymentPolicy entity, it is natural to perform the update using the DeploymentPolicy ID. If optimization is necessary, it would be preferable to introduce a separate action or an explicit option rather than overloading the existing one.

src/ai/backend/manager/services/deployment/service.py
jopemachine: > I'm asking because I don't quite remember. Is the reason for generating specs in the Service Layer due to dependencies with the repository layer? (I was wondering if it could be done in the API Layer if it's just simple assembly code.)

Yes, I think so, if CreatorSpec have to be constructed by the caller, it would introduce a dependency on the repository module. So, it seems more appropriate to define a new type and pass that instead.

src/ai/backend/manager/services/deployment/service.py
```diff
@@ -474,6 +484,60 @@ async def search_deployment_policies(
...
+
+        Args:
+            action: Action containing the creator for the policy
+
+        Returns:
+            CreateDeploymentPolicyActionResult: Result containing the created policy data
+        """
+        creator = action.creator
+        deployment_id = creator.deployment_id
+        if deployment_id is None:
```
seedspirit: If `deployment_id` is essential, would it not be preferable to change the type hinting for the creator's `deployment_id` from `UUID | None` to simply UUID? Then validate it during the API layer transformation using pydantic or logic?

seedspirit (dismissed):

PR #9347, opened 2026-02-25
feat(BA-4657): apply RBAC Creator pattern to Keypair

fregataa:
## Summary
- Add `KEYPAIR` to `EntityType` enum (needed for `RBACElementType.to_entity_type()` bridge)
- Create `KeyPairCreatorSpec` implementing `CreatorSpec[KeyPairRow]`
- Replace raw keypair insert in `UserDBSource` with `RBACEntityCreator` (both single and bulk create paths)
- Scope: `USER` (owning user)

## Test plan
- [x] Verify keypair creation during user creation produces RBAC scope association
- [x] Verify bulk user creation keypair path works correctly


changes/9347.feature.md
```diff
@@ -0,0 +1 @@
+Apply RBAC Creator pattern to Keypair entity for scope-entity association.
```
jopemachine, line 1: Just a question.

As far as I know, there used to be a rule that a news fragment should not end with a period. Is that rule no longer applied?

PR #9461, opened 2026-02-27
fix(BA-1146): use plain JSON body in update_password_no_auth

fregataa:
## Summary
- Replace `_put_secure_body(rqst, body)` with `rqst.set_json(body)` in SDK's `Auth.update_password_no_auth()` — this request goes directly to the Manager which has no decrypt middleware, so encrypting the body causes "Malformed body" (400)

## Test plan
- [ ] Test expired password update on HTTP endpoint (browser flow)
- [x] Run `test_auth.py` unit tests

Resolves BA-1146

### DIFF
--- a/src/ai/backend/web/server.py
+++ b/src/ai/backend/web/server.py
@@ -39,7 +39,6 @@ from setproctitle import setproctitle
 
 from ai.backend.client.config import APIConfig
 from ai.backend.client.exceptions import BackendAPIError, BackendClientError
-from ai.backend.client.request import Request
 from ai.backend.client.session import AsyncSession as APISession
 from ai.backend.common import config
 from ai.backend.common.clients.http_client.client_pool import ClientPool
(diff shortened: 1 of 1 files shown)


src/ai/backend/client/func/auth.py
```diff
@@ -104,7 +104,7 @@ async def update_password_no_auth(
             "current_password": current_password,
             "new_password": new_password,
         }
-        _put_secure_body(rqst, body)
+        rqst.set_json(body)
```
HyeockJinKim, line 107: Is modifying only the client SDK really the solution to this issue? I thought we were testing in the WebUI environment. Please let me know if I'm mistaken.

fregataa: Webservers uses SDK to communicate with a Manager

fregataa:
Closed this PR since #9478 will resolve the issue

PR #9462, opened 2026-02-27
docs: [WIP] BEP-1051 Kata Containers Agent Backend

kyujin-cho:
## Summary

- Propose KataAgent as a third `AbstractAgent` backend (alongside DockerAgent and KubernetesAgent)
- VFIO GPU passthrough for hardware-enforced workload isolation in multi-tenant GPU environments
- Segmented BEP with main document + 7 sub-documents

## Documents

| Document | Scope |
|----------|-------|
| `BEP-1049-kata-containers-agent.md` | Main: motivation, design overview, implementation plan, decision log |
| `BEP-1049/kata-agent-backend.md` | KataAgent, KataKernel, KataKernelCreationContext classes |
| `BEP-1049/configuration-deployment.md` | `[kata]` config section, hypervisor selection, host requirements |
| `BEP-1049/storage-compatibility.md` | virtio-fs mount translation, intrinsic mount evaluation, direct storage access analysis |
| `BEP-1049/networking.md` | Calico CNI integration, inter-VM networking, network policy |
| `BEP-1049/vfio-accelerator-plugin.md` | CUDAVFIOPlugin, IOMMU group detection, device passthrough |
| `BEP-1049/scheduler-integration.md` | AgentRow.backend column, scaling group policy, VM overhead |
| `BEP-1049/migration-compatibility.md` | Additive rollout, backward compatibility, rollback plan |

## Key Decisions

1. Full `AbstractAgent` backend, not a Docker RuntimeClass flag
2. VFIO-only for GPU (whole-device passthrough, no fractional)
3. Separate `CUDAVFIOPlugin` (not a mode on existing CUDAPlugin)
4. Homogeneous scaling groups (one backend type per group)
5. virtio-fs for all storage (no new storage interface)
6. Calico CNI for multi-host networking and session isolation
7. VM memory overhead as agent-level deduction from `available_slots`
8. Skip lxcfs and libbaihook.so for Kata (guest kernel provides accurate /proc, /sys, sysconf natively)
9. Skip jail ptrace sandbox for Kata (VM boundary is stronger isolation)
10. Agent socket via TCP instead of Unix domain socket (UDS cannot cross VM boundary)

## Status

Work in progress — open for early review and feedback.

### DIFF
--- a/proposals/BEP-1049-kata-containers-agent.md
+++ b/proposals/BEP-1049-kata-containers-agent.md
@@ -9,22 +9,32 @@ Implemented-Version:
 
 <!-- context-for-ai
 type: master-bep
-scope: Add Kata Containers as a third container backend (KataAgent) with VFIO GPU passthrough
+scope: Add Kata Containers as a third container backend (KataAgent) with CoCo by default, VFIO GPU passthrough, GPUDirect RDMA, and direct guest-side storage mounts
 detail-docs: [configuration-deployment.md, kata-agent-backend.md, storage-compatibility.md, networking.md, vfio-accelerator-plugin.md, scheduler-integration.md, migration-compatibility.md]
 key-constraints:
   - Full AbstractAgent implementation, NOT a RuntimeClass annotation
+  - CoCo (Confidential Containers) by default — host is always untrusted
   - Whole-GPU VFIO passthrough only, NO fractional GPU
   - DiscretePropertyAllocMap for cuda.device slot type
   - Must not break existing DockerAgent or KubernetesAgent deployments
+  - VFolder storage via direct guest-side NFS/Lustre/WekaFS mount (native kernel client, preserves RDMA), NOT virtio-fs
+  - Guest-side image pull via image-rs (host must not see image contents)
+  - Monolithic attested guest rootfs with all drivers baked in
 key-decisions:
   - KataAgent as independent backend, not a Docker runtime flag
+  - CoCo by default; no non-CoCo Kata mode
   - VFIO passthrough for GPU, not nvidia-container-runtime in guest
   - New CUDAVFIOPlugin separate from existing CUDAPlugin
+  - RDMAVFIOPlugin as non-schedulable co-plugin (auto-attaches co-located InfiniBand HCA)
   - Homogeneous scaling groups (one backend type per group)
   - VM memory overhead as agent-level deduction from available_slots
-  - No new storage interface; virtio-fs is the transparent compatibility layer for bind mounts
+  - VFolder data via direct guest-side NFS/Lustre mount (native, RDMA-preserving); virtio-fs only for scratch/config
+  - Storage topology visible in guest /proc/mounts (acceptable — same model as Docker; isolation via vfolder permissions, not mount concealment)
   - Calico CNI for inter-VM networking (multi-host, network policy); /etc/hosts for hostname resolution
-implementation: 4 phases
+  - DCGM Exporter + Node Exporter for metrics (containerd exec blocked by CoCo policy)
+  - Agent-side volume mount config delivered via virtio-fs (storage-mounts.json in /home/config)
+  - MNNVL (NVL72) deferred to separate BEP
+implementation: 3 phases
 -->
 
 # Kata Containers Agent Backend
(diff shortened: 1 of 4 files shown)


proposals/BEP-1049/configuration-deployment.md
```diff
@@ -0,0 +1,206 @@
...
+
+# --- VM Defaults ---
+default-vcpus = 2               # Initial vCPUs per VM (hot-plugged as needed)
+default-memory-mb = 2048        # Initial memory per VM in MB
+vm-overhead-mb = 64             # Per-VM memory overhead deducted from host capacity
+
+# --- Guest Image ---
+kernel-path = "/opt/kata/share/kata-containers/vmlinux.container"
+initrd-path = ""                # Empty = use rootfs image instead of initrd
+rootfs-path = "/opt/kata/share/kata-containers/kata-containers.img"
```
kyujin-cho, line 59: How are we going to integrate Backend.AI's container image management architecture here? Based on the configuration example here it seems that current kata container architecture only supports pre-configured image only

proposals/BEP-1049/kata-agent-backend.md
```diff
@@ -0,0 +1,295 @@
...
+
+```python
+async def prepare_scratch(self):
+    # Same directory structure as Docker, but these will be
+    # mounted into the guest VM via virtio-fs instead of bind mounts
+    scratch_dir = self.scratch_root / str(self.kernel_id)
+    scratch_dir.mkdir(parents=True, exist_ok=True)
+    # Write environ.txt, resource.txt, etc. into scratch_dir
+    # These become visible inside the guest at the virtio-fs mount point
+```
```
kyujin-cho, line 335: Given that VM hypervisor guarantees full resource isolation support, do we still require a separate resource definition file to be provisioned beforehands?

proposals/BEP-1049/kata-agent-backend.md
```diff
@@ -0,0 +1,295 @@
...
+**`prepare_scratch()`** — Create directories that will be shared via virtio-fs:
+
+```python
+async def prepare_scratch(self):
+    # Same directory structure as Docker, but these will be
+    # mounted into the guest VM via virtio-fs instead of bind mounts
+    scratch_dir = self.scratch_root / str(self.kernel_id)
+    scratch_dir.mkdir(parents=True, exist_ok=True)
+    # Write environ.txt, resource.txt, etc. into scratch_dir
+    # These become visible inside the guest at the virtio-fs mount point
```
kyujin-cho: We already create a boot disk for each VM with designated capacity limits. Do we still require scratch folders for VM based containers?

proposals/BEP-1049/networking.md
```diff
@@ -0,0 +1,345 @@
...
+  └────┬─────┘                                      │
+───────┼────────────────────────────────────────────┼──── VM boundary (KVM)
+       │  virtio-net                                │
+  ┌────┴─────┐                                      │
+  │   eth0   │  (guest interface, same IP as host   │
+  │          │   veth — transparent to CNI)          │
+  └──────────┘                                      │
+```
+
+This TC filter mechanism is Kata's default and requires no additional configuration. The guest VM's `eth0` gets the same IP address that the CNI assigned to the host-side veth, making the VM transparent to the network infrastructure.
```
kyujin-cho, line 54: Then how can kata container handle situations where multiple containers running simultaneously on single hypervisor?

proposals/BEP-1049/configuration-deployment.md
kyujin-cho: Kata uses two separate filesystem layers — the guest VM rootfs (`kata-containers.img`) is just a minimal mini-OS (kata-agent + systemd) for booting the VM. The user-selected container image (e.g., `tensorflow:latest`) flows through the standard containerd pull path on the host, and the Kata shim mounts it into the guest via virtio-fs (overlayfs snapshotter) or block device passthrough (devicemapper snapshotter).

No changes to Backend.AI's image management are needed — the image registry, selection, and pull flow are identical to Docker. Added a "Guest VM Image vs Container Image" section to configuration-deployment.md clarifying this two-layer architecture.

proposals/BEP-1049/kata-agent-backend.md
kyujin-cho: Yes, scratch directories are still required. The VM boot disk (`kata-containers.img`) is a read-only, shared mini-OS containing only the kata-agent — it is not per-session storage and has no user workspace.

Scratch dirs serve different purposes:
- `/home/config`: Agent-written config files consumed by the kernel runner
- `/home/work`: User workspace and vfolder mount point

These are created on the host and shared into the guest via virtio-fs, just as they are bind-mounted in Docker.

Updated the `prepare_scratch()` section with this clarification.

proposals/BEP-1049/kata-agent-backend.md
kyujin-cho: The VM hypervisor enforces resource **limits** (vCPU count, memory cap), but `resource.txt` and `environ.txt` serve a different purpose — they communicate resource **metadata** to the kernel runner application. The kernel runner reads these to:

- Display allocated resources to the user (e.g., `nvidia-smi`-like info, CPU/memory allocation in the session UI)
- Set environment variables (`CUDA_VISIBLE_DEVICES`, `BACKENDAI_KERNEL_ID`, etc.)
- Configure accelerator device mappings
- Provide SSH keys and credentials

The VM does not know about these application-level concerns. Updated the doc to clarify this distinction.

proposals/BEP-1049/networking.md
kyujin-cho: Kata uses a **sandbox model**: one sandbox = one VM. A single kata-agent inside the guest manages multiple containers in the same VM via ttRPC `CreateContainer` calls. All containers in the same sandbox share the VM's network namespace (communicate via `localhost`), consistent with Kubernetes pod semantics.

For Backend.AI multi-container sessions:
- **Single-host cluster**: All containers can share one Kata sandbox (one VM), communicating via localhost
- **Multi-host cluster** (distributed training): Each host runs a separate VM; inter-VM traffic flows through Calico CNI (BGP/VXLAN)

Added a "Sandbox Model: Multiple Containers Per VM" section to networking.md explaining this.

proposals/BEP-1049/kata-agent-backend.md
kyujin-cho: Follow-up: I've added a full analysis of the per-VM cloned disk approach to [storage-compatibility.md — Per-VM Cloned Disk: Hybrid Storage Model](https://github.com/lablup/backend.ai/blob/docs/bep-1049-kata-containers-agent/proposals/BEP-1049/storage-compatibility.md#per-vm-cloned-disk-hybrid-storage-model).

**Key findings:**

The virtio-fs-for-everything model creates 20-30+ `virtiofsd` processes per VM (~300-450 MB RSS overhead). A conventional hypervisor-style approach (qcow2 CoW cloning per VM) can eliminate most of these, but there's a critical constraint: **block devices are guest-local** — the host cannot write to them after VM boot. This breaks Backend.AI's bidirectional config injection flow (agent writes `environ.txt` / `resource.txt` → kernel runner reads them inside the guest).

**Recommended hybrid model:**
- **Block devices** (virtio-blk / devicemapper / EROFS) for read-only infrastructure: container OCI image, krunner binaries, Python libraries, timezone
- **virtio-fs** only for bidirectional data exchange: scratch dirs (`/home/config`, `/home/work`) and vfolders

This reduces virtiofsd count from 20-30+ to **2 + N_vfolders** per VM, saving ~270-375 MB RSS per VM. Phase 1 starts with virtio-fs for everything (simpler), transitioning to the hybrid model in Phase 2.

PR #9469, opened 2026-02-27
fix(BA-4392): Share single Docker client across container stats collection

### LINKED ISSUE #8809: Add connection pooling for Docker stats collection
## What

Refactor Docker stats collection in `docker/intrinsic.py` to reuse Docker client connections and limit concurrency.

## Why

Currently, each container stat query creates a new `Docker()` instance with its own `aiohttp.ClientSession`, causing 200-300+ FDs to be opened simultaneously for large deployments (called every 5 seconds).

## How

- Reuse a shared `Docker()` client instance across stat queries instead of creating a new one per container
- Apply to both `CPUPlugin` (intrinsic.py:275-286) and `MemPlugin` (intrinsic.py:727-761)
- Ensure proper cleanup of the shared Docker client on agent shutdown

JIRA Issue: BA-4392

seedspirit:
resolves #8809 (BA-4392)

- [x] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/tests/unit/agent/test_docker_intrinsic.py
+++ b/tests/unit/agent/test_docker_intrinsic.py
@@ -93,23 +93,19 @@ class TestCPUPluginDockerClientLifecycle(BaseDockerIntrinsicTest):
         cgroup_stat_context.agent.get_cgroup_path = mock_get_cgroup_path
         return cgroup_stat_context
 
-    async def test_init_creates_docker_client(self) -> None:
+    async def test_init_creates_docker_client(self, cpu_plugin: CPUPlugin) -> None:
         """Verify init() creates a Docker client instance."""
-        plugin = CPUPlugin.__new__(CPUPlugin)
-        plugin.local_config = {"agent": {"docker-mode": "default"}}
         with patch("ai.backend.agent.docker.intrinsic.Docker") as mock_docker_cls:
             mock_docker_cls.return_value = AsyncMock()
-            await plugin.init()
+            await cpu_plugin.init()
             mock_docker_cls.assert_called_once()
-            assert plugin._docker is not None
+            assert cpu_plugin._docker is not None
 
-    async def test_cleanup_closes_docker_client(self) -> None:
+    async def test_cleanup_closes_docker_client(self, cpu_plugin: CPUPlugin) -> None:
         """Verify cleanup() closes the Docker client."""
-        plugin = CPUPlugin.__new__(CPUPlugin)
-        plugin.local_config = {"agent": {"docker-mode": "default"}}
-        plugin._docker = AsyncMock()
-        await plugin.cleanup()
-        plugin._docker.close.assert_called_once()
+        cpu_plugin._docker = AsyncMock()
+        await cpu_plugin.cleanup()
+        cpu_plugin._docker.close.assert_called_once()
 
     async def test_api_mode_uses_instance_docker_client(
         self,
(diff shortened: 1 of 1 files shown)


src/ai/backend/agent/docker/intrinsic.py
```diff
@@ -724,41 +726,42 @@ async def sysfs_impl(
...
-                    mem_total_bytes,
-                    io_read_bytes,
-                    io_write_bytes,
-                    net_rx_bytes,
-                    net_tx_bytes,
-                    scratch_sz,
-                )
+        async def api_impl(
+            container_id: str,
+        ) -> tuple[int, int, int, int, int, int, int] | None:
```
fregataa, line 738: Can we change the return type to dataclass?

src/ai/backend/agent/docker/intrinsic.py
```diff
@@ -773,10 +776,11 @@ async def api_impl(container_id: str) -> tuple[int, int, int, int, int, int, int
         per_container_net_rx_bytes = {}
         per_container_net_tx_bytes = {}
         per_container_io_scratch_size = {}
-        tasks = []
-        for cid in container_ids:
-            tasks.append(asyncio.create_task(impl(cid)))
-        results = await asyncio.gather(*tasks)
+        async with closing_async(Docker()) as docker:
```
fregataa: Is the `docker` instance passed to impl() coroutine correctly?

src/ai/backend/agent/docker/intrinsic.py
```diff
@@ -272,33 +272,36 @@ async def sysfs_impl(container_id: str) -> float | None:
...
         if ctx.mode == StatModes.CGROUP:
-            impl = sysfs_impl
+            tasks = []
+            for cid in container_ids:
+                tasks.append(asyncio.create_task(sysfs_impl(cid)))
+            results = await asyncio.gather(*tasks)
         elif ctx.mode == StatModes.DOCKER:
-            impl = api_impl
+            async with closing_async(Docker()) as docker:
+
```
HyeockJinKim: Does this still seem to be generated every time?

src/ai/backend/agent/docker/intrinsic.py
seedspirit: I'll leave this for a follow-up [issue](https://lablup.atlassian.net/browse/BA-4761)

src/ai/backend/agent/docker/intrinsic.py
seedspirit: `docker` is captured by the closure. Since `api_impl` is defined within the block `async with closing_async(Docker()) as docker:`, it references the `docker` variable from the enclosing scope. Python closures are late-bound, meaning they resolve at function call time. By the time `api_impl` is called, the `docker` instance has already been created, so it is shared correctly.

src/ai/backend/agent/docker/intrinsic.py
seedspirit: Yes, it is created per tick. However, previously, a `Docker()` instance was created per container (N instances/tick), which caused an FD explosion issue when the number of containers increased. This change improves the system by creating only one per tick and having all containers share it. Reusing it between ticks requires additional work due to the stateless callback structure of `aiotools.create_timer`, so it's excluded from the scope of this PR.

fregataa: StatContext isn't built every tick

seedspirit: @fregataa It meant that a Docker instance was created each time.

PR #9517, opened 2026-02-27
feat(BA-4792): implement assignments_by_role_loader DataLoader

fregataa:
## Summary
- Add DataLoader for batch loading role assignments by role_id
- Prevents N+1 queries when loading users assigned to roles in GraphQL

## Changes
- Add `AssignedUserConditions.by_role_ids()` query condition method
- Implement `load_assignments_by_role_ids()` DataLoader function  
- Register `assignments_by_role_loader` in DataLoaders class
- Export new function in rbac module

## JIRA Issues
- BA-4780: role_assignment_loader ✓ (already implemented)
- BA-4792: assignments_by_role_loader ✓ (new implementation)

## Related
Part of BA-4775 (RBAC GraphQL Type Enhancement)

## Test Plan
- [x] Quality checks pass (fmt, fix, lint, check)
- [x] Pre-commit hooks pass

### DIFF
--- a/tests/unit/manager/api/rbac/test_dataloader.py
+++ b/tests/unit/manager/api/rbac/test_dataloader.py
@@ -1,0 +1,125 @@
+"""Tests for RBAC GraphQL DataLoader utilities."""
+
+from __future__ import annotations
+
+import uuid
+from unittest.mock import AsyncMock, MagicMock
+
+from ai.backend.manager.api.gql.data_loader.rbac.loader import (
+    load_assignments_by_role_ids,
+)
+from ai.backend.manager.data.permission.role import AssignedUserData
+
+
+class TestLoadAssignmentsByRoleIds:
+    """Tests for load_assignments_by_role_ids function."""
+
+    @staticmethod
+    def create_mock_assignment(role_id: uuid.UUID, user_id: uuid.UUID) -> MagicMock:
+        mock = MagicMock(spec=AssignedUserData)
+        mock.role_id = role_id
+        mock.user_id = user_id
+        return mock
+
+    @staticmethod
+    def create_mock_processor(assignments: list[MagicMock]) -> MagicMock:
+        mock_processor = MagicMock()
+        mock_action_result = MagicMock()
+        mock_action_result.result = MagicMock()
+        mock_action_result.result.items = assignments
+        mock_processor.search_users_assigned_to_role.wait_for_complete = AsyncMock(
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/api/gql/data_loader/rbac/loader.py
```diff
@@ -117,6 +118,37 @@ async def load_role_assignments_by_role_and_user_ids(
...
+        SearchUsersAssignedToRoleAction(querier=querier)
+    )
+
+    # Group assignments by role_id
+    assignment_map: dict[uuid.UUID, list[AssignedUserData]] = defaultdict(list)
+    for assignment in action_result.result.items:
+        assignment_map[assignment.role_id].append(assignment)
+
+    # Return in the same order as role_ids, with empty lists for roles with no assignments
+    return [assignment_map.get(role_id, []) for role_id in role_ids]
```
HyeockJinKim, line 149: Wouldn't it be better to return None if the value is not present?

fregataa (commented):
@HyeockJinKim 
This loader is intentionally different from other loaders because it serves a different purpose:
**Other loaders (entity loaders)**: Load individual entities by ID
- `load_roles_by_ids([role_id])` → `[RoleData | None]`
- `None` means "this entity doesn't exist"

**This loader (relationship loader)**: Load collections of related entities
- `load_assignments_by_role_ids([role_id])` → `[list[AssignedUserData]]`
- `[]` means "this role has no assignments" (a valid business state)

Returning `None` would incorrectly treat "no assignments" as an error condition, when it's actually a normal state. The empty list `[]` correctly represents "zero items in the collection."

tests/unit/manager/api/rbac/test_dataloader.py
```diff
@@ -0,0 +1,125 @@
...
+
+    @staticmethod
+    def create_mock_assignment(role_id: uuid.UUID, user_id: uuid.UUID) -> MagicMock:
+        mock = MagicMock(spec=AssignedUserData)
+        mock.role_id = role_id
+        mock.user_id = user_id
+        return mock
+
+    @staticmethod
+    def create_mock_processor(assignments: list[MagicMock]) -> MagicMock:
```
seedspirit, line 25: It seems better to expose the test intent by creating the processor itself as a fixture and then separately mocking the behavior of the mock processor according to the test scenario.

PR #9541, opened 2026-02-28
refactor(BA-4807): migrate all API modules to RouteRegistry.create_subapp()

HyeockJinKim:
## Summary
- Add `RouteRegistry.create_subapp()` with deferred `mount()` to support CORS setup before route registration, replacing legacy `global_subapp_pkgs` + `create_app()` + `init_subapp()` pipeline
- Wire all 39 API modules via centralized `_register_module()` match/case dispatch in `server.py`, with `register_modules()` public API for test fixtures
- Fix `admin_required`/`superadmin_required` decorators to correctly return HTTP 403 (not 401) for authenticated users with insufficient privileges

## Test plan
- [x] All 40 component tests pass (`pants test tests/component::`)
- [x] Lint and type check pass (`pants fmt fix lint check --changed-since=HEAD~1`)
- [ ] CI pipeline passes

Resolves BA-4807

----
📚 Documentation preview 📚: https://sorna--9541.org.readthedocs.build/en/9541/

----
📚 Documentation preview 📚: https://sorna-ko--9541.org.readthedocs.build/ko/9541/

### DIFF
--- a/tests/component/conftest.py
+++ b/tests/component/conftest.py
@@ -864,19 +864,19 @@ type ModuleDepsFactory = Callable[[RootContext], ModuleDeps]
 
 @pytest.fixture()
 def server_module_deps_factory() -> ModuleDepsFactory:
-    """Build ``ModuleDeps`` from a fully-initialised ``RootContext``.
+    """Build ``ModuleDeps`` from a ``RootContext``.
 
-    Override this fixture in domain-specific conftest.py when the
-    domain's cleanup contexts set additional attributes (e.g.
-    ``services_ctx``) that the default implementation does not read.
+    Falls back to ``MagicMock()`` for attributes not initialised by the
+    test's cleanup contexts.  Override in domain-specific conftest.py
+    when custom construction logic is needed.
     """
 
     def _factory(root_ctx: RootContext) -> ModuleDeps:
         return ModuleDeps(
             cors_options=root_ctx.cors_options,
-            processors=root_ctx.processors,
-            services_ctx=MagicMock(),
-            storage_manager=root_ctx.storage_manager,
+            processors=getattr(root_ctx, "processors", None) or MagicMock(),
+            services_ctx=getattr(root_ctx, "services_ctx", None) or MagicMock(),
+            storage_manager=getattr(root_ctx, "storage_manager", None) or MagicMock(),
             auth_config=root_ctx.config_provider.config.auth,
         )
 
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/server.py
```diff
@@ -1379,42 +1404,249 @@ def init_subapp(pkg_name: str, root_app: web.Application, create_subapp: AppCrea
...
-    if pidx == 0:
-        log.info("Loading new-style module: logs")
-    register_logs_routes(registry, dep_resources.processing.processors)
+def _register_module(
+    name: str,
+    root_app: web.Application,
+    cors: Mapping[str, aiohttp_cors.ResourceOptions],
+    processors: Any,
+    root_ctx: RootContext,
+) -> None:
```
HyeockJinKim: Instead of registering the module in server.py, it seems better to provide a function that registers it in api/rest, and have server.py only call that function.

src/ai/backend/manager/server.py
```diff
@@ -1379,42 +1404,249 @@ def init_subapp(pkg_name: str, root_app: web.Application, create_subapp: AppCrea
...
+            reg = RouteRegistry.create_subapp(root_app, "acl", cors)
+            register_acl_routes(reg)
+        case ".scaling_group":
+            reg = RouteRegistry.create_subapp(root_app, "scaling-groups", cors)
+            register_scaling_group_routes(reg)
+        case ".logs":
+            reg = RouteRegistry.create_subapp(root_app, "logs", cors)
+            reg.app["logs.context"] = LogsPrivateContext()
+            reg.app.on_startup.append(logs_init)
+            reg.app.on_shutdown.append(logs_shutdown)
```
HyeockJinKim: It is necessary to verify whether on_startup and on_shutdown should be configured individually.
And rather than setting app-specific values in a map, I want dependency values to be included when setting up the handler.

src/ai/backend/manager/server.py
```diff
@@ -1379,42 +1404,249 @@ def init_subapp(pkg_name: str, root_app: web.Application, create_subapp: AppCrea
...
+def _register_module(
+    name: str,
+    root_app: web.Application,
+    cors: Mapping[str, aiohttp_cors.ResourceOptions],
+    processors: Any,
+    root_ctx: RootContext,
+) -> None:
+    """Register a single API module by its package name."""
+    match name:
+        case ".auth":
```
HyeockJinKim: Instead of registering module names with case sensitivity, it seems better to provide the necessary dependencies as imported registration functions. (Don't use module name as str)

PR #9553, opened 2026-03-01
feat(BA-4811): update association_scopes_entities on role-user mapping changes

fregataa:
## Summary
- When a role is assigned to a user (`assign_role`, `map_user_to_role`), create an `association_scopes_entities` record (USER scope → ROLE entity) so the role becomes visible to the user through RBAC
- When a role is revoked from a user (`revoke_role`), delete the corresponding `association_scopes_entities` record
- Add tests verifying association creation on assign, deletion on revoke, correct scope/entity types, and isolation between users

## Test plan
- [x] `test_assign_role_creates_association` — verifies association record is created
- [x] `test_revoke_role_deletes_association` — verifies association record is deleted
- [x] `test_assign_role_association_has_correct_types` — verifies USER scope and ROLE entity types
- [x] `test_revoke_preserves_other_associations` — verifies revoking one user's role doesn't affect another

Resolves BA-4811

### DIFF
--- a/changes/9553.feature.md
+++ b/changes/9553.feature.md
@@ -1,0 +1,1 @@
+Create and delete `association_scopes_entities` records on user-role assignment and revocation so that roles are visible to users through RBAC.


fregataa:
Closing this PR. Currently, roles are not managed as RBAC entities, and only superadmins are allowed to CRUD roles other than their own. Therefore, scope chain traversal for role-user mappings is unnecessary, and there is no need to add association_scopes_entities records for this case.

PR #9554, opened 2026-03-01
feat(BA-4812): add my_roles GraphQL query

fregataa:
## Summary
- Add `my_roles` GraphQL query that returns roles assigned to the current authenticated user
- Implement `RoleConditions.by_assigned_user_id` to filter roles via `user_roles` table subquery
- No RBAC check required — users can always see their own roles
- Support pagination (cursor-based and offset-based), filtering, and ordering

## Test plan
- [x] Unit tests for `my_roles` resolver (authenticated call, unauthenticated rejection, pagination params)
- [x] Unit test for `RoleConditions.by_assigned_user_id` condition
- [x] `pants fmt`, `fix`, `lint` pass
- [x] `pants check` (mypy) passes
- [x] `pants test --changed-since=HEAD~1` passes

Resolves BA-4812

----
📚 Documentation preview 📚: https://sorna--9554.org.readthedocs.build/en/9554/

----
📚 Documentation preview 📚: https://sorna-ko--9554.org.readthedocs.build/ko/9554/

PR #9555, opened 2026-03-01
feat(BA-4813): update search_roles() to reflect RBAC visibility

fregataa:
## Summary
- Add `RoleSearchScope` that filters roles visible to a user through direct assignment (`user_roles`) or RBAC scope chain (`association_scopes_entities`)
- Update `search_roles()` across all layers (db_source, repository, service, action) to accept optional RBAC scope
- REST handler now allows non-admin users to search roles (scoped to their own), removing the superadmin-only restriction

## Test plan
- [x] 5 new tests covering RBAC visibility: admin sees all, direct assignment, scope association, no assignments, combined sources
- [x] All existing tests pass unchanged
- [x] pants fmt/fix/lint/check pass

Resolves BA-4813


fregataa:
Closed this PR since it has already been addressed by #9554

PR #9557, opened 2026-03-01
feat(BA-4816): add model folder pre-validation for inference sessions

HyeockJinKim:
## Summary
- Add `InferenceModelFolderRule` validator that checks inference sessions have at least one model-type virtual folder mounted before RPC dispatch to the agent
- Skip validation for `runtime_variant=custom` (uses `model_definition.yml` directly) and non-inference session types
- Prevents unnecessary RPC failures (~900/hr observed in dogbowl over 2 days)

## Test plan
- [x] Non-inference session types pass without model folder
- [x] Inference session with model folder passes
- [x] Inference session without model folder raises `InvalidAPIParameters`
- [x] Inference session with `runtime_variant=custom` passes without model folder
- [x] Inference session with `runtime_variant=None` requires model folder
- [x] Inference session with only GENERAL/DATA folders raises error
- [x] All existing validator tests still pass

Resolves BA-4816

### DIFF
--- a/tests/unit/manager/repositories/prometheus_query_preset/test_prometheus_query_preset_options.py
+++ b/tests/unit/manager/repositories/prometheus_query_preset/test_prometheus_query_preset_options.py
@@ -136,7 +136,6 @@ class TestPrometheusQueryPresetOptions:
         preset_ids = await self._seed_presets(case.seeds, db_with_cleanup)
         return SearchContext(preset_ids=preset_ids, case=case)
 
-    @pytest.mark.asyncio
     @pytest.mark.parametrize(
         "test_case",
         [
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/sokovan/scheduling_controller/validators/inference.py
```diff
@@ -0,0 +1,41 @@
...
+        self,
+        spec: SessionCreationSpec,
+        context: SessionCreationContext,
+    ) -> None:
+        if spec.session_type != SessionTypes.INFERENCE:
+            return
+
+        runtime_variant = spec.creation_spec.get("runtime_variant")
+        if runtime_variant == RuntimeVariant.CUSTOM.value:
+            return
```
HyeockJinKim: This check appears unrelated to vfolder mounts.

PR #9562, opened 2026-03-01
feat(BA-4819): add alias filter conditions to ImageConditions

fregataa:
## Summary
- Add 4 subquery-based alias filter methods (`by_alias_contains`, `by_alias_equals`, `by_alias_starts_with`, `by_alias_ends_with`) to `ImageConditions` class
- Each method uses `ImageRow.id.in_(subquery)` pattern against `ImageAliasRow` to filter images by their aliases
- Supports `case_insensitive` (ilike vs like) and `negated` (sa.not_()) flags, consistent with existing name/architecture filters

## Test plan
- [x] Verify all 4 methods follow the same IN subquery pattern
- [x] Verify case_insensitive flag toggles ilike/like correctly
- [x] Verify negated flag wraps condition with sa.not_()
- [x] CI passes (lint, type check, tests)

Resolves BA-4819

### DIFF
--- a/tests/unit/manager/repositories/image/test_image_repository.py
+++ b/tests/unit/manager/repositories/image/test_image_repository.py
@@ -370,12 +370,12 @@ class TestImageRepositorySearch:
 
         yield image_ids
 
-    async def test_filter_by_alias_contains(
+    async def test_filter_by_single_alias_condition(
         self,
         image_repository: ImageRepository,
         images_with_aliases: list[UUID],
     ) -> None:
-        """Test by_alias_contains matches images whose alias contains the substring."""
+        """Test filtering images with a single alias condition."""
         querier = BatchQuerier(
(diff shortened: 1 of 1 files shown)


tests/unit/manager/repositories/image/test_image_repository.py
```diff
@@ -314,3 +316,118 @@ async def test_search_images_no_results(
...
+                        image_id=image_row.id,
+                    )
+                    db_sess.add(alias)
+
+            image_ids = [row.id for row in image_rows]
+            await db_sess.commit()
+
+        yield image_ids
+
+    async def test_filter_by_alias(
```
seedspirit: Since the same format of test methods is being repeated, it seems we could make it much more concise using parametrize or similar approaches.

jopemachine (approved):

PR #9566, opened 2026-03-02
feat(BA-4821): Add deployment strategy evaluation framework

### LINKED ISSUE #9565: [Phase 7] Deployment strategy evaluation framework
## Motivation

Backend.AI needs to support advanced deployment strategies such as Blue-Green and Rolling Update for model serving endpoints. This requires a comprehensive framework that spans type definitions, database schema, strategy evaluation, handler orchestration, and coordinator integration. BEP-1049 defines the architectural design using a "Flat Registry with Pre-Step Pattern".

## Required Features

- DeploymentSubStep type system and DB schema for tracking deployment phases
- Strategy evaluation framework with per-strategy FSM dispatch and batched route mutations
- DEPLOYING lifecycle handlers for provisioning and progressing sub-steps
- Coordinator refactoring with flat HandlerRegistry, retry logic, and failure classification
- Sub-step aware event dispatching and scheduling

## Child Stories

- BA-4966: Update BEP-1049 deployment strategy design documents
- BA-4967: Add DeploymentSubStep type, DB schema, and repository extensions
- BA-4968: Add deployment strategy evaluation framework and handlers
- BA-4969: Refactor coordinator with flat registry and event sub_step support

## Impact

- Manager sokovan deployment module (coordinator, handlers, strategy)
- Common event types and Valkey schedule client
- Database schema (endpoints table sub_step column)
- Deployment service and controller

## Testing Scenarios

- All child story tests pass independently
- Manager starts successfully after all PRs are merged
- End-to-end deployment lifecycle with strategy evaluation works correctly

JIRA Issue: BA-4821

jopemachine:
resolves #9565 (BA-4821)

- [x] Test case(s) to:
- [ ] Documentation

----
📚 Documentation preview 📚: https://sorna--9566.org.readthedocs.build/en/9566/

----
📚 Documentation preview 📚: https://sorna-ko--9566.org.readthedocs.build/ko/9566/

### DIFF
--- a/src/ai/backend/manager/repositories/deployment/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/deployment/db_source/db_source.py
@@ -168,23 +168,43 @@ class StrategyTransaction:
     def __init__(
         self,
         db_sess: SASession,
-        apply_sub_step_assignments: Any,
     ) -> None:
         self._db_sess = db_sess
-        self._apply_sub_step_assignments = apply_sub_step_assignments
 
-    async def apply_strategy_evaluation(
+    async def update_sub_steps(
         self,
         assignments: dict[uuid.UUID, DeploymentSubStep],
-        rollout: BulkCreator[RoutingRow],
-        drain: BatchUpdater[RoutingRow] | None,
     ) -> None:
-        """Update sub_steps and apply route changes."""
-        await self._apply_sub_step_assignments(self._db_sess, assignments)
-        if rollout.specs:
-            await execute_bulk_creator(self._db_sess, rollout)
-        if drain:
-            await execute_batch_updater(self._db_sess, drain)
+        """Bulk-update the sub_step column, grouped by sub_step value."""
+        grouped: dict[DeploymentSubStep, list[uuid.UUID]] = {}
+        for endpoint_id, sub_step in assignments.items():
+            grouped.setdefault(sub_step, []).append(endpoint_id)
+        for sub_step, endpoint_ids in grouped.items():
+            stmt = (
+                sa.update(EndpointRow)
+                .where(
+                    EndpointRow.id.in_(endpoint_ids),
+                    # Only update endpoints still in DEPLOYING — if the lifecycle
+                    # changed concurrently (e.g. to DESTROYING), skip the stale update.
+                    EndpointRow.lifecycle_stage == EndpointLifecycle.DEPLOYING,
+                )
+                .values(sub_step=sub_step)
+            )
+            await self._db_sess.execute(stmt)
+
+    async def create_routes(
+        self,
+        rollout: BulkCreator[RoutingRow],
+    ) -> None:
+        """Bulk-create new-revision routes."""
+        await execute_bulk_creator(self._db_sess, rollout)
+
+    async def drain_routes(
+        self,
+        drain: BatchUpdater[RoutingRow],
+    ) -> None:
+        """Mark old-revision routes for draining."""
+        await execute_batch_updater(self._db_sess, drain)
 
     async def complete_deployment_revision_swap(
         self,
@@ -2407,47 +2427,6 @@ class DeploymentDBSource:
         async with self._begin_session_read_committed() as db_sess:
             return await execute_purger(db_sess, purger)
 
-    async def update_sub_steps(
-        self,
-        assignments: dict[uuid.UUID, DeploymentSubStep],
-    ) -> None:
-        """Bulk-update the sub_step column for multiple endpoints.
-
-        Args:
-            assignments: Mapping from endpoint ID to the sub_step value to set.
-        """
-        async with self._begin_session_read_committed() as db_sess:
-            await self._apply_sub_step_assignments(db_sess, assignments)
-
-    async def apply_strategy_evaluation(
-        self,
-        assignments: dict[uuid.UUID, DeploymentSubStep],
-        rollout: BulkCreator[RoutingRow],
-        drain: BatchUpdater[RoutingRow] | None,
-    ) -> None:
-        """Atomically update sub_steps and apply route changes in a single transaction.
-
-        This prevents an inconsistent state where sub_step is updated (e.g. COMPLETED)
-        but route mutations fail, which could cause the handler to swap revisions
-        without the expected routes being in place.
-
-        Args:
-            assignments: Mapping from endpoint ID to sub_step value.
-            rollout: Bulk creator for new-revision routes.
-            drain: Batch updater for draining old-revision routes.
-        """
-        async with self._begin_session_read_committed() as db_sess:
-            # 1. Update sub_steps
-            await self._apply_sub_step_assignments(db_sess, assignments)
-
-            # 2. Rollout new routes
-            if rollout.specs:
-                await execute_bulk_creator(db_sess, rollout)
-
-            # 3. Drain old routes
-            if drain:
-                await execute_batch_updater(db_sess, drain)
-
     @actxmgr
     async def begin_strategy_transaction(self) -> AsyncIterator[StrategyTransaction]:
         """Begin a transaction that spans multiple strategy-related DB operations.
@@ -2457,29 +2436,7 @@ class DeploymentDBSource:
         route mutations, revision swap, and deploying_revision clear.
         """
         async with self._begin_session_read_committed() as db_sess:
-            yield StrategyTransaction(db_sess, self._apply_sub_step_assignments)
-
-    @staticmethod
-    async def _apply_sub_step_assignments(
-        db_sess: SASession,
-        assignments: dict[uuid.UUID, DeploymentSubStep],
-    ) -> None:
-        """Group assignments by sub_step and execute bulk UPDATEs."""
-        grouped: dict[DeploymentSubStep, list[uuid.UUID]] = {}
-        for endpoint_id, sub_step in assignments.items():
-            grouped.setdefault(sub_step, []).append(endpoint_id)
-        for sub_step, endpoint_ids in grouped.items():
-            stmt = (
-                sa.update(EndpointRow)
-                .where(
-                    EndpointRow.id.in_(endpoint_ids),
-                    # Only update endpoints still in DEPLOYING — if the lifecycle
-                    # changed concurrently (e.g. to DESTROYING), skip the stale update.
-                    EndpointRow.lifecycle_stage == EndpointLifecycle.DEPLOYING,
-                )
-                .values(sub_step=sub_step)
-            )
-            await db_sess.execute(stmt)
+            yield StrategyTransaction(db_sess)
 
     # ========== Access Token Operations ==========
 
--- a/src/ai/backend/manager/repositories/deployment/repository.py
+++ b/src/ai/backend/manager/repositories/deployment/repository.py
@@ -1222,28 +1222,6 @@ class DeploymentRepository:
         """
         return await self._db_source.delete_deployment_policy(purger)
 
-    @deployment_repository_resilience.apply()
-    async def update_sub_steps(
-        self,
-        assignments: dict[uuid.UUID, DeploymentSubStep],
-    ) -> None:
-        """Bulk-update the sub_step column for multiple endpoints."""
-        await self._db_source.update_sub_steps(assignments)
-
-    @deployment_repository_resilience.apply()
-    async def apply_strategy_evaluation(
-        self,
-        assignments: dict[uuid.UUID, DeploymentSubStep],
-        rollout: BulkCreator[RoutingRow],
-        drain: BatchUpdater[RoutingRow] | None,
-    ) -> None:
-        """Atomically apply sub_step assignments and route mutations in a single transaction."""
-        await self._db_source.apply_strategy_evaluation(
-            assignments,
-            rollout,
-            drain,
-        )
-
     def begin_strategy_transaction(self) -> AsyncContextManager[StrategyTransaction]:
         """Begin a transaction that spans multiple strategy-related DB operations.
 
--- a/src/ai/backend/manager/sokovan/deployment/strategy/applier.py
+++ b/src/ai/backend/manager/sokovan/deployment/strategy/applier.py
@@ -94,15 +94,16 @@ class StrategyResultApplier:
 
         # All DB mutations in a single transaction via StrategyTransaction.
         async with self._deployment_repo.begin_strategy_transaction() as txn:
-            # 1. Sub-step assignments + route rollout/drain
-            await txn.apply_strategy_evaluation(summary.assignments, rollout, drain)
+            await txn.update_sub_steps(summary.assignments)
 
-            # 2. Revision swap for COMPLETED deployments
+            if rollout.specs:
+                await txn.create_routes(rollout)
+            if drain:
+                await txn.drain_routes(drain)
+
             swapped = 0
             if completed_ids:
                 swapped = await txn.complete_deployment_revision_swap(completed_ids)
-
-            # 3. Clear deploying_revision for ROLLED_BACK deployments
             if rolled_back_ids:
                 await txn.clear_deploying_revision(rolled_back_ids)
 
--- a/tests/unit/manager/sokovan/deployment/strategy/test_applier.py
+++ b/tests/unit/manager/sokovan/deployment/strategy/test_applier.py
@@ -10,9 +10,6 @@ from uuid import UUID, uuid4
 import pytest
 
 from ai.backend.manager.data.deployment.types import DeploymentSubStep
-from ai.backend.manager.models.routing import RoutingRow
-from ai.backend.manager.repositories.base.creator import BulkCreator
-from ai.backend.manager.repositories.base.updater import BatchUpdater
 from ai.backend.manager.sokovan.deployment.strategy.applier import (
     StrategyApplyResult,
     StrategyResultApplier,
@@ -45,7 +42,9 @@ def _build_summary(
 @pytest.fixture
 def mock_txn() -> AsyncMock:
     txn = AsyncMock()
-    txn.apply_strategy_evaluation = AsyncMock()
+    txn.update_sub_steps = AsyncMock()
+    txn.create_routes = AsyncMock()
+    txn.drain_routes = AsyncMock()
     txn.complete_deployment_revision_swap = AsyncMock(return_value=0)
     txn.clear_deploying_revision = AsyncMock()
     return txn
@@ -54,7 +53,6 @@ def mock_txn() -> AsyncMock:
 @pytest.fixture
 def mock_deployment_repo(mock_txn: AsyncMock) -> AsyncMock:
     repo = AsyncMock()
-    repo.apply_strategy_evaluation = AsyncMock()
 
     @asynccontextmanager
     async def _begin_strategy_transaction() -> AsyncIterator[AsyncMock]:
@@ -149,7 +147,9 @@ class TestStrategyResultApplier:
     ) -> None:
         result = await applier.apply(empty_summary)
 
-        mock_txn.apply_strategy_evaluation.assert_not_called()
+        mock_txn.update_sub_steps.assert_not_called()
+        mock_txn.create_routes.assert_not_called()
+        mock_txn.drain_routes.assert_not_called()
         mock_txn.complete_deployment_revision_swap.assert_not_called()
         mock_txn.clear_deploying_revision.assert_not_called()
         assert result.completed_ids == set()
@@ -157,7 +157,7 @@ class TestStrategyResultApplier:
         assert result.routes_created == 0
         assert result.routes_drained == 0
 
-    async def test_assignments_only_calls_apply_strategy_evaluation(
+    async def test_assignments_only(
         self,
         applier: StrategyResultApplier,
         mock_txn: AsyncMock,
@@ -165,13 +165,15 @@ class TestStrategyResultApplier:
     ) -> None:
         result = await applier.apply(provisioning_summary)
 
-        mock_txn.apply_strategy_evaluation.assert_called_once()
+        mock_txn.update_sub_steps.assert_called_once()
+        mock_txn.create_routes.assert_called_once()
+        mock_txn.drain_routes.assert_not_called()
         mock_txn.complete_deployment_revision_swap.assert_not_called()
         mock_txn.clear_deploying_revision.assert_not_called()
         assert result.routes_created == 0
         assert result.routes_drained == 0
 
-    async def test_rollout_specs_creates_bulk_creator(
+    async def test_rollout_calls_create_routes(
         self,
         applier: StrategyResultApplier,
         mock_txn: AsyncMock,
@@ -179,13 +181,12 @@ class TestStrategyResultApplier:
     ) -> None:
         result = await applier.apply(summary_with_rollout)
 
-        mock_txn.apply_strategy_evaluation.assert_called_once()
-        args = mock_txn.apply_strategy_evaluation.call_args
-        rollout_arg: BulkCreator[RoutingRow] = args[0][1]
+        mock_txn.create_routes.assert_called_once()
+        rollout_arg = mock_txn.create_routes.call_args[0][0]
         assert len(rollout_arg.specs) == 1
         assert result.routes_created == 1
 
-    async def test_drain_route_ids_creates_batch_updater(
+    async def test_drain_calls_drain_routes(
         self,
         applier: StrategyResultApplier,
         mock_txn: AsyncMock,
@@ -193,13 +194,10 @@ class TestStrategyResultApplier:
     ) -> None:
         result = await applier.apply(summary_with_drain)
 
-        mock_txn.apply_strategy_evaluation.assert_called_once()
-        args = mock_txn.apply_strategy_evaluation.call_args
-        drain_arg: BatchUpdater[RoutingRow] | None = args[0][2]
-        assert drain_arg is not None
+        mock_txn.drain_routes.assert_called_once()
         assert result.routes_drained == 1
 
-    async def test_no_drain_routes_passes_none(
+    async def test_no_drain_routes_skips_drain(
         self,
         applier: StrategyResultApplier,
         mock_txn: AsyncMock,
@@ -207,9 +205,7 @@ class TestStrategyResultApplier:
     ) -> None:
         await applier.apply(provisioning_summary)
 
-        args = mock_txn.apply_strategy_evaluation.call_args
-        drain_arg = args[0][2]
-        assert drain_arg is None
+        mock_txn.drain_routes.assert_not_called()
 
     async def test_completed_calls_revision_swap(
         self,
@@ -222,7 +218,7 @@ class TestStrategyResultApplier:
 
         result = await applier.apply(summary)
 
-        mock_txn.apply_strategy_evaluation.assert_called_once()
+        mock_txn.update_sub_steps.assert_called_once()
         mock_txn.complete_deployment_revision_swap.assert_called_once_with(completed_ids)
         mock_txn.clear_deploying_revision.assert_not_called()
         assert result.completed_ids == completed_ids
@@ -237,7 +233,7 @@ class TestStrategyResultApplier:
 
         result = await applier.apply(summary)
 
-        mock_txn.apply_strategy_evaluation.assert_called_once()
+        mock_txn.update_sub_steps.assert_called_once()
         mock_txn.clear_deploying_revision.assert_called_once_with(rolled_back_ids)
         mock_txn.complete_deployment_revision_swap.assert_not_called()
         assert result.rolled_back_ids == rolled_back_ids
@@ -253,7 +249,9 @@ class TestStrategyResultApplier:
 
         result = await applier.apply(summary)
 
-        mock_txn.apply_strategy_evaluation.assert_called_once()
+        mock_txn.update_sub_steps.assert_called_once()
+        mock_txn.create_routes.assert_called_once()
+        mock_txn.drain_routes.assert_called_once()
         mock_txn.complete_deployment_revision_swap.assert_called_once_with({completed_id})
         mock_txn.clear_deploying_revision.assert_called_once_with({rolled_back_id})
         assert result.completed_ids == {completed_id}


jopemachine (commented):
## PR #9566 Security & Performance Review

**Reviewer**: AI Code Reviewer
**Risk Level**: MEDIUM
**Scope**: Framework PR for BEP-1049 deployment strategy evaluation infrastructure

---

### Summary

This PR introduces the DEPLOYING lifecycle with a strategy evaluator framework, sub-step handlers (PROVISIONING, PROGRESSING, ROLLED_BACK), and coordinator integration. The architecture is well-structured with clean separation between the evaluator (which decides route mutations) and handlers (which manage lifecycle transitions). However, there are several issues that should be addressed before or shortly after merging.

### Findings Overview

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | HIGH | `coordinator.py` | Race condition: revision swap and lifecycle transition use separate DB sessions |
| 2 | HIGH | `coordinator.py` | Hardcoded "Rolling update" history message applies to all strategies |
| 3 | MEDIUM | `evaluator.py` / `coordinator.py` | Errored deployments silently dropped -- no retry bound or failure transition |
| 4 | MEDIUM | `coordinator.py` | Post-process assumes all handler executions succeeded |
| 5 | MEDIUM | `coordinator.py` | Skipped deployments (no policy) are never acted upon |
| 6 | LOW | `strategy/types.py` | `CycleEvaluationResult` docstring says "rolling update" but type is strategy-agnostic |
| 7 | LOW | (all) | Zero test coverage for the entire framework |

Detailed review comments follow on specific lines below.

src/ai/backend/manager/sokovan/deployment/coordinator.py
```diff
@@ -314,6 +353,134 @@ async def _handle_status_transitions(
...
+        """Transition completed DEPLOYING deployments to READY.
+
+        1. Swap deploying_revision → current_revision.
+        2. Update lifecycle to READY with history recording.
+        3. Send notification events.
+        """
+        endpoint_ids = {deployment.id for deployment in completed}
+
+        # Swap revisions
+        await self._deployment_repository.complete_deployment_revision_swap(endpoint_ids)
```
jopemachine: **[HIGH] Race condition: revision swap and lifecycle transition use separate DB sessions**

`complete_deployment_revision_swap` (line 433) and `update_endpoint_lifecycle_bulk_with_history` (line 466) execute in **separate database sessions** (each repository method opens its own transaction, per the repository layer guardrails).

If the process crashes or the distributed lock expires between these two calls, the deployment ends up in an inconsistent state:
- `current_revision` is already swapped to the new revision
- `deploying_revision` is already NULL
- But `lifecycle` is still `DEPLOYING`

On the next evaluation cycle, this deployment will be picked up again with `deploying_revision_id=None`, which means the evaluator will either skip it (no deploying revision to evaluate) or produce undefined behavior.

**Suggested fix**: Combine the revision swap and lifecycle transition into a single repository method that executes both operations within one database session/transaction. Alternatively, add a defensive check in `_process_with_evaluator` to detect deployments stuck in DEPLOYING with `deploying_revision_id=None` and transition them to READY (since the swap already happened) or an error state.

src/ai/backend/manager/sokovan/deployment/coordinator.py
```diff
@@ -314,6 +353,134 @@ async def _handle_status_transitions(
...
+            ],
+        )
+
+        timestamp_now = datetime.now(UTC).isoformat()
+        history_specs = [
+            DeploymentHistoryCreatorSpec(
+                deployment_id=deployment.id,
+                phase=lifecycle_type.value,
+                result=SchedulingResult.SUCCESS,
+                message="Rolling update completed successfully",
```
jopemachine: **[HIGH] Hardcoded "Rolling update" history message applies to all strategies**

The history message `"Rolling update completed successfully"` is hardcoded here, but `_transition_completed_deployments` is called for **all** deployment strategies -- including blue-green deployments. When child PR #9568 (blue-green) is merged, completed blue-green deployments will be recorded with an incorrect "Rolling update" message in their deployment history.

Since the `DeploymentPolicyData` is available during evaluation, consider:
1. Including the strategy type in the `EvaluationResult.completed` list (e.g. as a tuple `(DeploymentInfo, DeploymentStrategy)`).
2. Or, passing the strategy name into `_transition_completed_deployments` so the history message can be parameterized:

```python
message=f"{strategy.value} deployment completed successfully"
```

src/ai/backend/manager/sokovan/deployment/strategy/evaluator.py
```diff
@@ -0,0 +1,159 @@
...
+                result.skipped.append(deployment)
+                continue
+
+            routes: list[RouteInfo] = list(route_map.get(deployment.id, []))
+
+            try:
+                cycle_result = self._evaluate_single(deployment, routes, policy.strategy, policy)
+            except Exception as e:
+                log.warning("deployment {}: evaluation error — {}", deployment.id, e)
+                result.errors.append((deployment, str(e)))
```
jopemachine: **[MEDIUM] Errored deployments silently dropped -- no retry bound or failure transition**

`result.errors` is populated here when evaluation throws an exception, but this list is **never acted upon** by the caller (`_process_with_evaluator` in `coordinator.py`). The coordinator only processes `eval_result.groups` and `eval_result.completed` -- it completely ignores `eval_result.errors` and `eval_result.skipped`.

Consequences:
- Errored deployments remain in `DEPLOYING` state indefinitely.
- On every evaluation cycle (every 5-30 seconds), the same deployment will be re-evaluated and fail again, creating an infinite retry loop for permanent errors (e.g. misconfigured policy, unsupported strategy).
- No deployment history entry is recorded for the failure, making debugging difficult.
- No notification is sent to inform the user.

**Suggested fix**: In `_process_with_evaluator`, after processing groups and completed deployments, add handling for `eval_result.errors`:
1. For transient errors: increment a retry counter and log a warning. After N retries, treat as permanent.
2. For permanent errors: transition to an error state (e.g. `READY` with error sub-status) and record a failure history entry.

The same concern applies to `eval_result.skipped` -- deployments with no policy will loop forever. Consider adding a bounded retry or error transition for those as well.

src/ai/backend/manager/sokovan/deployment/coordinator.py
```diff
@@ -314,6 +353,134 @@ async def _handle_status_transitions(
...
+                    await self._handle_status_transitions(handler, sub_result, all_records)
+
+            # Post-process outside recorder scope
+            for sub_step, group in eval_result.groups.items():
+                handler_key = (lifecycle_type, sub_step)
+                handler = self._deployment_handlers.get(handler_key)
+                if handler is None:
+                    continue
+                try:
+                    result_for_post = DeploymentExecutionResult(successes=group.deployments)
```
jopemachine: **[MEDIUM] Post-process assumes all handler executions succeeded**

The post-process loop (lines 399-413) constructs `DeploymentExecutionResult(successes=group.deployments)` for every group, regardless of the actual outcome of `handler.execute()` on line 396.

The actual `sub_result` from `handler.execute(group.deployments)` is passed to `_handle_status_transitions` (line 397), but it is then **discarded**. The post-process step unconditionally treats all deployments in the group as successes.

For the current in-progress handlers (`DeployingInProgressHandler`), this is harmless because `execute()` always returns all deployments as successes. However, this creates a subtle contract violation that could cause bugs when:
- A future handler implementation returns some deployments as errors.
- The post-process logic (e.g. re-scheduling DEPLOYING) runs for failed deployments too, potentially re-triggering the same failed operation.

**Suggested fix**: Capture `sub_result` from `handler.execute()` in a mapping and use it in the post-process loop:

```python
sub_results: dict[DeploymentSubStep, DeploymentExecutionResult] = {}
for sub_step, group in eval_result.groups.items():
    ...
    sub_result = await handler.execute(group.deployments)
    sub_results[sub_step] = sub_result
    ...

# Post-process
for sub_step, group in eval_result.groups.items():
    ...
    await handler.post_process(sub_results[sub_step])
```

src/ai/backend/manager/sokovan/deployment/strategy/evaluator.py
```diff
@@ -0,0 +1,159 @@
...
+        # ── 2. Per-deployment evaluation ──
+        all_scale_out: list[Creator[RoutingRow]] = []
+        all_scale_in_ids: list[UUID] = []
+
+        for deployment in deployments:
+            policy = policy_map.get(deployment.id)
+            if policy is None:
+                log.warning("deployment {}: no policy found — skipping", deployment.id)
+                result.skipped.append(deployment)
+                continue
```
jopemachine, line 89: **[MEDIUM] Skipped deployments (no policy) are never acted upon**

When a deployment has no associated policy (`policy_map.get(deployment.id)` returns `None`), it is added to `result.skipped` (line 77) and the evaluator moves on. However, the coordinator in `_process_with_evaluator` does not process `eval_result.skipped` at all.

This means:
- Deployments without a policy remain stuck in `DEPLOYING` forever.
- Every evaluation cycle (5-30 seconds) will re-query and re-skip these deployments, wasting database queries.
- There is no user-visible indication that the deployment is stuck due to a missing policy.

This could happen if:
- A policy is deleted while a deployment is in progress.
- A race condition during deployment creation leaves the endpoint in DEPLOYING before the policy row is committed.

**Suggested fix**: In the coordinator, handle `eval_result.skipped` by either:
1. Logging a warning with the deployment IDs for operational visibility.
2. After a configurable number of cycles, transitioning them to an error state.
3. At minimum, add a log statement in the coordinator to surface these skipped deployments.

src/ai/backend/manager/sokovan/deployment/strategy/types.py
```diff
@@ -0,0 +1,48 @@
...
+@dataclass
+class RouteChanges:
+    """Route mutations to apply for a single deployment cycle."""
+
+    scale_out_specs: list[Creator[RoutingRow]] = field(default_factory=list)
+    scale_in_route_ids: list[UUID] = field(default_factory=list)
+
+
+@dataclass
+class CycleEvaluationResult:
```
jopemachine: **[LOW] Docstring says "rolling update" but the type is strategy-agnostic**

The docstring reads:
> Result of evaluating a single deployment's rolling update cycle.

But `CycleEvaluationResult` is used by both `rolling_update_evaluate` and `blue_green_evaluate`. Consider updating to:

```python
"""Result of evaluating a single deployment strategy cycle."""
```

src/ai/backend/manager/sokovan/deployment/strategy/evaluator.py
```diff
@@ -0,0 +1,159 @@
...
+from ai.backend.manager.repositories.deployment.repository import DeploymentRepository
+
+from .blue_green import blue_green_evaluate
+from .rolling_update import rolling_update_evaluate
+from .types import CycleEvaluationResult, EvaluationGroup, EvaluationResult
+
+log = BraceStyleAdapter(logging.getLogger(__name__))
+
+
+class DeploymentStrategyEvaluator:
```
jopemachine, line 42: **[LOW] Zero test coverage for the entire framework**

This PR introduces ~750 lines of new code across 6 new files (evaluator, handlers, types, strategy stubs, coordinator integration) with **zero test additions**. Key components that should have test coverage:

1. **`DeploymentStrategyEvaluator.evaluate()`** -- the core orchestration logic:
   - Verify correct grouping of deployments by sub-step.
   - Verify that `completed` deployments are separated from in-progress ones.
   - Verify that `errors` and `skipped` are populated correctly.
   - Verify that route mutations are aggregated and applied in batch.

2. **`_process_with_evaluator()`** -- coordinator integration:
   - Verify the lock is acquired.
   - Verify handler dispatch by composite key.
   - Verify status transitions are recorded.
   - Verify completed deployments trigger revision swap + READY transition.

3. **Handler classes** (`DeployingProvisioningHandler`, `DeployingProgressingHandler`, `DeployingRolledBackHandler`):
   - Verify `execute()` returns expected results.
   - Verify `status_transitions()` returns correct lifecycle states.
   - Verify `post_process()` marks lifecycle and route provisioning.

Given that child PRs #9567 and #9568 will build on this framework, having tests for the framework itself would catch regressions early. Consider adding unit tests in `tests/unit/manager/sokovan/deployment/strategy/` before or alongside the child PRs.

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -2257,6 +2257,74 @@ async def delete_deployment_policy(
...
+
+        Args:
+            endpoint_ids: Set of endpoint IDs to swap revisions for.
+        """
+        if not endpoint_ids:
+            return
+        async with self._begin_session_read_committed() as db_sess:
+            stmt = (
+                sa.update(EndpointRow)
+                .where(EndpointRow.id.in_(endpoint_ids))
```
jopemachine: **[MEDIUM] `complete_deployment_revision_swap` can set `current_revision` to NULL**

This SQL statement sets `current_revision = deploying_revision` unconditionally. If `deploying_revision` is already NULL for any of the endpoint IDs (e.g. due to a previous partial failure, duplicate call, or race condition), this will **overwrite `current_revision` with NULL**, effectively destroying the current revision reference.

This is particularly dangerous in combination with the race condition noted in `coordinator.py` -- if this method is called twice for the same endpoint (e.g. due to lock expiry and re-entry), the second call would set `current_revision = NULL`.

**Suggested fix**: Add a WHERE clause to guard against this:

```python
stmt = (
    sa.update(EndpointRow)
    .where(
        EndpointRow.id.in_(endpoint_ids),
        EndpointRow.deploying_revision.isnot(None),  # Guard
    )
    .values(
        current_revision=EndpointRow.deploying_revision,
        deploying_revision=None,
    )
)
```

This ensures that if `deploying_revision` is already NULL (because the swap already happened), the UPDATE is a no-op rather than corrupting `current_revision`.

jopemachine:
## Code Review Summary -- PR #9566

**Deployment Strategy Evaluation Framework (BEP-1049)**

---

### Overall Assessment

The architecture is well-designed. The evaluator pattern (bulk-load, per-deployment dispatch, aggregate mutations, batch apply) is clean and scalable. The handler hierarchy with composite keys (`lifecycle_type, sub_step`) is a good extension of the existing handler pattern. The separation between route mutation (evaluator) and lifecycle management (handlers) is well-considered.

### Findings Addressed (8 total)

| # | Severity | Location | Finding |
|---|----------|----------|---------|
| 1 | **HIGH** | `coordinator.py:433` | Race condition: `complete_deployment_revision_swap` and `update_endpoint_lifecycle_bulk_with_history` use separate DB sessions. A crash between them leaves deployments in an inconsistent state. |
| 2 | **HIGH** | `coordinator.py:458` | Hardcoded `"Rolling update completed successfully"` message is used for all strategies, including blue-green. |
| 3 | **MEDIUM** | `evaluator.py:85` | `eval_result.errors` is populated but never acted upon by the coordinator. Errored deployments remain stuck in DEPLOYING with infinite retry. |
| 4 | **MEDIUM** | `coordinator.py:406` | Post-process unconditionally treats all deployments as successes, discarding the actual `sub_result` from `handler.execute()`. |
| 5 | **MEDIUM** | `evaluator.py:77` | Skipped deployments (no policy) are never acted upon by the coordinator. They remain stuck in DEPLOYING forever. |
| 6 | **LOW** | `strategy/types.py:25` | `CycleEvaluationResult` docstring says "rolling update" but the type is strategy-agnostic. |
| 7 | **LOW** | (all new files) | Zero test coverage across ~750 lines of new framework code. |
| 8 | **MEDIUM** | `db_source.py:2299` | `complete_deployment_revision_swap` can set `current_revision` to NULL if `deploying_revision` is already NULL (double-call safety). |

### Recommendations

**Before merge (HIGH)**:
- **Finding 1 + 8**: Add a WHERE guard (`deploying_revision IS NOT NULL`) to `complete_deployment_revision_swap` at minimum. Ideally, combine the swap and lifecycle transition into a single transaction.
- **Finding 2**: Parameterize the history message with the actual strategy name.

**Before child PRs merge (MEDIUM)**:
- **Findings 3 + 5**: Add coordinator handling for `eval_result.errors` and `eval_result.skipped` (at minimum, log them; ideally, add bounded retry + error transition).
- **Finding 4**: Capture and pass through the actual `sub_result` to `post_process`.

**Backlog (LOW)**:
- **Finding 6**: Fix docstring.
- **Finding 7**: Add unit tests for the evaluator, coordinator integration, and handler classes.

src/ai/backend/manager/data/deployment/types.py
```diff
@@ -157,6 +157,19 @@ class DeploymentSubStatus(enum.StrEnum):
     """
 
 
+class DeploymentSubStep(DeploymentSubStatus):
+    """Sub-steps for the DEPLOYING lifecycle phase.
```
jopemachine, line 161: Defines sub-steps within the DEPLOYING lifecycle phase. Inherits from `DeploymentSubStatus` so it integrates with the existing sub-status tracking system.

- `PROVISIONING`: New-revision routes are being created; waiting for them to become healthy.
- `PROGRESSING`: Routes are actively being replaced (rolling: gradual replacement, blue-green: health check / promote delay wait).
- `ROLLED_BACK`: All new-revision routes failed; deployment rolled back to previous revision.

src/ai/backend/manager/data/deployment/types.py
```diff
@@ -353,6 +366,7 @@ class DeploymentInfo:
     network: DeploymentNetworkSpec
     model_revisions: list[ModelRevisionSpec]
     current_revision_id: UUID | None = None
+    deploying_revision_id: UUID | None = None
```
jopemachine, line 379: Tracks the revision currently being deployed. The strategy FSM uses this to classify routes as "old" (`current_revision_id`) vs "new" (`deploying_revision_id`). Without this field, the evaluator cannot determine which routes belong to the new deployment.

src/ai/backend/manager/defs.py
```diff
@@ -110,6 +110,7 @@ class LockID(enum.IntEnum):
     LOCKID_DEPLOYMENT_CHECK_PENDING = 226  # For operations checking PENDING sessions
     LOCKID_DEPLOYMENT_CHECK_REPLICA = 227  # For operations checking REPLICA sessions
     LOCKID_DEPLOYMENT_DESTROYING = 228  # For operations destroying deployments
+    LOCKID_DEPLOYMENT_DEPLOYING = 229  # For operations deploying (rolling update) deployments
     # Sokovan target status locks (prevent concurrent operations on same status)
```
jopemachine, line 114: Distributed lock ID for the DEPLOYING lifecycle. Prevents multiple manager instances from concurrently evaluating the same set of DEPLOYING deployments. Follows the existing numbering sequence after `LOCKID_DEPLOYMENT_DESTROYING = 228`.

src/ai/backend/manager/models/endpoint/row.py
```diff
@@ -837,6 +837,7 @@ def _to_deployment_info_from_revision(
                 ),
             ],
             current_revision_id=self.current_revision,
+            deploying_revision_id=self.deploying_revision,
         )
```
jopemachine, line 862: Passes `deploying_revision` from the ORM row into `DeploymentInfo`. This is done in both `_to_deployment_info_from_revision()` and `_to_deployment_info_legacy()` so the evaluator always receives the deploying revision regardless of which conversion path is used.

src/ai/backend/manager/sokovan/deployment/types.py
```diff
@@ -16,6 +16,7 @@ class DeploymentLifecycleType(StrEnum):
     CHECK_REPLICA = "check_replica"
     SCALING = "scaling"
     RECONCILE = "reconcile"
+    DEPLOYING = "deploying"
     DESTROYING = "destroying"
```
jopemachine, line 20: Registers `DEPLOYING` as a lifecycle type that the coordinator processes periodically, alongside existing types (`CHECK_PENDING`, `CHECK_REPLICA`, `SCALING`, `RECONCILE`, `DESTROYING`).

src/ai/backend/manager/sokovan/deployment/strategy/types.py
```diff
@@ -0,0 +1,65 @@
...
+    DeploymentInfo,
+    DeploymentSubStep,
+)
+from ai.backend.manager.models.routing import RoutingRow
+from ai.backend.manager.repositories.base import Creator
+
+
+@dataclass
+class RouteChanges:
+    """Route mutations to apply for a single deployment cycle."""
```
jopemachine, line 24: Route mutations produced by a single deployment's strategy FSM evaluation.

- `scale_out_specs`: `Creator[RoutingRow]` instances for new routes to create.
- `scale_in_route_ids`: IDs of existing routes to terminate (set to `TERMINATING` + `INACTIVE` + `traffic_ratio=0`).

These are collected across all deployments and applied in a single batch DB transaction by the evaluator.

src/ai/backend/manager/sokovan/deployment/strategy/types.py
```diff
@@ -0,0 +1,65 @@
...
+class EvaluationGroup:
+    """Deployments grouped by their sub-step result."""
+
+    sub_step: DeploymentSubStep
+    deployments: list[DeploymentInfo] = field(default_factory=list)
+
+
+@dataclass
+class EvaluationResult:
+    """Aggregate result of evaluating all DEPLOYING deployments."""
```
jopemachine: Aggregate result of evaluating all DEPLOYING deployments in one cycle.

- `groups`: Deployments grouped by sub-step (`PROVISIONING`, `PROGRESSING`). The coordinator looks up the handler for each sub-step and calls `execute()`.
- `completed`: Deployments that satisfied all FSM conditions. The coordinator performs an atomic revision swap + READY transition.
- `completed_strategies`: Maps each completed deployment ID to its strategy type. Included in the history message for observability.
- `skipped`: Deployments with no deployment policy found. History recorded as `SKIPPED`.
- `errors`: Deployments that raised exceptions during FSM evaluation, with error message. History recorded as `NEED_RETRY`, stays `DEPLOYING`.

src/ai/backend/manager/sokovan/deployment/strategy/evaluator.py
```diff
@@ -0,0 +1,160 @@
...
+    """Evaluates DEPLOYING deployments and produces grouped results + route mutations."""
```
jopemachine: Central orchestrator for the DEPLOYING lifecycle. `evaluate()` follows this flow:

1. **Bulk-load** policies and active routes for all endpoints (2 queries, avoids N+1).
2. **Per-deployment**: dispatch to the appropriate strategy FSM (`rolling_update_evaluate` or `blue_green_evaluate`) based on the policy's `strategy` field. Validates that `strategy_spec` matches the expected type.
3. **Aggregate** all `scale_out_specs` and `scale_in_route_ids` from individual results.
4. **Batch apply** route mutations in a single DB transaction via `scale_routes()`.
5. **Classify** deployments into completed / groups (by sub-step) / errors / skipped.

The evaluator does NOT handle lifecycle transitions or history recording — that responsibility stays with the coordinator.

src/ai/backend/manager/sokovan/deployment/strategy/rolling_update.py
```diff
@@ -0,0 +1,26 @@
...
+    DeploymentInfo,
+    RouteInfo,
+)
+from ai.backend.manager.models.deployment_policy import RollingUpdateSpec
+
+from .types import CycleEvaluationResult
+
+
+def rolling_update_evaluate(
+    deployment: DeploymentInfo,
```
jopemachine: Stub for the rolling update strategy FSM. Defines the function signature so the evaluator's dispatch logic is complete and type-safe. The actual implementation is done in a separate branch/issue.

src/ai/backend/manager/sokovan/deployment/strategy/blue_green.py
```diff
@@ -0,0 +1,26 @@
...
+    DeploymentInfo,
+    RouteInfo,
+)
+from ai.backend.manager.models.deployment_policy import BlueGreenSpec
+
+from .types import CycleEvaluationResult
+
+
+def blue_green_evaluate(
+    deployment: DeploymentInfo,
```
jopemachine: Stub for the blue-green strategy FSM. The full implementation (8-step FSM: classify routes → create green → wait provisioning → rollback if all failed → check healthy count → manual promotion gate → promote delay check → atomic promotion) is done in branch BA-3436_2.

src/ai/backend/manager/sokovan/deployment/handlers/deploying.py
```diff
@@ -0,0 +1,189 @@
...
+log = BraceStyleAdapter(logging.getLogger(__name__))
+
+
+# ---------------------------------------------------------------------------
+# In-progress handlers (PROVISIONING / PROGRESSING)
+# ---------------------------------------------------------------------------
+
+
+class DeployingInProgressHandler(DeploymentHandler):
+    """Base handler for in-progress DEPLOYING sub-steps.
```
jopemachine: Base handler for in-progress DEPLOYING sub-steps (PROVISIONING and PROGRESSING both extend this).

- `execute()`: Returns success for all deployments — the evaluator has already applied route mutations, so no additional work is needed here.
- `post_process()`: Re-schedules the `DEPLOYING` lifecycle (so the coordinator checks again next cycle) and triggers `PROVISIONING` on the route controller (so newly created routes get sessions assigned).
- `lock_id = None`: The coordinator already holds the distributed lock via `_process_with_evaluator()`, so the handler does not acquire another one.

src/ai/backend/manager/sokovan/deployment/handlers/deploying.py
```diff
@@ -0,0 +1,189 @@
...
+        )
+
+
+# ---------------------------------------------------------------------------
+# Rolled-back handler
+# ---------------------------------------------------------------------------
+
+
+class DeployingRolledBackHandler(DeploymentHandler):
+    """Handler for DEPLOYING / ROLLED_BACK sub-step.
```
jopemachine: Handler for the `ROLLED_BACK` sub-step.

- `execute()`: Calls `clear_deploying_revision()` to set `deploying_revision = NULL` without touching `current_revision` (the old revision is preserved since the new one failed).
- `status_transitions()`: Transitions `DEPLOYING → READY` with `sub_status = ROLLED_BACK`.
- `post_process()`: No-op — rollback is final, no re-scheduling needed.

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -2257,6 +2257,144 @@ async def delete_deployment_policy(
         async with self._begin_session_read_committed() as db_sess:
             return await execute_purger(db_sess, purger)
 
+    async def fetch_deployment_policies_by_endpoint_ids(
+        self,
```
jopemachine: Bulk-loads `DeploymentPolicyRow` for multiple endpoints in a single query. Returns `{endpoint_id: DeploymentPolicyData}`. Used by the evaluator to avoid N+1 when loading policies for all DEPLOYING deployments.

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -2257,6 +2257,144 @@ async def delete_deployment_policy(
...
+        async with self._db.begin_readonly_session_read_committed() as db_sess:
+            query = sa.select(DeploymentPolicyRow).where(
+                DeploymentPolicyRow.endpoint.in_(endpoint_ids)
+            )
+            result = await db_sess.execute(query)
+            rows = result.scalars().all()
+            return {row.endpoint: row.to_data() for row in rows}
+
+    async def complete_deployment_revision_swap(
+        self,
```
jopemachine: Simple version of revision swap: sets `current_revision = deploying_revision` and `deploying_revision = NULL`. Can be used standalone, but for the main completion flow, `complete_deployment_and_transition_to_ready()` is preferred because it combines this with lifecycle update and history recording in one transaction.

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -2257,6 +2257,144 @@ async def delete_deployment_policy(
...
+                .where(EndpointRow.id.in_(endpoint_ids))
+                .values(
+                    current_revision=EndpointRow.deploying_revision,
+                    deploying_revision=None,
+                )
+            )
+            await db_sess.execute(stmt)
+
+    async def complete_deployment_and_transition_to_ready(
+        self,
```
jopemachine: Atomic completion: revision swap + lifecycle update + history recording in a single transaction.

Key details:
- **Idempotency guard**: `WHERE deploying_revision IS NOT NULL` prevents double-swap if called twice.
- **History merge logic**: Reuses existing `_get_last_deployment_histories_bulk` / `should_merge_with` pattern — if the last history record has the same status, increments `attempts` instead of creating a new row.
- Order: swap first → lifecycle update → history, all within one `_begin_session_read_committed()` scope.

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -2257,6 +2257,144 @@ async def delete_deployment_policy(
...
+                    .where(DeploymentHistoryRow.id.in_(merge_ids))
+                    .values(attempts=DeploymentHistoryRow.attempts + 1)
+                )
+
+            if create_rows:
+                db_sess.add_all(create_rows)
+                await db_sess.flush()
+
+    async def clear_deploying_revision(
+        self,
```
jopemachine: Sets `deploying_revision = NULL` without modifying `current_revision`. Used by `DeployingRolledBackHandler` — on rollback, the new revision is discarded but the old (current) revision stays intact.

src/ai/backend/manager/sokovan/deployment/coordinator.py
```diff
@@ -51,13 +53,20 @@
...
     DestroyingDeploymentHandler,
     ReconcileDeploymentHandler,
     ScalingDeploymentHandler,
 )
+from .strategy.evaluator import DeploymentStrategyEvaluator
 from .types import DeploymentExecutionResult, DeploymentLifecycleType
 
+# Handler key: either a simple lifecycle type or a (lifecycle, sub-step) tuple
+HandlerKey = DeploymentLifecycleType | tuple[DeploymentLifecycleType, DeploymentSubStep]
+
```
jopemachine, line 77: Extends the handler key type to support composite keys.

Existing lifecycle types use a simple key (e.g., `CHECK_PENDING`). DEPLOYING uses composite keys like `(DEPLOYING, PROVISIONING)` to look up sub-step-specific handlers. This avoids adding a second dispatch layer and keeps the handler map flat.

src/ai/backend/manager/sokovan/deployment/coordinator.py
```diff
@@ -314,6 +354,186 @@ async def _handle_status_transitions(
             except Exception as e:
                 log.warning("Failed to send lifecycle notification: {}", e)
 
+    async def _process_with_evaluator(
+        self,
```
jopemachine, line 555: Main processing flow for evaluator-based lifecycles:

1. Acquire distributed lock (`LOCKID_DEPLOYMENT_DEPLOYING`)
2. Load all DEPLOYING deployments
3. Run `evaluator.evaluate()` → strategy FSM evaluation + route mutations applied
4. For each sub-step group, look up and execute the corresponding handler
5. Record status transitions and history via `_handle_status_transitions()`
6. Evaluation errors → record `NEED_RETRY` history (keeps DEPLOYING for next cycle retry)
7. Skipped deployments (no policy) → record `SKIPPED` history with warning log
8. Run `post_process()` for each sub-step handler (re-scheduling, route provisioning triggers)
9. Completed deployments → call `_transition_completed_deployments()` for atomic finalization

src/ai/backend/manager/sokovan/deployment/coordinator.py
```diff
@@ -314,6 +354,186 @@ async def _handle_status_transitions(
...
+            # Transition completed deployments: swap revision and move to READY
+            if eval_result.completed:
+                await self._transition_completed_deployments(
+                    lifecycle_type,
+                    eval_result.completed,
+                    strategies=eval_result.completed_strategies,
+                )
+
+    async def _transition_completed_deployments(
+        self,
```
jopemachine, line 538: Atomic completion for deployments that finished their strategy FSM.

Performs three operations in a single DB transaction via `complete_deployment_and_transition_to_ready()`:
1. **Revision swap**: `current_revision = deploying_revision`, `deploying_revision = NULL` (with idempotency guard)
2. **Lifecycle update**: `DEPLOYING → READY`
3. **History recording**: Records success with the strategy name (e.g., "strategy: rolling")

After the transaction, sends notification events for each completed deployment.

src/ai/backend/manager/sokovan/deployment/coordinator.py
```diff
@@ -386,6 +606,13 @@ def _create_task_specs() -> list[DeploymentTaskSpec]:
                 long_interval=30.0,
                 initial_delay=10.0,
             ),
+            # Deploying (rolling update) - both short and long cycles
+            DeploymentTaskSpec(
+                DeploymentLifecycleType.DEPLOYING,
```
jopemachine: Registers DEPLOYING as a periodic coordinator task.

- `short_interval=5s`: When there are active DEPLOYING deployments, check every 5 seconds.
- `long_interval=30s`: When idle (no DEPLOYING deployments), check every 30 seconds.
- `initial_delay=10s`: Wait 10 seconds after coordinator startup before first check.

src/ai/backend/manager/sokovan/deployment/strategy/evaluator.py
```diff
@@ -0,0 +1,160 @@
...
+            # Group by sub-step
+            if cycle_result.completed:
+                result.completed.append(deployment)
+                result.completed_strategies[deployment.id] = policy.strategy
+            else:
+                group = result.groups.setdefault(
+                    cycle_result.sub_step,
+                    EvaluationGroup(sub_step=cycle_result.sub_step),
+                )
+                group.deployments.append(deployment)
```
HyeockJinKim: Scale out and scale in should operate independently of deploy, right? Is it correct that scale out and scale in are included?

jopemachine: @HyeockJinKim This logic is used when calling `activate_revision` to spin up new routes for the new revision and bring down the existing routes.

Could you clarify what exactly you are expecting?

* Do you prefer not to use the terms scale_in and scale_out?
* Do you want this operation to be delegated to the coordinator?
* Or do you feel the overall structure itself is incorrect?

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -2248,6 +2254,122 @@ async def delete_deployment_policy(
...
+            if merge_ids:
+                await db_sess.execute(
+                    sa.update(DeploymentHistoryRow)
+                    .where(DeploymentHistoryRow.id.in_(merge_ids))
+                    .values(attempts=DeploymentHistoryRow.attempts + 1)
+                )
+
+            if create_rows:
+                db_sess.add_all(create_rows)
+                await db_sess.flush()
```
HyeockJinKim: The upper-level Coordinator will already be storing the history, so please verify that no duplicates are saved here.

src/ai/backend/manager/sokovan/deployment/strategy/evaluator.py
```diff
@@ -0,0 +1,157 @@
...
+from .types import CycleEvaluationResult, EvaluationGroup, EvaluationResult, RouteChanges
+
+log = BraceStyleAdapter(logging.getLogger(__name__))
+
+
+class DeploymentStrategyEvaluator:
+    """Evaluates DEPLOYING deployments and produces grouped results + route mutations."""
+
+    def __init__(self, deployment_repo: DeploymentRepository) -> None:
+        self._deployment_repo = deployment_repo
```
HyeockJinKim: I remember it was about storing values for each strategy and using them later, so please confirm this for me.

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
jopemachine: This is update of `DeploymentHistoryRow`, not creation.

src/ai/backend/manager/sokovan/deployment/strategy/evaluator.py
jopemachine: I have introduced the `BaseDeploymentStrategy` and `DeploymentStrategyRegistry` types and updated the code accordingly.

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -2087,24 +2087,30 @@ async def update_endpoint(
 
             return row.to_deployment_info()
 
-    async def update_current_revision(
+    async def start_deploying_revision(
```
fregataa: Should `update_current_revision()` be deleted?

src/ai/backend/manager/sokovan/deployment/coordinator.py
```diff
@@ -314,6 +366,205 @@ async def _handle_status_transitions(
...
+                        message=f"Evaluation error: {reason}",
+                        from_status=EndpointLifecycle.DEPLOYING,
+                        to_status=None,
+                        sub_steps=[],
+                    )
+                    for deployment, reason in eval_result.errors
+                ]
+                await self._deployment_repository.update_endpoint_lifecycle_bulk_with_history(
+                    [], BulkCreator(specs=error_history_specs)
+                )
```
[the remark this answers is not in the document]
fregataa: Please check this comment and resolve if it is false

src/ai/backend/manager/sokovan/deployment/coordinator.py
```diff
@@ -314,6 +366,205 @@ async def _handle_status_transitions(
...
+                        message="No deployment policy found",
+                        from_status=EndpointLifecycle.DEPLOYING,
+                        to_status=None,
+                        sub_steps=[],
+                    )
+                    for deployment in eval_result.skipped
+                ]
+                await self._deployment_repository.update_endpoint_lifecycle_bulk_with_history(
+                    [], BulkCreator(specs=skipped_history_specs)
+                )
```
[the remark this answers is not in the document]
fregataa: Same here

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -2248,6 +2254,144 @@ async def delete_deployment_policy(
...
+            endpoint_ids: Set of endpoint IDs to swap revisions for.
+        """
+        if not endpoint_ids:
+            return
+        async with self._begin_session_read_committed() as db_sess:
+            stmt = (
+                sa.update(EndpointRow)
+                .where(EndpointRow.id.in_(endpoint_ids))
+                .values(
+                    current_revision=EndpointRow.deploying_revision,
```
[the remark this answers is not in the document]
fregataa: "Setting current_revision to null if deploying_revision is null" is an expected behavior?

src/ai/backend/manager/sokovan/deployment/coordinator.py
```diff
@@ -314,6 +387,241 @@ async def _handle_status_transitions(
...
+                        sub_step,
+                        DeploymentExecutionResult(successes=group.deployments),
+                    )
+                    await handler.post_process(actual_result)
+                except Exception as e:
+                    log.error(
+                        "Error during post-processing for sub-step {}: {}",
+                        sub_step.value,
+                        e,
+                    )
```
fregataa: What's the purpose of this post process?

src/ai/backend/manager/sokovan/deployment/coordinator.py
```diff
@@ -314,6 +387,241 @@ async def _handle_status_transitions(
...
+            len(changes.rollout_specs),
+            len(changes.drain_route_ids),
+        )
+
+    async def _transition_completed_deployments(
+        self,
+        lifecycle_type: DeploymentLifecycleType,
+        completed: list[DeploymentInfo],
+        strategies: dict[UUID, DeploymentStrategy],
+        records: Mapping[UUID, ExecutionRecord],
```
fregataa: Those are all necessary to determine status?

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
jopemachine: It doesn’t seem necessary to have a function that updates only current_revision.

When the status becomes deploying, we should update deploying_revision rather than current_revision.
Once the deployment finishes, both current_revision and deploying_revision should be updated.

src/ai/backend/manager/sokovan/deployment/coordinator.py
jopemachine: I think there appears to be an issue with the current process for handling substeps.
(Substeps should be handled in a more generalized way and should not be coupled with the evaluator implementation.)
I will review and refactor it, then follow up with another reply.

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -2143,24 +2142,30 @@ async def update_endpoint(
 
             return row.to_deployment_info()
 
-    async def update_current_revision(
+    async def start_deploying_revision(
```
HyeockJinKim: Don't rename it to start_... even though it doesn't start with db.

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -2143,24 +2142,30 @@ async def update_endpoint(
...
+            # Set deploying_revision and transition to DEPLOYING
             update_query = (
                 sa.update(EndpointRow)
                 .where(EndpointRow.id == endpoint_id)
-                .values(current_revision=revision_id)
+                .values(
+                    deploying_revision=revision_id,
+                    lifecycle_stage=EndpointLifecycle.DEPLOYING,
+                )
             )
```
HyeockJinKim: Wasn't there also a substatus?

src/ai/backend/manager/sokovan/deployment/handlers/deploying.py
```diff
@@ -0,0 +1,308 @@
...
+
+    @classmethod
+    @override
+    def target_statuses(cls) -> list[EndpointLifecycle]:
+        return [EndpointLifecycle.DEPLOYING]
+
+    @classmethod
+    @override
+    def status_transitions(cls) -> DeploymentStatusTransitions:
+        return DeploymentStatusTransitions(success=None, failure=None)
```
HyeockJinKim: From here on, it seems like something strange is happening.

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
jopemachine: I removed it in thess commits [fix: Remove substeps](https://github.com/lablup/backend.ai/pull/9566/commits/42446f4c15d2cf5d224ab5ade7b957cf21d193d8)

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
jopemachine: Are you suggesting that method names should not start with start?
What would be the reason for that?

jopemachine:
## Code Review: PR #9566 — Deployment Strategy Evaluation Framework

**Reviewer**: AI Code Reviewer
**Branch**: BA-4821 -> main
**Risk Level**: Medium

---

### Overall Assessment

The PR introduces a well-structured deployment strategy evaluation framework with a flat handler registry, pre-step evaluation pattern, and failure classification. The architecture is clean and extensible. However, there are several issues across security, correctness, race conditions, and code quality that should be addressed before merging.

---

### CRITICAL Findings

None identified. The PR does not introduce direct security vulnerabilities like injection, authentication bypass, or credential exposure.

---

### HIGH Findings

#### H1. Race condition: pre-step and handlers use separate DB sessions (no distributed lock)

**File**: `src/ai/backend/manager/sokovan/deployment/coordinator.py`, lines 335-340

The DEPLOYING lifecycle path runs `_run_pre_step()` then iterates sub-step handlers **without** acquiring a distributed lock. The pre-step updates `sub_step` in DB, then each handler independently queries deployments by sub_step. Between the pre-step and handler execution (or between two coordinator cycles running concurrently), another coordinator instance could:

1. Run its own pre-step and re-assign sub_steps
2. Process the same deployments concurrently

The simple lifecycle handlers acquire `handler.lock_id`, but the DEPLOYING sub-step handlers all declare `lock_id = None`. The `LOCKID_DEPLOYMENT_DEPLOYING = 229` constant is defined but never used.

**Recommendation**: Wrap the entire DEPLOYING lifecycle (pre-step + all sub-step handlers) in a single distributed lock using `LOCKID_DEPLOYMENT_DEPLOYING`. Example:

```python
else:
    async with AsyncExitStack() as stack:
        lock_lifetime = self._config_provider.config.manager.session_schedule_lock_lifetime
        await stack.enter_async_context(
            self._lock_factory(LockID.LOCKID_DEPLOYMENT_DEPLOYING, lock_lifetime)
        )
        await self._run_pre_step(lifecycle_type)
        for sub_step in sub_steps:
            handler = self._registry.handlers[(lifecycle_type, sub_step)]
            await self._run_handler(handler, lifecycle_type)
```

#### H2. `DeploymentInfo` is not frozen but is mutated in place

**File**: `src/ai/backend/manager/data/deployment/types.py`, line 368  
**File**: `src/ai/backend/manager/sokovan/deployment/coordinator.py`, lines 375-382

The `data/` layer CLAUDE.md explicitly states: *"All data classes MUST be `@dataclass(frozen=True)`"*. `DeploymentInfo` is a mutable `@dataclass` (not frozen), and the coordinator mutates `phase_attempts` and `phase_started_at` in-place after loading from DB. This violates the data layer's immutability contract.

**Recommendation**: Either make `DeploymentInfo` frozen and pass phase info through a separate wrapper, or document this as an intentional exception. The current approach risks subtle bugs if the same `DeploymentInfo` instance is referenced elsewhere.

#### H3. `DeployingProgressingHandler` silently drops PROGRESSING deployments

**File**: `src/ai/backend/manager/sokovan/deployment/handlers/deploying.py`, lines 297-308

The `execute()` method uses a `match` statement that processes COMPLETED and ROLLED_BACK deployments but falls through to `pass` for PROGRESSING. Since these deployments are neither in `successes` nor `errors`, the coordinator's `_handle_status_transitions()` will not record any history or perform any lifecycle transition for them. This means:
- No history entry is recorded for PROGRESSING deployments in this handler
- The `status_transitions().success = READY` would transition COMPLETED to READY (correct), but if there are no successes, no batch update is generated at all

This is likely intentional (PROGRESSING deploys stay in DEPLOYING), but the handler's `status_transitions()` returns `success=READY` which is misleading. If a PROGRESSING deployment somehow ends up as the only deployment processed, no action is taken but `post_process()` still re-schedules.

**Recommendation**: Add a comment or consider returning PROGRESSING deployments in a `skipped` field to make the intent explicit and ensure history recording captures the re-scheduling.

#### H4. Timezone-naive comparison risk in failure classification

**File**: `src/ai/backend/manager/sokovan/deployment/coordinator.py`, lines 646-651

```python
elapsed = (current_time - deployment.phase_started_at).total_seconds()
```

`current_time` comes from `get_db_now()` (which returns `sa.func.now()` — the DB server's current time). `phase_started_at` is set from `history.created_at`. If the DB returns timezone-aware datetimes but `history.created_at` is timezone-naive (or vice versa), this subtraction will raise a `TypeError`. The code does not normalize timezones.

**Recommendation**: Ensure both values use the same timezone. Consider applying `.replace(tzinfo=UTC)` or `astimezone(UTC)` to both before comparison.

---

### MEDIUM Findings

#### M1. `update_sub_steps` uses N separate UPDATE statements

**File**: `src/ai/backend/manager/repositories/deployment/db_source/db_source.py`, lines 2173-2182 (in diff)

```python
for sub_step, endpoint_ids in sub_step_map.items():
    if not endpoint_ids:
        continue
    stmt = sa.update(EndpointRow).where(EndpointRow.id.in_(endpoint_ids)).values(sub_step=sub_step)
    await db_sess.execute(stmt)
```

Each sub_step value generates a separate UPDATE. With 4 possible sub_steps, this means up to 4 round-trips per cycle. While not critical, this could be optimized using a `CASE WHEN` expression for a single UPDATE.

#### M2. Pre-step queries DEPLOYING deployments again (redundant DB query)

**File**: `src/ai/backend/manager/sokovan/deployment/handlers/deploying.py`, lines 112-114  
**File**: `src/ai/backend/manager/sokovan/deployment/coordinator.py`, lines 354-362

The pre-step's `run()` queries all DEPLOYING deployments. Then each sub-step handler in `_run_handler()` also queries DEPLOYING deployments filtered by sub_step. This means the same endpoints are loaded from DB at least twice per cycle (once in pre-step, once in handler). For many concurrent deployments, this adds unnecessary latency.

**Recommendation**: Consider having the pre-step return its results and passing the already-loaded deployments to `_run_handler()`.

#### M3. `_record_route_changes` calls `DeploymentRecorderContext.current_pool()` outside recorder scope

**File**: `src/ai/backend/manager/sokovan/deployment/strategy/evaluator.py`, lines 107-126

The evaluator's `_record_route_changes` accesses `DeploymentRecorderContext.current_pool()`. However, the evaluator is called from the pre-step's `run()`, which runs **before** the coordinator enters any `DeploymentRecorderContext.scope()` in `_run_handler()`. If no scope is active, `current_pool()` will either raise an error or return stale data.

**Recommendation**: Verify that `DeploymentRecorderContext` has an active scope when the evaluator runs, or move recording to within the handler scope. This may be the most impactful bug in the PR.

#### M4. Missing `__init__.py` BUILD file for new `strategy/` directory

**File**: `src/ai/backend/manager/sokovan/deployment/strategy/`

Per the project's CLAUDE.md: "MUST add BUILD files to new test directories." While this rule mentions test directories, new source packages like `strategy/` with `__init__.py`, `types.py`, `evaluator.py`, `blue_green.py`, and `rolling_update.py` may also need pants BUILD targets depending on the project's build configuration.

#### M5. `DeploymentSubStep` inherits from `DeploymentSubStatus`

**File**: `src/ai/backend/manager/data/deployment/types.py`, lines 162-175

```python
class DeploymentSubStep(DeploymentSubStatus):
    PROVISIONING = "provisioning"
    PROGRESSING = "progressing"
    COMPLETED = "completed"
    ROLLED_BACK = "rolled_back"
```

`DeploymentSubStep` extends `DeploymentSubStatus` (a `StrEnum`). Since `DeploymentSubStep` values are distinct from the parent's values, `isinstance(x, DeploymentSubStatus)` will be `True` for all `DeploymentSubStep` values. This is used in `_derive_handler_key` to distinguish sub-step from non-sub-step handlers. While this works, the inheritance relationship is semantically confusing — a "sub-step" is-a "sub-status" but their enum members are completely disjoint. Consider using composition or a shared protocol instead.

#### M6. `_derive_handler_key` uses only the first `target_statuses()` entry

**File**: `src/ai/backend/manager/sokovan/deployment/coordinator.py`, lines 295-312

```python
first_sub = statuses[0].sub_status
```

`DeployingProgressingHandler.target_statuses()` returns three entries (PROGRESSING, COMPLETED, ROLLED_BACK). Only the first (PROGRESSING) is used as the registry key. This means COMPLETED and ROLLED_BACK are effectively handled by the PROGRESSING handler, which works but is implicit. The `sub_steps_for()` method derives sub-steps from registered keys, so it will only see PROVISIONING and PROGRESSING — never COMPLETED or ROLLED_BACK. This means the coordinator will never iterate over COMPLETED/ROLLED_BACK as separate sub-step handlers, which is the intended design (the PROGRESSING handler covers them). But if someone later registers a separate COMPLETED handler, the current key derivation would silently conflict.

#### M7. `update_sub_steps` within its own session may conflict with pre-step's route changes

**File**: `src/ai/backend/manager/sokovan/deployment/handlers/deploying.py`, lines 119-120

```python
await self._update_sub_steps(eval_result)
await self._apply_route_changes(eval_result)
```

These two operations use separate DB sessions (each method calls the repository, which opens its own `_begin_session_read_committed()`). If `_apply_route_changes` fails after `_update_sub_steps` succeeds, sub_steps will be updated but routes will not be created/drained. This partial state could cause the handler to attempt processing deployments that haven't had their routes set up correctly.

**Recommendation**: Consider executing both in a single transaction for atomicity.

---

### LOW Findings

#### L1. `build_lifecycle_notification_event` moved from coordinator method to module-level function

While this is fine architecturally, the function is now exported in `__init__.py` alongside handler classes, making it a public API of the handlers package. It's purely a coordinator utility and should perhaps live in its own module or in the coordinator.

#### L2. Missing newline at end of `changes/9566.feature.md`

Minor file formatting issue.

#### L3. Hardcoded 30-minute timeout for DEPLOYING

**File**: `src/ai/backend/manager/sokovan/deployment/coordinator.py`, line 86

```python
EndpointLifecycle.DEPLOYING: 1800.0,  # 30 minutes
```

This timeout should ideally be configurable per deployment policy or strategy, as blue-green deployments may take longer than rolling updates.

#### L4. Both strategy implementations raise `NotImplementedError`

**Files**: `strategy/rolling_update.py`, `strategy/blue_green.py`

Both `evaluate_cycle()` methods raise `NotImplementedError`. This is expected for a framework PR, but means the DEPLOYING lifecycle will crash at runtime if triggered. Consider adding a guard in the evaluator or a warning log.

---

### Summary

| Severity | Count | Key Items |
|----------|-------|-----------|
| CRITICAL | 0 | - |
| HIGH | 4 | Race condition (H1), frozen dataclass violation (H2), silent drop (H3), timezone (H4) |
| MEDIUM | 7 | Redundant queries (M2), recorder scope issue (M3), partial transaction (M7) |
| LOW | 4 | Timeout configurability (L3), NotImplementedError guard (L4) |

**Most impactful issues to address**: H1 (race condition with missing distributed lock) and M3 (recorder context scope issue during evaluator execution) are the most likely to cause production bugs.

jopemachine:
The PR has grown larger than expected, so I will split the work and proceed in smaller parts.

PR #9569, opened 2026-03-02
feat(BA-4765): Deprioritize agents with repeated session start failures

### LINKED ISSUE #9482: Deprioritize agents with repeated session start failures in scheduler
Background
When a specific agent repeatedly fails to start sessions (failed-to-start), the current scheduler only deprioritizes the session (DEPRIORITIZE_AMOUNT=10), not the agent. This causes the same failing agent to be selected again for subsequent scheduling attempts of similar workloads.

Scope
- Introduce per-agent failure counter tracking in the scheduler (e.g. failed-to-start count within a time window)
- Add threshold-based agent deprioritization/exclusion logic in agent selection strategies (concentrated, dispersed, etc.)
- Auto-recover agent priority after a cooldown period or successful session start
- Expose threshold/cooldown as scaling group configuration

Key Files
- manager/sokovan/scheduler/coordinator.py: _classify_failures()
- manager/sokovan/scheduler/handlers/lifecycle/deprioritize_sessions.py
- manager/sokovan/scheduler/provisioner/selectors/selector.py: AgentInfo
- manager/sokovan/scheduler/provisioner/provisioner.py: _allocate_workload()
- manager/data/agent/types.py: AgentStatus

JIRA Issue: BA-4765

jopemachine:
Resolves #9482 (BA-4765).

## Summary

- When a session fails to start on an agent (kernel creation RPC failure, image pull timeout, etc.), the scheduler now records those agents in Valkey and avoids selecting them on retry
- Failed agent tracking is **per-session scoped** — other sessions are unaffected and can still use those agents
- Uses Valkey SET with 1-hour TTL for automatic cleanup, requiring no DB migration

## Two Deprioritization Mechanisms (and Why Both Are Needed)

This PR introduces agent-level deprioritization, which is orthogonal to the existing session-level `DEPRIORITIZING` status:

| | Existing `DEPRIORITIZING` | This PR (agent-level) |
|---|---|---|
| **Acts on** | Session priority relative to other sessions | Specific agent × session combinations |
| **Mechanism** | Lowers `session.priority` score | Records failed agent IDs in Valkey |
| **Triggered when** | Session can't be scheduled at all (resource shortage) | Kernel start fails on a specific agent |
| **Effect** | Session is pushed behind other sessions in the queue | Provisioner skips the failed agent and tries another |

Without this PR, a session that fails to start on Agent X would go back to PENDING and the provisioner could select Agent X again — causing a failure loop. The `DEPRIORITIZING` status alone doesn't help here because it only controls scheduling order between sessions, not which agent is picked for a given session.

## Relationship to Concentrated / Dispersed Scheduling Policies

Failed agent filtering and the concentrated/dispersed placement policies are **independent concerns** that operate sequentially within a single agent selection call:

```
For each kernel in the session:
  1. Architecture filter    → arch_compatible_trackers
  2. Resource availability  → compatible_trackers
  3. Failed agent filter    → candidate_trackers   ← this PR
  4. Concentrated/Dispersed strategy               ← placement policy
```

- **Failed agent filtering** decides *which agents enter the candidate pool*
- **Concentrated/Dispersed** decides *how to pick from that pool*

For multi-kernel cluster sessions, each kernel goes through this pipeline in sequence. After each selection, the chosen agent's remaining resources are updated via a diff tracker, so subsequent kernel selections see the updated state. The placement policy then operates on the already-narrowed candidate pool for each kernel.

## How It Works

1. **Recording failures:** Two recording points capture failed agents:
   - **Launcher** (`_start_single_session`): after `asyncio.gather` returns, identifies agents with failed kernel creation RPCs
   - **Coordinator** (`_apply_kernel_pending_resets`): before resetting kernels to PENDING (timeout path), queries current agent assignments and records them

2. **Loading at scheduling time:** The provisioner loads each session's failed agents from Valkey (`session:failed_agents:{session_id}`) and attaches them to `SessionWorkload.failed_agent_ids`

3. **Filtering in agent selection:** The selector adds a third filtering pass that removes previously-failed agents from the compatible list. If **all** compatible agents have failed, the filter is skipped to avoid blocking scheduling entirely

### DIFF
--- a/src/ai/backend/common/clients/valkey_client/valkey_schedule/client.py
+++ b/src/ai/backend/common/clients/valkey_client/valkey_schedule/client.py
@@ -387,6 +387,9 @@ class ValkeyScheduleClient:
         for session_id in session_ids:
             batch.smembers(self._get_session_failed_agents_key(session_id))
         batch_result = await self._client.client.exec(batch, raise_on_error=True)
+        # batch_result is None only for atomic batches (transactions) when a WATCH-guarded key
+        # was modified before EXEC — non-atomic pipelines (is_atomic=False) never return None.
+        # This guard is kept for defensive typing since exec() is annotated Optional[List[...]].
         if batch_result is None:
             return [frozenset() for _ in session_ids]
         return [
--- a/src/ai/backend/manager/sokovan/scheduler/provisioner/selectors/selector.py
+++ b/src/ai/backend/manager/sokovan/scheduler/provisioner/selectors/selector.py
@@ -7,6 +7,7 @@ the row-based implementation details of the legacy selectors.
 
 from __future__ import annotations
 
+import logging
 from abc import ABC, abstractmethod
 from collections import defaultdict
 from collections.abc import Mapping, Sequence
@@ -22,6 +23,7 @@ from ai.backend.common.types import (
     SessionId,
     SessionTypes,
 )
+from ai.backend.logging.utils import BraceStyleAdapter
 
 from .exceptions import (
     ContainerLimitExceededError,
@@ -33,6 +35,8 @@ from .exceptions import (
 if TYPE_CHECKING:
     from ai.backend.manager.repositories.scheduler.types.agent import AgentMeta
     from ai.backend.manager.sokovan.data import AgentOccupancy
+
+log = BraceStyleAdapter(logging.getLogger(__spec__.name))
 
 
 @dataclass
@@ -449,7 +453,25 @@ class AgentSelector:
                 if tracker.original_agent.agent_id not in criteria.failed_agent_ids
             ]
             if non_failed:
+                excluded = [
+                    tracker.original_agent.agent_id
+                    for tracker in compatible_trackers
+                    if tracker.original_agent.agent_id in criteria.failed_agent_ids
+                ]
+                log.debug(
+                    "failed-agent filter(session:{}): excluding {} → candidates: {}",
+                    criteria.session_metadata.session_id,
+                    excluded,
+                    [tracker.original_agent.agent_id for tracker in non_failed],
+                )
                 candidate_trackers = non_failed
+            else:
+                log.debug(
+                    "failed-agent filter(session:{}): all {} compatible agents have failed, "
+                    "skipping filter to avoid blocking",
+                    criteria.session_metadata.session_id,
+                    len(compatible_trackers),
+                )
             # If ALL compatible agents have failed, keep all of them to avoid blocking
 
         # Use strategy to select from candidates
--- a/src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
@@ -2286,7 +2286,7 @@ class ScheduleDBSource:
             result = await db_sess.execute(stmt)
             return cast(CursorResult[Any], result).rowcount
 
-    async def get_kernel_agent_ids_for_sessions(
+    async def get_agent_ids_for_sessions(
         self, session_ids: list[SessionId]
     ) -> dict[SessionId, list[AgentId]]:
         """
--- a/src/ai/backend/manager/repositories/scheduler/repository.py
+++ b/src/ai/backend/manager/repositories/scheduler/repository.py
@@ -502,11 +502,11 @@ class SchedulerRepository:
         return await self._db_source.update_kernel_status_terminated(kernel_id, reason, exit_code)
 
     @scheduler_repository_resilience.apply()
-    async def get_kernel_agent_ids_for_sessions(
+    async def get_agent_ids_for_sessions(
         self, session_ids: list[SessionId]
     ) -> dict[SessionId, list[AgentId]]:
         """Get agent IDs assigned to kernels for the given sessions."""
-        return await self._db_source.get_kernel_agent_ids_for_sessions(session_ids)
+        return await self._db_source.get_agent_ids_for_sessions(session_ids)
 
     @scheduler_repository_resilience.apply()
     async def reset_kernels_to_pending_for_sessions(
--- a/src/ai/backend/manager/sokovan/scheduler/coordinator.py
+++ b/src/ai/backend/manager/sokovan/scheduler/coordinator.py
@@ -1270,7 +1270,7 @@ class ScheduleCoordinator:
 
         # Record current agent assignments before they are cleared by the reset.
         # This is best-effort: Valkey issues must not block kernel resets.
-        agent_ids_by_session = await self._repository.get_kernel_agent_ids_for_sessions(session_ids)
+        agent_ids_by_session = await self._repository.get_agent_ids_for_sessions(session_ids)
         record_sessions: list[SessionId] = []
         record_tasks: list[Awaitable[None]] = []
         for session_id, agent_ids in agent_ids_by_session.items():


jopemachine (commented):
## Code Review: Agent Deprioritization on Session Start Failures

Overall this is a well-structured feature with clean separation between recording, loading, and filtering. The Valkey-based approach with TTL cleanup is a good design choice. However, I found a few issues that should be addressed before merging.

### Findings Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | **HIGH** | `selector.py` | Deprioritization pass filters out designated agents, causing false `NoAvailableAgentError` |
| 2 | **MEDIUM** | `provisioner.py` | Sequential Valkey round-trips for N pending sessions (O(N) latency) |
| 3 | **MEDIUM** | `client.py` | `clear_session_failed_agents` defined but never called |
| 4 | **LOW** | tests | Minimal test coverage for new logic paths |

Details are posted as inline comments below.

src/ai/backend/manager/sokovan/scheduler/provisioner/selectors/selector.py
```diff
@@ -421,6 +423,17 @@ async def _select_agent_tracker_for_requirements(
...
+        # Third pass: deprioritize agents that previously failed for this session
+        if criteria.failed_agent_ids:
+            non_failed = [
+                t
+                for t in compatible_trackers
+                if t.original_agent.agent_id not in criteria.failed_agent_ids
+            ]
+            if non_failed:
+                compatible_trackers = non_failed
+            # If ALL compatible agents have failed, keep all of them to avoid blocking
```
jopemachine: **[HIGH] Deprioritization pass can filter out designated agents, causing false `NoAvailableAgentError`**

The "third pass" that removes failed agents from `compatible_trackers` is executed **before** the designated agent check (line 438). If a superadmin explicitly designates an agent for a session, and that agent previously failed, the deprioritization pass removes it from the list. The designated agent check then iterates only over the filtered `compatible_trackers` and raises `NoAvailableAgentError` because the designated agent is gone.

**Reproduction scenario:**
1. Session S is assigned to designated agent A
2. Agent A fails (kernel creation RPC timeout)
3. Session S goes back to PENDING, agent A is recorded as failed in Valkey
4. On retry, the scheduler loads `failed_agent_ids = {A}`
5. Third pass removes agent A from `compatible_trackers`
6. Designated agent check cannot find A -> raises `NoAvailableAgentError`
7. Session is permanently stuck despite the superadmin's explicit agent designation

**Suggested fix:** Move the designated agent check **before** the deprioritization pass:

```python
# Handle designated agent FIRST (user's explicit choice takes precedence)
if designated_agent_ids:
    for tracker in compatible_trackers:
        if tracker.original_agent.agent_id in designated_agent_ids:
            return tracker
    # ... existing error handling ...

# THEN deprioritize for non-designated sessions
if criteria.failed_agent_ids:
    non_failed = [
        t for t in compatible_trackers
        if t.original_agent.agent_id not in criteria.failed_agent_ids
    ]
    if non_failed:
        compatible_trackers = non_failed
```

src/ai/backend/manager/sokovan/scheduler/provisioner/provisioner.py
```diff
@@ -172,9 +173,18 @@ async def schedule_scaling_group(
...
             session.to_session_workload() for session in scheduling_data.pending_sessions.sessions
         ]
+
+        # Load per-session failed agents from Valkey for retry deprioritization
+        workloads: list[SessionWorkload] = []
+        for wl in base_workloads:
+            failed_agents = await self._valkey_schedule.get_session_failed_agents(wl.session_id)
+            if failed_agents:
+                wl = dataclasses.replace(wl, failed_agent_ids=failed_agents)
+            workloads.append(wl)
```
jopemachine: **[MEDIUM] Sequential Valkey round-trips for N pending sessions adds O(N) latency to scheduling loop**

This loop makes N sequential `await` calls to Valkey, one per pending session. In production with large queues (100+ pending sessions), this adds significant latency to the scheduling hot path. Each round-trip is ~0.1-1ms over the network, so 100 sessions could add 10-100ms of pure network wait time.

**Suggested fix:** Use `asyncio.gather` to parallelize:

```python
# Load per-session failed agents from Valkey for retry deprioritization
failed_agents_list = await asyncio.gather(*[
    self._valkey_schedule.get_session_failed_agents(wl.session_id)
    for wl in base_workloads
])
workloads = [
    dataclasses.replace(wl, failed_agent_ids=fa) if fa else wl
    for wl, fa in zip(base_workloads, failed_agents_list, strict=True)
]
```

Or even better, consider a batch Valkey pipeline approach using `Batch` (similar to `get_queue_positions` in the same client) to reduce this to a single round-trip.

src/ai/backend/common/clients/valkey_client/valkey_schedule/client.py
```diff
@@ -334,6 +335,65 @@ async def get_queue_positions(self, session_ids: Sequence[SessionId]) -> list[in
...
+
+    @valkey_schedule_resilience.apply()
+    async def clear_session_failed_agents(self, session_id: SessionId) -> None:
+        """
+        Clear the failed agents record for a specific session.
+
+        :param session_id: The session to clear
+        """
+        key = self._get_session_failed_agents_key(session_id)
+        await self._client.client.delete([key])
```
jopemachine: **[MEDIUM] `clear_session_failed_agents` is defined but never called anywhere in the codebase**

While the TTL (1 hour) provides eventual cleanup, there is no explicit cleanup when a session successfully starts. This means stale failure data persists for up to 1 hour after a successful retry.

Consider either:
- Adding a call to `clear_session_failed_agents` on successful session start in the launcher (after verifying all `asyncio.gather` results succeeded)
- Or removing the dead method and documenting that TTL-based cleanup is the intentional design choice (to avoid the extra Valkey call on the happy path)

tests/unit/manager/sokovan/scheduler/provisioner/test_provisioner.py
```diff
@@ -227,6 +227,7 @@ def test_provisioner(
     """Create SessionProvisioner with mock dependencies."""
     valkey_schedule = MagicMock()
     valkey_schedule.set_pending_queue = AsyncMock(return_value=None)
```
jopemachine, line 229: **[LOW] Minimal test coverage for new logic paths**

The only test change is adding a mock return value to prevent existing tests from breaking. No new tests are added for:

1. **Selector deprioritization logic**: The "third pass" has multiple code paths (some agents failed, all agents failed fallback, interaction with designated agents) that are untested.
2. **Valkey client methods**: `record_session_failed_agents`, `get_session_failed_agents`, and `clear_session_failed_agents` have no unit tests.
3. **Provisioner integration**: Loading of failed agents from Valkey and attachment to `SessionWorkload` is untested.

Consider adding at least:
- A selector test where `failed_agent_ids` is non-empty and verifies the correct agent is selected
- A selector test where **all** compatible agents are in `failed_agent_ids` and verifies the fallback behavior (all agents kept)
- A selector test with both `failed_agent_ids` and `designated_agent_ids` to verify the interaction (this would catch the HIGH issue above)

jopemachine:
> Wouldn't it be better to store the agent priority in the database rather than working with Redis?

I considered treating agent priority as a kind of ephemeral cache and storing it in Redis. What would be the advantages of storing it in the database instead? @HyeockJinKim

PR #9571, opened 2026-03-02
fix: default group gid in AccountManagerConfig

Lubov66:


Lubov66:
CI fails in assign-pr-number with Input required and not supplied: token for actions/checkout@v6.

Seems unrelated to this PR. Should I patch the workflow?

HyeockJinKim:
It appears to be a CI issue, and since the PR itself seems fine, I will merge it. cc @fregataa

PR #9578, opened 2026-03-03
test(BA-4827): Add unit tests for SSH stage provisioners

seedspirit:
## Summary
- Follow up of BA-3128 (#8783), adding unit tests in ssh provisioners
- Add unit tests for SSH stage provisioner classes (`SSHProvisioner`, `ContainerSSHProvisioner`) in `src/ai/backend/agent/stage/kernel_lifecycle/docker/`
- Tests cover UID/GID override priority, UID_MATCH feature behavior, root vs non-root chown, SSH keypair file creation with permissions, port mapping, container SSH dir setup, existing `id_rsa` skip logic, and `.ssh` mount detection
- Uses `tmp_path` and `unittest.mock.patch(os.geteuid)` — no Docker infrastructure required

## Test plan
- [x] All unit tests pass locally via `pants test`
- [x] `pants fmt`, `pants fix`, `pants lint` all pass
- [ ] CI pipeline passes

Resolves BA-4827

### DIFF
--- a/tests/unit/agent/stage/kernel_lifecycle/docker/test_ssh_provisioners.py
+++ b/tests/unit/agent/stage/kernel_lifecycle/docker/test_ssh_provisioners.py
@@ -194,11 +194,13 @@ class TestSSHProvisionerWriteConfigFunc:
 
 
 class TestContainerSSHProvisioner:
-    _DEFAULT_AGENT_CONFIG = AgentConfig(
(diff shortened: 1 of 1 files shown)


fregataa (commented):
Do the test create a real path and file or just mock?

seedspirit:
@fregataa Test use pytest's `tmp_path` which create unique real path per test

PR #9594, opened 2026-03-03
feat(BA-4837): Add extra_headers support to v2 BackendAIAnonymousClient

fregataa:
## Summary
- Add `extra_headers: Mapping[str, str] | None = None` parameter to `_request()`, `_typed_request()`, `authorize()`, `signup()`, and `update_password_no_auth()` on `BackendAIAnonymousClient`
- Use `multidict.CIMultiDict` in `_build_headers()` to ensure case-insensitive header merging, preventing duplicate headers from case mismatches
- Merge extra headers into the outgoing request after `_build_headers()`, enabling per-request header injection (e.g. `X-Forwarded-For/Host/Proto`)
- Add tests verifying extra headers are merged and default behavior is unchanged

## Test plan
- [x] `test_extra_headers_merged_into_request` — forwarding headers appear in outgoing request
- [x] `test_request_without_extra_headers` — no forwarding headers when `extra_headers` is omitted


jopemachine:
If the set of headers allowed in extra_headers is predefined, how about defining an explicit type instead of using a generic Mapping?

PR #9595, opened 2026-03-03
feat(BA-4838): add NoAuth strategy and BackendAIClientRegistry to webserver

fregataa:
## Summary
- Add `NoAuth(AuthStrategy)` to `client/v2/auth.py` that returns empty headers from `sign()`, for use by the webserver which has no real keypair
- Replace `anon_client_ctx` / `app["anon_client"]` with `client_registry_ctx` / `app["client_registry"]` using `BackendAIClientRegistry.create(config, NoAuth())`
- Extract `build_forwarding_headers()` as a reusable pure function and refactor `fill_forwarding_hdrs_to_api_session()` to use it internally

## Test plan
- [x] Verify webserver starts without errors
- [ ] Verify `app["client_registry"]` is accessible and properly initialized
- [x] Verify login/logout/update-password flows still work

Resolves BA-4838

### DIFF
--- a/changes/9595.feature.md
+++ b/changes/9595.feature.md
@@ -1,1 +1,1 @@
-Register a persistent `BackendAIClientRegistry` on the webserver app (`app["client_registry"]`) using the new `NoAuth` strategy, and extract `build_forwarding_headers()` as a reusable pure function for building proxy forwarding headers.
+Register a persistent `BackendAIClientRegistry` on the webserver and use it for the `update-password-no-auth` API handler
--- a/src/ai/backend/client/v2/auth.py
+++ b/src/ai/backend/client/v2/auth.py
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 from collections.abc import Mapping
 from datetime import datetime
+from typing import override
 
 from yarl import URL
 
@@ -29,6 +30,7 @@ class NoAuth(AuthStrategy):
     manager rather than crash internally.
     """
 
+    @override
     def sign(
         self,
         method: str,
@@ -52,6 +54,7 @@ class HMACAuth(AuthStrategy):
         self.secret_key: str = secret_key
         self.hash_type: str = hash_type
 
+    @override
     def sign(
         self,
         method: str,


src/ai/backend/web/server.py
```diff
@@ -764,6 +767,28 @@ async def _shutdown(_app: web.Application) -> None:
     yield client_pool
 
 
+@asynccontextmanager
+async def anon_client_ctx(
+    config: WebServerUnifiedConfig,
+) -> AsyncGenerator[BackendAIAnonymousClient]:
+    v2_config = V2ClientConfig(
+        endpoint=URL(str(config.api.endpoint[0])),
```
jopemachine, line 768: If there are multiple endpoints, are the other endpoints simply ignored?

fregataa: yes. Do we have a plan to support multiple endpoints in the future? @HyeockJinKim

jopemachine: If multiple values are provided for endpoint, it would be better to emit a warning rather than silently ignoring them.

src/ai/backend/web/server.py
```diff
@@ -764,6 +760,21 @@ async def _shutdown(_app: web.Application) -> None:
...
+) -> AsyncGenerator[BackendAIClientRegistry]:
+    client_config = V2ClientConfig(
+        endpoint=URL(str(config.api.endpoint[0])),
+        skip_ssl_verification=not config.api.ssl_verify,
+    )
+    registry = await BackendAIClientRegistry.create(client_config, NoAuth())
+    try:
+        yield registry
+    finally:
+        await registry.close()
```
jopemachine, line 775: Instead of client_registry, wouldn’t it be better to use a name that clearly indicates that no authentication is involved?

changes/9595.feature.md
```diff
@@ -0,0 +1 @@
+Register a persistent `BackendAIClientRegistry` on the webserver app (`app["client_registry"]`) using the new `NoAuth` strategy, and extract `build_forwarding_headers()` as a reusable pure function for building proxy forwarding headers.
```
jopemachine: It seems news fragment also should be updated

src/ai/backend/client/v2/auth.py
```diff
@@ -21,6 +21,26 @@ def sign(
...
 
+class NoAuth(AuthStrategy):
+    """Auth strategy that provides no credentials.
+
+    Used by the webserver, which has no real keypair and only proxies requests.
+    Any accidental call to an authenticated endpoint will get a 401 from the
+    manager rather than crash internally.
+    """
+
+    def sign(
```
jopemachine, line 34:
```suggestion
    @override
    def sign(
```

ironAiken2 (dismissed):
LGTM

PR #9624, opened 2026-03-04
feat(BA-4865): Apply RBAC validator infrastructure to Session actions

fregataa:
## Summary
- Refactored Session actions to use `BaseScopeAction` and `BaseSingleEntityAction` base classes for RBAC validation
- Added `SessionScopeAction` and `SessionSingleEntityAction` with `target_element()` methods for BEP-1048 compliance
- Updated 6 scope-level actions (Create, Search, Match) and 4 single-entity actions (Get, Destroy, Execute, Modify) to implement RBAC validation infrastructure
- Followed VFolder patterns from BA-4620 for consistent RBAC validation across all guarded entities

## Test plan
- [x] Lint and type checks pass
- [ ] Unit tests pass (CI will verify)
- [ ] Integration tests with RBAC validators (follow-up work needed)
- [ ] API layer updates to set scope/session_id fields (tracked in BA-2946)

## Next Steps
This PR provides the **infrastructure foundation** for Session RBAC validation. Follow-up work required:
1. Update API/processor layer to populate `_scope_type`, `_scope_id`, and `session_id` fields
2. Connect RBAC validators to action processors
3. Apply same pattern to remaining 20+ single-entity Session actions
4. Add integration tests for RBAC permission checks

Resolves BA-4865

### DIFF
--- a/tests/unit/manager/api/session/test_dataloader.py
+++ b/tests/unit/manager/api/session/test_dataloader.py
@@ -5,9 +5,22 @@ from __future__ import annotations
 import uuid
 from unittest.mock import AsyncMock, MagicMock
 
+from ai.backend.common.contexts.user import with_user
+from ai.backend.common.data.user.types import UserData, UserRole
 from ai.backend.common.types import KernelId
 from ai.backend.manager.api.gql.data_loader.kernel.loader import load_kernels_by_ids
 from ai.backend.manager.data.kernel.types import KernelInfo
+
+
+def _make_user_data() -> UserData:
+    return UserData(
+        user_id=uuid.uuid4(),
+        is_authorized=True,
+        is_admin=False,
+        is_superadmin=False,
+        role=UserRole.USER,
+        domain_name="default",
+    )
 
 
 class TestLoadKernelsByIds:
@@ -47,7 +60,8 @@ class TestLoadKernelsByIds:
         )
 
         # When
-        result = await load_kernels_by_ids(mock_processor, [id1, id2, id3])
+        with with_user(_make_user_data()):
+            result = await load_kernels_by_ids(mock_processor, [id1, id2, id3])
 
         # Then
         assert result == [kernel1, kernel2, kernel3]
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/services/session/processors.py
```diff
@@ -172,42 +173,88 @@ def __init__(
...
-        self.modify_session = ActionProcessor(service.modify_session, action_monitors)
         self.check_and_transit_status = ActionProcessor(
             service.check_and_transit_status, action_monitors
         )
 
+        # Scope actions with RBAC validation
+        self.create_cluster = ActionProcessor(
+            service.create_cluster,
+            action_monitors,
+            validators=[cast(ActionValidator, scope_validator)],
```
jopemachine, line 210: It seems that casting is required because there is no inheritance relationship between the validators.
Is there a reason why they are not designed with an inheritance structure? @fregataa

src/ai/backend/manager/services/session/base.py
```diff
@@ -19,3 +20,24 @@ class SessionBatchAction(BaseBatchAction):
...
+
+@dataclass
+class SessionScopeAction(BaseScopeAction):
+    """Base class for session actions that operate within a scope (domain/project).
+
+    Used for operations like creating or searching sessions within a specific scope.
+    Subclasses must implement scope_type(), scope_id(), and target_element() methods.
+
+    Note: Scope should typically be USER scope (user_id), not GLOBAL.
+    """
```
jopemachine, line 31: The comment is confusing.

The line above mentions domain and project in the context of scope, while the line below mainly refers to user scope.

src/ai/backend/manager/services/session/processors.py
fregataa: The cast is needed because ScopeActionValidator and SingleEntityActionValidator don't inherit from ActionValidator — they are parallel hierarchies.
Since this is a shared infrastructure concern (actions/validator), not specific to Session RBAC, it should be addressed in a separate issue rather than in this branch.

PR #9630, opened 2026-03-04
feat(BA-4866): extract RBACValidators injection to dependency layer

fregataa:
## Summary
- this work **DOES NOT** apply any validator to action processors.
- Create `Validators` and `RBACValidators` dataclasses for structured validator injection
- Wire validators through the DI layer (`server.py`, `ProcessingComposer`, `ProcessorsDependency`) into `VFolderProcessors`, following the same pattern as `action_monitors`
- Validators are injected but not yet applied to individual action processors (follow-up task)

## Test plan
- [x] `pants fmt/fix/lint` passed
- [x] `pants check` (mypy) passed
- [x] CI passes

Resolves BA-4866

### DIFF
--- a/tests/component/artifact_registry/conftest.py
+++ b/tests/component/artifact_registry/conftest.py
@@ -4,6 +4,7 @@ import uuid
 from collections.abc import AsyncIterator, Callable, Coroutine
 from dataclasses import dataclass
 from typing import Any
+from unittest.mock import MagicMock
 
 import pytest
 import sqlalchemy as sa
@@ -11,6 +12,7 @@ from sqlalchemy.ext.asyncio.engine import AsyncEngine as SAEngine
 
 from ai.backend.common.bgtask.bgtask import BackgroundTaskManager
 from ai.backend.common.data.artifact.types import ArtifactRegistryType
+from ai.backend.manager.actions.validators import ActionValidators
 from ai.backend.manager.api.rest.artifact_registry.handler import ArtifactRegistryHandler
 from ai.backend.manager.api.rest.artifact_registry.registry import (
     register_artifact_registry_routes,
@@ -82,7 +84,9 @@ def artifact_processors(
         storage_manager=storage_manager,
         config_provider=config_provider,
     )
-    return ArtifactProcessors(service=service, action_monitors=[])
+    return ArtifactProcessors(
+        service=service, action_monitors=[], validators=MagicMock(spec=ActionValidators)
+    )
 
 
 @pytest.fixture()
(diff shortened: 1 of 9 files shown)


src/ai/backend/manager/actions/validators/rbac/__init__.py
```diff
@@ -0,0 +1,12 @@
+from dataclasses import dataclass
+
+from ai.backend.manager.actions.validators.rbac.scope import ScopeActionRBACValidator
+from ai.backend.manager.actions.validators.rbac.single_entity import (
+    SingleEntityActionRBACValidator,
+)
+
+
+@dataclass
+class RBACValidators:
```
seedspirit, line 10: I'm not sure define dataclass in `__init__.py`. I think init.py should only be used for importing things to make importing more convenient. What do others think? @jopemachine @HyeockJinKim

src/ai/backend/manager/dependencies/processing/composer.py
```diff
@@ -231,11 +237,18 @@ async def compose(
...
+            scope=ScopeActionRBACValidator(permission_controller_repository),
+            single_entity=SingleEntityActionRBACValidator(permission_controller_repository),
+        )
+
         processors = await stack.enter_dependency(
             ProcessorsDependency(),
             ProcessorsProviderInput(
                 service_args=service_args,
                 action_monitors=[reporter_monitor, prometheus_monitor, audit_log_monitor],
+                validators=ActionValidators(rbac=rbac_validators),
```
seedspirit, line 262: While this may not be within the scope of this PR's changes, shouldn't validators also be implemented as `list[ActionValidator]` as `action_monitors`?

src/ai/backend/manager/actions/validators/rbac/__init__.py
jopemachine: I think it would be better to keep things like types.py in a separate file.

src/ai/backend/manager/actions/validators/__init__.py
```diff
@@ -0,0 +1,8 @@
+from dataclasses import dataclass
+
+from .rbac import RBACValidators
+
+
+@dataclass
+class ActionValidators:
```
jopemachine, line 7: Just a question,
Are there any plans to include implementations in ActionValidators other than RBAC?

jopemachine:
There are a large number of conflicts

src/ai/backend/manager/actions/validators/rbac/__init__.py
fregataa: types.py is not a good idea because usually types modules should be the one that other modules depend on, not the one that depends on other modules

src/ai/backend/manager/dependencies/processing/composer.py
fregataa: This one does not need to be a list.
Processors get the signature as a list because they iterate validators and check one by one. But this one does not need to be iterated.

src/ai/backend/manager/actions/validators/__init__.py
fregataa: No plan for other validators yet

jopemachine (approved):

PR #9634, opened 2026-03-04
fix(BA-4868): add status field to UpdateRoleInput GraphQL input

fregataa:
## Summary
- Add `status: RoleStatusGQL | None` field to `UpdateRoleInput` GraphQL input type
- Wire the status field through `to_updater()` to `RoleUpdaterSpec` so role status can be updated via the mutation
- `RoleUpdaterSpec` already supported status updates; this change exposes it at the GraphQL layer

## Test plan
- [x] Verify `updateRole` mutation accepts `status` parameter
- [x] Verify role status can be changed to ACTIVE/INACTIVE/DELETED via GraphQL

Resolves BA-4868

### DIFF
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -12091,6 +12091,7 @@ input UpdateRoleInput
   id: UUID!
   name: String = null
   description: String = null
+  status: RoleStatus = null
 }
 
 """Added in 25.19.0. Input for updating route traffic status."""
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/api/gql/rbac/types/role.py
```diff
@@ -357,6 +357,7 @@ class UpdateRoleInput:
     id: uuid.UUID
     name: str | None = None
     description: str | None = None
+    status: RoleStatusGQL | None = None
 
     def to_updater(self) -> Updater[RoleRow]:
         spec = RoleUpdaterSpec(
```
seedspirit, line 363: It doesn't seem to be within the current scope of modifications, but there's a field called `source` in `RoleUpdaterSpec`. Is this not being used?

seedspirit (approved):

ironAiken2 (approved):
LGTM

PR #9636, opened 2026-03-04
feat(BA-2938): Migrate Session data to RBAC database

fregataa:
## Summary

- Add SESSION entity-type permissions to all role+scope combinations (skip domain member roles)
- Create AUTO edges from User scope → Session (via user_uuid) and Project scope → Session (via group_id)
- Follow new RBAC pattern with direct role+scope in permissions table (no permission_groups)
- Use keyset pagination for scalability with large session tables

This migration brings Session entities into the RBAC system, enabling fine-grained access control. Unlike VFolder, Session has no invitation/sharing mechanism, making the migration simpler with only entity-type permissions and AUTO edges.

## Test plan

- [x] Migration applies successfully (`alembic upgrade head`)
- [x] Migration rolls back correctly (`alembic downgrade -1`)
- [x] All quality checks pass (fmt, lint, check)
- [x] CI tests pass

Resolves BA-2938

### DIFF
--- a/src/ai/backend/manager/models/alembic/versions/30c8308738ee_migrate_session_data_to_rbac.py
+++ b/src/ai/backend/manager/models/alembic/versions/30c8308738ee_migrate_session_data_to_rbac.py
@@ -26,6 +26,8 @@ depends_on = None
 # Constants
 BATCH_SIZE = 1000
 MEMBER_ROLE_SUFFIX = "member"
+SESSION_ENTITY_TYPE = EntityType.SESSION.value
+AUTO_RELATION_TYPE = "auto"
 
 
 def _add_entity_type_permissions(db_conn: Connection) -> None:
@@ -93,7 +95,7 @@ def _add_entity_type_permissions(db_conn: Connection) -> None:
             "member_ops": member_ops,
             "owner_ops": owner_ops,
             "member_pattern": f"%{MEMBER_ROLE_SUFFIX}",
-            "entity_type": EntityType.SESSION.value,
+            "entity_type": SESSION_ENTITY_TYPE,
         },
     )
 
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/models/alembic/versions/30c8308738ee_migrate_session_data_to_rbac.py
```diff
@@ -0,0 +1,211 @@
...
+    # Process User scope edges
+    last_id = "00000000-0000-0000-0000-000000000000"
+    while True:
+        query = sa.text("""
+            SELECT id::text AS id, user_uuid::text AS user_uuid
+            FROM sessions
+            WHERE id::text > :last_id
+            ORDER BY id
+            LIMIT :limit
+        """)
```
fregataa: Reasons why I don't use OFFSET here:
- https://www.postgresql.org/docs/current/queries-limit.html#QUERIES-LIMIT
- https://docs.gitlab.com/development/database/keyset_pagination/

Using OFFSET to migrate huge data is inefficient.

src/ai/backend/manager/models/alembic/versions/30c8308738ee_migrate_session_data_to_rbac.py
```diff
@@ -0,0 +1,204 @@
...
+    db_conn: Connection,
+    scope_type: str,
+    scope_id_column: str,
+) -> None:
+    """Associate sessions to a given scope type using keyset pagination.
+
+    Creates AUTO edges from the specified scope to each session.
+    """
+    entity_type = EntityType.SESSION.value
+    relation_type = "auto"
```
jopemachine: Wouldn't it be better to define relation_type as a constant?

jopemachine:
Is it appropriate to classify a data migration PR as a feat?

fregataa:
> Is it appropriate to classify a data migration PR as a feat?

yes it is because it effects to RBAC features

PR #9650, opened 2026-03-05
feat(BA-4872): Add server-side API endpoint for searching deployment policies

jopemachine:
## Summary

- Add `POST /deployments/policies/search` REST API endpoint in the manager
- Wire the handler to the existing `search_deployment_policies` service/repository methods
- Add `build_querier` method to `DeploymentPolicyAdapter` for pagination support

Resolves BA-4872

## Test plan

- [x] Verified `backend.ai deployment policy list` returns all policies with correct data
- [x] Verified `--limit` and `--offset` pagination options work

### DIFF
--- a/src/ai/backend/manager/api/rest/deployment/adapter.py
+++ b/src/ai/backend/manager/api/rest/deployment/adapter.py
@@ -38,6 +38,7 @@ from ai.backend.common.dto.manager.deployment import (
     RouteDTO,
     RouteFilter,
     RouteOrder,
+    SearchDeploymentPoliciesRequest,
     SearchDeploymentsRequest,
     SearchRevisionsRequest,
     SearchRoutesRequest,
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/api/rest/deployment/adapter.py
```diff
@@ -544,6 +545,11 @@ def convert_to_dto(self, data: DeploymentPolicyData) -> DeploymentPolicyDTO:
             updated_at=data.updated_at,
         )
 
+    def build_querier(self, request: SearchDeploymentPoliciesRequest) -> BatchQuerier:
+        """Build a BatchQuerier for deployment policies from search request."""
+        pagination = OffsetPagination(limit=request.limit, offset=request.offset)
```
seedspirit, line 550: Does it currently only support offset based pagination?

jopemachine:
Since users are likely to consider the deployment policy as an attribute of a deployment, I will include it in the deployment search API instead of creating a separate deployment policy search API.

PR #9653, opened 2026-03-05
fix(BA-4624): Validate cloud detection responses and harden metadata JSON parsing

seedspirit:
resolves BA-4624

- [x] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/tests/unit/common/test_identity.py
+++ b/tests/unit/common/test_identity.py
@@ -273,7 +273,7 @@ class TestDetectCloudServices:
         body: str,
     ) -> None:
         mock_responses.get(aws_metadata_url, body=body)
-        with pytest.raises(CloudDetectionError, match="Not a AWS metadata response"):
+        with pytest.raises(CloudDetectionError, match="AWS detection failed"):
             await _detect_aws(client_session)
 
     @pytest.mark.parametrize("status", [404, 500, 503], ids=["404", "500", "503"])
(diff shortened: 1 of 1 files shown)


src/ai/backend/common/identity.py
```diff
@@ -67,15 +71,28 @@ async def _detect_azure(session: aiohttp.ClientSession) -> CloudProvider:
         "http://169.254.169.254/metadata/instance/compute",
         params={"api-version": "2021-02-01"},
         headers={"Metadata": "true"},
-    ):
+    ) as resp:
+        try:
+            data = await resp.json()
+        except json.JSONDecodeError:
+            raise ValueError("Not an Azure metadata response") from None
```
jopemachine: Let's use an error based on BackendAIError.

src/ai/backend/common/identity.py
```diff
@@ -67,15 +71,28 @@ async def _detect_azure(session: aiohttp.ClientSession) -> CloudProvider:
...
+        # ref: https://learn.microsoft.com/azure/virtual-machines/instance-metadata-service?tabs=linux
+        if not isinstance(data, dict) or "vmId" not in data:
```
jopemachine, line 95: How about defining the schema as a type?

seedspirit: Given the large volume of data provided and the potential for change(usually added), it seems best to proceed with the current implementation.

tests/unit/common/test_identity.py
```diff
@@ -211,3 +223,362 @@ async def test_get_instance_type(provider: str | None) -> None:
         elif provider is None:
             ret = await ai.backend.common.identity.get_instance_type()
             assert ret == "default"
+
+
+_AWS_URL = "http://169.254.169.254/latest/meta-data/"
```
jopemachine, line 228: This change is not part of this PR, but it would be better to extract 169.254.169.254 into a variable instead of repeating it.

HyeockJinKim:
Please check the CI test. @seedspirit

PR #9720, opened 2026-03-05
feat(BA-4905): add LoginSecurityPolicy model, login_sessions table, and repository layer

HyeockJinKim:
## Summary
- Add `LoginSecurityPolicy` Pydantic model with `max_concurrent_logins: int | None` field and `login_security_policy` JSONB column on `UserRow`
- Add `LoginSessionRow` ORM model and Alembic migration for the `login_sessions` table with proper FK, indexes, and `LoginSessionExpiryReason` enum
- Implement `LoginSessionRepository` (cache-first via Valkey Sorted Set + DB fallback) and `LoginSessionService` (create/expire/evict/check concurrency) following existing repository and service patterns

## Test plan
- [ ] Alembic migration applies cleanly: `alembic upgrade head`
- [ ] `LoginSessionRepository` ZADD/ZSCORE/ZCARD/ZPOPMIN/ZREM operations work against Valkey
- [ ] `LoginSessionService.check_concurrency_limit` enforces `max_concurrent_logins` correctly

Resolves BA-4905

### DIFF
--- a/changes/9720.feature.md
+++ b/changes/9720.feature.md
@@ -1,0 +1,1 @@
+Add LoginSecurityPolicy model, login_sessions table, LoginSessionRepository (Valkey Sorted Set + DB), and LoginSessionService for concurrent login session management.


src/ai/backend/manager/data/user/types.py
```diff
@@ -87,6 +87,7 @@ class UserData:
     container_uid: int | None = field(compare=False)
     container_main_gid: int | None = field(compare=False)
     container_gids: list[int] | None = field(compare=False)
+    login_security_policy: dict[str, Any] | None = field(default=None, compare=False)
```
HyeockJinKim, line 90: Avoid using `dict`. We recommend handling data through `PydanticModel` or `dataclass`.

src/ai/backend/manager/models/user/row.py
```diff
@@ -237,6 +239,9 @@ class UserRow(Base):  # type: ignore[misc]
     container_gids: Mapped[list[int] | None] = mapped_column(
         "container_gids", sa.ARRAY(sa.Integer), nullable=True, server_default=sa.null()
     )
+    login_security_policy: Mapped[LoginSecurityPolicy | None] = mapped_column(
+        "login_security_policy", PydanticColumn(LoginSecurityPolicy), nullable=True
+    )
```
HyeockJinKim, line 244: Use the default value rather than leaving it optional.

PR #9741, opened 2026-03-06
feat(BA-4768): Support cloning vfolders to a different quota scope

seedspirit:
## Summary
- Add `target_quota_scope_id` parameter to the vfolder clone API, enabling clients to clone vfolders into a different quota scope (e.g. user personal quota instead of the source's scope)
- When `target_quota_scope_id` is not specified, it defaults to the requester's own user scope (`QuotaScopeID(USER, requester_uuid)`), maintaining backward compatibility
- Validate requester's access to the target quota scope before proceeding with the clone operation

## Test plan
- [x] Unit test: clone with `target_quota_scope_id=None` defaults to user scope (backward compat)
- [x] Unit test: clone with explicit `target_quota_scope_id` validates access and succeeds
- [x] Unit test: clone with unauthorized `target_quota_scope_id` raises `InvalidAPIParameters`
- [x] Existing clone tests updated and passing (4 tests)
- [x] `pants fmt fix lint check` all pass

Resolves BA-4768

### DIFF
--- a/tests/unit/storage-proxy/volumes/test_vfs.py
+++ b/tests/unit/storage-proxy/volumes/test_vfs.py
@@ -5,11 +5,12 @@ import uuid
 from collections.abc import AsyncIterator, Iterator
 from pathlib import Path, PurePosixPath
 from typing import TYPE_CHECKING
-from unittest.mock import MagicMock
+from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 
 from ai.backend.common.types import QuotaScopeID, QuotaScopeType
+from ai.backend.storage.errors import QuotaScopeCreationFailedError
 from ai.backend.storage.types import VFolderID
 from ai.backend.storage.volumes.vfs import BaseVolume
 
@@ -77,8 +78,14 @@ class TestBaseVolume:
         dst_qsid = QuotaScopeID(QuotaScopeType.USER, uuid.uuid4())
         dst_vfid = VFolderID(dst_qsid, uuid.uuid4())
         yield dst_vfid
-        await base_volume.delete_vfolder(dst_vfid)
-        await base_volume.quota_model.delete_quota_scope(dst_qsid)
+        try:
+            await base_volume.delete_vfolder(dst_vfid)
+        except FileNotFoundError:
+            pass
+        try:
+            await base_volume.quota_model.delete_quota_scope(dst_qsid)
+        except FileNotFoundError:
+            pass
 
     async def test_add_file_writes_content_correctly(
         self,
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/services/vfolder/services/vfolder.py
```diff
@@ -708,7 +709,23 @@ async def clone(self, action: CloneVFolderAction) -> CloneVFolderActionResult:
 
         # Create source and target VFolderID
         source_folder_id = VFolderID(source_vfolder_data.quota_scope_id, source_vfolder_data.id)
-        target_quota_scope_id = "..."  # TODO: implement
+
+        # Resolve target quota scope: use the provided one or default to requester's user scope
+        if action.target_quota_scope_id is not None:
+            target_quota_scope_id = action.target_quota_scope_id
+            if target_quota_scope_id.scope_type != QuotaScopeType.USER:
+                raise VFolderInvalidParameter("Clone target must be a user quota scope.")
```
seedspirit, line 738: I wrote the code assuming that this is only possible when the target `QuotaScopeType` is "User." Is that correct? @fregataa

fregataa: Project quota scope is available and Project vfolders are able to be cloned.
But we do not support cloning a folder to be a Project folder, as cloned results become User vfolders.
 I think it is good to leave a TODO comment

src/ai/backend/storage/volumes/vfs/__init__.py
```diff
@@ -455,9 +455,15 @@ async def clone_vfolder(
         if vfolder_usage.used_bytes > fs_usage.capacity_bytes - fs_usage.used_bytes:
             raise ProcessExecutionError("Not enough space available for clone.")
 
-        # create the target vfolder
+        # create the target vfolder (auto-create quota scope if missing)
```
jopemachine, line 466: Rather than catching an exception to create it, wouldn’t it be better to first check for existence and create it only if it doesn’t exist?

src/ai/backend/storage/volumes/vfs/__init__.py
```diff
@@ -455,9 +455,15 @@ async def clone_vfolder(
...
         src_vfpath = self.mangle_vfpath(src_vfid)
-        await self.create_vfolder(dst_vfid)
+        try:
+            await self.create_vfolder(dst_vfid)
+        except QuotaScopeNotFoundError:
+            if dst_vfid.quota_scope_id is None:
+                raise
+            await self.quota_model.create_quota_scope(dst_vfid.quota_scope_id)
+            await self.create_vfolder(dst_vfid)
```
fregataa: Please write a unit test for regression

src/ai/backend/storage/volumes/vfs/__init__.py
```diff
@@ -455,8 +455,12 @@ async def clone_vfolder(
         if vfolder_usage.used_bytes > fs_usage.capacity_bytes - fs_usage.used_bytes:
             raise ProcessExecutionError("Not enough space available for clone.")
 
-        # create the target vfolder
+        # create the target vfolder (auto-create quota scope if missing)
         src_vfpath = self.mangle_vfpath(src_vfid)
+        if not self.quota_model.mangle_qspath(dst_vfid).exists():
+            if dst_vfid.quota_scope_id is None:
+                raise QuotaScopeNotFoundError
+            await self.quota_model.create_quota_scope(dst_vfid.quota_scope_id)
```
fregataa: we have to consider the case when a target quota scope is not created

PR #9742, opened 2026-03-06
test(BA-4921): expand check_permission_with_scope_chain() test coverage

fregataa:
## Summary
- Add 15 new test cases for `check_permission_with_scope_chain()` covering edge cases: operation mismatch, deleted/unassigned roles, multiple roles, mixed AUTO/REF edge chains, deep 3-level AUTO chains, self-scope combinations, user isolation, multiple permissions on same role, and entity type mismatch
- Extend `permission_setup` fixture to support `user_scope` key and optional `entity_type` parameter
- Add new fixtures: `user_with_deleted_role`, `user_with_unassigned_role`, `user_with_two_roles`, `multi_role_permission_setup`, `project_in_domain_ref`, `domain_in_user_scope_auto`, `other_user_with_permission`

## Test plan
- [x] `pants lint` passes
- [x] `pants check` passes (mypy)
- [x] `pants test tests/unit/manager/repositories/permission_controller/test_check_permission_with_scope_chain.py` — all 24 test cases pass

Resolves BA-4921

### DIFF
--- a/tests/unit/manager/repositories/permission_controller/test_check_permission_with_scope_chain.py
+++ b/tests/unit/manager/repositories/permission_controller/test_check_permission_with_scope_chain.py
@@ -221,12 +221,12 @@ class TestCheckPermissionWithScopeChain:
         expected: bool,
     ) -> None:
         """VFOLDER in PROJECT (auto): direct scope permission checks."""
-        f = user_with_active_role
+        fixture = user_with_active_role
         result = await db_source.check_permission_with_scope_chain(
-            user_id=f.user_id,
+            user_id=fixture.user_id,
             target_element_ref=RBACElementRef(
                 element_type=RBACElementType.VFOLDER,
-                element_id=f.vfolder_id,
+                element_id=fixture.vfolder_id,
             ),
             operation=check_op,
         )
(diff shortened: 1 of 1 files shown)


tests/unit/manager/repositories/permission_controller/test_check_permission_with_scope_chain.py
```diff
@@ -36,6 +36,15 @@
 from ai.backend.testutils.db import with_tables
 
 
+@dataclass
+class PermissionEntry:
```
jopemachine, line 40: How about `VFolderPermissionEntry`?

tests/unit/manager/repositories/permission_controller/test_check_permission_with_scope_chain.py
```diff
@@ -447,3 +460,596 @@ async def test_self_scope_without_any_association(
...
+        db_source: PermissionDBSource,
+        user_with_active_role: ScopeChainFixture,
+        vfolder_in_project_auto: None,
+        project_in_domain_ref: None,
+        permission_setup: None,
+        check_op: OperationType,
+        expected: bool,
+    ) -> None:
+        """VFOLDER→(AUTO)→PROJECT→(REF)→DOMAIN: REF in the middle blocks chain."""
+        f = user_with_active_role
```
jopemachine: I think if possible, it would be better not to use abbreviated variable names.
With a bit of thought, one can figure out that it stands for fixture, but it still requires an extra moment to interpret, which could be avoided.

tests/unit/manager/repositories/permission_controller/test_check_permission_with_scope_chain.py
fregataa: There is test codes with Session type, so VFolder permission is not a proper name

fregataa:
Continue after BA-4952 or BA-4953 is resolved

PR #9760, opened 2026-03-06
fix(BA-4965): sync circuit route_info on route updates for accurate metric collection

lizable:
## Summary
This PR Resolves [BA-4965](https://lablup.atlassian.net/browse/BA-4965) by 
- Fix auto-scaling scale-out failing to re-trigger after the first scale-in cycle
- `update_circuit_route_info()` updated only the proxy backend but not `circuit.route_info`, causing metric collection to keep scraping terminated routes and failing entirely
- Add per-route exception handling so that one unreachable route does not abort collection for the remaining healthy routes

## Root Cause

1. Scale-in terminates the oldest route first (equal termination priority, stable sort)
2. `update_circuit_route_info()` updates `self.backends[key]` but not `self.circuits[key].route_info`
3. Metric collection (`gather_prometheus_inference_measures`) uses stale `circuit.route_info` and attempts to reach the terminated route's `/metrics` endpoint
4. Without per-route exception handling, the entire circuit collection fails and `_app_inference_metrics` is not updated
5. Stale metrics (value=0) persist in Valkey, so the scale-out condition (`> threshold`) is never met
6. Scale-in continues to work because the stale value 0 satisfies `<= 0`

## Detailed Analysis: Why stale metrics cause asymmetric behavior

Even though metric collection fails after scale-in, `collect_inference_metric()` still pushes the **last successfully collected values** to Valkey on every cycle:

```python
# After gather fails (exception caught), _app_inference_metrics is NOT updated
# but the old values are still pushed to Valkey for ALL circuits:
app_metrics_updates = {
    circuit.endpoint_id: {
        key: obj.to_serializable_dict()
        for key, obj in circuit._app_inference_metrics.items()
    }
    for circuit in inference_circuits  # includes failed circuits
}
await root_ctx.valkey_stat.store_inference_metrics(app_metrics_updates, ...)
```

Timeline:

1. During load: collection succeeds → `vllm_num_requests_running = 10`
2. Load stops: collection succeeds → `vllm_num_requests_running = 0`
3. Scale-in triggers → terminates the original route (oldest, first in stable sort)
4. All subsequent collections fail (terminated route unreachable)
5. `_app_inference_metrics` frozen at last successful value = **0**
6. This stale value **0** is pushed to Valkey every cycle, keeping the TTL alive

Result:

- **Scale-out**: `0 > 5` → False → never triggers again
- **Scale-in**: `0 <= 0` → True → still triggers (confirmed: changing `min_replicas` to 0 allowed scale-in from 1→0)

The metrics are not truly missing — they are **zombie values** from the last successful collection, which happen to be 0 (idle state). This creates an asymmetry where scale-in conditions are perpetually satisfied while scale-out conditions can never be met.

## Why stale per-replica metrics cleanup is unnecessary

After a route is removed, its per-replica metrics remain in `circuit._replica_inference_metrics` and are pushed to Valkey. However, this does **not** require explicit cleanup because:

- **Auto-scaling evaluates only app-level metrics** (`inference.{endpoint_id}.app`), not per-replica metrics (`inference.{endpoint_id}.replica.{route_id}`) — stale replica entries have zero impact on scaling decisions
- **Valkey TTL (120s) naturally expires** stale replica entries without any intervention
- **Once `circuit.route_info` is synced** (the root cause fix), new collection cycles no longer add data for removed routes — existing stale entries simply stop being refreshed and expire

Removing the per-cycle cleanup loop avoids unnecessary iteration over all metric keys and route IDs every collection cycle.

## Changes

- `base.py`: Sync `circuit.route_info` in `update_circuit_route_info()`, updating it **after** `update_backend()` succeeds to ensure consistency — if backend update fails, `route_info` remains unchanged
- `metrics.py`: Add per-route try/except in `gather_prometheus_inference_measures()` to skip unreachable routes gracefully, with `exc_info=True` for detailed error logging (timeout vs connection error vs HTTP status)
- `metrics.py`: Wrap metrics parsing/processing in a separate per-route try/except so that one malformed metrics payload only skips that route without aborting the entire collection
- `test_frontend_route_sync.py`: Regression tests for route_info sync, including backend update failure path (route_info must remain unchanged on failure)
- `test_metrics_resilience.py`: Regression tests for per-route error resilience, including malformed metrics payload isolation

## Test plan

- [ ] Trigger scale-out, then scale-in completes, apply load again — verify scale-out re-triggers
- [ ] Confirm metric collection continues for remaining healthy routes after a route is terminated during scale-in
- [ ] Check appproxy logs for warning messages when a terminated route is skipped

[BA-4965]: https://lablup.atlassian.net/browse/BA-4965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/9781.enhance.md
+++ b/changes/9781.enhance.md
@@ -1,0 +1,1 @@
+Add per-plugin timeout (120s) to `gather_container_measures` calls so a single hung plugin does not block stat collection from all other plugins
--- a/changes/9782.fix.md
+++ b/changes/9782.fix.md
@@ -1,0 +1,1 @@
+Pre-validate namespace path before `netstat_ns()` to prevent thread pool exhaustion from hung threads on stale network namespaces
--- a/changes/9869.enhance.md
+++ b/changes/9869.enhance.md
@@ -1,0 +1,1 @@
+Extract shared PaginationInfo DTO class and replace per-module duplicates with unified import.
--- a/src/ai/backend/agent/docker/intrinsic.py
+++ b/src/ai/backend/agent/docker/intrinsic.py
@@ -715,20 +715,36 @@ class MemoryPlugin(AbstractComputePlugin):
             sandbox_key = data["NetworkSettings"]["SandboxKey"]
             net_rx_bytes = 0
             net_tx_bytes = 0
-            try:
-                nstat = await netstat_ns(sandbox_key)
-            except OSError as e:
+            if not sandbox_key:
                 log.warning(
-                    "MemoryPlugin: cannot read net stats for container {0}: {1!r}",
+                    "MemoryPlugin: empty SandboxKey for container {0},"
+                    " skipping net stat collection",
                     container_id[:7],
-                    e,
                 )
-                return None
-            for name, net_stat in nstat.items():
-                if name == "lo":
-                    continue
-                net_rx_bytes += net_stat.bytes_recv
-                net_tx_bytes += net_stat.bytes_sent
+            else:
+                ns_path = Path(sandbox_key)
+                if not ns_path.exists():
+                    log.warning(
+                        "MemoryPlugin: network namespace path does not exist for container"
+                        " {0} (sandbox_key={1!r}), skipping net stat collection",
+                        container_id[:7],
+                        sandbox_key,
+                    )
+                else:
+                    try:
+                        nstat = await netstat_ns(ns_path)
+                    except OSError as e:
+                        log.warning(
+                            "MemoryPlugin: cannot read net stats for container {0}: {1!r}",
+                            container_id[:7],
+                            e,
+                        )
+                        return None
+                    for name, net_stat in nstat.items():
+                        if name == "lo":
+                            continue
+                        net_rx_bytes += net_stat.bytes_recv
+                        net_tx_bytes += net_stat.bytes_sent
             loop = current_loop()
             scratch_sz = await loop.run_in_executor(None, get_scratch_size, container_id)
             return (
--- a/src/ai/backend/agent/stats.py
+++ b/src/ai/backend/agent/stats.py
@@ -26,6 +26,7 @@ import attrs
 from ai.backend.common import msgpack
 from ai.backend.common.identity import is_containerized
 from ai.backend.common.metrics.metric import StageObserver
+from ai.backend.common.metrics.types import UTILIZATION_METRIC_INTERVAL
 from ai.backend.common.types import (
     PID,
     ContainerId,
@@ -67,6 +68,8 @@ __all__ = (
 )
 
 log = BraceStyleAdapter(logging.getLogger(__spec__.name))
+
+_PLUGIN_TIMEOUT: float = max(UTILIZATION_METRIC_INTERVAL - 1.0, 1.0)
 
 
 def check_cgroup_available() -> bool:
@@ -735,7 +738,10 @@ class StatContext:
         for computer in self.agent.computers.values():
             _tasks.append(
                 asyncio.create_task(
-                    computer.instance.gather_container_measures(self, container_ids),
+                    asyncio.wait_for(
+                        computer.instance.gather_container_measures(self, container_ids),
+                        timeout=_PLUGIN_TIMEOUT,
+                    ),
                 )
             )
         self._stage_observer.observe_stage(
--- a/src/ai/backend/common/dto/manager/agent/response.py
+++ b/src/ai/backend/common/dto/manager/agent/response.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 from pydantic import BaseModel, Field
 
 from ai.backend.common.api_handlers import BaseResponseModel
+from ai.backend.common.dto.manager.pagination import PaginationInfo
 
 __all__ = (
     "AgentDTO",
@@ -31,14 +32,6 @@ class AgentDTO(BaseModel):
     addr: str = Field(description="Agent address")
     architecture: str = Field(description="CPU architecture")
     version: str = Field(description="Agent version string")
-
-
-class PaginationInfo(BaseModel):
-    """Pagination information."""
-
-    total: int = Field(description="Total number of items")
-    offset: int = Field(description="Number of items skipped")
-    limit: int | None = Field(default=None, description="Maximum items returned")
 
 
 class GetAgentDetailResponse(BaseResponseModel):
--- a/src/ai/backend/common/dto/manager/auto_scaling_rule/response.py
+++ b/src/ai/backend/common/dto/manager/auto_scaling_rule/response.py
@@ -12,6 +12,7 @@ from uuid import UUID
 from pydantic import BaseModel, Field
 
 from ai.backend.common.api_handlers import BaseResponseModel
+from ai.backend.common.dto.manager.pagination import PaginationInfo
 from ai.backend.common.types import AutoScalingMetricSource
 
 __all__ = (
@@ -52,14 +53,6 @@ class GetAutoScalingRuleResponse(BaseResponseModel):
     """Response for getting an auto-scaling rule."""
 
     auto_scaling_rule: AutoScalingRuleDTO = Field(description="Auto-scaling rule data")
-
-
-class PaginationInfo(BaseModel):
-    """Pagination information."""
-
-    total: int = Field(description="Total number of items")
-    offset: int = Field(description="Number of items skipped")
-    limit: int | None = Field(default=None, description="Maximum items returned")
 
 
 class SearchAutoScalingRulesResponse(BaseResponseModel):
--- a/src/ai/backend/common/dto/manager/compute_session/response.py
+++ b/src/ai/backend/common/dto/manager/compute_session/response.py
@@ -4,9 +4,10 @@ from datetime import datetime
 from typing import Any
 from uuid import UUID
 
-from pydantic import BaseModel, Field
+from pydantic import Field
 
 from ai.backend.common.api_handlers import BaseResponseModel
+from ai.backend.common.dto.manager.pagination import PaginationInfo
 
 __all__ = (
     "PaginationInfo",
@@ -15,14 +16,6 @@ __all__ = (
     "GetComputeSessionDetailResponse",
     "SearchComputeSessionsResponse",
 )
-
-
-class PaginationInfo(BaseModel):
-    """Pagination information."""
-
-    total: int = Field(description="Total number of items")
-    offset: int = Field(description="Number of items skipped")
-    limit: int | None = Field(default=None, description="Maximum items returned")
 
 
 class ContainerDTO(BaseResponseModel):
--- a/src/ai/backend/common/dto/manager/deployment/response.py
+++ b/src/ai/backend/common/dto/manager/deployment/response.py
@@ -18,6 +18,7 @@ from ai.backend.common.data.model_deployment.types import (
     RouteStatus,
     RouteTrafficStatus,
 )
+from ai.backend.common.dto.manager.pagination import PaginationInfo
 from ai.backend.common.types import ClusterMode, RuntimeVariant
 
 __all__ = (
@@ -52,14 +53,6 @@ __all__ = (
     "PaginationInfo",
     "CursorPaginationInfo",
 )
-
-
-class PaginationInfo(BaseModel):
-    """Pagination information."""
-
-    total: int = Field(description="Total number of items")
-    offset: int = Field(description="Number of items skipped")
-    limit: int | None = Field(default=None, description="Maximum items returned")
 
 
 class NetworkConfigDTO(BaseModel):
--- a/src/ai/backend/common/dto/manager/domain/response.py
+++ b/src/ai/backend/common/dto/manager/domain/response.py
@@ -11,6 +11,7 @@ from typing import Any
 from pydantic import BaseModel, Field
 
 from ai.backend.common.api_handlers import BaseResponseModel
+from ai.backend.common.dto.manager.pagination import PaginationInfo
 
 __all__ = (
     "CreateDomainResponse",
@@ -38,14 +39,6 @@ class DomainDTO(BaseModel):
     )
     allowed_docker_registries: list[str] = Field(description="Allowed docker registries")
     integration_id: str | None = Field(default=None, description="External integration ID")
-
-
-class PaginationInfo(BaseModel):
-    """Pagination information."""
-
-    total: int = Field(description="Total number of items")
-    offset: int = Field(description="Number of items skipped")
-    limit: int | None = Field(default=None, description="Maximum items returned")
 
 
 class CreateDomainResponse(BaseResponseModel):
--- a/src/ai/backend/common/dto/manager/fair_share/response.py
+++ b/src/ai/backend/common/dto/manager/fair_share/response.py
@@ -9,6 +9,7 @@ from uuid import UUID
 from pydantic import BaseModel, Field
 
 from ai.backend.common.api_handlers import BaseResponseModel
+from ai.backend.common.dto.manager.pagination import PaginationInfo
 
 __all__ = (
     # Common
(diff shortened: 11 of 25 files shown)


src/ai/backend/appproxy/worker/metrics.py
```diff
@@ -386,6 +394,16 @@ async def collect_inference_metric(root_ctx: RootContext, _interval: float) -> N
                                     sum=replica_inference_metrics.sum,
                                 )
 
+        # Clean up stale replica metrics for routes no longer in circuit.route_info
```
jopemachine: I think the patched `update_circuit_route_info` will synchronize the circuits, so why does a cleanup logic need to be included here?

jopemachine:
Since this is a bug fix PR, it seems necessary to add related regression tests

tests/unit/appproxy/worker/test_frontend_route_sync.py
```diff
@@ -185,3 +194,60 @@ async def test_inactive_circuit_update_is_noop(
...
+        3. Only route_a's endpoint is accessed; route_b's is never contacted
+
+        Before the fix, circuit.route_info stayed [route_a, route_b] after
+        scale-in, so the metric collector would try to scrape the terminated
+        route_b's /metrics endpoint, causing collection failure.
+        """
+        port = 10200
+        route_a = _make_route(kernel_host="10.0.0.1", kernel_port=8080)
+        route_b = _make_route(kernel_host="10.0.0.2", kernel_port=8081)
+        circuit = _make_circuit([route_a, route_b], port=port)
```
jopemachine: Please extract these inputs as fixtures

src/ai/backend/appproxy/worker/metrics.py
```diff
@@ -85,42 +85,52 @@ async def gather_prometheus_inference_measures(
...
+        try:
+            async with client_session.get(request_path) as resp:
+                resp.raise_for_status()
+                metrics_text = await resp.text()
+        except Exception:
+            log.warning(
+                "Failed to collect metrics from route {} ({}:{}), skipping",
+                route.route_id,
+                route.current_kernel_host,
+                route.kernel_port,
```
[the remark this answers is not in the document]
jopemachine: I think it make sense. Let’s catch the error and log it so we can identify what kind of error occurred.

tests/unit/appproxy/worker/test_frontend_route_sync.py
```diff
@@ -0,0 +1,264 @@
...
+        await port_frontend.update_circuit_route_info(circuit, [])
+
+        mock_update_backend.assert_not_called()
+
+    async def test_metric_collection_uses_updated_route_info(
+        self,
+        mocker: pytest_mock.MockerFixture,
+        port_frontend: PortFrontend,
+        route_a: RouteInfo,
+        route_b: RouteInfo,
```
jopemachine: Instead of splitting fixtures like route_a and route_b, let's define the problematic scenario itself as a fixture and inject that. (Even if we do need to inject routes as separate fixtures, labels like a and b are hard to read.)

jopemachine:
Is there a related ticket in JIRA? If not, it would be good to create one and link it. @lizable

lizable:
Here's Jira Issue link: https://lablup.atlassian.net/browse/BA-4965

jopemachine:
It looks like the scenarios and test code can be unified using parametrize as shown below. The following is an example snippet.

```python
@dataclass(frozen=True)
class RouteInput:
    """Configuration for a single route in a test scenario."""

    kernel_host: str
    kernel_port: int
    health_status: ModelServiceStatus | None
    response: str | Exception


@dataclass(frozen=True)
class MetricsScenario:
    """Describes a metric collection test case."""

    routes: list[RouteInfo]
    responses: dict[str, str | Exception]
    # None means measures should be empty
    expected_running_value: Decimal | None
    expected_replica_count: int


def _create_metrics_scenario(
    route_configs: list[RouteInput],
    *,
    expected_running_value: Decimal | None,
    expected_replica_count: int,
) -> MetricsScenario:
    """Build a scenario from (host, port, health_status, response) tuples."""
    routes: list[RouteInfo] = []
    responses: dict[str, str | Exception] = {}
    for cfg in route_configs:
        route = _make_route(
            kernel_host=cfg.kernel_host,
            kernel_port=cfg.kernel_port,
            health_status=cfg.health_status,
        )
        routes.append(route)
        responses[_route_endpoint(route)] = cfg.response
    return MetricsScenario(
        routes=routes,
        responses=responses,
        expected_running_value=expected_running_value,
        expected_replica_count=expected_replica_count,
    )


class TestGatherPrometheusInferenceMeasures:
    """Regression tests for per-route error resilience in metric collection."""

    @pytest.mark.parametrize(
        "scenario",
        [
            pytest.param(
                _create_metrics_scenario(
                    [
                        RouteInput(
                            kernel_host="10.0.0.1",
                            kernel_port=8080,
                            health_status=HEALTHY,
                            response=SAMPLE_PROMETHEUS_OUTPUT,
                        ),
                        RouteInput(
                            kernel_host="10.0.0.2",
                            kernel_port=8081,
                            health_status=HEALTHY,
                            response=ConnectionError("Connection refused"),
                        ),
                    ],
                    expected_running_value=Decimal(3),
                    expected_replica_count=1,
                ),
                id="partial-failure-connection-error",
            ),
            pytest.param(
                _create_metrics_scenario(
                    [
                        RouteInput(
                            kernel_host="10.0.0.1",
                            kernel_port=8080,
                            health_status=HEALTHY,
                            response=ConnectionError("Connection refused"),
                        ),
                        RouteInput(
                            kernel_host="10.0.0.2",
                            kernel_port=8081,
                            health_status=HEALTHY,
                            response=TimeoutError("Timed out"),
                        ),
                    ],
                    expected_running_value=None,
                    expected_replica_count=0,
                ),
                id="all-routes-unreachable",
            ),
            pytest.param(
                _create_metrics_scenario(
                    [
                        RouteInput(
                            kernel_host="10.0.0.1",
                            kernel_port=8080,
                            health_status=HEALTHY,
                            response=SAMPLE_PROMETHEUS_OUTPUT,
                        ),
                        RouteInput(
                            kernel_host="10.0.0.2",
                            kernel_port=8081,
                            health_status=HEALTHY,
                            response=SAMPLE_PROMETHEUS_OUTPUT,
                        ),
                    ],
                    expected_running_value=Decimal(6),
                    expected_replica_count=2,
                ),
                id="two-healthy-routes",
            ),
            pytest.param(
                _create_metrics_scenario(
                    [
                        RouteInput(
                            kernel_host="10.0.0.1",
                            kernel_port=8080,
                            health_status=HEALTHY,
                            response=SAMPLE_PROMETHEUS_OUTPUT,
                        ),
                        RouteInput(
                            kernel_host="10.0.0.2",
                            kernel_port=8081,
                            health_status=ModelServiceStatus.UNHEALTHY,
                            response=SAMPLE_PROMETHEUS_OUTPUT,
                        ),
                    ],
                    expected_running_value=Decimal(3),
                    expected_replica_count=1,
                ),
                id="unhealthy-route-skipped",
            ),
            pytest.param(
                _create_metrics_scenario(
                    [
                        RouteInput(
                            kernel_host="10.0.0.1",
                            kernel_port=8080,
                            health_status=HEALTHY,
                            response=SAMPLE_PROMETHEUS_OUTPUT,
                        ),
                        RouteInput(
                            kernel_host="10.0.0.3",
                            kernel_port=8082,
                            health_status=HEALTHY,
                            response=MALFORMED_PROMETHEUS_OUTPUT,
                        ),
                    ],
                    expected_running_value=Decimal(3),
                    expected_replica_count=1,
                ),
                id="malformed-payload",
            ),
        ],
    )
    async def test_metric_collection(self, scenario: MetricsScenario) -> None:
        client_pool = _make_client_pool(scenario.responses)

        measures = await gather_prometheus_inference_measures(client_pool, scenario.routes)

        if scenario.expected_running_value is None:
            assert measures == []
        else:
            running_measures = [
                m for m in measures if m.key == "vllm:num_requests_running"
            ]
            assert len(running_measures) == 1, (
                f"Expected exactly 1 'vllm:num_requests_running' measure, "
                f"got {len(running_measures)}"
            )
            running_measure = running_measures[0]
```

lizable:
> It looks like the scenarios and test code can be unified using parametrize as shown below. The following is an example snippet.
> 
> ```python
> @dataclass(frozen=True)
> class RouteInput:
>     """Configuration for a single route in a test scenario."""
> 
>     kernel_host: str
>     kernel_port: int
>     health_status: ModelServiceStatus | None
>     response: str | Exception
> 
> 
> @dataclass(frozen=True)
> class MetricsScenario:
>     """Describes a metric collection test case."""
> 
>     routes: list[RouteInfo]
>     responses: dict[str, str | Exception]
>     # None means measures should be empty
>     expected_running_value: Decimal | None
>     expected_replica_count: int
> 
> 
> def _create_metrics_scenario(
>     route_configs: list[RouteInput],
>     *,
>     expected_running_value: Decimal | None,
>     expected_replica_count: int,
> ) -> MetricsScenario:
>     """Build a scenario from (host, port, health_status, response) tuples."""
>     routes: list[RouteInfo] = []
>     responses: dict[str, str | Exception] = {}
>     for cfg in route_configs:
>         route = _make_route(
>             kernel_host=cfg.kernel_host,
>             kernel_port=cfg.kernel_port,
>             health_status=cfg.health_status,
>         )
>         routes.append(route)
>         responses[_route_endpoint(route)] = cfg.response
>     return MetricsScenario(
>         routes=routes,
>         responses=responses,
>         expected_running_value=expected_running_value,
>         expected_replica_count=expected_replica_count,
>     )
> 
> 
> class TestGatherPrometheusInferenceMeasures:
>     """Regression tests for per-route error resilience in metric collection."""
> 
>     @pytest.mark.parametrize(
>         "scenario",
>         [
>             pytest.param(
>                 _create_metrics_scenario(
>                     [
>                         RouteInput(
>                             kernel_host="10.0.0.1",
>                             kernel_port=8080,
>                             health_status=HEALTHY,
>                             response=SAMPLE_PROMETHEUS_OUTPUT,
>                         ),
>                         RouteInput(
>                             kernel_host="10.0.0.2",
>                             kernel_port=8081,
>                             health_status=HEALTHY,
>                             response=ConnectionError("Connection refused"),
>                         ),
>                     ],
>                     expected_running_value=Decimal(3),
>                     expected_replica_count=1,
>                 ),
>                 id="partial-failure-connection-error",
>             ),
>             pytest.param(
>                 _create_metrics_scenario(
>                     [
>                         RouteInput(
>                             kernel_host="10.0.0.1",
>                             kernel_port=8080,
>                             health_status=HEALTHY,
>                             response=ConnectionError("Connection refused"),
>                         ),
>                         RouteInput(
>                             kernel_host="10.0.0.2",
>                             kernel_port=8081,
>                             health_status=HEALTHY,
>                             response=TimeoutError("Timed out"),
>                         ),
>                     ],
>                     expected_running_value=None,
>                     expected_replica_count=0,
>                 ),
>                 id="all-routes-unreachable",
>             ),
>             pytest.param(
>                 _create_metrics_scenario(
>                     [
>                         RouteInput(
>                             kernel_host="10.0.0.1",
>                             kernel_port=8080,
>                             health_status=HEALTHY,
>                             response=SAMPLE_PROMETHEUS_OUTPUT,
>                         ),
>                         RouteInput(
>                             kernel_host="10.0.0.2",
>                             kernel_port=8081,
>                             health_status=HEALTHY,
>                             response=SAMPLE_PROMETHEUS_OUTPUT,
>                         ),
>                     ],
>                     expected_running_value=Decimal(6),
>                     expected_replica_count=2,
>                 ),
>                 id="two-healthy-routes",
>             ),
>             pytest.param(
>                 _create_metrics_scenario(
>                     [
>                         RouteInput(
>                             kernel_host="10.0.0.1",
>                             kernel_port=8080,
>                             health_status=HEALTHY,
>                             response=SAMPLE_PROMETHEUS_OUTPUT,
>                         ),
>                         RouteInput(
>                             kernel_host="10.0.0.2",
>                             kernel_port=8081,
>                             health_status=ModelServiceStatus.UNHEALTHY,
>                             response=SAMPLE_PROMETHEUS_OUTPUT,
>                         ),
>                     ],
>                     expected_running_value=Decimal(3),
>                     expected_replica_count=1,
>                 ),
>                 id="unhealthy-route-skipped",
>             ),
>             pytest.param(
>                 _create_metrics_scenario(
>                     [
>                         RouteInput(
>                             kernel_host="10.0.0.1",
>                             kernel_port=8080,
>                             health_status=HEALTHY,
>                             response=SAMPLE_PROMETHEUS_OUTPUT,
>                         ),
>                         RouteInput(
>                             kernel_host="10.0.0.3",
>                             kernel_port=8082,
>                             health_status=HEALTHY,
>                             response=MALFORMED_PROMETHEUS_OUTPUT,
>                         ),
>                     ],
>                     expected_running_value=Decimal(3),
>                     expected_replica_count=1,
>                 ),
>                 id="malformed-payload",
>             ),
>         ],
>     )
>     async def test_metric_collection(self, scenario: MetricsScenario) -> None:
>         client_pool = _make_client_pool(scenario.responses)
> 
>         measures = await gather_prometheus_inference_measures(client_pool, scenario.routes)
> 
>         if scenario.expected_running_value is None:
>             assert measures == []
>         else:
>             running_measures = [
>                 m for m in measures if m.key == "vllm:num_requests_running"
>             ]
>             assert len(running_measures) == 1, (
>                 f"Expected exactly 1 'vllm:num_requests_running' measure, "
>                 f"got {len(running_measures)}"
>             )
>             running_measure = running_measures[0]
> ```

Refactored test_metrics_resilience.py: replaced 5 NamedTuples/fixtures/test methods with RouteInput/MetricsScenario dataclasses and a single parametrized test_metric_collection.
test_frontend_route_sync.py was intentionally left as-is — each test has unique setup and assertions (timing capture, pytest.raises, endpoint tracking), making parametrize unsuitable.

tests/unit/appproxy/worker/test_metrics_resilience.py
```diff
@@ -103,198 +106,182 @@ async def successful_get(path: str) -> Any:
...
+
+
+@dataclass(frozen=True)
+class MetricsScenario:
+    """Describes a metric collection test case."""
+
+    routes: list[RouteInfo]
+    responses: dict[str, str | Exception]
+    # None means measures should be empty
+    expected_running_value: Decimal | None
```
jopemachine: What I provided was just example code.
The variable name expected_running_value does not seem to clearly convey its intended purpose.

tests/unit/appproxy/worker/test_metrics_resilience.py
```diff
@@ -25,6 +26,8 @@
 from ai.backend.appproxy.worker.types import Measurement
 from ai.backend.common.types import ModelServiceStatus
 
+HEALTHY = ModelServiceStatus.HEALTHY
```
jopemachine: Why do we need this?

lizable: Right. I think it's mistake. I'll fix it.

tests/unit/appproxy/worker/test_metrics_resilience.py
lizable: I improved variable names across test files to better reflect their intent, such as renaming expected_running_value to expected_num_requests_running to match the         
  vllm:num_requests_running metric key used in assertions.

tests/unit/appproxy/worker/test_metrics_resilience.py
```diff
@@ -0,0 +1,293 @@
...
+        last_health_check=None,
+        consecutive_failures=0,
+    )
+
+
+def _route_endpoint(route: RouteInfo) -> str:
+    return f"http://{route.current_kernel_host}:{route.kernel_port}"
+
+
+def _make_client_pool(
```
jopemachine: Let's use aioresponses

lizable: I replaced the hand-written mock helpers (_make_client_pool, mock_load_client_session) with aioresponses, which intercepts real aiohttp.ClientSession requests instead of 
  faking them manually. I moved the repeated setup code into a shared mock_metrics_client_pool fixture in conftest.py to avoid duplication across test files. I also
  switched endpoint call verification to use aioresponses' built-in request tracking (mocked.requests) instead of a custom list.

PR #9765, opened 2026-03-06
feat(BA-4926): adapt validation rules for Strawberry schema execution

fregataa:
## Summary
- Add `GQLValidationExtension` Strawberry `SchemaExtension` that conditionally applies `CustomIntrospectionRule` and query depth limiting
- Reads `allow_graphql_schema_introspection` and `max_gql_query_depth` from config via `StrawberryGQLContext` at request time
- Register the extension at schema creation time in `CustomizedSchema`

## Test plan
- [ ] Verify introspection queries are blocked when `allow_graphql_schema_introspection` is false
- [ ] Verify `__typename` is still allowed even when introspection is disabled
- [ ] Verify query depth limiting works when `max_gql_query_depth` is configured

Resolves BA-4926

### DIFF
--- a/src/ai/backend/manager/api/gql/extensions/validation.py
+++ b/src/ai/backend/manager/api/gql/extensions/validation.py
@@ -1,26 +1,13 @@
 from __future__ import annotations
 
 from collections.abc import Iterator
-from typing import Any
 
-from graphql import GraphQLError, ValidationRule
+from graphql import ValidationRule
 from strawberry.extensions.base_extension import SchemaExtension
 from strawberry.extensions.query_depth_limiter import create_validator as create_depth_validator
 
 from ai.backend.manager.api.gql.types import StrawberryGQLContext
-
-
-class CustomIntrospectionRule(ValidationRule):
-    """Blocks introspection queries (fields starting with ``__``) except ``__typename``."""
-
-    def enter_field(self, node: Any, *_args: Any) -> None:
-        field_name = node.name.value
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/api/gql/extensions/validation.py
```diff
@@ -0,0 +1,47 @@
...
+    """Blocks introspection queries (fields starting with ``__``) except ``__typename``."""
+
+    def enter_field(self, node: Any, *_args: Any) -> None:
+        field_name = node.name.value
+        if field_name.startswith("__"):
+            if field_name == "__typename":
+                return
+            self.report_error(
+                GraphQLError(f"Cannot query '{field_name}': introspection is disabled.", node)
+            )
```
jopemachine: Since this implementation is identical to the existing one in the REST API, would it make sense to move it to common and reuse a single implementation?

src/ai/backend/manager/api/gql/extensions/validation.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class GQLValidationExtension(SchemaExtension):
+    """Conditionally applies introspection blocking and query depth limiting.
+
+    Reads configuration from ``StrawberryGQLContext.config_provider`` to decide
+    which validation rules to add at request time.
+    """
+
+    def on_validate(self) -> Iterator[None]:
```
jopemachine, line 20: Let's add `override` decorator

PR #9767, opened 2026-03-06
feat(BA-4928): implement Strawberry UnfrozenRequired extension

fregataa:
## Summary
- Add `GQLMutationUnfrozenRequiredExtension` that blocks mutations when manager is FROZEN via `on_execute()` hook
- Add `manager_status` field to `StrawberryGQLContext` and wire it in handler

## Test plan
- [ ] Verify mutations are blocked when manager status is FROZEN
- [ ] Verify queries still work when manager is FROZEN

Resolves BA-4928

### DIFF
--- a/changes/9767.feature.md
+++ b/changes/9767.feature.md
@@ -1,0 +1,1 @@
+Add Strawberry UnfrozenRequired extension to block mutations when manager is frozen.


fregataa:
리뷰 결과 not-planned로 결정. PR을 닫습니다.

PR #9790, opened 2026-03-09
fix(BA-4891): Fix kernel container_id attribute shadowed by UserDict data dict in agent

### LINKED ISSUE #9683: Fix kernel container_id attribute shadowed by UserDict data dict in agent
AbstractKernel inherits from UserDict[str, Any], causing a mismatch between the instance attribute and the internal UserDict data dict for container_id.

- AbstractKernel.__init__ initializes self.container_id = None as an instance attribute (kernel.py:232)

- After container startup, agent.py:3066 calls self.kernel_registry[kernel_id].data.update(container_data), which only updates the UserDict data dict

- Due to Python's attribute lookup order, the instance attribute takes precedence, so kernel_obj.container_id always returns None

- repr(kernel_obj) uses UserDict.__repr__() which displays the data dict, so container_id appears to have a valid value

Impact: In collect_container_stat(), the condition kernel_obj.container_id is None evaluates to True, causing container stats collection to be skipped entirely.

Fix: Either sync the instance attribute after data.update(), or unify container_id access to always go through self.data.

JIRA Issue: BA-4891

fregataa:
## Summary
- Fixed kernel container_id attribute shadowing issue where instance attribute remained None after agent restarts
- AbstractKernel inherits from UserDict, causing container_id to exist in both instance attribute and data dict
- Python attribute lookup prioritizes instance attributes, so kernel_obj.container_id always returned None after restarts
- This caused collect_container_stat() to skip container stats collection entirely

## Implementation
Applied three-point fix to sync instance attribute with data dict:
1. In _handle_start_event: sync container_id from event (critical for agent restart recovery)
2. In inject_container_lifecycle_event: sync instance attribute after setting data dict
3. After data.update(container_data): sync instance attribute if container_id present

## Test plan
- [x] Unit tests verify all three fix points sync container_id correctly
- [x] Consistency checks pass (instance attribute matches data dict)
- [x] container_id is None check works correctly for stat collection
- [x] pants fmt, fix, lint all pass

resolves #9683 (BA-4891)
resolves BA-4946

### DIFF
--- a/tests/unit/agent/test_container_id_sync.py
+++ b/tests/unit/agent/test_container_id_sync.py
@@ -8,6 +8,7 @@ and the UserDict data dict in a single call.
 from __future__ import annotations
 
 from unittest.mock import Mock
+from uuid import uuid4
 
 import pytest
 
@@ -20,8 +21,8 @@ from ai.backend.common.types import AgentId, ContainerId, KernelId, SessionId
 @pytest.fixture
 def mock_kernel_obj() -> DockerKernel:
     """Create a mock kernel object for testing."""
-    kernel_id = KernelId("test-kernel-id")
-    session_id = SessionId("test-session-id")
+    kernel_id = KernelId(uuid4())
+    session_id = SessionId(uuid4())
     agent_id = AgentId("test-agent-id")
 
     ownership_data = KernelOwnershipData(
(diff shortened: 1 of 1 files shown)


src/ai/backend/agent/kernel.py
```diff
@@ -233,6 +234,10 @@ def __init__(
         self.state = KernelLifecycleStatus.PREPARING
         self.session_type = session_type
 
+    def set_container_id(self, cid: ContainerId) -> None:
+        self.container_id = cid
```
jopemachine, line 238: Would it be difficult to unify access to container_id?

fregataa: it is the same, but it is harder to track all access to container id field than track setter parts

PR #9802, opened 2026-03-10
fix: remove pants: no-infer-dep from TYPE_CHECKING imports in processors.py

fregataa:
## Summary
- Remove all `# pants: no-infer-dep` comments from `TYPE_CHECKING` imports in `src/ai/backend/manager/services/processors.py`
- The `TYPE_CHECKING` guard alone is sufficient to prevent runtime imports; the `no-infer-dep` comments were also hiding these modules from mypy's sandbox, causing `import-not-found` errors

## Context
PR #9775 (BA-4935) moved all imports in `processors.py` into `if TYPE_CHECKING:` blocks with `# pants: no-infer-dep` to break the transitive dependency chain. While `TYPE_CHECKING` correctly prevents runtime imports, adding `# pants: no-infer-dep` also prevented Pants from including these modules in mypy's sandbox. This caused 49+ `import-not-found` errors for any PR whose transitive dependencies include `processors.py` (e.g., PR #9742).

## Test plan
- [x] `pants check src/ai/backend/manager/services/processors.py` passes (was failing with 49+ errors)
- [x] `pants lint src/ai/backend/manager/services/processors.py` passes
- [x] `pants fmt src/ai/backend/manager/services/processors.py` — no changes needed

### DIFF
--- a/changes/9802.fix.md
+++ b/changes/9802.fix.md
@@ -1,0 +1,1 @@
+Fix mypy `import-not-found` errors in `processors.py` by removing `# pants: no-infer-dep` from `TYPE_CHECKING` imports


fregataa:
Closed this PR since applying `no-infer-dep` is intended

PR #9817, opened 2026-03-10
feat(BA-4967): Add DeploymentSubStep type, DB schema, and repository extensions

jopemachine:
Resolves #9817 BA-4967

## Summary
- Add `DeploymentSubStep` enum and `DeploymentStatusTransitions` type
- Add `sub_step` column DB migration with composite index
- Add ORM column mapping for `sub_step`
- All additive changes, no impact on existing code

Split from #9566. Full feature works after all split PRs are merged.

## Test plan
- [ ] DB migration applies cleanly (`alembic upgrade head`)
- [ ] Manager startup check

### DIFF
--- a/src/ai/backend/manager/repositories/deployment/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/deployment/db_source/db_source.py
@@ -2201,7 +2201,7 @@ class DeploymentDBSource:
             row = result.one_or_none()
             if row is None:
                 return None
-            return row[0]
+            return cast(uuid.UUID | None, row[0])
 
     # -------------------------------------------------------------------------
     # Auto-Scaling Policy Methods (DeploymentAutoScalingPolicyRow)


jopemachine:
## Security & Performance Review -- PR #9817

Reviewed all 11 changed files. Findings below ordered by severity.

---

### [HIGH] Race Condition in `set_deploying_revision` (TOCTOU)

**File**: `src/ai/backend/manager/repositories/deployment/db_source/db_source.py`, lines 2283-2304

The method performs a SELECT to read `current_revision`, then a separate UPDATE with a `deploying_revision IS NULL` guard. Under READ COMMITTED isolation, another transaction can modify the row between these two statements. This means `previous_revision_id` may be stale by the time the UPDATE executes.

While the `deploying_revision IS NULL` guard prevents double-activation (good), the returned `previous_revision_id` could reference an already-swapped revision if a concurrent `complete_deployment_revision_swap` committed between the SELECT and UPDATE. If the caller uses this stale value for rollback decisions, it could target the wrong revision.

**Recommendation**: Use a single UPDATE ... RETURNING statement or SELECT ... FOR UPDATE to make this atomic:
```sql
UPDATE endpoints
SET deploying_revision = :rev, lifecycle_stage = 'DEPLOYING'
WHERE id = :id AND deploying_revision IS NULL
RETURNING current_revision
```

---

### [HIGH] Alembic Migration Chain Conflict

**File**: `src/ai/backend/manager/models/alembic/versions/359f0bbd936c_add_sub_step_column_to_endpoints.py`

This migration declares `down_revision = "b1009fe7f865"`, but two other migrations already reference that same parent:
- `6f2f5d828a52_add_session_is_preemptible.py`
- `df13b2272b61_add_server_default_to_roles_updated_at.py`

Having three migrations branching from the same parent creates an Alembic "multiple heads" situation. `alembic upgrade head` will fail unless a merge migration is created. This blocks deployment.

**Recommendation**: Rebase the down_revision to point to the actual latest head, or create a merge migration.

---

### [MEDIUM] Repository Layer Leaks ORM Row Object

**File**: `src/ai/backend/manager/repositories/deployment/repository.py`, lines 1233-1245

`get_last_deployment_histories` returns `dict[uuid.UUID, DeploymentHistoryRow]` directly to callers. Per the repository layer guardrails (`repositories/CLAUDE.md`), ORM Row objects must not cross above the repository layer. The return type should be converted to the corresponding `data/` type (`DeploymentHistoryData`).

**Impact**: Tight coupling between callers and ORM schema; breaks layer isolation.

---

### [MEDIUM] No DB-Level CHECK Constraint on `sub_step`

**File**: Migration `359f0bbd936c` and `endpoint/row.py`

The `sub_step` column is defined as `sa.String(length=32)` with no CHECK constraint limiting values to valid `DeploymentSubStep` enum members. While the ORM uses `StrEnumType(DeploymentSubStep)` for type coercion, any direct SQL update or raw query can insert arbitrary strings into this column.

Since the `StrategyTransaction.update_sub_steps` method already filters by `lifecycle_stage == DEPLOYING`, adding a CHECK constraint (or using a PostgreSQL enum type) would provide defense-in-depth at the DB level.

---

### [MEDIUM] `StrategyTransaction` Lacks Resilience/Retry Policy

**File**: `src/ai/backend/manager/repositories/deployment/repository.py`, line 1222

The `begin_strategy_transaction()` method is the only public repository method that does not apply `deployment_repository_resilience`. All other public methods use `@deployment_repository_resilience.apply()`. Because `StrategyTransaction` methods involve multiple writes (sub_step update, route creation, revision swap), a transient DB error mid-transaction would fail without retry. This is likely intentional (retrying a partially-committed multi-step transaction is complex), but it should be explicitly documented why resilience is omitted.

---

### [MEDIUM] `DeploymentStatusTransitions.failure` Field Removed (Breaking Change)

**File**: `src/ai/backend/manager/data/deployment/types.py`, lines 178-211

The `failure` field was replaced with `need_retry`, `expired`, and `give_up`. Any existing code referencing `.failure` will break at runtime with `AttributeError`. Since this PR states it is "split from #9566" and "all additive changes, no impact on existing code," this is a contradiction -- the field rename is a breaking change.

**Recommendation**: Grep the full codebase for `.failure` usage on `DeploymentStatusTransitions` to confirm no callers exist yet. If callers exist in other split PRs, document the merge order dependency.

---

### [LOW] Hardcoded Lock ID Without Gap Protection

**File**: `src/ai/backend/manager/defs.py`, line 113

`LOCKID_DEPLOYMENT_DEPLOYING = 229` is inserted between `LOCKID_DEPLOYMENT_DESTROYING = 228` and `LOCKID_SOKOVAN_TARGET_PENDING = 230`. The gap is exactly 1, leaving no room for future deployment lock IDs. This is not a bug, but future additions will require non-contiguous numbering.

---

### [LOW] Missing `__all__` Update in `db_source/__init__.py`

**File**: `src/ai/backend/manager/repositories/deployment/db_source/__init__.py`

`StrategyTransaction` is now exported in `__all__`, which is correct. No issue here -- just confirming this was properly handled.

---

### Summary

| Severity | Count | Key Areas |
|----------|-------|-----------|
| HIGH     | 2     | Race condition in set_deploying_revision; Alembic migration chain conflict |
| MEDIUM   | 3     | ORM row leak; missing CHECK constraint; breaking field rename |
| LOW      | 2     | Lock ID gap; __all__ (clean) |

**Most urgent**: The Alembic migration chain conflict (HIGH) will prevent `alembic upgrade head` from succeeding and must be resolved before merge.

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -155,6 +157,100 @@ class EndpointWithRoutesRawData:
...
+                .values(sub_step=sub_step)
+            )
+            await self._db_sess.execute(stmt)
+
+    async def create_routes(
+        self,
+        rollout: BulkCreator[RoutingRow],
+    ) -> None:
+        """Bulk-create new-revision routes."""
+        await execute_bulk_creator(self._db_sess, rollout)
```
HyeockJinKim: It seems like some unnecessary implementation has been added.

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -155,6 +157,100 @@ class EndpointWithRoutesRawData:
...
+
+    async def drain_routes(
+        self,
+        drain: BatchUpdater[RoutingRow],
+    ) -> None:
+        """Mark old-revision routes for draining."""
+        await execute_batch_updater(self._db_sess, drain)
```
HyeockJinKim: same here.

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -476,26 +573,32 @@ async def get_scaling_group_cleanup_configs(
...
         self,
         db_sess: SASession,
         statuses: list[EndpointLifecycle],
+        sub_steps: list[DeploymentSubStep] | None = None,
     ) -> list[EndpointRow]:
-        """Fetch endpoints by lifecycle statuses."""
+        """Fetch endpoints by lifecycle statuses, optionally filtered by sub_steps."""
+        where_clause: sa.ColumnElement[bool] = EndpointRow.lifecycle_stage.in_(statuses)
+        if sub_steps is not None:
+            where_clause = sa.and_(where_clause, EndpointRow.sub_step.in_(sub_steps))
```
HyeockJinKim, line 499: Please adjust it so that both values must be satisfied, not by separating them as if there were no sub-step.

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -708,6 +811,28 @@ async def _get_last_deployment_histories_bulk(
...
+
+        Returns the most recent history record for each deployment. The caller
+        should compare history.phase with the current phase to determine
+        if attempts should be used or reset to 0.
+        """
+        if not deployment_ids:
+            return {}
+
+        async with self._begin_readonly_session_read_committed() as db_sess:
+            return await self._get_last_deployment_histories_bulk(db_sess, deployment_ids)
```
HyeockJinKim, line 733: What is the context in which this was added?

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -2290,6 +2425,17 @@ async def delete_deployment_policy(
...
+    @actxmgr
+    async def begin_strategy_transaction(self) -> AsyncIterator[StrategyTransaction]:
+        """Begin a transaction that spans multiple strategy-related DB operations.
+
+        All operations on the yielded ``StrategyTransaction`` share a single
+        DB session/transaction, ensuring atomicity across sub_step assignment,
+        route mutations, revision swap, and deploying_revision clear.
+        """
+        async with self._begin_session_read_committed() as db_sess:
+            yield StrategyTransaction(db_sess)
```
HyeockJinKim: I don't understand why this requires a separate transaction.

src/ai/backend/manager/repositories/deployment/repository.py
```diff
@@ -1218,6 +1222,33 @@ async def delete_deployment_policy(
...
 
+    def begin_strategy_transaction(self) -> AsyncContextManager[StrategyTransaction]:
+        """Begin a transaction that spans multiple strategy-related DB operations.
+
+        All operations on the yielded ``StrategyTransaction`` share a single
+        DB session/transaction, ensuring atomicity across sub_step assignment,
+        route mutations, revision swap, and deploying_revision clear.
+        """
+        return self._db_source.begin_strategy_transaction()
+
```
HyeockJinKim: We are not using transactions outside the database in the first place.

PR #9818, opened 2026-03-10
feat(BA-5013): Add deployment strategy evaluation framework and deploying handlers

jopemachine:
## Summary
- Add strategy evaluation framework: evaluator, types, blue-green strategy, rolling-update strategy
- Add `StrategyResultApplier` for persisting evaluation results
- Add `DeployingProvisioningHandler` and `DeployingProgressingHandler` sub-step handlers
- Add `StrategyTransaction` stub and `begin_strategy_transaction` in `DeploymentRepository`
- Unit tests for `StrategyResultApplier`

Implements the BEP-1049 zero-downtime deployment strategy architecture. Depends on #9874 (handler base type changes).

## Jira
[BA-5013](https://lablup.atlassian.net/browse/BA-5013)

## Test plan
- [x] Strategy applier unit tests pass
- [x] `pants check` passes
- [x] `pants lint` passes

[BA-5013]: https://lablup.atlassian.net/browse/BA-5013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ai/backend/manager/sokovan/deployment/coordinator.py
+++ b/src/ai/backend/manager/sokovan/deployment/coordinator.py
@@ -173,8 +173,10 @@ class DeploymentCoordinator:
             if handler.lock_id is not None:
                 lock_lifetime = self._config_provider.config.manager.session_schedule_lock_lifetime
                 await stack.enter_async_context(self._lock_factory(handler.lock_id, lock_lifetime))
+            target_statuses = handler.target_statuses()
+            lifecycle_stages = [s.lifecycle for s in target_statuses]
             deployments = await self._deployment_repository.get_endpoints_by_statuses(
-                handler.target_statuses()
+                lifecycle_stages
             )
             if not deployments:
                 log.trace("No deployments to process for handler: {}", handler.name())
@@ -215,7 +217,8 @@ class DeploymentCoordinator:
         """
         handler_name = handler.name()
         target_statuses = handler.target_statuses()
-        from_status = target_statuses[0] if target_statuses else None
+        from_status = target_statuses[0].lifecycle if target_statuses else None
+        target_lifecycle_stages = [s.lifecycle for s in target_statuses]
 
         # Collect all batch updaters and history specs
         batch_updaters: list[BatchUpdater[EndpointRow]] = []
@@ -246,7 +249,7 @@ class DeploymentCoordinator:
                     spec=EndpointLifecycleBatchUpdaterSpec(lifecycle_stage=next_lifecycle),
                     conditions=[
                         DeploymentConditions.by_ids(endpoint_ids),
-                        DeploymentConditions.by_lifecycle_stages(target_statuses),
+                        DeploymentConditions.by_lifecycle_stages(target_lifecycle_stages),
                     ],
                 )
             )
@@ -285,7 +288,7 @@ class DeploymentCoordinator:
                     spec=EndpointLifecycleBatchUpdaterSpec(lifecycle_stage=failure_lifecycle),
                     conditions=[
                         DeploymentConditions.by_ids(endpoint_ids),
-                        DeploymentConditions.by_lifecycle_stages(target_statuses),
+                        DeploymentConditions.by_lifecycle_stages(target_lifecycle_stages),
                     ],
                 )
             )
--- a/tests/unit/manager/sokovan/deployment/test_coordinator_history.py
+++ b/tests/unit/manager/sokovan/deployment/test_coordinator_history.py
@@ -164,7 +164,9 @@ def mock_handler_with_success(
     mock = MagicMock(spec=DeploymentHandler)
     mock.name = MagicMock(return_value="check_pending")
     mock.lock_id = None
-    mock.target_statuses = MagicMock(return_value=[EndpointLifecycle.PENDING])
+    mock.target_statuses = MagicMock(
+        return_value=[DeploymentLifecycleStatus(lifecycle=EndpointLifecycle.PENDING)]
+    )
     mock.status_transitions = MagicMock(
         return_value=DeploymentStatusTransitions(
             success=DeploymentLifecycleStatus(lifecycle=EndpointLifecycle.CREATED),
@@ -189,7 +191,9 @@ def mock_handler_with_failure(
     mock = MagicMock(spec=DeploymentHandler)
     mock.name = MagicMock(return_value="check_pending")
     mock.lock_id = None
-    mock.target_statuses = MagicMock(return_value=[EndpointLifecycle.PENDING])
+    mock.target_statuses = MagicMock(
+        return_value=[DeploymentLifecycleStatus(lifecycle=EndpointLifecycle.PENDING)]
+    )
     mock.status_transitions = MagicMock(
         return_value=DeploymentStatusTransitions(
             success=DeploymentLifecycleStatus(lifecycle=EndpointLifecycle.CREATED),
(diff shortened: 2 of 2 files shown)


jopemachine:
## Code Review: PR #9818 — Deployment Strategy Evaluation Framework

### Summary

This PR introduces a strategy evaluation framework for deployment lifecycle management (BEP-1049). The overall architecture is well-structured with clean separation of concerns (evaluator, applier, types, strategy registry, handlers). Below are the findings organized by severity.

---

### Findings

#### [HIGH-1] `DeploymentStatusTransitions` field mismatch — will fail at runtime

**Files:** `deploying.py:179-184`, `destroying.py:61-66`, `reconcile.py:62-65`

The `DeploymentStatusTransitions` dataclass (in `data/deployment/types.py`) currently only defines `success` and `failure` fields:

```python
@dataclass(frozen=True)
class DeploymentStatusTransitions:
    success: DeploymentLifecycleStatus | None = None
    failure: DeploymentLifecycleStatus | None = None
```

But this PR uses `need_retry`, `expired`, and `give_up` keyword arguments that do not exist on this class. This will raise `TypeError` at import time. The docstring even says "Deployment handlers only have success/failure outcomes (no expired/give_up)."

The PR description states "Split from #9566. Full feature works after all split PRs are merged." — presumably another sibling PR modifies `DeploymentStatusTransitions`. This is acceptable for a split PR workflow, but the dependency should be clearly documented and the merge order enforced.

**Recommendation:** Add a note in the PR description listing the sibling PRs that must be merged first or together, and update the `DeploymentStatusTransitions` docstring to match the new fields.

---

#### [HIGH-2] `DeploymentSubStep` does not exist on this branch

**Files:** `deploying.py:23`, `applier.py:11`, `types.py:15`, `test_applier.py:12`

`DeploymentSubStep` is imported from `ai.backend.manager.data.deployment.types` but is not defined in that module on this branch. This means all files in this PR will fail to import. Same as HIGH-1, this appears to be a sibling PR dependency.

---

#### [MEDIUM-1] Broad `except Exception` in evaluator swallows all errors

**File:** `evaluator.py:96`

```python
except Exception as e:
    log.warning("deployment {}: evaluation error — {}", deployment.id, e)
    result.errors.append(EvaluationErrorData(deployment=deployment, reason=str(e)))
    continue
```

This catches all exceptions including programming errors (e.g., `AttributeError`, `KeyError`, `TypeError`). While this prevents one deployment's evaluation failure from blocking others, it hides bugs during development and makes debugging difficult. The error is only logged at `warning` level and the exception traceback is lost (`reason=str(e)` discards the traceback).

**Recommendation:** Either narrow the catch to domain-specific exceptions (`InvalidDeploymentStrategy`, `InvalidDeploymentStrategySpec`), or at minimum log with `log.warning(..., exc_info=True)` / `log.exception(...)` to preserve the traceback. Consider re-raising unexpected exceptions after logging.

---

#### [MEDIUM-2] Test assertion bug in `test_assignments_only`

**File:** `test_applier.py:169`

The test `test_assignments_only` uses the `provisioning_summary` fixture which creates a summary with only a PROVISIONING assignment and **no rollout specs**. Looking at the applier code:

```python
rollout: BulkCreator[RoutingRow] = BulkCreator(
    specs=[c.spec for c in changes.rollout_specs],  # empty list
)
...
if rollout.specs:  # False — empty list
    await txn.create_routes(rollout)  # NOT called
```

But the test asserts:
```python
mock_txn.create_routes.assert_called_once()  # Will fail — create_routes is NOT called
```

This assertion is incorrect. When the summary has only assignments and no rollout specs, `create_routes` should NOT be called. The assertion should be `assert_not_called()`.

---

#### [MEDIUM-3] `DeployingProvisioningHandler` discards `apply_result` return value

**File:** `deploying.py:101`

```python
await self._applier.apply(summary)
```

Unlike `DeployingProgressingHandler` (line 189) which captures the `apply_result` to determine completed/rolled-back deployments, the provisioning handler discards the result entirely. The `successes` list is built from `summary.assignments.keys()` instead, which means:

1. If the applier's transaction fails partway, the handler still reports all evaluated deployments as successes.
2. The handler cannot distinguish between deployments that were actually persisted vs. those whose DB writes failed.

**Recommendation:** Capture and use the `apply_result` to build the success list, or document why it is intentionally discarded for the provisioning sub-step.

---

#### [MEDIUM-4] Both DEPLOYING handlers share the same lock (`LOCKID_DEPLOYMENT_DEPLOYING`) but operate on different sub-steps

**Files:** `deploying.py:76`, `deploying.py:155`

Both `DeployingProvisioningHandler` and `DeployingProgressingHandler` use `LockID.LOCKID_DEPLOYMENT_DEPLOYING`. If the coordinator runs these handlers concurrently, they will contend on the same advisory lock. This is likely intentional for safety (preventing concurrent provisioning and progressing), but it means:

- A slow provisioning cycle blocks progressing evaluation
- A slow progressing cycle blocks provisioning evaluation

**Recommendation:** Document whether this mutual exclusion is intentional. If provisioning and progressing should be able to run concurrently, consider separate lock IDs.

---

#### [MEDIUM-5] `post_process()` unconditionally re-schedules regardless of result

**Files:** `deploying.py:118-122`, `deploying.py:225-230`

Both deploying handlers' `post_process()` methods unconditionally call `mark_lifecycle_needed()` regardless of whether the execution result had successes or not (or even if ALL deployments errored). This means:

- If all deployments failed strategy evaluation, the next cycle is still triggered
- For `DeployingProgressingHandler`, even when all deployments are COMPLETED or ROLLED_BACK, the DEPLOYING lifecycle is re-scheduled

The other handlers (e.g., `DestroyingDeploymentHandler.post_process`) check `if result.successes:` before triggering downstream lifecycle scheduling.

**Recommendation:** Consider guarding re-scheduling behind a check for whether there are still active (non-terminal) deployments in the DEPLOYING state.

---

#### [LOW-1] Unused `conftest.py` fixture

**File:** `tests/unit/manager/sokovan/deployment/strategy/conftest.py`

The `database_connection` fixture is defined but not used by any test in `test_applier.py` (all tests use mocks). This is dead code that creates an unnecessary dependency on `postgres_container`.

**Recommendation:** Remove it until integration tests that need it are added.

---

#### [LOW-2] Strategy stubs raise `NotImplementedError` — no guard in evaluator

**Files:** `blue_green.py:35`, `rolling_update.py:35`

Both strategy implementations raise `NotImplementedError` in `evaluate_cycle()`. The evaluator's broad `except Exception` (MEDIUM-1 above) will catch this and log it as a warning, but it means any deployment configured with these strategies will silently fail with a generic "evaluation error" message.

**Recommendation:** Consider adding an explicit check in the evaluator or strategy registry that prevents dispatching to unimplemented strategies, with a clearer error message.

---

#### [LOW-3] `StrategyEvaluationSummary` uses mutable default fields

**File:** `types.py:50-67`

`StrategyEvaluationSummary` is a mutable `@dataclass` (not frozen) with `field(default_factory=...)` for `assignments`, `route_changes`, and `errors`. The evaluator creates a new instance per `evaluate()` call, so the default-factory pattern is safe. However, the `RouteChanges` sub-dataclass is also mutable and shared across deployments via `result.route_changes`. This is intentional (aggregation), but means concurrent modification would be unsafe if the evaluator were ever parallelized.

**Recommendation:** Document that the evaluator must process deployments sequentially, not concurrently, or use thread-safe collections.

---

#### [LOW-4] `evaluate_cycle` is synchronous — potential blocking in async context

**File:** `evaluator.py:95`

`strategy.evaluate_cycle()` is a synchronous method called from within the async `evaluate()` method. Currently the strategy implementations are stubs, but if future implementations involve CPU-heavy computation (e.g., complex route analysis), this could block the event loop.

**Recommendation:** This is fine for now since the method is designed as a pure FSM evaluation, but consider documenting that `evaluate_cycle` must remain non-blocking, or provide an async override path for strategies that need I/O.

---

### Overall Assessment

The architecture is well-designed with clear responsibilities: evaluator orchestrates, strategy FSM decides, applier persists. The handler-coordinator integration pattern is consistent with existing handlers. Key areas to address before merging:

1. **HIGH items** are acceptable given the split-PR workflow, but the merge order and dependencies should be explicitly documented.
2. **MEDIUM-1** (broad exception catch) and **MEDIUM-2** (test bug) should be fixed.
3. **MEDIUM-3** and **MEDIUM-5** represent correctness gaps that could lead to inconsistent state if the applier transaction partially fails or when all deployments reach terminal states.

jopemachine:
Superseded by #9874 (handler type changes) and #9880 (strategy framework). Split for smaller review scope.

PR #9819, opened 2026-03-10
feat(BA-4969): Refactor deployment coordinator with `sub_step` support

jopemachine:
Resolves #9819 BA-4969

## Summary
- Refactor coordinator to use flat HandlerRegistry with (lifecycle, sub_step) keys
- Add retry logic with _AttemptContext and failure classification
- Add sub_step parameter to deployment events and schedule client
- Update event dispatcher handlers for sub_step support
- Update deployment service and controller

Split from #9566. Full feature works after all split PRs are merged.

## Test plan
- [ ] `pants test tests/unit/manager/sokovan/deployment::` passes
- [ ] Manager startup check

### DIFF
--- a/changes/9819.feature.md
+++ b/changes/9819.feature.md
@@ -1,1 +1,1 @@
-Refactor deployment coordinator with flat HandlerRegistry, retry logic, and sub_step event support.
+Refactor deployment coordinator with flat HandlerRegistry and sub_step event support.
--- a/src/ai/backend/manager/sokovan/deployment/coordinator.py
+++ b/src/ai/backend/manager/sokovan/deployment/coordinator.py
@@ -382,6 +382,11 @@ class DeploymentCoordinator:
         and phase_started_at from DeploymentWithHistory embedded in each error,
         then applies per-category transitions. All transitions are processed
         in a single transaction.
+
+        Args:
+            handler: The route handler that was executed
+            result: The result of the handler execution
+            records: Execution records from the recorder context
         """
         handler_name = handler.name()
         target_statuses = handler.target_statuses()
@@ -484,33 +489,33 @@ class DeploymentCoordinator:
         next_lifecycle = lifecycle_status.lifecycle
         from_status = target_lifecycles[0].lifecycle if target_lifecycles else None
         target_lifecycle_stages = [status.lifecycle for status in target_lifecycles]
-        endpoint_ids = [dwh.deployment_info.id for dwh in deployments]
+        endpoint_ids = [deployment.deployment_info.id for deployment in deployments]
         history_specs = [
             DeploymentHistoryCreatorSpec(
-                deployment_id=dwh.deployment_info.id,
+                deployment_id=deployment.deployment_info.id,
                 phase=handler_name,
                 result=SchedulingResult.SUCCESS,
                 message=f"{handler_name} completed successfully",
                 from_status=from_status,
                 to_status=next_lifecycle,
                 sub_steps=self._build_history_sub_steps(
-                    dwh.deployment_info.id,
+                    deployment.deployment_info.id,
                     records,
-                    dwh.deployment_info.sub_step,
+                    deployment.deployment_info.sub_step,
                     SchedulingResult.SUCCESS,
                 ),
             )
-            for dwh in deployments
+            for deployment in deployments
         ]
         events = [
             self._build_lifecycle_notification_event(
-                deployment=dwh.deployment_info,
+                deployment=deployment.deployment_info,
                 from_status=from_status,
                 to_status=next_lifecycle,
                 transition_result="success",
                 timestamp=timestamp_now,
             )
-            for dwh in deployments
+            for deployment in deployments
         ]
         updater = self._build_lifecycle_updater(
             endpoint_ids, lifecycle_status, target_lifecycle_stages


jopemachine:
## Code Review: PR #9819 - Coordinator Flat Registry and Event sub_step Support

Reviewed files:
- `src/ai/backend/manager/sokovan/deployment/coordinator.py`
- `src/ai/backend/common/clients/valkey_client/valkey_schedule/client.py`
- `src/ai/backend/common/events/event_types/schedule/anycast.py`
- `src/ai/backend/manager/event_dispatcher/handlers/schedule.py`
- `src/ai/backend/manager/services/deployment/service.py`
- `src/ai/backend/manager/sokovan/deployment/deployment_controller.py`
- `tests/unit/manager/sokovan/deployment/test_coordinator_history.py`

---

### Finding 1 [MEDIUM] - Handler registry key derivation uses only the first target_status, creating a silent dispatch mismatch

**File:** `coordinator.py`, `_derive_handler_key()` (lines 352-369)

The `DeployingProgressingHandler.target_statuses()` returns three entries: `PROGRESSING`, `COMPLETED`, and `ROLLED_BACK`. However, `_derive_handler_key()` uses only `statuses[0].sub_status` (i.e., `PROGRESSING`) as the registry key. This means:

1. There is no handler registered for `(DEPLOYING, COMPLETED)` or `(DEPLOYING, ROLLED_BACK)`.
2. If the DB has endpoints with `sub_step=COMPLETED` or `sub_step=ROLLED_BACK` and an event arrives targeting those sub_steps specifically, `process_deployment_lifecycle()` will find no handler and silently skip them (the log warning at line 379 fires but no processing happens).
3. The current design relies on the PROGRESSING handler fetching all three sub_steps from the DB (line 399-406), so it works as a batch, but the registry key semantic is misleading.

**Recommendation:** Either document this intentional "primary sub-step keying" design prominently, or register the handler under all three keys for explicit dispatch.

---

### Finding 2 [MEDIUM] - Event serialization sends `None` as a tuple element, which may not round-trip safely through msgpack

**File:** `anycast.py`, `DoDeploymentLifecycleEvent.serialize()` and `DoDeploymentLifecycleIfNeededEvent.serialize()` (lines 112-113, 149-150)

Both events serialize as `(self.lifecycle_type, self.sub_step)` where `self.sub_step` can be `None`. When serialized through msgpack, `None` becomes msgpack `nil`, which deserializes back to `None` in Python. However, the `AbstractEvent.serialize()` type hint declares `tuple[bytes, ...]` (line 70 of `types.py`), but these overrides return `tuple[Any, ...]`. While this may work in practice because the event dispatcher handles `Any` tuples, the type contract is violated.

Additionally, the `deserialize` method uses `value[1] if len(value) > 1 else None` (line 118, 156), which is a defensive guard for backward compatibility. However, if an older producer sends events with only `(lifecycle_type,)` (a single-element tuple), and a new consumer receives it, the `sub_step` correctly defaults to `None`. Conversely, if a new producer sends `(lifecycle_type, sub_step)` and an older consumer that only expects `(lifecycle_type,)` receives it, the older consumer will fail because it indexes only `value[0]`.

**Impact:** During rolling upgrades where managers are at different versions, event deserialization could break. The existing `DoSokovanProcessIfNeededEvent` and `DoSokovanProcessScheduleEvent` events only serialize/deserialize `(schedule_type,)`, so the new two-element tuple is a breaking change for these event types.

**Recommendation:** Since `DoDeploymentLifecycleEvent` and `DoDeploymentLifecycleIfNeededEvent` are new event classes (not modifications of existing ones), this is actually safe -- the concern only applies if the event class was renamed or reused. Verify that these event classes did not exist before this PR with a different tuple format.

---

### Finding 3 [LOW] - _AttemptContext.is_expired uses `replace(tzinfo=UTC)` for naive datetimes, which is incorrect for non-UTC databases

**File:** `coordinator.py`, lines 114-123

The code does:
```python
current_utc = (
    current_dbtime.astimezone(UTC)
    if current_dbtime.tzinfo
    else current_dbtime.replace(tzinfo=UTC)
)
```

`replace(tzinfo=UTC)` assumes the naive datetime is already in UTC. If the database returns naive datetimes in a different timezone (e.g., local server time), this would produce incorrect timeout calculations. However, the comment at line 112-113 acknowledges this: "DB may return either depending on driver."

**Recommendation:** This is acceptable given the documented assumption that DB time is in UTC. Consider adding an assertion or log warning if `current_dbtime` is naive to detect misconfigured environments early.

---

### Finding 4 [LOW] - Notification events are fire-and-forget with no batching

**File:** `coordinator.py`, lines 530-535

```python
for event in notification_events:
    try:
        await self._event_producer.anycast_event(event)
    except Exception as e:
        log.warning("Failed to send lifecycle notification: {}", e)
```

Each notification is sent individually in a loop with per-event exception swallowing. For deployments with many endpoints, this could result in many individual I/O calls. If anycast_event is a network call (Valkey/Redis publish), this could be slow.

**Recommendation:** Consider batching notification events if the event producer supports it. Also, the broad `except Exception` here is acceptable since this is a best-effort notification path, but be aware it will swallow unexpected errors (e.g., `TypeError` from bad serialization).

---

### Finding 5 [LOW] - `_build_history_sub_steps` creates timestamps with `datetime.now(UTC)` instead of using DB time

**File:** `coordinator.py`, lines 704-713

```python
now = datetime.now(UTC)
sub_steps.append(
    SubStepResult(
        step=sub_status.value,
        result=scheduling_result,
        started_at=now,
        ended_at=now,
    )
)
```

This uses the application server's clock rather than the database clock. In `_classify_failures` (line 491), the code correctly fetches `current_dbtime = await self._deployment_repository.get_db_now()`. However, `_build_history_sub_steps` uses `datetime.now(UTC)`. For consistency and to avoid clock drift issues in distributed deployments, the DB time should be used for all recorded timestamps.

**Recommendation:** Pass the DB time into `_build_history_sub_steps` or use the `timestamp_now` already available in the caller scope.

---

### Finding 6 [LOW] - `_build_lifecycle_notification_event` also uses `datetime.now(UTC)` for the event timestamp

**File:** `coordinator.py`, lines 738-739

```python
return NotificationTriggeredEvent(
    rule_type=NotificationRuleType.ENDPOINT_LIFECYCLE_CHANGED.value,
    timestamp=datetime.now(UTC),
    ...
)
```

Same clock concern as Finding 5. The `timestamp_now` (ISO string of `datetime.now(UTC)`) is used for the `event_timestamp` field of the message, but a fresh `datetime.now(UTC)` is created for the event's own `timestamp` field. These can differ by a few milliseconds, which is probably acceptable, but using a single timestamp source would be more consistent.

---

### Finding 7 [MEDIUM] - DeploymentTaskSpec sub_step ordering depends on dict insertion order

**File:** `coordinator.py`, `HandlerRegistry.sub_steps_for()` (lines 163-169) and `_create_task_specs()` (lines 798-807)

`sub_steps_for()` returns sub-steps by iterating over `self.handlers` dict keys:
```python
return [
    sub_step
    for lt, sub_step in self.handlers
    if lt == lifecycle_type and sub_step is not None
]
```

The order of task specs for DEPLOYING sub-steps depends on the insertion order of `self.handlers` (Python 3.7+ dicts preserve insertion order). While this currently works because handlers are inserted in order (PROVISIONING first, then PROGRESSING), this is an implicit contract. If handler registration order changes, task scheduling intervals and initial delays could be applied to the wrong sub-step tasks.

**Recommendation:** Consider sorting the sub-steps explicitly or documenting the ordering contract.

---

### Finding 8 [MEDIUM] - `_run_handler` exception in `handler.execute()` is not caught, causing the entire lifecycle processing to fail

**File:** `coordinator.py`, `_run_handler()` (lines 392-426)

```python
async def _run_handler(self, handler):
    ...
    with DeploymentRecorderContext.scope(handler_name, entity_ids=deployment_ids) as pool:
        result = await handler.execute(deployments)  # No try/except
        all_records = pool.build_all_records()
        await self._handle_status_transitions(handler, result, all_records, attempt_ctx_map)

    try:
        await handler.post_process(result)
    except Exception as e:
        log.error(...)
```

`handler.execute()` is called without a try/except. If `execute()` raises an unhandled exception, the entire `process_deployment_lifecycle` call fails, and no status transitions or history records are created. The deployments remain in their current state and will be retried on the next cycle, but there is no failure record. Meanwhile, `post_process()` is correctly wrapped in try/except.

**Recommendation:** Wrap `handler.execute()` in a try/except to create failure records for all deployments when execution fails unexpectedly, allowing the retry/give_up classification to work correctly.

---

### Finding 9 [LOW] - Test fixtures use `Generator` instead of `Iterator` for non-yielding patterns

**File:** `test_coordinator_history.py`, lines 255, 288

The coordinator fixtures use `Generator[DeploymentCoordinator, None, None]` as the return type, but since they only `yield` once and do no teardown, they could simply use `Iterator`. This is a minor stylistic concern.

---

### Finding 10 [LOW] - `process_deployment_lifecycle` acquires a lock even when no deployments exist

**File:** `coordinator.py`, `process_deployment_lifecycle()` (lines 371-390)

The distributed lock is acquired before calling `_run_handler()`, which then checks if deployments exist (line 408). This means the lock is acquired and held even when there are no deployments to process. For frequently-polled lifecycle types with no pending work, this adds unnecessary lock contention.

**Recommendation:** Consider checking for deployments before acquiring the lock, or document why the lock is needed even when no deployments exist (e.g., to prevent concurrent handler initialization).

---

### Summary

| # | Severity | Category | Description |
|---|----------|----------|-------------|
| 1 | MEDIUM | Correctness | Handler registry key uses only first target_status; `COMPLETED` and `ROLLED_BACK` sub-steps have no explicit registry entries |
| 2 | MEDIUM | Compatibility | Event serialization tuple size changed; safe for new event classes but verify no renaming occurred |
| 3 | LOW | Correctness | Naive datetime `replace(tzinfo=UTC)` assumes DB returns UTC |
| 4 | LOW | Performance | Notification events sent individually without batching |
| 5 | LOW | Consistency | `_build_history_sub_steps` uses app clock instead of DB clock |
| 6 | LOW | Consistency | Notification event uses separate `datetime.now(UTC)` call |
| 7 | MEDIUM | Correctness | Task spec sub_step ordering depends on implicit dict insertion order |
| 8 | MEDIUM | Robustness | Unhandled exception in `handler.execute()` skips failure recording |
| 9 | LOW | Code Quality | Test fixture uses `Generator` instead of `Iterator` |
| 10 | LOW | Performance | Lock acquired before checking if deployments exist |

No CRITICAL or HIGH severity issues found. The overall design is sound with a clean separation of concerns between handler registry, dispatch, retry classification, and status transitions.

src/ai/backend/manager/repositories/deployment/repository.py
```diff
@@ -1140,6 +1140,21 @@ async def update_current_revision(
...
+        revision_id: uuid.UUID,
+    ) -> tuple[uuid.UUID | None, int]:
+        """Set the deploying revision for a deployment.
+
+        Returns:
+            Tuple of (previous_revision_id, rowcount).
+            rowcount == 0 means a concurrent activation guard fired.
+        """
+        previous = await self._db_source.update_current_revision(endpoint_id, revision_id)
+        return previous, 1
```
HyeockJinKim: 1?

HyeockJinKim (approved):

PR #9897, opened 2026-03-11
test(BA-4984): add component tests for auth core (authorize, signup, signout)

seedspirit:
## Summary
- Add component tests for authentication core covering authorize, signup, and signout flows
- Test KEYPAIR authorization with valid credentials, wrong password, INACTIVE/BEFORE_VERIFICATION users, password expiry, and unsupported auth type
- Test signup with valid data and duplicate email rejection, signout with user deactivation and keypair invalidation

## Test plan
- [x] `pants test tests/component/auth/` passes all 9 tests
- [x] `pants lint` passes with no issues

Resolves BA-4984

### DIFF
--- a/tests/component/auth/test_auth_core.py
+++ b/tests/component/auth/test_auth_core.py
@@ -163,8 +163,8 @@ class TestSignup:
                             association_groups_users.c.user_id == user_uuid
                         )
                     )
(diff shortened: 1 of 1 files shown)


jopemachine:
Does it need to be separated from `test_auth.py`?

PR #9899, opened 2026-03-11
feat(BA-4923): Implement Prometheus instant query in client and preset execution service

seedspirit:
## Summary
- Add `PrometheusQueryInstantResponse` DTO for instant query results (singular `value` instead of `values` array)
- Add `query_instant()` method to `PrometheusClient` that POSTs to `/api/v1/query` with rendered PromQL and optional `time` parameter
- Update `execute_preset()` service method to call `query_instant()` when `time_range` is `None` instead of raising `InvalidAPIParameters`
- Add unit tests for both the client method and the service branching logic

## Test plan
- [ ] Unit tests for `query_instant()` client method (success, time param, HTTP errors, connection errors)
- [ ] Unit tests for service branching (instant query when `time_range=None`, range query when `time_range` provided)
- [ ] Existing range query tests continue to pass (no regression)

Resolves BA-4923

### DIFF
--- a/tests/unit/common/clients/prometheus/test_client.py
+++ b/tests/unit/common/clients/prometheus/test_client.py
@@ -225,6 +225,9 @@ class TestQueryInstant:
 
         assert isinstance(result, PrometheusResponse)
         mock_session.post.assert_called_once()
+        form_data = mock_session.post.call_args.kwargs["data"]
+        field_values = {field[0]["name"]: field[2] for field in form_data._fields}
+        assert field_values["time"] == "1704067200.123"
 
     @pytest.fixture
     def error_4xx_response(self, mock_session: Mock) -> AsyncMock:


src/ai/backend/common/dto/clients/prometheus/response.py
```diff
@@ -22,23 +24,34 @@ class MetricResponseInfo(BaseModel):
...
     metric: MetricResponseInfo
     values: list[MetricResponseValue]
 
+    @model_validator(mode="before")
+    @classmethod
+    def _normalize_values(cls, data: Any) -> Any:
+        if isinstance(data, dict) and "value" in data and "values" not in data:
+            data["values"] = [data.pop("value")]
+        return data
+
```
fregataa, line 42: Is this for handling both value and values ?

seedspirit: yes, trying to handle both instant query and range query in one response class

fregataa: I think it is not good to implement such adapter function in data type. please add an issue to refactor this part

PR #9902, opened 2026-03-11
test(BA-5008): add component tests for VFS storage and storage namespace

seedspirit:
## Summary
- Add component tests for VFS storage CRUD operations (create, get by ID/name, search, update, delete)
- Add VFS storage quota management tests using AsyncMock for StorageProxyManagerFacingClient (get/set/unset quota)
- Add component tests for storage namespace operations (register, unregister, per-storage listing, grouped query, BatchQuerier search, duplicate registration error)

## Test plan
- [x] All 19 tests pass (11 VFS storage + 8 storage namespace)
- [x] Quota tests use proper AsyncMock, no xfail markers
- [x] pants lint passes on all changed files

Resolves BA-5008


tests/component/vfs_storage/test_vfs_storage_full.py
jopemachine, line 1: It doesn't seem necessary to separate a file named full.

seedspirit:
> Is threre no SDK related to storage namespace? and what is different between `test_storage.py` and `test_storage_full.py`

@fregataa  Since there was no difference between `test_storage.py` and `test_storage_full.py`, I combined them into one.

seedspirit:
> Is threre no SDK related to storage namespace? and what is different between `test_storage.py` and `test_storage_full.py`

Yes, there is currently no dedicated `StorageNamespaceClient` in the client SDK. Storage namespace operations are only exposed via GraphQL on the manager side. As a result, the core storage namespace tests (`TestStorageNamespace`) call `StorageNamespaceProcessors` directly instead of going through the client SDK. Only the bucket-related tests (`TestStorageNamespaceRegisterUnregister`) use the client SDK, and even those go through `object_storage` rather than a storage namespace–specific SDK.

fregataa (approved):

PR #9989, opened 2026-03-12
refactor(BA-5047): migrate deployment action classes to RBAC base classes

fregataa:
## Summary
- Add `DeploymentScopeAction`, `DeploymentSingleEntityAction` and replica-specific RBAC base classes
- Migrate 5 single-entity actions: `GetDeploymentById`, `UpdateDeployment`, `DestroyDeployment`, `SyncReplicas`, `GetReplicaById`
- Each implements `target_element()` returning `RBACElementRef(MODEL_DEPLOYMENT, id)`
- Upgrade 5 processors to `SingleEntityActionProcessor`
- Keep `CreateDeployment`, `CreateLegacyDeployment`, `SearchDeployments`, `SearchReplicas` as base actions

## Test plan
- [x] `pants fmt/lint/check` pass on `services/deployment::`
- [x] Tests pass
- [x] CI passes

Resolves BA-5047

### DIFF
--- a/changes/9989.enhance.md
+++ b/changes/9989.enhance.md
@@ -1,0 +1,1 @@
+Refactor deployment action classes to RBAC-aware base classes (`DeploymentSingleEntityAction`) to enable RBAC validator wiring


fregataa:
Closing: the replica action's `target_element()` uses `replica_id` as `MODEL_DEPLOYMENT` element ID, which is semantically incorrect. The current API flow only passes `replica_id` without `deployment_id`, and `target_element()` is a synchronous method called before the service method executes (i.e., before DB lookup), so we cannot resolve the correct deployment_id at action construction time. This needs to be addressed alongside the API changes before wiring RBAC validators.

PR #9999, opened 2026-03-12
refactor(BA-4521): replace ScopeType/EntityType with RBACElementType in RBAC layers

fregataa:
## Summary
- Replace `ScopeType` and RBAC `EntityType` references with the unified `RBACElementType` enum across permission controller (BA-4525), domain repositories/services (BA-4526), and REST API/DTO layers (BA-4612)
- Use bridge methods (`to_scope_type()`, `to_entity_type()`, `to_element()`) at DB boundaries where ORM columns still store the old types
- Handle GLOBAL scope as `None` since it has no `RBACElementType` equivalent — handler returns empty result for GLOBAL scope searches

## Test plan
- [ ] Verify RBAC permission CRUD operations via REST API
- [ ] Verify scope search endpoints handle DOMAIN, PROJECT, USER, and GLOBAL correctly
- [ ] Verify entity search endpoints return correct results
- [ ] Verify vfolder creation/listing with scope types
- [ ] Run `pants check --changed-since=origin/main` in CI

Resolves BA-4525, BA-4526, BA-4612

### DIFF
--- a/changes/9999.enhance.md
+++ b/changes/9999.enhance.md
@@ -1,0 +1,1 @@
+Refactor RBAC layers to use unified RBACElementType enum replacing ScopeType and EntityType


src/ai/backend/manager/api/rest/rbac/scope_adapter.py
```diff
@@ -36,16 +36,22 @@
...
     """Adapter for converting scope requests to BatchQuerier objects."""
 
-    def build_querier(self, scope_type: ScopeType, request: SearchScopesRequest) -> BatchQuerier:
-        """Build a BatchQuerier based on scope type."""
+    def build_querier(
+        self, scope_type: RBACElementType | None, request: SearchScopesRequest
+    ) -> BatchQuerier:
+        """Build a BatchQuerier based on scope type.
+
+        ``scope_type`` is ``None`` for the GLOBAL scope, which has no
```
jopemachine, line 44: Would it be possible to add GLOBAL as an enum value?
I think it doesn’t seem good to treat scope_type as GLOBAL when it is None or an empty string.

fregataa:
Splitting into separate PRs per issue: BA-4525 and BA-4526. BA-4612 deferred.

PR #10032, opened 2026-03-13
feat(BA-3691): Apply RBAC validator for Model Artifact Registry actions

fregataa:
## Summary
- Added RBAC base classes (`ArtifactRegistryScopeAction`, `ArtifactRegistrySingleEntityAction`) that return `entity_type() = "ARTIFACT_REGISTRY"`
- Refactored artifact registry actions to extend RBAC base classes and implement required methods (`target_entity_id()`, `scope_type()`, `scope_id()`, `permission_operation_type()`)
- Wired RBAC validators to processors: `ScopeActionProcessor` with `validators.rbac.scope` for scope actions, `SingleEntityActionProcessor` with `validators.rbac.single_entity` for single-entity actions
- Kept internal/batch actions (get_huggingface_registries, get_reservoir_registries, get_registry_meta, get_registry_metas) on plain `ActionProcessor` without RBAC validators

## Test plan
- [x] pants fmt passes
- [x] pants fix passes
- [x] pants lint passes
- [ ] CI type checks pass
- [ ] CI tests pass

Resolves BA-3691

### DIFF
--- a/src/ai/backend/manager/services/artifact_registry/actions/reservoir/list.py
+++ b/src/ai/backend/manager/services/artifact_registry/actions/reservoir/list.py
@@ -1,44 +1,28 @@
 from dataclasses import dataclass
 from typing import override
 
-from ai.backend.common.data.permission.types import RBACElementType, ScopeType
+from ai.backend.manager.actions.action import BaseActionResult
 from ai.backend.manager.actions.types import ActionOperationType
-from ai.backend.manager.data.permission.types import RBACElementRef
 from ai.backend.manager.data.reservoir_registry.types import ReservoirRegistryData
-from ai.backend.manager.services.artifact_registry.actions.base import (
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/services/artifact_registry/actions/huggingface/list.py
```diff
@@ -1,28 +1,44 @@
...
 
     @override
-    def entity_id(self) -> str | None:
-        return None
+    def scope_type(self) -> ScopeType:
+        return ScopeType.GLOBAL
+
+    @override
+    def scope_id(self) -> str:
+        return ""
```
jopemachine, line 44: Is scope_id for the global scope supposed to be an empty string?
Wouldn't it be better to treat it as None or a dedicated sentinel value instead?

PR #10055, opened 2026-03-13
feat(BA-3696): Apply RBAC validator for User actions

fregataa:
## Summary
- Implemented RBAC validator integration for User service following the established pattern from Group, VFolder, and Session services
- Scope actions (create, search) now use `BaseScopeAction` base class with `ScopeActionProcessor` 
- Single-entity actions (get, update, delete, purge) now use `BaseSingleEntityAction` base class with `SingleEntityActionProcessor`
- Added RBAC-aware base classes `UserScopeAction` and `UserSingleEntityAction` to define entity type and required methods
- Updated all action result classes to inherit from RBAC result bases and implement scope/entity ID methods
- Wired RBAC validators from `ActionValidators` to processors for all user actions

## Test plan
- [x] `pants fmt` passes with no changes
- [x] `pants fix` passes with no changes  
- [x] CI type checks pass
- [x] CI tests pass

Resolves BA-3696

### DIFF
--- a/src/ai/backend/manager/repositories/user/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/user/db_source/db_source.py
@@ -107,8 +107,7 @@ from ai.backend.manager.repositories.user.types import (
     RoleUserSearchScope,
 )
 from ai.backend.manager.repositories.user.updaters import UserUpdaterSpec
-from ai.backend.manager.services.user.actions.create_user import UserCreateSpec
-from ai.backend.manager.services.user.actions.modify_user import UserUpdateSpec
+from ai.backend.manager.services.user.types import UserCreateSpec, UserUpdateSpec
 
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/services/user/actions/modify_user.py
```diff
@@ -1,50 +1,67 @@
...
 
     @override
     @classmethod
     def operation_type(cls) -> ActionOperationType:
         return ActionOperationType.UPDATE
 
+    @override
+    def target_entity_id(self) -> str:
+        if self.user_uuid is None:
+            raise ValueError("user_uuid must be set for RBAC validation")
```
jopemachine, line 44: It seems that a BackendAIError should be raised.

HyeockJinKim: Let's handle this as a follow-up task.

PR #10056, opened 2026-03-13
feat(BA-5038): Apply RBAC validator for Resource Group actions

fregataa:
## Summary
- Integrated RBAC validators into scaling group (resource group) action processors
- Scope actions (create, search) now use `ScopeActionProcessor` with GLOBAL scope
- Single-entity actions (modify, purge) now use `SingleEntityActionProcessor` with entity-specific validation
- Internal/system actions remain on plain `ActionProcessor` without RBAC validators

## Test plan
- [x] Run `pants fmt fix lint` on changed files - all passed
- [ ] CI type checking and tests will validate the implementation

Resolves BA-5038


fregataa:
Closed this PR since it is outdate, I will revisit it in the future

PR #10059, opened 2026-03-13
feat(BA-5040): Apply RBAC validator for Image actions

fregataa:
## Summary
- Added RBAC base class (`ImageSingleEntityAction`) for Image service following established patterns
- Refactored `ForgetImageByIdAction` and `PurgeImageByIdAction` to extend `ImageSingleEntityAction` and implement required RBAC methods (`target_entity_id()`, `target_element()`, `entity_type()`)
- Wired `SingleEntityActionValidator` to Image processors using `SingleEntityActionProcessor` for entity-level permission checks
- Superadmin-only mutations (alias, dealias, modify, untag, scan, etc.) remain on plain `ActionProcessor` — access is enforced by `check_admin_only` at the API layer
- Internal/system actions remain on plain `ActionProcessor` without RBAC validators

## Test plan
- [x] `pants fmt` passes
- [x] `pants fix` passes
- [x] CI type checks pass
- [x] CI tests pass

Resolves BA-5040

### DIFF
--- a/src/ai/backend/manager/services/image/processors.py
+++ b/src/ai/backend/manager/services/image/processors.py
@@ -164,6 +164,8 @@ class ImageProcessors(AbstractProcessorPackage):
         self.purge_image_by_id = SingleEntityActionProcessor(
             service.purge_image_by_id, action_monitors, validators=[validators.rbac.single_entity]
         )
+        # Superadmin-only mutations — access is enforced by check_admin_only at the API layer,
+        # so per-entity RBAC validation is not required here.
         self.alias_image = ActionProcessor(service.alias_image, action_monitors)
         self.alias_image_by_id = ActionProcessor(service.alias_image_by_id, action_monitors)
         self.dealias_image = ActionProcessor(service.dealias_image, action_monitors)
--- a/tests/component/image/conftest.py
+++ b/tests/component/image/conftest.py
@@ -9,6 +9,12 @@ import sqlalchemy as sa
 from sqlalchemy.ext.asyncio.engine import AsyncEngine as SAEngine
 
 from ai.backend.common.container_registry import ContainerRegistryType
+from ai.backend.manager.actions.validators import ActionValidators
+from ai.backend.manager.actions.validators.rbac import RBACValidators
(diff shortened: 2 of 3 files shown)


src/ai/backend/manager/services/image/processors.py
```diff
@@ -92,8 +93,10 @@
...
 class ImageProcessors(AbstractProcessorPackage):
     forget_image: ActionProcessor[ForgetImageAction, ForgetImageActionResult]
-    forget_image_by_id: ActionProcessor[ForgetImageByIdAction, ForgetImageByIdActionResult]
-    purge_image_by_id: ActionProcessor[PurgeImageByIdAction, PurgeImageByIdActionResult]
+    forget_image_by_id: SingleEntityActionProcessor[
+        ForgetImageByIdAction, ForgetImageByIdActionResult
+    ]
+    purge_image_by_id: SingleEntityActionProcessor[PurgeImageByIdAction, PurgeImageByIdActionResult]
     alias_image: ActionProcessor[AliasImageAction, AliasImageActionResult]
     alias_image_by_id: ActionProcessor[AliasImageByIdAction, AliasImageByIdActionResult]
```
jopemachine, line 101: AliasImage (alias_image_by_id) is a mutation that can only be invoked by a superadmin—does it not require an explicit permission check?
Or is it intentionally omitted because check_admin_only already guarantees that?

(It’s also missing in places like untag_image_from_registry—is this intentional as well?)

fregataa: good point. I discarded the change

jopemachine:
Overall, it’s unclear whether the actions without permission checks were intentionally left unmodified because they are deprecated, or if they were simply overlooked.

Would it make sense to add deprecation annotations to all actions that have not been updated?

HyeockJinKim (approved):

HyeockJinKim:
It seems the admin service should be separated; please review this. Rather than distinguishing between them at the API level, please verify at the service layer whether the respective operations should be restricted to admins only. (For GET requests, check if the user has access permissions; for SEARCH, check based on the scope; for CREATE, UPDATE, DELETE, and PURGE, check based on permissions.) @fregataa

PR #10077, opened 2026-03-13
feat(BA-5104): Update app-proxy health_check_config on deployment revision swap

jopemachine:
Resolves BA-5104

## Summary

- Update app-proxy `health_check_config` to reflect the deploying revision's model-definition.yaml during PROVISIONING
- Previously, health checks used stale config from the previous revision (e.g., wrong path or `expected_status_code`), preventing rollback from triggering correctly
- Fetch deploying revision's model definition from its vfolder via `deploying_revision_id` (not `target_revision()` which returns current revision's spec)

## Changes

- `AppProxyClient.update_health_check()` — new PUT `/v2/endpoints/{id}/health-check` method
- `DeploymentExecutor.update_endpoint_health_check()` — reads deploying revision's model-definition from DB/storage, extracts health check config, updates app-proxy
- `DeployingProvisioningHandler` — calls `update_endpoint_health_check()` at the start of each PROVISIONING cycle
- `coordinator.py` — passes `deployment_executor` to Provisioning/Progressing handlers

## Technical Notes

- `health_check_config` is stored per-endpoint (not per-revision) in app-proxy's `endpoints` table
- `DeploymentInfo.target_revision()` returns **current** revision's spec — must fetch deploying revision from DB directly
- Structural limitation: mixed-revision health checks during coexistence are not possible with per-endpoint config

## Test plan

- [ ] Rollback E2E: broken model (`health_check.path: /health`) triggers health check failure → rollback
- [ ] Rolling update E2E: normal model passes health check → revision swap completes
- [ ] Verify health_check_config is correctly updated in app-proxy DB during PROVISIONING

Jira: https://lablup.atlassian.net/browse/BA-5104

### DIFF
--- a/src/ai/backend/manager/sokovan/deployment/handlers/deploying.py
+++ b/src/ai/backend/manager/sokovan/deployment/handlers/deploying.py
@@ -130,24 +130,6 @@ class DeployingProvisioningHandler(DeploymentHandler):
         )
 
     @override
-    async def prepare(self, deployments: Sequence[DeploymentWithHistory]) -> None:
-        """Update health check config in app-proxy for newly entering deployments.
-
-        Called once per deployment when it first enters PROVISIONING, so that
-        the new revision's health check settings are used from the start.
-        """
-        for deployment in deployments:
-            try:
-                await self._deployment_executor.update_endpoint_health_check(
-                    deployment.deployment_info,
-                )
-            except Exception:
-                log.exception(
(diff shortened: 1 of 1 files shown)


jopemachine:
The app proxy DB will soon be moved to the manager, and the app proxy health check synchronization logic is scheduled to be removed.
So, I will skip this task.

PR #10096, opened 2026-03-16
fix(BA-5120): Fix `container_id` UUID parsing error in GET /session/{session_name}

seedspirit:
## Summary
- Fixed `ValueError: badly formed hexadecimal UUID string` in `GET /session/{session_name}` API
- Changed `LegacySessionInfo.container_id` type from `UUID` to `ContainerId` (str-based) to correctly handle Docker container IDs (64-char hex SHA-256)
- Updated `service.py` to use `ContainerId()` instead of `uuid.UUID()` when building session info

## Test plan
- [ ] Existing `test_success` passes with Docker-format (64-char hex) container ID
- [ ] CI type checks pass for affected files

Resolves BA-5120

### DIFF
--- a/changes/10096.fix.md
+++ b/changes/10096.fix.md
@@ -1,0 +1,1 @@
+Fix `ValueError: badly formed hexadecimal UUID string` in `GET /session/{session_name}` caused by parsing Docker container IDs as UUIDs.


src/ai/backend/manager/services/session/service.py
```diff
@@ -1089,9 +1090,9 @@ async def get_session_info(self, action: GetSessionInfoAction) -> GetSessionInfo
             architecture=sess.main_kernel.architecture or "",
             registry=sess.main_kernel.registry,
             tag=sess.tag,
-            container_id=uuid.UUID(sess.main_kernel.container_id)
+            container_id=ContainerId(sess.main_kernel.container_id)
             if sess.main_kernel.container_id
-            else uuid.uuid4(),
+            else ContainerId(""),
```
HyeockJinKim, line 1095: It seems safer to treat it as an optional value.

HyeockJinKim (approved):
I'll go ahead and apply it for now. I'll make the necessary adjustments later.

PR #10252, opened 2026-03-16
fix(BA-5274): Make `LegacySessionInfo.container_id` optional

seedspirit:
## Summary
- Changed the type of `LegacySessionInfo.container_id` from `ContainerId` to `ContainerId | None`
- Return `None` instead of `ContainerId("")` when `container_id` is missing
- Added tests for the `container_id=None` case

Follow-up from PR #10096 review comment by @HyeockJinKim: "It seems safer to treat it as an optional value."

## Test plan
- [x] `pants check` passes for affected files
- [x] `pants test` passes for `test_session_service.py`
- [x] Existing `test_success` asserts `container_id` is not None and matches expected value
- [x] New `test_success_with_no_container_id` asserts `container_id` is None when kernel has no container

### DIFF
--- a/tests/unit/manager/services/session/test_session_service.py
+++ b/tests/unit/manager/services/session/test_session_service.py
@@ -758,9 +758,7 @@ class TestGetSessionInfo:
         sample_access_key: AccessKey,
     ) -> None:
         """Test successfully getting session info"""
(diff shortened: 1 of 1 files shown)


fregataa:
Please check the ci failure

HyeockJinKim (approved):

PR #10271, opened 2026-03-17
feat(BA-5278): Use deploying-revision image for new route session creation

jopemachine:
## Summary

During rolling updates, new route sessions were incorrectly using the target revision image from deployment info rather than the image of the specific revision assigned to each route. This fix ensures that when provisioning a new session for a route, the image is resolved from the route's own revision record.

## Changes

- Add `revision_id: uuid.UUID | None = None` field to `RouteData` dataclass
- Populate `revision_id` from `RoutingRow.revision` in both `get_routes_by_endpoint` and `get_routes_by_statuses` in `db_source.py`
- Add `revision_id` parameter to `fetch_deployment_context()` in `db_source.py`; when provided, queries `DeploymentRevisionRow` with `selectinload(image_row)` and builds an `ImageIdentifier` from it instead of using `deployment_info.target_revision()`
- Add `revision_id` parameter to `fetch_deployment_context()` in `repository.py` and pass it through to `db_source`
- Pass `revision_id=route.revision_id` when calling `fetch_deployment_context()` in `_provision_route()` inside the route executor

## Testing

- Verified pre-commit checks (ruff format, ruff check, visibility) pass with no errors
- Rolling update integration test can be used to confirm new routes use the correct revision image

## Related Issues

Fixes BA-5278

### DIFF
--- a/src/ai/backend/manager/data/deployment/types.py
+++ b/src/ai/backend/manager/data/deployment/types.py
@@ -374,7 +374,7 @@ class DeploymentInfo:
     state: DeploymentState
     replica_spec: ReplicaSpec
     network: DeploymentNetworkSpec
-    model_revisions: list[ModelRevisionSpec] = field(default_factory=list)
+    model_revisions: list[ModelRevisionSpec]
     current_revision_id: UUID | None = None
     policy: DeploymentPolicyData | None = None
     deploying_revision_id: UUID | None = None


src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -1843,15 +1849,34 @@ async def fetch_deployment_context(
...
+            # deploying → current revision from deployment_info.
+            target_revision_id = (
+                revision_id
+                or deployment_info.deploying_revision_id
+                or deployment_info.current_revision_id
+            )
+            if target_revision_id is None:
+                raise DeploymentHasNoTargetRevision(
+                    "Deployment has no revision for image resolution"
+                )
```
HyeockJinKim: I don't think the revision ID should be optional.

src/ai/backend/manager/repositories/deployment/repository.py
```diff
@@ -707,16 +707,20 @@ async def delete_routes_by_route_ids(
     async def fetch_deployment_context(
         self,
         deployment_info: DeploymentInfo,
+        revision_id: UUID | None = None,
```
HyeockJinKim: Do you need deployment information?

src/ai/backend/manager/data/deployment/types.py
```diff
@@ -373,16 +374,18 @@ class DeploymentInfo:
     state: DeploymentState
     replica_spec: ReplicaSpec
     network: DeploymentNetworkSpec
-    model_revisions: list[ModelRevisionSpec]
+    model_revisions: list[ModelRevisionSpec] = field(default_factory=list)
```
HyeockJinKim: If possible, it seems best not to include default values for internal fields.

jopemachine: I will remove it

src/ai/backend/manager/models/endpoint/row.py
```diff
@@ -779,14 +780,16 @@ def to_deployment_info(self) -> DeploymentInfo:
...
+        # Build model_revisions list from loaded revision rows
+        if "revisions" in instance_state(self).dict and self.revisions:
+            model_revisions: list[ModelRevisionSpec] = []
+            for rev_row in self.revisions:
+                if rev_row.image_row is None:
+                    continue
+                if rev_row.id == self.current_revision or rev_row.id == self.deploying_revision:
+                    model_revisions.append(self._build_revision_spec(rev_row))
+            if model_revisions:
+                info = self._to_deployment_info_with_revisions(model_revisions)
```
HyeockJinKim, line 792: The code for checking the string key doesn't look great, but since it seems to be an existing issue, I'll leave it for now.

PR #10292, opened 2026-03-17
fix(BA-5287): Expose db and config_provider to webapp plugins in webapp_plugin_ctx

fregataa:
## Summary
- Inject `db` and `config_provider` into webapp plugin root app to restore access after DI refactoring
- Webapp plugins (OpenID, TOTP) can now access database and config via `request.app["_root_app"]["_db"]` and `request.app["_root_app"]["_config_provider"]`
- Added comprehensive unit test to verify injection works correctly

## Context
After the DI refactoring removed RootContext, webapp plugins lost access to `db` and `config_provider`. This fix restores that access by injecting these resources into the root app before plugins are initialized.

## Test plan
- [x] Unit test `test_webapp_plugin_ctx_sets_db_and_config_provider` verifies injection
- [x] Quality checks (fmt, fix, lint) passed
- [ ] Integration test with actual TOTP/OpenID plugins (manual verification)

Resolves BA-5287


src/ai/backend/manager/server.py
```diff
@@ -87,6 +87,8 @@ async def webapp_plugin_ctx(
         blocklist=r.bootstrap.config_provider.config.manager.disabled_plugins,
     )
     cors_options = r.system.cors_options
+    root_app["_db"] = r.infrastructure.db
```
jopemachine, line 90: This change appears to be a temporary workaround for compatibility, and it should be fundamentally addressed through a redesign of the plugin interface structure.

Given that, I think it would be better to explicitly discourage this pattern by marking it in the code with a TODO or a comment.

PR #10294, opened 2026-03-18
fix(BA-5287): Expose db and config_provider to webapp plugins in webapp_plugin_ctx (#10292)

lablup-octodog:
This is an auto-generated backport PR of #10292 to the 26.2 release.


HyeockJinKim:
This will only be an issue in version 26.3. @fregataa

PR #10305, opened 2026-03-18
fix(BA-5292): add otp alias to AuthorizeRequest.stoken for TOTP compatibility

fregataa:
## Summary
- Add "otp" to `AliasChoices` in `AuthorizeRequest.stoken` and `AuthorizeInput.stoken` validation_alias
- Add "otp" key to `AuthorizeAction.hook_params` for auth hook propagation
- Add test cases verifying "otp" alias works alongside existing "stoken"/"sToken" aliases

This fixes the core DTO parsing issue where the webui sends TOTP codes as "otp" but the field only accepted "stoken"/"sToken", causing hook_params["stoken"] to be empty and TOTP auth to fail even with correct codes.

## Test plan
- [x] Unit tests for "otp" alias in both DTO versions pass
- [x] All quality checks (fmt, fix, lint) pass
- [ ] CI tests pass

Resolves BA-5292

### DIFF
--- a/tests/unit/manager/services/auth/test_authorize.py
+++ b/tests/unit/manager/services/auth/test_authorize.py
@@ -92,6 +92,7 @@ async def test_authorize_success(
         password="correct_password",
         request=MagicMock(),
         stoken=None,
+        otp=None,
     )
 
     result = await auth_service.authorize(action)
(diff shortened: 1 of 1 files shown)


HyeockJinKim (approved):

src/ai/backend/manager/services/auth/actions/authorize.py
```diff
@@ -30,13 +31,15 @@ def operation_type(cls) -> ActionOperationType:
 
     @property
     def hook_params(self) -> dict[str, str]:
+        otp_value = self.otp or self.stoken or ""
```
jopemachine, line 34: Is otp default value empty string?

PR #10395, opened 2026-03-22
fix: Make MetricPolicy metric recording best-effort with gauge consistency

### LINKED ISSUE #10394: fix: MetricPolicy should not propagate prometheus mmap errors into retry loop
## Problem

When macOS (or any OS) wakes from sleep, prometheus multiprocess mmap'd `.db` files can become corrupted. This causes `ValueError: mmap slice assignment is wrong size` when any prometheus metric (Counter, Gauge, Histogram) is updated.

The `MetricPolicy` in the resilience pipeline sits **inside** the `RetryPolicy` loop. When `MetricPolicy.execute()` calls `observer.observe_layer_operation_triggered()` → prometheus `.inc()` → mmap write fails, the `ValueError` propagates out. Since it's not a `BackendAIError`, `RetryPolicy` treats it as retryable and retries — which triggers another metric write — which hits the same mmap error — creating a death spiral.

Each retry attempt also opens a new DB transaction, causing concurrent transaction count to exceed the threshold (40), effectively deadlocking the manager.

### Logs

```
2026-03-22 12:52:08.464 WARNING ai.backend.manager.models.utils [64694] The number of concurrent generic transactions (40) looks too high (warning threshold: 40).
2026-03-22 12:52:08.465 WARNING ai.backend.manager.models.utils [64694] The number of concurrent generic transactions (41) looks too high (warning threshold: 40).
2026-03-22 12:52:08.466 DEBUG ai.backend.common.resilience.policies.retry [64694] retryable exception encountered: mmap slice assignment is wrong size, attempt 5/10
2026-03-22 12:52:08.466 DEBUG ai.backend.common.resilience.policies.retry [64694] retryable exception encountered: mmap slice assignment is wrong size, attempt 6/10
```

## Root Cause

- `MetricPolicy` records prometheus metrics synchronously as part of the resilience policy chain
- Prometheus multiprocess mode uses mmap'd files that can become corrupted after OS sleep/wake
- The mmap error is not in the non-retryable exceptions list, so `RetryPolicy` retries it indefinitely
- Each retry re-enters `MetricPolicy`, causing the same error again

## Fix

Make all metric recording in `MetricPolicy` best-effort by wrapping calls in a `_safe_observe()` helper that catches and logs (at debug level) any exceptions. Metrics are observability — they should never interfere with the actual operation or trigger retries.

## Impact

- **Without fix**: Manager becomes unresponsive after OS sleep/wake, requires restart
- **With fix**: Metrics are temporarily lost after sleep/wake, but operations continue normally and metrics recover once mmap files are re-established

rapsealk:
## Summary
- Wrap all prometheus metric recording in `MetricPolicy` with a `_safe_observe()` helper that catches and suppresses exceptions
- Prevents mmap error death spiral after OS sleep/wake where corrupted prometheus `.db` files cause `ValueError` to propagate into the retry loop, exhausting the DB connection pool
- **Prevent negative in-flight gauge**: When `observe_layer_operation_triggered()` fails (no `.inc()`), skip `observe_layer_operation()` to avoid `.dec()` driving the gauge negative
- Log the first metric error at `warning` level with `exc_info` for diagnostics; subsequent errors are silently suppressed to avoid log spam
- Add comprehensive tests covering all metric error suppression paths (trigger, completion, failure path, combined failures)

Fixes #10394

## Test plan
- [x] Existing `test_metric_policy.py` tests pass (12 tests)
- [x] `test_safe_observe_suppresses_trigger_metric_errors` — trigger fails, completion skipped, operation succeeds
- [x] `test_safe_observe_suppresses_completion_metric_errors` — trigger succeeds, completion fails, operation succeeds
- [x] `test_safe_observe_suppresses_failure_path_metric_errors` — operation fails + completion metric fails, original error propagates
- [x] `test_skips_completion_metric_when_trigger_fails` — verifies gauge consistency (no negative in-flight count)
- [x] `test_metric_error_logged_once_as_warning` — first error logged at warning, subsequent suppressed
- [ ] Manual: run manager, sleep/wake Mac, verify manager stays responsive


rapsealk:
## Follow-up: Other unprotected metric callers

This PR only protects `MetricPolicy` (the resilience pipeline path). However, the prometheus mmap corruption affects **all** metric recording in the process. The following callers are still unprotected and will raise on corrupted mmap files:

| Caller | File | Severity |
|--------|------|----------|
| HTTP middleware `finally` block | `src/ai/backend/common/metrics/http.py` | **Critical** — mmap error replaces the original HTTP exception, turning every 4xx/5xx into an opaque `ValueError` |
| GraphQL metric extension | `src/ai/backend/manager/api/gql/extensions/metric.py` | High — all GraphQL resolvers fail |
| Event dispatcher | `src/ai/backend/common/events/dispatcher.py` | High — kills event processing tasks |
| Background task metric hook | `src/ai/backend/common/bgtask/hooks/metric_hook.py` | Medium — `finally` block masks task exceptions |
| Prometheus action monitor | `src/ai/backend/manager/actions/monitors/prometheus.py` | Medium |
| Event propagator hub | `src/ai/backend/common/events/hub/hub.py` | Low |

**Recommendation:** Extract `_safe_observe` into a shared utility in `common/metrics/` and apply it to all callers above. The HTTP middleware `finally` block is the most urgent since it can mask real errors even without the sleep/wake scenario (any prometheus corruption would trigger it).

HyeockJinKim:
If an mmap issue occurs, Prometheus data collection itself will likely fail—so wouldn’t that eventually cause errors when collecting metrics from other sources as well? Fundamentally, once an error occurs at the collection point, we should stop collecting Prometheus metrics and instead synchronize the watcher based on the Health API. @rapsealk

src/ai/backend/common/resilience/policies/metrics.py
```diff
@@ -56,6 +60,26 @@ def __init__(self, args: MetricArgs) -> None:
...
+    def _safe_observe(self, fn: Callable[[], None]) -> bool:
+        """Call a metric-recording function, suppressing any errors.
+
+        Returns True if the function executed successfully, False otherwise.
+        """
+        try:
+            fn()
+            return True
+        except Exception:
+            if not self._metric_error_logged:
```
seedspirit, line 74: It seems metric errors could occur for various reasons—wouldn’t suppressing them make it harder to track them? Is the benefit of suppressing the logs for the same metric errors greater than the trade-off of having to accept that?

rapsealk: Good point. To clarify the trade-off:

1. **The first error IS logged** — at `warning` level with `exc_info=True` (full traceback), so operators can diagnose the root cause.
2. **Only subsequent identical errors are suppressed** — in the mmap corruption scenario, every single metric call on every request will fail with the same `ValueError` until the process restarts. Logging every one would flood the logs (thousands of identical tracebacks) and mask other real warnings.
3. **The `_metric_error_logged` flag is per-`MetricPolicy` instance**, so if different metric domains/layers fail, each one still gets its first warning logged.

So the actual trade-off is: we lose visibility into whether the error is *still happening* after the first occurrence. But since the root cause (corrupted mmap files) is persistent until restart, the first warning is sufficient for diagnosis. A metric recovery mechanism (as @HyeockJinKim suggested) would be the proper long-term solution.

rapsealk:
@HyeockJinKim Agree — once mmap corruption happens, all prometheus metric recording in the process is effectively broken. A circuit-breaker approach (detect the first error → disable all metric collection → rely on Health API) would be cleaner than wrapping every call site individually.

This PR is a focused fix for the immediate crash loop in `MetricPolicy` where the mmap error propagates into the retry policy and exhausts the DB connection pool. For the broader "stop collecting after corruption + Health API fallback" approach, I think that deserves its own issue/PR since it would touch the `LayerMetricObserver` singleton itself (or a new global metric health flag) plus all the callers listed in the follow-up comment.

Should I create a separate issue to track the circuit-breaker approach?

rapsealk:
Superseded by #10443.

PR #10458, opened 2026-03-24
feat(BA-5389): resolve model definition at revision creation time instead of PROVISIONING

### LINKED ISSUE #10457: Merge deployment revision DB model_definition into generated model definition
## Background

The `deployment_revisions` table has a `model_definition` JSONB column (`DeploymentRevisionRow.model_definition`), but it is completely ignored during inference session creation. The current `ModelDefinitionGeneratorRegistry.generate_model_definition()` only merges overrides from the storage file (via `model_definition_path`), never from the DB record.

## Problem

- `ModelRevisionSpec` does not include a `model_definition` field — the DB value is never propagated to the generator.
- `DeploymentRevisionRow.to_data()` does not map `model_definition` to the data object.
- `ModelDefinitionGeneratorRegistry` does not read or merge the revision DB `model_definition` value into the generated result.

## Tasks

1. Add `model_definition: dict[str, Any] | None` field to `ModelRevisionSpec` (`data/deployment/types.py`)
1. Map `model_definition` in `DeploymentRevisionRow.to_data()` and any repository methods that build `ModelRevisionSpec`
1. Update `ModelDefinitionGeneratorRegistry.generate_model_definition()` to `deep_merge` the DB `model_definition` value into the generated definition (respecting the existing merge order with storage file override)
1. Add tests to verify the merge behavior (DB value + generated + storage override)

## Affected Files

- `src/ai/backend/manager/data/deployment/types.py` — `ModelRevisionSpec`
- `src/ai/backend/manager/models/deployment_revision/row.py` — `to_data()`
- `src/ai/backend/manager/sokovan/deployment/definition_generator/registry.py` — `generate_model_definition()`
- `src/ai/backend/manager/repositories/deployment/` — repository methods building ModelRevisionSpec

JIRA Issue: BA-5389

jopemachine:
resolves #10457 (BA-5389)

## Summary

Move model definition resolution from every PROVISIONING cycle to **revision creation time**. The DB now stores the fully-resolved definition; PROVISIONING reads it directly without vfolder access.

### What changed

- `DeploymentService._resolve_model_definition()` resolves the final definition before creating the revision row
- Executor no longer calls `generate_model_definition()` — uses stored `target_revision.model_definition` directly
- New `ModelDefinitionContext` replaces `ModelRevisionSpec` as the generator input (only the fields generators actually need)
- `Mapping[str, Any]` → `ModelDefinition` strong typing across revision spec and creator
- GQL `model_definition` field annotated with `NEXT_RELEASE_VERSION`

### Merge priority (applied at revision creation)

1. Generator-produced definition (lowest)
2. Vfolder file override (non-CUSTOM only, requires `enable_model_definition_override` flag)
3. User-provided `model_definition` override (highest)

## Not included

The legacy session creation path (`_get_health_check_info` in `registry.py`) still reads model definition from vfolder at runtime. This serves the old endpoint system and should be migrated separately.

## Merge dependency

> Depends on [#10459](https://github.com/lablup/backend.ai/pull/10459) (BA-5390) which adds the write-side: accepting `model_definition` as input in `add_model_revision` API. Merge #10459 first.

- [x] Test case(s) to:
- [ ] Documentation

----
📚 Documentation preview 📚: https://sorna--10458.org.readthedocs.build/en/10458/

----
📚 Documentation preview 📚: https://sorna-ko--10458.org.readthedocs.build/ko/10458/

### DIFF
--- a/src/ai/backend/manager/models/endpoint/row.py
+++ b/src/ai/backend/manager/models/endpoint/row.py
@@ -36,7 +36,7 @@ from sqlalchemy.orm import (
     selectinload,
 )
 
-from ai.backend.common.config import ModelDefinition, model_definition_iv
+from ai.backend.common.config import model_definition_iv
 from ai.backend.common.types import (
     AccessKey,
     AutoScalingMetricComparator,
@@ -67,14 +67,10 @@ from ai.backend.manager.data.deployment.types import (
     DeploymentMetadata,
     DeploymentNetworkSpec,
     DeploymentState,
-    ExecutionSpec,
     ModelDeploymentAutoScalingRuleData,
     ModelRevisionSpec,
-    MountMetadata,
     ReplicaSpec,
-    ResourceSpec,
 )
-from ai.backend.manager.data.image.types import ImageIdentifier
 from ai.backend.manager.data.model_serving.types import (
     EndpointAutoScalingRuleData,
     EndpointData,
@@ -105,7 +101,6 @@ from ai.backend.manager.types import MountOptionModel, UserScope
 
 if TYPE_CHECKING:
     from ai.backend.manager.data.deployment.creator import DeploymentCreator
-    from ai.backend.manager.models.deployment_revision import DeploymentRevisionRow
 
 __all__ = (
     "EndpointAutoScalingRuleRow",
@@ -779,47 +774,11 @@ class EndpointRow(Base):  # type: ignore[misc]
             if rev_row.image_row is None:
                 continue
             if rev_row.id == self.current_revision or rev_row.id == self.deploying_revision:
-                model_revisions.append(self._build_revision_spec(rev_row))
+                model_revisions.append(rev_row.to_model_revision_spec())
 
         info = self._to_deployment_info_with_revisions(model_revisions)
         info.policy = policy_data
         return info
-
-    def _build_revision_spec(
-        self,
-        revision: DeploymentRevisionRow,
-    ) -> ModelRevisionSpec:
-        """Build a ModelRevisionSpec from a revision row."""
-        image_identifier = ImageIdentifier(
-            canonical=revision.image_row.name,
-            architecture=revision.image_row.architecture,
-        )
-        return ModelRevisionSpec(
-            revision_id=revision.id,
-            image_identifier=image_identifier,
-            resource_spec=ResourceSpec(
-                cluster_mode=ClusterMode(revision.cluster_mode),
-                cluster_size=revision.cluster_size,
-                resource_slots=revision.resource_slots,
-                resource_opts=revision.resource_opts,
-            ),
-            mounts=MountMetadata(
-                model_vfolder_id=revision.model or uuid.UUID(int=0),
-                model_definition_path=revision.model_definition_path,
-                model_mount_destination=revision.model_mount_destination,
-                extra_mounts=revision.extra_mounts or [],
-            ),
-            execution=ExecutionSpec(
-                startup_command=revision.startup_command,
-                bootstrap_script=revision.bootstrap_script,
-                environ=revision.environ,
-                runtime_variant=revision.runtime_variant,
-                callback_url=yarl.URL(revision.callback_url) if revision.callback_url else None,
-            ),
-            model_definition=ModelDefinition.model_validate(revision.model_definition)
-            if revision.model_definition
-            else None,
-        )
 
     def _to_deployment_info_with_revisions(
         self,


tests/unit/manager/sokovan/deployment/definition_generator/test_registry.py
```diff
@@ -207,19 +275,91 @@ async def test_override_merges_with_generated(
...
+    ) -> None:
+        """DB model_definition is merged on top of generated definition."""
+        result = await definition_generator_registry.generate_model_definition(
+            create_revision(variant, model_definition=db_model_definition)
+        )
+
+        model = result.models[0]
+        assert model.service is not None
+        assert model.service.start_command == "db-overridden-command"
+
```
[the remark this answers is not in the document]
jopemachine: This is intentional — the test's purpose is to verify that DB model_definition merging works correctly regardless of the variant. The DB override dict uses VLLM-like fields as a representative example, but the core behavior being tested (deep merge into the generated definition) is variant-agnostic. Variant-specific field validation is already covered by each generator's own tests.

changes/10458.feature.md
```diff
@@ -0,0 +1 @@
+Merge `deployment_revision`'s `model_definition` into generated model definition
```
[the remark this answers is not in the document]
jopemachine: The changelog entry refers to the `model_definition` field on a deployment revision entity, not the database table name. Using the singular form is correct here as it describes the conceptual feature, not the table.

src/ai/backend/common/config.py
```diff
@@ -20,6 +20,7 @@
 from .etcd import AsyncEtcd, ConfigScopes
 from .exception import ConfigurationError
 from .types import RedisHelperConfig
+from .utils import deep_merge
```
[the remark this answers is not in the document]
jopemachine: `deep_merge` is already defined in `ai.backend.common.utils` and reused across the codebase. Duplicating it locally in `config.py` would violate DRY. The import cost of `common.utils` is negligible in our server context since it's already imported transitively by many modules at startup.

src/ai/backend/manager/data/deployment/types.py
```diff
@@ -327,6 +327,10 @@ class ModelRevisionSpec(ConfiguredModel):
     resource_spec: ResourceSpec
     mounts: MountMetadata
     execution: ExecutionSpec
+    # Kept as Mapping (not ModelDefinition) because this holds a partial override
+    # from the DB JSONB column. It may lack required fields and is only valid
+    # after being deep-merged into a fully generated ModelDefinition.
+    model_definition: Mapping[str, Any] | None = None
```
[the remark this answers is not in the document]
jopemachine: Using `Mapping[str, Any] | None` here is intentional — the DB stores a partial override dict (not a fully validated `ModelDefinition`), so it should remain a raw mapping. The PR description will be updated to reflect this.

src/ai/backend/common/dto/manager/v2/deployment/request.py
```diff
@@ -187,6 +187,9 @@ class CreateRevisionInputDTO(BaseRequestModel):
     image: ImageInput = Field(description="Container image")
     model_runtime_config: ModelRuntimeConfigInput = Field(description="Runtime configuration")
     model_mount_config: ModelMountConfigInput = Field(description="Model mount configuration")
+    model_definition: ModelDefinition = Field(
+        description="Model definition to override the generated definition"
```
seedspirit: People who aren't familiar with the codebase might not know exactly what "generated definition" means. I suggest conveying the idea that it overrides the default values generated by the server.

src/ai/backend/manager/sokovan/deployment/definition_generator/registry.py
```diff
@@ -92,7 +96,17 @@ async def generate_model_definition(self, model_revision: ModelRevisionSpec) ->
...
         generated_definition = await generator.generate_model_definition(model_revision)
 
-        # Check if override is enabled and path exists
+        # Merge revision DB model_definition if present
+        if model_revision.model_definition:
+            try:
+                generated_definition = generated_definition.merge(model_revision.model_definition)
+            except Exception:
+                log.error(
+                    "Failed to merge revision DB model_definition, using base definition",
```
seedspirit: It seems an additional explanation of "base definition" is needed.

docs/manager/graphql-reference/supergraph.graphql
```diff
@@ -2807,6 +2807,7 @@ input CreateRevisionInput
   image: ImageInput!
   modelRuntimeConfig: ModelRuntimeConfigInput!
   modelMountConfig: ModelMountConfigInput!
+  modelDefinition: ModelDefinitionInput!
```
fregataa: Need a version annotation

docs/manager/graphql-reference/v2-schema.graphql
```diff
@@ -1533,6 +1533,7 @@ input CreateRevisionInput {
   image: ImageInput!
   modelRuntimeConfig: ModelRuntimeConfigInput!
   modelMountConfig: ModelMountConfigInput!
+  modelDefinition: ModelDefinitionInput!
```
fregataa: Same here

jopemachine:
> 1. Version annotations are needed
> 2. Does the new model definition value override a stored value or is merged with it? Could you describe how the model definition of model deployment and revision changes after this work?

The system no longer references the vfolder's model definition at the PROVISIONING stage.
The model_definition is now merged with the vfolder’s model definition at revision creation time and stored in the database, rather than being resolved during the PROVISIONING stage.

tests/unit/manager/sokovan/deployment/executor/test_deployment_executor.py
```diff
@@ -90,17 +90,18 @@ async def test_deployment_without_revision_is_skipped(
...
+            with DeploymentRecorderContext.scope("test", entity_ids=entity_ids):
+                # Act
+                result = await deployment_executor.check_pending_deployments([
+                    pending_deployment_no_revision
+                ])
 
         # Assert - Deployment skipped due to missing revision
         assert len(result.successes) == 0
         assert len(result.failures) == 0
-        assert len(result.skipped) == 1
```
fregataa, line 103: Why is it removed?

tests/unit/manager/sokovan/deployment/definition_generator/test_registry.py
```diff
@@ -207,19 +271,256 @@ async def test_override_merges_with_generated(
...
+            )
+        )
+
+        model = result.models[0]
+        assert model.service is not None
+        # from vfolder (overrode generator)
+        assert model.service.start_command == "vfolder-command"
+        assert model.service.health_check is not None
+        # from vfolder (preserved through user override)
+        assert model.service.health_check.initial_delay == 300.0
```
fregataa: vllm_vfolder_definition and vllm_user_override share the same start_command and initial_delay fixtures, which leads to weak assertion.

PR #10486, opened 2026-03-25
feat(BA-5400): add custom filename support to archive download ZIP response

seedspirit:
## Summary
- Add optional `filename` field to the archive download request body, allowing frontends to specify a custom ZIP filename
- Propagate the filename through the JWT token from Manager to Storage Proxy, which sets it in the `Content-Disposition` response header
- Validate filename against path traversal characters (`/`, `\`, `..`, null bytes) and empty/whitespace strings

## Test plan
- [ ] `pants test` passes for affected packages (common, manager, storage)
- [ ] DTO validation tests: accepts valid filenames (including Unicode), rejects path traversal attempts
- [ ] JWT roundtrip test: filename survives token encode/decode
- [ ] Content-Disposition header tests: custom filename used when provided, "archive.zip" default when omitted

Resolves BA-5400

### DIFF
--- a/tests/unit/storage/api/test_file_stream.py
+++ b/tests/unit/storage/api/test_file_stream.py
@@ -10,6 +10,7 @@ This test suite covers:
 from __future__ import annotations
 
 import urllib.parse
+from collections.abc import Callable
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
 from typing import Any, cast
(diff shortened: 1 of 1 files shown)


src/ai/backend/common/dto/storage/request.py
```diff
@@ -561,11 +566,31 @@ class ArchiveDownloadTokenData(BaseModel):
         min_length=1,
         description="List of relative file paths within the virtual folder to include in the archive.",
     )
+    filename: str | None = Field(
+        default=None,
+        max_length=255,
```
seedspirit, line 571: This is a placeholder value; we plan to set the actual value after verifying any front-end constraints.

tests/unit/storage/api/test_file_stream.py
```diff
@@ -338,15 +338,93 @@ async def test_handler_uses_filename_from_reader(
...
+        """
+        Test that handler correctly encodes Unicode filename in Content-Disposition.
+
+        Scenario:
+        1. Create JWT token with filename="데이터-export.zip"
+        2. Call download_archive
+        3. Verify Content-Disposition contains RFC-5987 encoded filename
+        """
+        unicode_filename = "데이터-export.zip"
+        token_data_with_filename = ArchiveDownloadTokenData(
```
jopemachine: Why not extracting this as fixture?

src/ai/backend/storage/api/client.py
```diff
@@ -509,7 +509,8 @@ async def download_archive(
             reader = ZipArchiveStreamReader(vfolder_root)
             reader.add_entries(sanitized)
 
-            headers = build_attachment_headers(reader.filename(), reader.content_type())
+            filename = token_data.filename if token_data.filename is not None else reader.filename()
+            headers = build_attachment_headers(filename, reader.content_type())
```
fregataa, line 513: Is there any test code for this change?

seedspirit: I will add test code

HyeockJinKim (approved):

PR #10497, opened 2026-03-25
feat(BA-5401): Ensure every deployment has a default rolling deployment policy

### LINKED ISSUE #10487: Add migration to assign default rolling deployment policy to existing deployments
## Background

Currently, existing deployments may not have an associated deployment policy. This means some deployments lack a defined update strategy, which can lead to unexpected behavior during updates or scaling operations.

## Goal

Write a database migration that assigns a default deployment policy with a "rolling" update strategy to all existing deployments that do not currently have one.

## Acceptance Criteria

- All existing deployments without a deployment policy are assigned a default rolling deployment policy after migration
- New deployments continue to work as expected
- Migration is reversible (downgrade path provided)

JIRA Issue: BA-5401

jopemachine:
Resolves #10487 (BA-5401)

## Summary

- Add FK constraint (`fk_deployment_policies_endpoint`) from `deployment_policies.endpoint` → `endpoints.id` with `ON DELETE CASCADE`
- Ensure every endpoint always gets a default rolling deployment policy on creation (both v1 and v2 code paths)
- Add alembic migration to backfill missing default policies and clean up orphaned ones before adding the FK

## Test plan

- [x] `check-alembic-migrations` CI passed
- [x] `lint-and-typecheck` CI passed
- [ ] Verify endpoint creation always produces exactly one deployment policy row

### DIFF
--- a/src/ai/backend/manager/models/alembic/versions/930e9f2dd502_assign_default_rolling_deployment_policy.py
+++ b/src/ai/backend/manager/models/alembic/versions/930e9f2dd502_assign_default_rolling_deployment_policy.py
@@ -26,7 +26,7 @@ def upgrade() -> None:
             uuid_generate_v4(),
             e.id,
             'ROLLING',
-            '{"max_surge": 1, "max_unavailable": 0}'::jsonb,
+            '{"max_surge": {"percent": 0.5}, "max_unavailable": {"percent": 0.0}}'::jsonb,
             NOW(),
             NOW()
         FROM endpoints e
--- a/src/ai/backend/manager/repositories/deployment/creators/policy.py
+++ b/src/ai/backend/manager/repositories/deployment/creators/policy.py
@@ -33,7 +33,7 @@ class DeploymentPolicyCreatorSpec(CreatorSpec[DeploymentPolicyRow]):
         return cls(
             endpoint_id=endpoint_id,
             strategy=DeploymentStrategy.ROLLING,
-            strategy_spec=RollingUpdateSpec(max_surge=1, max_unavailable=0),
+            strategy_spec=RollingUpdateSpec(),
         )
 
     @override


jopemachine:
This PR should be merged after updating the default values following https://github.com/lablup/backend.ai/pull/10563.

seedspirit (approved):

src/ai/backend/manager/models/alembic/versions/930e9f2dd502_assign_default_rolling_deployment_policy.py
```diff
@@ -0,0 +1,61 @@
...
+Revises: 17b679c98b50, 6f8a4c0d2e3g, 7h4m7ygnaao1
+Create Date: 2026-03-25
+
+"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "930e9f2dd502"
+down_revision = ("17b679c98b50", "6f8a4c0d2e3g", "7h4m7ygnaao1")
```
fregataa: Are they all needed for the downgrade? Let's keep it as linear as possible

jopemachine: down_revision updated

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -253,22 +253,25 @@ async def create_endpoint(
...
 
-            # Create deployment policy if provided
+            # Create deployment policy (use provided config or default rolling policy)
             if policy_config is not None:
-                policy_spec = DeploymentPolicyCreatorSpec(
+                policy_creator_spec = DeploymentPolicyCreatorSpec(
                     endpoint_id=endpoint.id,
                     strategy=policy_config.strategy,
                     strategy_spec=policy_config.strategy_spec,
                 )
```
fregataa, line 262: Let's add a plan to refactor this part to build default policy creator when action instance is created

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -335,17 +338,21 @@ async def create_endpoint_legacy(
...
-                policy_creator = RBACEntityCreator(
-                    spec=spec.policy,
-                    element_type=RBACElementType.DEPLOYMENT_POLICY,
-                    scope_ref=RBACElementRef(
-                        element_type=RBACElementType.MODEL_DEPLOYMENT,
-                        element_id=str(endpoint.id),
-                    ),
-                )
-                await execute_rbac_entity_creator(db_sess, policy_creator)
+                policy_spec = spec.policy
```
fregataa, line 343: Same here

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -335,17 +338,21 @@ async def create_endpoint_legacy(
...
+            else:
+                policy_spec = DeploymentPolicyCreatorSpec.build_default(endpoint.id)
```
seedspirit, line 345: If we define a class method that creates a CreatorSpec by injecting a policy—such as `with_policy`—wouldn’t that allow us to hide the `if` branches in the creation logic that handle situations where the policy is optional? Or is it more important to make the creation of the RBACEntityCreator more explicit in the service layer?

jopemachine: Because the if statement is small and sufficiently readable, I think it does not seem to introduce any meaningful difference in terms of readability.

PR #10534, opened 2026-03-26
test(BA-5375): add RBAC registry completeness test

fregataa:
## Summary
- Extract the inline RBAC action registry list from `factory.py` into a shared `rbac_registry.py` module for testability
- Add `TestRBACRegistryCompleteness` tests that verify every concrete `BaseRBACAction` subclass is registered, with no duplicates and no non-subclass entries
- Ensures new RBAC action implementations (BA-5317~BA-5348) cannot be silently excluded from the `get_entity_valid_operations()` API response

## Test plan
- [x] `pants test tests/unit/manager/actions/test_rbac_registry.py` passes
- [x] `pants test tests/unit/manager/actions/test_rbac_session.py` passes (existing tests unaffected)
- [x] `pants lint` / `pants fmt` / `pants check` clean (only pre-existing `processors.py` mypy error)

### DIFF
--- a/src/ai/backend/manager/actions/action/rbac_registry.py
+++ b/src/ai/backend/manager/actions/action/rbac_registry.py
@@ -1,0 +1,32 @@
+"""Central RBAC action registry list.
+
+This module provides the canonical list of all RBAC action classes
+that should be registered in the PermissionControllerService.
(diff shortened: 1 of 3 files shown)


fregataa:
Closed this PR since it is outdate, I will revisit it in the future

PR #10569, opened 2026-03-26
feat(BA-5425): Implement VFolder Adapter with admin search support

seedspirit:
## Summary
- Implement `VFolderAdapter.admin_search()` with full filter, order, and pagination support via `BatchQuerier` pipeline
- Add `VFolderConditions` with string filters (name, host), enum filters (status, usage_mode), datetime filters (created_at), and cursor pagination factories
- Add `VFolderOrders` with `ORDER_FIELD_MAP` and `resolve_order()` for all `VFolderOrderField` values
- Add `AdminSearchVFoldersInput` / `AdminSearchVFoldersPayload` DTOs
- Implement `_vfolder_data_to_node()` converting `VFolderData` directly to `VFolderNode` DTO (with `BinarySizeInfo` for usage fields)
- Make `VFolderAccessControlInfo.permission` nullable to match `VFolderRow` schema

## Test plan
- [x] `pants check` passes for all changed files
- [x] `pants lint` passes for all changed files
- [ ] `VFolderAdapter.admin_search()` returns correct results with filters and pagination
- [ ] Enum filters (status `in_`/`not_in`, usage_mode `in_`/`not_in`) generate correct SQL conditions
- [ ] String filters (name, host) support all variants (contains, equals, starts_with, ends_with) with case sensitivity
- [ ] DateTime filter (created_at) supports before/after/equals
- [ ] Cursor pagination (forward/backward) and offset pagination work correctly

Resolves BA-5425

### DIFF
--- a/src/ai/backend/manager/api/adapters/vfolder.py
+++ b/src/ai/backend/manager/api/adapters/vfolder.py
@@ -2,8 +2,6 @@
 
 from __future__ import annotations
 
-from uuid import UUID
-
 from ai.backend.common.dto.manager.v2.common import BinarySizeInfo
 from ai.backend.common.dto.manager.v2.vfolder.request import (
     SearchVFoldersInput,
@@ -136,14 +134,6 @@ class VFolderAdapter(BaseAdapter):
             has_next_page=action_result.has_next_page,
             has_previous_page=action_result.has_previous_page,
         )
-
-    async def get(self, vfolder_id: UUID) -> VFolderNode:
-        """Retrieve a single VFolder by ID."""
-        raise NotImplementedError
-
-    async def my_search(self) -> object:
-        """Self-service search for VFolders using current user context."""
-        raise NotImplementedError
 
     # -------------------------------------------------------------------------
     # Filter / Order conversion


src/ai/backend/common/dto/manager/v2/vfolder/response.py
```diff
@@ -204,3 +205,17 @@ class UnshareVFolderPayload(BaseResponseModel):
...
         description="List of email addresses that were unshared from"
     )
+
+
+# ============================================================
+# Search Payload Models
+# ============================================================
+
+
+class AdminSearchVFoldersPayload(BaseResponseModel):
```
fregataa: I think we don't need to divide `SearchVFoldersPayload` and `AdminSearchVFoldersPayload`

src/ai/backend/manager/api/adapters/vfolder.py
```diff
@@ -0,0 +1,215 @@
...
+def _to_binary_size_info(value: int) -> BinarySizeInfo:
+    """Convert bytes integer to BinarySizeInfo DTO."""
+    return BinarySizeInfo(value=value, display=f"{BinarySize(value):s}")
+
+
+class VFolderAdapter(BaseAdapter):
+    """Adapter for VFolder domain operations."""
+
+    @staticmethod
+    def _vfolder_data_to_node(data: VFolderData) -> VFolderNode:
```
jopemachine, line 72: Wouldn’t it be more natural to implement `from_dataclass` on VFolderNode?

seedspirit: Since the conversion itself is generally handled by the adapter layer, the current structure seems appropriate. https://github.com/lablup/backend.ai/blob/main/src/ai/backend/common/dto/manager/v2/CLAUDE.md

> ## Conversion
> - DTO is a pure data structure — no conversion logic inside DTOs
> - Conversion from domain Data types to DTOs is performed in the Adapter layer (`manager/api/adapters/{domain}.py`)
> - No direct DB model or domain Data type imports in DTO modules

PR #10604, opened 2026-03-26
feat(BA-5427): define VFolder filter and order-by GQL types

seedspirit:
## Summary
- Add `VFolderFilter` and `VFolderOrder` DTO classes to `common/dto/manager/v2/vfolder/request.py` with enum filter types for status and usage mode
- Create `gql/vfolder_v2/types/filters.py` with `VFolderV2FilterGQL` (name, host, status, usage_mode, created_at + AND/OR/NOT operators), `VFolderV2OrderFieldGQL` enum, and `VFolderV2OrderByGQL` input
- Update package-level `__init__.py` re-exports for both common DTO and manager GQL packages

## Test plan
- [x] `pants fmt` passes
- [x] `pants fix` passes
- [ ] `pants check` passes in CI
- [ ] Filter and order types are importable from package top-level

Resolves BA-5427

### DIFF
--- a/src/ai/backend/manager/api/gql/vfolder_v2/types/filters.py
+++ b/src/ai/backend/manager/api/gql/vfolder_v2/types/filters.py
@@ -31,10 +31,7 @@ from ai.backend.manager.api.gql.pydantic_compat import PydanticInputMixin
         description=(
             "VFolder operation status. "
             "READY: Folder is ready for use. "
-            "PERFORMING: An operation is in progress. "
             "CLONING: Folder is being cloned. "
-            "MOUNTED: Folder is mounted to a session. "
-            "ERROR: An error occurred. "
             "DELETE_PENDING: Folder is in trash bin (recoverable). "
             "DELETE_ONGOING: Deletion is in progress. "
             "DELETE_COMPLETE: Deletion is complete. "
@@ -45,10 +42,7 @@ from ai.backend.manager.api.gql.pydantic_compat import PydanticInputMixin
 )
 class VFolderV2OperationStatusGQL(StrEnum):
     READY = "ready"
-    PERFORMING = "performing"
     CLONING = "cloning"
-    MOUNTED = "mounted"
-    ERROR = "error"
     DELETE_PENDING = "delete-pending"
     DELETE_ONGOING = "delete-ongoing"
     DELETE_COMPLETE = "delete-complete"


src/ai/backend/common/dto/manager/v2/vfolder/request.py
```diff
@@ -229,3 +239,35 @@ class DeleteInvitationInput(BaseRequestModel):
...
 
     invitation_id: UUID = Field(description="Invitation ID to delete")
+
+
+# ============================================================
+# Search / Filter / Order
+# ============================================================
+
+
+class VFolderFilter(BaseRequestModel):
```
jopemachine: We should append V2 in accordance with our naming conventions.

```py
class VFolderV2Filter(BaseRequestModel):
```

jopemachine:
How is the change to ScopeChainPermissionCheckInput related to adding VFolder filter and order? It would be better to handle it in a separate PR from the type addition.

changes/10598.feature.md
jopemachine, line 1: Why is deleted?

seedspirit: It seems that content from another task was accidentally included when I performed a commit squash. I’ll fix it.

src/ai/backend/manager/api/gql/vfolder_v2/types/filters.py
```diff
@@ -0,0 +1,165 @@
...
+        description=(
+            "VFolder usage mode. "
+            "GENERAL: Normal virtual folder. "
+            "MODEL: Virtual folder for shared models. "
+            "DATA: Virtual folder for shared data."
+        ),
+    ),
+    name="VFolderUsageModeV2",
+)
+class VFolderUsageModeEnumGQL(StrEnum):
```
jopemachine: I think it should be `VFolderUsageModeGQL`

src/ai/backend/manager/api/gql/vfolder_v2/types/filters.py
```diff
@@ -0,0 +1,165 @@
...
+            "ERROR: An error occurred. "
+            "DELETE_PENDING: Folder is in trash bin (recoverable). "
+            "DELETE_ONGOING: Deletion is in progress. "
+            "DELETE_COMPLETE: Deletion is complete. "
+            "DELETE_ERROR: An error occurred during deletion."
+        ),
+    ),
+    name="VFolderOperationStatusV2",
+)
+class VFolderOperationStatusEnumGQL(StrEnum):
```
jopemachine: What I meant was not just changing the filter, but replacing all instances of VFolder with VFolderV2.

seedspirit:
FYI, I will add cloneable and by accessible storage host filter when implementing those features

src/ai/backend/manager/api/gql/vfolder_v2/types/filters.py
```diff
@@ -0,0 +1,165 @@
...
+        ),
+    ),
+    name="VFolderV2OperationStatus",
+)
+class VFolderV2OperationStatusGQL(StrEnum):
+    READY = "ready"
+    PERFORMING = "performing"
+    CLONING = "cloning"
+    MOUNTED = "mounted"
+    ERROR = "error"
```
fregataa: Let's remove PERFORMING MOUNTED and ERROR because the status are not used

fregataa (approved):

PR #10612, opened 2026-03-27
feat(BA-5024): add group memberships to user detail response

### LINKED ISSUE #9925: Add group memberships to user detail response
## Context

Part of epic: BA-5022 | Design proposal: [kubernetes-bridge#231](https://github.com/lablup/backend.ai-kubernetes-bridge/issues/231)

## Summary

Extend the `GET /admin/users/{id`} response to include a `groups` array listing the user's group memberships (group ID and name). The kubernetes-bridge uses this to validate that a K8s user belongs to the [http://Backend.AI](http://Backend.AI)  domain and group derived from the target namespace, without requiring a separate API call for membership checks.

## Deliverables

- `UserDTO` gains `groups: list[UserGroupMembershipDTO]`
- `UserData` gains `groups: list[UserGroupMembership]`
- New lightweight type `UserGroupMembership` with `id: UUID` and `name: str`
- New DTO type `UserGroupMembershipDTO` with `id: str` and `name: str`
- Single-user fetch query in `UserDBSource` loads group data via JOIN with `association_groups_users` and `GroupRow`
- `convert_to_dto()` maps the groups list

## Acceptance Criteria

- [ ] `GET /admin/users/{id`} for a user in two groups returns both groups with correct `id` and `name` values.
- [ ] `GET /admin/users/{id`} for a user in no groups returns `"groups": []` (never `null`).
- [ ] Group memberships reflect current state (if a user is added to or removed from a group, the next GET reflects the change).

## Technical Notes

- Requires a JOIN with `association_groups_users` table (`AssocGroupUserRow` with `user_id` and `group_id` columns)
- Design decision: choose between `joinedload` (single query, one row per group), `selectinload` (two queries), or explicit subquery. The existing `db_source.py` pattern for project-scoped user search already performs JOINs against `AssocGroupUserRow`, suggesting `joinedload` or explicit JOIN is consistent.
- The `groups` field must never be `null` in the JSON response – always an empty array for users with no memberships
- A user with many group memberships will inflate the response; this is acceptable for the expected cardinality (users typically belong to 1-5 groups)
- Must tolerate concurrent group/membership deletion (LEFT JOIN or equivalent)

## Prerequisites

None – independent of the `integration_id` changes.

JIRA Issue: BA-5024

hhoikoo:
resolves #9925 (BA-5024)

Extend `GET /admin/users/{id}` to include a `groups` array listing the user's group memberships (UUID and name). The kubernetes-bridge uses this to validate that a K8s user belongs to the Backend.AI domain and group derived from the target namespace, without requiring a separate API call for membership checks.

| Layer | File | Change |
|-------|------|--------|
| Data | `manager/data/user/types.py` | New `UserGroupMembership` dataclass (`id: UUID`, `name: str`); new `UserDetail` dataclass composing `UserData` + `groups: list[UserGroupMembership]` |
| DTO | `common/dto/manager/v2/user/response.py` | New `UserGroupMembershipInfo` (`id: UUID`, `name: str`); `UserNode.groups` field |
| ORM | `manager/models/user/row.py` | `to_data()` unchanged — groups are not part of `UserData` |
| Repository | `manager/repositories/user/db_source/db_source.py` | New `get_user_detail_by_uuid` method eagerly loads groups via `selectinload` + `joinedload` + `load_only(GroupRow.id, GroupRow.name)`, builds the `UserGroupMembership` list, and returns `UserDetail`. `get_user_by_uuid` remains unchanged (returns plain `UserData`). |
| Repository | `manager/repositories/user/repository.py` | New `get_user_detail_by_uuid` public method delegating to db_source |
| Service | `manager/services/user/service.py` | `get_user` calls `get_user_detail_by_uuid` instead of `get_user_by_uuid` |
| Action | `manager/services/user/actions/get_user.py` | `GetUserActionResult.user` typed as `UserDetail` instead of `UserData` |
| Adapter | `manager/api/adapters/user.py` | `_user_data_to_node` accepts optional `groups` kwarg; `get()` passes `detail.groups` from `UserDetail` |

- [x] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/api/rest/user/handler.py
+++ b/src/ai/backend/manager/api/rest/user/handler.py
@@ -125,7 +125,7 @@ class UserHandler:
             GetUserAction(user_uuid=path.parsed.user_id)
         )
 
-        resp = GetUserResponse(user=self._adapter.convert_to_dto(action_result.user))
+        resp = GetUserResponse(user=self._adapter.convert_to_dto(action_result.user.user))
         return APIResponse.build(status_code=HTTPStatus.OK, response_model=resp)
 
     # ------------------------------------------------------------------
@@ -170,7 +170,7 @@ class UserHandler:
         get_result = await self._user.get_user.wait_for_complete(
             GetUserAction(user_uuid=path.parsed.user_id)
         )
-        email = get_result.user.email
+        email = get_result.user.user.email
 
         # Build password info if password is being updated
         password_info: PasswordInfo | None = None
@@ -208,7 +208,7 @@ class UserHandler:
         )
 
         await self._user.delete_user.wait_for_complete(
-            DeleteUserAction(email=get_result.user.email)
+            DeleteUserAction(email=get_result.user.user.email)
         )
 
         resp = DeleteUserResponse(success=True)
@@ -236,9 +236,9 @@ class UserHandler:
         )
 
         user_info_ctx = UserInfoContext(
-            uuid=caller_result.user.uuid,
-            email=caller_result.user.email,
-            main_access_key=AccessKey(caller_result.user.main_access_key or ""),
+            uuid=caller_result.user.user.uuid,
+            email=caller_result.user.user.email,
+            main_access_key=AccessKey(caller_result.user.user.main_access_key or ""),
         )
 
         purge_shared = OptionalState[bool].nop()
@@ -252,7 +252,7 @@ class UserHandler:
             PurgeUserAction(
                 user_uuid=body.parsed.user_id,
                 user_info_ctx=user_info_ctx,
-                email=get_result.user.email,
+                email=get_result.user.user.email,
                 purge_shared_vfolders=purge_shared,
                 delegate_endpoint_ownership=delegate_endpoint,
             )


src/ai/backend/common/dto/manager/v2/user/response.py
```diff
@@ -154,6 +155,17 @@ class EntityTimestamps(BaseModel):
...
 
+class UserGroupMembershipDTO(BaseResponseModel):
+    """Group membership information for a user."""
+
+    id: str = Field(
+        description="Unique identifier of the group.",
+    )
+    name: str = Field(
+        description="Name of the group.",
+    )
```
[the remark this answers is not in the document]
hhoikoo: Fixed — changed to UUID for consistency with UserNode.id and other v2 DTOs.

tests/unit/manager/api/adapters/test_user_adapter.py
```diff
@@ -0,0 +1,85 @@
...
+
+def _create_user_data(
+    user_id: UUID | None = None,
+    groups: list[UserGroupMembership] | None = None,
+) -> UserData:
+    """Create a minimal UserData for testing adapter conversion."""
+    now = datetime.now(tz=UTC)
+    return UserData(
+        id=user_id or uuid4(),
+        uuid=user_id or uuid4(),
```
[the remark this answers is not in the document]
hhoikoo: Fixed — now generates a single UUID and reuses it for both id and uuid.

src/ai/backend/manager/repositories/user/db_source/db_source.py
```diff
@@ -134,7 +134,16 @@ async def get_user_by_uuid(self, user_uuid: UUID) -> UserData:
         Admin-only operation.
         """
         async with self._db.begin_readonly_session_read_committed() as db_session:
-            user_row = await self._get_user_by_uuid(db_session, user_uuid)
+            query = (
+                sa.select(UserRow)
+                .where(UserRow.uuid == user_uuid)
+                .options(
+                    selectinload(UserRow.groups).joinedload(AssocGroupUserRow.group),
```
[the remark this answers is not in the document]
hhoikoo: Fixed — added load_only(GroupRow.id, GroupRow.name) to limit fetched columns.

src/ai/backend/manager/repositories/user/db_source/db_source.py
```diff
@@ -134,7 +134,16 @@ async def get_user_by_uuid(self, user_uuid: UUID) -> UserData:
...
+                )
+            )
+            user_row = await db_session.scalar(query)
+            if user_row is None:
+                raise UserNotFound(f"User with UUID {user_uuid} not found.")
```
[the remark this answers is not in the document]
hhoikoo: Intentional — the inline query needs selectinload options that _get_user_by_uuid does not support. Adding an options parameter to a private helper used by other callers would leak eager-load concerns into unrelated code paths.

src/ai/backend/common/dto/manager/v2/user/response.py
```diff
@@ -154,6 +155,17 @@ class EntityTimestamps(BaseModel):
     )
 
 
+class UserGroupMembershipDTO(BaseResponseModel):
```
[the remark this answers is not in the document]
hhoikoo: Keeping BaseResponseModel — the v2 DTO CLAUDE.md states nested sub-models must also inherit from BaseResponseModel. The existing sub-models using BaseModel predate that guideline.

src/ai/backend/manager/models/user/row.py
```diff
@@ -405,6 +405,14 @@ def to_model_serving_user_data(self) -> ModelServingUserData:
         )
 
     def to_data(self) -> UserData:
+        groups: list[UserGroupMembership] = []
+        if "groups" in self.__dict__:
+            groups = [
+                UserGroupMembership(id=assoc.group.id, name=assoc.group.name)
+                for assoc in self.groups
+                if assoc.group is not None
```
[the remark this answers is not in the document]
hhoikoo: The only caller that sets groups in __dict__ is get_user_by_uuid, which chains selectinload(UserRow.groups).joinedload(AssocGroupUserRow.group) — so assoc.group is always eagerly loaded when this code runs. The if assoc.group is not None guard handles the theoretical edge case. Adding sa.inspect(assoc).unloaded checks would add overhead for a scenario that cannot occur with the current single caller.

tests/unit/manager/data/user/test_types.py
hhoikoo, line 1: really? no good existing place to put this test file in? or is this even testing anything useful?

src/ai/backend/manager/repositories/user/db_source/db_source.py
```diff
@@ -134,7 +134,18 @@ async def get_user_by_uuid(self, user_uuid: UUID) -> UserData:
         Admin-only operation.
         """
         async with self._db.begin_readonly_session_read_committed() as db_session:
-            user_row = await self._get_user_by_uuid(db_session, user_uuid)
```
hhoikoo, line 137: can we not modify this helper function itself somehow? not sure why you went for rewriting the query here

src/ai/backend/manager/models/user/row.py
```diff
@@ -405,6 +405,14 @@ def to_model_serving_user_data(self) -> ModelServingUserData:
         )
 
     def to_data(self) -> UserData:
+        groups: list[UserGroupMembership] = []
+        if "groups" in self.__dict__:
```
hhoikoo: why __dict__? really no better way of checking this?

src/ai/backend/common/dto/manager/v2/user/response.py
hhoikoo: don't think the DTO suffix follows the convention with the rest of the codebase

tests/unit/manager/data/user/test_types.py
hhoikoo: Removed — these tests only verified dataclass instantiation and default values, which the adapter tests already cover meaningfully.

src/ai/backend/manager/repositories/user/db_source/db_source.py
hhoikoo: Refactored — _get_user_by_uuid now accepts an optional options parameter. get_user_by_uuid passes the eager load options through it.

src/ai/backend/manager/models/user/row.py
hhoikoo: Changed to sa.inspect(self).unloaded — the idiomatic SQLAlchemy way to check relationship loaded state.

src/ai/backend/common/dto/manager/v2/user/response.py
hhoikoo: Renamed to UserGroupMembershipInfo to follow the v2 DTO naming convention ({Entity}{Group}Info) per common/dto/manager/v2/CLAUDE.md.

src/ai/backend/manager/models/user/row.py
```diff
@@ -405,6 +405,14 @@ def to_model_serving_user_data(self) -> ModelServingUserData:
         )
 
     def to_data(self) -> UserData:
+        groups: list[UserGroupMembership] = []
+        if "groups" not in sa.inspect(self).unloaded:
+            groups = [
+                UserGroupMembership(id=assoc.group.id, name=assoc.group.name)
+                for assoc in self.groups
+                if assoc.group is not None
```
HyeockJinKim: Do you need sa.inspec?

hhoikoo: Refactored — to_data() now accepts groups as an explicit keyword parameter instead of using sa.inspect. The caller (get_user_by_uuid in db_source) builds the UserGroupMembership list from the eagerly-loaded relationship and passes it in. This keeps to_data() a pure data mapper with no ORM introspection.

src/ai/backend/manager/models/user/row.py
```diff
@@ -404,7 +404,13 @@ def to_model_serving_user_data(self) -> ModelServingUserData:
             email=self.email,
         )
 
-    def to_data(self) -> UserData:
+    def to_data(
+        self,
+        *,
+        groups: list[UserGroupMembership] | None = None,
```
jopemachine: Including GroupData within UserData or UserData.to_data does not seem appropriate.

Would it be better to introduce a separate type—such as UserWithGroup or UserDetail—that explicitly represents a combined view of both user and group information?

src/ai/backend/manager/repositories/user/db_source/db_source.py
```diff
@@ -134,8 +135,21 @@ async def get_user_by_uuid(self, user_uuid: UUID) -> UserData:
...
-            return user_row.to_data()
+            user_row = await self._get_user_by_uuid(
+                db_session,
+                user_uuid,
+                options=[
+                    selectinload(UserRow.groups)
+                    .joinedload(AssocGroupUserRow.group)
+                    .load_only(GroupRow.id, GroupRow.name),
```
jopemachine, line 154: Conditionally loading groups depending on the situation is confusing.

Instead of adding options to _get_user_by_uuid, let’s define a separate UserDetail type and provide a separate method that returns UserDetail, so the behavior is explicit and less confusing.

src/ai/backend/manager/models/user/row.py
hhoikoo: Agreed — introduced a separate UserDetail dataclass that composes UserData + groups. UserData no longer carries group membership. The detail view (get_user_by_uuid) returns UserDetail via a dedicated get_user_detail_by_uuid method.

src/ai/backend/manager/repositories/user/db_source/db_source.py
hhoikoo: Agreed — removed the options parameter from _get_user_by_uuid and added a dedicated get_user_detail_by_uuid method that returns UserDetail. The behavior is now explicit: get_user_by_uuid returns plain UserData, get_user_detail_by_uuid returns UserDetail with groups.

src/ai/backend/manager/data/user/types.py
```diff
@@ -56,6 +56,12 @@ def _missing_(cls, value: Any) -> UserStatus | None:
         return None
 
 
+@dataclass
+class UserGroupMembership:
+    id: UUID
+    name: str
```
jopemachine, line 62: From the type name and field names alone, it’s not entirely clear what the values represent—it could be confusing whether they refer to a user or a group.

How about using group_id and group_name instead?

PR #10621, opened 2026-03-27
feat(BA-5481): add role_assignments field to UserV2GQL type

fregataa:
## Summary
- Add a `role_assignments` field to `UserV2GQL` that returns RBAC role assignments for the user
- Uses the existing `role_assignments_by_user_loader` DataLoader to prevent N+1 queries
- Decorated with `gql_added_field` using `NEXT_RELEASE_VERSION` for proper version tracking

## Test plan
- [ ] Verify `role_assignments` field resolves correctly in GraphQL playground
- [ ] Confirm DataLoader batches multiple user role assignment lookups

Resolves BA-5481

### DIFF
--- a/changes/10621.feature.md
+++ b/changes/10621.feature.md
@@ -1,0 +1,1 @@
+Add `role_assignments` field to `UserV2GQL` type exposing RBAC role assignments via DataLoader.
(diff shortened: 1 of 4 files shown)


jopemachine:
This PR appears simple, is there a reason why this is held in a draft state?

fregataa:
Closed this PR since it is not updated anymore

PR #10658, opened 2026-03-30
feat(BA-3517): Rename `service-definition.toml` to `deployment-config.yaml`

jopemachine:
Resolves BA-3517.

Rename the deployment configuration file from service-definition.toml to deployment-config.yaml to avoid naming conflicts with the existing "service definition" concept (container app definitions like JupyterLab, ttyd). The new name better reflects its purpose as deployment configuration.

- File format changed from TOML to YAML (consistent with model-definition.yaml)
- Backward compatibility: falls back to service-definition.toml if deployment-config.yaml is not found, with a deprecation log message
- Internal naming updated: ModelServiceDefinition → DeploymentConfig, fetch_service_definition → fetch_deployment_config, etc.

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/manager/repositories/deployment/repository.py
+++ b/src/ai/backend/manager/repositories/deployment/repository.py
@@ -96,6 +96,9 @@ from .types import RouteData, RouteServiceDiscoveryInfo
 
 log = BraceStyleAdapter(logging.getLogger(__name__))
 
+_DEPLOYMENT_CONFIG_FILENAME = "deployment-config.yaml"
+_LEGACY_SERVICE_DEFINITION_FILENAME = "service-definition.toml"
+
 
 @dataclass
 class AutoScalingMetricsData:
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/repositories/deployment/storage_source/storage_source.py
```diff
@@ -17,43 +14,6 @@ class DeploymentStorageSource:
     def __init__(self, storage_manager: StorageSessionManager) -> None:
         self._storage_manager = storage_manager
 
-    async def fetch_service_config(
```
jopemachine, line 20: This is being deleted since it is not used anywhere

src/ai/backend/manager/repositories/deployment/repository.py
```diff
@@ -471,38 +472,70 @@ async def fetch_model_definition(
...
             )
 
-        # Read service definition from storage
-        service_definition_content: dict[str, Any] | None = None
-        try:
-            service_definition_bytes = await self._storage_source.fetch_definition_file(
-                vfolder_location,
-                ["service-definition.toml"],
+        # Try deployment-config.yaml first (new format)
+        config = await self._try_fetch_config_file(vfolder_location, "deployment-config.yaml")
```
fregataa: It would be nice to set a const file name

src/ai/backend/manager/repositories/deployment/repository.py
```diff
@@ -471,38 +472,70 @@ async def fetch_model_definition(
...
+        if config is not None:
+            return config
+
+        # Fall back to legacy service-definition.toml
+        config = await self._try_fetch_config_file(vfolder_location, "service-definition.toml")
```
fregataa: Same goes to legacy service-definition.toml file

PR #10662, opened 2026-03-30
feat(BA-5318): add RBAC action declarations for PROJECT entity

fregataa:
## Summary
- Define 11 RBAC action classes for the PROJECT element type (create, get, search, update, soft-delete, hard-delete, and 5 grant operations)
- Register all PROJECT RBAC actions in `RBAC_ACTION_REGISTRY`
- Part of Epic BA-5316 (RBAC action declarations)

## Test plan
- [x] `pants lint` passes
- [x] `pants check` (mypy) passes
- [x] `pants fmt` / `pants fix` — no changes needed

### DIFF
--- a/changes/10662.feature.md
+++ b/changes/10662.feature.md
@@ -1,0 +1,1 @@
+Add RBAC action declarations for the PROJECT entity


jopemachine:
For this kind of work, I think it would be more efficient to handle it in a single PR rather than splitting it by entity.

PR #10666, opened 2026-03-30
fix(BA-5499): prevent health checker from recreating deleted etcd keys in Traefik mode

seedspirit:
## Summary
- Add per-circuit `asyncio.Lock` to `CircuitManager` to serialize `update_circuit_routes()` and `unload_circuits()` for the same circuit, preventing race conditions
- Re-read circuit from DB in `propagate_route_updates_to_workers()` before etcd write; skip propagation if circuit was deleted (`ObjectNotFound`)
- Use fresh `route_info` instead of stale in-memory snapshot to prevent stale backend URLs from persisting in Traefik's weighted service pool

## Test plan
- [x] Unit tests pass (`pants test tests/unit/appproxy/coordinator::`)
- [x] Type checking passes (`pants check src/ai/backend/appproxy/coordinator::`)
- [x] Lint passes (`pants lint --changed-since=origin/main`)
- [ ] Manual QA: scale-in scenario does not leave stale session service keys in etcd (see DEVIATION.md)

Resolves BA-5499

### DIFF
--- a/src/ai/backend/appproxy/coordinator/types.py
+++ b/src/ai/backend/appproxy/coordinator/types.py
@@ -84,7 +84,6 @@ class CircuitManager:
     traefik_etcd: TraefikEtcd | None
     local_config: ServerConfig
     _circuit_locks: dict[UUID, asyncio.Lock] = field(default_factory=dict)
-    _unloaded_circuits: set[UUID] = field(default_factory=set)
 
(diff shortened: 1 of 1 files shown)


seedspirit:
Close because it is duplicated with https://github.com/lablup/backend.ai/pull/10692

PR #10667, opened 2026-03-30
fix(BA-5498): Use vfolder last_used as fallback for ModelCard modified_at (#10654)

lablup-octodog:
This is an auto-generated backport PR of #10654 to the 26.3 release.


seedspirit:
Since it needs to use modified_at or updated_at, the work described in this PR doesn't seem to be correct, so I'm closing it.

PR #10674, opened 2026-03-31
feat(BA-5505): add cloneable filter to VFolder my_search

seedspirit:
## Summary
- Add `cloneable: Boolean` field to `VFolderFilter` DTO and `VFolderFilterGQL` GraphQL type
- Add `VFolderConditions.by_cloneable()` condition factory and wire it in `VFolderAdapter._convert_vfolder_filter()`
- Comprehensive unit and integration tests covering filter logic, pagination, and filter combination

## Test plan
- [x] BooleanFilter.build_query_condition() unit tests (true/false)
- [x] VFolderConditions.by_cloneable() SQL generation tests
- [x] Adapter _convert_vfolder_filter with cloneable={eq:true}, {eq:false}, and None
- [x] DB integration tests: cloneable=true (owned+shared), cloneable=false, no filter, pagination, combination with usage_mode

Resolves BA-5505

----
📚 Documentation preview 📚: https://sorna--10674.org.readthedocs.build/en/10674/

----
📚 Documentation preview 📚: https://sorna-ko--10674.org.readthedocs.build/ko/10674/

### DIFF
--- a/tests/unit/manager/repositories/vfolder/test_vfolder_search_filter.py
+++ b/tests/unit/manager/repositories/vfolder/test_vfolder_search_filter.py
@@ -35,8 +35,8 @@ from ai.backend.manager.repositories.vfolder.types import UserVFolderSearchScope
 from ai.backend.testutils.db import with_tables
 
 
-class TestVfolderSearchCloneableFilter:
-    """Tests for search_user_vfolders with cloneable condition filter."""
+class TestVfolderSearchFilter:
+    """Tests for search_user_vfolders with filters."""
 
     @pytest.fixture
     async def db_with_cleanup(


docs/manager/graphql-reference/supergraph.graphql
```diff
@@ -1692,6 +1692,15 @@ input BlueGreenConfigInput
...
 }
 
+"""
+Added in UNRELEASED. Filter for boolean fields supporting equality operation.
+"""
+input BooleanFilter
+  @join__type(graph: STRAWBERRY)
+{
+  equals: Boolean!
+}
```
HyeockJinKim: Since there's nothing besides `equal` for `bool`, there doesn't seem to be any reason to create one.

tests/unit/manager/models/vfolder/test_conditions.py
```diff
@@ -0,0 +1,35 @@
...
+        lower_sql = sql.lower()
+        assert "vfolders.cloneable" in lower_sql
+        assert any(literal in lower_sql for literal in expected_literals)
+
+    def test_by_cloneable_returns_callable(self) -> None:
+        """by_cloneable returns a QueryCondition (callable)."""
+        condition = VFolderConditions.by_cloneable(True)
+        assert callable(condition)
+        result = condition()
+        assert hasattr(result, "compile")
```
jopemachine: I think this is excessive testing.
We don’t need to add this kind of test every time a boolean filter is introduced.

tests/unit/manager/repositories/vfolder/test_vfolder_search_filter.py
```diff
@@ -0,0 +1,453 @@
+"""
+Tests for VfolderRepository.search_user_vfolders() with cloneable filter.
+Verifies that the cloneable condition correctly filters vfolders.
+"""
```
fregataa, line 4: Let's change the filename for general vfolder search filter because it is too verbose to test a single filter

PR #10690, opened 2026-03-31
fix(BA-5512): Use maxWaitSeconds as actual timeout in session creation wait loop

fregataa:
## Summary
- Use `max_wait` value as the actual timeout in `_wait_for_session_running()` instead of always using the hardcoded `DEFAULT_WAIT_TIMEOUT_SECONDS` (60s)
- When `max_wait > 0`, the timeout is now set to `max_wait` seconds; when `max_wait == 0`, the existing 60-second poll interval behavior is preserved
- Add unit tests covering timeout with custom max_wait, infinite retry with max_wait=0, early RUNNING event, and TERMINATED during wait

## Test plan
- [x] Unit tests for `_wait_for_session_running()` timeout scenarios
- [x] Verify `maxWaitSeconds: 15` returns `TIMEOUT` status within ~15 seconds (manual/integration)
- [x] Verify `maxWaitSeconds: 0` still retries indefinitely with 60s poll intervals

Resolves BA-5512

### DIFF
--- a/tests/unit/manager/test_registry.py
+++ b/tests/unit/manager/test_registry.py
@@ -671,42 +671,6 @@ class TestWaitForSessionRunning:
         elapsed = time.monotonic() - start
         assert elapsed < 1, f"Should return immediately, took {elapsed}s"
 
-    async def test_terminated_event_raises_session_not_found(
-        self,
-        session_id: SessionId,
-        mock_propagator: MagicMock,
-        mock_registry_obj: MagicMock,
-    ) -> None:
-        """TERMINATED event during wait should raise SessionNotFound."""
-
-        async def _yield_terminated(cache_id: str) -> AsyncIterator[AbstractEvent]:
-            yield _make_scheduling_event(session_id, str(SessionStatus.TERMINATED))
-
-        mock_propagator.receive = _yield_terminated
-
-        with pytest.raises(SessionNotFound):
-            await mock_registry_obj._wait_for_session_running(
-                session_id, mock_propagator, max_wait=15
-            )
-
-    async def test_cancelled_event_raises_session_not_found(
-        self,
-        session_id: SessionId,
-        mock_propagator: MagicMock,
-        mock_registry_obj: MagicMock,
-    ) -> None:
-        """CANCELLED event during wait should raise SessionNotFound."""
-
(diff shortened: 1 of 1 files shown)


tests/unit/manager/test_registry.py
```diff
@@ -550,3 +556,217 @@ async def test_cmd_variant_with_override_uses_yaml_path_and_pydantic_defaults(
         assert result.max_wait_time == 15.0
         assert result.expected_status_code == 200
         assert result.initial_delay == 60.0
+
+
+def _make_scheduling_event(
```
seedspirit, line 561: It appears to be available as a fixture

tests/unit/manager/test_registry.py
```diff
@@ -550,3 +556,217 @@ async def test_cmd_variant_with_override_uses_yaml_path_and_pydantic_defaults(
...
+
+        mock_propagator.receive = _hang_forever
+
+        start = time.monotonic()
+        with pytest.raises(TimeoutError):
+            await mock_registry_obj._wait_for_session_running(
+                session_id, mock_propagator, max_wait=2
+            )
+        elapsed = time.monotonic() - start
+        assert elapsed < 5, f"Timed out in {elapsed}s, expected ~2s"
```
seedspirit, line 613: Wouldn't it be better to set it to around 2.5 or 3? 5 seems a bit too generous.

tests/unit/manager/test_registry.py
```diff
@@ -550,3 +556,217 @@ async def test_cmd_variant_with_override_uses_yaml_path_and_pydantic_defaults(
...
+        mock_row = MagicMock()
+        mock_row.status = SessionStatus.RUNNING
+        mock_result.first.return_value = mock_row
+        mock_db_session.execute = AsyncMock(return_value=mock_result)
+
+        mock_db_ctx = MagicMock()
+        mock_db_ctx.__aenter__ = AsyncMock(return_value=mock_db_session)
+        mock_db_ctx.__aexit__ = AsyncMock(return_value=False)
+        mock_registry_obj.db = MagicMock()
+        mock_registry_obj.db.begin_readonly_session = MagicMock(return_value=mock_db_ctx)
```
seedspirit, line 645: It looks like we can separate it into fixtures for mocking

PR #10692, opened 2026-03-31
fix(BA-5499): Prevent race conditions in circuit route propagation in Traefik mode

### LINKED ISSUE #10656: Health checker recreates deleted etcd keys due to race condition with circuit deletion in Traefik mode
## Problem

Health checker periodically overwrites etcd service keys using a stale in-memory `circuit.route_info` snapshot. When routes change between the DB read and etcd write (session termination, scale-in, endpoint deletion), stale backend URLs persist in Traefik's weighted service pool.

\***Reported symptom:**\* Request to Model A's endpoint returns Model B's response. "Sync Route Info" button fixes it temporarily.

\***Mechanism:**\* Model A's weighted service (`bai_service_{circuit_id}`) contains a stale backend URL from a removed session. After the container at that address is freed and reassigned to Model B, Traefik load-balances some of Model A's traffic to Model B's container.

Traefik mode only (`enable_traefik = true`). Python frontends unaffected.

## Root Cause

`propagate_route_updates_to_workers()` writes etcd service keys from an in-memory circuit object without:
1. Verifying circuit still exists in DB
2. Re-reading fresh route_info before writing
3. Any concurrency control (no Lock on CircuitManager)

## Trigger Scenarios

- Scale-in: removed replica's service key recreated by health checker
- Session crash/kill: dead session's backend URL persists in weighted service
- Endpoint deletion: circuit's service keys recreated after unload
- Health state transition during concurrent route update: stale route set overwrites correct one

Not limited to scale-in/out — any route_info change during a health check cycle.

## Affected Files

- `appproxy/coordinator/health_checker.py:434-507` — `propagate_route_updates_to_workers()`
- `appproxy/coordinator/types.py:80-85` — `CircuitManager` (no Lock)
- `appproxy/coordinator/types.py:163-206` — `update_traefik_circuit_routes()`
- `appproxy/coordinator/server.py:495-578` — `on_route_update_event()`

## Success Criteria

- [ ] `propagate_route_updates_to_workers()` re-reads circuit from DB before etcd write; skips if deleted
- [ ] `CircuitManager` uses per-circuit `asyncio.Lock` to serialize `update_circuit_routes()` and `unload_circuits()`
- [ ] Scale-in does not leave stale session service keys in etcd
- [ ] Session crash does not leave stale backend URLs in weighted service pool
- [ ] pants test passes for affected packages

JIRA Issue: BA-5499

seedspirit:
resolves #10656 (BA-5499)

- [ ] Test case(s) to:
- [ ] Documentation

### DIFF
--- a/src/ai/backend/appproxy/coordinator/types.py
+++ b/src/ai/backend/appproxy/coordinator/types.py
@@ -95,7 +95,7 @@ class CircuitManager:
         self._circuit_locks.pop(circuit_id, None)
 
     @actxmgr
-    async def _circuit_lock(self, circuit_id: UUID) -> AsyncIterator[None]:
+    async def circuit_lock(self, circuit_id: UUID) -> AsyncIterator[None]:
         async with self._get_lock(circuit_id):
             yield
 
@@ -170,7 +170,7 @@ class CircuitManager:
         self.event_dispatcher.unsubscribe(worker_ready_event_handler)
 
     async def update_circuit_routes(self, circuit: Circuit, old_routes: list[RouteInfo]) -> None:
-        async with self._circuit_lock(circuit.id):
+        async with self.circuit_lock(circuit.id):
             await self._update_circuit_routes_unlocked(circuit, old_routes)
 
     async def _update_circuit_routes_unlocked(
@@ -242,7 +242,7 @@ class CircuitManager:
     async def unload_circuits(self, circuits: Sequence[Circuit]) -> None:
         for circuit in circuits:
             try:
-                async with self._circuit_lock(circuit.id):
+                async with self.circuit_lock(circuit.id):
                     if self.local_config.proxy_coordinator.enable_traefik:
                         await self.unload_traefik_circuit(circuit)
                     else:


src/ai/backend/appproxy/coordinator/types.py
```diff
@@ -219,11 +240,13 @@ async def update_legacy_circuit_routes(
...
-                await self.unload_traefik_circuit(circuit)
-        else:
-            await self.unload_legacy_circuits(circuits)
+        for circuit in circuits:
+            async with self.circuit_lock(circuit.id):
+                if self.local_config.proxy_coordinator.enable_traefik:
+                    await self.unload_traefik_circuit(circuit)
+                else:
+                    await self.unload_legacy_circuit(circuit)
+            self.release_circuit_lock(circuit.id)
```
fregataa: Let's wrap this part with try-except-finally:
- To ensure circuit lock release
- To keep iteration for unloading circuits

seedspirit:
> Do we have prometheus infra in coordinator server?



Yes, you can see in `CoordinatorMetricRegistry` (coordinator/types.py:53-73)

src/ai/backend/appproxy/coordinator/types.py
```diff
@@ -219,11 +240,15 @@ async def update_legacy_circuit_routes(
...
-            await self.unload_legacy_circuits(circuits)
+        for circuit in circuits:
+            try:
+                async with self.circuit_lock(circuit.id):
+                    if self.local_config.proxy_coordinator.enable_traefik:
+                        await self.unload_traefik_circuit(circuit)
+                    else:
+                        await self.unload_legacy_circuit(circuit)
+            finally:
+                self.release_circuit_lock(circuit.id)
```
fregataa: Let's add `except` block here to avoid iteration stop

src/ai/backend/appproxy/coordinator/types.py
```diff
@@ -83,6 +84,20 @@ class CircuitManager:
...
     traefik_etcd: TraefikEtcd | None
     local_config: ServerConfig
+    _circuit_locks: dict[UUID, asyncio.Lock] = field(default_factory=dict)
+
+    def _get_lock(self, circuit_id: UUID) -> asyncio.Lock:
+        if circuit_id not in self._circuit_locks:
+            self._circuit_locks[circuit_id] = asyncio.Lock()
+        return self._circuit_locks[circuit_id]
+
+    def release_circuit_lock(self, circuit_id: UUID) -> None:
```
jopemachine: Should it be public?

seedspirit: In the initial code design, there were parts exposed externally, but since it seems better to handle lock management internally, I will change them to private.

PR #10696, opened 2026-03-31
fix(BA-5519): read container net stats from /proc/[pid]/net/dev with netns fallback

fregataa:
## Summary
- Replace `netstat_ns()` + `setns()` with direct `/proc/[container_pid]/net/dev` reading for container network stats in cgroup mode
- Root cause: agent runs as daemon subprocess → `ProcessPoolExecutor` unavailable → thread pool fallback → `setns()` changes thread namespace but `psutil` reads `/proc/self/net/dev` (process-level = host namespace)
- Remove dead code: `netstat_ns()`, `netstat_ns_work()`, unused imports (`multiprocessing`, `ProcessPoolExecutor`, `nsenter`)

Fixes BA-5519

## Test plan
- [x] Verify container net_rx/net_tx values match `docker stats` output
- [x] Verify `/proc/[container_pid]/net/dev` shows only container interfaces (lo, eth0)
- [x] Verify PID=0 case (stopped container) returns 0 with warning log
- [x] Run existing unit tests for agent intrinsic plugins

### DIFF
--- a/tests/unit/agent/test_docker_intrinsic.py
+++ b/tests/unit/agent/test_docker_intrinsic.py
@@ -580,42 +580,31 @@ class TestReadProcNetDev:
         "        8888       10    0    0    0     0       0          0\n"
     )
 
-    def test_parses_standard_format(self, tmp_path: Path) -> None:
-        """Parses standard /proc/net/dev format: skips headers, excludes lo,
-        returns correct rx/tx sums."""
+    @pytest.mark.parametrize(
+        ("content_attr", "expected_rx", "expected_tx"),
+        [
+            ("SAMPLE_NET_DEV", 50000, 80000),
+            ("MULTI_IFACE_NET_DEV", 40000, 60000),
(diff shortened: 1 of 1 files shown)


tests/unit/agent/test_docker_intrinsic.py
```diff
@@ -566,14 +534,90 @@ async def test_cancelled_error_is_reraised(
...
+        """When only loopback is present, returns (0, 0)."""
+        net_dev = tmp_path / "net_dev"
+        net_dev.write_text(self.LO_ONLY_NET_DEV)
+        with patch(
+            "ai.backend.agent.docker.intrinsic.Path",
+            return_value=net_dev,
+        ):
+            result = read_proc_net_dev(42)
+        assert result.rx_bytes == 0
+        assert result.tx_bytes == 0
```
seedspirit: I think this test can be generalized with parametrize

seedspirit (commented):
The logic seems fine, but I think we'll need @HyeockJinKim's review as well.

PR #10704, opened 2026-04-01
feat(BA-5514): register roles in association_scopes_entities with optional scope

fregataa:
## Summary
- Add optional scope registration when creating roles via `ScopeInput` (scope_type + scope_id)
- When scope is provided, `RBACEntityCreator` atomically creates the role row and its scope association; when absent, only the role row is created
- Move `RBACElementTypeGQL` to a new `scope.py` module to break the circular import chain and co-locate it with `ScopeInputGQL`

## Test plan
- [x] `pants check` passes (0 new errors)
- [x] `tests/unit/manager/api/gql/rbac/test_rbac_element_type_sync.py` passes
- [x] `tests/unit/common/dto/manager/v2/rbac::` passes
- [x] CI green

Resolves BA-5514

----
📚 Documentation preview 📚: https://sorna--10704.org.readthedocs.build/en/10704/

----
📚 Documentation preview 📚: https://sorna-ko--10704.org.readthedocs.build/ko/10704/

### DIFF
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -3524,7 +3524,7 @@ input CreateRoleInput
   name: String!
   description: String = null
   source: RoleSource! = CUSTOM
-  scope: ScopeInput = null
+  scopes: [ScopeInput!] = null
 }
 
 """Added in UNRELEASED. Input for creating a runtime variant."""
--- a/docs/manager/graphql-reference/v2-schema.graphql
+++ b/docs/manager/graphql-reference/v2-schema.graphql
@@ -2145,7 +2145,7 @@ input CreateRoleInput {
   name: String!
   description: String = null
   source: RoleSource! = CUSTOM
-  scope: ScopeInput = null
+  scopes: [ScopeInput!] = null
 }
 
 """Added in UNRELEASED. Input for creating a runtime variant."""


src/ai/backend/common/dto/manager/v2/rbac/types.py
```diff
@@ -174,6 +175,13 @@ class EntityTypeFilter(BaseRequestModel):
     in_: list[str] | None = None
 
 
+class ScopeInput(BaseRequestModel):
```
jopemachine: Just a question: some of the DTO types have the DTO suffix while others don’t—wouldn’t it be better to make this consistent?

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
```diff
@@ -111,9 +118,16 @@ async def create_role(self, input_data: CreateRoleInput) -> RoleRow:
...
         async with self._db.begin_session() as db_session:
-            # 1. Create role
-            result = await execute_creator(db_session, input_data.creator)
-            role_row = result.row
+            if input_data.scope_refs:
+                rbac_creator = RBACEntityCreator(
+                    spec=input_data.creator.spec,
+                    element_type=RBACElementType.ROLE,
+                    scope_ref=input_data.scope_refs[0],
+                    additional_scope_refs=input_data.scope_refs[1:],
```
jopemachine, line 126: It seems preferable to distinguish them explicitly through separate methods, rather than implicitly relying on the semantic difference between the 0th element and the subsequent elements in the array.

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
```diff
@@ -111,9 +118,16 @@ async def create_role(self, input_data: CreateRoleInput) -> RoleRow:
...
+                )
+                role_row = (await execute_rbac_entity_creator(db_session, rbac_creator)).row
+            else:
```
seedspirit, line 129: Can you add repository test verifying scope_ref, additional_scope_refs data is inserted in intended way?

HyeockJinKim (dismissed):
The code isn't very clean, so it looks like it will need to be refactored later.

PR #10711, opened 2026-04-01
fix(BA-5527): inherit source vfolder's usage_mode when cloning

jopemachine:
Resolves BA-5527.

## Summary
- When cloning a vfolder without explicitly specifying `usage_mode`, the clone now inherits the source vfolder's `usage_mode` instead of always defaulting to `GENERAL`
- Previously, `CloneVFolderReq.usage_mode` was hardcoded to `VFolderUsageMode.GENERAL`, so cloning a MODEL_STORE vfolder always produced a GENERAL clone

## Changes
- `CloneVFolderReq.usage_mode`: `VFolderUsageMode` → `VFolderUsageMode | None` (default `None`)
- `CloneVFolderAction.usage_mode`: similarly changed to Optional
- `VFolderService.clone()`: added fallback logic `action.usage_mode or source_vfolder_data.usage_mode`
- REST handler log output: handle `None` for usage_mode

## Test plan
- [ ] Clone a MODEL_STORE vfolder without specifying usage_mode → verify clone is MODEL_STORE
- [ ] Clone a GENERAL vfolder without specifying usage_mode → verify clone is GENERAL
- [ ] Clone with explicit usage_mode specified → verify it uses the specified value

### DIFF
--- a/tests/unit/manager/services/vfolder/test_vfolder_service.py
+++ b/tests/unit/manager/services/vfolder/test_vfolder_service.py
@@ -413,12 +413,20 @@ class TestVFolderServiceCloneUsageModeInheritance:
         "case",
         [
             # None → inherit from source
-            CloneUsageModeCase(VFolderUsageMode.MODEL, None, VFolderUsageMode.MODEL),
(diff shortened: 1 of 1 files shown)


jopemachine:
This changes existing behavior, so it is not suitable for backporting.

PR #10728, opened 2026-04-02
fix(BA-5557): register appproxy endpoints for deployments entering DEPLOYING via ActivateRevision

jopemachine:
## Summary
- Deployments entering DEPLOYING state via `ActivateRevision` (bypassing `check_pending`) were not getting registered with appproxy, causing them to stay DEGRADED
- `DeployingProvisioningHandler` now directly calls `register_endpoint` for unregistered deployments using `deploying_revision_id`
- Made `register_endpoint` public on `DeploymentExecutor` for cross-handler reuse
- Restored `check_pending_deployments` name for the pending handler path

## Test plan
- [ ] Verify deployment via `CreateDeployment → AddRevision → ActivateRevision` registers endpoint in appproxy
- [ ] Verify legacy deployment path (`check_pending`) still works unchanged
- [ ] Verify deployments that already have URLs are skipped (no double registration)

Fixes [BA-5557](https://lablup.atlassian.net/browse/BA-5557)

[BA-5557]: https://lablup.atlassian.net/browse/BA-5557?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ai/backend/manager/sokovan/deployment/executor.py
+++ b/src/ai/backend/manager/sokovan/deployment/executor.py
@@ -128,18 +128,10 @@ class DeploymentExecutor:
     ) -> DeploymentExecutionResult:
         """Register appproxy endpoints and persist their URLs.
 
-        For each ``(deployment, revision_id)`` entry, calls
-        ``register_endpoint`` in appproxy, then on success persists the
-        returned URL. Entries whose scaling group has no proxy target are
-        reported as ``skipped``.
-
-        Retry safety: appproxy's ``create_or_update_endpoint`` is
-        idempotent on ``endpoint_id`` — a second call with the same
-        ``endpoint_id`` returns the existing circuit's URL rather than
-        creating a duplicate — so if URL persistence fails after a
-        successful registration, the next coordinator tick will call
-        ``register_endpoint`` again, receive the same URL, and recover.
-        No compensating unregister is needed.
+        Entries without a proxy target are reported as ``skipped``.
+        Retry-safe: appproxy's endpoint create is idempotent on
+        ``endpoint_id``, so a failed URL persist is recovered on the
+        next tick.
         """
         if not entries:
             return DeploymentExecutionResult()
@@ -155,7 +147,7 @@ class DeploymentExecutor:
         failures: list[DeploymentExecutionError] = []
         skipped: list[DeploymentWithHistory] = []
 
-        async def _run(
+        async def _register_and_persist(
             deployment: DeploymentWithHistory,
             revision_id: UUID,
             target: ScalingGroupProxyTarget,
@@ -185,7 +177,9 @@ class DeploymentExecutor:
         if not tasks:
             return DeploymentExecutionResult(skipped=skipped)
 
-        results = await asyncio.gather(*(_run(dep, rev, tgt) for dep, rev, tgt in tasks))
+        results = await asyncio.gather(
+            *(_register_and_persist(dep, rev, tgt) for dep, rev, tgt in tasks)
+        )
 
         for deployment, url, error in results:
             dep_id = deployment.deployment_info.id
@@ -495,26 +489,11 @@ class DeploymentExecutor:
         scaling_group_target: ScalingGroupProxyTarget,
         revision_id: UUID,
     ) -> str:
-        """Resolve the target revision's model definition and register the
-        endpoint in appproxy.
+        """Register the deployment's endpoint in appproxy and return its URL.
 
-        Side effects:
-            Creates or updates an endpoint in appproxy (WSProxy). The
-            underlying ``POST /v2/endpoints/{endpoint_id}`` handler is
-            idempotent on ``endpoint_id``: if an endpoint/circuit already
-            exists, the existing URL is returned instead of a new circuit
-            being created. The caller is expected to persist the returned
-            URL to the ``endpoints`` table; if persistence fails, the
-            next coordinator tick can safely call this method again and
-            will receive the same URL.
-
-        Preconditions:
-            - ``revision_id`` must resolve to a revision on ``deployment``.
-            - ``scaling_group_target`` must be a valid proxy target for
-              ``deployment.metadata.resource_group``.
-
-        Returns:
-            The registered endpoint URL.
+        Idempotent on ``deployment.id``: repeated calls return the same URL
+        without creating a duplicate circuit. The caller is expected to
+        persist the returned URL to the ``endpoints`` table.
         """
         pool = DeploymentRecorderContext.current_pool()
         recorder = pool.recorder(deployment.id)
--- a/src/ai/backend/manager/sokovan/deployment/handlers/deploying.py
+++ b/src/ai/backend/manager/sokovan/deployment/handlers/deploying.py
@@ -136,19 +136,11 @@ class DeployingProvisioningHandler(DeploymentHandler):
     async def _ensure_endpoints_registered(
         self, deployments: Sequence[DeploymentWithHistory]
     ) -> set[UUID]:
-        """Register endpoints in appproxy for deployments that have no URL yet.
+        """Register endpoints for deployments that entered DEPLOYING via
+        ActivateRevision and therefore skipped ``check_pending``.
 
-        Deployments that entered DEPLOYING via ActivateRevision skip
-        check_pending (which normally registers them), so this method
-        ensures they are registered before route provisioning begins.
-
-        Delegates to ``DeploymentExecutor.register_endpoints_bulk`` so the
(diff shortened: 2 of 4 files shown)


jopemachine:
## Code Review

### CRITICAL
None.

### HIGH

**H1. Inconsistent state between appproxy and DB on partial failure**
`src/ai/backend/manager/sokovan/deployment/handlers/deploying.py` (`_execute_registration_batch` → `update_endpoint_urls_bulk`)
`register_endpoint` has a real side effect on appproxy, but if the subsequent `update_endpoint_urls_bulk(url_updates)` fails, the endpoint exists in appproxy while the DB still has an empty URL. On the next tick, `_ensure_endpoints_registered` will pick the same deployment up and register it again (double registration). The PR description's "already have URLs are skipped" guarantee does not hold on this failure path.
Recommendation: (a) best-effort unregister for successfully-registered endpoints on persistence failure, or (b) persist URLs inside `register_endpoint` as part of the same atomic step, or (c) use a `deploying_revision_id`-based idempotency key on the appproxy side.

**H2. Race condition when the same deployment is processed concurrently**
`handlers/deploying.py` `_ensure_endpoints_registered` (filter: `not url and deploying_revision_id`)
Overlapping coordinator ticks / retries may invoke `register_endpoint` twice for the same deployment. This new path also races against the existing `check_pending` path (both evaluate the same filter independently). The codebase has a `LockID` scheme, but this new path registers without acquiring any deployment-level lock.
Recommendation: guard `_ensure_endpoints_registered` with an appropriate deployment-level lock (or reuse existing `LockID.DEPLOYMENT_*`), or add a pre-check in `register_endpoint` that detects an already-registered revision.

### MEDIUM

**M1. Exceptions are swallowed without status / metric updates**
`handlers/deploying.py` `_execute_registration_batch` — `asyncio.gather(..., return_exceptions=True)` logs failures via `log.error` only, and `execute()` proceeds into route provisioning as if everything succeeded. Running subsequent steps with no URL will regress the deployment back to DEGRADED.
Recommendation: drop failed deployments from the list, or explicitly record them as retry/error (e.g. via `DeploymentRecorderContext` or a lifecycle sub-step). Mirror `check_pending_deployments` which reflects failures in `DeploymentExecutionResult`.

**M2. `_ensure_endpoints_registered` is on the happy path of `execute()` with no failure isolation**
`handlers/deploying.py` (top of `execute`)
If an exception escapes this step, the entire `execute()` fails and blocks route provisioning for every other deployment in the batch (batch latency amplification). `_ensure_endpoints_registered` itself uses `return_exceptions=True`, but calls around it like `fetch_scaling_group_proxy_targets` propagate exceptions directly.
Recommendation: wrap this in a `try/except` so only failing entries are isolated, or split retry policy by dependency.

**M3. Pre-built coroutines may trigger "coroutine was never awaited" warnings**
`handlers/deploying.py` `_build_registration_batch`
`coroutines.append(self._deployment_executor.register_endpoint(...))` materializes coroutine objects up-front. If anything in the build loop raises before `gather`, the already-created coroutines are never awaited → `RuntimeWarning: coroutine ... was never awaited`.
Recommendation: store argument tuples `(info, target, rev_id)` instead, and construct coroutines inside `_execute_registration_batch` immediately before `gather`. Alternatively wrap build in try/finally that `.close()`s unused coroutines.

**M4. Regression test coverage is limited to a single happy path**
`tests/unit/manager/sokovan/deployment/handlers/test_deploying_handler.py`
Missing cases:
- proxy target missing → skipped
- `register_endpoint` failure → `update_endpoint_urls_bulk` records only successes
- deployment already has URL → not registered (this is the 3rd item in the PR description's test plan)
- deployment with `deploying_revision_id is None` is filtered out

Recommendation: add the above 4 cases via `@pytest.mark.parametrize` or as separate tests.

**M5. Promoting `register_endpoint` to public duplicates logic**
`src/ai/backend/manager/sokovan/deployment/executor.py` `async def register_endpoint`
A previously private helper is now public; the new handler re-fetches `fetch_scaling_group_proxy_targets` itself, duplicating what `check_pending_deployments` already does.
Recommendation: lift the shared logic into a high-level public API on `DeploymentExecutor` (e.g. `register_endpoints_bulk(deployments)`) so both call sites use a single entry point and stay DRY.

### LOW

- **L1.** `_EndpointRegistrationBatch` dataclass is used in exactly one place; a `list[tuple[...]]` would do if no reuse is planned.
- **L2.** Log message style: check team convention (structured `extra=` vs. brace-style formatting).
- **L3.** `register_endpoint` is now public but lacks a contract docstring describing preconditions, side effects, and who is responsible for URL persistence.
- **L4.** `fetch_scaling_group_proxy_targets` is called from both paths on the same tick with no memoization (only worth addressing if profiling shows it).
- **L5.** `changes/10728.fix.md` is a one-liner; add the user-visible symptom ("deployments created via ActivateRevision remained DEGRADED") for release-note readers.
- **L6.** `mock_revision_spec` in the new test fixture is declared but never used — remove.

### Recommended order

1. **[HIGH]** H1 — atomicity / compensation between appproxy registration and URL persistence
2. **[HIGH]** H2 — prevent concurrent double-registration (lock or idempotency key)
3. **[MEDIUM]** M1 + M2 — isolate failing deployments from downstream route provisioning
4. **[MEDIUM]** M3 — switch to argument-tuple pattern to avoid unawaited coroutines
5. **[MEDIUM]** M5 — consolidate duplicated registration logic into a high-level executor API
6. **[MEDIUM]** M4 — add the 4 missing regression tests
7. **[LOW]** docstring, release-note wording, dead-code cleanup

jopemachine:
### Follow-up on H2 (concurrent double-registration)

After tracing the coordinator's locking scheme, H2 turns out to be a non-issue in the current architecture, so no code change was made for it. Details:

1. **Handler-level serialization is distributed.** `DeploymentCoordinator.process_deployment_lifecycle` acquires the handler's `lock_id` through `DistributedLockFactory` before calling `handler.execute()` (coordinator.py:347-351). `DeployingProvisioningHandler.lock_id == LOCKID_DEPLOYMENT_DEPLOYING (229)` is a distributed lock, so even across multiple manager processes, two DEPLOYING ticks cannot overlap on the same deployment.

2. **Cross-handler races touch disjoint deployment sets.** `CheckPendingHandler` and `DeployingProvisioningHandler` use different lock IDs (226 vs 229) and can run in parallel, but `get_deployments_for_handler` filters by `lifecycle_stages`:
   - CheckPending → `PENDING`, `CREATED`
   - DeployingProvisioning → `DEPLOYING` (sub_step `DEPLOYING_PROVISIONING`)

   A deployment has exactly one lifecycle state, so the two handlers select disjoint sets and cannot register the same deployment simultaneously.

3. **State transition mid-tick is safe.** Lifecycle transitions happen in `post_process` / transitions after `execute()`. By the time a deployment transitions from `PENDING` into `DEPLOYING`, the CheckPending registration has already persisted the URL — so the next tick's DeployingProvisioning handler filters it out via `if not info.network.url` in `_ensure_endpoints_registered`. If CheckPending's registration failed and the compensating unregister ran, the deployment stays in `PENDING` and never reaches DeployingProvisioning.

4. **BA-5557 scenario.** The "created without a revision → ActivateRevision → DEPLOYING" path skips CheckPending entirely and only DeployingProvisioning handles it, which is already single-flighted by its own distributed lock.

Given the above, adding extra locks or pre-check queries would only add complexity without closing a real race window. Marking H2 as resolved by existing architecture.

jopemachine:
### Follow-up: unbounded concurrency in \`register_endpoints_bulk\`

Noting this as a known limitation, not fixed in this PR because it predates the refactor and affects the pre-existing \`check_pending_deployments\` path too.

\`register_endpoints_bulk\` does \`asyncio.gather(*(_register_and_persist(...) for ...))\` with no concurrency cap. Combined with \`get_deployments_for_handler\`, which has no \`LIMIT\`, a single coordinator tick can in theory fan out to N simultaneous appproxy HTTP calls and N simultaneous DB \`UPDATE\`s, where N is the number of deployments waiting in the handler's target state. Under a large batch this could exhaust the HTTP client pool, the DB connection pool, or pressure appproxy.

In practice this hasn't surfaced because the realistic per-tick batch size is bounded by how many deployments sit in \`PENDING\` / \`DEPLOYING\` at once, but there is no explicit guard.

Follow-up (separate PR): bound via an \`asyncio.Semaphore\` sized to min(HTTP client pool, DB connection pool), or introduce an explicit batch-size LIMIT on the handler fetch.

PR #10766, opened 2026-04-03
feat(BA-5578): refactor ValkeyClient usage to async context manager pattern

jopemachine:
Resolves BA-5578.

## Summary
- Add `acquire()` async context manager to `AbstractValkeyClient` and `MonitoringValkeyClient` that wraps operations and tracks consecutive connection failures
- When the failure threshold is exceeded, the connection is torn down and reconnected rather than continuing to retry on a broken connection
- Migrate all 13 domain-specific Valkey clients to use the new `async with self._client.acquire() as conn:` pattern
- Add tests covering the retry-based disconnection logic

## Changes
- **Core** (`client.py`): Define `_VALKEY_CONNECTION_ERRORS` tuple, add `acquire()` base implementation to `AbstractValkeyClient`, override in `MonitoringValkeyClient` with `_operation_failure_count` tracking and threshold-based `_reconnect()` trigger
- **Domain clients** (13 files): Replace all `self._client.client.xxx()` calls with `async with self._client.acquire() as conn:` pattern
- **Tests**: Add `TestMonitoringValkeyClientAcquire` class with 8 test cases covering success/failure counting, threshold detection, reconnection, and error type tracking

## Test plan
- [x] `pants fmt/fix/lint --changed-since=HEAD` passes
- [x] `pants check --changed-since=HEAD` (mypy) passes
- [x] `pants test --changed-since=HEAD` passes
- [ ] Verify in Standalone mode
- [ ] Verify in Sentinel mode

### DIFF
--- a/tests/unit/common/clients/valkey_client/test_sentinel_valkey_client.py
+++ b/tests/unit/common/clients/valkey_client/test_sentinel_valkey_client.py
@@ -1,15 +1,17 @@
 from __future__ import annotations
 
 import asyncio
+from collections.abc import AsyncIterator
+from contextlib import asynccontextmanager
 from unittest.mock import AsyncMock, MagicMock
 
 import pytest
 from glide import GlideClient
 from glide.exceptions import ClosingError  # type: ignore[import-not-found]
 
-import ai.backend.common.clients.valkey_client.client as client_module
 from ai.backend.common.clients.valkey_client.client import (
     MonitoringValkeyClient,
+    MonitoringValkeyClientSpec,
     ValkeySentinelClient,
     ValkeySentinelTarget,
     ValkeyStandaloneClient,
@@ -77,12 +79,21 @@ def connected_sentinel_client(
 
 
 def _create_mock_sentinel_client() -> AsyncMock:
+    mock_glide = AsyncMock()
+    mock_glide.ping = AsyncMock(return_value=b"PONG")
+
     client = AsyncMock(spec=ValkeySentinelClient)
     client.need_reconnect = AsyncMock(return_value=False)
     client.ping = AsyncMock()
     client.connect = AsyncMock()
     client.disconnect = AsyncMock()
-    client.client = MagicMock()
+
+    @asynccontextmanager
+    async def mock_client_cm() -> AsyncIterator[AsyncMock]:
+        yield mock_glide
+
+    client.client = mock_client_cm
+    client._mock_glide = mock_glide
     return client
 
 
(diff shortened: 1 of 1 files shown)


src/ai/backend/common/clients/valkey_client/client.py
```diff
@@ -479,14 +524,46 @@ async def ping(self) -> None:
...
+        Acquire the operation client for use, tracking consecutive failures.
+
+        On connection errors, increments the failure counter. When the
+        consecutive failure threshold is exceeded, marks the client for
+        reconnection by the monitor loop rather than reconnecting directly.
+
+        On successful operations, resets the failure counter.
+        """
+        try:
+            yield self._operation_client._raw_client
```
fregataa: Should not refer a private field of _operation_client

src/ai/backend/common/clients/valkey_client/client.py
```diff
@@ -533,8 +610,17 @@ async def _monitor_connection(self) -> None:
...
-                    await asyncio.sleep(_DEFAULT_MONITOR_INTERVAL)
-                    if await self._check_connection():
+                    # Wait for either the monitor interval or an explicit reconnect request
+                    try:
+                        await asyncio.wait_for(
+                            self._reconnect_event.wait(),
+                            timeout=self._spec.monitor_interval,
+                        )
+                    except TimeoutError:
+                        pass
```
fregataa, line 622: Does it try to reconnect when waiting for `_reconnect_event` raises timeout?

src/ai/backend/common/clients/valkey_client/client.py
```diff
@@ -424,32 +469,32 @@ class MonitoringValkeyClient(AbstractValkeyClient):
...
+        spec: MonitoringValkeyClientSpec | None = None,
     ) -> None:
         self._operation_client = operation_client
         self._monitor_client = monitor_client
+        self._spec = spec or MonitoringValkeyClientSpec()
         self._monitor_task = None
-        self._reconnectable_exceptions = reconnectable_exceptions
-        self._consecutive_failure_threshold = consecutive_failure_threshold
+        self._needs_reconnect = False
+        self._reconnect_event = asyncio.Event()
```
fregataa: Is there any case:
- `_needs_reconnect` is False and `_reconnect_event` is set
- `_needs_reconnect` is True and `_reconnect_event` is not set

src/ai/backend/common/clients/valkey_client/client.py
jopemachine: No, regardless of whether a timeout occurs, _check_connection is invoked to detect ping failures, maintaining the same behavior as before.

src/ai/backend/common/clients/valkey_client/client.py
jopemachine: _needs_reconnect is a flag that was mistakenly left in and not removed in the previous commit.

PR #10768, opened 2026-04-03
fix: Sort occupied resource slots by rank from resource_slot_types table

### LINKED ISSUE #10767: Inconsistent resource slot ordering in 'Occupied Slots' column of agent list
## Summary

The `backend.ai admin agent list` command displays resource slots in an inconsistent order in the **Occupied Slots** column. While **Available Slots** always shows a consistent order (`cpu`, `mem`, `cuda.shares`), the **Occupied Slots** column can vary per row.

## Observed Behavior

```
Available Slots                      Occupied Slots
-----------------------------------  ------------------------------------
cpu:222, mem:2.0T, cuda.shares:8.00  cpu:0, mem:0B, cuda.shares:0
cpu:222, mem:2.0T, cuda.shares:8.00  mem:960.2G, cpu:49, cuda.shares:4.00   ← mem before cpu
cpu:222, mem:2.0T, cuda.shares:8.00  cpu:0, mem:0B, cuda.shares:0
```

Notice on the second agent, **Occupied Slots** starts with `mem` instead of `cpu`, breaking the expected `cpu → mem → cuda.shares` order that **Available Slots** consistently uses.

## Expected Behavior

Both columns should display resource slots in a consistent, deterministic order (e.g., `cpu, mem, cuda.shares`).

## Root Cause Analysis

The issue originates from the `actual_occupied_slots()` method in `src/ai/backend/manager/models/agent/row.py`:

```python
agent_resource_rows: Mapped[list[AgentResourceRow]] = relationship("AgentResourceRow")

def actual_occupied_slots(self) -> ResourceSlot:
    occupied = ResourceSlot()
    for resource_row in self.agent_resource_rows:
        occupied[resource_row.slot_name] = resource_row.used
    return occupied
```

The `agent_resource_rows` relationship has **no `order_by` clause**, so SQLAlchemy returns rows in whatever order the database provides — which is non-deterministic.

In contrast, `available_slots` is stored as a JSON column where key order is preserved from insertion time, so it appears consistent.

The downstream formatter (`ResourceSlotFormatter` in `src/ai/backend/client/output/formatters.py`) simply iterates `dict.items()` without sorting:

```python
return ", ".join(f"{k}:{v}" for k, v in value.items())
```

## Possible Fixes

1. **Sort keys in the formatter** — normalize display order in `ResourceSlotFormatter.format_console()` (client-side fix, covers all slot columns)
2. **Add `order_by` to the relationship** — ensure `agent_resource_rows` are always loaded in a consistent order (server-side fix)
3. **Sort in `ResourceSlot.to_json()`** — normalize at the serialization boundary (common layer fix, broadest impact)

Option 1 or 3 would be the safest and most broadly effective.

rapsealk:
## Summary

- Sort occupied resource slots by `resource_slot_types.rank` from the database instead of using a client-side hardcoded sort
- Add `slot_type_row` relationship on `AgentResourceRow` and eagerly load it at all query sites
- `actual_occupied_slots()` now sorts `agent_resource_rows` by `slot_type_row.rank`

## Related

- Closes #10767

## Changes

**Server-side (`manager/`):**
- `models/resource_slot/row.py`: Added `slot_type_row` relationship on `AgentResourceRow` → `ResourceSlotTypeRow` (with `lazy="raise"` to prevent accidental lazy loads)
- `models/agent/row.py`: `actual_occupied_slots()` now sorts rows by `slot_type_row.rank` before building the `ResourceSlot`
- `repositories/agent/repository.py`, `repositories/agent/db_source/db_source.py`, `api/gql_legacy/agent.py`: Chain `.joinedload(AgentResourceRow.slot_type_row)` on all `selectinload(AgentRow.agent_resource_rows)` calls

**Client-side (`client/`):**
- `output/formatters.py`: Reverted to plain `value.items()` iteration — ordering is now guaranteed by the server

## Test plan

- [ ] Run `backend.ai admin agent list` and verify Occupied Slots matches the rank order defined in `resource_slot_types`
- [ ] Verify Available Slots column order is unaffected
- [ ] Verify agents with zero-valued slots maintain consistent ordering


tests/unit/manager/models/resource_slot/test_resource_slot_tables.py
```diff
@@ -473,3 +475,117 @@ async def test_numeric_precision_large_memory_allocation(
             assert row is not None
             assert row.requested == one_tib
             assert row.used == one_tib
+
+
+class TestActualOccupiedSlotsOrdering:
```
seedspirit, line 480: What is the difference between the two tests below? At first glance, it appears that both sort the `rank` values in ascending order and display the `slot_keys`. If they are identical, they could likely be grouped together using `parametrize` or similar.

Additionally, by creating a fixture factory or fixture repository to inject sample data into the database, it seems possible to keep only the parts necessary for testing within the test methods themselves.

jopemachine:
This PR appears to be closer to a feature than a fix.

tests/unit/manager/models/resource_slot/test_resource_slot_tables.py
rapsealk: Good catch — consolidated the two tests into a single `@pytest.mark.parametrize`d test in b84426a0e. The two cases differ only in rank assignments and expected ordering, so parametrize fits well here. Also simplified the resource row setup since the test only cares about key ordering, not actual capacity/used values.

rapsealk:
Thanks for the feedback @jopemachine. I'd argue `fix:` is appropriate here — the prior behavior of `actual_occupied_slots()` returned slots in non-deterministic order (dependent on DB row insertion/retrieval order), which could produce inconsistent output across calls. Sorting by `resource_slot_types.rank` fixes that non-determinism rather than adding new capability. That said, happy to change to `feat:` if you still feel it's more appropriate.

HyeockJinKim (approved):

PR #10777, opened 2026-04-03
fix: Validate OpenID plugin config in init() to prevent startup crash

### LINKED ISSUE #10775: OpenID plugin crashes manager server on startup when not configured
## Bug Description

The OpenID plugin (`OIDCWebAppPlugin` and `OIDCHookPlugin`) crashes the manager server during startup when no OpenID configuration exists in etcd. This is the common case for local development environments.

The crash occurs because:
- `OIDCWebAppPlugin._webapp_init()` unconditionally accesses `self.plugin_config["openid"]` (line 239), raising `KeyError: 'openid'`
- `OIDCHookPlugin.pre_auth_hook()` accesses `self.plugin_config["secret"]` (line 49) without validation
- Neither plugin validates required configuration in their `init()` method

The `_webapp_init` crash is particularly problematic because it occurs during aiohttp application startup, **outside** the plugin framework's try/except handler that normally catches and gracefully skips misconfigured plugins.

## Steps to Reproduce

1. Have a local dev environment without OpenID configured in etcd
2. Run `./backend.ai mgr start-server --debug`
3. Server crashes with:
```
KeyError: 'openid'
```

## Expected Behavior

The OpenID plugin should validate its configuration in `init()` and raise an error there, so the plugin framework can gracefully skip it (like the overlay network plugin does).

## Actual Behavior

Server crashes during aiohttp startup with `KeyError: 'openid'`.

## Fix

Add configuration validation in `init()` for both `OIDCWebAppPlugin` and `OIDCHookPlugin` so missing config is caught early by the plugin framework's error handler.

## Related PRs

- https://github.com/lablup/backend.ai/pull/10762

rapsealk:
## Summary

- Add config validation in `OIDCWebAppPlugin.init()` to check for required `openid` key
- Add config validation in `OIDCHookPlugin.init()` to check for required `secret` key
- This allows the plugin framework's try/except handler to gracefully skip unconfigured plugins, matching the pattern used by other plugins (e.g., overlay network)

Previously, the server crashed with `KeyError: 'openid'` during aiohttp startup when no OpenID config existed in etcd (common in local dev).

Fixes #10775

## Related PRs

- #10762 (OpenID migration — introduced the bug)
- #10776 (BUILD duplicate keys fix — companion fix)

## Test plan

- [ ] Start manager without OpenID config in etcd — should skip plugin gracefully with an error log
- [ ] Start manager with OpenID config in etcd — should initialize normally

### DIFF
--- a/src/ai/backend/manager/plugin/openid/hook.py
+++ b/src/ai/backend/manager/plugin/openid/hook.py
@@ -28,12 +28,16 @@ class OIDCHookPlugin(HookPlugin):
     #   - plugin_config is populated from the "/config/plugins/hook/openid/" etcd key
     _config: OIDCPluginConfig
 
+    def __init__(self, plugin_config: Mapping[str, Any], local_config: Mapping[str, Any]) -> None:
+        super().__init__(plugin_config, local_config)
+        self._config = OIDCPluginConfig(**plugin_config)
+
     async def update_plugin_config(self, plugin_config: Mapping[str, Any]) -> None:
         self.plugin_config = plugin_config
         self._config = OIDCPluginConfig(**plugin_config)
 
     async def init(self, context: Any = None) -> None:
-        self._config = OIDCPluginConfig(**self.plugin_config)
+        pass
 
     async def cleanup(self) -> None:
         pass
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/plugin/openid/hook.py
```diff
@@ -29,7 +29,11 @@ async def update_plugin_config(self, plugin_config: Mapping[str, Any]) -> None:
         self.plugin_config = plugin_config
 
     async def init(self, context: Any = None) -> None:
-        pass
+        if "secret" not in self.plugin_config:
```
jopemachine: Could we define and use self.plugin_config as a Pydantic model?

PR #10794, opened 2026-04-05
feat(BA-5515): add scoped role search API

fregataa:
## Summary
- Add GQL resolver (`projectRoles`), REST v2 endpoint (`POST /v2/rbac/roles/projects/{project_id}/search`), SDK method (`project_search_roles`), and CLI commands (`rbac role project-search`, `project role search`) for searching roles registered in a scope
- Add `ScopedRoleSearchScope` (using `RBACElementType`) with proper adapter method (`search_roles_in_scope`) replacing the `base_conditions` injection pattern
- Add `SearchRolesInScopeAction` extending `BaseScopeAction` with `ScopeActionProcessor` and `ScopeActionRBACValidator` to enforce scope-level permission checks
- Remove unused `RoleConditions.by_registered_in_scope()` and unnecessary `ObjectPermissionRow` JOIN from scoped role search query
- Add component tests and db_source unit tests covering scope filtering, scope type isolation, and empty scope scenarios

## Test plan
- [x] Component tests pass (`pants test tests/component/rbac/test_rbac_scoped_role_search.py`)
- [x] Unit tests pass (`pants test tests/unit/manager/repositories/permission_controller/test_search_roles_in_scope.py`)
- [x] GQL `projectRoles(projectId: UUID!)` returns only roles registered in the given project
- [x] REST `POST /v2/rbac/roles/projects/{project_id}/search` works with filter/order/pagination
- [x] CLI `./bai rbac role project-search {project_id}` returns expected results
- [x] CLI `./bai project role search {project_id}` returns expected results
- [x] Roles not registered in the project scope are excluded from results
- [x] RBAC validation rejects users without project-scope READ permission

Resolves BA-5515

----
📚 Documentation preview 📚: https://sorna--10794.org.readthedocs.build/en/10794/

----
📚 Documentation preview 📚: https://sorna-ko--10794.org.readthedocs.build/ko/10794/

### DIFF
--- a/tests/component/rbac/test_rbac_scoped_role_search.py
+++ b/tests/component/rbac/test_rbac_scoped_role_search.py
@@ -52,7 +52,9 @@ def permission_controller_processors(
     database_engine: ExtendedAsyncSAEngine,
 ) -> PermissionControllerProcessors:
     repo = PermissionControllerRepository(database_engine)
-    service = PermissionControllerService(repo, rbac_action_registry=[])
+    service = PermissionControllerService(
+        repo, group_repository=MagicMock(), rbac_action_registry=[]
+    )
     validators = MagicMock()
     validators.rbac.scope.validate = AsyncMock()
     return PermissionControllerProcessors(


src/ai/backend/manager/api/gql/rbac/resolver/role.py
```diff
@@ -208,6 +210,55 @@ async def my_roles(
...
+    filter: RoleFilter | None = None,
+    order_by: list[RoleOrderBy] | None = None,
+    before: str | None = None,
+    after: str | None = None,
+    first: int | None = None,
+    last: int | None = None,
+    limit: int | None = None,
+    offset: int | None = None,
+) -> RoleConnection:
+    result = await info.context.adapters.rbac.admin_search_roles_gql(
```
jopemachine: `info.context.adapters.rbac` is an adapter used not only by the GQL layer but also by the REST API. Why does it include a method like admin_search_roles_gql, which appears to be GQL-specific?

fregataa: The _gql suffix on adapter methods and GQL in shared DTO names (AdminSearchRolesGQLInput, etc.) was incorrectly carried over from the Strawberry type naming convention (GQL suffix rule), which should only apply to GraphQL types, not to shared adapters/DTOs. I'll address this in a separate issue to rename all affected adapter methods and DTO classes (removing the GQL suffix) #10809

src/ai/backend/client/cli/v2/rbac/role.py
```diff
@@ -92,6 +92,79 @@ async def _run() -> None:
...
+        RoleStatusFilter,
+    )
+
+    filter_dto: RoleFilter | None = None
+    if any([name_contains is not None, source is not None, status is not None]):
+        filter_dto = RoleFilter(
+            name=StringFilter(contains=name_contains) if name_contains is not None else None,
+            source=RoleSourceFilter(equals=source) if source is not None else None,
+            status=RoleStatusFilter(equals=status) if status is not None else None,
+        )
```
jopemachine, line 145: Has the usage of filters in this form in the v2 CLI been discussed and established as a convention?

src/ai/backend/manager/api/gql/rbac/resolver/role.py
```diff
@@ -208,6 +210,55 @@ async def my_roles(
...
+            first=first,
+            after=after,
+            last=last,
+            before=before,
+            limit=limit,
+            offset=offset,
+        ),
+        base_conditions=[
+            RoleConditions.by_registered_in_scope(ScopeType.PROJECT, str(project_id)),
+        ],
```
HyeockJinKim: don't use this base conditions.

src/ai/backend/manager/api/gql/rbac/resolver/role.py
HyeockJinKim: It seems a bit odd that "admin" and "gql" are included in the adapter. Could you please check this?
It’s fine to handle "gql" later, but since it’s not for "admin," please review whether separate action is required.

src/ai/backend/client/v2/domains_v2/rbac.py
```diff
@@ -56,7 +56,7 @@ async def create_role(self, request: CreateRoleInput) -> CreateRolePayload:
             response_model=CreateRolePayload,
         )
 
-    async def search_roles(self, request: AdminSearchRolesGQLInput) -> AdminSearchRolesPayload:
```
seedspirit, line 59: Though "Admin" prefix removed from the request, but it remains in the payload. Is this design intended?

src/ai/backend/client/cli/v2/project/commands.py
```diff
@@ -29,3 +34,84 @@ async def _run() -> None:
...
+@role.command(name="search")
+@click.argument("project_id", type=click.UUID)
+@click.option("--limit", type=int, default=None, help="Maximum items to return.")
+@click.option("--offset", type=int, default=None, help="Number of items to skip.")
+@click.option(
+    "--order-by",
+    multiple=True,
+    help="Order by field:direction (e.g., name:asc, created_at:desc).",
+)
+@click.option("--name-contains", type=str, default=None, help="Filter roles by name (contains).")
```
jopemachine, line 56: I’m wondering whether there has been any discussion around which filters should be supported or excluded when accepting filter arguments in the CLI.

For example, is there a specific reason why --name-contains is supported while something like --name-starts-with is not?

fregataa: I assume that it is because contains can support all other patterns and can be used generally

jopemachine:
CI is broken

jopemachine:
It would be better to explicitly include “project-scoped” in the PR title, but since this is not critical, I will proceed with merging.

PR #10797, opened 2026-04-05
fix(BA-5520): use target ownership policy for VFolder clone max_vfolder_count check

fregataa:
## Summary
- Fix `clone()` and `clone_v2()` methods to check the user's resource policy (not the source project's policy) when validating max_vfolder_count during VFolder cloning
- Previously, cloning a project-owned VFolder incorrectly passed `source_vfolder_data.group` to `get_max_vfolder_count`, which checked the source project's resource policy instead of the requester's user resource policy
- Add regression tests verifying `get_max_vfolder_count` is called with `(user_uuid, None)` when cloning project-owned VFolders

## Test plan
- [x] Unit tests verify `get_max_vfolder_count` is called with correct arguments `(user_uuid, None)` for both `clone()` and `clone_v2()`
- [x] CI passes lint and type checks
- [x] CI passes test suite

Resolves BA-5520

### DIFF
--- a/tests/component/vfolder/test_vfolder_clone.py
+++ b/tests/component/vfolder/test_vfolder_clone.py
@@ -1,0 +1,367 @@
+"""Component tests for VFolder clone policy enforcement.
+
+Verifies that clone operations check the *user's* resource policy for
+max_vfolder_count, not the source project's resource policy (BA-5520).
+
+Storage proxy is mocked; the database is real.
+"""
+
+from __future__ import annotations
+
+import uuid
+from collections.abc import AsyncIterator, Callable, Coroutine
+from typing import TYPE_CHECKING, Any
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+import sqlalchemy as sa
+import yarl
+from sqlalchemy.ext.asyncio.engine import AsyncEngine as SAEngine
+
+from ai.backend.client.exceptions import BackendAPIError
+from ai.backend.client.v2.auth import HMACAuth
+from ai.backend.client.v2.config import ClientConfig
+from ai.backend.client.v2.registry import BackendAIClientRegistry
+from ai.backend.client.v2.v2_registry import V2ClientRegistry
+from ai.backend.common.bgtask.types import TaskID
+from ai.backend.common.dto.manager.v2.vfolder.request import CloneVFolderInput
+from ai.backend.common.dto.manager.vfolder import CloneVFolderReq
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/services/vfolder/services/vfolder.py
```diff
@@ -1681,8 +1681,9 @@ async def clone_v2(self, action: CloneVFolderV2Action) -> CloneVFolderV2ActionRe
             user_uuid=action.user_id,
         )
 
+        # clone_v2 always creates user-owned vfolders, so check user resource policy (group_uuid=None)
```
jopemachine, line 1684: If, as the comment suggests, group_uuid must always be passed as None, why is this parameter needed? If it is not actually required, it would be better to remove it.

seedspirit: This is because the logic is designed such that when `get_max_vfolder_count` receives a project ID (group_id), it first validates `group_row.resource_policy_row.max_vfolder_count`, and only if the project ID is `None` does it proceed to validate the user’s resource policy.

In other words, since this bug fundamentally stems from the fact that `get_max_vfolder_count` is performing two roles, it would be best to either separate these functions or (since there is no use case for this method to verify the max vfolder count for a group) modify the function to simply retrieve the user’s max vfolder count.

fregataa: it is a fix task not a refactoring, I'd like to keep this work as small as possible

fregataa: > This is because the logic is designed such that when `get_max_vfolder_count` receives a project ID (group_id), it first validates `group_row.resource_policy_row.max_vfolder_count`, and only if the project ID is `None` does it proceed to validate the user’s resource policy.
> 
> In other words, since this bug fundamentally stems from the fact that `get_max_vfolder_count` is performing two roles, it would be best to either separate these functions or (since there is no use case for this method to verify the max vfolder count for a group) modify the function to simply retrieve the user’s max vfolder count.

get_max_vfolder_count serving dual roles (user vs project policy) is indeed the root cause that made this bug possible. Separating these responsibilities would prevent similar mistakes in the future.

That said, a bug fix PR should stay minimal and focused. Mixing a refactoring into a fix increases review scope and rollback risk. The right approach is to merge this fix as-is and track the refactoring as a follow-up task.

PR #10802, opened 2026-04-06
fix(BA-5593): add superadmin permission check to resource preset GQL queries

jopemachine:
## Summary
- Add `@privileged_query(UserRole.SUPERADMIN)` to `resolve_resource_preset`, `resolve_resource_preset_by_id`, and `resolve_resource_presets` in the legacy GraphQL schema
- These resolvers previously had no permission check, allowing any authenticated user to list all resource presets
- The v2 Strawberry resolver already enforces `check_admin_only()` on all resource preset endpoints — this fix aligns the legacy resolver

## Test plan
- [x] Regular user calling `resource_presets` query receives `"Forbidden operation. (superadmin privilege required)"` error
- [x] Regular user calling `resource_preset(name: "...")` query receives the same error
- [x] Superadmin can still list and retrieve resource presets normally

Resolves BA-5593

### DIFF
--- a/changes/10802.fix.md
+++ b/changes/10802.fix.md
@@ -1,0 +1,1 @@
+Add superadmin permission check to legacy resource preset GraphQL queries (BA-5593)


jopemachine:
Closing this PR — after further investigation, `resource_presets` is intentionally accessible to all authenticated users.

**Evidence:**
- The REST API `GET /resource/presets` uses `@auth_required` only (not `@superadmin_required`), and is documented as a user-facing API (`docs/manager/user-api/resource-presets.rst`)
- The WebUI `ResourcePresetSelect` component calls this GQL query in the session creation form (`ResourceAllocationFormItems.tsx`) — restricting it to superadmin would break session creation for regular users
- Only the GQL mutations (create/modify/delete) are restricted to `SUPERADMIN` via `allowed_roles`, which is correct

Adding `@privileged_query(UserRole.SUPERADMIN)` here would be a regression, not a fix.

Issue #10800, opened 2026-04-06
Regular users can view full agent infrastructure details via `agent_nodes` GraphQL query


jopemachine:
## Problem

The legacy agent_nodes resolver (gql_legacy/schema.py line 1422) uses RBAC via get_permission_ctx() with AgentPermission.CREATE_COMPUTE_SESSION as the default permission. Because regular users are granted this permission through their default RBAC role, the permission context builder generates a non-null query_condition, and the full agent list is returned to them.

## Verified Behavior

Tested with a regular user (user@lablup.com). The user receives the exact same agent data as the superadmin, including:

- addr: internal network address (e.g. tcp://127.0.0.1:6011)
- available_slots: full resource capacity (CPU cores, memory, CUDA devices)
- occupied_slots: current resource utilization
- scaling_group, region, architecture

This exposes server infrastructure details (IP addresses, hardware specs, utilization) to non-admin users.

## Root Cause

The RBAC permission AgentPermission.CREATE_COMPUTE_SESSION is too broadly granted. While users need to create compute sessions, they should not need direct visibility into agent-level infrastructure details. The older resolve_agents (line 1272) and resolve_agent_list (line 1287) are properly protected with @privileged_query(UserRole.SUPERADMIN), but agent_nodes bypasses this via the RBAC permission model.

## Acceptance Criteria

- Regular users should not be able to retrieve agent infrastructure details (addr, available_slots, occupied_slots, scaling_group, region)
- Either restrict agent_nodes to admin roles, or apply field-level filtering so non-admin users only see non-sensitive fields
- Verify that the v2 Strawberry agent resolver (gql/agent/resolver.py) has the same issue and fix if needed

JIRA Issue: BA-5594

PR #10803, opened 2026-04-06
fix(BA-5594): restrict agent_nodes and agents_v2 queries to superadmin only

jopemachine:
## Summary
- Regular users could access full agent infrastructure details (addr, available_slots, occupied_slots, scaling_group, region) via `agent_nodes` and `agents_v2` GraphQL queries
- Legacy `agent_nodes`: Added superadmin role check in `AgentNode.get_connection()` — returns empty results for non-superadmin (consistent with existing `resolve_agents`/`resolve_agent_list` which use `@privileged_query(UserRole.SUPERADMIN)`)
- V2 `agents_v2`: Added `check_admin_only()` guard
- `agent_summary_list` is unaffected — it's the intentional non-admin schema that excludes sensitive fields (no `addr`, `region`, `hardware_metadata`)

## Test plan
- [ ] Query `agent_nodes` as regular user → should return empty results
- [ ] Query `agent_nodes` as superadmin → should work as before
- [ ] Query `agents_v2` as regular user → should return 403
- [ ] Query `agents_v2` as superadmin → should work as before
- [ ] Verify `agent_summary_list` still works for regular users
- [ ] Verify WebUI dashboard displays correctly for both admin and non-admin users

Resolves BA-5594 (#10800)

### DIFF
--- a/tests/unit/manager/api/gql/agent/test_agents_v2_admin_only.py
+++ b/tests/unit/manager/api/gql/agent/test_agents_v2_admin_only.py
@@ -1,0 +1,68 @@
+"""
+Regression test: agents_v2 must be restricted to superadmin only. (BA-5594)
+"""
+
+from __future__ import annotations
+
+import uuid
+from enum import Enum, auto
+from typing import NamedTuple
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+from aiohttp import web
+
+from ai.backend.common.contexts.user import with_user
+from ai.backend.common.data.user.types import UserData
+from ai.backend.manager.api.gql.agent.resolver import agents_v2
+from ai.backend.manager.models.user import UserRole
+
+
+class ExpectedResult(Enum):
+    ALLOWED = auto()
+    FORBIDDEN = auto()
+
+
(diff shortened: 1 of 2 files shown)


HyeockJinKim (dismissed):

src/ai/backend/manager/api/gql_legacy/agent.py
```diff
@@ -252,6 +252,8 @@ async def get_connection(
         last: int | None = None,
     ) -> ConnectionResolverResult[AgentNode]:
         graph_ctx: GraphQueryContext = info.context
+        if graph_ctx.user["role"] != UserRole.SUPERADMIN:
+            return ConnectionResolverResult([], None, None, None, 0)
```
fregataa, line 257: Could you check whether it needs to support `hide-agent` option?

jopemachine: Since this guard restricts the query to SUPERADMIN unconditionally, I think we don't need to additionally check the hide_agents option here. 

* hide_agents: hides agents from non-admin (regular) users
* This PR: shows agents only to admin

PR #10811, opened 2026-04-06
fix(BA-5519): read container net stats from /proc/[pid]/net/dev with netns fallback (#10696)

lablup-octodog:
This is an auto-generated backport PR of #10696 to the 26.2 release.


jopemachine:
Please check this unmerged backport PR @fregataa

PR #10820, opened 2026-04-06
feat(BA-5599): extend assign_role API with optional project_id

fregataa:
## Summary
- Add optional `project_id` to `AssignRoleInput` / `BulkAssignRoleInput` (REST v2, GQL)
- When `project_id` is provided, the service layer calls `GroupRepository.bind_user_to_project()` to create `association_groups_users` + user-project RBAC scope binding alongside the role assignment
- On `revoke_role`, `PermissionDBSource` returns `RoleRevocationResult` containing per-project remaining role counts (single query via `association_scopes_entities` JOIN `user_roles`); the service layer calls `GroupRepository.unbind_user_from_project()` for projects where count reaches 0
- New `GroupDBSource.bind_user_to_project()` / `unbind_user_from_project()` methods for project membership management
- Service layer orchestrates cross-repository calls (`PermissionControllerRepository` + `GroupRepository`) — db_source layers remain single-responsibility

### Changed files
| Layer | Files |
|-------|-------|
| DTO | `common/dto/manager/v2/rbac/request.py` |
| GQL | `api/gql/rbac/types/role.py` |
| Data | `data/permission/role.py` (`ProjectRoleCount`, `RoleRevocationResult`, `UserRoleAssignmentInput.project_id`) |
| Adapter | `api/adapters/rbac.py` |
| Action | `services/permission_contoller/actions/bulk_assign_role.py` |
| Service | `services/permission_contoller/service.py`, `services/factory.py` |
| Repository | `repositories/permission_controller/{repository,db_source}.py`, `repositories/group/{repository,db_source}.py` |

## Test plan
- [x] 6 db-level tests (`test_role_assignment.py`): revoke remaining count, bind/unbind idempotency
- [x] 7 service-level tests (`test_role_assignment_service.py`): assign/revoke/bulk with mock repositories
- [x] All existing permission controller tests pass (fixtures updated for new constructor)
- [x] `pants fmt`, `fix`, `lint` pass

Resolves BA-5599

----
📚 Documentation preview 📚: https://sorna--10820.org.readthedocs.build/en/10820/

----
📚 Documentation preview 📚: https://sorna-ko--10820.org.readthedocs.build/ko/10820/

### DIFF
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -1303,6 +1303,7 @@ input AssignRoleInput
 {
   userId: UUID!
   roleId: UUID!
+  projectId: UUID
 }
 
 """Added in 24.03.9."""
@@ -1791,6 +1792,7 @@ input BulkAssignRoleInput
 {
   roleId: UUID!
   userIds: [UUID!]!
+  projectId: UUID
 }
 
 """Added in 26.3.0. Payload for bulk role assignment mutation."""
--- a/docs/manager/graphql-reference/v2-schema.graphql
+++ b/docs/manager/graphql-reference/v2-schema.graphql
@@ -954,6 +954,7 @@ type ArtifactVerifierResultEntry {
 input AssignRoleInput {
   userId: UUID!
   roleId: UUID!
+  projectId: UUID
 }
 
 """Added in 24.09.0. Filter criteria for querying audit logs."""
@@ -1205,6 +1206,7 @@ type BulkAssignRoleError {
 input BulkAssignRoleInput {
   roleId: UUID!
   userIds: [UUID!]!
+  projectId: UUID
 }
 
 """Added in 26.3.0. Payload for bulk role assignment mutation."""


src/ai/backend/manager/repositories/group/db_source/db_source.py
```diff
@@ -729,6 +729,47 @@ async def unassign_users_from_project(
...
+        """
+        async with self._db.begin_session() as session:
+            already_bound = await session.scalar(
+                sa.select(AssocGroupUserRow.user_id).where(
+                    AssocGroupUserRow.user_id == user_id,
+                    AssocGroupUserRow.group_id == project_id,
+                )
+            )
+            if already_bound is not None:
+                return
```
seedspirit, line 745: Rather than checking if it already exists, it seems like we could use an `upsert`. Is the reason we aren't using upsert because it's difficult to handle upsert operations within the `execute_rbac_scope_binder` function?

src/ai/backend/manager/repositories/group/db_source/db_source.py
```diff
@@ -729,6 +729,47 @@ async def unassign_users_from_project(
...
+                    AssocGroupUserRow.user_id == user_id,
+                    AssocGroupUserRow.group_id == project_id,
+                )
+            )
+            await session.execute(
+                sa.delete(AssociationScopesEntitiesRow).where(
+                    AssociationScopesEntitiesRow.scope_type == ScopeType.PROJECT,
+                    AssociationScopesEntitiesRow.scope_id == str(project_id),
+                    AssociationScopesEntitiesRow.entity_type == EntityType.USER,
+                    AssociationScopesEntitiesRow.entity_id == str(user_id),
```
seedspirit, line 769: nitpick: using `sa.and_ will` will be enhance readability

src/ai/backend/manager/repositories/permission_controller/repository.py
```diff
@@ -160,11 +161,15 @@ async def assign_role(self, data: UserRoleAssignmentInput) -> UserRoleAssignment
         return result.to_data()
 
     @permission_controller_repository_resilience.apply()
-    async def revoke_role(self, data: UserRoleRevocationInput) -> UserRoleRevocationData:
-        user_role_id = await self._db_source.revoke_role(data)
+    async def revoke_role(
+        self, data: UserRoleRevocationInput
+    ) -> tuple[UserRoleRevocationData, list[ProjectRoleCount]]:
```
seedspirit: rather than using tuple, how about  re-using `RoleRevocationResult` again here?

src/ai/backend/manager/services/permission_contoller/actions/bulk_assign_role.py
```diff
@@ -15,6 +16,7 @@
 @dataclass
 class BulkAssignRoleAction(RoleAction):
     bulk_creator: BulkCreator[UserRoleRow]
+    project_id: uuid.UUID | None = field(default=None)
```
seedspirit, line 19: nitpick: 
```suggestion
    project_id: UUID | None = field(default=None)
```

src/ai/backend/manager/services/permission_contoller/service.py
```diff
@@ -181,23 +187,40 @@ async def purge_role(self, action: PurgeRoleAction) -> PurgeRoleActionResult:
...
+
+        When project_id is provided, also binds the user to the project.
         """
-        Assigns a role to a user.
-        """
+        if action.input.project_id is not None:
+            await self._group_repository.bind_user_to_project(
+                action.input.user_id, action.input.project_id
+            )
         data = await self._repository.assign_role(action.input)
```
seedspirit, line 199: Since these operations seem like they should be grouped within the same transaction, wouldn't it be better to handle user project binding and unbinding within the `PermissionControllerRepository`?

seedspirit: also same with in `revoke_role` and `bulk_assign_role`

seedspirit (changes requested):

src/ai/backend/manager/repositories/group/db_source/db_source.py
fregataa: Calling execute_rbac_scope_binder() directly without the check would raise IntegrityError on the business row (association_groups_users) since execute_bulk_creator does a plain INSERT — only the RBAC row side uses on_conflict_do_nothing. The exists check guards against this.

src/ai/backend/manager/services/permission_contoller/service.py
fregataa: This is intentional. The service layer orchestrates the two repositories separately so that the association_groups_users management can be deprecated and removed later without modifying the permission controller internals. Keeping them decoupled at the service layer makes that future separation straightforward.

seedspirit (approved):
LGTM

PR #10821, opened 2026-04-06
feat(BA-5504): add updated_at column to vfolders

seedspirit:
## Summary

- Add `updated_at` column to the `vfolders` table; existing rows are backfilled with `COALESCE(last_used, created_at)` and the column is then set to `NOT NULL` with `server_default = now()`
- The column is automatically refreshed on every ORM-issued UPDATE via SQLAlchemy `onupdate=sa.func.now()`
- After in-place ORM mutations (`move_vfolders_to_trash`, `restore_vfolders_from_trash`), the row is `session.refresh(..., attribute_names=["updated_at"])` to surface the new value to callers — same pattern as `artifact/db_source/db_source.py:409`

## Out of scope

Tracking storage-side content mutations (upload / rename / move / delete / mkdir) is intentionally NOT covered here. That work will be done in a follow-up via a dedicated `content_updated_at` column and event-based refresh, instead of wrapping every call site in a context manager.

## Test plan

- [x] `pants test` for vfolder unit / repository / component / integration suites
- [x] Migration validated locally: `alembic upgrade` + `downgrade` + `upgrade` cycle; existing rows backfilled with `COALESCE(last_used, created_at)`; column is `NOT NULL` with `now()` default
- [ ] CI passes

Resolves BA-5504


src/ai/backend/manager/repositories/vfolder/repository.py
```diff
@@ -393,6 +393,20 @@ async def update_vfolder_attribute(self, updater: Updater[VFolderRow]) -> VFolde
...
+        Set updated_at to the current timestamp on a VFolder.
+        Called when vfolder contents are mutated (upload, rename, move, delete).
+        """
+        async with self._db.begin_session() as session:
+            query = (
+                sa.update(VFolderRow)
+                .where(VFolderRow.id == vfolder_id)
+                .values(updated_at=sa.func.now())
+            )
+            await session.execute(query)
```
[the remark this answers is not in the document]
seedspirit: A vfolder_id that does not exist in the call path cannot be provided. The track_content_update method is called only after the file operation service has verified the existence of the vfolder

tests/unit/manager/services/vfolder/test_vfolder_file_service.py
```diff
@@ -541,3 +560,138 @@ async def test_over_50_dirs_raises_invalid_parameter(
...
+        action = CreateDownloadSessionAction(
+            keypair_resource_policy=keypair_resource_policy,
+            user_uuid=user_uuid,
+            vfolder_uuid=vfolder_uuid,
+            path="data/file.bin",
+            archive=False,
+        )
+        await file_service.download_file(action)
+
+        mock_vfolder_repository.track_content_update.assert_not_called()
```
jopemachine: These tests look like they could be consolidated into a single parametrized test, for example by extracting the actions into fixtures.

src/ai/backend/manager/models/alembic/versions/5a139f0e951e_add_updated_at_to_vfolders.py
```diff
@@ -0,0 +1,27 @@
...
+down_revision = "9dc6609c92ce"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "vfolders",
+        sa.Column("updated_at", sa.DateTime(timezone=True), nullable=True),
+    )
```
jopemachine, line 23: Is it okay not to set last_used or a default value for updated_at?

seedspirit: Since the two columns have different meanings, setting `default_value` to `last_used` seems a bit odd. However, since value backfilling appears necessary at the time of migration, I’ve changed the code to backfill using either `last_used` or `created_at` during migration.

HyeockJinKim (approved):

PR #10823, opened 2026-04-06
feat(BA-5607): add FLAG value type to `PresetValueType` for flag-style args

seedspirit:
## Summary
- Add `FLAG = "flag"` to `PresetValueType` enum to support flag-style CLI args that take no value (e.g., `--enable-prefix-caching`)
- When `value_type="flag"` and value is truthy (`"true"`/`"1"`), the flag key is appended to args without a value; when falsy, it is omitted entirely
- Update validation, preset application logic (model_serving.py, db_source.py), and field descriptions across DTOs and GQL types

## Test plan
- [ ] Create a preset with `value_type="flag"`, `preset_target="args"`, `key="--enable-prefix-caching"`, `default_value="true"` — should pass validation
- [ ] Verify that when the preset is applied, only the flag key is appended (no value) when truthy
- [ ] Verify that when `default_value="false"`, the flag is omitted from the args list

Resolves BA-5607

----
📚 Documentation preview 📚: https://sorna--10823.org.readthedocs.build/en/10823/

----
📚 Documentation preview 📚: https://sorna-ko--10823.org.readthedocs.build/ko/10823/

### DIFF
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -3580,12 +3580,12 @@ input CreateRuntimeVariantPresetInput
   """
   How the value is applied: 'env' for environment variable, 'args' for command-line argument.
   """
-  presetTarget: String!
+  presetTarget: PresetTarget!
 
   """
   Data type for validation (e.g., 'str', 'int', 'float', 'bool', 'flag').
   """
-  valueType: String!
+  valueType: PresetValueType!
 
   """The default value shown to users when creating a deployment."""
   defaultValue: String = null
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/event_dispatcher/handlers/model_serving.py
```diff
@@ -158,10 +162,14 @@ async def handle_route_creation(
...
                         if vp is None:
                             continue
-                        if vp.preset_target == "env":
+                        if vp.preset_target == PresetTarget.ENV:
                             environ[vp.key] = pv.value
-                        elif vp.preset_target == "args":
-                            args_parts.append(f"{vp.key} {pv.value}")
+                        elif vp.preset_target == PresetTarget.ARGS:
+                            if vp.value_type == PresetValueType.FLAG:
+                                if (pv.value or "").strip().lower() in ("true", "1"):
```
jopemachine, line 169: I'm not sure why this parsing logic is needed here.
If possible, the type information should be included in the value of PresetValueEntry at the API boundary. If that's not feasible, would it be better to move it into a method on PresetValueEntry?

seedspirit: Since input validation at the API boundary is currently set up as shown below, it seems appropriate to implement the parsing logic described above.

```
def _validate_default_value(default_value: str | None, value_type: str) -> None:
    if default_value is None:
        return
    validator = VALUE_TYPE_VALIDATORS.get(value_type)
    if validator is None:
        return
    if validator is bool:
        if default_value.lower() not in ("true", "false", "1", "0"):
            raise InvalidAPIParameters(
                f"default_value '{default_value}' is not a valid {value_type}"
            )
    else:
        try:
            validator(default_value)
        except (ValueError, TypeError) as e:
            raise InvalidAPIParameters(
                f"default_value '{default_value}' is not a valid {value_type}: {e}"
            ) from e
```

PR #10824, opened 2026-04-06
fix(BA-5608): union owner+requester scaling group access in delegation

jopemachine:
Resolves [BA-5608](https://lablup.atlassian.net/browse/BA-5608).

## Summary

When a session is created with `owner_access_key` (delegation), the scaling group access check is now the **union** of the owner's and the requester's allowed scaling groups. This lets a superadmin/admin delegate sessions using a scaling group granted to their own keypair (or domain/group) even if the delegated user does not have direct access to it.

## Reproduction (before fix)

1. Superadmin creates a session via `POST /session` with `owner_access_key` set to a regular user's access key.
2. The request specifies a scaling group (e.g., `H100`) accessible to the superadmin (via `sgroups_for_keypairs`/`sgroups_for_domains`) but NOT associated with the delegated user's project group or keypair.
3. The request fails with `InvalidAPIParameters: Scaling group 'H100' is not accessible`, even though the superadmin has access.

After this fix, the delegation succeeds because the owner's allowed scaling groups are unioned with the requester's.

## Changes

- New `SessionCreationSpec.requester_access_key` (Optional) field to thread the original requester's access key down to the scheduling controller.
- `AgentRegistry.create_session` / `create_cluster` / `enqueue_session` now accept and forward `requester_access_key`.
- `SessionService.create_from_params` / `create_from_template` / `create_cluster` pass `action.requester_access_key` to the registry.
- Two enforcement points apply the union:
  1. `SchedulingController._resolve_scaling_group()`
  2. `SchedulerDBSource._query_allowed_scaling_groups_for_spec()` (used by `fetch_session_creation_data()` which feeds the `ScalingGroupFilter`)
- Regression tests in `TestResolveScalingGroupForDelegation` cover owner-only access, requester-only access (the fixed scenario), and the neither-party-has-access failure mode.

## Test plan

- [x] Unit tests pass (`tests/unit/manager/sokovan/scheduling_controller/test_integration.py`)
- [x] Manual e2e: created `restricted-sg` associated only with admin's keypair; verified delegation now succeeds with `owner_access_key=<user>` (previously failed with the "not accessible" error)

## Related

- #10837 / BA-5618 — Delegated session owner attribution (separate but related issue discovered during this investigation; tracked separately for focused review)

[BA-5608]: https://lablup.atlassian.net/browse/BA-5608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/unit/manager/sokovan/scheduling_controller/test_integration.py
+++ b/tests/unit/manager/sokovan/scheduling_controller/test_integration.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import uuid
+from dataclasses import dataclass
 from datetime import datetime, timedelta
 from pathlib import PurePosixPath
 from typing import Any, cast
@@ -11,6 +12,7 @@ from unittest.mock import AsyncMock, MagicMock
 import pytest
 from dateutil.tz import tzutc
 
+from ai.backend.common.exception import InvalidAPIParameters
 from ai.backend.common.plugin.hook import HookResult, HookResults
 from ai.backend.common.types import (
     AccessKey,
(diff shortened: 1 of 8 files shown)


src/ai/backend/manager/utils.py
```diff
@@ -287,7 +305,11 @@ async def query_userinfo_from_session(
         # Normal case when the user is creating her/his own session.
         owner_domain = requester_domain
         owner_uuid = requester_uuid
-        owner_role = UserRole.USER
+        actual_owner_role = requester_role
+        # Own-session creation always goes through the USER path to enforce
+        # explicit project membership via the agus table, regardless of the
+        # requester's actual role.
+        role_for_group_resolution = UserRole.USER
```
jopemachine: This should be refactored

tests/unit/manager/sokovan/scheduling_controller/test_integration.py
```diff
@@ -1107,3 +1109,115 @@ async def test_normal_terminate_broadcasts_terminating_event(
...
+                for name in case.mocked_allowed
+            ]
+        )
+
+        spec = self._make_spec(case.requested_sg, owner_ak, case.requester_ak)
+
+        if case.expected is None:
+            with pytest.raises(InvalidAPIParameters, match="not accessible"):
+                await scheduling_controller._resolve_scaling_group(spec)
+        else:
```
seedspirit, line 1219: It seems better to split the methods into success and failure tests.

jopemachine:
Closing this PR as it was based on a misunderstanding of the issue.

PR #10837, opened 2026-04-07
fix(BA-5608): attribute delegated sessions to the owner, not the requester

jopemachine:
Resolves BA-5608.

## Summary

When an admin creates a session via `owner_access_key`, the session must be attributed to the resolved owner — not the requester. Previously the `UserScope` was constructed with the requester's `user_uuid`/`user_role`, causing several downstream issues:

- `sessions.user_uuid` stored the admin's UUID, so the owner did not see the delegated session in their list and the admin's list was polluted.
- `registry.py` uses `user_scope.user_uuid` to look up the `UserRow` for container UID/GID, so the container ran as the admin instead of the delegated user.
- The customized image owner check compared against `user_scope.user_uuid`, blocking delegated sessions from using the owner's customized images.

## Scaling group access policy (owner-only)

An earlier draft (#10824, now closed) proposed **unioning** the owner's and requester's allowed scaling groups so that an admin could delegate sessions into an SG only their own keypair had access to. We decided against that: delegation should let an admin act *as* the owner, not grant the owner extra access the admin happens to hold.

With this PR, scaling group access for delegated sessions is resolved **using the owner's permissions only**:

- `SchedulingController._resolve_scaling_group()` calls `query_allowed_scaling_groups(domain_name, group_id, access_key)` where all three values come from `SessionCreationSpec` built from the owner's context (`owner_access_key` is passed down to `registry.enqueue_session()` as `access_key`, and `query_userinfo()` already returns the owner's `domain_name`/`group_id`).
- No union fallback to the requester's allowed SGs.
- A new regression test in `TestOwnerOnlyScalingGroupForDelegation` pins this invariant: if the requested SG is not in the owner's allowed set, `_resolve_scaling_group` raises `InvalidAPIParameters`, and the lookup is verified to use the owner's access key.

**Operational note:** admins who want to delegate a session into an SG the target user doesn't currently have access to must first grant the SG to the user (via domain/project/keypair association), then delegate.

## Changes

- Extend `query_userinfo()` / `query_userinfo_from_session()` to return a `SessionOwnerContext` dataclass that includes the resolved owner's role alongside `owner_uuid`, `group_id`, and `resource_policy`.
- Validate that the owner's role from the DB is non-null (raise `InternalServerError` if it is).
- In the non-delegation branch of `query_userinfo()`, return the requester's actual role instead of the previously hardcoded `UserRole.USER`. The hardcoded value is preserved as a separate local variable so the existing project-membership enforcement is unchanged.
- In `SessionService` (`create_from_params`, `create_from_template`, `create_cluster`), construct `UserScope` using the owner's `user_uuid`/`user_role`.
- Replace the redundant role lookup in `ModelServingRepository` with the role returned from `query_userinfo()`.
- Regression tests:
  - `test_owner_access_key_uses_owner_user_scope` — verifies `UserScope` carries the owner's identity.
  - `TestOwnerOnlyScalingGroupForDelegation` — verifies scaling group access is owner-only (no union fallback).

## Test plan

- [x] Unit tests pass (`test_session_lifecycle_service.py`, `test_integration.py`)
- [x] `pants lint/check/test` clean on changed files
- [x] Manual e2e: admin creates session with `owner_access_key=<user>`, verified `sessions.user_uuid` is the owner's UUID
- [x] Manual e2e: admin delegates into an SG only accessible to the target user → succeeds
- [x] Manual e2e: admin delegates into an SG only accessible to the admin's own keypair → rejected with `Scaling group '...' is not accessible` (desired behavior)

## Related

- #10824 — earlier draft using union semantics; **closed** in favor of the owner-only policy documented here.

[BA-5608]: https://lablup.atlassian.net/browse/BA-5608

### DIFF
--- a/tests/component/session/test_session_create_delegation.py
+++ b/tests/component/session/test_session_create_delegation.py
@@ -12,6 +12,8 @@ import sqlalchemy as sa
 from sqlalchemy.ext.asyncio.engine import AsyncEngine as SAEngine
 
 from ai.backend.client.v2.registry import BackendAIClientRegistry
+from ai.backend.common.dto.manager.session.request import CreateFromParamsRequest
+from ai.backend.common.types import SessionTypes
 from ai.backend.manager.models.group import GroupRow
 from ai.backend.manager.registry import AgentRegistry
 from ai.backend.manager.repositories.session.repository import SessionRepository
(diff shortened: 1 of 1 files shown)


tests/component/session/test_session_create_delegation.py
```diff
@@ -0,0 +1,144 @@
...
+            enqueue_only=True,
+        )
+
+        # Reuse the v2 client's signed transport for a v1 endpoint
+        # (same pattern as ``tests/component/user/test_keypair_ops.py``).
+        await admin_registry._client._request(
+            "POST",
+            "/session",
+            json=request.model_dump(mode="json", exclude_none=True),
+        )
```
HyeockJinKim, line 124: Use client sdk.

HyeockJinKim (approved):

PR #10861, opened 2026-04-08
fix(BA-5625): free orphan `resource_allocations` rows for CANCELLED/TERMINATED kernels

jopemachine:
Resolves BA-5625.

## Summary

Adds an Alembic data-cleanup migration that removes the data inconsistency between `kernels.status = 'CANCELLED'` and the normalized `resource_allocations` / `agent_resources` tables.

By design `KernelStatus.resource_occupied_statuses() == {RUNNING, TERMINATING}` (`src/ai/backend/manager/data/kernel/types.py:63-70`), so a `CANCELLED` kernel must never own a live allocation. Today some clusters have orphan rows where `resource_allocations.free_at IS NULL` even though the kernel is `CANCELLED`, and in pathological cases the matching `agent_resources.used` cache counter is also over-charged. This PR cleans up both.

## Why the orphans exist

Two factors leak rows into this state:

1. **Migration timeline gap.** `4b7b4b040a70_add_resource_slot_normalization_tables` (2026-02-08) backfilled `resource_allocations`, but the `free_at` column itself was added two days later in `321c033a321b_set_agent_resources_used_column_not_` (2026-02-10). Any kernel cancelled in that window — or right after the backfill — ended up with `free_at = NULL` once the column appeared.
2. **Cancellation paths that don't free allocations.** Several CANCELLED-transition paths in `src/ai/backend/manager/repositories/scheduler/db_source/db_source.py` flip `KernelRow.status` to `CANCELLED` without freeing the matching `ResourceAllocationRow`:
   - `_cancel_pending_sessions` (line 814) — never touches `free_at`.
   - `cancel_kernels_by_failed_image_pull` (line 2600) — never touches `free_at`.
   - `mark_session_cancelled` (line 3218) — does set `free_at`, but **without a kernel-status filter**, so it can flip a `RUNNING` kernel directly to `CANCELLED` while leaving `agent_resources.used` charged.

`be8ccc830` (BA-5060, _"fix orphaned resource_allocations rows causing ghost usage in check-presets"_) added a kernel-status filter to occupancy queries as a runtime workaround. That hides the symptom in `check-presets` and friends but leaves the table dirty and, more importantly, leaves `agent_resources.used` (which the scheduler reads directly for placement) silently over-counted in any case where a `RUNNING → CANCELLED` flip ever happened.

## What the migration does

- File: `src/ai/backend/manager/models/alembic/versions/4f08ccd6cb8e_free_canceled_kernel_resource_allocations.py`
- `down_revision = "99805291b6e0"` (current head, verified linear — exactly one migration points at this rev)
- `# Part of: 26.3.0 (main)`

The `upgrade()` runs in two ordered steps inside Alembic's transactional DDL so they are atomic:

### Step 1 — defensively decrement `agent_resources.used`

```sql
WITH agent_decrements AS (
    SELECT k.agent     AS agent_id,
           ra.slot_name AS slot_name,
           SUM(ra.used) AS delta
    FROM resource_allocations ra
    JOIN kernels k ON ra.kernel_id = k.id
    WHERE k.status   = 'CANCELLED'
      AND ra.free_at IS NULL
      AND ra.used    IS NOT NULL
      AND k.agent    IS NOT NULL
    GROUP BY k.agent, ra.slot_name
)
UPDATE agent_resources ar
SET used = GREATEST(ar.used - ad.delta, 0)
FROM agent_decrements ad
WHERE ar.agent_id  = ad.agent_id
  AND ar.slot_name = ad.slot_name;
```

For every orphan allocation that has a non-NULL `used` (i.e. the kernel was charged against an agent before being cancelled), aggregate per `(agent_id, slot_name)` and subtract that delta from the cache counter. `GREATEST(used - delta, 0)` keeps the counter non-negative on partially-corrupted data. **In a healthy cluster this is a zero-row no-op** because cancellation normally happens before the RUNNING transition that would set `used`.

### Step 2 — mark every orphan row as freed

```sql
UPDATE resource_allocations AS ra
SET free_at = NOW()
FROM kernels AS k
WHERE ra.kernel_id = k.id
  AND k.status = 'CANCELLED'
  AND ra.free_at IS NULL;
```

The partial index `ix_ra_occupied (WHERE free_at IS NULL)` automatically excludes the cleaned rows from active occupancy queries.

### Design notes

- **Order matters.** Step 2 must run after Step 1, otherwise filling in `free_at` would hide the orphans from Step 1's predicate.
- **`UPDATE` not `DELETE`.** Matches the runtime free pattern (`update_kernel_status_cancelled`, `free_kernel_resources`). Preserves the historical allocation record while the partial index keeps the rows out of active queries.
- **Idempotent by construction.** Both steps re-run as zero-row no-ops because the `free_at IS NULL` predicate excludes anything Step 2 already touched. No `inspector` guards needed.
- **`downgrade()` is `pass`.** The cleanup is one-way; we have no record of which rows were touched and the original `NULL` state was the bug. Same pattern as e.g. `9dc6609c92ce_register_existing_project_roles_in_ase.py:45-46`.

### Out of scope (follow-up)

The leak source in `_cancel_pending_sessions` / `cancel_kernels_by_failed_image_pull` / `mark_session_cancelled` is **not** fixed in this PR so the migration can be backported as a clean hotfix. A separate PR should mirror the `update_kernel_status_terminated` free-and-decrement behavior (or extract a shared helper) into those paths to stop new zombies from accumulating.

## Diagnostic queries (run before applying to size the cleanup)

```sql
-- How many orphan rows exist?
SELECT COUNT(*) AS zombie_count
FROM resource_allocations ra
JOIN kernels k ON ra.kernel_id = k.id
WHERE k.status = 'CANCELLED' AND ra.free_at IS NULL;

-- How many of them have a non-NULL `used` (i.e. agent_resources.used is over-counted)?
SELECT COUNT(*) AS used_nonnull_zombies,
       COUNT(DISTINCT k.agent) AS affected_agents
FROM resource_allocations ra
JOIN kernels k ON ra.kernel_id = k.id
WHERE k.status = 'CANCELLED'
  AND ra.free_at IS NULL
  AND ra.used IS NOT NULL;

-- Per-agent/slot delta that Step 1 will apply
SELECT k.agent, ra.slot_name, SUM(ra.used) AS would_subtract
FROM resource_allocations ra
JOIN kernels k ON ra.kernel_id = k.id
WHERE k.status = 'CANCELLED'
  AND ra.free_at IS NULL
  AND ra.used IS NOT NULL
  AND k.agent IS NOT NULL
GROUP BY k.agent, ra.slot_name
ORDER BY k.agent, ra.slot_name;
```

After `alembic upgrade head` the first query must return `0`.

## Test plan

- [ ] Run the diagnostic queries on staging, capture `zombie_count` and `used_nonnull_zombies`.
- [ ] Apply migration; rerun the first query — must be `0`.
- [ ] Re-run `alembic upgrade head` — expect a no-op (idempotency).
- [ ] Confirm a `RUNNING` control kernel still has `free_at IS NULL` afterwards and still appears in `_get_resource_occupancy` totals.
- [ ] Confirm an already-freed `CANCELLED` row keeps its original `free_at` timestamp (not overwritten).
- [ ] Confirm that for any `(agent_id, slot_name)` where Step 1 subtracted a non-zero delta, `agent_resources.used` is now exactly `(previous value - delta)` and is not negative.
- [ ] `pants fmt / fix / lint / check` on the changed file — done locally, all green.
- [ ] `alembic downgrade -1` succeeds without error (acceptable as documented one-way no-op).

### DIFF
--- a/changes/10861.fix.md
+++ b/changes/10861.fix.md
@@ -1,1 +1,1 @@
-Free orphan `resource_allocations` rows whose kernel is already in `CANCELLED` state, removing the data inconsistency that the `BA-5060` runtime workaround was masking.
+Free orphan `resource_allocations` rows whose kernel is already in `CANCELLED` or `TERMINATED` state, removing the data inconsistency that the `BA-5060` runtime workaround was masking.


jopemachine:
> Could you please test the combination of the kernel and resource allocation that were marked as "cancelled" or "terminated" in the database?

I confirmed it works as expected

fregataa (approved):

PR #10867, opened 2026-04-08
feat(BA-5628): Add orphaned resource allocation cleanup to recalc

seedspirit:
## Summary
- Add defensive logic to `recalc_resource_usage()` that detects `resource_allocations` rows where `free_at IS NULL` but the kernel is in a terminal status (CANCELLED, TERMINATED, ERROR), and sets `free_at = now()`
- New `cleanup_orphaned_allocations()` method in ResourceSlot repository/db_source layers
- Each corrected orphaned allocation is logged at warning level with kernel_id and slot_name

## Test plan
- [ ] Verify recalc correctly identifies orphaned allocations (terminal kernel + free_at IS NULL)
- [ ] Verify orphaned allocations get free_at set to now()
- [ ] Verify existing reconcile_agent_resources() behavior is unchanged
- [ ] pants lint/check passes for affected packages

### DIFF
--- a/tests/unit/manager/test_reconcile_agent_resources.py
+++ b/tests/unit/manager/test_reconcile_agent_resources.py
@@ -740,19 +740,19 @@ class TestOrphanedAllocationCleanup:
             row = (
                 await db_sess.execute(sa.select(ra.c.free_at).where(ra.c.kernel_id == kernel))
             ).one()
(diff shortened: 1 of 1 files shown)


fregataa (approved):
let's make a plan to refactor/deprecate the recalc API

PR #10870, opened 2026-04-08
feat(BA-5627): allocate resources per-kernel instead of per-session

fregataa:
## Summary
- Move resource allocation from session RUNNING transition hook (`RunningTransitionHook._update_occupied_slots`) to kernel RUNNING transition (`update_kernel_status_running`), ensuring resources are allocated atomically with the kernel status change in the same DB transaction
- Remove the now-unnecessary session-level allocation logic and its `scheduler_repository` dependency chain from `RunningTransitionHook`, `HookRegistryArgs`, and `SchedulerComponents`
- Extract allocation logic into `_allocate_kernel_resources_in_session()` private method using SQLAlchemy 2.0 ORM API

## Test plan
- [ ] Verify single-kernel session starts correctly with resources allocated at kernel RUNNING time
- [ ] Verify multi-kernel session allocates resources per-kernel as each kernel transitions to RUNNING
- [ ] Verify resource deallocation still works correctly on kernel termination
- [ ] Verify idempotency: re-calling allocation for an already-allocated kernel is a no-op
- [ ] Verify BATCH and INFERENCE session-type hooks still execute correctly on session RUNNING transition

Resolves BA-5627

### DIFF
--- a/src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
@@ -2103,7 +2103,7 @@ class ScheduleDBSource:
         )
         async with self._begin_session_read_committed() as db_sess:
             # Check current kernel status and fetch agent_id before update
-            check_stmt = sa.select(KernelRow.status, KernelRow.starts_at, KernelRow.agent_id).where(
+            check_stmt = sa.select(KernelRow.status, KernelRow.starts_at, KernelRow.agent).where(
                 KernelRow.id == kernel_id
             )
             check_result = await db_sess.execute(check_stmt)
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
```diff
@@ -2157,7 +2158,70 @@ async def update_kernel_status_running(
...
+        Idempotent: rows where used_at is already set are skipped.
+        No-op if agent_id is None or occupied_slots is empty.
+
+        Raises:
+            AgentResourceCapacityExceeded: If any slot would exceed agent capacity.
+        """
+        if not agent_id or not occupied_slots:
+            return
+        slots = resource_slot_to_quantities(occupied_slots)
+        for s in slots:
```
HyeockJinKim, line 2193: Isn't this step unnecessary? I think we can just handle it based on the kernel ID anyway.

PR #10872, opened 2026-04-08
fix(BA-5629): unblock stuck session termination and promotion flows

jopemachine:
## Summary
- coordinator-side lifecycle, promotion, kernel, and observer passes were iterating over schedulable or active-agent resource-group subsets, which could skip sessions pinned to resource groups that were no longer schedulable or no longer had ALIVE agents.
- switch those coordinator passes to enumerate all resource groups referenced by agents so termination and status-promotion checks still run for already-placed sessions.
- add force-specific terminatable status sets for sessions and kernels so forced termination can move already `TERMINATING` sessions directly to `TERMINATED`, while leaving pending cancellation as an explicit earlier step in the flow.
- add regression coverage for resource-group discovery, force-terminating from `TERMINATING`, resource deallocation, and scheduling-history recording.

Resolves BA-5629.

## Test plan
- [x] `./py -m pytest tests/unit/manager/repositories/scheduler/test_resource_deallocation.py -vv`
- [x] `./py -m pytest tests/unit/manager/repositories/scheduler/test_scheduling_history_recording.py -vv`
- [x] `./py -m pytest tests/unit/manager/repositories/scheduler/test_scaling_group_queries.py -vv`

### DIFF
--- a/src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
@@ -1065,11 +1065,11 @@ class ScheduleDBSource:
             return [row.scaling_group for row in result.fetchall()]
 
     async def get_all_scaling_groups(self) -> list[str]:
-        """Get all scaling groups referenced by agents, ignoring agent state."""
+        """Get all defined scaling groups."""
         async with self._begin_readonly_session_read_committed() as session:
-            query = sa.select(AgentRow.scaling_group).group_by(AgentRow.scaling_group)
+            query = sa.select(ScalingGroupRow.name)
             result = await session.execute(query)
-            return [row.scaling_group for row in result.fetchall()]
+            return [row.name for row in result.fetchall()]
 
     async def get_terminating_sessions_by_ids(
         self,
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/data/kernel/types.py
```diff
@@ -100,6 +100,22 @@ def terminatable_statuses(cls) -> frozenset[KernelStatus]:
...
+    @classmethod
+    @lru_cache(maxsize=1)
+    def force_terminatable_statuses(cls) -> frozenset[KernelStatus]:
+        """Return statuses that can transition directly to TERMINATED."""
+        return frozenset(
+            status
+            for status in cls
+            if status
+            not in (
+                cls.PENDING,
```
HyeockJinKim, line 112: Come to think of it, if a task is rescheduled, it should be marked as "terminated" even if it was previously "pending." Please include this in the update.

HyeockJinKim: Oh, I meant to say that in the session, not here.

HyeockJinKim (dismissed):

HyeockJinKim (approved):

PR #10880, opened 2026-04-08
feat: add --otel-endpoint option to TUI installer

devyubin:
resolves BA-5635

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/install/context.py
```diff
@@ -586,6 +586,21 @@ async def configure_manager(self) -> None:
...
         await self.etcd_put_json("config", data)
+        public_addr = self.install_variable.public_facing_address
+        manager_port = self.install_info.service_config.manager_addr.bind.port
+        self.sed_in_place(
+            toml_path,
+            re.compile(r"^internal-addr = .*$", flags=re.MULTILINE),
+            f'internal-addr = {{ host = "0.0.0.0", port = 18080 }}\n'
+            f'announce-addr = {{ host = "{public_addr}", port = {manager_port} }}\n'
+            f'announce-internal-addr = {{ host = "{public_addr}", port = 18080 }}',
+        )
```
seedspirit, line 597: This doesn't seem to be related to the Otel settings—could you explain why it was added?

configs/app-proxy-coordinator/halfstack.toml
```diff
@@ -41,6 +41,11 @@ drivers = ["console"]
 colored = true
 format = "verbose"
 
+[otel]
+enabled = true
+log-level = "INFO"
+endpoint = "http://127.0.0.1:4317"
```
seedspirit, line 47: No need to seperate port?

src/ai/backend/install/context.py
```diff
@@ -763,6 +807,12 @@ async def configure_webserver(self) -> None:
             data["ui"]["menu_inactivelist"] = ",".join(service.webui_menu_inactivelist)  # type: ignore
         with conf_path.open("w") as fp:
             tomlkit.dump(data, fp)
+        if self.install_variable.otel_endpoint:
+            self.sed_in_place(
+                conf_path,
+                'endpoint = "http://127.0.0.1:4317"',
```
seedspirit, line 813: Does `install/context.py` use hard-coded values instead of the values set in TOML?

src/ai/backend/install/context.py
devyubin: This is not directly related to OTel, but is necessary for Prometheus metric scraping to work in the TUI installer environment.

The default value of announce-internal-addr is host.docker.internal:18080. On Linux server environments where the TUI installer runs, Prometheus cannot resolve this hostname, so it fails to scrape /metrics.

By setting announce-internal-addr to the actual public_facing_address, Prometheus service discovery (/metrics/service_discovery) returns a reachable target address. I included it in this PR since it's part of the same monitoring/observability setup, but I can split it into a separate PR if preferred.

configs/app-proxy-coordinator/halfstack.toml
devyubin: No, the port doesn't need to be separated. 4317 is the standard OTLP port. All components use the same collector endpoint, so a single endpoint string (host + port together) is the standard configuration format used by OTel SDKs. The halfstack config uses http://127.0.0.1:4317 as a sensible default for local development, and the TUI installer replaces this full URL with the user-provided otel_endpoint value when configured.

src/ai/backend/install/context.py
devyubin: Yes this is the established pattern throughout the installer. All config customization (etcd port, agent RPC port, alembic DB address, wsproxy URL, etc.) works the same way. Copy the halfstack template, then sed_in_place the known default values with user-provided values. The OTel endpoint follows the same convention.

HyeockJinKim (approved):

PR #10882, opened 2026-04-08
fix(BA-5636): handle non-scope element types in RBAC permission check

fregataa:
## Summary
- Fix `RBACTypeConversionError` when calling `get_ssh_keypair` endpoint by handling element types (like KEYPAIR) that have no corresponding `ScopeType`
- In `check_permission_with_scope_chain()`, wrap `to_scope_type()` in try/except so non-scope element types gracefully fall back to scope-chain-only permission checks instead of crashing
- Add unit tests covering KEYPAIR permission grants via scope chain, operation mismatches, and missing permissions

## Test plan
- [x] Unit tests for non-scope element type permission checks (3 new test cases)
- [x] Existing permission controller tests pass (no regression)
- [ ] Manual verification: `get_ssh_keypair` endpoint succeeds for authorized users
- [ ] Manual verification: `get_ssh_keypair` endpoint returns permission denied for unauthorized users

Resolves BA-5636

### DIFF
--- a/changes/10882.fix.md
+++ b/changes/10882.fix.md
@@ -1,0 +1,1 @@
+Fix `RBACTypeConversionError` for non-scope element types (e.g., KEYPAIR) in permission checks by falling back to scope-chain-only validation.


fregataa:
closed this PR since #10883 covers the same issue

PR #10889, opened 2026-04-08
feat(BA-5639): Add vLLM runtime variant preset fixtures

seedspirit:
## Summary

- Add preset fixture data for the vLLM runtime variant
- Add automatic `runtime_variant_name` → UUID FK resolution during fixture populate
- Wire fixture populate order in `install-dev.sh` and installer context

## Fixture Items (29 vLLM presets)

| Category | Presets | Rank |
|---|---|---|
| **Model Loading** | model, dtype, quantization, max-model-len, served-model-name, trust-remote-code | 100–600 |
| **Resource & Memory** | gpu-memory-utilization, tensor-parallel-size, pipeline-parallel-size, data-parallel-size, kv-cache-dtype, enable-prefix-caching | 700–1200 |
| **Serving Performance** | max-num-seqs, max-num-batched-tokens, enable-chunked-prefill, scheduling-policy, performance-mode | 1300–1700 |
| **Multimodal** | limit-mm-per-prompt | 1800 |
| **Tool & Reasoning** | enable-auto-tool-choice, tool-call-parser, reasoning-parser | 1900–2100 |
| **Network** | host, port | 2200–2300 |
| **Monitoring** | enable-log-requests, disable-log-stats | 2400–2500 |
| **LoRA** | enable-lora, lora-modules | 2600–2700 |
| **Environment** | cuda-visible-devices, hf-token-env | 2800–2900 |

## Test plan

- [x] Unit test: preset populates with correct runtime_variant FK resolution
- [x] Local server: verified 29 items via `./bai admin runtime-variant-preset search`

### DIFF
--- a/src/ai/backend/manager/models/base.py
+++ b/src/ai/backend/manager/models/base.py
@@ -122,14 +122,16 @@ class FixtureReferenceSpec:
 
     For each matching fixture row, the value in `fixture_alias_column` is
     looked up against `lookup_table.lookup_match_column`, and the resulting
-    `lookup_table.lookup_value_column` is written into `fixture_fk_column`.
+    `lookup_table.lookup_referenced_column` is written into `fixture_fk_column`.
+    In SQL terms, the fixture's FK column ends up referencing
+    `lookup_table(lookup_referenced_column)`.
     """
 
     fixture_alias_column: str
     fixture_fk_column: str
     lookup_table: str
     lookup_match_column: str
-    lookup_value_column: str = "id"
+    lookup_referenced_column: str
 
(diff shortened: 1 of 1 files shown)


src/ai/backend/manager/models/base.py
```diff
@@ -948,6 +951,40 @@ async def populate_fixture(
                     await conn.execute(update_stmt, update_data)
 
 
+async def _resolve_runtime_variant_preset_references(
```
jopemachine: This function does not seem appropriate to be located here.

PR #10896, opened 2026-04-08
fix(BA-5640): wrap slot_types strings with SlotName in gather_node_measures_with_slots

fregataa:
## Summary
- Wrap plain string slot names with `SlotName()` in `gather_node_measures_with_slots` to prevent `AttributeError` when accelerator plugins (e.g., CUDA) provide `slot_types` as plain strings
- The multi-slot branch in `_extract_slot_measurement_scale_factor` accesses `slot_name.device_name`, which requires a `SlotName` instance

## Test plan
- [ ] Verify no `AttributeError` on `slot_name.device_name` with accelerator plugins providing plain strings
- [ ] Verify single-slot plugins (CPU, mem) continue to work unchanged
- [ ] pants lint/check pass

Resolves BA-5640

### DIFF
--- a/changes/10896.fix.md
+++ b/changes/10896.fix.md
@@ -1,0 +1,1 @@
+Fix `AttributeError` on `slot_name.device_name` when accelerator plugins provide plain strings in `slot_types`.


seedspirit:
@fregataa  Wouldn't we also need backports for versions prior to 26.4?

PR #10934, opened 2026-04-09
fix(BA-5656): parse BinarySize-suffixed values in resource_slots migration

fregataa:
## Summary
- Migration `e1f2a3b4c5d6` (normalize_resource_slots) used raw `::numeric(24, 6)` casts on JSONB values, which fails when values contain BinarySize suffixes (e.g. `32g`, `4m`, `512k`)
- Add a temporary PL/pgSQL helper function `_tmp_parse_binary_size()` that handles both plain numeric strings and BinarySize-suffixed strings, converting them to byte counts matching Backend.AI's `BinarySize` semantics
- The helper is created before backfill and dropped immediately after, leaving no artifacts in the schema

## Test plan
- [x] Migration upgrade succeeds with BinarySize-suffixed values (`32g`, `4g`, `512m`, `256m`, `512k`, `1t`, `4GiB`)
- [x] Plain numeric values pass through without regression (`0.5`, `1`, `1073741824`)
- [x] All three backfill targets verified: `model_card_resource_requirements`, `preset_resource_slots`, `deployment_revision_resource_slots`
- [x] Temporary function is cleaned up after migration
- [x] `pants lint` passes
- [ ] CI checks pass

Resolves BA-5656

### DIFF
--- a/changes/10934.fix.md
+++ b/changes/10934.fix.md
@@ -1,0 +1,1 @@
+Fix migration failure on BinarySize-suffixed resource_slots values (e.g. "32g", "4m").
--- a/src/ai/backend/manager/models/alembic/versions/e1f2a3b4c5d6_normalize_resource_slots.py
+++ b/src/ai/backend/manager/models/alembic/versions/e1f2a3b4c5d6_normalize_resource_slots.py
@@ -93,8 +93,52 @@ def upgrade() -> None:
         ["slot_name"],
     )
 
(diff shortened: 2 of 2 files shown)


seedspirit (commented):
It seems impossible to revert to the previous value after downgrading. Is this intentional? (It doesn't appear there is a separate method for recovery.)

seedspirit (approved):

PR #10941, opened 2026-04-09
fix(BA-5659): skip deserialization for events with no registered consumer

fregataa:
## Summary
- In `_consume_loop`, check the event name from the raw payload before calling `from_anycast()` to avoid full deserialization of messages with no registered consumer
- This prevents `ModuleNotFoundError` when a component (e.g., appproxy coordinator) receives events containing types from other components' packages (e.g., `ai.backend.manager`)
- The coordinator shares the `"events"` Redis stream with the manager via a separate consumer group, so it receives all manager events but only needs `EndpointRouteListUpdatedEvent`

## Test plan
- [x] Deploy appproxy coordinator in a separate venv (without `ai.backend.manager`) and verify no `ModuleNotFoundError` in `_consume_loop`
- [x] Verify `EndpointRouteListUpdatedEvent` is still consumed and processed correctly
- [x] Confirm unhandled events are properly acknowledged (no PEL buildup in Redis)

Resolves BA-5659

### DIFF
--- a/src/ai/backend/common/events/dispatcher.py
+++ b/src/ai/backend/common/events/dispatcher.py
@@ -650,6 +650,10 @@ class EventDispatcher(EventDispatcherGroup):
                 return
             try:
                 msg = cast(BroadcastMessage, msg)
+                # Check the event name before full deserialization (ref: BA-5659)
+                event_name = msg.payload["name"]
+                if not self._subscribers[event_name]:
+                    continue
                 msg_payload = MessagePayload.from_broadcast(msg.payload)
                 await self.dispatch_subscribers(
                     msg_payload.name,
--- a/src/ai/backend/common/events/dispatcher.py
+++ b/src/ai/backend/common/events/dispatcher.py
@@ -616,6 +616,12 @@ class EventDispatcher(EventDispatcherGroup):
                 return
             try:
                 mq_msg = cast(MQMessage, msg)
+                # Check the event name before full deserialization to avoid
+                # deserializing messages that have no registered consumer.
+                event_name = mq_msg.payload[b"name"].decode("utf-8")
+                if not self._consumers[event_name]:
+                    await self._msg_queue.done(mq_msg.msg_id)
+                    continue
                 msg_payload = MessagePayload.from_anycast(mq_msg.payload)
                 post_callback = _ConsumerPostCallback(
                     mq_msg.msg_id,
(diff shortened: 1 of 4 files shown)


src/ai/backend/common/events/dispatcher.py
```diff
@@ -572,40 +574,53 @@ async def _handle(
...
+            mq_msg.msg_id,
+            self._msg_queue,
+            len(self._consumers[event_name]) if event_name in self._consumers else 0,
+        )
+        if event_name not in self._consumers or not self._consumers[event_name]:
+            await post_callback.done()
             return
-        for consumer in consumers_handlers:
+        if self._log_events:
+            log.debug("DISPATCH_CONSUMERS(ev:{})}", event_name)
```
seedspirit: nitpick: typo 
```suggestion
            log.debug("DISPATCH_CONSUMERS(ev:{})", event_name)
```

src/ai/backend/common/events/dispatcher.py
```diff
@@ -572,40 +574,53 @@ async def _handle(
...
-        consumers_handlers = self._consumers[event_name].copy()
-        if not consumers_handlers:
-            # If there are no consumer handlers, we can just call post callbacks and return.
-            for post_callback in post_callbacks:
-                await post_callback.done()
+        event_name = deserialize_event_name_for_anycast(mq_msg.payload)
+        post_callback = _ConsumerPostCallback(
+            mq_msg.msg_id,
+            self._msg_queue,
+            len(self._consumers[event_name]) if event_name in self._consumers else 0,
```
seedspirit: How about make local var with `self._consumers[event_name]` first and re-use local var instead of calling `self._consumers[event_name]` repeatedly? It could defense race condition by `self._consumer[event_name]` change

src/ai/backend/common/events/dispatcher.py
```diff
@@ -572,40 +574,53 @@ async def _handle(
...
+    async def _dispatch_subscribers(
         self,
-        event_name: str,
-        source: AgentId,
-        args: tuple[Any, ...],
-        metadata: MessageMetadata | None = None,
+        broadcast_msg: BroadcastMessage,
     ) -> None:
+        event_name = deserialize_event_name_for_broadcast(broadcast_msg.payload)
+        if event_name not in self._subscribers or not self._subscribers[event_name]:
```
seedspirit: same here

seedspirit (approved):

PR #10944, opened 2026-04-10
fix(BA-4096): avoid nested bind mount of DO_NOT_STORE_PERSISTENT_FILES_HERE.md

seedspirit:
## Summary
- Stop bind-mounting `DO_NOT_STORE_PERSISTENT_FILES_HERE.md` at `/home/work/DO_NOT_STORE_PERSISTENT_FILES_HERE.md`; this was a nested bind mount (on top of the `/home/work` bind mount) that runc rejects on macOS Docker Desktop with VirtioFS because virtiofs resolves the nested path to `/run/host_virtiofs/...`, outside the container rootfs.
- Copy the file into the scratch `work/` directory during scratch provisioning instead, matching how dotfiles (`.bashrc`, `.vimrc`, etc.) and the Kubernetes agent path already handle it.
- Updated both the legacy active path (`agent.py:mount_krunner`) and the not-yet-wired `KernelRunnerMountProvisioner` so the fix isn't lost when the new path is activated.

## Test plan
- [x] On macOS with Docker Desktop VirtioFS, create a session and confirm it starts without the "outside of rootfs" error
- [x] Inside the new container, `cat /home/work/DO_NOT_STORE_PERSISTENT_FILES_HERE.md` shows the original content
- [ ] On Linux (native Docker), session creation still works and the file is present at `/home/work/DO_NOT_STORE_PERSISTENT_FILES_HERE.md`

Resolves BA-4096


fregataa (commented):
Is this tested on Linux machine?

seedspirit:
> Is this tested on Linux machine?

@fregataa Yup, also tested in linux

fregataa (approved):

PR #10955, opened 2026-04-10
fix(BA-5667): replace atomic list fields wholesale in ModelDefinition.merge()

agatha197:
## Summary

- Fix `ModelDefinition.merge()` corrupting `start_command` via index-based list merging
- After `deep_merge`, restore atomic list fields (`start_command`, `pre_start_actions`) from the override dict
- Scoped to `ModelDefinition.merge()` — does not change the general-purpose `deep_merge` utility

## Motivation

`deep_merge` merges lists element-by-element by index, which is correct for structural lists (e.g., `models`) but wrong for atomic lists like `start_command` that represent a single logical value and should be replaced wholesale.

Closes BA-5667

### DIFF
--- a/changes/10955.fix.md
+++ b/changes/10955.fix.md
@@ -1,0 +1,1 @@
+Fix `ModelDefinition.merge()` corrupting `start_command` by replacing atomic list fields wholesale instead of merging element-by-element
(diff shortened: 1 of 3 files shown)


fregataa:
Closed this PR since it will be addressed in another branch

PR #10959, opened 2026-04-10
fix(BA-5667): replace deep_merge with Pydantic-aware field-by-field merge in ModelDefinition

fregataa:
## Summary
- Replace `ModelDefinition.merge()`'s dict-based `deep_merge()` with explicit per-model merge functions (`merge_definition`, `merge_config`, `merge_service_config`, `merge_metadata`) that use Pydantic `model_fields_set` to determine merge strategy
- Atomic list fields like `start_command` are now replaced wholesale instead of being corrupted by index-based list merging
- Properly distinguish between unset fields (keep base) and explicitly-set `None` (clear the field)
- Add field coverage tests that detect when a new field is added to a config model but the corresponding merge function is not updated

## Test plan
- [x] Existing `test_registry.py` merge tests pass (vfolder + user override scenario)
- [x] New `TestMergeFieldCoverage` tests verify all fields are handled by each merge function
- [x] `pants lint`, `pants check` pass

Resolves BA-5667

### DIFF
--- a/changes/10959.fix.md
+++ b/changes/10959.fix.md
@@ -1,0 +1,1 @@
+Fix `ModelDefinition.merge()` corrupting `start_command` via index-based list merging by replacing `deep_merge()` with Pydantic-aware field-by-field merge functions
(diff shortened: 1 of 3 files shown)


agatha197 (approved):
LGTM

PR #10963, opened 2026-04-10
feat(BA-5669): Add optional Harbor registry installation in TUI dev installer

Yaminyam:
## Summary

Add a new `--with-harbor` option to the TUI installer (dev mode) that stands up a local Harbor container registry using Harbor's official offline installer. Harbor is **not** added to halfstack — it is managed separately via `./dev harbor`.

## How it works

- `CliArgs` / `InstallVariable` gain `with_harbor`, `harbor_http_port`, `harbor_admin_password`, `harbor_download_uri`
- `ServiceConfig` gains `harbor_enabled`, `harbor_hostname`, `harbor_http_port`, `harbor_admin_password`
- `DevContext.configure_harbor()`:
  1. Downloads Harbor offline installer archive (cached by filename)
  2. Extracts into `<base>/harbor/`
  3. Loads bundled `harbor.yml.tmpl` via `ruamel.yaml` round-trip (preserving comments), sets `hostname`, `http.port`, `harbor_admin_password`, `database.password`, `data_volume`, removes `https` block
  4. Runs Harbor's `install.sh` to generate `docker-compose.yml` and all per-service config files, then immediately runs `docker compose down` so that nothing is left running — lifecycle is owned explicitly by the user
- `DevContext.configure()` calls `configure_harbor()` when `with_harbor=True`, otherwise the method is a no-op
- New `./dev harbor <start|stop|restart|status|logs>` subcommand wraps `docker compose` against `<base>/harbor`
- `InstallReport` post-install guide shows a **Harbor** tab with URL and admin credentials when Harbor is enabled

## Design notes

- Follows the same in-place-edit pattern as other service configs (tomlkit for TOML → ruamel.yaml for YAML). No Jinja, no separate template files bundled in the installer package.
- Kept out of halfstack because Harbor is optional and Backend.AI is fully functional without it.
- `./dev harbor` is separate from the tmux-managed services because Harbor runs as a bunch of Docker containers, not a single foreground process.

## Test plan
- [ ] `./backendai-install install --with-harbor` on a clean dev env produces a working `harbor/docker-compose.yml`
- [ ] `./dev harbor start` brings Harbor up
- [ ] Harbor UI is accessible at `http://127.0.0.1:8084` with admin / Harbor12345
- [ ] `./dev harbor stop` brings it down cleanly
- [ ] Without `--with-harbor`, the installer behaves exactly as before
- [ ] `pants lint/check` pass (verified)


dev
```diff
@@ -189,6 +226,8 @@ Commands:
   stop    <service|all>     Stop a service
   restart <service|all>     Restart a service
   log     <service> [-f]    Show recent log output (-f to follow)
+  harbor  <start|stop|restart|status|logs>
```
HyeockJinKim, line 316: The consistency is all over the place... I'll fix it later.

HyeockJinKim (dismissed):

HyeockJinKim:
Please resolve this conflict. While we're resolving this conflict, it would be best to ensure consistency with other services. @Yaminyam

HyeockJinKim (approved):

PR #10951, opened 2026-04-10
fix(BA-5665,BA-5680): pass Manager model_definition to Agent, remove Agent-side generation

jopemachine:
> **⚠️ DRAFT — Do not merge.**
> This PR passes `model_definition` from Manager → Agent via `internal_data`. However, the WebUI uses the legacy `POST /services` path which does not store `model_definition` in DB, so Agent would get `ModelDefinitionNotFoundError` for WebUI-created services.
>
> **Superseded by #10984** — simpler approach where Agent reads VFolder `model-definition.yaml` directly (same as Manager), requiring no Manager-side changes and working for both legacy and sokovan paths.

## Summary
- Agent was generating model definitions independently per runtime variant, ignoring Manager's VFolder-merged health_check overrides (initial_delay, max_retries, max_wait_time)
- Pass Manager-generated `ModelDefinition` through `creation_spec` → `internal_data` to Agent
- Remove Agent-side model definition generation entirely (~400 lines) — Agent now uses Manager-provided definition as single source of truth
- Add Agent logging for health_check source and actual timeout values

## Jira
- [BA-5665](https://lablup.atlassian.net/browse/BA-5665) — Agent ignores VFolder health_check override
- [BA-5680](https://lablup.atlassian.net/browse/BA-5680) — Remove Agent-side model definition generation

## Why this approach was abandoned
The WebUI calls `POST /services` (legacy path), which does NOT go through sokovan's `ModelDefinitionGeneratorRegistry`. Legacy endpoints have `deployment_revisions.model_definition = NULL`, so `internal_data.model_definition` is never set. After the BA-5680 refactor that removed Agent-side fallback, Agent raises `ModelDefinitionNotFoundError` for all WebUI-created services.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[BA-5665]: https://lablup.atlassian.net/browse/BA-5665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10984, opened 2026-04-12
fix(BA-5665): apply VFolder model-definition.yaml override for all runtime variants on Agent

jopemachine:
## Summary
- Agent was ignoring VFolder `model-definition.yaml` for non-custom runtime variants (vllm, tgi, nim, sglang, modular-max), using only hardcoded defaults
- Now reads VFolder yaml and deep-merges it on top of variant defaults using `ModelDefinition.merge()`, mirroring Manager's `_apply_vfolder_override()` in `definition_generator/registry.py`

## Jira
[BA-5665](https://lablup.atlassian.net/browse/BA-5665)

## Changes (1 file: `agent/agent.py`)
- Extract `_read_model_definition_from_vfolder()` (required, for custom) and `_try_read_model_definition_from_vfolder()` (optional, returns None if not found) from inline custom variant code
- After variant-specific base generation, call `_try_read_model_definition_from_vfolder()` and `base.merge(override)` for non-custom variants
- Custom variant uses the same extracted helper

## Approach
Agent reads VFolder yaml directly — same approach Manager already uses. Works for both legacy (`POST /services`) and sokovan (`POST /v2/deployments`) paths without Manager-side changes.

See #10951 for the alternative approach (Manager → Agent passing) which was drafted because it breaks the legacy WebUI path.

## Test plan
- [ ] Deploy custom variant service with VFolder health_check override → verify applied
- [ ] Deploy non-custom variant with VFolder health_check override → verify applied
- [ ] Deploy non-custom variant without VFolder yaml → verify defaults used
- [ ] Verify Agent log: "applied VFolder model-definition.yaml override"

[BA-5665]: https://lablup.atlassian.net/browse/BA-5665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/10984.fix.md
+++ b/changes/10984.fix.md
@@ -1,0 +1,1 @@
+Apply VFolder model-definition.yaml health_check overrides for non-custom runtime variants (vllm, tgi, nim, sglang, modular-max) on the Agent side, mirroring Manager's existing merge behavior.
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -3417,15 +3417,23 @@ class AbstractAgent[
                 model_folder, kernel_config
             )
             if vfolder_override is not None:
-                base = ModelDefinition.model_validate(raw_definition)
-                override = ModelDefinition.model_validate(vfolder_override)
-                merged = base.merge(override)
-                raw_definition = merged.model_dump(mode="json")
-                log.info(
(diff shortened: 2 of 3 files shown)


jopemachine:
Although this PR only removes dead code, I am keeping it as a draft for now to avoid any unintended side effects

PR #10990, opened 2026-04-12
feat(BA-5681): auto-sync user-scope entries on role assign/unassign

fregataa:
## Summary
- Auto-sync user-scope membership entries in `association_scopes_entities` when roles are assigned or revoked
- On assign: INSERT user-scope entries for all scopes bound to the role (`ON CONFLICT DO NOTHING`)
- On revoke: DELETE user-scope entries only when no other role covers the same scope (correlated `NOT EXISTS` check)
- Scope-type-agnostic — works for project, domain, and any future scope types
- Integrated into `assign_role()`, `revoke_role()`, `bulk_assign_role()`, and `bulk_revoke_role()`

## Test plan
- [x] Unit tests: assign creates user-scope entries (single scope, multi-scope, no-scope, idempotent)
- [x] Unit tests: revoke removes entries (only role, overlapping roles, multi-scope partial, no-scope)
- [x] Existing role assignment tests still pass

Resolves BA-5681

### DIFF
--- a/src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
@@ -148,11 +148,12 @@ class PermissionDBSource:
         for row in rows:
             try:
                 key = uuid.UUID(row.entity_id)
-            except ValueError:
+            except (ValueError, TypeError):
                 log.warning(
-                    "Skipping scope entry with non-UUID entity_id: {}"
-                    " (entity_type: {}, expected: {})",
+                    "Skipping scope entry with non-UUID entity_id: {!r}"
+                    " (type: {}, entity_type: {}, expected: {})",
                     row.entity_id,
+                    type(row.entity_id).__name__,
                     row.entity_type,
                     EntityType.ROLE,
                 )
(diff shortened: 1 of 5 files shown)


src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
```diff
@@ -106,6 +120,122 @@ class PermissionDBSource:
...
+                )
+            )
+            .scalars()
+            .all()
+        )
+        result: dict[uuid.UUID, list[AssociationScopesEntitiesRow]] = {
+            uuid.UUID(sid): [] for sid in str_ids
+        }
+        for row in rows:
+            result[uuid.UUID(row.entity_id)].append(row)
```
HyeockJinKim: It seems possible that the entity's ID could be a string value—is there a guarantee that converting it to a UUID will always work?

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
```diff
@@ -106,6 +107,83 @@ class PermissionDBSource:
     def __init__(self, db: ExtendedAsyncSAEngine) -> None:
         self._db = db
 
+    @staticmethod
+    async def _sync_user_scopes_on_assign(
+        db_session: SASession,
+        user_ids: Collection[uuid.UUID],
```
seedspirit, line 113: nitpick: 
```suggestion
        user_ids: Sequence[uuid.UUID],
```

seedspirit (approved):

seedspirit:
It seems okay but @HyeockJinKim might have to look once again

PR #10998, opened 2026-04-13
feat(BA-4050): Add Prometheus-based kernel live stat query action

seedspirit:
## Summary
- Add `query_kernel_live_stat_batch` to `UtilizationMetricService` that issues gauge / diff / rate PromQL queries in parallel via `asyncio.gather`, reducing latency from 3×RTT to 1×RTT
- Add `KernelNode._entry_to_live_stat_mapping` to transform Prometheus pipeline output into the legacy Valkey-compatible GQL shape
- Add `KernelLiveStatAction` / `KernelLiveStatActionResult` action pair and wire through `UtilizationMetricProcessors`
- Add unit tests for the batch query pipeline and component-level Valkey equivalence tests

## Test plan
- [x] Unit tests: `tests/unit/manager/services/utilization_metric/test_kernel_live_stat_batch.py`
- [x] Component tests: `tests/component/metric/test_kernel_live_stat_valkey_equivalence.py`
- [ ] CI checks

Resolves BA-4050


src/ai/backend/manager/services/metric/root_service.py
```diff
@@ -14,5 +53,151 @@ def __init__(
...
+        action: KernelLiveStatAction,
+    ) -> KernelLiveStatActionResult:
+        if not action.kernel_ids:
+            return KernelLiveStatActionResult(stats=KernelLiveStatBatchResult(entries={}))
+        try:
+            gauge, diff, rate = await asyncio.gather(
+                self._query_gauge_kernel_live_stat(action.kernel_ids),
+                self._query_diff_kernel_live_stat(action.kernel_ids),
+                self._query_rate_kernel_live_stat(action.kernel_ids),
+            )
```
seedspirit, line 89: It was difficult to write a PromQL query that could handle everything at once, so I used `asyncio.gather` for query. In the case of the Prometheus client, it seems there are no issues since it uses a client pool, but I'm not entirely sure if this approach is the right one.

src/ai/backend/manager/services/metric/actions/live_stat.py
```diff
@@ -0,0 +1,33 @@
...
+
+from ai.backend.common.data.permission.types import EntityType
+from ai.backend.common.types import KernelId
+from ai.backend.manager.actions.action import BaseAction, BaseActionResult
+from ai.backend.manager.actions.types import ActionOperationType
+from ai.backend.manager.services.metric.types import KernelLiveStatBatchResult
+
+
+@dataclass(frozen=True)
+class KernelLiveStatAction(BaseAction):
```
jopemachine: How about QueryKernelLiveStatAction?

src/ai/backend/manager/services/metric/types.py
```diff
@@ -80,6 +88,97 @@ class ContainerMetricResult:
     values: list[MetricResultValue]
 
 
+@dataclass(frozen=True)
+class KernelMetricValue:
```
jopemachine: Wouldn’t it be better to move this under the data package?

src/ai/backend/manager/services/metric/root_service.py
fregataa: ~~Could you leave some details of the difficulty of batch query?~~
I got it. It is accurate to let prometheus compute such diff/rate metrics

fregataa (commented):
How about implementing `query_kernel_..._metrics()` methods in the prometheus client class rather than service?

src/ai/backend/manager/services/metric/root_service.py
```diff
@@ -14,5 +55,111 @@ def __init__(
...
+                template = (
+                    "sum by ({group_by})(" + CONTAINER_UTILIZATION_METRIC_NAME + "{{{labels}}})"
+                )
+            case UtilizationMetricType.RATE:
+                template = (
+                    "sum by ({group_by})(rate("
+                    + CONTAINER_UTILIZATION_METRIC_NAME
+                    + "{{{labels}}}[{window}]))"
+                    " / " + str(UTILIZATION_METRIC_INTERVAL)
+                )
```
jopemachine, line 151: How about
 
```python
template = (
    "sum by ({group_by})("
    "rate({metric}{{{labels}}}[{window}])"
    ")"
    " / {interval}"
).format(
    group_by=group_by,
    metric=CONTAINER_UTILIZATION_METRIC_NAME,
    labels=labels,
    window=window,
    interval=UTILIZATION_METRIC_INTERVAL,
)
```

or 

```
template = (
    f"sum by ({group_by})("
    f"rate({CONTAINER_UTILIZATION_METRIC_NAME}{{{{{labels}}}}}[{window}])"
    f")"
    f" / {UTILIZATION_METRIC_INTERVAL}"
)
```

src/ai/backend/manager/services/metric/types.py
```diff
@@ -80,6 +89,46 @@ class ContainerMetricResult:
...
+    def from_prometheus_response(cls, response: PrometheusResponse) -> Self:
+        grouped: dict[KernelId, list[KernelMetricValue]] = {}
+        for metric in response.data.result:
+            info = metric.metric
+            if (
+                info.kernel_id is None
+                or info.container_metric_name is None
+                or info.value_type is None
+                or not metric.values
+            ):
```
jopemachine, line 106: How about creating a method for indicating these to the Metric type?

jopemachine, line 106: How about creating a method for indicating these to the Metric type?

seedspirit:
> How about implementing `query_kernel_..._metrics()` methods in the prometheus client class rather than service?

@fregataa 
It would be better if the common client did not include domain logic such as query instances or query ranges. If domain logic must be included, it would be better to create an additional client that incorporates domain logic, similar to ValkeyLive or ValkeyStat. However, I overlooked the fact that using the client layer within the repository layer rather than the service layer is the current coding convention at Backend.AI, so I will make the change accordingly.

src/ai/backend/manager/services/factory.py
```diff
@@ -299,7 +300,11 @@ def create_services(args: ServiceArgs) -> Services:
         utilization_metric=UtilizationMetricService(
             args.prometheus_client,
             args.config_provider.config.metric.timewindow,
-            repositories.metric.repository,
+            MetricRepository(
+                args.db,
+                prometheus_client=args.prometheus_client,
+                timewindow=args.config_provider.config.metric.timewindow,
+            ),
         ),
```
seedspirit, line 308: The `ContainerUtilizationMetricService` also passes `prometheus_client` and `timewindow` twice, and not using repositories because it still uses `prometheus_client` directly within the service. This will be addressed in a subsequent issue involving the refactoring of the `ContainerUtilizationMetricService`.

src/ai/backend/manager/data/metric/types.py
```diff
@@ -0,0 +1,86 @@
...
+    RATE = enum.auto()
+    """
+    Represents a rate of changes calculated from underlying gauge/accumulation values
+    (e.g., I/O bps calculated from RX/TX accum.bytes)
+    """
+    DIFF = enum.auto()
+    """
+    Represents a difference of changes calculated from underlying gauge/accumulation values
+    (e.g., Utilization msec from CPU usage)
+    """
```
HyeockJinKim: Why not use StrEnum?

HyeockJinKim: and don't use auto.

src/ai/backend/manager/repositories/metric/repository.py
```diff
@@ -19,9 +42,122 @@
...
+                metric_name_filter=DIFF_METRICS,
+                value_type_filter=ValueType.CURRENT,
+            ),
+            self._query_kernel_live_stat(
+                kernel_ids,
+                metric_type=UtilizationMetricType.RATE,
+                metric_name_filter=RATE_METRICS,
+                value_type_filter=ValueType.CURRENT,
+            ),
+        )
```
HyeockJinKim, line 110: I’m thinking I could just save this as a preset and use it, but I feel like I’m not making the most of the features I’ve already implemented.

src/ai/backend/manager/services/metric/actions/live_stat.py
```diff
@@ -0,0 +1,33 @@
...
+
+from ai.backend.common.data.permission.types import EntityType
+from ai.backend.common.types import KernelId
+from ai.backend.manager.actions.action import BaseAction, BaseActionResult
+from ai.backend.manager.actions.types import ActionOperationType
+from ai.backend.manager.data.metric.types import KernelLiveStatBatchResult
+
+
+@dataclass(frozen=True)
+class QueryKernelLiveStatAction(BaseAction):
```
jopemachine: Even if it’s not immediately necessary, would it make sense to follow the existing code conventions by introducing a BaseAction-based KernelLiveStatAction, and then deriving QueryKernelLiveStatAction from KernelLiveStatAction?

src/ai/backend/manager/repositories/metric/repository.py
```diff
@@ -19,9 +42,122 @@
...
+    _timewindow: str
 
-    def __init__(self, db: ExtendedAsyncSAEngine) -> None:
+    def __init__(
+        self,
+        db: ExtendedAsyncSAEngine,
+        prometheus_client: PrometheusClient,
+        timewindow: str,
+    ) -> None:
         self._db = db
```
jopemachine, line 80: It seems worth separating db_source into its own file and class.
(This does not need to be done in this PR.)

seedspirit: It would be a good idea to apply this when we introduce the DB Source layer later on. Since there are currently no places using the database, I implemented it in the manner described above for now.

src/ai/backend/manager/repositories/metric/repository.py
```diff
@@ -19,9 +42,122 @@
...
 class MetricRepository:
     _db: ExtendedAsyncSAEngine
+    _prometheus_client: PrometheusClient
+    _timewindow: str
 
-    def __init__(self, db: ExtendedAsyncSAEngine) -> None:
+    def __init__(
+        self,
+        db: ExtendedAsyncSAEngine,
+        prometheus_client: PrometheusClient,
```
fregataa, line 77: repository should not have prometheus client

fregataa (changes requested):
service should call repository function and client function separately

src/ai/backend/manager/services/metric/root_service.py
```diff
@@ -16,3 +29,28 @@ def __init__(
...
+    ) -> QueryKernelLiveStatActionResult:
+        if not action.kernel_ids:
+            return QueryKernelLiveStatActionResult(
+                stats=KernelLiveStatBatchResult.empty(action.kernel_ids)
+            )
+        try:
+            values_by_kernel = await self._metric_repository.query_kernel_live_stats(
+                action.kernel_ids,
+            )
+        except (PrometheusConnectionError, FailedToGetMetric):
```
fregataa, line 45: service should not handle prometheus error

fregataa (approved):

PR #11019, opened 2026-04-14
fix(BA-5699): handle missing model-definition.yaml during session termination

fregataa:
## Summary
- Add `update_endpoint_route_info_for_termination()` to deployment repository that wraps `get_endpoint_health_check_config()` in try-except, falling back to `None` on failure
- Use this new method in `TerminatedTransitionHook._execute_inference_terminated()` so session termination is not blocked when `model-definition.yaml` is missing from storage

Resolves BA-5699


src/ai/backend/manager/repositories/deployment/repository.py
```diff
@@ -1128,6 +1128,32 @@ async def update_endpoint_route_info(
...
 
+    @deployment_repository_resilience.apply()
+    async def update_endpoint_route_info_for_termination(
+        self,
+        endpoint_id: uuid.UUID,
+    ) -> None:
+        connection_info = await self._db_source.generate_route_connection_info(endpoint_id)
+
+        try:
+            health_check_config = await self._db_source.get_endpoint_health_check_config(
```
jopemachine, line 1139: It does not seem necessary to update get_endpoint_health_check_config in the session termination hook.

HyeockJinKim (approved):

PR #11042, opened 2026-04-14
fix(BA-5708): bind aiohttp session identity to manager session_token in token login

fregataa:
## Summary
The sToken/keypair login path (`token_login_handler`) never copied the manager-issued `session_token` into the aiohttp session identity, so the browser cookie held an unrelated random value. On logout the webserver forwarded that random value to the manager, the DELETE matched zero rows, and the `login_sessions` row created during keypair authorize was leaked.

## Test plan
- [x] Log in via the keypair-auth hook plugin (sToken redirect flow) and verify a row is created in `login_sessions`.
- [x] Log out from the WebUI and verify the row is removed.
- [x] Repeat for the regular email/password login to confirm no regression.

Resolves BA-5708

### DIFF
--- a/changes/11042.fix.md
+++ b/changes/11042.fix.md
@@ -1,0 +1,1 @@
+Fix orphan `login_sessions` rows after WebUI logout when authenticated via the keypair (sToken) login flow.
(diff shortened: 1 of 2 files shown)


jopemachine:
Would it be possible to add a simple test case?

HyeockJinKim (approved):

PR #11069, opened 2026-04-14
fix: replace bare except with Exception in check-docker.py

harshadkhetpal:
## Summary

Replaces bare `except:` clauses with `except Exception:` in `scripts/check-docker.py`.

## Motivation

Bare `except:` catches `BaseException`, which includes `KeyboardInterrupt` and `SystemExit`. This can mask critical signals like Ctrl-C and make debugging harder. PEP 8 flags this as [E722](https://www.flake8rules.com/rules/E722.html).

## Change

```python
# Before
try:
    ...
except:
    ...

# After
try:
    ...
except Exception:
    ...
```

## Testing

No behavior change for normal error paths — only behavior change is that `KeyboardInterrupt`/`SystemExit` now propagate as intended.

### DIFF
--- a/scripts/check-docker.py
+++ b/scripts/check-docker.py
@@ -54,7 +54,7 @@ async def detect_snap_docker():
         async with sess.get("http://localhost/v2/snaps?names=docker") as r:
             try:
                 r.raise_for_status()
-            except:
+            except BaseException:
                 raise RuntimeError("Failed to query Snapd package information")
             response_data = await r.json()
             for pkg_data in response_data["result"]:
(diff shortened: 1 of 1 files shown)


scripts/check-docker.py
```diff
@@ -80,7 +80,7 @@ async def detect_system_docker():
         async with sess.get("http://localhost/version") as r:
             try:
                 r.raise_for_status()
-            except:
+            except BaseException:
```
HyeockJinKim, line 83: `except BaseException` is effectively the same as a bare `except:`. it catches everything including KeyboardInterrupt, SystemExit, and GeneratorExit.
Since the intent here is to handle HTTP/request errors, this should be except Exception: instead, so that system-level signals can still propagate normally.

PR #11071, opened 2026-04-14
fix(BA-5721): bypass RBAC permission checks for superadmin in action validators

fregataa:
## Summary
- Add early-return superadmin bypass to `ScopeActionRBACValidator`, `SingleEntityActionRBACValidator` so superadmin actions (e.g. project creation at `default` domain) no longer fail with `NotEnoughPermission`.
- Add validator tests under `tests/unit/manager/actions/validators/` that drive each validator against a real `PermissionControllerRepository` with seeded users/roles/permissions — covers superadmin bypass, missing user, allowed, and denied paths.

## Test plan
- [x] `pants test tests/unit/manager/actions/validators/test_rbac_validators.py`
- [x] As superadmin, create a project at `default` domain → succeeds
- [x] As non-superadmin without `PROJECT:CREATE`, project creation → still raises `NotEnoughPermission`

Resolves BA-5721

### DIFF
--- a/tests/unit/manager/actions/validators/test_rbac_validators.py
+++ b/tests/unit/manager/actions/validators/test_rbac_validators.py
@@ -1,10 +1,12 @@
 """Tests for RBAC action validators, focused on the superadmin bypass (BA-5721).
 
 These tests drive the validators against a real ``PermissionControllerRepository``
-backed by a real Postgres connection. Permission rows are seeded directly into
-the RBAC tables instead of mocked, so the bypass path (no DB lookup) and the
-non-superadmin enforcement path (real CTE-based scope chain check) are both
-exercised end to end.
+backed by a real Postgres connection. Permission rows are seeded directly at the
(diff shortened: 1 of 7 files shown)


HyeockJinKim (approved):

seedspirit (approved):

tests/unit/manager/actions/validators/test_rbac_validators.py
```diff
@@ -0,0 +1,481 @@
...
+    async def test_non_superadmin_without_permission_does_not_raise(
+        self,
+        repository: PermissionControllerRepository,
+        scope_action: _ProjectCreateAction,
+        trigger_meta: BaseActionTriggerMeta,
+        regular_user_without_permission: UserData,
+    ) -> None:
+        validator = LegacyScopeActionRBACValidator(repository)
+        with with_user(regular_user_without_permission):
+            await validator.validate(scope_action, trigger_meta)
```
jopemachine, line 481: It looks like these tests for multiple validators could be consolidated using parametrize

PR #11080, opened 2026-04-15
feat(BA-5682): backfill project-user scopes from association_groups_users

fregataa:
## Summary
- Add Alembic migration `c8d4e5f6a7b9` that copies every `(user_id, group_id)` row from `association_groups_users` into `association_scopes_entities` as a project-scope user membership with `relation_type='ref'`, per BEP-1048 (Project ─ref→ User).
- Single `INSERT ... SELECT ... ON CONFLICT ON CONSTRAINT uq_scope_id_entity_id DO NOTHING`, so the migration is idempotent and naturally deduplicates. Rows are migrated regardless of `is_active` to match legacy lookup semantics.
- `downgrade()` is a no-op: once the auto-sync path lands, backfilled rows cannot be distinguished from live ones, so the schema is torn down by its own future migration rather than here.

## Test plan
- [ ] `alembic upgrade head` on a DB populated with `association_groups_users` rows; verify matching `(project, group_id, user, user_id, ref)` rows appear in `association_scopes_entities`.
- [ ] Re-run `alembic upgrade head` and confirm no duplicate rows / no errors (idempotency).
- [ ] Verify inactive users/groups are migrated as-is.
- [ ] RBAC project membership lookups return the same set as the legacy table.

Resolves BA-5682

### DIFF
--- a/changes/11080.feature.md
+++ b/changes/11080.feature.md
@@ -1,0 +1,1 @@
+Backfill project-user memberships from `association_groups_users` into `association_scopes_entities` as `project ─ref→ user` relations for RBAC.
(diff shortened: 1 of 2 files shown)


fregataa:
Closed this PR since #5734 already has resolved the same issue

PR #11107, opened 2026-04-15
feat(BA-5740): add scope resolver on EntityRefGQL

fregataa:
## Summary
- `Role.scopes` returns `EntityRefGQL` rows filtered by `entity_type=ROLE, entity_id=<role_id>`, which meant the scope side (`scope_type`/`scope_id`) had no resolver and clients could read the raw `scope_id` but not fetch the scope target (e.g. the project name).
- Adds a `scope` resolver on `EntityRefGQL` that mirrors the existing `entity` resolver and returns `EntityNode | None`, so queries like `Role.scopes { edges { node { scope { ... on ProjectV2 { basicInfo { name } } } } } }` now work.
- Extracts the shared `RBACElementType` → DataLoader dispatch into a single `_load_rbac_element` helper reused by both `entity` and `scope` resolvers.

## Test plan
- [x] `pants fmt` / `pants fix` / `pants lint` on the diff — all green
- [x] Live GraphQL test against local manager: `adminRole(id: ...) { scopes { edges { node { scope { ... on ProjectV2 { basicInfo { name } } } } } } }` resolves the associated project correctly (verified `project-86975a60-admin` → `ba-5725-test-project`)
- [x] `entity` resolver still returns the role itself (no regression)

Resolves BA-5740

### DIFF
--- a/changes/11107.feature.md
+++ b/changes/11107.feature.md
@@ -1,0 +1,1 @@
+Add `scope` resolver on `EntityRefGQL` so RBAC role scope connections can resolve the scope target (e.g., project, domain) directly in GraphQL.
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -6070,6 +6070,11 @@ type EntityRef implements Node
 
   """The resolved entity object."""
   entity: EntityNode
+
+  """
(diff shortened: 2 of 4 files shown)


ironAiken2 (approved):
LGTM

seedspirit (approved):

PR #11112, opened 2026-04-15
fix: populate model_definition in legacy endpoint creation flow

jopemachine:
## Summary
- Legacy `create_endpoint_legacy` path was not resolving and storing `model_definition` content in `DeploymentRevisionRow.model_definition` JSONB column
- Unlike the v2 `add_model_revision` flow which calls `_resolve_model_definition()`, the legacy flow left this column as NULL
- This caused the executor (`executor.py`) to miss health check configuration for legacy-created deployments since it reads from `target_revision.model_definition`

## Changes
- `DeploymentService.create_legacy_deployment()`: resolve model_definition via `ModelDefinitionGeneratorRegistry` before calling controller
- `DeploymentController.create_deployment()`: accept optional `model_definition` and set on revision spec
- `LegacyEndpointCreatorSpec`: add `model_definition` field, propagate from `ModelRevisionSpec`
- `db_source.create_endpoint_legacy()`: store `model_definition` in `DeploymentRevisionRow`

## Test plan
- [x] `pants lint/check` pass
- [x] `tests/unit/manager/repositories/deployment/test_deployment_repository.py` — 53 passed
- [x] `tests/unit/manager/services/deployment/test_deployment_service.py` — passed
- [x] `tests/unit/manager/sokovan/deployment/` — all passed
- [x] `tests/unit/manager/sokovan/deployment/definition_generator/` — passed

### DIFF
--- a/src/ai/backend/manager/services/deployment/service.py
+++ b/src/ai/backend/manager/services/deployment/service.py
@@ -528,11 +528,10 @@ class DeploymentService:
         """
         log.info("Creating deployment with name: {}", action.draft.name)
         draft_revision = action.draft.draft_model_revision
-        model_definition = await self._resolve_model_definition_from_draft(draft_revision)
-        deployment_info = await self._deployment_controller.create_deployment(
-            action.draft,
-            model_definition=model_definition,
+        action.draft.model_definition = await self._resolve_model_definition_from_draft(
+            draft_revision
         )
+        deployment_info = await self._deployment_controller.create_deployment(action.draft)
         await self._deployment_controller.mark_lifecycle_needed(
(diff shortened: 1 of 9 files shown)


src/ai/backend/manager/services/deployment/service.py
```diff
@@ -929,6 +935,22 @@ async def _resolve_model_definition(
         )
         return await self._model_definition_generator_registry.generate_model_definition(context)
 
+    async def _resolve_model_definition_from_draft(
```
jopemachine, line 937: How about move this to ModelRevisionSpecDraft method?

PR #11134, opened 2026-04-16
feat(BA-5735): add admin API to refresh revisions for active deployments

HyeockJinKim:
## Summary

- Added a parameterless superadmin endpoint that rebuilds and activates a fresh revision for every deployment in `EndpointLifecycle.active_states()`. The operation reuses `DeploymentController.add_revision` + `activate_revision`, so preset / deployment-config / model_definition are re-resolved. It's primarily meant to repair deployments whose current revision has stale or missing `model_definition` after backing-store migrations, but generalizes to any case where a fresh re-resolve is needed.
- Each deployment is processed independently — one failure (e.g. endpoint with no current revision) does not abort the rest. Per-deployment outcomes (`success`, `new_revision_id`, `failure_reason`) are returned so operators can triage failures.
- Exposed through REST (`POST /v2/deployments/admin/refresh-revisions`), GraphQL (`mutation admin_refresh_deployment_revisions`), Client SDK, and CLI (`./bai admin deployment revision refresh`).

## Test plan

- [x] `pants fmt`, `pants fix`, `pants lint` pass
- [x] `pants test --changed-since=HEAD --changed-dependents=transitive` passes
- [x] `pants check --changed-since=HEAD` introduces no new errors beyond existing `services/processors.py` import-stub warnings
- [x] E2E: verified on local dev stack with `./bai admin deployment revision refresh`; response contains both success (new revision UUID) and failure entries (deployments with null `current_revision_id` are reported, not aborting)
- [ ] Reviewer to confirm: per-deployment transaction isolation works as expected when one deployment errors mid-run
- [ ] Reviewer to confirm: `admin_refresh_deployment_revisions` mutation appears in `docs/manager/graphql-reference` schema dump once CI regenerates

### DIFF
--- a/changes/11134.feature.md
+++ b/changes/11134.feature.md
@@ -1,0 +1,1 @@
+Add admin API to refresh revisions for all active deployments, rebuilding each revision through `DeploymentController` so preset, deployment-config, and model_definition are re-resolved. Partial success is reported per deployment.
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -260,6 +260,16 @@ type AdminGetSSHKeypairPayload
 }
 
 """
+Added in UNRELEASED. Payload for admin bulk revision refresh mutation result.
+"""
+type AdminRefreshDeploymentRevisionsPayload
+  @join__type(graph: STRAWBERRY)
+{
+  """Per-deployment refresh outcomes (partial success by design)"""
+  results: [RevisionRefreshResult!]!
+}
+
+"""
 Added in 26.4.2. Admin input for registering (overwriting) a user's SSH keypair.
 """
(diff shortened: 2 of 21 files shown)


jopemachine:
Is it necessary to redeploy all deployments solely to synchronize with the vfolder?

HyeockJinKim:
I had interpreted this as an action intended to trigger the process during a full migration or in an abnormal state.

jopemachine:
> I had interpreted this as an action intended to trigger the process during a full migration or in an abnormal state.

Refreshing or redeploying all endpoints could be a heavy operation in some cases. From a scalability perspective, I think it would be better for the service to take only specific endpoints as arguments and refresh them, while the CLI is responsible for querying the endpoints and passing them to the service.

jopemachine:
Also, please add the missing test cases.

PR #11143, opened 2026-04-16
fix(BA-5754): add RBAC validation to v2 session GET endpoint

### LINKED ISSUE #11140: Apply RBAC validator to session GET endpoint via SingleEntityActionProcessor
## Motivation

- The get_session_info processor currently uses a plain ActionProcessor with no RBAC validation, allowing session info retrieval without permission checks
- Other single-entity operations like modify_session already use SingleEntityActionProcessor with single_entity_validator — the same pattern should be applied to session GET

## Objective

- Apply SingleEntityActionProcessor and RBAC single_entity_validator to get_session_info so that session retrieval is subject to permission checks

## Details

1. In services/session/processors.py, change get_session_info from ActionProcessor to SingleEntityActionProcessor and pass single_entity_validator as validators
1. Verify whether GetSessionInfoAction already implements the SingleEntityAction interface (entity_id, etc.) or needs modification
1. Follow the existing modify_session pattern (processors.py lines 284-288) as reference

## Impact

- RBAC validation will be enforced on session GET calls, blocking unauthorized users from retrieving session info
- REST and GQL handlers that call get_session_info will be affected — any previously unguarded access paths will now require proper permissions

JIRA Issue: BA-5754

seedspirit:
resolve #11140 (BA-5754)

## Summary
- Add `SingleEntityActionProcessor` with RBAC validation to the v2 `GET /sessions/{id}` endpoint via `GetSessionAction`
- Add component tests verifying RBAC enforcement: owner access succeeds, non-owner denied (403), superadmin bypass, nonexistent session (404)

## Test plan
- [x] `pants lint tests/component/session_v2::` passes
- [x] `pants check tests/component/session_v2::` passes
- [x] `pants test tests/component/session_v2::` passes (5 tests)

### DIFF
--- a/tests/component/session_v2/test_session_get_rbac.py
+++ b/tests/component/session_v2/test_session_get_rbac.py
@@ -1,0 +1,75 @@
+"""Component tests for v2 Session GET endpoint RBAC validation.
+
+Tests that the v2 GET /sessions/{id} endpoint enforces RBAC:
+- Regular users can GET their own session (owner permission via scope chain)
+- Regular users WITHOUT permission are denied on other users' sessions (403)
+- Superadmin bypasses RBAC and can access any session
+- Superadmin querying nonexistent session gets 404
+"""
+
+from __future__ import annotations
+
+import uuid
+from typing import TYPE_CHECKING
+
+import pytest
+
+from ai.backend.client.v2.exceptions import NotFoundError, PermissionDeniedError
+from ai.backend.client.v2.v2_registry import V2ClientRegistry
+
+if TYPE_CHECKING:
+    from .conftest import SessionSeedData
+
+
+class TestSessionGetV2RBAC:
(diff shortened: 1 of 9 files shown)


jopemachine:
Similar patterns appear in deployment and resource_preset adapters as well, so I think they also should be reviewed.

tests/component/session_v2/test_session_get_rbac.py
jopemachine, line 1: This is just a question out of my own curiosity, but is there a reason RBAC tests are categorized as component tests rather than unit tests?

seedspirit: Since methods at both the processor level (including the validator) and the repository level (for RBAC checks) are used, it would be difficult to classify this as a unit

jopemachine (approved):

PR #11150, opened 2026-04-16
fix(BA-5758): replace broken revision name filter with `revision_number`

jopemachine:
## Summary
- `deployment_revisions` has no `name` column — revisions are identified by `revision_number`, and `revision-{N}` is only a display-side composition in Python. The entire `RevisionFilter.name` / `RevisionOrderField.NAME` path was non-functional end-to-end: pydantic `ValidationError` in the CLI and `AttributeError` on `DeploymentRevisionRow.name` in the repository.
- Drop the dead `name` field/enum value from the DTO (v2 + legacy), GraphQL (`ModelRevisionFilter`, `ModelRevisionOrderFieldGQL`), manager adapter (v2 + REST v1), row-level conditions/orders, and both revision search CLIs. Add a working `revision_number: IntFilter` and `RevisionOrderField.REVISION_NUMBER` equivalent (`make_int_conditions(DeploymentRevisionRow.revision_number)` plus a matching ordering method).
- CLI: `./bai deployment revision search --name-contains <str>` → `--revision-number <int>` (exact match), same for the admin-scoped `./bai admin deployment revision search` variant. The corresponding v2 `RevisionOrderField` enum test is updated to match.

Resolves BA-5758.

### DIFF
--- a/src/ai/backend/manager/models/deployment_revision/conditions.py
+++ b/src/ai/backend/manager/models/deployment_revision/conditions.py
@@ -4,11 +4,10 @@ from __future__ import annotations
 
 import uuid
 from collections.abc import Collection
-from typing import cast
 
 import sqlalchemy as sa
 
-from ai.backend.common.data.filter_specs import StringInMatchSpec, StringMatchSpec
+from ai.backend.manager.models.condition_utils import make_int_conditions
 from ai.backend.manager.models.deployment_revision import DeploymentRevisionRow
 from ai.backend.manager.repositories.base import QueryCondition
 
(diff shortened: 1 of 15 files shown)


src/ai/backend/manager/models/deployment_revision/conditions.py
```diff
@@ -34,66 +51,64 @@ def inner() -> sa.sql.expression.ColumnElement[bool]:
...
         def inner() -> sa.sql.expression.ColumnElement[bool]:
             if spec.case_insensitive:
-                condition = DeploymentRevisionRow.name.ilike(f"{spec.value}%")
+                name_cond = EndpointRow.name.ilike(f"{spec.value}%")
             else:
-                condition = DeploymentRevisionRow.name.like(f"{spec.value}%")
+                name_cond = EndpointRow.name.like(f"{spec.value}%")
             if spec.negated:
-                condition = sa.not_(condition)
-            return cast(sa.sql.expression.ColumnElement[bool], condition)
```
HyeockJinKim, line 68: Why did you change this to EndpointRow?

HyeockJinKim (changes requested):

src/ai/backend/manager/models/deployment_revision/conditions.py
jopemachine: I will remove this filter

tests/unit/common/dto/manager/v2/deployment/test_types.py
jopemachine, line 114: Let's remove these useless tests @HyeockJinKim

fregataa (approved):

seedspirit (approved):

HyeockJinKim (approved):

PR #11165, opened 2026-04-16
feat(BA-5765): add RBAC-enforced VFolder purge mutation

fregataa:
## Summary
- Add `PurgeVFolderV2RBACAction` (`SingleEntityActionProcessor` + `single_entity_rbac_validators`).
- Service `purge_v2_rbac()` — `get_by_id` + `delete_vfolders_forever` + storage removal, no manual user/host checks.
- Adapter `purge()` routes to RBAC processor; `bulk_purge()` stays on legacy path.

## Test plan
- [x] `pants fmt/fix/lint/check` — green
- [x] Component test: `TestPurgeVFolderRBAC` (regular user 403) via SDK

Resolves BA-5765

### DIFF
--- a/src/ai/backend/manager/api/adapters/vfolder.py
+++ b/src/ai/backend/manager/api/adapters/vfolder.py
@@ -406,8 +406,8 @@ class VFolderAdapter(BaseAdapter):
 
     async def purge(self, vfolder_id: UUID) -> PurgeVFolderPayload:
         """Permanently delete a vfolder. RBAC enforced."""
-        action = PurgeVFolderV2RBACAction(vfolder_id=vfolder_id)
-        await self._processors.vfolder.purge_v2_rbac.wait_for_complete(action)
+        action = PurgeVFolderV2Action(vfolder_id=vfolder_id)
+        await self._processors.vfolder.purge_v2.wait_for_complete(action)
         return PurgeVFolderPayload(id=vfolder_id)
 
     async def deploy(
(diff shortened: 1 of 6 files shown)


jopemachine:
Is there a reason why `bulk_purge` is not replaced with the `purge_v2_rbac`?
If bulk_purge is migrated to the RBAC version, there would be no remaining usages of purge_v2, so it seems safe to replace it.

src/ai/backend/manager/api/adapters/vfolder.py
```diff
@@ -392,12 +395,9 @@ async def delete(self, vfolder_id: UUID) -> DeleteVFolderPayload:
         return DeleteVFolderPayload(id=vfolder_id)
 
     async def purge(self, vfolder_id: UUID) -> PurgeVFolderPayload:
-        """Permanently delete a vfolder."""
-        me = current_user()
-        if me is None:
-            raise UnreachableError("User context is not available")
-        action = PurgeVFolderV2Action(user_id=me.user_id, vfolder_id=vfolder_id)
```
HyeockJinKim, line 399: PurgeVFolderV2Action seems like it might become a dangling reference. If we need to move forward with this, please clean up PurgeVFolderV2Action.

fregataa:
> Is there a reason why `bulk_purge` is not replaced with the `purge_v2_rbac`? If bulk_purge is migrated to the RBAC version, there would be no remaining usages of purge_v2, so it seems safe to replace it.

Because batch RBAC validation is not ready yet. I target this job to support purge single VFolder only and I will impl batch purge with RBAC in the future. Check #11186 @jopemachine

jopemachine:
Please resolve the conflict

jopemachine (approved):

Issue #10825, opened 2026-04-06
Project admin page: Assign User API


fregataa:
Provide project admins with the ability to discover and assign users to their project via two complementary mechanisms, selectable per organization's security policy.

== Current state ==

- User assign/unassign endpoints are implemented (BA-5484, BA-5485, BA-5486), but require the caller to already know user UUIDs.

- User list service/repository for project admin is implemented (BA-5458).

- User list API handler (BA-5459) was closed as Not Planned since project_users_v2 already covers listing assigned users.

- Internal filtering logic (same domain + not already assigned) exists in group/db_source but is not exposed as a standalone search endpoint.

== Membership model change (PR #10977) ==

Project membership is now determined by role assignment, not by association_groups_users.

- **Assign user to project** = assign a project-scoped role (e.g., Project-A-Member) to the user via user_roles.

- **Unassign user from project** = remove all role assignments for roles bound to that project scope.

- **Query users in project** = unified pattern:

\```Java
SELECT DISTINCT ur.user_id

FROM user_roles ur

JOIN association_scopes_entities ase

  ON ase.entity_type = 'role'

  AND ase.entity_id = ur.role_id::text

  AND ase.scope_type = 'project'

  AND ase.scope_id = :project_id

\
```

- association_groups_users is sunset; user-scope membership is derived from role-scope binding.

- System roles (Project Admin, Project Member) are auto-created when a project is created.

- BEP documents (BEP-1008, BEP-1012, BEP-1048) updated accordingly.



== Design ==

Two approaches provided simultaneously. Organizations choose based on their security requirements.



=== Approach A: Scoped search via USER_NAME entity ===

- Add a new RBAC entity type (USER_NAME or USER_IDENTITY) controlling access to basic user identity info.

- Included fields: uuid, username, email. Optionally full_name for disambiguation.

- Only return users with status=ACTIVE (do not expose inactive/deleted users).

- Project admin role can optionally be granted this entity permission (opt-in).

- When granted: project admin can search/autocomplete users in the same domain who are not yet assigned to the project.

- Support filtering by email/username with minimum input length (3 chars) to prevent enumeration.

- Support pagination (offset-based and/or cursor-based).

Use case: Organizations where user discovery is needed (e.g., internal team management).



=== Approach B: Blind input API (baseline) ===

- Accept email or username as direct input for assignment.

- Backend internally resolves to user UUID within the same domain, then assigns a project-scoped role to the user.

[...]

PR #11181, opened 2026-04-18
feat(BA-5613): implement blind role invitation data layer and service

fregataa:
## Summary
- Add `RoleInvitationRow` model, `RoleInvitationState` enum, and alembic migration for the `role_invitations` table with a partial unique index (`state != 'accepted'`)
- Implement RBAC action declarations (`Create/Read/Accept/Reject/Cancel`) registered in `RBAC_ACTION_REGISTRY` for permission validation
- Integrate invitation CRUD into `PermissionDBSource` and `PermissionControllerRepository` — accept atomically transitions state and assigns the role in a single DB session via `_assign_role_in_session()`
- Email-only user resolution with `execute_bulk_creator_partial` for partial-failure support

## Test plan
- [x] Verify alembic migration applies cleanly (`alembic upgrade head`)
- [x] CI: lint, type-check, existing test suite passes

Resolves BA-5613

----
📚 Documentation preview 📚: https://sorna--11181.org.readthedocs.build/en/11181/

----
📚 Documentation preview 📚: https://sorna-ko--11181.org.readthedocs.build/ko/11181/


src/ai/backend/manager/actions/action/rbac_role_invitation.py
```diff
@@ -0,0 +1,135 @@
...
+)
+from ai.backend.manager.data.role_invitation.types import RoleInvitationData
+
+
+@dataclass
+class CreateRoleInvitationByEmailAction(BaseRBACAction):
+    """Project admin creates role invitations by invitee emails."""
+
+    invitee_emails: list[str]
+    domain_name: str
```
HyeockJinKim: Please check if a domain value is required.

src/ai/backend/manager/actions/action/rbac_role_invitation.py
```diff
@@ -0,0 +1,135 @@
...
+    inviter_user_id: UUID | None
```
HyeockJinKim: Does "inviter" have to be optional?

src/ai/backend/manager/data/role_invitation/types.py
```diff
@@ -0,0 +1,24 @@
...
+
+@dataclass(frozen=True)
+class RoleInvitationData:
+    id: uuid.UUID
+    inviter_user_id: uuid.UUID | None
+    invitee_user_id: uuid.UUID
+    role_id: uuid.UUID
+    state: RoleInvitationState
+    created_at: datetime
+    modified_at: datetime | None
```
HyeockJinKim: Both `modified_at` and `updated_at` are present. I'd like to standardize them to one format.

src/ai/backend/manager/actions/action/rbac_role_invitation.py
```diff
@@ -0,0 +1,134 @@
...
+    RBACRequiredPermission,
+)
+from ai.backend.manager.data.role_invitation.types import RoleInvitationData
+
+
+@dataclass
+class CreateRoleInvitationByUsernameAction(BaseRBACAction):
+    """Project admin creates role invitations by invitee usernames or emails."""
+
+    invitee_username_or_emails: list[str]
```
seedspirit: It seems better to accept only either the UUID or the email address, but is the reason for allowing both the username and email address to be accepted to maintain backward compatibility?

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
```diff
@@ -1083,3 +1110,183 @@ async def bulk_revoke_role(
...
+    @staticmethod
+    async def _resolve_invitation_identifiers(
+        session: SASession,
+        identifiers: list[str],
+    ) -> dict[str, uuid.UUID]:
+        """Resolve usernames or emails to user UUIDs (ACTIVE users only).
+
+        Each identifier is matched against both ``username`` and ``email``.
+        If more than one ACTIVE user matches a given identifier the entry is
+        silently skipped (ambiguous).
```
seedspirit: While both the `username` and `email` are unique, does "duplication" refer to a situation where User A's username and User B's email address overlap? Also, wouldn't it be better for handling the situation going forward to return an explicit error and prompt the user to contact the administrator, rather than silently skipping the validation?

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
```diff
@@ -1083,3 +1110,183 @@ async def bulk_revoke_role(
...
+
+        Already canceled invitations are silently returned (idempotent).
+        Raises ObjectNotFound if not found.
+        Raises GenericBadRequest if accepted or rejected.
+        """
+        cond_id = RoleInvitationConditions.by_id(invitation_id)
+        stmt = sa.select(RoleInvitationRow).where(cond_id())
+        row = await session.scalar(stmt)
+        if row is None:
+            raise ObjectNotFound(object_name="RoleInvitation")
```
seedspirit: Please change it to Custom Error

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
```diff
@@ -1083,3 +1110,183 @@ async def bulk_revoke_role(
...
+        if row.state == RoleInvitationState.REJECTED:
+            return row
+        if row.state != RoleInvitationState.PENDING:
+            raise GenericBadRequest(
+                f"Cannot reject: invitation is {row.state.value}, expected pending"
+            )
+        return row
```
seedspirit: It appears that Bad Request handling applies only when the state is `PENDING`; for all other cases, it can be changed to a row return.

src/ai/backend/manager/models/alembic/versions/ad7acfe8aa1c_add_role_invitations_table.py
```diff
@@ -0,0 +1,244 @@
...
+
+def upgrade() -> None:
+    op.create_table(
+        "role_invitations",
+        sa.Column("id", GUID, primary_key=True, server_default=sa.text("uuid_generate_v4()")),
+        sa.Column(
+            "inviter_user_id",
+            GUID,
+            nullable=True,
+        ),
```
seedspirit, line 193: Are there cases where the `inviter` user ID is null? Also, is there a reason why only this column lacks a foreign key constraint?

seedspirit (changes requested):

src/ai/backend/manager/repositories/role_invitation/types.py
```diff
@@ -0,0 +1,61 @@
...
+
+        def inner() -> sa.sql.expression.ColumnElement[bool]:
+            return RoleInvitationRow.invitee_user_id == invitee_user_id
+
+        return inner
+
+    @property
+    def existence_checks(self) -> Sequence[ExistenceCheck[UUID]]:
+        return []
+
```
jopemachine, line 44: Is there a reason it returns an empty list?

src/ai/backend/manager/repositories/permission_controller/repository.py
```diff
@@ -336,3 +348,41 @@ async def check_permission_with_scope_chain(
...
+        self,
+        action: CreateRoleInvitationByUsernameAction,
+    ) -> CreateRoleInvitationResult:
+        return await self._db_source.create_invitation_by_username(action)
+
+    @permission_controller_repository_resilience.apply()
+    async def search_invitations(
+        self,
+        querier: BatchQuerier,
+        scope: SearchScope | None = None,
```
jopemachine: Why not creating SearchScope?

src/ai/backend/manager/actions/action/rbac_role_invitation.py
fregataa: This is an intentional design, it supports users to send invitation without knowing other user's id.
Imagine a case of project admin invite a project role to a user outside of its project. How does the admin know the user's id?
There is another plan to impl grant role feature.
Please check this issue. https://github.com/lablup/backend.ai/issues/10825

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
fregataa: user1 - `username: UserA, email: userA@email.com`
user2 - `username: userA@email, email: user@mail.com`
silent skipping is intentional. if it raises error, the client can abuse this API

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
fregataa: It is weird to accept such transition

src/ai/backend/manager/actions/action/rbac_role_invitation.py
jopemachine: > This is an intentional design, it supports users to send invitation without knowing other user's id. Imagine a case of project admin invite a project role to a user outside of its project. How does the admin know the user's id? There is another plan to impl grant role feature. Please check this issue. #10825

It would be better to rename CreateRoleInvitationByUsernameAction to CreateRoleInvitationAction and separate invitee_usernames and invitee_emails into distinct fields.

Or we could split them into CreateRoleInvitationByUsernameAction and CreateRoleInvitationByUserEmailAction.

fregataa: 1. There is another plan to impl invite users by user id - e.g. superadmin invites users to a new role. That case, action name should be plain `CreateRoleInvitationAction`
2. I intentionally allow username or email because users usually don't know whether "a certain name value" is username or email

jopemachine: I don’t think the fact that users may confuse usernames and emails means we should mix them in the code. It would be better to handle them as separate fields, since they are distinct types of data.

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
```diff
@@ -346,16 +372,22 @@ async def purge_role(self, purger: Purger[RoleRow]) -> RoleRow:
...
+                user_id=data.user_id,
+                role_id=data.role_id,
+                granted_by=data.granted_by,
             )
-            result = await execute_creator(db_session, creator)
-            await self._sync_user_scopes_on_assign(db_session, [data.user_id])
-            return result.row
+        )
+        result = await execute_creator(db_session, creator)
+        await PermissionDBSource._sync_user_scopes_on_assign(db_session, [data.user_id])
```
seedspirit:
```suggestion
        await self._sync_user_scopes_on_assign(db_session, [data.user_id])
```

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
```diff
@@ -1172,3 +1204,192 @@ async def bulk_revoke_role(
...
+    async def cancel_invitation(
+        self,
+        action: CancelRoleInvitationAction,
+    ) -> RoleInvitationData:
+        async with self._db.begin_session_read_committed() as session:
+            row = await self._get_invitation_to_cancel(session, action.invitation_id)
+            row.state = RoleInvitationState.CANCELED
+            await session.flush()
+            await session.refresh(row)
+            return row.to_data()
```
seedspirit: how about pattern like

```python
stmt = (
      sa.update(RoleInvitationRow)
      .where(
          RoleInvitationRow.id == action.invitation_id,
          RoleInvitationRow.state.in_([RoleInvitationState.PENDING, RoleInvitationState.REJECTED]),
      )
      .values(state=RoleInvitationState.REJECTED)
      .returning(RoleInvitationRow)
  )
  row = (await session.execute(stmt)).scalar_one_or_none()
  if row is None:
      # distinct not found vs invalid state with one more select
      ...
  return row.to_data()
```

seedspirit: or we can generalize accept state transition like this

```python
async def _transition_invitation_state(
        session: SASession,
        invitation_id: uuid.UUID,
        *,
        target_state: RoleInvitationState,
        allowed_states: tuple[RoleInvitationState, ...],
        operation: str,
    ) -> RoleInvitationRow:
        """Transition an invitation state atomically and return the updated row."""
        update_stmt = (
            sa.update(RoleInvitationRow)
            .where(
                RoleInvitationRow.id == invitation_id,
                RoleInvitationRow.state.in_(allowed_states),
            )
            .values(state=target_state)
            .returning(RoleInvitationRow)
        )
        stmt = sa.select(RoleInvitationRow).from_statement(update_stmt)
        row = cast(RoleInvitationRow | None, (await session.execute(stmt)).scalar_one_or_none())
        if row is None:
			current_state = await session.scalar(
            	sa.select(RoleInvitationRow.state).where(RoleInvitationRow.id == invitation_id)
        	)
        	if current_state is None:
            	raise RoleInvitationNotFound()
        	raise RoleInvitationInvalidState(
            	f"Cannot {operation}: invitation is {current_state.value}, expected pending"
        	)
        
		return row
```

seedspirit (changes requested):

seedspirit (approved):

PR #11189, opened 2026-04-19
feat(BA-5776): implement bulk permission check query

fregataa:
## Summary
- Add `BatchPermissionCheckInput` data type and `check_batch_permission_with_scope_chain()` that validates multiple entities of the same `RBACElementType` in a single DB query
- Uses a recursive CTE carrying `entity_id` through scope chain traversal, preserving the same two-layer semantics (scope chain + self-scope) as the single-entity version
- Exposed via `PermissionControllerRepository` with the standard resilience decorator

## Test plan
- [x] Empty input returns empty dict
- [x] All granted via project scope (shared AUTO edges)
- [x] All granted via domain chain (2-level traversal)
- [x] No permission → all denied
- [x] Mixed edges partial grant (AUTO/REF/none)
- [x] Self-scope grants individually
- [x] Operation mismatch → all denied
- [x] Inactive role → all denied
- [x] Cross-scope entity type override
- [x] Self-scope + chain combined
- [x] Multi-project two-domain isolation
- [x] Other user permission isolation

Resolves BA-5776

### DIFF
--- a/changes/11189.feature.md
+++ b/changes/11189.feature.md
@@ -1,1 +1,1 @@
-Add batch scope-chain permission check query for validating multiple entities in a single DB round-trip.
+Add bulk scope-chain permission check query for validating multiple entities in a single DB round-trip.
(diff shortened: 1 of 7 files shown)


seedspirit (approved):
It seems good, but I think @HyeockJinKim have to review once more

PR #11191, opened 2026-04-20
feat(BA-5777): add bulk RBAC filtering infrastructure

fregataa:
## Summary
- Add bulk RBAC filtering infrastructure under `manager/actions/`: `BaseBulkAction[T]` (generic dataclass with `entity_ids: list[str]` + abstract `typed_entity_ids()`), `BulkActionValidator` returning `BulkValidationResult(allowed_entity_ids, denied_entities)`, and `BulkActionProcessor` that narrows the action between validators and returns `BulkProcessResult(result, validator_decisions)` for partial-success responses.
- Bulk actions intentionally carry only `entity_ids` so the processor can reconstruct a narrowed action via `type(action)(entity_ids=...)` — no factory hook.
- Remove the orphan `check_and_transit_status_multi` path and 10 unused intermediate `*BulkAction` base classes across services.
- Follow-up: wire `BulkActionRBACValidator.validate()` to `PermissionControllerRepository.check_bulk_permission_with_scope_chain()` (BA-5776).

Commits:
1. Add infrastructure.
2. Rename `Batch*` → `Bulk*` to match updated BA-5777 wording.
3. Address review: `BulkValidatorDecision` → `ValidatorDecision`, replace `_validator_scope` CM with a plain async helper, drop dead `name` param in test validator.

## Test plan
- [x] `pants fmt / lint / check`
- [x] `pants test tests/unit/manager/actions/test_bulk_processor.py`
- [x] CI green

Resolves BA-5777


src/ai/backend/manager/actions/processor/bulk.py
```diff
@@ -0,0 +1,142 @@
...
+    BulkValidationResult,
+)
+
+from .base import ActionRunner
+
+log = BraceStyleAdapter(logging.getLogger(__spec__.name))
+
+
+@dataclass(frozen=True)
+class BulkValidatorDecision:
```
jopemachine: It seems it is actually SingleValidatorDecision

src/ai/backend/manager/actions/processor/bulk.py
```diff
@@ -0,0 +1,142 @@
...
+    _runner: ActionRunner[TBulkAction, TBulkActionResult]
+
+    def __init__(
+        self,
+        func: Callable[[TBulkAction], Awaitable[TBulkActionResult]],
+        monitors: Sequence[ActionMonitor] | None = None,
+        validators: Sequence[BulkActionValidator] | None = None,
+    ) -> None:
+        self._runner = ActionRunner(func, monitors)
+
```
jopemachine, line 69: Nit: For consistency, it seems better to keep monitors as an empty list as well.

src/ai/backend/manager/actions/processor/bulk.py
```diff
@@ -0,0 +1,142 @@
...
+        self._validators = validators or []
+
+    @asynccontextmanager
+    async def _validator_scope(
+        self,
+        validator: BulkActionValidator,
+        action: TBulkAction,
+        meta: BaseActionTriggerMeta,
+    ) -> AsyncIterator[BulkValidationResult]:
+        """Run one validator inside a bookend scope.
```
jopemachine: Nit: bookend seems typo

src/ai/backend/manager/actions/processor/bulk.py
```diff
@@ -0,0 +1,142 @@
...
+            ) as validation:
+                decisions.append(
+                    BulkValidatorDecision(
+                        validator_name=validator.name(),
+                        results=validation,
+                    )
+                )
+                current_action = self._process_action(current_action, validation)
+
+        action_result = await self._runner.run(current_action, action_trigger_meta)
```
jopemachine: Could you explain this code?
I don't understand why context manager is needed, and why current_action isn’t used inside the loop but instead run is called after the async with block.

tests/unit/manager/actions/test_bulk_processor.py
```diff
@@ -0,0 +1,242 @@
...
+
+    @override
+    def entity_ids(self) -> list[str]:
+        return list(self.processed_ids)
+
+
+class _AllowSetValidator(BulkActionValidator):
+    """Approves any ID in ``allowed``; anything else visible is denied."""
+
+    def __init__(self, allowed: set[str], name: str = "allow-set") -> None:
```
jopemachine: I don't think name is necessory for the constructor

src/ai/backend/manager/actions/processor/bulk.py
fregataa: SingleValidatorDecision does not match to the processor name

src/ai/backend/manager/actions/processor/bulk.py
fregataa: it is not

src/ai/backend/manager/actions/processor/bulk.py
```diff
@@ -0,0 +1,138 @@
...
+        validation = await validator.validate(action, meta)
+        duration = (datetime.now(UTC) - started_at).total_seconds()
+        log.debug(
+            "bulk validator {} saw {} ids, denied {} in {:.3f}s",
+            validator.name(),
+            len(action.entity_ids),
+            len(validation.denied_entities),
+            duration,
+        )
+        return validation
```
seedspirit: remove

src/ai/backend/manager/actions/processor/bulk.py
```diff
@@ -0,0 +1,138 @@
...
+        Returns the incoming action unchanged when the validator denied
+        nothing; otherwise constructs a fresh instance of the same class
+        via its ``entity_ids``-only constructor so the original stays
+        immutable.
+        """
+        if not validation.denied_entities:
+            return current_action
+        allowed_set = set(validation.allowed_entity_ids)
+        filtered_ids = [eid for eid in current_action.entity_ids if eid in allowed_set]
+        return type(current_action)(entity_ids=filtered_ids)
```
seedspirit: how about just filter entity ids, no need to process action

src/ai/backend/manager/actions/processor/bulk.py
```diff
@@ -0,0 +1,138 @@
...
+
+        for validator in self._validators:
+            validation = await self._run_validator(validator, current_action, action_trigger_meta)
+            decisions.append(
+                ValidatorDecision(
+                    validator_name=validator.name(),
+                    results=validation,
+                )
+            )
+            current_action = self._process_action(current_action, validation)
```
seedspirit: Instead of reassigning `current_action`, assigning it to a variable with a new name, such as `resolved_action`, would likely make the code less confusing to read.

seedspirit (approved):

PR #11197, opened 2026-04-21
docs(BA-5782): Write BEP for Scoped App Config Redesign

jopemachine:
resolves BA-5782

BEP-1052: Scoped App Config Redesign — proposes a redesigned GraphQL/REST surface for the WebUI's app configuration (anonymous theme/branding, authenticated user defaults / personal settings, admin everything) so that the three audiences are expressed as **separate scoped entities** instead of per-field access control on a single entity.

## Schema

- **`app_configs` table** keyed by `(scope_type, scope_id, name)` — a scope holds many *named documents* (e.g. `theme`, `menu`, `branding`), not one. `extra_config` is the only payload column. `status` (`ALIVE` / `DELETED`) provides soft-delete; record removal flips status, never drops the row. `name` is restricted to `^[A-Za-z0-9_.\-]+$` (1–128) so it rides in REST path segments and `MyAppConfig.id` without escaping.
- **Scopes** (`AppConfigScopeType`):
  - `PUBLIC` — pre-login readable (anonymous)
  - `DOMAIN` — admin-enforced domain policy; same-domain users may read, admin writes, **not in the merge**
  - `DOMAIN_USER_DEFAULTS` — per-user default values admins provide that users may override; participates in the merge as the base
  - `USER` — the user's customized value
- **Read merge** (per-user only, exposed on the new `MyAppConfig` type): **2-way deep merge** `DOMAIN_USER_DEFAULTS ⊕ USER` (low → high priority). Views: `userCustomizedConfig`, `domainDefaultConfig`, `mergedConfig`. Arrays are leaves — the higher-priority side replaces the array wholesale (element-wise array merge has no unambiguous semantics). `DOMAIN` and `PUBLIC` are not in the merge — they have their own read paths. A single SQL query resolves the caller's `domain_name` via a `users` subquery and fetches both merge-source rows, bounded to at most 2 rows by the natural-key UniqueConstraint.

## Types

- `PublicAppConfig`, `DomainAppConfig`, `DomainUserDefaultsAppConfig`, `UserAppConfig` — all mirror the same shape (`id`, `name`, `config`, timestamps). Each is a thin view over an `app_configs` row for its scope.
- `MyAppConfig` — the per-user **resolved / merged view** for the current caller. Implements `Node` via a server-encoded composite id `base64("MyAppConfig:{user_id}:{name}")`. `node(id)` resolver returns the merged view only when `decoded.user_id == current_user.id` (or the caller is admin).
- `AppConfig` — generic raw-row type returned by admin mutations, `adminAppConfigs`, and `node(id)` for non-`MyAppConfig` rows.

## GraphQL

- Per-scope typed Connections — `publicAppConfigs`, `myAppConfigs`, `DomainV2.appConfigs`, `UserV2.appConfigs`, plus admin `adminAppConfigs` for cross-scope search.
- Unified `AppConfigFilter` / `AppConfigOrderBy` shared by every Connection, with `AND` / `OR` / `NOT` composition. `filter.status` opts into `DELETED` rows for admin recovery / audit flows. `scopeType` / `scopeId` filter fields are no-op on per-scope Connections.
- **Writes are bulk-only** (8 mutations total, 4 admin × 4 my). Each item runs in its own DB transaction — partial success; payload carries a success list + `failed` list. Single-item callers pass a 1-element array.
  - **Admin path** — `adminBulk{Create,Update,Delete,Restore}AppConfigs`. Items carry `AppConfigKey { scopeType, scopeId, name }`; any scope can be mixed in one call. Admin-only. Returns lists of raw `AppConfig`.
  - **My path** — `bulk{Create,Update,Delete,Restore}MyAppConfigs`. Items carry only `name`; server injects `scopeType = USER` + `scopeId = current_user.user_id`. Returns lists of merged `MyAppConfig`.
- `create` — strict insert (fails the item if any row exists for the key, soft-deleted included); `update` — replace the existing `ALIVE` row wholesale; `delete` — soft-delete (`ALIVE → DELETED`); `restore` — explicit inverse (`DELETED → ALIVE`, value unchanged). No upsert.
- Permissions (service layer):
  - `publicAppConfigs` — anonymous OK.
  - `myAppConfigs` — authenticated self.
  - `DomainV2.appConfigs` — same-domain users or admin.
  - `UserV2.appConfigs` — owner or admin.
  - `adminAppConfigs`, `admin*AppConfigs` mutations — admin-only regardless of item scope.
  - `node(id) → MyAppConfig` — self (via decoded `user_id`) or admin.

## REST

Aligned with `src/ai/backend/manager/api/rest/v2/CLAUDE.md` conventions under the existing `app-configs` registry.

- **Reads** use the scope-parameterized template:
  - `GET /v2/app-configs/{scope_type}/{scope_id}[/{name}]` — per-scope permissions match the GQL matrix (public: anonymous; domain / domain_user_defaults: same-domain users; user: self or admin).
  - `GET /v2/app-configs/my[/{name}]` — self-service; response is the merged `MyAppConfig` snake_case projection.
- **Writes are bulk-only**:
  - `POST /v2/app-configs/bulk-{create,update,delete,restore}` — admin.
  - `POST /v2/app-configs/my/bulk-{create,update,delete,restore}` — self.
  - Request / response bodies mirror the GQL input / payload in snake_case (partial success).
- `POST /v2/app-configs/search` — admin cross-scope search (same body schema as `adminAppConfigs`).

## Client integration

§6 documents the WebUI bootstrap flow. The set of pre-login public documents is owned by the frontend (hard-coded in the WebUI bundle) — the server does not publish a bootstrap list. Two-step flow: `publicAppConfigs` queries pre-login; single `myAppConfigs` query post-login (admins use the same for their own session; DOMAIN policy is fetched separately from the admin path if needed).

See the BEP for the full schema, permission matrix, and 7 caller scenarios (S1–S7).

----
📚 Documentation preview 📚: https://sorna--11197.org.readthedocs.build/en/11197/

----
📚 Documentation preview 📚: https://sorna-ko--11197.org.readthedocs.build/ko/11197/


proposals/BEP-1052-scoped-app-config-redesign.md
```diff
@@ -0,0 +1,876 @@
...
+}
+```
+
+- Server: `check_admin_only()` → locates the DOMAIN row by id →
+  replaces `config` and `userAppConfigDefaults` wholesale.
+- Effect: on the next `myAppConfig` call, every user in that domain
+  receives the new defaults merged with their own
+  `userCustomizedConfig`.
+
+### S5. Admin edits a specific user's config on their behalf
```
seedspirit: It seems like a User could reset the configuration set by the Admin using a mutation like `updateMyAppConfig`. Is this acceptable?

jopemachine: According to the requirements document, users should be able to modify their own app configuration. There was no requirement stating that administrators must be able to enforce specific settings.

fregataa (commented):
Domain config and User config are stored as opaque JSON in the same table. Since the backend enforces no schema, nothing prevents the same key from appearing in both scopes.
Is it entirely the client's responsibility to decide which scope a key belongs to and how to resolve conflicts? If so, it would be good to document it explicitly.

jopemachine:
> Domain config and User config are stored as opaque JSON in the same table. Since the backend enforces no schema, nothing prevents the same key from appearing in both scopes. Is it entirely the client's responsibility to decide which scope a key belongs to and how to resolve conflicts? If so, it would be good to document it explicitly.

Domain config and user config will not conflict because their scope_id and scope_type do not overlap. Please refer to the app_configs table.

fregataa (commented):
1. UserAppConfig -> UserAppConfig and ResolvedAppConfig
2. It would be good to determine what kind of configs will be stored in domain and domain_user_default configs
3. Purpose of public config: Only for login page?
4. Merge conflict(or priority) between domain/domain_user_default/user config. Plus, how should public config be handled in merge process.
5. Upsert config seems ok
6. How should list (array) objects be merged?
7. Client should handle 404 from query result of public app config

proposals/BEP-1052-kr.md
```diff
@@ -0,0 +1,1627 @@
...
+  """
+  현재 사용자의 soft-deleted USER 문서를 복구 (`DELETED → ALIVE`).
+  저장 값은 그대로 유지 — 값을 바꾸려면 이어서 `updateMyAppConfig`
+  호출. row가 없거나 이미 `ALIVE`이면 에러. 응답은 재계산된
+  `MyAppConfig`.
+  """
+  restoreMyAppConfig(input: RestoreMyAppConfigInput!): RestoreMyAppConfigPayload!
+}
+
+enum AppConfigScopeGQL {
```
seedspirit: nitpick: How about `AppConfigScopeTypeGQL`?

proposals/BEP-1052-kr.md
```diff
@@ -0,0 +1,1627 @@
...
+동일한 식별 모델을 GQL 입력으로 노출.
+- `PUBLIC`:               `scopeId`는 고정 문자열 `"public"`.
+- `DOMAIN`:               `scopeId`는 `domain_name`.
+- `DOMAIN_USER_DEFAULTS`: `scopeId`는 `domain_name`.
+- `USER`:                 `scopeId`는 `user_id` (UUID string).
+- `name`은 문서 이름 (스코프 안에서 unique).
+"""
+input AppConfigKey {
+  scope: AppConfigScopeGQL!
+  scopeId: String!
```
seedspirit: nitpick: `scopeType` and `scopeId`

jopemachine:
1. Split into two types: UserAppConfig (raw USER-row view) and MyAppConfig (the merged view).                                                       
                                                                                     
  2. Made domain readable by same-domain users and took it out of the myAppConfigs merge.                                                             
                                                                                                                                                      
  3. It's the schema for content accessible without authentication.                                                                                   
                                                                                     
  4. In the merge, user_app_config has higher priority than domain_user_defaults_app_config; domain and public do not participate in the merge.       
                                                                                     
  5. Both are fine, but the UI I had in mind has a separate create mutation.                                                                          
                                                                                     
  6. Lists are treated as leaves — the higher-priority side replaces the array wholesale (element-wise merge has no unambiguous semantics around      
  stable identity / ordering).                                                       
                                                                                                                                                      
  7. GraphQL publicAppConfigs returns empty edges on missing documents while REST GET /v2/app-configs/public/public/{name} returns 404, so the client 
  needs to handle both.   @fregataa

HyeockJinKim (approved):

PR #11070, opened 2026-04-14
fix(BA-5720): handle null DeviceRequests in CUDA container measures

seedspirit:
resolve #11065 (BA-5720)

## Summary
- Docker API can return `null` for `HostConfig.DeviceRequests` (key exists but value is `None`)
- `.get("DeviceRequests", [])` only provides a default when the key is **missing**, not when the value is `None` → `TypeError: 'NoneType' object is not iterable`
- Fix: `.get("DeviceRequests") or []` to handle both missing key and `None` value

## Test plan
- [ ] Verify no error when `DeviceRequests` is `None`
- [ ] Verify existing behavior when `DeviceRequests` is a normal list

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11209, opened 2026-04-21
fix(BA-5774): skip already-attached networks to avoid duplicate-attach 403

fregataa:
## Summary

- Before calling `network.connect()` for each entry in `docker-networks`, inspect the container and skip networks Docker has already attached (e.g. `bridge` auto-attached when `NetworkMode` is unset).
- Keep a narrow `DockerError` catch for `403 "already exists"` around `connect()` as a defense against a `container.show()` ↔ `connect()` race.
- Lets a single accelerator plugin `docker-networks` config (e.g. `["bridge", "macvlan-roce-0", "macvlan-roce-1"]`) serve both single-node and multi-node sessions.

Resolves **BA-5774**.

### Background

When an accelerator plugin config (e.g. `cuda.toml`) lists `bridge` in `docker-networks`, **single-node** session creation failed every time with:

```
DockerError(403, 'endpoint with name kernel.<img>.<kid> already exists in network bridge')
```

Root cause: after container start, the agent iterated the `docker-networks` list and called `network.connect()` for each entry **without checking whether the container was already attached**. Docker auto-attaches `bridge` when `NetworkMode` is unset (single-node), so the explicit `connect()` duplicated the attach.

Removing `bridge` from the list fixed single-node but broke multi-node sessions, where `bridge` is needed for control-plane connectivity because overlay is used as the primary `NetworkMode`. A single config could not cover both modes — this patch fixes that.

## Test plan

- [ ] Single-node session creation succeeds with `docker-networks = ["bridge", "macvlan-roce-0", "macvlan-roce-1"]` (no more 403)
- [ ] Multi-node (overlay) session creation still attaches `bridge` successfully for control-plane connectivity
- [ ] Session with `docker-networks` containing only non-auto-attached networks still connects correctly
- [x] Unit/integration tests: `pants test ::` on `src/ai/backend/agent/docker/`

### DIFF
--- a/src/ai/backend/agent/docker/agent.py
+++ b/src/ai/backend/agent/docker/agent.py
@@ -1325,9 +1325,13 @@ class DockerKernelCreationContext(AbstractKernelCreationContext[DockerKernel]):
             # serve both single-node and multi-node without duplicate-attach 403s.
             if additional_network_names:
                 container_info = await container.show()
-                already_attached_networks = set(
-                    container_info.get("NetworkSettings", {}).get("Networks", {}).keys()
-                )
+                networks = container_info.get("NetworkSettings", {}).get("Networks", {}) or {}
+                # `get_docker_networks()` may return either network names or IDs
+                # (see `Resources.get_docker_networks` contract), so collect both
+                # the dict keys (names) and `NetworkID` values to filter against.
+                already_attached_networks = set(networks.keys()) | {
+                    n["NetworkID"] for n in networks.values() if n.get("NetworkID")
+                }
             else:
                 already_attached_networks = set()
 
--- a/src/ai/backend/agent/docker/agent.py
+++ b/src/ai/backend/agent/docker/agent.py
@@ -1320,9 +1320,24 @@ class DockerKernelCreationContext(AbstractKernelCreationContext[DockerKernel]):
                 n = await self.computers[dev_name].instance.get_docker_networks(device_alloc)
                 additional_network_names |= set(n)
 
-            for name in additional_network_names:
+            # Skip networks already attached by Docker (e.g. `bridge` auto-attached
+            # when NetworkMode is unset). A single `docker-networks` config can then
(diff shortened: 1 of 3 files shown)


src/ai/backend/agent/docker/agent.py
```diff
@@ -1042,6 +1042,37 @@ async def _apply_seccomp_profile(self, container_config: MutableMapping[str, Any
...
+            return
+        container_info = await container.show()
+        networks = container_info.get("NetworkSettings", {}).get("Networks", {}) or {}
+        already_attached = set(networks.keys()) | {
+            n["NetworkID"] for n in networks.values() if n.get("NetworkID")
+        }
+        for ref in requested - already_attached:
+            network = await docker.networks.get(ref)
+            try:
+                await network.connect({"Container": container._id})
```
seedspirit, line 1068: nitpick: 
```suggestion
                await network.connect({"Container": container.id})
```

src/ai/backend/agent/docker/agent.py
```diff
@@ -1042,6 +1042,37 @@ async def _apply_seccomp_profile(self, container_config: MutableMapping[str, Any
...
+        Connect the container to each requested network, skipping any that
+        Docker has already attached (e.g. `bridge` auto-attached when
+        `NetworkMode` is unset). `requested_networks` may contain network
+        names or IDs per the `Resources.get_docker_networks` contract.
+        """
+        requested = set(requested_networks)
+        if not requested:
+            return
+        container_info = await container.show()
+        networks = container_info.get("NetworkSettings", {}).get("Networks", {}) or {}
```
seedspirit, line 1061: Does this work safely? I’ve had instances before where `container.show()` failed because some keys had `None` as their value, so I’d appreciate it if you could check this.

https://github.com/lablup/backend.ai/pull/11070

fregataa: it works in my case but it seems good to cover such case

seedspirit (approved):

PR #11215, opened 2026-04-22
ci(BA-5855): Auto-correct mismatched PR numbers in news fragments

### LINKED ISSUE #11214: Validate news-fragment PR number matches current PR in assign-pr-number workflow
## Main idea

`scripts/assign-pr-number.py` currently handles two cases for files under `changes/`:

1. **Unnumbered** (e.g., `.fix.md`, `fix.md`) → renamed to `<PR#>.fix.md`.
2. **Already numbered** (e.g., `11170.fix.md`) → left untouched if the number matches the current PR, silently ignored otherwise.

The second case is a gap. When an author (or an AI coding agent) writes `changes/9999.fix.md` using an **issue number** — or any stale/wrong PR number — the workflow treats it as valid and merges it as-is. The wrong number then appears in the release notes and links to an unrelated PR/issue.

The current numbering automation was introduced in #989, which covered only the unnumbered case.

**Proposal:** In `scripts/assign-pr-number.py`, when `rx_numbered_fragment` matches but the prefix number ≠ the current PR number, rename the file to `<current-PR#>.<type>.md` (same git-mv path as the unnumbered case) instead of the current no-op branch. This auto-heals the common authoring mistake and keeps the news fragment convention (filename = PR number) self-enforcing.

Concretely, in the existing loop around `scripts/assign-pr-number.py:33-39`, the `rx_numbered_fragment` branch should add an `else` that appends the file to the rename list when its leading number differs from `pr_number`, reusing the `renamed_pairs` machinery at lines 41-52.

Also update `changes/README.md` to make the naming rule explicit: the filename number must be the **PR number** (not an issue number, not a Linear/ticket number), and unnumbered drafts like `.fix.md` are acceptable — CI will fill in the PR number on push.

## Alternative ideas

- **Fail loudly instead of auto-rename:** exit non-zero with a message like `changes/9999.fix.md has a number that does not match PR #11170`. Safer if we want authors to notice, noisier in practice.
- **Add a `changes/CLAUDE.md` for AI agents:** soft guidance only — doesn't catch human typos or copy-pasted issue links, so it should supplement, not replace, the workflow fix.

## Anything else?

- Exempted files (`README.md`, `template.md`) must continue to be skipped.
- Consider whether branch-backport PRs (where the fragment was already numbered for the original PR on `main`) should be rewritten to the backport PR's number or preserved — current behavior preserves; the proposal would rewrite. Worth confirming with maintainers before implementing.

## Related

- #989 — original workflow that introduced PR-number automation for news fragments.

rapsealk:
## Summary

- Extend `scripts/assign-pr-number.py` so the `assign-pr-number` workflow also rewrites news fragments whose filename prefix does not match the current PR number (e.g., `9999.fix.md` written with an issue number instead of a PR number), not only unnumbered placeholders like `.fix.md`.
- Scope the rewrite to **fragments added or modified by this PR** (computed via `git diff --diff-filter=AM origin/$GITHUB_BASE_REF...HEAD`), so historical fragments accumulated on the base branch from past merged PRs are never touched.
- Bump the workflow's `fetch-depth` from `2` to `0` so the base ref is reachable for the diff.
- Document the filename convention in `changes/README.md`: the leading number must be the **pull request number**, and unnumbered drafts are acceptable because CI will fill it in on push.

Resolves #11214 (BA-5855). Builds on the numbering automation originally introduced in #989.

## Behavior

For files in `changes/`, running `python scripts/assign-pr-number.py <PR>` under CI (with `GITHUB_BASE_REF` set):

| File                                                  | Before                       | After                                                                |
| ----------------------------------------------------- | ---------------------------- | -------------------------------------------------------------------- |
| `.fix.md` (added by this PR, unnumbered)              | rename → `<PR>.fix.md`       | rename → `<PR>.fix.md` (unchanged)                                   |
| `<PR>.fix.md` (added by this PR, correct)             | no-op                        | no-op (unchanged)                                                    |
| `9999.fix.md` (added by this PR, wrong number)        | no-op (silently merged wrong)| **rename → `<PR>.fix.md`** (new)                                     |
| `11170.fix.md` (from a past merged PR, on base branch)| no-op                        | no-op (explicitly preserved by the diff-based scope)                 |

Local runs without `GITHUB_BASE_REF` fall back to the pre-existing behavior: rename unnumbered placeholders, leave all numbered fragments alone.

## Test plan

- [x] Simulated historical fragment (`11170.fix.md`) present on base + this-PR's `.doc.md`, `12345.feature.md`, `9999.fix.md` — only the unnumbered and the mismatched fragments are renamed; the historical one is left alone.
- [x] Simulated the local-run fallback (no `GITHUB_BASE_REF`) — mismatched fragment is preserved; only unnumbered is renamed.
- [x] Simulated the no-op path (matching fragment only) — prints `already exists`, emits `has_renamed_pairs=false`.
- [x] `pants --changed-since=HEAD --changed-dependents=transitive fmt lint check` passes.
- [x] CI's `assign-pr-number` job ran on this PR's first push and correctly renamed `changes/.misc.md` → `changes/11215.misc.md`.

## Notes for reviewers

One open question: **backport PRs**. A backport to a release branch typically lands with the fragment already numbered after the original `main` PR (e.g., copying `11170.fix.md` into the backport branch). Because backport PRs *do* touch `changes/11170.fix.md` in their diff, this change would rewrite that prefix to the backport PR's number. If the project convention is to preserve the original PR number in release-branch fragments, the rewrite should be gated further (e.g., only when `GITHUB_BASE_REF == "main"`). Current behavior on `main` is unaffected.


rapsealk:
## Review

The scoping via `git diff ... --diff-filter=AM` successfully protects historical fragments (verified: `git diff --name-status origin/main...HEAD` on this branch shows zero deletions, only 4 A/M entries) and the new self-healing path closes a real authoring footgun. However, the mismatched-rename loop in `scripts/assign-pr-number.py` silently overwrites on collision, which is a production blocker I cannot wave through.

### Strengths

- Scoping is the right design — `origin/$GITHUB_BASE_REF...HEAD` (three-dot, against the merge-base) is the correct git incantation, and restricting the new rename-mismatch path to files inside `pr_touched_fragments` keeps the blast radius small.
- Local-run fallback (`base_ref = os.getenv(...) or ""` → `pr_touched_fragments = None`) preserves the prior behavior exactly: unnumbered files get renamed, mismatched loop is skipped. Good conservatism.
- The two-loop split (existing unnumbered loop untouched, new mismatched loop gated on `pr_touched_fragments is not None`) means the original unnumbered path has no regression risk.
- `changes/README.md` update is accurate and matches the new behavior.
- CI loop termination is safe: after the auto-commit renames `<x>.type.md` → `<PR>.type.md`, the re-triggered workflow sees it as correctly numbered and no-ops. No infinite commit loop.

### Findings

**Blocker — silent data loss on rename collision** (`scripts/assign-pr-number.py:91-100`).
`Path.rename()` on POSIX silently overwrites an existing target. The new mismatched loop does not check for target existence:

```python
for original_filename, news_type in mismatched_fragments:
    numbered_filename = f"{pr_number}.{news_type}.md"
    file_path = base_path / original_filename
    file_path.rename(base_path / numbered_filename)
```

Three concrete collision cases, all producing silent content loss:

1. PR contains both `11215.fix.md` (correct, lands in `existing_fragments`) and `9999.fix.md` (mismatched). The mismatched loop renames `9999.fix.md` → `11215.fix.md`, **overwriting the correct fragment**.
2. PR contains two mismatched fragments of the same type, e.g. `9999.fix.md` and `8888.fix.md`. Both want to become `11215.fix.md`; the second rename silently clobbers the first.
3. PR contains `.fix.md` (unnumbered) and `9999.fix.md` (mismatched). The unnumbered loop creates `11215.fix.md`; then the mismatched loop overwrites it.

I just verified the overwrite behavior locally (`Path.rename` on an existing target on Darwin/Linux drops the destination without error). Please either (a) use `os.rename` with an explicit `exists()` pre-check and `sys.exit(1)` with a clear message, or (b) fold both loops into one and coalesce by `news_type` before renaming, keeping the first and warning about duplicates. Given this script silently deleted 16 historical fragments in an earlier iteration of the PR, an explicit guard here is warranted.

**Should-fix — follow-up PRs that modify a historical fragment get renumbered** (`scripts/assign-pr-number.py:54-65`).
`--diff-filter=AM` includes `M` (modified). If a later PR legitimately edits `changes/11170.fix.md` (e.g., to fix a typo in a merged fragment before release), `pr_touched_fragments` will contain it, the prefix `11170` won't match that PR's number, and the script will rename it to `<new-PR>.fix.md` — destroying the original PR-number link. Two reasonable mitigations: (a) only rewrite when the diff status is `A` (added), not `M`; or (b) only rewrite when the file did not exist at the base commit (`git cat-file -e origin/$BASE:changes/<file>` → fail). Option (a) is simpler and matches the stated intent ("authors (or AI agents) writing a wrong number on a new fragment").

**Should-fix — unguarded `subprocess.run(..., check=True)` in `get_pr_touched_fragments`** (`scripts/assign-pr-number.py:26-41`).
If `origin/$GITHUB_BASE_REF` is not fetched (shallow clone regression, `actions/checkout` config change, someone running the script locally after `git fetch --depth=1`), `check=True` will raise `CalledProcessError` with a stderr dump and no actionable message for the user. Catch `CalledProcessError`, log a clear "base ref not fetched — did you set `fetch-depth: 0`?", and either fall back to `None` (local-style behavior) or `sys.exit(1)` deliberately.

**Should-fix — `fetch-depth: 0` on a ~1 GB-history repo is a meaningful cost** (`.github/workflows/assign-pr-number.yml:18`).
This workflow runs on every PR push and only needs the merge base. A more targeted `git fetch --no-tags --prune --depth=50 origin $GITHUB_BASE_REF` (or `actions/checkout` with `fetch-depth: 2` plus an explicit deepen step) would usually suffice, and `--deepen` can be used if the merge base is further back. Not a blocker, but worth a line of justification in the PR.

**Nit — `file[0 : file.find(".")]`** (`scripts/assign-pr-number.py:71`).
Pre-existing, but since this PR reuses it: prefer `file.split(".", 1)[0]` or, since you already have the regex match, add a named group `(?P<number>\d+)` to `rx_numbered_fragment` and use `match.group("number")`. More readable and doesn't silently return `""` on a pathological input.

**Nit — `--diff-filter=AM` misses `R`** (`scripts/assign-pr-number.py:33`).
If a reviewer locally runs `git mv 9999.fix.md 8888.fix.md` (still wrong, but renamed), git reports it as `R`, not `A`/`M`, so it will not be in `pr_touched_fragments` and the mismatch healing will skip it. Adding `R` to the filter (`--diff-filter=AMR`) covers this. Edge case, but cheap.

**Nit — type hint shadowing** (`scripts/assign-pr-number.py:54-56`).
`pr_touched_fragments: set[str] | None = (get_pr_touched_fragments(...) if base_ref else None)` is fine, but you can simplify with `pr_touched_fragments = get_pr_touched_fragments(...) if base_ref else None` — the annotation isn't adding information here.

**Nit — no unit tests** (`scripts/`).
No tests exist under `tests/` for `scripts/assign-pr-number.py`. For CI glue this is defensible, but the rename-collision scenarios above are exactly the kind of logic that pays off a small pytest file with a tmp `changes/` directory. Consider adding coverage for: collision between mismatched + existing, collision between two mismatched same-type, unnumbered-wins-over-mismatched ordering, and local-fallback (no `GITHUB_BASE_REF`) leaving numbered fragments alone.

### Open questions for the author

- Backport branches (flagged in the PR body): current behavior will rewrite `11170.fix.md` to the backport PR's number when the backport lands on a release branch. What's the team's convention — do release-branch fragments keep the original `main` PR number, or take the backport PR number? If the former, gate the mismatched-loop behind `if base_ref == "main":` (or a configurable list of trunk branches). Worth resolving before merge since the `25.*` release branches will hit this on the first backport after this PR merges.
- Why `fetch-depth: 0` rather than a deepen-as-needed step? If there's a pragmatic reason (simplicity, avoids merge-base-not-found flakes), a one-line comment in the workflow would help future maintainers.
- Is exiting non-zero on a mismatch (the "fail loudly" alternative from #11214) still off the table? The auto-heal is friendly, but a hard failure has the property that humans notice the mistake rather than the script papering over it — and it eliminates the overwrite-collision risk entirely.

rapsealk:
LGTM ✅

Post-polish (`beebc4cf2`) the PR addresses all four open review concerns and the backport question raised in the PR body:

- **Backport gate** — `rewrite_mismatched = base_ref == "main"` (`scripts/assign-pr-number.py:66`) preserves the original-PR numbering on release-branch backports so release notes keep linking back to the original change.
- **Atomic collision detection** — planned renames are collected up-front and checked against both on-disk files and intra-plan duplicates before any `Path.rename()` fires; no partial-rename state on failure.
- **Unified classification loop** — numbered / unnumbered / invalid resolution happens in one pass, eliminating the prior dual-loop shape.
- **README accuracy** — corrected the towncrier config reference to `[[tool.towncrier.type]]` + `directory`, and scoped the auto-rewrite description to *PRs targeting `main`* and *fragments added by the PR*.

Verified locally against 10 scenarios (main-PR mixed renames, backport preservation, local fallback, same-type collisions with disk and within plan, mismatch vs. correct collision, no-op paths, invalid filenames). CI is green, including the `assign-pr-number` workflow dogfooding itself on this very push.

fregataa (approved):

Issue #11216, opened 2026-04-22
Improve agent container resource-mounting performance (aiodocker + runtime layer)


rapsealk:
## Main idea

The agent's container resource-mounting pipeline (CPU, memory, GPU) works correctly but has accumulated several latent performance issues and half-landed optimizations. This epic tracks the work to address them and to prepare the runtime layer for CDI / containerd / Kata adoption.

Investigation summary (evidence below): the allocation pipeline — `KernelResourceSpec` → `allocate()` → per-plugin `generate_docker_args()` → `docker.containers.create()` — is well-architected, but stats collection polls the Docker API per-container, Docker connections are opened per-operation, NUMA memory pinning is implemented-but-commented-out, scratch-dir size collection is hard-disabled, and GPU mounting is tied to Docker-specific `DeviceRequests` (blocks non-Docker runtimes).

### P0 — localized, low-risk, measurable wins

- [ ] **Enable `CpusetMems` when allocation is NUMA-local**
  `src/ai/backend/agent/docker/intrinsic.py:429–442` — `CpusetMems` is commented out despite `AffinityMap` already computing NUMA nodes. Gate it on `len(unique_numa_nodes) == 1`. Expected 2–3× improvement on remote-memory access for NUMA hosts.
- [ ] **Pool the Docker daemon connection**
  `src/ai/backend/agent/docker/agent.py:826–835, 1240` — `async with closing_async(Docker())` is called 5–10× per kernel start (once per computer plugin + creation). Keep a long-lived client on the agent.
- [ ] **Stream container stats instead of per-cycle polling**
  `src/ai/backend/agent/docker/intrinsic.py:134–169` — `stats(stream=False)` with a 2 s per-container timeout (`_CONTAINER_STAT_TIMEOUT`) means a single hung container delays the stat cycle. Move to a streamed reader per container.
- [ ] **Default to sysfs-first stats on Linux; Docker API only for network/IO and fallback**
  `src/ai/backend/agent/docker/intrinsic.py:262–347, 625–900` — dual path already exists (`StatModes.CGROUP` vs `DOCKER`). Drops CPU/memory stat cost from ~100–500 ms to <10 ms.

### P1 — blocked or requires design

- [ ] **Re-enable scratch directory size collection** via inotify-driven incremental tracking
  `src/ai/backend/agent/docker/intrinsic.py:631–648` currently hard-returns `0` — comment notes the blocking `rglob()` caused "indefinite accumulation of stat collector tasks."
[...]

PR #11222, opened 2026-04-22
refactor(BA-5857): Pin container memory to NUMA-local node via CpusetMems

### LINKED ISSUE #11217: Enable `CpusetMems` when CPU allocation is NUMA-local
Parent epic: #11216

## Main idea

`src/ai/backend/agent/docker/intrinsic.py:429–442` — the `CpusetMems` host-config key is commented out, so container memory is not pinned to the same NUMA node as the allocated CPU cores. NUMA tracking itself is fully implemented: `AffinityMap` (`src/ai/backend/agent/affinity_map.py:145–160`) already computes the distance matrix, and `CPUPlugin.list_devices()` (`intrinsic.py:213`) records a `numa_node` per core.

Gate `CpusetMems` on the allocated cores belonging to a single NUMA node, and set it to that node ID. Fall back to the current behavior (omit the key) when the allocation spans multiple nodes.

```python
# intrinsic.py, around line 434
numa_nodes = {core.numa_node for core in allocated_cores}
host_config = {"Cpus": len(cores), "CpusetCpus": ",".join(sorted_core_ids)}
if len(numa_nodes) == 1:
    host_config["CpusetMems"] = str(next(iter(numa_nodes)))
```

Expected 2–3× improvement on remote-memory access latency for NUMA hosts when the allocation already happens to be node-local (which `AllocationStrategy.FILL` prefers today — see comment at `resources.py:176`).

## Alternative ideas

- Allow multi-node pinning (`CpusetMems="0,1"`) when the allocation spans exactly the selected set. Lower win, wider applicability.
- Always emit `CpusetMems` equal to the union of all nodes covered by allocated cores — functionally equivalent to the current default on most systems, so low value.

## Anything else?

Needs verification on a NUMA host. Single-socket machines are unaffected (single node = current behavior).

rapsealk:
Closes #11217 (BA-5857)
Refs #11216

## Summary
- Set `HostConfig.CpusetMems` to the NUMA node ID when all allocated CPU cores are on the same node.
- Omit `CpusetMems` for multi-node allocations so the kernel's default NUMA memory placement policy applies.
- Short-circuit on `libnuma.num_nodes() <= 1` so non-NUMA Linux hosts and macOS / WSL / Docker Desktop don't get `CpusetMems="0"` pinned on every container.
- Use per-allocated-core `libnuma.node_of_cpu()` instead of a host-wide `list_devices()` scan.

## Why
Pre-PR the NUMA memory pinning code was commented out, so CPU-only workloads on NUMA hosts could pay 2-3× remote-memory access latency. Part of the agent-runtime performance epic #11216.

## Test plan
- [ ] `pants check src/ai/backend/agent/docker::` and `pants test tests/unit/agent::` pass
- [ ] On a NUMA host, verify container `HostConfig.CpusetMems` reflects the allocation node
- [ ] On a single-socket / non-NUMA host, verify `CpusetMems` is absent


rapsealk:
Review summary from an independent pass + resolutions pushed to this branch.

| Finding | Resolution |
|---|---|
| `libnuma.node_of_cpu` returns `0` on non-NUMA/non-Linux hosts → every container would get `CpusetMems="0"` pinned | Short-circuit on `libnuma.num_nodes() > 1` before the per-core loop (commit `0eda461ec`) |
| Perf: `list_devices()` iterates every host core just to build a lookup | Switched to per-core `libnuma.node_of_cpu(core)` (commit `64529b24a`) |
| Comment claimed Docker `HostConfig` can't express multi-node `CpusetMems` — factually wrong | Reworded as deliberate policy: defer to kernel default NUMA placement for multi-node allocations (commit `64529b24a`) |
| Twin tombstone in `dummy/intrinsic.py:148` | Removed (commit `0eda461ec`) |
| No unit tests for the new branching | 4 tests added: single-node, multi-node, unknown/negative-node, non-NUMA short-circuit |

Scope stayed tight: only `docker/intrinsic.py`, `dummy/intrinsic.py`, and `tests/unit/agent/test_docker_intrinsic.py` touched. `libnuma.node_of_cpu` is a ctypes call into an in-memory lookup, so calling it synchronously inside `generate_docker_args` is fine per the agent's async rules. Scheduler default is `AffinityPolicy.PREFER_SINGLE_NODE`, so the common case hits the pin-to-node branch; multi-node correctly falls through to kernel default rather than forcing `"0,1"` across all nodes.

src/ai/backend/agent/docker/intrinsic.py
```diff
@@ -433,12 +433,29 @@ async def generate_docker_args(
...
+                node = libnuma.node_of_cpu(core)
+                if node < 0:
+                    allocated_nodes.clear()
+                    break
+                allocated_nodes.add(node)
+            # Pin memory only when the CPU allocation is fully node-local.
+            # For multi-node CPU allocations, intentionally leave CpusetMems unset
+            # so Docker/kernel default NUMA memory placement policy can apply.
+            if len(allocated_nodes) == 1:
+                host_config["CpusetMems"] = str(next(iter(allocated_nodes)))
```
jopemachine, line 456: How about extracting this logic into a separate private method?

tests/unit/agent/test_docker_intrinsic.py
```diff
@@ -610,3 +612,137 @@ def test_raises_oserror_for_nonexistent_pid(self) -> None:
...
+
+    @staticmethod
+    def _device_alloc(core_ids: list[int]) -> dict[SlotName, dict[DeviceId, Decimal]]:
+        return {
+            SlotName("cpu"): {DeviceId(str(cid)): Decimal("1") for cid in core_ids},
+        }
+
+    @staticmethod
+    @contextmanager
+    def _patch_node_of_cpu(
```
jopemachine, line 632: Since this method is intended to reproduce specific test scenarios, it would be more readable to inject fixtures for each scenario and apply parametrization.

PR #11223, opened 2026-04-22
refactor(BA-5860): Default to sysfs-first CPU/memory stats on native Linux

### LINKED ISSUE #11220: Default to sysfs-first CPU/memory stats on Linux
Parent epic: #11216

## Main idea

The agent already has two stat-collection implementations: a sysfs direct-read path (fast, ~1–10 ms, reads `/sys/fs/cgroup/...` directly) and a Docker-API path (slow, 100–500 ms per container). Selection is driven by `StatModes.CGROUP` vs `StatModes.DOCKER` in `src/ai/backend/agent/docker/intrinsic.py:262–347` (CPU) and `:625–900` (memory).

Today the selection is a static config decision. On Linux native hosts (the primary deployment target), sysfs is always available and strictly faster, but the default selection logic does not prefer it. Linuxkit / Docker Desktop hosts (line 370) legitimately need the API path.

Change the default so Linux native hosts pick `CGROUP` mode automatically, fall back to `DOCKER` only when sysfs reads fail or the host is linuxkit/Docker Desktop. Keep the explicit config override.

Per-resource behavior to preserve:
- CPU: cgroup `cpu.stat` / `cpuacct.usage` read — already implemented for both cgroupv1 (line 275 match) and cgroupv2 (line 287).
- Memory: `memory.current`, `memory.stat` — already implemented for both versions (lines 661, 694).
- Network / block IO: keep using Docker API; these are not cleanly available from container cgroups and require netns entry which is costlier than the API call.

## Alternative ideas

- Remove the `DOCKER` mode entirely for CPU and memory. Cleaner, but loses the escape hatch for hosts where sysfs paths move (rootless, nested containers, unusual cgroup drivers).
- Add a runtime healthcheck on startup that benchmarks both paths and picks the winner. Overkill for what is essentially a Linux-native vs not decision.

## Anything else?

Pairs well with issue about streaming stats (#11219) — once sysfs is the default for CPU/memory, the streamed path only needs to cover network/IO and the linuxkit fallback, which keeps the stream connection count and complexity down.

rapsealk:
Closes #11220 (BA-5860)
Refs #11216

## Summary
- Default CPU/memory stat collection to `StatModes.CGROUP` on native Linux (sysfs path, ~100-500 ms → <10 ms per sample).
- Keep `StatModes.DOCKER` as the forced path on linuxkit / Docker Desktop and as a per-read fallback when a cgroup sysfs read fails.
- Dedup "cgroup sysfs failed, falling back to Docker API" warnings per `(plugin, container_id)` via a bounded `cachetools.LRUCache` so a systemic sysfs failure doesn't storm logs.
- Replace `raise RuntimeError("should not reach here")` in exhaustive `match` defaults with `typing.assert_never` (satisfies the agent rule banning built-in exceptions in business logic + strengthens mypy exhaustiveness).
- No changes to network/IO stat collection.

## Operator note — possible metric step-down on upgrade
Hosts previously running `stats-type: docker` may see:
- **Memory**: reported usage steps down by the inactive-file cache amount (sysfs now matches `docker stats` CLI semantics).
- **Block I/O** (cgroup v1 only): readings may shift because sysfs reads `blkio.throttle.io_service_bytes` while the Docker API path reads `blkio_stats.io_service_bytes_recursive`.

Dashboards and autoscaling thresholds tuned to the old values should be re-evaluated.

## Why
Agent-runtime performance epic #11216; sysfs is an order of magnitude faster for CPU/memory stats than the Docker API round-trip.

## Test plan
- [ ] `pants test tests/unit/agent::` passes (CPU + Memory plugin fallback, linuxkit guard, warn dedup)
- [ ] On a native Linux host, confirm CPU/memory stats populate via cgroup path (no Docker API calls for those fields)
- [ ] Simulate a cgroup read failure; confirm the Docker-API fallback triggers and the one-time warning fires


rapsealk:
Review summary from an independent pass + resolutions pushed to this branch (commit `9cc2b7e71`).

| Finding | Resolution |
|---|---|
| Memory metric will step-down on upgrade (sysfs excludes inactive-file cache; matches `docker stats` CLI) — not called out for operators | `changes/11223.enhance.md` extended to warn that dashboards and autoscaling thresholds tuned to the previous `stats-type: docker` values may need re-tuning |
| `_cgroup_fallback_warned: set[str]` grew unbounded on container churn | Replaced with `cachetools.LRUCache(maxsize=1024)` — same pattern used elsewhere in the agent |
| `raise RuntimeError("should not reach here")` at two `match ctx.mode:` sites violates the `BackendAIError`-only rule | Both sites now use `assert_never(ctx.mode)`, which also strengthens mypy exhaustiveness |
| No mirror tests for `MemoryPlugin` fallback / linuxkit guard; no dedup-warn tests | Added 4 tests: mirror CPU fallback + linuxkit tests for MemoryPlugin, plus `TestWarnCgroupFallbackOnce` covering dedup and LRU eviction |

Cgroup v1/v2 both handled (pre-existing). Network/IO still goes through Docker API per spec. Selection via `StatModes.get_preferred_mode()` correctly picks up `is_containerized()` — agents running inside a container stay on DOCKER mode. Cgroup-first wraps sysfs with a per-read Docker-API fallback, so transient sysfs errors downgrade gracefully rather than silently zeroing metrics. Linuxkit guard belt-and-suspenders against Docker Desktop / colima.

PR #11224, opened 2026-04-22
refactor(BA-5859): Stream container stats from Docker instead of polling

### LINKED ISSUE #11219: Stream container stats instead of per-cycle polling
Parent epic: #11216

## Main idea

`src/ai/backend/agent/docker/intrinsic.py:134–169` (`fetch_api_stats`) calls `container.stats(stream=False)` on every collection cycle, with a per-container 2 s timeout (`_CONTAINER_STAT_TIMEOUT`, line 73). `CPUPlugin` (line 301) and `MemoryPlugin` (line 791) both go through this path when `StatModes.DOCKER` is selected, and linuxkit hosts are forced onto it (line 370).

Two problems:
1. Every cycle pays a full HTTP round-trip to dockerd per container — typically 100–500 ms.
2. A single unresponsive container consumes the full 2 s timeout before the gather completes, delaying every other metric on the same cycle.

Switch to a long-lived streamed reader per container using `container.stats(stream=True)`, keep the latest sample in memory, and have the stat collector read from the in-memory cache instead of making new API calls. Start/stop the stream alongside the container's lifecycle.

## Alternative ideas

- Keep per-cycle polling but parallelize with a hard global deadline instead of per-container timeout. Smaller win, doesn't fix the round-trip cost.
- Do nothing here and rely on the sysfs-first path (tracked separately) on Linux; stream only on hosts where sysfs isn't available (linuxkit, Docker Desktop). This is probably the pragmatic order: land sysfs-first first, then stream only for the remaining fallback population.

## Anything else?

Needs care around stream lifecycle: leaked readers = leaked connections against dockerd. Should be wired to the same container-lifecycle hooks used for teardown. Also consider backpressure — dockerd emits stats every ~1 s, which is likely faster than needed; may want to downsample.

rapsealk:
Closes #11219 (BA-5859)
Refs #11216

## Summary
- `DockerStatsStreamer`: a long-lived reader keyed by container ID. Callers read the most recent sample from memory instead of `container.stats(stream=False)` per collection cycle.
- Eager `start()` / `stop()` wired through container-start / container-clean events on the agent (no lazy-on-first-get race).
- Bounded exponential-backoff reconnect (1s → 30s, 8 retries) on transient `ClientConnectionError` / `TimeoutError`.
- `CancelledError` re-raised; non-transient `DockerError` (e.g. 404 for container-gone) exits cleanly without retry spin.
- Renamed `_CONTAINER_STAT_TIMEOUT` → `_CONTAINER_INSPECT_TIMEOUT` to reflect its remaining scope.

## Why
Cuts API-path stat latency from 100-500 ms per container to a memory read, and removes the 2s per-container timeout as a serial-blocker. Part of epic #11216.

## Known trade-offs (tracked as follow-ups)
- CPU and Memory intrinsic plugins currently each own a `DockerStatsStreamer` → 2 streams per container. Consolidation tracked as #11232 / PR #11234.
- After #11223 (sysfs-first) merges, CPU/Memory `api_impl` paths will be sysfs-only and the stream migrates to network/IO consumers. `TODO(#11223)` comments mark the affected paths.

## Test plan
- [ ] `pants test tests/unit/agent::` passes (lifecycle, reconnect, shutdown, 404 cleanup)
- [ ] Run an agent locally; confirm CPU/memory stats populate for running containers
- [ ] Stop a container mid-stream; confirm the reader task exits cleanly and `get_latest()` returns `None`
- [ ] `docker restart` mid-session; confirm reconnect kicks in and stats resume


rapsealk:
Review summary from an independent pass + blocker resolutions pushed to this branch (commit `43e9ff7c0`).

**Verdict: ship with changes — blockers resolved.**

| Finding | Severity | Resolution |
|---|---|---|
| `DockerStatsStreamer.stop()` defined but never called — container destruction leaked the reader task until dockerd closed the stream | **Blocker** | Eager `start()` now fires from `_handle_start_event`; `stop()` from `_handle_clean_event`. No-op default hooks `notify_container_started/destroyed` added to `AbstractComputePlugin` so k8s/dummy intrinsics inherit them. `TODO(#11219)` removed |
| Lazy-on-first-call start meant the first stat cycle returned `None` for CPU/memory on newly-created containers (regression vs. `stats(stream=False)`) | **Blocker** | Eager start gives dockerd ~1s head-start before the first collection cycle. Opt-in `wait_for_first_sample(cid, timeout=0.5)` available when callers want a blocking guarantee |
| `except (asyncio.CancelledError, Exception): pass` swallowed cancellation and `BackendAIError` alike | Major | Reader re-raises `CancelledError`; logs others with cid; new `ContainerStatsStreamError(BackendAIError)` on permanent failure |
| No reconnect on `ClientConnectionError` | Major | Bounded exponential backoff (1s → 30s, 8 retries), then raises `ContainerStatsStreamError` |
| `close()` didn't cancel in-flight readers | Minor | Now awaits cancellation of all active tasks |
| `_CONTAINER_STAT_TIMEOUT` misnamed after refactor (only used for `container.show()` now) | Nit | Renamed to `_CONTAINER_INSPECT_TIMEOUT` |

Tests added (9 new): `test_start_spawns_reader_and_first_sample_lands`, `test_stop_cancels_reader_and_drops_cache`, `test_close_cancels_all_in_flight_tasks`, `test_first_sample_wait_times_out_when_no_frames`, `test_reconnect_after_client_connection_error`, `test_cancelled_error_reraises_from_reader`, `test_notify_started_calls_plugin_hook`, `test_notify_destroyed_calls_plugin_hook`, `test_notify_tolerates_plugin_exceptions`.

**Coordination note**: if #11223 (sysfs-first) merges first, the Docker stream is only needed for network/IO fields — streaming all fields on every container is wasteful. Worth a follow-up to narrow the streamed fields, but out of scope here.

**Deferred**: `CPUPlugin` and `MemoryPlugin` each own a separate `DockerStatsStreamer`, so two streams open per container. Consolidating is a follow-up refactor.

jopemachine:
For cumulative counters, if the Docker client in DockerStatsStreamer fails for an extended period, the same value may be repeatedly collected in latest. This can lead to a spike immediately after recovery.

jopemachine (dismissed):

PR #11226, opened 2026-04-22
refactor(BA-5858): Reuse a long-lived aiodocker client across container operations

### LINKED ISSUE #11218: Pool the Docker daemon connection in the agent
Parent epic: #11216

## Main idea

The agent currently opens and closes a fresh aiodocker connection for each container operation. `src/ai/backend/agent/docker/agent.py:1240` and the accelerator-config generation loop (`agent.py:826–835`) use `async with closing_async(Docker())`, which means a single kernel start opens on the order of 5–10 Docker connections (one for `generate_docker_args` per computer plugin, one for `generate_accelerator_configs` per plugin, one for `containers.create()`, one for `start()`, plus recovery/inspect paths). Each connection pays TCP/Unix-socket setup and HTTP parser initialization.

Replace the per-call pattern with a long-lived `Docker` client owned by the agent, passed into accelerator plugins instead of created inside them. Preserve the current `closing_async()` semantics only for the agent shutdown path.

Concrete places to change:

- `src/ai/backend/agent/docker/agent.py:826` (`apply_accelerator_allocation`)
- `src/ai/backend/agent/docker/agent.py` (`generate_accelerator_configs`)
- `src/ai/backend/agent/docker/agent.py:1240` (`start_container`)
- Computer plugin `generate_docker_args(docker, ...)` signature already accepts a `docker` arg — callers just need to stop creating throwaway clients.

## Alternative ideas

- Connection pool (max N idle) rather than a singleton — useful if concurrent kernel starts end up serialized on a single connection. Profile first before adding the complexity.
- Keep per-op connections but reduce the number of ops (batch `generate_docker_args` into one pass). Smaller win, does not help recovery paths.

## Anything else?

Needs a sanity check that aiodocker's `Docker` is safe to share across concurrent tasks (it is, as of current versions — connector is an `aiohttp.ClientSession`). Should pair with a small benchmark: kernel start latency before/after, measured at the agent.

rapsealk:
Closes #11218 (BA-5858)
Refs #11216

## Summary
- Own a single `aiodocker.Docker` on the agent and thread it into all per-op sites (`start_container`, `apply_accelerator_allocation`, `generate_accelerator_configs`, recovery/inspect, image / container / network ops).
- Kernel start now opens ~1 connection instead of ~5-10 (per-op `async with closing_async(Docker())` removed).
- `shutdown()` wraps its body in `try/finally` for unconditional `docker.close()` — a raising `super().shutdown()` no longer leaks the aiohttp session.
- `__ainit__` boot-failure path releases the client if startup raises (narrowed to `except Exception:` to let `KeyboardInterrupt` / `SystemExit` propagate).
- `monitor_docker_events()` intentionally keeps its own `closing_async(Docker())` — needs per-op isolation for reconnect semantics.

## Why
Part of epic #11216. A shared client amortizes TCP/Unix-socket + HTTP parser + connector init across the agent's lifetime and lets aiohttp reuse keepalive sockets across ops.

## Known trade-offs (tracked as follow-ups)
- After a dockerd restart, the shared session's pooled keepalive socket can be stale and surface a one-shot `aiohttp.ClientConnectionError` on the first post-restart call. Follow-up: #11233 / PR #11235.
- Accelerator plugins (`src/ai/backend/accelerator/*`) still instantiate their own `Docker()` inside `init()`. Deferred under epic #11216.

## Test plan
- [ ] `pants test tests/component/agent/docker::` passes (shutdown-lifecycle tests + existing image-pull / purge coverage)
- [ ] Start a kernel, confirm it launches end-to-end
- [ ] Run `agent.shutdown()`; confirm no `Unclosed client session` warning
- [ ] Simulate a raising `super().shutdown()`; confirm the client is still closed


rapsealk:
Review summary + tradeoff notes. Resolutions pushed to this branch.

| Finding | Resolution |
|---|---|
| `shutdown()` closed `self.docker` outside the `try/finally` wrapping `super().shutdown()` — a raising inner shutdown leaked the `aiohttp.ClientSession` | Wrapped the full `shutdown()` body in `try: ... finally: await self.docker.close()` (commit `801ed1e90`) |
| `__ainit__` raising after `self.docker = Docker()` also leaked | Narrow `try/except BaseException: await self.docker.close(); raise` around post-construction startup (includes `KeyboardInterrupt`/`SystemExit`) |
| `test_auto_pull_*` broke because class-level `patch("...Docker")` no longer intercepts the shared instance | Realigned tests to patch `agent.docker.images.inspect` directly (commit `27f9c3616`) |
| No regression coverage for the shutdown lifecycle | Added tests asserting session closed after normal `shutdown()` AND after a raising `super().shutdown()` |

**Tradeoffs vs. the original `async with closing_async(Docker())` pattern**:

*What the original gave up that this PR lost*: automatic cleanup (no explicit lifecycle to get wrong — the shutdown-leak bug only exists because we now own it), fault isolation between ops, transparent daemon-restart recovery (shared `ClientSession` survives but a stale keep-alive socket can surface a one-shot `ClientConnectionError`), trivial class-level mocking (hence the test realignment), no pickling hazard.

*What the new pattern wins*: ~5–10× fewer connections per kernel start (TCP/Unix-socket handshake + HTTP parser + connector init amortized over agent lifetime), keep-alive socket reuse, lower resource footprint.

Net: favorable for a long-lived agent. `monitor_docker_events()` correctly keeps its own `closing_async(Docker())` since stream reconnect semantics need per-op isolation. Accelerator plugins (`src/ai/backend/accelerator/*/plugin.py`) still instantiate `Docker()` inside `init()` — deferred to a follow-up within epic #11216.

jopemachine (approved):

PR #11228, opened 2026-04-22
fix(agent): Scope aiodocker client in DockerKernel.commit() via closing_async

### LINKED ISSUE #11227: Reuse the shared aiodocker client in DockerKernel methods
Parent epic: #11216
Follow-up to: #11218 / #11226

## Main idea

PR #11226 routed every per-op `Docker()` instantiation through a shared long-lived client **except** in `DockerKernel`, because plumbing the shared client through would change `DockerKernel.__init__`'s contract. The remaining throwaway connections are:

- `src/ai/backend/agent/docker/kernel.py:121` — `get_logs()`
- `src/ai/backend/agent/docker/kernel.py:224` — `commit()` (bare `Docker()`, not even `closing_async` — also worth fixing for correctness)
- `src/ai/backend/agent/docker/kernel.py:327` — `download_file()`
- `src/ai/backend/agent/docker/kernel.py:356` — `download_single()`
- `src/ai/backend/agent/docker/kernel.py:486` — module-level `prepare_krunner_env_impl()` (separate case — not a `DockerKernel` method)

## Design

- Add `docker: Docker` as a keyword-only argument to `DockerKernel.__init__`, store on `self`.
- Update construction sites in `DockerAgent` (`agent.py`) to pass `self.docker`.
- Update the four `DockerKernel` methods above to use `self.docker` instead of `async with closing_async(Docker())` / bare `Docker()`.
- `prepare_krunner_env_impl` is a module-level function — evaluate the caller: if the caller has a shared `Docker` client handy, accept it as a parameter; otherwise leave it as-is and note why.
- Handle pickling: `DockerKernel` has `__getstate__`/`__setstate__` for RPC marshalling — the `docker` field must NOT be pickled. Exclude it explicitly in `__getstate__` and re-inject on restore (the agent can re-attach `self.docker` when materializing a kernel from persisted state).

## Alternative ideas

- Instead of a constructor arg, expose a setter (`kernel.attach_docker(self.docker)`) called from `DockerAgent`. Slightly less explicit, but avoids touching `__init__`'s signature. Lower preference — constructor injection is clearer.
- Inject via a context var. Rejected — implicit and harder to reason about.

## Anything else?

- `monitor_docker_events()` (in `agent.py`) must keep its own `closing_async(Docker())` — it reconnects on stream drops. Do NOT touch.
- `PersistentServiceContainer` in `docker/utils.py` is bootstrap-time only; not part of this issue.

rapsealk:
## Summary
- Wrap the bare `Docker()` in `DockerKernel.commit()` with `closing_async(...)` so the `aiohttp.ClientSession` is always closed.
- Matches the pattern already used by sibling methods (`get_logs`, `download_file`, `download_single`).

## Why not the broader shared-client refactor?
The original scope of #11227 threaded a shared `Docker` client into `DockerKernel` and its recovery path. After review, the ongoing invariant-maintenance cost (pickle exclusion, `attach_docker()` ordering on recovery, constructor plumbing) outweighed the latency gains on these rarely-called methods. The only real bug — a leaked session in `commit()` — is fixed here in isolation. See the prior discussion on this PR for the full tradeoff analysis.

## Test plan
- [ ] `pants check src/ai/backend/agent/docker::` passes
- [ ] Commit a kernel and verify no aiohttp-session leak warnings on agent shutdown.

Closes #11227
Refs #11216

### DIFF
--- a/changes/11228.fix.md
+++ b/changes/11228.fix.md
@@ -1,0 +1,1 @@
+Fix a leaked aiodocker session in `DockerKernel.commit()` by scoping the client via `closing_async`.
--- a/src/ai/backend/agent/docker/kernel.py
+++ b/src/ai/backend/agent/docker/kernel.py
@@ -221,8 +221,7 @@ class DockerKernel(AbstractKernel):
         try:
             async with FileLock(path=lock_path, timeout=0.1, remove_when_unlock=True):
                 log.info("Container (k: {}) is being committed", kernel_id)
-                docker = Docker()
-                try:
+                async with closing_async(Docker()) as docker:
                     # There is a known issue at certain versions of Docker Engine
                     # which prevents container from being committed when request config body is empty
                     # https://github.com/moby/moby/issues/45543
@@ -289,8 +288,6 @@ class DockerKernel(AbstractKernel):
                                         await write_task
                         finally:
                             await docker.images.delete(image_id)
-                finally:
-                    await docker.close()
         except TimeoutError:
             log.warning("Session is already being committed.")
 


rapsealk:
Review summary + tradeoff notes — **worth considering whether to close this PR in favor of a narrower fix**.

**Resolutions pushed to this branch (commit `6a312195e`)**:
- Pickle round-trip test locking `__getstate__` exclusion of `_docker` (invariant most at risk from future contributors).
- Class-level docstring documenting the `_docker`-is-local-only / must-re-attach invariant.
- `attach_docker()` logs `log.warning` on client-swap (mismatched connector); same-connector reattach silent.
- `DockerAgent.docker: Docker` class annotation aligned with #11226.

---

**Why this PR is worth a second look on scope:**

The original `async with closing_async(Docker()) as docker:` in `DockerKernel.get_logs/download_file/download_single` gave up more than it appears at first:

| Property preserved by original | Cost paid by this PR |
|---|---|
| No pickling hazard | `DockerKernel` is RPC-marshalled → `__getstate__`/`__setstate__` exclusion is now load-bearing; one wrong line in a future PR silently breaks RPC |
| No recovery-ordering window | Unpickled kernels have no `_docker` until `_load_kernel_registry_from_recovery` runs `attach_docker()` on each — any method fired in that window raises `AttributeError` |
| Simple construction | Kw-only `docker=` threaded through `DockerKernel.__init__`, `DockerKernelCreationContext`, `KernelObjectSpec`, and the recovery loader (plus a `cast("DockerAgent", ...)`) |
| Trivial test seam | Tests mocking `Docker` now need to reach through `kernel._docker` |
| Fault isolation | A bad session state now fails every subsequent op on that kernel |

**Gains**: `get_logs`, `download_file`, `download_single` are low-frequency, not per-kernel-start hot path. The only unambiguous win is the **`commit()` leak fix** — the old `commit()` used a bare `Docker()` with no `closing_async`, a real bug on every commit.

---

**Suggested alternative**: close this PR and open a minimal one that:
1. Wraps the existing `commit()` bare `Docker()` in `closing_async(...)` — fixes the real leak in ~3 lines.
2. Leaves `get_logs`, `download_file`, `download_single` on their current `closing_async(Docker())` pattern.

That fixes the actual bug, keeps the fault-isolation / no-pickling-hazard / simple-testing properties of the original design for rarely-called methods, and avoids paying ongoing invariant-maintenance cost for marginal latency wins.

Arguments for **keeping** this PR:
- Consistency with #11226 (agent owns shared client; kernels inheriting it is the natural next step).
- The invariant documentation (docstring + pickle test + `attach_docker` warn) is now in place — cost is paid.
- Future per-kernel Docker operations would automatically benefit.

My lean: narrow fix is cleaner, but the call belongs to @lablup maintainers. Happy to spin up the `commit()`-only alternative if preferred.

rapsealk:
Closing as not planned.

Rereading the original `DockerKernel.commit()` on main: it was already closing the session correctly via `try/finally: await docker.close()` — not leaking. The narrow `async with closing_async(...)` rewrite in the current HEAD of this PR is cosmetic (style/idiom consistency with sibling methods), not a fix. The `fix(agent): ...` title over-promises.

The broader shared-client extension into `DockerKernel` (the original scope of this PR — pickling exclusion, `attach_docker()`, recovery ordering, constructor plumbing through ~4 sites) was judged unfavorable on cost/benefit for `get_logs` / `download_file` / `download_single` / `commit`, which are not per-kernel-start hot path. See the prior discussion in this thread.

If the `try/finally` → `async with closing_async(...)` idiom cleanup is worth doing, it can be swept in organically during a future touch to `docker/kernel.py`, not as its own PR.

Closing #11227 (not planned) alongside this PR.

PR #11234, opened 2026-04-22
refactor(BA-5861): Share a single DockerStatsStreamer across CPU/Memory plugins

### LINKED ISSUE #11232: Share a single DockerStatsStreamer across CPU/Memory intrinsic plugins
Parent epic: #11216
Follow-up to: #11224

## Main idea

After #11224, `CPUPlugin` and `MemoryPlugin` each own their own `DockerStatsStreamer`, and each constructs its own `aiodocker.Docker()` client in `init()`. That means every running container opens **two** persistent Docker stats streams across **two** separate `aiohttp.ClientSession` instances.

Aiohttp's default `TCPConnector(limit_per_host=30)` means at roughly 50+ containers per agent you start queueing stream setups and stacking backoff retries. Memory and connector overhead are also double what they need to be.

This becomes even more obviously worth refactoring once #11223 (sysfs-first CPU/memory stats) lands, because CPU/Memory `api_impl` paths will fall through to Docker only as a fallback — the persistent stream is really only needed for network / blkio fields neither plugin consumes today.

## Design

- Own a single `DockerStatsStreamer` on `DockerAgent` (or `AbstractAgent`, if the abstraction generalizes to kubernetes).
- Both `CPUPlugin` and `MemoryPlugin` consume via a shared reference (set during plugin init, or accessed through the existing `computer_ctx`).
- Lifecycle (`start` / `stop`) still hangs off `_handle_start_event` / `_handle_clean_event`; the agent just dispatches to a single streamer instead of each plugin.
- Post-#11223: consider moving the streamer out of the intrinsic plugins entirely and into a network/IO consumer, since that's what actually needs the stream.

## Dependencies

- Best sequenced AFTER #11223 (sysfs-first), since that clarifies what fields the stream still needs to deliver.

## Scope

- Refactor ownership of `DockerStatsStreamer`.
- Unify the two streamers into one instance per agent.
- Update `notify_container_started/destroyed` dispatch path accordingly.
- No behavioral change for operators — this is purely a resource-consumption fix.

## Out of scope

- Changing the streaming protocol or cadence.
- Kubernetes agent (`src/ai/backend/agent/kubernetes/`) — if it's affected, evaluate in parallel.

rapsealk:
Closes #11232 (BA-5861)
Refs #11216

## Summary
- Consolidate `DockerStatsStreamer` ownership onto `DockerAgent` — one streamer per agent instead of one per intrinsic plugin.
- Each container now opens a single persistent Docker stats stream instead of two.
- Agent dispatches `start` / `stop` directly via new `AbstractAgent._on_container_started` / `_on_container_destroyed` hooks; the old plugin-side `notify_container_started` / `notify_container_destroyed` interface is removed.
- Plugins receive the shared streamer via an `attach_stats_streamer()` setter called **before** `await super().__ainit__()` — the streamer is live before `scan_running_kernels()` can inject START events into the lifecycle queue on warm restart.

## Why
At ~50+ containers, the per-plugin-streamer layout hit aiohttp's default `limit_per_host=30` connector limit and stacked backoff retries. See #11232 for the full rationale.

## Stacked on
#11224 (`perf/11219-stream-container-stats`). Rebase to main once #11224 merges.

## Test plan
- [ ] `pants test tests/unit/agent:: tests/component/agent/docker::` passes — includes a regression test asserting `_stats_streamer` is attached before `scan_running_kernels` runs (warm-restart ordering guard).
- [ ] Warm-restart the agent with running containers; confirm streams reattach without `AttributeError`.
- [ ] Start ~10 kernels, verify only N streams to dockerd (one per container, not two) via `ss -tnp | grep dockerd | wc -l`.
- [ ] `docker restart` mid-session, verify streams re-establish and stats resume.


src/ai/backend/agent/docker/agent.py
```diff
@@ -1517,6 +1521,28 @@ async def __ainit__(self) -> None:
...
+        # ``super().__ainit__()`` runs. ``AbstractAgent.__ainit__`` calls
+        # ``scan_running_kernels()`` and then starts the lifecycle handler
+        # task (``process_lifecycle_events``); on warm restart that pipeline
+        # can fire container-start events which land in
+        # ``_on_container_started`` -> ``self._stats_streamer.start(...)``
+        # before any code AFTER ``super().__ainit__()`` gets to run.
+        self._stats_streamer = DockerStatsStreamer(self.docker)
+        for computer_ctx in self.computers.values():
+            instance = computer_ctx.instance
+            if hasattr(instance, "attach_stats_streamer"):
```
jopemachine: We prefer to avoid dynamic access patterns such as hasattr whenever possible.

rapsealk: Good point — removed the `hasattr` probe in 7cf076af1. Moved `attach_stats_streamer` onto `AbstractComputePlugin` with a no-op default so non-Docker plugins (K8s, Dummy, third-party accelerators) inherit it safely, and `DockerAgent.__ainit__` now calls it unconditionally on every compute plugin.

PR #11235, opened 2026-04-22
refactor(BA-5862): Retry once on stale aiodocker connections after dockerd restart

### LINKED ISSUE #11233: Retry once on stale-connection errors from shared aiodocker client after dockerd restart
Parent epic: #11216
Follow-up to: #11226

## Main idea

#11226 pooled a single long-lived `aiodocker.Docker` (wrapping an `aiohttp.ClientSession`) on `DockerAgent`, replacing the per-op `async with closing_async(Docker()) as docker:` pattern. That's a big latency win, but it introduces one production failure mode that the old pattern didn't have:

After `systemctl restart docker` (or any dockerd bounce), the agent's long-lived `aiohttp.ClientSession` still holds keepalive sockets in its connector pool. The first call to dockerd after the restart picks a stale socket, fails with `aiohttp.ClientConnectionError` or `ServerDisconnectedError`, and surfaces as a spurious failure in `purge_images`, `scan_images`, `check_image`, or any per-op site. Aiohttp reconnects on the *next* call, so the error is one-shot — but that one shot is enough to fail a user-visible operation.

Per-op `Docker()` never hit this because each call opened a fresh socket.

## Design

Add a thin once-retry wrapper around the per-op aiodocker sites that catches `aiohttp.ClientConnectionError` / `aiohttp.ServerDisconnectedError` exactly once, then re-runs the call. On the second failure, propagate normally. No exponential backoff — the retry is specifically for the stale-connection case, which resolves on the next attempt.

Sites to cover (grep `self.docker.` in `src/ai/backend/agent/docker/agent.py`):
- `apply_accelerator_allocation`
- `start_container`
- `get_intrinsic_mounts`
- `destroy_kernel` / `clean_kernel`
- `extract_image_command`
- `enumerate_containers`
- `scan_images`
- `push_images` / `pull_images` / `purge_images`
- `check_image`
- `create_local_network` / `destroy_local_network`
- `resolve_image_distro`

Do NOT wrap:
- `monitor_docker_events()` — it already has its own reconnect loop over `closing_async(Docker())`.
- Streaming reads (`DockerStatsStreamer`, introduced in #11224) — those already have bounded-backoff reconnect.

## Alternative ideas

- Set a short `force_close=True` on the connector so every request uses a fresh socket. Simpler, but negates #11226's latency win. Rejected.
- Health-probe on interval and recreate the client when dockerd version endpoint fails. More infrastructure, catches the same case. Overkill for now.

## Out of scope

- Accelerator plugins' `Docker()` usage — tracked separately within epic #11216.

rapsealk:
Closes #11233 (BA-5862)
Refs #11216

## Summary
- Add `_retry_on_stale_connection` helper that catches `(ServerDisconnectedError, ClientOSError)` once, logs a warning, and retries.
- Explicitly **excludes** `ServerTimeoutError` — even though it inherits from `ClientConnectionError`, it represents response-too-slow, not a stale socket. Legitimate long-running `images.push` / `images.pull` timeouts now propagate normally.
- Apply to retry-safe per-op aiodocker calls on `DockerAgent` (image / container / network inspect / list / delete / create / start / stop — 26 sites across 13 methods).
- Explicitly do NOT wrap: `monitor_docker_events` (own reconnect loop), `DockerStatsStreamer` (bounded backoff), `__ainit__` bootstrap probes (boot-failures should be loud), streaming log readers, `commit` and other non-idempotent verbs.

## Why
After `systemctl restart docker`, the shared aiohttp session's pooled keepalive sockets are stale. aiohttp reconnects on the next call, but the first one-shot failure surfaces as spurious user-facing errors (`purge_images`, `check_image`, etc.). See #11233.

## Stacked on
#11226 (`perf/11218-docker-client-pool`). Rebase to main once #11226 merges.

## Test plan
- [ ] `pants test tests/component/agent/docker::` passes — 5 tests: retry-once-then-succeeds, propagates persistent failure, non-matching errors not retried, `ServerTimeoutError` NOT retried (regression guard), `container.create` retry path via `resolve_image_distro`.
- [ ] `docker restart`; run `bai admin image list` from the CLI; verify it succeeds on the first call post-restart (would fail pre-this-PR).
- [ ] Run a deliberately-slow `images.pull` past its timeout; verify it propagates without silent retry.


HyeockJinKim (changes requested):
I'm not sure if that's a very meaningful action. If the connection was lost at a single point in time, would retrying immediately yield meaningful results?

PR #11236, opened 2026-04-22
feat(BA-5797): add effective permissions resolver for entities

fregataa:
## Summary
- Add `resolve_effective_permissions` across the full stack (data types → DB source → repository → service/action) to answer "what operations can user X perform on entities [A, B, C]?"
- Reuses the existing scope chain CTE from BA-5776, removing the operation filter and collecting all permitted operations via UNION of scope-chain and self-scope queries
- Returns `dict[str, set[OperationType]]` mapping entity ID → permitted operations

## Test plan
- [x] Component tests for DB source query (scope chain + self-scope union)
- [x] Repository method integration test with resilience policy

Resolves BA-5797

### DIFF
--- a/changes/11236.feature.md
+++ b/changes/11236.feature.md
@@ -1,1 +1,1 @@
-Add an effective permissions resolver API that returns all permitted operations a user can perform on given entities by traversing the RBAC scope chain.
+Add an effective permissions resolver in the permission controller service and repository layers that returns all permitted operations a user can perform on given entities by traversing the RBAC scope chain.
--- a/changes/11236.feature.md
+++ b/changes/11236.feature.md
@@ -1,0 +1,1 @@
+Add an effective permissions resolver API that returns all permitted operations a user can perform on given entities by traversing the RBAC scope chain.
--- a/src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
@@ -40,6 +40,8 @@ from ai.backend.manager.data.permission.role import (
     BulkRoleRevocationFailure,
     BulkRoleRevocationResultData,
     BulkUserRoleRevocationInput,
+    EffectivePermissionsInput,
+    EffectivePermissionsResult,
     ProjectRoleCount,
     RoleListResult,
     RolePermissionsUpdateInput,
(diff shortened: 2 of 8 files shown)


changes/11236.feature.md
```diff
@@ -0,0 +1 @@
+Add an effective permissions resolver API that returns all permitted operations a user can perform on given entities by traversing the RBAC scope chain.
```
jopemachine: The news fragment indicates that an API is being added, but this PR only introduces the service layer in practice.

src/ai/backend/manager/services/permission_contoller/actions/resolve_effective_permissions.py
```diff
@@ -0,0 +1,50 @@
...
+    target_entity_ids: list[str]
+    permission_entity_type: EntityType | None = None
+
+    @override
+    def entity_id(self) -> str | None:
+        return str(self.user_id)
+
+    @override
+    @classmethod
+    def entity_type(cls) -> EntityType:
```
jopemachine, line 33: Is it intentional that entity_id is a user_id even though entity_type is not USER?

fregataa: We decided to not expand/use entity type enum anymore and this abstract method is also deprecated. it is implemented to avoid type check for now

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
```diff
@@ -1088,6 +1091,88 @@ async def check_bulk_permission_with_scope_chain(
...
+                    perm.c.scope_type == target_scope_type,
+                    perm.c.scope_id.in_(data.target_entity_ids),
+                    perm.c.entity_type == permission_entity_type,
+                )
+            )
+        )
+
+        combined_query = sa.union_all(scope_chain_query, self_scope_query)
+
+        permissions: defaultdict[str, set[OperationType]] = defaultdict(set)
```
seedspirit, line 1168: How about init result dict keys with requested entity ids first? Since the current query does not return information for entity IDs that do not match, it seems better to ensure that it returns an empty set for entity IDs that do not match at all.
```suggestion
permissions: dict[str, set[OperationType]] = {
    entity_id: set() for entity_id in data.target_entity_ids
}
```

jopemachine (approved):

tests/unit/manager/repositories/permission_controller/test_resolve_effective_permissions.py
```diff
@@ -0,0 +1,888 @@
...
+                    )
+                )
+                await db_sess.flush()
+
+    # ── Helpers ──
+
+    def _make_input(
+        self,
+        fixture: EffectiveFixture,
+        entity_type: EntityType = EntityType.VFOLDER,
```
jopemachine, line 427: Nit: entity_type seems useless

PR #11240, opened 2026-04-23
feat(BA-5778): wire BulkActionRBACValidator to bulk permission check

fregataa:
## Summary
- Implement `BulkActionRBACValidator.validate()` to call `check_bulk_permission_with_scope_chain()` and split `entity_ids` into allowed/denied entries for the Bulk Processor's partial-success reporting (BA-5777).
- Superadmins bypass the DB lookup; missing user context raises `UserNotFound`, mirroring `SingleEntityActionRBACValidator`.
- Add `TestBulkActionRBACValidator` covering superadmin bypass, missing user, partial allow/deny split via self-scope permission, full-deny, and empty-batch cases.

## Test plan
- [x] `pants fmt / fix / lint` on changed files
- [x] `pants check tests/unit/manager/actions/validators/test_rbac_validators.py src/ai/backend/manager/actions/validators/rbac/bulk.py`
- [x] `pants test tests/unit/manager/actions/validators/test_rbac_validators.py`
- [x] `pants test tests/unit/manager/actions/test_bulk_processor.py` (no regression)

Resolves BA-5778

### DIFF
--- a/src/ai/backend/manager/actions/validators/rbac/bulk.py
+++ b/src/ai/backend/manager/actions/validators/rbac/bulk.py
@@ -1,6 +1,7 @@
 from typing import Any, override
 
 from ai.backend.common.contexts.user import current_user
+from ai.backend.common.exception import UnreachableError
 from ai.backend.manager.actions.action import BaseActionTriggerMeta
 from ai.backend.manager.actions.action.bulk import BaseBulkAction
 from ai.backend.manager.actions.validator.bulk import (
@@ -9,7 +10,6 @@ from ai.backend.manager.actions.validator.bulk import (
     DeniedEntity,
 )
 from ai.backend.manager.data.permission.role import BulkPermissionCheckInput
-from ai.backend.manager.errors.user import UserNotFound
 from ai.backend.manager.repositories.permission_controller.repository import (
     PermissionControllerRepository,
 )
@@ -35,7 +35,7 @@ class BulkActionRBACValidator(BulkActionValidator):
     ) -> BulkValidationResult:
         user = current_user()
         if user is None:
-            raise UserNotFound("User not found in context")
+            raise UnreachableError("User context is not available")
         entity_ids = list(action.entity_ids)
         if user.is_superadmin:
             return BulkValidationResult(
--- a/src/ai/backend/manager/actions/validators/rbac/bulk.py
+++ b/src/ai/backend/manager/actions/validators/rbac/bulk.py
@@ -1,14 +1,20 @@
 from typing import Any, override
 
+from ai.backend.common.contexts.user import current_user
 from ai.backend.manager.actions.action import BaseActionTriggerMeta
 from ai.backend.manager.actions.action.bulk import BaseBulkAction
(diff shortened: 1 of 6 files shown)


src/ai/backend/manager/actions/validators/rbac/bulk.py
```diff
@@ -27,9 +33,31 @@ def name(cls) -> str:
     async def validate(
         self, action: BaseBulkAction[Any], meta: BaseActionTriggerMeta
     ) -> BulkValidationResult:
-        # TODO: wire this to PermissionControllerRepository.check_bulk_permission_with_scope_chain().
-        # Until then, every entity is treated as allowed so legacy behavior is preserved.
+        user = current_user()
+        if user is None:
+            raise UserNotFound("User not found in context")
```
seedspirit: Unreachable error seems more appropriate for the current convention.
```suggestion
            raise UnreachableError("User context is not available")
```

seedspirit (commented):
Does "Bulk Validation" mean that permission checks for multiple operations performed by a single user are carried out in bulk? Could you please confirm whether this refers to validating operations performed by a single user rather than multiple users?

fregataa:
> Does "Bulk Validation" mean that permission checks for multiple operations performed by a single user are carried out in bulk? Could you please confirm whether this refers to validating operations performed by a single user rather than multiple users?

bulk validation means validating multiple entities in a single user, single entity type, single operation type

src/ai/backend/manager/actions/validators/rbac/bulk.py
```diff
@@ -1,15 +1,21 @@
...
     BulkActionValidator,
     BulkValidationResult,
+    DeniedEntity,
 )
+from ai.backend.manager.data.permission.role import BulkPermissionCheckInput
 from ai.backend.manager.repositories.permission_controller.repository import (
     PermissionControllerRepository,
 )
 
+_DENY_REASON = "permission_denied"
```
jopemachine, line 17: This value doesn’t seem particularly useful—was there a reason for extracting it as a constant?

jopemachine:
CI is broken

jopemachine (approved):

src/ai/backend/manager/actions/validators/rbac/bulk.py
```diff
@@ -27,9 +33,31 @@ def name(cls) -> str:
...
+        user = current_user()
+        if user is None:
+            raise UnreachableError("User context is not available")
+        entity_ids = list(action.entity_ids)
+        if user.is_superadmin:
+            return BulkValidationResult(
+                allowed_entity_ids=entity_ids,
+                denied_entities=[],
+            )
+        permission_map = await self._repository.check_bulk_permission_with_scope_chain(
```
jopemachine, line 45: check_bulk_permission_with_scope_chain currently returns only a boolean indicating whether the permission check passed. Ideally, it should also return a meaningful denied reason. At the moment, the denied reason is just a constant value that isn’t particularly useful.

PR #11249, opened 2026-04-23
fix(BA-5809): map new users to project member role on creation

fregataa:
## Summary
- Route project membership during user creation through `_add_user_to_project_in_session`, the same RBAC-aware path used by `modifyGroup` / `modify_user`.
- Drop `additional_scope_refs=project_scope_refs` from `RBACEntityCreator` (now only the DOMAIN primary scope), delete the now-redundant `_add_user_to_groups` helper, and inline a regression test asserting that newly created users are mapped to their project's member role.
- Fixes folder count / user storage usage coming back empty for users created with project assignments under RBAC enforcement.

## Test plan
- [x] `pants fmt / fix / lint --changed-since=origin/main` pass.
- [x] `pants test tests/unit/manager/repositories/user/test_user_repository.py` — all 13 relevant tests pass, including the new `test_create_user_validated_maps_project_member_role`.

Resolves BA-5809

### DIFF
--- a/src/ai/backend/manager/repositories/user/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/user/db_source/db_source.py
@@ -50,7 +50,6 @@ from ai.backend.manager.models.group import (
     AssocGroupUserRow,
     GroupRow,
     ProjectType,
-    association_groups_users,
 )
 from ai.backend.manager.models.kernel import (
     AGENT_RESOURCE_OCCUPYING_KERNEL_STATUSES,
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/repositories/user/db_source/db_source.py
jopemachine, line 170: Does `create_user_with_keypair` not require the patch?

fregataa:
Closed since we decided to migrate association_groups_users to association_scopes_entities

PR #11289, opened 2026-04-24
feat(BA-5684): migrate association_groups_users data to rbac

fregataa:
## Summary
Backfill `association_scopes_entities` with `PROJECT/USER` membership rows derived from `association_groups_users` so RBAC read paths see consistent membership on deployments that predate the auto-sync path.

- `INSERT ... SELECT` from agus (no JOIN; FKs are `ON DELETE CASCADE`)
- `ON CONFLICT (scope_type, scope_id, entity_id) DO NOTHING` — idempotent and safe under concurrent auto-sync
- `downgrade()` is a no-op; re-running `upgrade()` stays idempotent

## Performance

Measured on local halfstack PostgreSQL, all rows fresh (no pre-existing ASE matches):

| agus rows | First run (new inserts) | Re-run (all conflict) |
|-----------|-------------------------|-----------------------|
| 10,000 | 197 ms | 106 ms |
| 100,000 | 2,234 ms | 1,126 ms |

Scales near-linearly (~11x time for 10x data). Query plan is a sequential scan of agus plus a unique-index conflict check against `uq_scope_id_entity_id`.

## Test plan
- [x] Local halfstack: baseline drift healed, pre-existing ASE rows untouched (ON CONFLICT), re-run inserts 0 rows
- [x] Timed against 10k and 100k synthetic agus rows (see table above)

Resolves BA-5684

### DIFF
--- a/changes/11289.feature.md
+++ b/changes/11289.feature.md
@@ -1,0 +1,1 @@
+Add an idempotent Alembic migration that backfills `association_scopes_entities` with PROJECT/USER membership rows derived from `association_groups_users`, so RBAC read paths see consistent membership state on deployments that predate auto-sync.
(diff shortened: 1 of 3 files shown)


HyeockJinKim (approved):
Please just share the results of the test run.

PR #11309, opened 2026-04-26
Security: SSH host key verification disabled (MITM risk)

tomaioo:
## Summary

Security: SSH host key verification disabled (MITM risk)

## Problem

**Severity**: `High` | **File**: `scripts/agent/deploy-static/deploy_static_files.py:L13`

The deployment script sets `AutoAddPolicy`, which trusts unknown SSH host keys automatically. This allows man-in-the-middle attacks during deployment, potentially exposing credentials and enabling command interception/modification.

## Solution

Use strict host key verification (`RejectPolicy`) and pre-provision known host keys (e.g., via `load_system_host_keys()` or a managed known_hosts file). Fail deployment on host key mismatch.

## Changes

- `scripts/agent/deploy-static/deploy_static_files.py` (modified)

### DIFF
--- a/scripts/agent/deploy-static/deploy_static_files.py
+++ b/scripts/agent/deploy-static/deploy_static_files.py
@@ -11,7 +11,8 @@ if len(sys.argv) == 1:
 STATIC_FILE = 'https://backend-ai-k8s-agent-static.s3.ap-northeast-2.amazonaws.com/bai-static.tar.gz'
 
 ssh = paramiko.SSHClient()
-ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+ssh.load_system_host_keys()
+ssh.set_missing_host_key_policy(paramiko.RejectPolicy())
 
 for ip in sys.argv[1:]:
     ssh.connect(ip)


rapsealk:
Thanks for the security fix — the direction is correct (`AutoAddPolicy` is a textbook MITM hazard, especially in this script where the trusted SSH session executes `sudo rm -rf /opt/backend.ai` and `wget | tar` chains). A few items to address before this can land:

### Changelog required

This repo uses [towncrier](https://github.com/lablup/backend.ai/blob/main/changes/README.md) and every PR needs a news fragment. Please add one under `changes/`:

- File: `changes/11309.fix.md` (use the PR number, `.fix` for the bug-fix type)
- Suggested content (single imperative sentence):

  > Reject unknown SSH host keys in `deploy_static_files.py` to prevent MITM during static-file deployment.

While here, the PR title `Security: SSH host key verification disabled (MITM risk)` describes the *problem*, not the *fix*, and lacks the conventional-commit prefix this repo enforces (the title becomes the squash-merge commit message). Please rename to:

> `fix: Reject unknown SSH host keys in deploy_static_files`

### Supporting references

This isn't just a stylistic preference — `AutoAddPolicy` is flagged as a vulnerability by Paramiko's own docs and three independent security authorities:

- **Paramiko upstream** ([API reference](https://docs.paramiko.org/en/latest/api/client.html)) — `RejectPolicy` is the **documented default**: *"Policy for automatically rejecting the unknown hostname & key."* `AutoAddPolicy` is described as *"Policy for automatically adding the hostname and new host key to the local HostKeys object, and saving it"* — i.e., trust-on-first-use with no verification. This PR is restoring upstream's intended secure default.
- **GitHub CodeQL** — [`py/paramiko-missing-host-key-validation`](https://codeql.github.com/codeql-query-help/python/py-paramiko-missing-host-key-validation/) — CWE-295 (Improper Certificate Validation): *"Accepting unknown host keys exposes connections to man-in-the-middle attacks."* Recommended fix is `RejectPolicy`.
- **AWS CodeGuru / Amazon Q** — [`do-not-auto-add-or-warning-missing-hostkey-policy`](https://docs.aws.amazon.com/codeguru/detector-library/python/do-not-auto-add-or-warning-missing-hostkey-policy/) — **High severity**, CWE-322. Compliant example uses `RejectPolicy`, identical to this PR.
- **Code Pathfinder SAST rule** — [`PYTHON-LANG-SEC-071`](https://codepathfinder.dev/registry/python/lang/PYTHON-LANG-SEC-071) "Paramiko Implicit Host Key Trust (AutoAddPolicy)."

### Operational concern with `load_system_host_keys()`

Per [Paramiko's docs](https://docs.paramiko.org/en/latest/api/client.html), `load_system_host_keys()` *"attempts to read from the user's local known hosts file without raising errors if unavailable"* — meaning in CI runners, fresh jumphosts, or root-owned deploy contexts (where `~/.ssh/known_hosts` is empty or absent), it silently loads nothing. The very first run will then abort with `SSHException: Server ... not found in known_hosts` and leave the operator guessing. Consider one of:

1. Add `ssh.load_host_keys(<managed_path>)` pointing at a known_hosts file checked into the deploy environment, or
2. Add a short comment/README note documenting the pre-provisioning step (`ssh-keyscan <ip> >> ~/.ssh/known_hosts`) so the next on-call doesn't get paged.

### Per-host error handling

With `RejectPolicy`, the first unknown host in `sys.argv[1:]` now raises and aborts the entire batch loop with an uncaught exception — one bad fingerprint skips every remaining IP, and the operator only sees a Python traceback rather than which IP failed. Wrapping the loop body in `try/except paramiko.SSHException` and logging the offending IP would make this safe to use against a list of nodes.

### Unrelated but adjacent (not blocking)

These pre-date your PR; flagging for awareness, not as a request:

- `deploy_static_files.py:33` shell-interpolates `pwd`/`whoami` from remote stdout into a `sudo`-bearing command string.
- One `SSHClient` is reused across all IPs without `close()` between iterations.
- `import os`, `import subprocess`, and the `STATIC_FILE` constant are unused.

### Process

The `license/cla` check is still pending — please sign the CLA via the link in the bot comment so this is unblockable.

PR #11317, opened 2026-04-26
fix(BA-5845): bind signup users to default project via association_scopes_entities

fregataa:
## Summary
- Stop writing to `association_groups_users` from the signup flow; route project membership through a new `UserRepository.assign_project_membership` that performs the ASE scope binding plus the project member role mapping in one transaction.
- `AuthService.signup` resolves the default project id via `GroupRepository.project_id_by_name_in_domain` and delegates the membership write to the user repository, so the auth repository no longer needs to know about projects or RBAC scope tables.
- Closes a gap exposed by BA-5818: signup-created users can now be recognised by ASE-based membership checks (session / deployment / model-serving creation) and gain the project member role at the same time.

## Test plan
- [x] `pants fmt :: && pants fix :: && pants lint --changed-since=origin/main`
- [x] `pants check --changed-since=origin/main`
- [x] `pants test tests/unit/manager/services/auth:: tests/component/auth::`
- [x] New component test asserts an ASE row exists with `(scope_type=PROJECT, scope_id=<default_project>, entity_type=USER, entity_id=<user>)` after signup.
- [x] Manual smoke: signup → login → search vfolders in default project as USER role.

Resolves BA-5845


src/ai/backend/manager/services/auth/service.py
```diff
@@ -494,19 +502,22 @@ async def signup(self, action: SignupAction) -> SignupActionResult:
...
                 keypair_data=kp_data,
-                group_name=group_name,
-                domain_name=action.domain_name,
             )
         except UserCreationError as e:
             raise InternalServerError("Error creating user account") from e
 
+        # Assign the new user to the default project
+        group_name = user_data_overriden.get("group", "default")
+        project_id = await self._group_repository.project_id_by_name_in_domain(
```
jopemachine, line 516: I'm not convinced that the action converting a name to an ID needs to be defined per scope. Wouldn’t it be enough to keep the permission check in the service layer and have only one repository method?

src/ai/backend/manager/services/auth/service.py
```diff
@@ -494,19 +502,22 @@ async def signup(self, action: SignupAction) -> SignupActionResult:
...
                 user_data=data,
                 keypair_data=kp_data,
-                group_name=group_name,
-                domain_name=action.domain_name,
             )
         except UserCreationError as e:
             raise InternalServerError("Error creating user account") from e
 
+        # Assign the new user to the default project
+        group_name = user_data_overriden.get("group", "default")
```
jopemachine: Having a default string value here feels awkward and brittle.

tests/component/auth/test_auth.py
```diff
@@ -98,11 +102,57 @@ class _RSAKeypairData:
     private_key: str
 
 
+@dataclass
+class _SignupDefaultProjectData:
+    """Holds the test domain's ``default`` project and tracks signup-created
+    user emails so the fixture can clean them up on teardown."""
+
+    group_id: uuid.UUID
```
jopemachine: Nit: project_id

src/ai/backend/manager/services/auth/service.py
fregataa: `group_name` is supplied by PRE_SIGNUP hook plugins as a free-form string, so dropping the name-based resolver requires redefining the hook contract and migrating every plugin.
That's out of scope for this fix — I'd prefer to consolidate the resolver in a follow-up once the hook contract is updated.

seedspirit (approved):

PR #11327, opened 2026-04-27
fix(BA-5794): auto-load .env in v2 CLI to restore v1 behavior

jopemachine:
## Summary

- v2 CLI was reading `BACKEND_*` env vars via `os.environ` directly without ever invoking `load_dotenv()`, so users with a local `.env` file had to manually `export` variables — a regression from v1 where `get_env()` implicitly loaded `.env` on every lookup.
- Add a process-level `_load_dotenv_once()` helper in `cli/v2/helpers.py` and invoke it from the `v2()` Click group callback (covers all subcommands including `login`/`logout`) and from `load_v2_config()` (defense-in-depth for non-Click callers).
- Use `override=True` to preserve v1 semantics where `.env` wins over pre-existing shell env.

## Test plan

- [x] New unit tests in `tests/unit/client/cli/test_v2_dotenv.py` cover: `.env` is picked up from cwd, `override=True` semantics, idempotency of `_load_dotenv_once()`, and `load_v2_config()` reads values from `.env`.
- [x] `pants fmt fix lint check` pass on the changed files.
- [ ] Manual: drop a `.env` with `BACKEND_ENDPOINT=...` in cwd, run `backend.ai v2 config show`, confirm the value is applied.

Resolves BA-5794

### DIFF
--- a/src/ai/backend/client/cli/v2/helpers.py
+++ b/src/ai/backend/client/cli/v2/helpers.py
@@ -33,9 +33,10 @@ DEFAULTS = {
 def _load_cwd_dotenv() -> None:
     """Load ``.env`` from the current working directory (walking upward).
 
-    Mirrors v1 CLI behavior where ``get_env()`` implicitly called ``load_dotenv()``
-    on every lookup. ``override=True`` preserves v1 semantics where ``.env`` wins
-    over pre-existing shell env vars.
+    Called from the ``v2()`` Click group callback so all v2 subcommands see
+    ``BACKEND_*`` variables defined in a project-local ``.env``. Mirrors v1
+    CLI behavior; ``override=True`` preserves v1 semantics where ``.env``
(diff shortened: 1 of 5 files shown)


src/ai/backend/client/cli/v2/helpers.py
```diff
@@ -45,6 +57,10 @@ class V2ConnectionConfig:
 def load_v2_config() -> V2ConnectionConfig:
     """Load v2 connection config from ``~/.backend.ai/``.
 
+    ``.env`` auto-loading happens once in the ``v2`` Click group callback;
+    callers reaching this function outside that group must invoke
+    ``_load_cwd_dotenv()`` themselves if they want ``.env`` picked up.
+
```
jopemachine: Delete this

HyeockJinKim (approved):

Issue #5228, opened 2025-07-21
AppProxy returns HTML instead of JSON for invalid API parameters


rapsealk:
## Summary

When the AppProxy receives invalid or malformed API parameters, it responds with an HTML error page instead of a JSON-formatted error response. This behavior breaks clients expecting a structured API response.

POST /v2/endpoints/{endpoint_id}

[https://github.com/lablup/backend.ai/blob/e62b356ef330fbe1c0b02c504a0e53d044d7cc52/src/ai/backend/wsproxy/api/utils.py#L240-L241](https://github.com/lablup/backend.ai/blob/e62b356ef330fbe1c0b02c504a0e53d044d7cc52/src/ai/backend/wsproxy/api/utils.py#L240-L241) 

## Steps to Reproduce

```python
resp = sess.Service.create(
    self.config["environment"]["image"],
    model_id_or_name=self.config["service"]["model"],
    initial_session_count=self.config["service"]["replicas"],
    service_name=service_name,
    runtime_variant=self.config["service"].get("runtime_variant"),
    model_mount_destination=model_mount_destination,
    model_definition_path=self.config["service"].get("model_definition_path"),
    cluster_mode=cluster_mode,
    cluster_size=cluster_size,
    extra_mounts=mounts,  # TODO: No model-type folders here
    extra_mount_map=mount_map,
    extra_mount_options=mount_options,
    envs={**self.config["environment"]["envs"], **envs},
    resources=self.config["resources"],
    resource_opts=self.config.get("resource_opts", {}),
    domain_name=ownership["domain_name"],
    group_name=ownership["project"],
    scaling_group=ownership["scaling-group"],
    expose_to_public=config.open_to_public,
)
```

## Expected Behavior

-

## Actual Behavior

-

## Logs/Errors

```
2025-07-21 14:20:27.741 INFO ai.backend.manager.registry [25961] Created endpoint 09998a83-250a-4c1c-b662-8b77e33ec119 on appproxy: <!DOCTYPE html>
<html>
<head>
<meta charset="UTF-8">
[...]

PR #11333, opened 2026-04-27
fix(BA-1929): Surface AppProxy client endpoint errors as domain exceptions

### LINKED ISSUE #11331: AppProxyClient endpoint methods do not surface error responses as domain exceptions
## Summary

`AppProxyClient` in `src/ai/backend/manager/clients/appproxy/client.py` has inconsistent error handling across its methods. Only `fetch_status` maps low-level `aiohttp` exceptions to the `AppProxyConnectionError` / `AppProxyResponseError` domain exceptions defined in `manager/errors/appproxy.py`. The four endpoint methods — `create_endpoint`, `create_endpoints_bulk`, `delete_endpoint`, `delete_endpoints_bulk` — leak raw aiohttp exceptions or, in the case of `delete_endpoint`, silently swallow non-2xx responses entirely.

This was flagged during review of #11328 (BA-1929) as an out-of-scope but real robustness gap.

## Concrete defects

1. **`delete_endpoint`** uses `async with ... as resp: pass` with no `resp.raise_for_status()` and no body read. A 4xx/5xx response from the coordinator is silently dropped — the manager logs and returns successfully, even though the deletion never happened.
2. **`create_endpoint`, `create_endpoints_bulk`, `delete_endpoints_bulk`** call `resp.raise_for_status()` followed by `await resp.json()`. Both can raise raw `aiohttp.ClientResponseError` / `aiohttp.ContentTypeError`, neither of which is wrapped into a `BackendAIError` subclass. Callers in `manager/sokovan/deployment/executor.py` then see a non-domain exception, and the eventual `DeploymentExecutionError` does not carry the AppProxy domain.
3. The structured `BackendAIError` JSON body returned by the coordinator on validation failures (after #11329) is not preserved through to the caller. The error gets re-raised as an aiohttp exception with only the status code.

## Proposed fix

Apply the same `try / except (ClientConnectorError, ClientResponseError, ContentTypeError, JSONDecodeError)` pattern that `fetch_status` uses to all four endpoint methods, and add `raise_for_status()` to `delete_endpoint`. Where possible, attach the parsed coordinator error body to `AppProxyResponseError.extra_data` so the upstream JSON error survives the translation.

## Out of scope

- Re-architecting the resilience policy or retry behavior.
- Changing the public signatures of the four methods.

## Related

- #5228 — original bug report (HTML vs JSON)
- #11328 — manager-side `Accept` header fix (BA-1929)
- #11329 — coordinator-side default-to-JSON fix (BA-1929)

rapsealk:
## Summary

Closes #11331. Builds on #11328 (LTS hotfix) so this branch contains its commits — once #11328 merges, the diff here will collapse to just the error-mapping change.

`AppProxyClient`'s four mutating methods had inconsistent error handling:

- `delete_endpoint` did `async with ... as resp: pass` — non-2xx responses were silently dropped, so a failed deletion looked successful to the manager.
- `create_endpoint`, `create_endpoints_bulk`, `delete_endpoints_bulk` called `resp.raise_for_status()` then `await resp.json()`, leaking raw `aiohttp.ClientResponseError` / `aiohttp.ContentTypeError` to callers. Neither inherits from `BackendAIError`, so the deployment executor saw a non-domain exception and the eventual `DeploymentExecutionError` lost the AppProxy domain context.
- The structured `BackendAIError` JSON body returned by the coordinator (now reliable after #11328 / #11329) was discarded.

This PR introduces two small helpers on `AppProxyClient`:

- `_request` (async context manager): wraps `ClientConnectorError` → `AppProxyConnectionError`, and any non-2xx status → `AppProxyResponseError` with the upstream body attached as `extra_data` (parsed JSON when possible, raw text otherwise).
- `_parse_json`: maps `ContentTypeError` / `JSONDecodeError` from the success path → `AppProxyResponseError`.

All four endpoint methods now route through `_request`. `fetch_status` is left alone — it already maps to the domain exceptions and exercises a different code path.

## Suggested merge order

1. **#11328** — manager-side `Accept: application/json` (LTS hotfix half)
2. **#11329** — coordinator-side default-to-JSON (LTS hotfix half)
3. **#11333** — this PR (broader hardening)

#11328 and #11329 are independent of each other and resolve the user-visible #5228 symptom on their own. #11333 depends on #11328 and should land last.

## Why this can wait for the next normal cycle (not a hotfix)

#11328 + #11329 already eliminate the user-reported HTML-vs-JSON symptom from #5228. This PR is the broader hardening so the manager doesn't lose error context the next time the coordinator returns a non-2xx — separately reviewable, separately revertable.

## Test plan

- [ ] Force a 400 from the coordinator (invalid `CreateEndpointRequestBody`) and confirm the manager raises `AppProxyResponseError` with the coordinator's `BackendAIError` JSON body attached as `extra_data`.
- [ ] Force a 500 from the coordinator and confirm the same.
- [ ] Force a non-JSON 4xx response (e.g. via reverse proxy) and confirm `AppProxyResponseError` carries the raw text in `extra_data["body"]`.
- [ ] Take down the AppProxy and confirm `AppProxyConnectionError` is raised, not raw `ClientConnectorError`.
- [ ] Force `delete_endpoint` to receive a 4xx and confirm it now raises (was silently swallowed before).
- [ ] Verify `pants check` / `pants lint` pass.

## Notes for reviewer

- This branch is based on `fix/BA-1929-appproxy-client-accept-json` (#11328) so the `Accept: application/json` change is a parent commit, not a duplicate.
- After #11328 lands, this branch can be rebased onto main; the diff will then show only the error-mapping commit.

### DIFF
--- a/changes/11328.fix.md
+++ b/changes/11328.fix.md
@@ -1,1 +1,0 @@
-Send `Accept: application/json` from the manager's AppProxy client so endpoint create/delete failures return parseable JSON instead of HTML error pages.
--- a/changes/11328.fix.md
+++ b/changes/11328.fix.md
@@ -1,0 +1,1 @@
+Send `Accept: application/json` from the manager's AppProxy client so endpoint create/delete failures return parseable JSON instead of HTML error pages.
--- a/src/ai/backend/manager/clients/appproxy/client.py
+++ b/src/ai/backend/manager/clients/appproxy/client.py
@@ -97,12 +97,6 @@ class AppProxyClient:
                 extra_msg=f"Invalid response from AppProxy at {self._address}"
             ) from e
 
-    def _auth_headers(self) -> dict[str, str]:
-        return {
-            "Accept": "application/json",
-            "X-BackendAI-Token": self._token,
-        }
-
     @appproxy_client_resilience.apply()
     async def create_endpoint(
         self,
@@ -112,7 +106,10 @@ class AppProxyClient:
         async with self._client_session.post(
             f"/v2/endpoints/{endpoint_id}",
             json=body.model_dump(mode="json"),
-            headers=self._auth_headers(),
+            headers={
+                "Accept": "application/json",
+                "X-BackendAI-Token": self._token,
+            },
         ) as resp:
             resp.raise_for_status()
             result: dict[str, Any] = await resp.json()
@@ -133,7 +130,10 @@ class AppProxyClient:
         async with self._client_session.post(
             "/v2/endpoints/bulk",
             json=body.model_dump(mode="json"),
-            headers=self._auth_headers(),
+            headers={
+                "Accept": "application/json",
+                "X-BackendAI-Token": self._token,
+            },
         ) as resp:
             resp.raise_for_status()
             payload = await resp.json()
@@ -146,7 +146,10 @@ class AppProxyClient:
     ) -> None:
         async with self._client_session.delete(
             f"/v2/endpoints/{endpoint_id}",
-            headers=self._auth_headers(),
+            headers={
+                "Accept": "application/json",
+                "X-BackendAI-Token": self._token,
+            },
         ):
             pass
 
@@ -165,7 +168,10 @@ class AppProxyClient:
             "DELETE",
             "/v2/endpoints/bulk",
             json=body.model_dump(mode="json"),
-            headers=self._auth_headers(),
+            headers={
+                "Accept": "application/json",
+                "X-BackendAI-Token": self._token,
+            },
         ) as resp:
             resp.raise_for_status()
             payload = await resp.json()
--- a/src/ai/backend/manager/clients/appproxy/client.py
+++ b/src/ai/backend/manager/clients/appproxy/client.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 import json
 import logging
+from collections.abc import AsyncIterator
+from contextlib import asynccontextmanager
 from typing import Any
 from uuid import UUID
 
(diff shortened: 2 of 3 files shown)


changes/11328.fix.md
seedspirit: It appears that a news snippet from another PR has been mixed in; this needs to be verified.

src/ai/backend/manager/clients/appproxy/client.py
```diff
@@ -2,6 +2,8 @@
 
 import json
 import logging
+from collections.abc import AsyncIterator
+from contextlib import asynccontextmanager
```
seedspirit: nitpick: 
```suggestion
from contextlib import asynccontextmanager as actxmgr
```

src/ai/backend/manager/clients/appproxy/client.py
```diff
@@ -99,22 +99,93 @@ async def fetch_status(self) -> AppProxyStatusResponse:
...
+                    try:
+                        error_body: Any = json.loads(text) if text else None
+                    except json.JSONDecodeError:
+                        error_body = text
+                    log.error(
+                        "AppProxy at {} returned {} during {}: {!r}",
+                        self._address,
+                        resp.status,
+                        operation,
+                        error_body,
```
seedspirit, line 139: Wouldn't a structure where we use `yield` when `resp.status // 100 == 2` (or up to 300) and handle errors for the rest be more readable? Also, rather than parsing the JSON again after `resp.text`, wouldn’t it be better to try `resp.json` first and fall back to `resp.text` only if that fails? You might find some useful code patterns by looking at the `StorageProxyHTTPClient` code.

seedspirit (changes requested):

fregataa (commented):
I think an error handling layer is needed, rather than a request proxy thing

PR #11344, opened 2026-04-27
feat(BA-5528): add deployment chat CLI for vLLM-backed model services

jopemachine:
## 📚 Stacked PRs

1. 👉 **#11344** — `feat(BA-5528): add deployment chat CLI for vLLM-backed model services` ← you are here
2. ⬇️ #11412 — `feat(BA-5903): persist deployment chat history and replay as request context`

## Summary
- Add `./bai deployment chat <id> "<content>"` for one-shot OpenAI-compatible chat against deployed inference services. Requests are sent directly to the deployment's inference endpoint with optional `Authorization: Bearer <token>` (the value the runtime — vLLM/SGLang/NIM/TGI/custom — was started with), bypassing the Backend.AI manager. Use `--params` to forward runtime-variant-specific sampling knobs.
- Add `./bai deployment chat-config set/show/clear/clear-cache` to register, inspect, and remove per-deployment chat state.
- Auto-derive the request `model` when the user did not specify one: the CLI calls `GET /v1/models` on the inference endpoint, picks `data[0].id`, and caches it as `cache.default_model` for subsequent calls (matches the webui `ChatCard.tsx` fallback). The user is no longer required to run `chat-config set --model` before the first `chat`.
- Persist state under `~/.backend.ai/deployment_chat/`, grouped per-feature (matching the existing `~/.backend.ai/session/` layout used by `./bai login`):
  - `cache.json` — auto-managed: `endpoint_url`, `default_model` (auto-fetched from `/v1/models`), `last_synced_at` (24-h TTL).
  - `config.json` — user-managed: per-deployment `{ token, model }` entries. The user's `model` takes precedence over `cache.default_model`.
- Both files are written via plain `path.write_text()` to match the existing CLI credential-storage convention (`client/cli/v2/config_cmd.py`). On `401`/`403` from the inference endpoint, the cached token for that deployment is cleared and the user is prompted to re-register.
- Add an SDK-side `BackendAIAppProxyClient` base in `client/v2/base_client.py` for direct-to-deployment HTTP traffic (Bearer-token auth, app-proxy-aware JSON parsing) and a thin `DeploymentChatClient` subclass exposing `chat_completion()` and `list_models()` (returning a typed `ListModelsResponse`).

## Model resolution order

When the runtime needs a `model` field for a `chat` call, the CLI walks this list and stops at the first hit:

1. `--model <name>` on the `chat` command line.
2. `config.<deployment-id>.model` — the user's pinned model in `config.json`.
3. `cache.<deployment-id>.default_model` — the auto-derived value in `cache.json`.
4. `GET /v1/models` on the inference endpoint, taking `data[0].id`. The result is written to `cache.default_model` so subsequent calls skip the round trip.

This means a fresh deployment works with zero configuration as long as the runtime serves `/v1/models`; you only need `chat-config set --model` for multi-model deployments where `[0]` is not the right pick.

## Command usage

```bash
# One-shot chat — model is auto-derived on first call from /v1/models
./bai deployment chat <deployment-id> "Hello, who are you?"

# Override the model for one call
./bai deployment chat <deployment-id> "..." --model llama-3-8b-instruct

# Forward runtime-specific sampling knobs as a JSON object
./bai deployment chat <deployment-id> "..." \
    --params '{"temperature": 0.7, "max_tokens": 256}'

# Register a Bearer token for a token-gated deployment
./bai deployment chat-config set <deployment-id> --token <runtime-token>

# Pin a model (overrides the cached default; useful for multi-model deployments)
./bai deployment chat-config set <deployment-id> --model llama-3-8b-instruct

# Set both at once
./bai deployment chat-config set <deployment-id> \
    --token <runtime-token> --model llama-3-8b-instruct

# Inspect what's currently registered/cached (token is masked)
./bai deployment chat-config show <deployment-id>

# Remove the user-managed config entry (token + model) for a deployment
./bai deployment chat-config clear <deployment-id>

# Force-invalidate the auto-managed cache entry (endpoint_url, default_model)
./bai deployment chat-config clear-cache <deployment-id>
```

`chat-config set` writes to `config.json` only — it does not contact the manager, so it stays usable while the deployment is still provisioning or the manager is unreachable. `chat-config clear` and `clear-cache` operate on the two storage files independently: clearing user config never touches the cache, and vice versa.

## On-disk state

State lives under `~/.backend.ai/deployment_chat/` so it stays grouped with the other Backend.AI CLI state directories.

`cache.json` — auto-managed by the CLI; do not hand-edit.

```json
{
  "deployments": {
    "d55e251a-3a70-408d-97a9-ca305502aa58": {
      "endpoint_url": "https://app-proxy.example.com/v1/some-deployment",
      "default_model": "llama-3-8b-instruct",
      "last_synced_at": "2026-04-29T12:34:56.789012+00:00"
    }
  }
}
```

- `endpoint_url` — fetched from the manager's `deployment.network_access.endpoint_url` and refreshed on a 24-hour TTL.
- `default_model` — auto-derived from `GET /v1/models` on first use; never written by `chat-config set`.
- `last_synced_at` — UTC timestamp of the last manager fetch; entries past `CACHE_ENTRY_TTL` (24 h) are treated as a cache miss.

`config.json` — user-managed: one `{ token, model }` entry per deployment.

```json
{
  "deployments": {
    "d55e251a-3a70-408d-97a9-ca305502aa58": {
      "token": "sk-runtime-token-here",
      "model": "llama-3-8b-instruct"
    }
  }
}
```

Either field may be `null` — `chat-config set` upserts only the fields you pass, and an entry is dropped automatically once both fields are cleared. The token is also cleared automatically on `401`/`403` from the inference endpoint so the next `chat` call surfaces the re-register hint instead of silently re-sending a stale credential.

Resolves BA-5528.


src/ai/backend/client/cli/v2/deployment/chat_config.py
```diff
@@ -0,0 +1,154 @@
...
+    DeploymentChatCacheEntry,
+    IncompatibleChatCacheError,
+    load_chat_cache,
+    mask_token,
+    save_chat_cache,
+)
+from ai.backend.client.cli.v2.helpers import create_v2_registry, load_v2_config
+
+
+def _abort(message: str) -> NoReturn:
```
jopemachine: Don't make this

src/ai/backend/client/cli/v2/deployment/chat_config.py
```diff
@@ -0,0 +1,154 @@
...
+        save_chat_cache(cache)
+        click.echo(f"Removed chat cache entry for deployment {deployment_id}.")
+    else:
+        click.echo(f"No chat cache entry for deployment {deployment_id}.")
+
+
+def _print_entry(deployment_id: UUID, entry: DeploymentChatCacheEntry) -> None:
+    click.echo(f"deployment_id : {deployment_id}")
+    click.echo(f"endpoint_url  : {entry.endpoint_url}")
+    click.echo(f"vllm_api_key  : {mask_token(entry.vllm_api_key)}")
```
jopemachine: Don't make this, this is not only for vllm

src/ai/backend/client/v2/chat_dto.py
jopemachine, line 1: Why did you create dtos in client package?

src/ai/backend/client/v2/domains_v2/deployment_chat.py
```diff
@@ -0,0 +1,116 @@
...
+
+from ai.backend.client.exceptions import BackendAPIError, BackendClientError
+from ai.backend.client.v2.chat_dto import ChatCompletionRequest, ChatCompletionResponse
+
+
+class DeploymentChatAuthError(BackendAPIError):
+    """Raised when the inference endpoint rejects the configured API key."""
+
+
+class DeploymentChatClient:
```
jopemachine: This is only for vllm.
Could we make the client usable regardless of the runtime variant?
If not, we should split the types.

tests/unit/client/cli/test_deployment_chat_cache.py
```diff
@@ -0,0 +1,204 @@
...
+    @pytest.mark.skipif(os.name == "nt", reason="POSIX-only permission check")
+    def test_save_resets_too_open_permissions(self, tmp_path: Path) -> None:
+        path = tmp_path / "cache.json"
+        path.write_text("{}", encoding="utf-8")
+        os.chmod(path, 0o644)
+        cache = DeploymentChatCache()
+        cache.upsert(uuid4(), _entry())
+        save_chat_cache(cache, path)
+        mode = stat.S_IMODE(path.stat().st_mode)
+        assert mode == 0o600
```
jopemachine: Why do we need this?

tests/unit/client/cli/test_deployment_chat_cache.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def test_mask_short_token(self) -> None:
+        assert mask_token("short") == "***"
+
+    def test_mask_none(self) -> None:
+        assert mask_token(None) == "<unset>"
+
+
+class TestAtomicWrite:
+    def test_save_does_not_leave_temp_files_on_success(self, tmp_path: Path) -> None:
```
jopemachine: Why is this TestAtomicWrite?

changes/11344.feature.md
```diff
@@ -0,0 +1 @@
+Add `./bai deployment chat` and `./bai deployment chat-config` v2 CLI commands for one-shot OpenAI-compatible chat with deployed vLLM models, including a local cache (`~/.backend.ai/deployment_chat.json`, `0600`) of per-deployment endpoint URLs and API keys.
```
jopemachine: Too verbose

src/ai/backend/client/cli/v2/deployment/chat.py
```diff
@@ -0,0 +1,153 @@
...
+    default=None,
+    type=float,
+    help="Sampling temperature.",
+)
+@click.option(
+    "--max-tokens",
+    default=None,
+    type=int,
+    help="Maximum number of tokens to generate.",
+)
```
jopemachine: Suggest which useful parameters should be added for vLLM.

src/ai/backend/client/v2/chat_dto.py
```diff
@@ -0,0 +1,66 @@
...
+
+class ChatCompletionRequest(BaseModel):
+    model: str
+    messages: list[ChatMessage]
+    temperature: float | None = None
+    max_tokens: int | None = Field(default=None, ge=1)
+
+
+class ChatCompletionMessage(BaseModel):
+    model_config = ConfigDict(extra="allow")
```
jopemachine: Why extra allow needed?

src/ai/backend/client/v2/chat_dto.py
```diff
@@ -0,0 +1,66 @@
...
+
+class ChatCompletionUsage(BaseModel):
+    model_config = ConfigDict(extra="allow")
+
+    prompt_tokens: int | None = None
+    completion_tokens: int | None = None
+    total_tokens: int | None = None
+
+
+class ChatCompletionResponse(BaseModel):
```
jopemachine: It also seems vllm-dedicated

src/ai/backend/client/cli/v2/deployment/chat.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+@click.command(name="chat")
+@click.argument("deployment_id", type=click.UUID)
+@click.argument("content", type=str)
+@click.option(
+    "--model",
+    default=None,
+    type=str,
+    help="Model name to send (defaults to cached default_model or 'default').",
```
jopemachine: It doesn't seem 'default' nessory

src/ai/backend/client/cli/v2/deployment/chat.py
```diff
@@ -0,0 +1,168 @@
...
+            skip_ssl_verification=connection.skip_ssl_verification,
+        ) as client:
+            try:
+                response = await client.chat_completion(
+                    endpoint_entry.endpoint_url,
+                    api_key,
+                    body,
+                )
+            except DeploymentChatAuthError as e:
+                chat_config.clear_token(deployment_id)
```
jopemachine: Why did you remove chat config? It doesn't make sense

src/ai/backend/client/cli/v2/deployment/chat.py
```diff
@@ -0,0 +1,168 @@
...
+                    f"The inference endpoint rejected the configured API key for "
+                    f"deployment {deployment_id}. The cached key has been cleared.\n"
+                    "Register a new one with:\n"
+                    f"  ./bai deployment chat-config set {deployment_id} --token <api_key>"
+                ) from e
+            except BackendAPIError as e:
+                raise click.ClickException(
+                    f"Inference endpoint error ({e.status} {e.reason}): {e.data}"
+                ) from e
+        sys.stdout.write(json.dumps(response, indent=2, ensure_ascii=False, default=str) + "\n")
```
jopemachine: Just use print

src/ai/backend/client/cli/v2/deployment/chat_config.py
```diff
@@ -0,0 +1,253 @@
...
+    type=str,
+    help=(
+        "Override the auto-discovered served model name. "
+        "If omitted, the model is fetched from the inference endpoint's /v1/models."
+    ),
+)
+def set_(
+    deployment_id: UUID,
+    api_key: str | None,
+    no_token: bool,
```
jopemachine: Do not make this. Make separate clear command

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -0,0 +1,253 @@
...
+
+    connection = load_v2_config()
+    try:
+        cache = load_chat_cache()
+    except IncompatibleChatCacheError as e:
+        raise click.ClickException(str(e)) from e
+    try:
+        chat_config_store = load_chat_config()
+    except IncompatibleChatConfigError as e:
+        raise click.ClickException(str(e)) from e
```
jopemachine: Merge this

src/ai/backend/client/cli/v2/deployment/chat_config.py
```diff
@@ -0,0 +1,253 @@
...
+            ),
+        )
+        save_chat_cache(cache)
+        if resolved_key is None:
+            chat_config_store.clear_token(deployment_id)
+        else:
+            chat_config_store.set_token(deployment_id, resolved_key)
+        save_chat_config(chat_config_store)
+
+        click.echo(f"Updated chat cache entry for deployment {deployment_id}.")
```
jopemachine: Is click.echo this convention? not print or cli framework?

src/ai/backend/client/cli/v2/deployment/chat_config.py
```diff
@@ -0,0 +1,253 @@
...
+        DeploymentChatAuthError,
+        DeploymentChatClient,
+    )
+
+    async with DeploymentChatClient(skip_ssl_verification=skip_ssl_verification) as client:
+        try:
+            payload = await client.list_models(endpoint_url, api_key)
+        except (DeploymentChatAuthError, BackendAPIError, BackendClientError):
+            return fallback
+    data = payload.get("data") if isinstance(payload, dict) else None
```
jopemachine: Make response type based pydantic

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -0,0 +1,253 @@
...
+        save_chat_cache(cache)
+    if removed_token:
+        save_chat_config(chat_config_store)
+    if removed_entry or removed_token:
+        click.echo(f"Removed chat cache entry for deployment {deployment_id}.")
+    else:
+        click.echo(f"No chat cache entry for deployment {deployment_id}.")
+
+
+def _print_entry(
```
jopemachine: Move it under the cache entry type method

src/ai/backend/client/cli/v2/deployment_chat_cache.py
```diff
@@ -0,0 +1,111 @@
...
+
+    endpoint_url: str
+    default_model: str | None = None
+    last_synced_at: datetime
+
+
+class DeploymentChatCache(BaseModel):
+    """In-memory representation of the chat cache file."""
+
+    schema_version: int = Field(default=CHAT_CACHE_SCHEMA_VERSION)
```
jopemachine: Remove this.

src/ai/backend/client/v2/deployment_chat.py
```diff
@@ -0,0 +1,162 @@
...
+
+class DeploymentChatAuthError(BackendAPIError):
+    """Raised when the inference endpoint rejects the configured API key."""
+
+
+class DeploymentChatClient:
+    """Direct HTTP client for OpenAI-compatible inference endpoints."""
+
+    _session: aiohttp.ClientSession
+    _owns_session: bool
```
jopemachine: Do not introduce _owns_session; instead, align the session handling with the other client types.

src/ai/backend/client/v2/deployment_chat.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def __init__(
+        self,
+        *,
+        session: aiohttp.ClientSession | None = None,
+        skip_ssl_verification: bool = False,
+        connect_timeout: float | None = 10.0,
+        read_timeout: float | None = 300.0,
+    ) -> None:
```
jopemachine: Create DeploymentChatClientArgs type

src/ai/backend/client/v2/deployment_chat.py
```diff
@@ -0,0 +1,162 @@
...
+
+    async def close(self) -> None:
+        if self._owns_session and not self._session.closed:
+            await self._session.close()
+
+    async def chat_completion(
+        self,
+        endpoint_url: str,
+        api_key: str | None,
+        request: dict[str, Any],
```
jopemachine: I don't this variable name make sense.

tests/unit/client/cli/test_deployment_chat_utils.py
```diff
@@ -0,0 +1,175 @@
...
+def _entry(
+    *,
+    endpoint: str = "https://infer.example.test/api",
+    default_model: str | None = None,
+) -> DeploymentChatCacheEntry:
+    return DeploymentChatCacheEntry(
+        endpoint_url=endpoint,
+        default_model=default_model,
+        last_synced_at=datetime(2026, 4, 27, 12, 0, tzinfo=UTC),
+    )
```
jopemachine: _make_entry

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -0,0 +1,363 @@
...
+        "Omit when the runtime was started without an API key."
+    ),
+)
+@click.option(
+    "--default-model",
+    default=None,
+    type=str,
+    help=(
+        "Override the auto-discovered served model name. "
+        "If omitted, the model is fetched from the inference endpoint's /v1/models."
```
jopemachine: Isn't this vllm dedicated?

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -0,0 +1,363 @@
...
+            DeploymentChatCacheEntry(
+                endpoint_url=endpoint_url,
+                default_model=served_model,
+                last_synced_at=datetime.now(UTC),
+            ),
+        )
+        save_chat_cache(cache)
+        if resolved_key is not None:
+            chat_config_store.set_token(deployment_id, resolved_key)
+            save_chat_config(chat_config_store)
```
jopemachine: There seems to be no need to call save_chat_cache twice.

src/ai/backend/client/cli/v2/deployment/chat/utils.py
```diff
@@ -0,0 +1,150 @@
...
+        return "***"
+    return f"{token[:3]}***...***{token[-4:]}"
+
+
+def _read_json(path: Path) -> dict[str, Any] | None:
+    if not path.exists():
+        return None
+    try:
+        with path.open("r", encoding="utf-8") as f:
+            raw = json.load(f)
```
jopemachine: Use src/ai/backend/common/json.py

src/ai/backend/client/cli/v2/deployment/chat/utils.py
```diff
@@ -0,0 +1,150 @@
...
+) -> None:
+    schema = raw.get("schema_version")
+    if schema is not None and isinstance(schema, int) and schema > supported_version:
+        raise error_cls(
+            f"{filename} schema version {schema} is newer than supported "
+            f"{supported_version}; please upgrade the client."
+        )
+
+
+def _atomic_write(path: Path, payload: str) -> None:
```
jopemachine: Please choose a sensible name.

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -0,0 +1,312 @@
...
+        if endpoint_entry.default_model:
+            print(f"  default_model: {endpoint_entry.default_model}")
+        print(f"  api_key:       {mask_token(resolved_key)}")
+
+    _run_async(_run)
+
+
+@chat_config.command(name="show")
+@click.argument("deployment_id", type=click.UUID, required=False)
+def show(deployment_id: UUID | None) -> None:
```
jopemachine: Making deployment_id nullable to return everything is quite awkward.
We could introduce a separate command like show_all, but since there’s no clear demand for that right now, it would be better to simply make deployment_id required.

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -0,0 +1,312 @@
...
+        return
+    for dep_id in dep_ids:
+        _print_entry(dep_id, cache.get(dep_id), chat_config_store.get_token(dep_id))
+        print()
+
+
+@chat_config.command(name="clear")
+@click.argument("deployment_id", type=click.UUID)
+def clear(deployment_id: UUID) -> None:
+    """Remove the chat cache entry and stored token for a deployment."""
```
jopemachine: It is quite awkward for a single clear command to remove both cache and config (token). Split them into separate commands.

src/ai/backend/client/cli/v2/deployment/chat/commands.py
jopemachine: I said move it under the cache entry type method

src/ai/backend/client/cli/v2/deployment/chat/utils.py
```diff
@@ -0,0 +1,121 @@
...
+from ai.backend.client.cli.v2.helpers import CONFIG_DIR
+from ai.backend.common.json import load_json
+
+CHAT_CACHE_FILE = CONFIG_DIR / "deployment_chat.json"
+CHAT_CONFIG_FILE = CONFIG_DIR / "deployment_chat_config.json"
+
+
+def load_chat_cache(path: Path = CHAT_CACHE_FILE) -> DeploymentChatCache:
+    """Load the chat cache; return an empty cache when the file is absent or unreadable."""
+    raw = _read_json(path)
```
jopemachine: Make pydantic type and model_validate into it. Do not use raw dict

src/ai/backend/client/cli/v2/deployment/chat/utils.py
jopemachine: Remove this. Inline

jopemachine:
Please review the code thoroughly to see whether there are any parts that may not work across all runtime variants, including custom

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -0,0 +1,291 @@
...
+    if cache.remove(deployment_id):
+        save_chat_cache(cache)
+        print(f"Removed cache entry for deployment {deployment_id}.")
+    else:
+        print(f"No cache entry for deployment {deployment_id}.")
+
+
+@chat_config.command(name="clear-token")
+@click.argument("deployment_id", type=click.UUID)
+def clear_token(deployment_id: UUID) -> None:
```
jopemachine: Why clear token? It should be clear config

src/ai/backend/client/cli/v2/deployment/chat/types.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class DeploymentChatCacheEntry(BaseModel):
+    """One deployment's auto-managed endpoint metadata."""
+
+    endpoint_url: str
+    default_model: str | None = None
+    last_synced_at: datetime
+
+    def format_summary(self) -> list[str]:
```
jopemachine: Create Formatter class and make methods for formatting in them (Do not make format_summary, print_summary... etc)

jopemachine:
Reviewed the code with each runtime variant in mind. Compatibility matrix for the current implementation:

| Assumption | vLLM | SGLang | NIM | TGI | custom |
|---|---|---|---|---|---|
| `endpoint_url` from manager | ✓ | ✓ | ✓ | ✓ | ✓ |
| `Authorization: Bearer` header (sent only when token registered) | ✓ | ✓ | ✓ | proxy-dependent | depends |
| Path `/v1/chat/completions` | ✓ | ✓ | ✓ | **only in messages-api mode** | depends |
| Body keys `{model, messages: [{role, content}]}` | ✓ | ✓ | ✓ | only in messages-api mode | depends |
| Response is dict and printed verbatim (no parsing) | ✓ | ✓ | ✓ | ✓ | ✓ |
| `mask_token` / cache I/O / chat-config commands | runtime-agnostic | | | | |

**Variant-agnostic parts** (no fixes needed):
- Endpoint URL resolution via `registry.deployment.get(...)` — manager provides the right URL regardless of runtime variant.
- Auth: Bearer is the most common scheme and silently absent when no token is registered. Containers without auth ignore the missing header; containers with non-Bearer auth (custom header / query param) are out of scope.
- Response handling: SDK returns the raw dict, CLI prints it verbatim. No variant-specific parsing.
- Cache / config storage, `mask_token`, file format: pure persistence, generic.

**Hard-coded OpenAI assumptions** (variant-specific):
1. **Path** was hard-coded to `/v1/chat/completions`.
2. **Body shape** is `{model, messages, ...extras}` with messages as `[{role: \"user\", content}]`.

**Fix in 208ff1863**: expose `--path` on `chat` so users can target other routes (e.g. TGI's native `/generate` if they roll their own body via `--params`, or custom containers with non-default paths). The SDK already had `path` as a kwarg; only the CLI bridge was missing.

**Documented limitation**: body keys are still chat-shaped. Containers that expose a non-chat inference contract (e.g. raw TGI `{inputs, parameters}`) cannot be driven by this command — for those, users should curl directly or use a custom client. Help text now records this explicitly.

If the requirement is for **arbitrary runtime variants including non-chat APIs** to be drivable through this command, the fix is bigger: replace the body-merge with a body-from-`--params` mode (or a separate \`./bai deployment infer\` command for raw POST). Let me know if you want me to take that further.

src/ai/backend/client/cli/v2/deployment/chat/utils.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def load_chat_config(path: Path = CHAT_CONFIG_FILE) -> DeploymentChatConfig:
+    """Load the chat config; return an empty config when the file is absent or unreadable."""
+    raw = _read_json(path)
+    if raw is None:
+        return DeploymentChatConfig()
+    try:
+        return DeploymentChatConfig.model_validate(raw)
+    except ValidationError:
```
jopemachine: Notify the user that the chat config is in an invalid format.

src/ai/backend/client/cli/v2/deployment/chat/types.py
```diff
@@ -0,0 +1,70 @@
...
+            for line in entry.format_summary():
+                print(line)
+        else:
+            print("endpoint_url  : -")
+            print("default_model : -")
+            print("last_synced_at: -")
+        print(f"api_key       : {token_display}")
+
+
+class DeploymentChatCache(BaseModel):
```
jopemachine, line 51: TTL is required for this cache

tests/unit/client/cli/test_deployment_chat_types.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def test_format_summary_dash_for_missing_default_model(self) -> None:
+        entry = _make_entry(default_model=None)
+        lines = entry.format_summary()
+        assert any("default_model : -" in line for line in lines)
+
+
+class TestCacheMutations:
+    def test_upsert_overwrites_existing_make_entry(self) -> None:
+        cache = DeploymentChatCache()
```
jopemachine: Create fixtures for all input-related objects, including DeploymentChatCache.

src/ai/backend/client/cli/v2/deployment/chat/formatter.py
```diff
@@ -0,0 +1,42 @@
...
+                "last_synced_at: -",
+            ]
+        return [
+            f"endpoint_url  : {entry.endpoint_url}",
+            f"default_model : {entry.default_model or '-'}",
+            f"last_synced_at: {entry.last_synced_at.isoformat()}",
+        ]
+
+    @classmethod
+    def render(
```
jopemachine: Name properly

src/ai/backend/client/cli/v2/deployment/chat/utils.py
```diff
@@ -0,0 +1,94 @@
...
+        return DeploymentChatCache.model_validate(raw)
+    except ValidationError:
+        print(
+            f"WARNING: {path} is in an invalid format and was ignored.",
+            file=sys.stderr,
+        )
+        return DeploymentChatCache()
+
+
+def save_chat_cache(cache: DeploymentChatCache, path: Path = CHAT_CACHE_FILE) -> None:
```
jopemachine: Remove path argument.

src/ai/backend/client/cli/v2/deployment/chat/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    """Write the chat cache and enforce ``0600`` permissions."""
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(cache.model_dump_json(indent=2), encoding="utf-8")
+    path.chmod(stat.S_IRUSR | stat.S_IWUSR)
+
+
+def load_chat_config(path: Path = CHAT_CONFIG_FILE) -> DeploymentChatConfig:
```
jopemachine: Remove path argument.

src/ai/backend/client/cli/v2/deployment/chat/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    except ValidationError:
+        print(
+            f"WARNING: {path} is in an invalid format and was ignored. "
+            "Re-register tokens with `./bai deployment chat-config set`.",
+            file=sys.stderr,
+        )
+        return DeploymentChatConfig()
+
+
+def save_chat_config(config: DeploymentChatConfig, path: Path = CHAT_CONFIG_FILE) -> None:
```
jopemachine: Remove path argument.

src/ai/backend/client/cli/v2/deployment/chat/utils.py
```diff
@@ -0,0 +1,94 @@
...
+def mask_token(token: str | None) -> str:
+    """Render a token as ``sk-***...***xxxx`` for diagnostic display."""
+    if token is None:
+        return "<unset>"
+    if len(token) <= 8:
+        return "***"
+    return f"{token[:3]}***...***{token[-4:]}"
+
+
+def _read_json(path: Path) -> dict[str, Any] | None:
```
jopemachine: Check whether this utility function already exists in the existing codebase.

src/ai/backend/client/cli/v2/deployment/chat/types.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class DeploymentChatCacheEntry(BaseModel):
+    """One deployment's auto-managed endpoint metadata."""
+
+    endpoint_url: str
+    default_model: str | None = None
+    last_synced_at: datetime
+
+    def is_fresh(self, *, now: datetime, ttl: timedelta = CACHE_ENTRY_TTL) -> bool:
```
jopemachine: Make it is_expired

src/ai/backend/client/cli/v2/deployment/chat/utils.py
```diff
@@ -0,0 +1,54 @@
...
+    DeploymentChatConfig,
+)
+from ai.backend.client.cli.v2.helpers import CONFIG_DIR
+from ai.backend.common.json import load_json
+
+CHAT_CACHE_FILE = CONFIG_DIR / "deployment_chat.json"
+CHAT_CONFIG_FILE = CONFIG_DIR / "deployment_chat_config.json"
+
+
+def save_chat_cache(cache: DeploymentChatCache) -> None:
```
jopemachine: Like load, move this to an instance method on each type.

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -0,0 +1,344 @@
...
+        detail = title or msg or str(e)
+        raise click.ClickException(f"{status}: {detail}") from e
+
+
+async def _resolve_endpoint_entry(
+    cache: DeploymentChatCache,
+    deployment_id: UUID,
+    connection: V2ConnectionConfig,
+    *,
+    default_model_override: str | None = None,
```
jopemachine: It seems _resolve_endpoint_entry is not used

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -0,0 +1,344 @@
...
+        msg = data.get("msg", "") if isinstance(data, dict) else ""
+        status = e.args[0] if e.args else "?"
+        detail = title or msg or str(e)
+        raise click.ClickException(f"{status}: {detail}") from e
+
+
+async def _resolve_endpoint_entry(
+    cache: DeploymentChatCache,
+    deployment_id: UUID,
+    connection: V2ConnectionConfig,
```
jopemachine: connection_config

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -0,0 +1,310 @@
...
+    from ai.backend.client.v2.deployment_chat import (
+        DeploymentChatClient,
+        DeploymentChatClientArgs,
+    )
+    from ai.backend.client.v2.exceptions import DeploymentChatAuthError
+
+    connection_config = load_v2_config()
+
+    cache = DeploymentChatCache.load()
+    chat_config_store = DeploymentChatConfig.load()
```
jopemachine: chat_config

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -0,0 +1,310 @@
...
+                f"No --model given and no default_model cached for deployment {deployment_id}.\n"
+                "Set one with:\n"
+                f"  ./bai deployment chat-config set {deployment_id} --default-model <name>"
+            )
+
+        body: dict[str, Any] = {
+            **extra_body,
+            "model": request_model,
+            "messages": [{"role": "user", "content": content}],
+        }
```
jopemachine: Create Request type based on pydantic

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -0,0 +1,310 @@
...
+    The deployment's ``endpoint_url`` is auto-managed; ``--token`` and
+    ``--default-model`` are user-supplied.
+    """
+
+
+@chat_config.command(name="set")
+@click.argument("deployment_id", type=click.UUID)
+@click.option(
+    "--token",
+    "api_key",
```
jopemachine: Let’s standardize on the term "token" instead of "API key"

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -0,0 +1,310 @@
...
+                "--default-model will be applied on the first 'chat' call after the "
+                "deployment is READY.",
+                file=sys.stderr,
+            )
+
+        print(f"Updated chat config for deployment {deployment_id}.")
+        if cached_default_model:
+            print(f"  default_model: {cached_default_model}")
+        if api_key is not None:
+            print(f"  api_key:       {mask_token(api_key)}")
```
jopemachine: token

src/ai/backend/client/cli/v2/deployment/chat/formatter.py
```diff
@@ -0,0 +1,51 @@
...
+from ai.backend.client.cli.v2.deployment.chat.types import DeploymentChatCacheEntry
+
+
+def mask_token(token: str | None) -> str:
+    """Render a token as ``sk-***...***xxxx`` for diagnostic display."""
+    if token is None:
+        return "<unset>"
+    if len(token) <= 8:
+        return "***"
+    return f"{token[:3]}***...***{token[-4:]}"
```
jopemachine: I think it'd better to hardcode a placeholder string like ********, regardless of length.

src/ai/backend/client/cli/v2/deployment/chat/types.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def _write_text_file(path: Path, text: str, *, mode: int | None) -> None:
+    """Write ``text`` to ``path`` via a tmp-and-rename so partial writes can't
+    corrupt the destination. When ``mode`` is given, the temp file is created
+    with that POSIX permission so the final file never exists with weaker
+    permissions, even briefly.
+    """
+    path.parent.mkdir(parents=True, exist_ok=True)
+    tmp = path.with_name(path.name + ".tmp")
```
jopemachine: How does the CLI store login credentials?
I’d like to align this with the existing convention used there.

src/ai/backend/client/cli/v2/deployment/chat/types.py
```diff
@@ -0,0 +1,130 @@
...
+    flags = os.O_WRONLY | os.O_CREAT | os.O_TRUNC
+    open_mode = mode if mode is not None else 0o666
```
jopemachine: Let’s extract 0o666 into a named constant instead of using a magic number.

src/ai/backend/client/cli/v2/deployment/chat/types.py
```diff
@@ -0,0 +1,130 @@
...
+            return cls()
+
+    def save(self) -> None:
+        """Persist the config with ``0600`` permissions because it stores plaintext API keys."""
+        from ai.backend.client.cli.v2.deployment.chat.utils import CHAT_CONFIG_FILE
+
+        _write_text_file(CHAT_CONFIG_FILE, self.model_dump_json(indent=2), mode=0o600)
+
+
+def _write_text_file(path: Path, text: str, *, mode: int | None) -> None:
```
jopemachine: Move this to utils.py

src/ai/backend/client/v2/deployment_chat.py
```diff
@@ -0,0 +1,143 @@
...
+DEFAULT_MODELS_PATH = "/v1/models"
+
+
+@dataclass(frozen=True)
+class DeploymentChatClientArgs:
+    """Connection knobs for :class:`DeploymentChatClient`."""
+
+    skip_ssl_verification: bool = False
+    connect_timeout: float | None = 10.0
+    read_timeout: float | None = 300.0
```
jopemachine: Let’s extract these into constants

src/ai/backend/client/v2/deployment_chat.py
```diff
@@ -0,0 +1,143 @@
...
+        raise BackendAPIError(resp.status, resp.reason or "HTTP error", data)
+
+    @staticmethod
+    def _ensure_dict(payload: object) -> dict[str, Any]:
+        if not isinstance(payload, dict):
+            raise BackendClientError(f"inference endpoint returned non-JSON response: {payload!r}")
+        return payload
+
+    @staticmethod
+    async def _read_payload(resp: aiohttp.ClientResponse) -> object:
```
jopemachine: If we call _ensure_dict immediately after _read_payload, it seems we only intend to support JSON responses.
In that case, do we really need separate _read_payload and _ensure_dict helpers? Wouldn’t it be better to read the response directly with resp.json() and include resp.text() in the error message when JSON parsing fails?

src/ai/backend/client/v2/deployment_chat.py
```diff
@@ -0,0 +1,143 @@
...
+    ) -> dict[str, Any]:
+        """GET the list of models served by the deployment endpoint."""
+        return await self._request("GET", endpoint_url, path, api_key)
+
+    async def _request(
+        self,
+        method: str,
+        endpoint_url: str,
+        path: str,
+        api_key: str | None,
```
jopemachine: token

src/ai/backend/client/v2/deployment_chat.py
```diff
@@ -0,0 +1,143 @@
...
+        *,
+        path: str = DEFAULT_CHAT_PATH,
+    ) -> dict[str, Any]:
+        """POST a chat completion request to the deployment endpoint."""
+        return await self._request("POST", endpoint_url, path, api_key, body=body)
+
+    async def list_models(
+        self,
+        endpoint_url: str,
+        api_key: str | None,
```
jopemachine: token

src/ai/backend/client/v2/deployment_chat.py
```diff
@@ -0,0 +1,143 @@
...
+    @staticmethod
+    def _ensure_dict(payload: object) -> dict[str, Any]:
+        if not isinstance(payload, dict):
+            raise BackendClientError(f"inference endpoint returned non-JSON response: {payload!r}")
+        return payload
+
+    @staticmethod
+    async def _read_payload(resp: aiohttp.ClientResponse) -> object:
+        # Inference endpoints normally return JSON, but proxies in front of them
+        # (nginx, app-proxy, cloud LB) often emit HTML/plain-text bodies on 5xx.
```
jopemachine: This requests to the app-proxy

tests/unit/client/cli/test_deployment_chat_utils.py
```diff
@@ -0,0 +1,133 @@
...
+        assert loaded.get_token(dep_id) == "sk-secret-token-1234"
+
+
+class TestPermissions:
+    @pytest.mark.skipif(os.name == "nt", reason="POSIX-only permission check")
+    def test_config_save_enforces_0600(self, config_path: Path) -> None:
+        cfg = DeploymentChatConfig()
+        cfg.set_token(uuid4(), "sk-x")
+        cfg.save()
+        assert stat.S_IMODE(config_path.stat().st_mode) == 0o600
```
jopemachine: Remove this

tests/unit/client/v2/test_deployment_chat_client.py
```diff
@@ -0,0 +1,228 @@
...
+    runner = web.AppRunner(app)
+    await runner.setup()
+    site = web.TCPSite(runner, "127.0.0.1", 0)
+    await site.start()
+    server_obj = site._server
+    assert server_obj is not None
+    sockets = server_obj.sockets
+    assert sockets
+    port = sockets[0].getsockname()[1]
+    return _FakeServer(runner, port, recorded)
```
jopemachine: I think using something like aioresponses to mock responses should be sufficient; spinning up an actual server for tests is unnecessary.

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -0,0 +1,310 @@
...
+@click.option(
+    "--model",
+    default=None,
+    type=str,
+    help="Model name to send (defaults to cached default_model).",
+)
+@click.option(
+    "--path",
+    "request_path",
+    default="/v1/chat/completions",
```
jopemachine: Is this default sensible value?

src/ai/backend/client/cli/v2/deployment/chat/types.py
```diff
@@ -0,0 +1,140 @@
...
+
+    deployments: dict[UUID, DeploymentChatCacheEntry] = Field(default_factory=dict)
+
+    def get(self, deployment_id: UUID) -> DeploymentChatCacheEntry | None:
+        return self.deployments.get(deployment_id)
+
+    def set(self, deployment_id: UUID, entry: DeploymentChatCacheEntry) -> None:
+        self.deployments[deployment_id] = entry
+
+    def remove(self, deployment_id: UUID) -> bool:
```
jopemachine: Rename to pop

src/ai/backend/client/cli/v2/deployment/chat/types.py
```diff
@@ -0,0 +1,140 @@
...
+
+    tokens: dict[UUID, str] = Field(default_factory=dict)
+
+    def get_token(self, deployment_id: UUID) -> str | None:
+        return self.tokens.get(deployment_id)
+
+    def set_token(self, deployment_id: UUID, token: str) -> None:
+        self.tokens[deployment_id] = token
+
+    def clear_token(self, deployment_id: UUID) -> bool:
```
jopemachine: Rename to pop_token

src/ai/backend/client/cli/v2/deployment/chat/types.py
```diff
@@ -0,0 +1,140 @@
...
+class ChatCompletionRequest(BaseModel):
+    """Body for ``POST /v1/chat/completions`` (OpenAI-compatible).
+
+    Extra fields are forwarded so callers can pass runtime-variant-specific
+    knobs (e.g. ``temperature``, ``top_p``, vLLM/NIM extensions) through
+    ``./bai deployment chat --params`` without the CLI having to enumerate
+    them.
+    """
+
+    model_config = ConfigDict(extra="allow")
```
jopemachine: Why is it needed?

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -0,0 +1,302 @@
...
+
+@chat_config.command(name="set")
+@click.argument("deployment_id", type=click.UUID)
+@click.option(
+    "--token",
+    default=None,
+    type=str,
+    help=(
+        "Token the inference runtime accepts as a Bearer credential. "
+        "Omit when the runtime was started without a token."
```
jopemachine: "Omit when the deployment is open to public"

src/ai/backend/client/cli/v2/deployment/chat/formatter.py
```diff
@@ -0,0 +1,55 @@
...
+
+def mask_token(token: str | None) -> str:
+    """Render a stored token as a fixed placeholder for diagnostic display.
+
+    The placeholder is length-independent so the masked output never leaks the
+    token's length, prefix, or suffix.
+    """
+    if token is None:
+        return "<unset>"
+    return TOKEN_PLACEHOLDER
```
jopemachine: Inline

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -16,14 +16,37 @@
...
     ChatCompletionRequest,
     DeploymentChatCache,
     DeploymentChatCacheEntry,
     DeploymentChatConfig,
+    DeploymentChatHistory,
 )
 from ai.backend.client.cli.v2.helpers import create_v2_registry, load_v2_config
 
 
+def _extract_assistant_content(response: dict[str, Any]) -> str | None:
```
jopemachine: Create response type based on pydantic

src/ai/backend/client/cli/v2/deployment/chat/commands.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+# ---------------------------------------------------------------------------
+# chat
+# ---------------------------------------------------------------------------
+
+
+@click.command(name="chat")
+@click.argument("deployment_id", type=click.UUID)
+@click.argument("content", type=str)
```
jopemachine: message

jopemachine: Renamed CLI argument `content` -> `message` in 543cdb192. Updated the click argument, function param, help text, and request-body construction site. The wire-format JSON key stays `content` (OpenAI spec).

src/ai/backend/client/cli/v2/deployment/chat/types.py
```diff
@@ -0,0 +1,245 @@
...
+        entry = self.deployments.get(deployment_id)
+        if entry is None or entry.model is None:
+            return False
+        entry.model = None
+        if entry.is_empty():
+            del self.deployments[deployment_id]
+        return True
+
+    def pop(self, deployment_id: DeploymentID) -> bool:
+        return self.deployments.pop(deployment_id, None) is not None
```
seedspirit: In a typical pop operation, one would expect the operation to return the object that was popped rather than a boolean value

src/ai/backend/client/cli/v2/deployment/chat/types.py
```diff
@@ -0,0 +1,245 @@
...
+    Stored separately from the cache (auto-managed endpoint metadata) and
+    config (user-managed token/model) so that clearing one does not affect
+    the others. The transcripts are FIFO-truncated at
+    :data:`MAX_PERSISTED_HISTORY_MESSAGES` to bound disk usage.
+    """
+
+    deployments: dict[DeploymentID, list[ChatMessage]] = Field(default_factory=dict)
+
+    def get(self, deployment_id: DeploymentID) -> list[ChatMessage] | None:
+        return self.deployments.get(deployment_id)
```
seedspirit, line 206: Does the message "No chat history found for deployment_id" mean only one of the following: either the deployment ID is invalid, or the deployment ID is valid but there is no chat history? Or does it mean both? If it means the former, it would be better to return an empty list rather than `None`.

src/ai/backend/client/cli/v2/deployment/chat/types.py
```diff
@@ -0,0 +1,245 @@
...
+        return messages[-limit:]
+
+    def append(
+        self,
+        deployment_id: DeploymentID,
+        role: str,
+        content: str,
+        *,
+        created_at: datetime,
+        max_persisted: int = MAX_PERSISTED_HISTORY_MESSAGES,
```
seedspirit, line 224: Wouldn't it be better to change the structure so that `max_persisted` is stored in `DeploymentChatConfig` or similar, rather than injecting or apply it during the chat append operation?

src/ai/backend/common/dto/clients/openai_compat/response.py
```diff
@@ -0,0 +1,88 @@
...
+class ChatCompletionResponseChoice(BaseModel):
+    """One entry in ``choices[]`` on a non-streaming chat-completion response.
+
+    Streaming responses use ``delta`` instead of ``message``; this model
+    intentionally requires ``message`` so a streaming chunk that slips
+    through (the SDK never sets ``stream=true`` itself) fails loudly via
+    ``ValidationError`` rather than corrupting persisted history.
+    """
+
+    model_config = ConfigDict(extra="allow")
```
seedspirit: nitpick: The setting `model_config = ConfigDict(extra="allow")` seems redundant; it might be a good idea to create a `BaseModel` class that inherits these settings

src/ai/backend/client/cli/v2/deployment/chat/types.py
jopemachine: This PR was not originally intended to become this large.
Given the current size of the PR, I won't include further enhancement work in this PR.

src/ai/backend/client/v2/base_client.py
```diff
@@ -453,3 +461,100 @@ async def typed_request(
...
+                        f"deployment endpoint returned non-JSON response "
+                        f"(status={resp.status}): {raw!r}"
+                    ) from e
+                if not isinstance(payload, dict):
+                    raise BackendClientError(
+                        f"deployment endpoint returned non-object payload "
+                        f"(type={type(payload).__name__}, status={resp.status}): {payload!r}"
+                    )
+                self._raise_for_status(resp, payload)
+                return payload
```
fregataa: this part can be handled in a separate method

src/ai/backend/client/v2/deployment_chat.py
```diff
@@ -0,0 +1,61 @@
...
+
+class DeploymentChatClient(BackendAIAppProxyClient):
+    """OpenAI Chat Completions client for direct-to-deployment inference traffic.
+
+    Sends ``POST /v1/chat/completions`` with an OpenAI-shaped
+    ``{model, messages, ...}`` JSON body. Compatible runtimes: vLLM,
+    SGLang, NVIDIA NIM, and TGI in Messages API mode. Vanilla TGI
+    (``/generate``) and arbitrary custom containers need a different
+    :class:`BackendAIAppProxyClient` subclass.
+    """
```
fregataa, line 37: is there any reason this client is not registered in client registry?

jopemachine: The V2Client registry is based on BackendAIAuthClient, which communicates with the manager.
Since DeploymentChatClient is a client that communicates with the app proxy, it seems conceptually misaligned with the registry.

fregataa (approved):

PR #11351, opened 2026-04-27
refactor(BA-5821): migrate user entity off association_groups_users

fregataa:
## Summary
- Migrate every READ and WRITE in user, auth, and gql_legacy keypair/user code paths from `association_groups_users` (agus) to `association_scopes_entities` (ASE) with `(scope=PROJECT, entity=USER)`.
- `UserProjectEntityUnbinder` (and its `UsersForProjectPurgerSpec`) retargets the framework's `BatchPurger` at the ASE table so unassign now removes the canonical membership row directly; the user creation path drops the duplicate agus INSERT since `RBACEntityCreator` already writes the ASE row via `additional_scope_refs`.
- Pre-existing unit tests that asserted on agus state are updated to assert ASE state, matching the new contract.

## Test plan
- [x] `pants fmt :: && pants fix :: && pants lint --changed-since=origin/main`
- [x] `pants test tests/unit/manager/repositories/auth:: tests/unit/manager/repositories/user:: tests/unit/manager/repositories/group::` (all green)
- [x] New component tests at `tests/component/user_membership/` exercise `user.search_by_project` across the four BA-5811 scenarios (member finds self, non-member 403, unassign loses access, cross-project isolation) for both the `project.assign_users` and `rbac.assign_role` SDK paths — 8/8 pass.

Resolves BA-5821


tests/unit/manager/repositories/user/test_user_repository.py
```diff
@@ -659,15 +659,6 @@ async def test_create_user_validated_auto_includes_model_store_project(
...
             assert model_store_assoc is not None
 
-            # Also verify user was added to the model store group
-            group_assoc = await session.scalar(
-                sa.select(AssocGroupUserRow).where(
-                    AssocGroupUserRow.user_id == result.user.uuid,
-                    AssocGroupUserRow.group_id == model_store_project_id,
-                )
-            )
-            assert group_assoc is not None
```
seedspirit, line 669: It might be a good idea to include logic that validates this using an AssociationScopesEntitiesRow scalar query. I think we should ensure that the items we’ve been verifying through testing continue to be verified.

jopemachine (approved):

PR #11360, opened 2026-04-28
feat(BA-5878): re-supply legacy live_stat stats.{max,avg,rate} from Prometheus

seedspirit:
## Why

The legacy `live_stat` resolvers — `mem_max_bytes`, `cpu_using`, `io_max_scratch_size`, `net_rx_bytes`, `net_tx_bytes`, and friends — read `stats.max` / `stats.avg` / `stats.rate` keys produced by an in-memory `MovingStatistics` accumulator on the **agent**. Each running container's metric carries a `MovingStatistics` instance holding running per-sample max / sum / count over every sample observed (sampled every **5 s** via `UTILIZATION_METRIC_INTERVAL = 5.0`), plus the latest 2 samples for rate calculation. Each metric declares a `stats_filter` whitelist controlling which slots get published.

That accumulator has serious failure modes:

- **Agent restart wipes the history.** First post-restart sample becomes the new baseline, so `stats.max` silently rewinds to "max since the agent process happened to last restart" — not lifetime peak.
- **`avg` flattens forever.** `_count` is unbounded; a session that ran hot then went idle drifts toward the idle reading.
- **No retention** across host reboot, **no consistency** across sessions whose agents restarted at different times.
- `stats.rate` is computed from the last two samples only, so the first post-restart reading is `0`.
- For `net_rx` / `net_tx` the agent's `stats_filter` is empty, so the agent never actually publishes `stats.rate` — the legacy GraphQL `net_rx_bytes` / `net_tx_bytes` resolvers always return `0`.

## What this PR provides

A **parallel manager-side supply** for the same `stats.{max,avg,rate}` slots, computed on demand from Prometheus via window queries (`MetricConfig.timewindow`, default **5m**). The legacy agent producer stays in place — this is a drop-in alternative source that future legacy compat code can route to.

The conceptual switch is **"lifetime peak as remembered by the agent"** → **"recent peak as recorded by Prometheus."** The new supply:

- Survives agent / manager / host restart (data lives in Prometheus TSDB)
- Is consistent across sessions (wall-clock window, not per-agent uptime)
- Doesn't accumulate forever (old peaks roll out of the window)
- Closes the long-standing `net_rx` / `net_tx` `stats.rate` hole that the agent's empty `stats_filter` left

The query results carry clean `value_type=max` / `value_type=avg` / `value_type=rate` labels (without the legacy `stats.` prefix). A future `LegacyLiveStatConverter` will re-attach the `stats.` prefix when serving legacy clients; the manager pipeline itself doesn't need to carry the legacy encoding.

## Coverage matrix (what each pipeline publishes)

| Metric | Legacy `stats.max` | This PR `stats.max` | Legacy `stats.avg` | This PR `stats.avg` | Legacy `stats.rate` | This PR `stats.rate` |
|---|---|---|---|---|---|---|
| `cpu_util` | ✓ | ✓ | ✓ | ✓ | — | — |
| `mem` | ✓ | ✓ | — | — | — | — |
| `io_scratch_size` | ✓ | ✓ | — | — | — | — |
| `io_read` | — | — | — | — | ✓ | ✓ |
| `io_write` | — | — | — | — | ✓ | ✓ |
| `net_rx` | — | — | — | — | — *(see note)* | ✓ |
| `net_tx` | — | — | — | — | — *(see note)* | ✓ |
| accel `*_mem` | ✓ | ✓ | — | — | — | — |
| accel `*_util` | ✓ | ✓ | ✓ | ✓ | — | — |
| accel `*_power` | ✓ | ✓ | ✓ | ✓ | — | — |
| accel `*_temperature` | ✓ | ✓ | ✓ | ✓ | — | — |

Notes:

- **Legacy `stats.rate` for `net_rx` / `net_tx` is broken — always returns `0`.** The agent's measurement entries for these omit `stats_filter` (default empty), so the agent never publishes `stats.rate` to Valkey for net_rx/net_tx. The legacy GraphQL `net_rx_bytes` / `net_tx_bytes` resolvers read the missing key and fall back to `0`. The agent's per-second rate IS published, but as `current` (via `current_hook = lambda m: m.stats.rate`), which the legacy resolver doesn't read — an internal contract mismatch.
- **This PR's `stats.rate` for `net_rx` / `net_tx` — passthrough, no `rate()` wrap.** The `current` series in Prometheus already carries per-second rate (thanks to the agent's `current_hook`), so the new query just sums across replicas and `label_replace`s the `value_type` label from `current` to `rate`. Applying PromQL `rate()` to an already-rate series would compute rate-of-rate and produce nonsense.
- **This PR's `stats.rate` for `io_read` / `io_write` — `rate(...[5m])` over the counter.** No `current_hook` here, so `current` is the cumulative byte counter. The new query applies PromQL `rate(...[5m])` to convert it to per-second bytes, then `label_replace`s the result to `rate`. Conceptually matches what the legacy `stats_filter={"rate"}` slot held, except computed over a 5-minute window instead of the agent's last-2-sample delta.

## Live verification

Manager restarted on this branch, halfstack with Prometheus 3.1.0, one idle running kernel (1 CPU / 1 GiB):

| metric | current | stats.max | stats.avg | stats.rate |
|---|---|---|---|---|
| `mem` | 133,910,528 | **133,910,528** | — *(not produced by agent)* | — |
| `cpu_util` | 5371.729 | **5371.729** | **5371.729** | — |
| `io_scratch_size` | 0 | **0** *(empty disk, not a missing field)* | — | — |
| `net_rx` | 27530 | — | — | **27530** |
| `net_tx` | 30378 | — | — | **30378** |
| `io_read` | 0 | — | — | **0** |
| `io_write` | 0 | — | — | **0** |

Idle workload across the 5-minute window → `current` / `stats.max` / `stats.avg` converge. Same kernel via the legacy GraphQL resolver returned `live_stat: null` (Valkey path empty in this env), confirming all values above came from the new pipeline.

## Bug fixed during verification

Prometheus's RE2 engine rejects `\-` as `parse error: unknown escape sequence U+002D '-'`. Python's `re.escape` over-escapes hyphens, so every kernel-live-stat query (UUIDs always contain hyphens) would have failed to parse:

```
kernel_id=~"…-…"   → status: success
kernel_id=~"…\-…"  → status: error
```

Fixed by stripping the redundant escape before sending.

## Test plan

- [x] `pants test tests/unit/manager/services/utilization_metric/test_container_metric.py`
- [x] `pants fmt fix lint check --changed-since=HEAD~1`
- [x] Live verification against Prometheus on a running kernel — `stats.max` / `stats.avg` / `stats.rate` populated as expected for `mem` / `cpu_util` / `io_scratch_size` / `net_rx` / `net_tx` / `io_read` / `io_write`. Wiring the new pipeline behind the legacy GraphQL resolvers (the `LegacyLiveStatConverter`) is the follow-up.

### DIFF
--- a/src/ai/backend/common/clients/prometheus/types.py
+++ b/src/ai/backend/common/clients/prometheus/types.py
@@ -8,6 +8,9 @@ class ValueType(StrEnum):
     CURRENT = "current"
     CAPACITY = "capacity"
     PCT = "pct"
+    MAX = "max"
+    AVG = "avg"
+    RATE = "rate"
 
 
 @dataclass(frozen=True)
--- a/changes/11360.feature.md
+++ b/changes/11360.feature.md
@@ -1,0 +1,1 @@
+Provide a manager-side parallel supply for legacy `live_stat` `stats.max` / `stats.avg` / `stats.rate` fields, computed from Prometheus on demand instead of from the agent's in-memory `MovingStatistics` accumulator. Survives agent / manager / host restart, stays consistent across sessions, and uses a sliding window (default 5m) instead of unbounded lifetime accumulation.
--- a/src/ai/backend/common/clients/prometheus/client.py
+++ b/src/ai/backend/common/clients/prometheus/client.py
@@ -89,7 +89,7 @@ class PrometheusClient:
             response = await self._query_instant(preset)
             merged = merged.merged_with(
                 KernelMetricValuesByKernel.from_prometheus_response(response)
-            )        
+            )
         return KernelLiveStatValues.with_capacity_sentinels(merged.values_by_kernel)
 
     async def execute_preset(
--- a/src/ai/backend/common/clients/prometheus/client.py
+++ b/src/ai/backend/common/clients/prometheus/client.py
@@ -84,13 +84,12 @@ class PrometheusClient:
         kernel_ids: Sequence[KernelId],
     ) -> KernelLiveStatValues:
         queries = self._fixed_query_builder.get_container_live_stat_queries(kernel_ids)
-        gauge_response = await self._query_instant(queries.gauge)
-        diff_response = await self._query_instant(queries.diff)
-        rate_response = await self._query_instant(queries.rate)
-        gauge = KernelMetricValuesByKernel.from_prometheus_response(gauge_response)
-        diff = KernelMetricValuesByKernel.from_prometheus_response(diff_response)
-        rate = KernelMetricValuesByKernel.from_prometheus_response(rate_response)
-        merged = gauge.merged_with(diff).merged_with(rate)
+        merged = KernelMetricValuesByKernel(values_by_kernel={})
+        for preset in queries.to_list():
+            response = await self._query_instant(preset)
+            merged = merged.merged_with(
+                KernelMetricValuesByKernel.from_prometheus_response(response)
+            )        
         return KernelLiveStatValues.with_capacity_sentinels(merged.values_by_kernel)
 
     async def execute_preset(
--- a/src/ai/backend/common/clients/prometheus/fixed_query_builder.py
+++ b/src/ai/backend/common/clients/prometheus/fixed_query_builder.py
@@ -6,6 +6,14 @@ from typing import Final
 from ai.backend.common.clients.prometheus.metric_types import (
     DIFF_METRICS,
     RATE_METRICS,
+    STATS_AVG_GAUGE_METRIC_PATTERNS,
+    STATS_AVG_GAUGE_METRICS,
+    STATS_AVG_OVER_RATE_METRICS,
+    STATS_MAX_GAUGE_METRIC_PATTERNS,
+    STATS_MAX_GAUGE_METRICS,
+    STATS_MAX_OVER_RATE_METRICS,
+    STATS_RATE_COUNTER_METRICS,
+    STATS_RATE_GAUGE_METRICS,
     ContainerLiveStatQueries,
     ContainerMetricOptionalLabel,
     MetricType,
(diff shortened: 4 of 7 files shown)


src/ai/backend/agent/stats.py
```diff
@@ -773,11 +750,6 @@ async def collect_container_stat(
                         kernel_id = kernel_id_map[ContainerId(cid)]
                     except KeyError:
                         continue
-                    self.observe_container_metric(
-                        kernel_id,
-                        metric_key,
-                        measure,
-                    )
```
fregataa, line 780: Is this duplicate??

src/ai/backend/common/clients/prometheus/types.py
```diff
@@ -8,6 +8,22 @@ class ValueType(StrEnum):
     CURRENT = "current"
     CAPACITY = "capacity"
     PCT = "pct"
+    MAX = "max"
+    AVG = "avg"
+    RATE = "rate"
```
jopemachine, line 13: Just a question, this seems PR is adding max, and avg
Why rate is included here?

seedspirit: I added this to maintain backward compatibility with the `stat.rate` legacy feature, but since it’s unrelated to this PR, I’ll remove it.

src/ai/backend/agent/stats.py
seedspirit: Yes, that is duplicate code. However, the direction of the current PR has shifted from having the new agent emit lifetime value as a new metric to Prometheus, to querying existing metrics using functions like `max over time`; therefore, we have decided not to make this code change.

HyeockJinKim (approved):

PR #11398, opened 2026-04-28
fix(BA-5893): replace native sessionresult enum with VARCHAR + StrEnumType

fregataa:
## Summary
- Convert `kernels.result` and `sessions.result` from PostgreSQL native enum types (`sessionresult` / `sessionresults`) to `VARCHAR(64)`, with the SQLAlchemy ORM using `StrEnumType(SessionResult, use_name=True)`. Stored values are unchanged (uppercase enum names: `UNDEFINED`, `SUCCESS`, `FAILURE`).
- Add new alembic revision `dec0deba5893` that is idempotent across all observed diverged enum-type states (only singular, only plural, both, neither) and drops both enum types after migrating columns.
- Wrap every `sessionresult` / `sessionresults` reference in existing alembic revisions (`405aa2c39458`, `b6b884fbae1f`, `ffcf0ed13a26`, `0b1efbb2db84`, `1cc9b47e0a8e`) with `try/except` so `alembic upgrade` no longer fails when a machine's enum-type history diverges from the migration chain.

## Test plan
- [x] `pants fmt :: && pants fix :: && pants lint --changed-since=origin/main && pants check --changed-since=origin/main`
- [x] Local DB migration round-trip verified: starting from existing data with `UNDEFINED`/`SUCCESS`/`FAILURE` rows, `alembic upgrade head` converts both columns to `VARCHAR(64)`, drops both enum types, and preserves all values; `alembic downgrade -1` recreates the singular enum type and restores both columns; re-running `alembic upgrade head` is a no-op for already-converted columns.
- [x] Verified column metadata after upgrade (`udt_name = varchar`, default `'UNDEFINED'`, `ix_kernels_result` / `ix_sessions_result` indexes preserved).

Resolves BA-5893


HyeockJinKim (approved):
Please take a look at the session creation and session details page and let me know what you think.

src/ai/backend/manager/models/alembic/versions/1cc9b47e0a8e_fix_sessionresult_enum_type_coexistence_backport.py
```diff
@@ -15,48 +18,100 @@
 branch_labels = None
 depends_on = None
 
+log = logging.getLogger(__name__)
+
+# PostgreSQL SQLSTATE codes accepted as "expected divergence" when guarding
+# sessionresult-related DDL with try/except. Anything else is reraised.
```
jopemachine, line 24: Nit: If there is a link where this information can be verified, it would be good to include it as a comment.

PR #11400, opened 2026-04-28
fix(BA-5894): bypass legacy permission check in vfolder REST middleware for UUID-based access

fregataa:
## Summary
- The legacy `_vfolder_resolver` middleware calls `get_accessible_vfolder` for every vfolder REST endpoint, which uses the pre-RBAC permission resolver. That resolver only recognizes domain-admin/superadmin special cases plus the static group default permission column, so a project admin granted update permission via the new RBAC association graph receives `VFolderNotFound` before the downstream RBAC validator on `update_vfolder_attribute` can run.
- Branch the resolver behavior on the path parameter type. UUID path: fetch the row via a new `get_vfolder_row` processor (no permission filter, no status check) — permission evaluation is delegated to the downstream RBAC validator on the action. Name path: keep the existing `get_accessible_vfolder` flow, since name-based resolution requires user context to disambiguate folders sharing a name.
- Handler signatures are unchanged. The injected `request["vfolder_row"]` mapping shape matches what `query_accessible_vfolders` returns.

## Test plan
- [x] As a project admin, `POST /folders/<UUID>/update-options` with a project-owned vfolder updates `cloneable` / `mount_permission` successfully (previously returned `VFolderNotFound`).
- [x] As a regular user without RBAC update permission, the same call is rejected by the RBAC validator (`NotEnoughPermission`).
- [x] Name-based access (`POST /folders/<NAME>/update-options`) preserves legacy behavior — a non-owner still receives `VFolderNotFound`.

Resolves BA-5894

### DIFF
--- a/src/ai/backend/manager/repositories/vfolder/repository.py
+++ b/src/ai/backend/manager/repositories/vfolder/repository.py
@@ -205,7 +205,7 @@ class VfolderRepository:
             return self._vfolder_row_to_data(vfolder_row)
 
     @vfolder_repository_resilience.apply()
-    async def get_row_by_id(self, vfolder_id: uuid.UUID) -> Mapping[str, Any]:
+    async def get_row_by_id(self, vfolder_id: VFolderUUID) -> Mapping[str, Any]:
         """
         Fetch a vfolder row as a plain mapping by UUID, without permission filtering.
 
--- a/src/ai/backend/manager/repositories/vfolder/repository.py
+++ b/src/ai/backend/manager/repositories/vfolder/repository.py
@@ -205,6 +205,43 @@ class VfolderRepository:
             return self._vfolder_row_to_data(vfolder_row)
 
     @vfolder_repository_resilience.apply()
+    async def get_row_by_id(self, vfolder_id: uuid.UUID) -> Mapping[str, Any]:
+        """
+        Fetch a vfolder row as a plain mapping by UUID, without permission filtering.
+
+        For use only by the REST middleware; do not call from new code.
+        """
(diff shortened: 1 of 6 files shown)


src/ai/backend/manager/repositories/vfolder/repository.py
```diff
@@ -204,6 +204,43 @@ async def get_by_id(self, vfolder_id: uuid.UUID) -> VFolderData:
...
+                "creator_id": vfolder_row.creator_id,
+                "user_email": None,
+                "group_name": None,
+                "is_owner": False,
+                "permission": vfolder_row.permission,
+                "unmanaged_path": vfolder_row.unmanaged_path,
+                "cloneable": vfolder_row.cloneable,
+                "status": vfolder_row.status,
+                "cur_size": vfolder_row.cur_size,
+            }
```
jopemachine, line 242: Why are we not returning a dataclass type?

fregataa: because it should be used in a legacy REST middleware, which requires mapping type.
Check this reference
https://github.com/lablup/backend.ai/blob/main/src/ai/backend/manager/api/rest/vfolder/registry.py
https://github.com/lablup/backend.ai/blob/main/src/ai/backend/manager/dto/context.py

jopemachine: In that case, would it make sense to reflect the legacy nature in the action name, such as GetLegacyVFolderAction?
Alternatively, we could use a different name like GetVFolderMappingAction to distinguish it.

jopemachine (approved):

PR #11402, opened 2026-04-29
fix(BA-5891): allow optional start_command in model service config

seedspirit:
## Summary
- Make `start_command` optional in model service configs and fall back to the container image's `Config.Cmd` when omitted.
- Update trafaret schema, Pydantic config models, v2 DTOs, GraphQL types, and `model-definition.schema.json` to allow `null` / missing `start_command`.
- Resolve the missing command in `AbstractAgent` at kernel-creation time via `extract_image_command()`; log a warning and skip the model service when neither source provides a command.

## Test plan
- [x] `pants fmt / fix / lint --changed-since=origin/main`
- [x] Unit tests: `tests/unit/agent/test_model_service_start_command.py`, updated `tests/unit/common/test_config.py`
- [ ] Manual: deploy a model with no `start_command` and verify the agent uses the image's default Cmd
- [ ] Manual: deploy with an image that has no `Config.Cmd` and verify a warning is logged and the service is skipped

Resolves BA-5891

----
📚 Documentation preview 📚: https://sorna--11402.org.readthedocs.build/en/11402/

----
📚 Documentation preview 📚: https://sorna-ko--11402.org.readthedocs.build/ko/11402/

### DIFF
--- a/src/ai/backend/agent/docker/agent.py
+++ b/src/ai/backend/agent/docker/agent.py
@@ -1662,8 +1662,9 @@ class DockerAgent(AbstractAgent[DockerKernel, DockerKernelCreationContext]):
                 return None
             if isinstance(command, str):
                 return [command]
-            if isinstance(command, Sequence):
-                return cast(list[str], list(command))
+
+            if isinstance(command, list):
+                return cast(list[str], command)
             return None
 
     @override
--- a/src/ai/backend/agent/docker/agent.py
+++ b/src/ai/backend/agent/docker/agent.py
@@ -1654,10 +1654,10 @@ class DockerAgent(AbstractAgent[DockerKernel, DockerKernelCreationContext]):
         return cast(str, self.docker_info["CgroupVersion"])
 
     @override
-    async def extract_image_command(self, image: str) -> str | None:
+    async def extract_image_command(self, image: str) -> str | list[str] | None:
         async with closing_async(Docker()) as docker:
             result = await docker.images.get(image)
-            return cast(str | None, result["Config"].get("Cmd"))
+            return cast(str | list[str] | None, result["Config"].get("Cmd"))
 
     @override
     async def enumerate_containers(
--- a/src/ai/backend/agent/docker/agent.py
+++ b/src/ai/backend/agent/docker/agent.py
@@ -1654,10 +1654,17 @@ class DockerAgent(AbstractAgent[DockerKernel, DockerKernelCreationContext]):
         return cast(str, self.docker_info["CgroupVersion"])
 
     @override
-    async def extract_image_command(self, image: str) -> str | list[str] | None:
+    async def extract_image_command(self, image: str) -> list[str] | None:
         async with closing_async(Docker()) as docker:
             result = await docker.images.get(image)
-            return cast(str | list[str] | None, result["Config"].get("Cmd"))
+            command = result["Config"].get("Cmd")
+            if command is None:
+                return None
+            if isinstance(command, str):
+                return [command]
+            if isinstance(command, Sequence):
+                return cast(list[str], list(command))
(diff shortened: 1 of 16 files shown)


src/ai/backend/agent/docker/agent.py
```diff
@@ -1654,10 +1654,10 @@ def get_cgroup_version(self) -> str:
         return cast(str, self.docker_info["CgroupVersion"])
 
     @override
-    async def extract_image_command(self, image: str) -> str | None:
+    async def extract_image_command(self, image: str) -> str | list[str] | None:
         async with closing_async(Docker()) as docker:
             result = await docker.images.get(image)
-            return cast(str | None, result["Config"].get("Cmd"))
+            return cast(str | list[str] | None, result["Config"].get("Cmd"))
```
jopemachine: Why was this changed to allow list[str]?

seedspirit: Docker normalizes both exec-form and shell-form `CMD` into a JSON array. So `Config.Cmd` is almost always `list[str] | None`. The previous `str | None` annotation was therefore inaccurate. But I remained `str` for legacy compatibility

jopemachine: I asked because this seems unrelated to the main changes in this PR.

src/ai/backend/common/dto/manager/v2/deployment/types.py
```diff
@@ -293,15 +297,28 @@ class ModelServiceConfigInfoDTO(BaseResponseModel):
...
+        description="Shell used to wrap a string-form start_command at parse time.",
     )
     port: int = Field(description="Port number for the model service.")
     health_check: ModelHealthCheckInfoDTO | None = Field(
         default=None, description="Health check configuration for the model service."
     )
 
+    @model_validator(mode="before")
+    @classmethod
+    def _wrap_string_start_command(cls, data: Any) -> Any:
```
jopemachine: It doesn't seem appropriate for conversion logic to live in the DTO. 
How about moving it to the adapter layer or finding another approach?

src/ai/backend/agent/dummy/agent.py
```diff
@@ -308,10 +308,10 @@ def get_cgroup_version(self) -> str:
         return ""
 
     @override
-    async def extract_image_command(self, image: str) -> str | None:
+    async def extract_image_command(self, image: str) -> str | list[str] | None:
```
HyeockJinKim: Why are these parts written as `str | list[str] | None`? Is there a reason to treat `str` separately?

src/ai/backend/common/dto/manager/v2/deployment/types.py
seedspirit: Since we're also using this DTO for validating request inputs, it seems a bit ambiguous, but I'll try to figure out a solution.

src/ai/backend/agent/docker/agent.py
```diff
@@ -1654,10 +1654,17 @@ def get_cgroup_version(self) -> str:
...
         async with closing_async(Docker()) as docker:
             result = await docker.images.get(image)
-            return cast(str | None, result["Config"].get("Cmd"))
+            command = result["Config"].get("Cmd")
+            if command is None:
+                return None
+            if isinstance(command, str):
+                return [command]
+            if isinstance(command, Sequence):
+                return cast(list[str], list(command))
```
fregataa: It is not appropriate to check by seqeunce type because `list()` accepts iterable inputs, not sequence
https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

HyeockJinKim (approved):

fregataa (approved):

PR #11412, opened 2026-04-29
feat(BA-5903): persist deployment chat history and replay as request context

jopemachine:
## 📚 Stacked PRs

1. ⬇️ #11344 — `feat(BA-5528): add deployment chat CLI for vLLM-backed model services`
2. 👉 **#11412** — `feat(BA-5903): persist deployment chat history and replay as request context` ← you are here

## Summary

Builds on top of [BA-5528](https://github.com/lablup/backend.ai/pull/11344) to add **client-side chat history** so multi-turn conversations against an OpenAI-compatible deployment endpoint keep their state across `./bai deployment chat` invocations. Also tightens the SDK boundary by typing the chat-completion response as a Pydantic model.

Resolves [BA-5903](https://lablup.atlassian.net/browse/BA-5903).

## What's added

- `~/.backend.ai/deployment_chat/history.json` — per-deployment rolling transcript, isolated from the existing `cache.json` (auto endpoint metadata) and `config.json` (user token/model).
- `chat` command auto-injects the most recent K user/assistant turns into the request `messages[]` array, then records the round on success.
  - `--history-limit N` (default 10) controls the context window. `0` skips context for the turn but still records the round.
  - Persisted history per deployment is FIFO-capped at 100 messages so disk usage stays bounded.
  - Half-recorded rounds are never written: if assistant `content` is missing (tool-call only / empty `choices`), the user side is dropped too.
- New `chat-history show / clear` subcommands for inspection and reset.
- `DeploymentChatClient.chat_completion()` now returns a Pydantic `ChatCompletionResponse` with strict validation on `choices[0].message.content` (the path the CLI consumes) and `extra="allow"` everywhere else, so vLLM `prompt_logprobs` / `usage` / `system_fingerprint` / NIM extras still survive the JSON pretty-print.

## Why prefix-cache-friendly

vLLM/SGLang/NIM all advertise OpenAI-compat `/v1/chat/completions`, which is stateless by spec — the client must re-send full message history each turn. Prefix caching on the inference server reuses the KV cache for the unchanged prefix, so the GPU cost only grows with the new tokens, not the whole conversation. Storing history client-side and replaying it is the conventional pattern.

## Test plan

- [x] `pants fmt/lint/check` clean
- [x] Unit tests:
  - History storage round-trip and FIFO truncation (`test_deployment_chat_types.py`, `test_deployment_chat_utils.py`)
  - Loader resilience against corrupt / schema-mismatched `history.json`
  - `ChatCompletionResponse.assistant_message` extraction across OpenAI-compat edge cases (empty `choices`, tool-call only, streaming `delta` shape) (`test_deployment_chat_client.py`)
- [ ] Smoke test against a live vLLM deployment: two-turn conversation where the second turn references the first.

[BA-5528]: https://lablup.atlassian.net/browse/BA-5528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BA-5903]: https://lablup.atlassian.net/browse/BA-5903?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ai/backend/client/cli/v2/deployment/chat/types.py
+++ b/src/ai/backend/client/cli/v2/deployment/chat/types.py
@@ -12,6 +12,7 @@ from pydantic import BaseModel, Field, ValidationError
 from ai.backend.client.cli.v2.deployment.chat.utils import (
     CHAT_CACHE_FILE,
     CHAT_CONFIG_FILE,
+    CHAT_HISTORY_FILE,
     read_json_file,
     write_json_file,
 )
@@ -19,6 +20,20 @@ from ai.backend.common.identifier.deployment import DeploymentID
 
 CACHE_ENTRY_TTL = timedelta(hours=24)
 """Endpoint cache entries older than this are treated as a cache miss."""
+
+DEFAULT_CHAT_HISTORY_LIMIT = 10
+"""Default number of past messages forwarded as context on each ``chat`` call.
+
+Mirrors the typical 5-turn rolling window of OpenAI-compatible chat UIs.
+Override per-call with ``--history-limit``; setting it to 0 disables context.
+"""
+
+MAX_PERSISTED_HISTORY_MESSAGES = 100
+"""Hard cap on messages kept in ``history.json`` per deployment.
+
+The file holds plain text; capping it keeps disk usage bounded even when the
+user never runs ``chat-history clear``. Older messages are dropped FIFO.
+"""
 
 
 class DeploymentChatCacheEntry(BaseModel):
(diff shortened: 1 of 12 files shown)


src/ai/backend/common/data/deployment_chat/types.py
```diff
@@ -185,3 +199,93 @@ def save(self) -> None:
...
+
+    deployments: defaultdict[
+        UUID,
+        Annotated[list[ChatHistoryMessage], Field(default_factory=list)],
+    ] = Field(default_factory=lambda: defaultdict(list))
+
+    def get(self, deployment_id: UUID) -> list[ChatHistoryMessage] | None:
+        return self.deployments.get(deployment_id)
+
+    def recent(self, deployment_id: UUID, limit: int) -> list[ChatHistoryMessage]:
```
jopemachine: Consider to use more proper name

jopemachine:
Add news fragment

src/ai/backend/common/data/deployment_chat/types.py
jopemachine: Renamed to `tail` in 111989546 — matches the docstring ("Return the tail of the transcript") and the pandas/Unix idiom for "last N items".

src/ai/backend/common/data/deployment_chat/types.py
```diff
@@ -105,3 +119,73 @@ def pop_model(self, deployment_id: UUID) -> bool:
...
+    stripped before the message is replayed into the chat-completions request
+    body (the wire format is just ``{role, content}``).
+    """
+
+    role: str
+    content: str
+    created_at: datetime
+
+
+def _new_history_store() -> defaultdict[UUID, list[ChatHistoryMessage]]:
```
jopemachine: Why is it required?

src/ai/backend/common/data/deployment_chat/types.py
```diff
@@ -105,3 +119,73 @@ def pop_model(self, deployment_id: UUID) -> bool:
 
     def pop(self, deployment_id: UUID) -> bool:
         return self.deployments.pop(deployment_id, None) is not None
+
+
+class ChatHistoryMessage(BaseModel):
```
jopemachine: Should it be ChatMessageHistory?

src/ai/backend/common/data/deployment_chat/types.py
jopemachine: slice

src/ai/backend/common/dto/clients/openai_compat/response.py
```diff
@@ -26,3 +26,63 @@ class ListModelsResponse(BaseModel):
...
+    through via ``extra="allow"`` so they survive the round-trip back
+    to the user's stdout when the CLI pretty-prints the response.
+    """
+
+    model_config = ConfigDict(extra="allow")
+
+    choices: list[ChatCompletionResponseChoice]
+
+    @property
+    def assistant_content(self) -> str | None:
```
jopemachine: assistant_message

src/ai/backend/common/data/deployment_chat/types.py
jopemachine: Switched the store from `defaultdict[UUID, list[ChatMessage]]` (with the typed factory) to plain `dict[UUID, list[ChatMessage]]` and use `dict.setdefault()` in `append()` (commit 611e4b4e8). `append` was the only spot that benefited from the auto-create behavior, so `setdefault` is just as short while letting mypy infer types from the plain annotation — no factory, no `Annotated` wrapper, no `Never`-vs-typed workaround.

src/ai/backend/common/data/deployment_chat/types.py
jopemachine: Renamed to `ChatMessage` in 611e4b4e8. The class represents a single chat turn, so a name like `ChatMessageHistory` would fit better on a collection (which is what `DeploymentChatHistory` already is). `ChatMessage` reads cleanly on its own and the surrounding `DeploymentChatHistory` carries the "history" framing.

src/ai/backend/common/dto/clients/openai_compat/response.py
jopemachine: Renamed `ChatCompletionResponse.assistant_content` to `assistant_message` in 72d839bd1. Reads as "the assistant's message" rather than "content of [what?]"; CLI caller and tests updated.

src/ai/backend/client/cli/v2/deployment/chat/types.py
```diff
@@ -150,3 +165,80 @@ def save(self) -> None:
...
+    the others. The transcripts are FIFO-truncated at
+    :data:`MAX_PERSISTED_HISTORY_MESSAGES` to bound disk usage.
+    """
+
+    deployments: dict[UUID, list[ChatMessage]] = Field(default_factory=dict)
+
+    def get(self, deployment_id: UUID) -> list[ChatMessage] | None:
+        return self.deployments.get(deployment_id)
+
+    def tail(self, deployment_id: UUID, limit: int) -> list[ChatMessage]:
```
jopemachine: slice

jopemachine: Renamed `tail` -> `slice` in 287f830a2. Method/test renames + caller updated.

src/ai/backend/client/cli/v2/deployment/chat/types.py
```diff
@@ -150,3 +165,80 @@ def save(self) -> None:
...
+class DeploymentChatHistory(BaseModel):
+    """Per-deployment rolling chat transcripts.
+
+    Stored separately from the cache (auto-managed endpoint metadata) and
+    config (user-managed token/model) so that clearing one does not affect
+    the others. The transcripts are FIFO-truncated at
+    :data:`MAX_PERSISTED_HISTORY_MESSAGES` to bound disk usage.
+    """
+
+    deployments: dict[UUID, list[ChatMessage]] = Field(default_factory=dict)
```
fregataa: nit: ai.backend.common.identifier.deployment.DeploymentID is good for this case
Same goes to signatures of other methods

jopemachine: Switched to `DeploymentID` across the chat data classes (Cache/Config/History), the formatter, and the click handler signatures in 4c38ac1d2 (#11344) and 696b49302 (#11412). The click `type=click.UUID` parser still emits a plain `UUID` at runtime; since `DeploymentID = NewType("DeploymentID", UUID)` is structurally identical, no explicit wrap is needed at the boundary.

fregataa (approved):
Just an idea. How about saving a message to history first and send a chat request?

PR #7880, opened 2026-01-08
feat(BA-3802): Migrate to SQLAlchemy 2.0 with comprehensive type safety improvements

HyeockJinKim:
- Updated various files to enhance code formatting, including consistent use of inline conditionals and improved line breaks for better readability.
- Replaced `timezone.utc` with `UTC` for consistency in datetime handling across multiple files.
- Simplified query construction in several repository files for clarity.
- Adjusted import statements to remove redundancies and improve organization.
- Enhanced logging messages for better clarity in error handling.
- Improved type hinting and annotations for better code comprehension.

resolves #NNN (BA-MMM)

**Checklist:** (if applicable)

- [ ] Milestone metadata specifying the target backport version
- [ ] Mention to the original issue
- [ ] Installer updates including:
  - Fixtures for db schema changes
  - New mandatory config options
- [ ] Update of end-to-end CLI integration tests in `ai.backend.test`
- [ ] API server-client counterparts (e.g., manager API -> client SDK)
- [ ] Test case(s) to:
  - Demonstrate the difference of before/after
  - Demonstrate the flow of abstract/conceptual models with a concrete implementation
- [ ] Documentation
  - Contents in the `docs` directory
  - docstrings in public interfaces and type annotations

----
📚 Documentation preview 📚: https://sorna--7880.org.readthedocs.build/en/7880/

----
📚 Documentation preview 📚: https://sorna-ko--7880.org.readthedocs.build/ko/7880/

PR #11421, opened 2026-04-29
fix(BA-5910): Fix session creation failure due to `callback_url` type error

### LINKED ISSUE #11420: POST /session 500: `callback_url` bound to asyncpg as `yarl.URL` instead of `str` (manager 26.4)
## Summary

When a session-create request includes a `callback_url`, the manager fails to
persist the kernel row because `callback_url` reaches asyncpg as a `yarl.URL`
object instead of a `str`. The request returns `500 Internal Server Error` and
no session is enqueued.

This is data-independent: any caller that supplies a `callback_url` will hit
this on manager 26.4.

## Environment

- Manager: `bai-26.4-manager` (static-python venv, Python 3.13)
- DB driver: asyncpg via SQLAlchemy `postgresql+asyncpg`
- Endpoint: `POST /session` (GET_OR_CREATE)

## Reproduction

1. `POST /session` with any payload that includes a `callback_url`, e.g.
   ```json
   {
     "callback_url": "http://example.invalid/webhook?token=<opaque>"
   }
   ```
1. Manager returns `500 Internal server error`.
1. Manager log shows the traceback below; nothing is enqueued.

The session in the captured incident was a `BATCH` job, `cluster_mode=single-node`,
single kernel — but the failure is unrelated to those fields.

## Root cause

The kernel `INSERT` binds parameter `$46` (the `callback_url` column, typed
`VARCHAR`) with a `yarl.URL(...)` instance. asyncpg's text codec only accepts
`str`:

```
File "asyncpg/pgproto/codecs/text.pyx", line 12, in as_pg_string_and_size
TypeError: expected str, got URL
```

surfaced as:

```
asyncpg.exceptions.DataError: invalid input for query argument $46:
URL('http://example.invalid/webhook?token=...') (expected str, got URL)
```

Mapping placeholders in the failing statement (four hard-coded `NULL`s for
`terminated_at`, `starts_at`, `status_info`, `status_data` shift the indexes),
`$46` is the `callback_url` column. Statement skeleton:

```
INSERT INTO kernels (id, session_id, ..., status_history, callback_url,
  startup_command, ...)
VALUES ($1::UUID, $2::UUID, ..., $45::JSONB, $46::VARCHAR, $47::VARCHAR, ...)
```

### Call path

```
api.rest.middleware.exception                          (POST /session → 500)
└─ services/session/service.py:525     create_from_params
   └─ registry.py:552 / 996 / 966      create_session
                                       → enqueue_session
                                       → _enqueue_session_via_sokovan
      └─ sokovan/scheduling_controller/scheduling_controller.py:270
                                       enqueue_session
         └─ repositories/scheduler/repository.py:241
            └─ repositories/scheduler/db_source/db_source.py:1412
                                       enqueue_session
               └─ repositories/base/rbac/entity_creator.py:186
                                       execute_rbac_bulk_entity_creator
                  └─ db_sess.flush()   ← KernelRow.callback_url = URL(...)
```

Somewhere on the create-or-get path the raw `callback_url` string is normalized
through `yarl.URL(...)` (likely for parsing/validation) and the `URL` instance
is then assigned directly to `KernelRow.callback_url` without being stringified
back. The `kernels.callback_url` column is plain `sa.String` / `VARCHAR` with
no `TypeDecorator`, so SQLAlchemy hands the object straight to asyncpg.

## Expected behaviour

`callback_url` is persisted as the URL's string form (or `NULL` if not
provided), and the session is enqueued.

## Suggested fix

Two complementary changes:

1. **Boundary coercion (minimal fix):** in
   `services/session/service.py::create_from_params` (or the session-creation
   params model that holds `callback_url`), store `str(url) if url else None`
   so the value handed to `KernelRow` / `SessionRow` is always `str | None`.
1. **Defensive (preferred, in addition):** give `kernels.callback_url` (and
   the matching column on `SessionRow`) a SQLAlchemy `TypeDecorator` that does
   `str(value)` in `process_bind_param`. This prevents the same class of
   regression from recurring on any other URL-shaped column.

## Regression note

Earlier 26.x manager builds materialized `callback_url` as a plain `str` into
`KernelRow`. A `git log -p -S 'callback_url'` over `services/session/` and
`repositories/scheduler/db_source/` between the last known-good build and 26.4
should pinpoint the commit that introduced the `yarl.URL` leak.

## Logs

Three correlated `ERROR` entries from a single worker, all wrapping the same
underlying `asyncpg.exceptions.DataError` shown above:

- `ai.backend.manager.services.session.service` — `GET_OR_CREATE: unexpected error!`
- `ai.backend.manager.actions.processor.base`   — `Action processing error: Internal server error.`
- `ai.backend.manager.api.rest.middleware.exception`
  — `Internal server error raised inside handlers: (POST /session): <InternalServerError(500): Internal server error.>`

(Full tracebacks available; sensitive values redacted.)

JIRA Issue: BA-5910

rapsealk:
## Summary

`KernelRow.callback_url` was the lone outlier among `*Row.callback_url` mappings: it declared `Mapped[yarl.URL | None]` but used a plain `sa.UnicodeText` column type instead of the project's `URLColumn` `TypeDecorator`. Realign it with `SessionRow.callback_url` and `DeploymentRevisionRow.callback_url`, both of which already use `URLColumn`.

No DB schema change: `URLColumn.impl = UnicodeText`, so the underlying column is unchanged. Only the Python ↔ SQL coercion is restored.

## Why

Without `URLColumn.process_bind_param` stringifying the value, SQLAlchemy handed the `yarl.URL` instance straight to asyncpg, whose text codec only accepts `str`. Every `POST /session` carrying a `callback_url` failed at enqueue with:

```
asyncpg.exceptions.DataError: invalid input for query argument $46:
URL('...') (expected str, got URL)
```

Fixes #11420.

## Regression source

Introduced in **#7880 (BA-3802)** — _"feat: Migrate to SQLAlchemy 2.0 with comprehensive type safety improvements"_, commit `e58da417f3`.

That PR converted both `SessionRow` and `KernelRow` from `sa.Column` to `Mapped[...] / mapped_column(...)` declarative form. `SessionRow.callback_url` correctly preserved `URLColumn`; `KernelRow.callback_url` was switched to `sa.UnicodeText` while keeping the `Mapped[yarl.URL | None]` annotation — almost certainly a copy/paste slip during the dual migration. The bad mapping has stood on `main` ever since (no follow-up edits to those lines).

## What this PR changes

- `src/ai/backend/manager/models/kernel/row.py`: `sa.UnicodeText` → `URLColumn` on `callback_url`; add the import.
- `tests/unit/manager/models/test_kernel_row.py` (new): pin the column type so a future refactor can't silently re-introduce the same slip.
- `changes/11421.fix.md`: news fragment.

## Test plan

- [x] `pants fmt` / `lint` / `check` clean on touched files
- [x] `pants test tests/unit/manager/models/test_kernel_row.py` passes
- [ ] Manually verify `POST /session` with `callback_url` set succeeds end-to-end on a halfstack
- [ ] CI green


HyeockJinKim (approved):

rapsealk:
> 저장되는 값은 동일한가요?

Yes — looking at `URLColumn` itself rather than just the symmetry with `SessionRow`:

```python
class URLColumn(TypeDecorator[yarl.URL]):
    impl = UnicodeText
    def process_bind_param(self, value, _dialect):
        return str(value)
```

Two things follow:

**DDL is unchanged.** `impl = UnicodeText` means the underlying column stays the same `UnicodeText` it was before — no migration, no impact on existing rows, no rewrite of stored data.

**Stored bytes match the pre-regression behaviour.** On the call path, the REST handler always wraps the request payload in a `yarl.URL` before it reaches `KernelRow` (e.g. `api/rest/session/handler.py:336, 424`), so `URLColumn.process_bind_param` ends up writing `str(yarl.URL(<input string>))`. That is exactly what was written before #7880 (when the column was already `URLColumn`). For the URL that triggered #11420 — `http://…:9500/api/session-event-webhook/?token=<jwt>` — `str(yarl.URL(s)) == s`, byte-identical to the original input string. The only inputs where `str(yarl.URL(s)) != s` are the ones yarl normalises (uppercase host → lowercase, default port stripped, etc.); those normalisations also happened pre-#7880, so this PR doesn't introduce any new transformation — it restores the historical behaviour.

So:
- vs. **pre-#7880**: stored value is identical.
- vs. **post-#7880 (current `main`)**: INSERT was failing with `asyncpg.DataError`, so there was no stored value to compare against. After this PR it stores the same shape it always used to.

Happy to pin this with an explicit round-trip assertion in the regression test (`str(URLColumn().process_bind_param(yarl.URL(s), …)) == s` over a few representative URLs) if you'd like that locked down in CI.

rapsealk:
CI failures are infra flakes, not code issues:

- **`api-updated`** — Apollo's `rover.apollo.dev` CDN returned `Connection reset by peer (os error 104)` while downloading the `supergraph` compose binary.
- **`test-integration (1/2)`** — `pants` launcher download returned `HTTP 502` in the very first step; the job died in 22s before any test ran.
- **`test-integration-gate`** — cascade failure of the above.

None of these checks are red on `main`. Reran the failed jobs.

PR #11430, opened 2026-04-29
fix(BA-5914): Record destroyed session IDs in audit log entity_id

### LINKED ISSUE #11429: Session 'destroy' audit log records '(unknown)' for entity_id
## Summary

Audit-log entries for session destroy (`DELETE` operation, `entity_type=session`) record `entity_id = \\"(unknown)\\"` instead of the destroyed session's UUID. Session create entries record the UUID correctly, so the discrepancy makes destroy events untraceable in the audit log.

## Repro

1. Create a session.
1. Destroy it.
1. Query `audit_logs` for the most recent rows.

Observed:

|operation|entity_type|entity_id|
|---|---|---|
|`CREATE`|`session`|`<session-uuid>`|
|`DELETE`|`session`|`(unknown)`|

Expected: the `DELETE` row records the destroyed session's UUID.

## Root cause

`ActionProcessor` resolves `entity_id` via `action.entity_id() or result.entity_id()` (`src/ai/backend/manager/actions/processor/base.py:85-87`), then `AuditLogMonitor` falls back to the sentinel `\\"(unknown)\\"` when both are `None` (`src/ai/backend/manager/actions/monitors/audit_log.py:16,37`).

For session destroy, both return `None`:

- `DestroySessionAction.entity_id()` → `None` — the action only carries `session_name`, not an id (`src/ai/backend/manager/services/session/actions/destroy_session.py:21-22`).
- `DestroySessionActionResult.entity_id()` → `None`, with a `# TODO: Change this to entity_ids` comment (`destroy_session.py:38-41`).

The destroy service already resolves `session_name → session_ids: list[UUID]` at `src/ai/backend/manager/services/session/service.py:786`, but those ids are dropped on the floor instead of being attached to the result.

For comparison, session create works because `CreateFromParamsActionResult.entity_id()` returns `str(self.session_id)` (`src/ai/backend/manager/services/session/actions/create_from_params.py:94-95`).

## Proposed fix (option 1, this issue)

Thread `session_ids` through `DestroySessionActionResult` and have `entity_id()` return them (joined as a comma-separated string when there are multiple, since `recursive=True` can target several sessions). This is a minimum-blast-radius fix: it does not require changing `BaseActionResultMeta`, the `audit_logs` schema, or the audit log monitor.

A follow-up could properly model multi-entity audit events at the framework layer (`BaseActionResultMeta.entity_ids`), but that is out of scope here.

## Impact

Audit-log integrity for session lifecycle events. Destroy events cannot be correlated with the affected session today.

JIRA Issue: BA-5914

rapsealk:
## Summary

- `DestroySessionActionResult` now carries the list of session ids resolved by the service and returns them (comma-joined for recursive destroys) from `entity_id()`.
- `AuditLogMonitor` therefore records the affected session UUID(s) for `DELETE` events, instead of the `(unknown)` blank-id sentinel.

Fixes #11429 (BA-5914)

## Background

`ActionProcessor` resolves the audit-log entity id via `action.entity_id() or result.entity_id()` (`manager/actions/processor/base.py`). For session create, the result returns `str(self.session_id)`; for session destroy, both the action and the result returned `None`, so `AuditLogMonitor` fell back to `_BLANK_ID = "(unknown)"`. The destroy service already resolves `session_name` to `list[SessionId]` via `SessionRepository.get_target_session_ids()` — those ids are now threaded into the result.

A proper multi-id audit-log model (`BaseActionResultMeta.entity_ids`, schema change) is left as a follow-up; comma-joined ids are a minimum-blast-radius fix that fits the existing `audit_logs.entity_id` (`sa.String`, no length cap).

## Test plan

- [x] Unit tests in `tests/unit/manager/services/session/test_session_service.py` updated to assert `result.session_ids` and `result.entity_id()` for the single-session, recursive (multi), and no-target cases.
- [x] `pants fmt`, `pants fix`, `pants lint`, `pants check`, `pants test` on the changed files all pass.
- [ ] Manual: create a session, destroy it, then verify the latest `audit_logs` row for `entity_type=session, operation=DELETE` carries the destroyed session's UUID instead of `(unknown)`.

### DIFF
--- a/changes/11430.fix.md
+++ b/changes/11430.fix.md
@@ -1,1 +1,1 @@
-Record the destroyed session UUID(s) in audit log entries for session `DELETE` events instead of the `(unknown)` placeholder.
+Session destroy audit log entries now record the destroyed session UUID(s) in `entity_id` instead of the `(unknown)` placeholder, so destroy events can be correlated with the affected session.
--- a/changes/11430.fix.md
+++ b/changes/11430.fix.md
@@ -1,0 +1,1 @@
+Record the destroyed session UUID(s) in audit log entries for session `DELETE` events instead of the `(unknown)` placeholder.
--- a/src/ai/backend/manager/services/session/actions/destroy_session.py
+++ b/src/ai/backend/manager/services/session/actions/destroy_session.py
@@ -1,7 +1,7 @@
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Any, override
 
-from ai.backend.common.types import AccessKey
+from ai.backend.common.types import AccessKey, SessionId
 from ai.backend.manager.actions.action import BaseActionResult
 from ai.backend.manager.actions.types import ActionOperationType
 from ai.backend.manager.models.user import UserRole
@@ -34,8 +34,13 @@ class DestroySessionAction(SessionAction):
 class DestroySessionActionResult(BaseActionResult):
     # TODO: Add proper type
     result: Any
+    session_ids: list[SessionId] = field(default_factory=list)
 
-    # TODO: Change this to `entity_ids`
+    # TODO: Change this to `entity_ids` once BaseActionResultMeta supports
+    # multiple ids; until then, comma-join so audit logs capture every
+    # affected session (recursive destroy can target several).
     @override
     def entity_id(self) -> str | None:
-        return None
+        if not self.session_ids:
+            return None
+        return ",".join(str(sid) for sid in self.session_ids)
--- a/src/ai/backend/manager/services/session/service.py
+++ b/src/ai/backend/manager/services/session/service.py
@@ -812,7 +812,7 @@ class SessionService:
 
         # Return response - same format for both recursive and non-recursive
         resp = {"stats": last_stat}
-        return DestroySessionActionResult(result=resp)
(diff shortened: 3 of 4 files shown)


rapsealk:
Heads-up for whoever owns the audit-log UI / GraphQL filters:

After this PR, `audit_logs.entity_id` for session `DELETE` events can be a comma-joined string of session UUIDs (recursive destroy can target multiple sessions), not a single UUID. The legacy GQL filter at `src/ai/backend/manager/api/gql_legacy/audit_log.py:81` does string equality, so filtering destroy events by exact session UUID now needs `LIKE %uuid%` (or substring match) instead of `=`.

Strictly an improvement over the pre-PR state — those rows were `(unknown)` and unfilterable at all — but worth knowing before you next touch that filter or the audit-log dashboard.

The proper fix is the follow-up TODO already noted in `destroy_session.py`: extend `BaseActionResultMeta` to carry `entity_ids: list[str]` and add a corresponding `audit_logs` column. That would replace the comma-join hack and also remove the theoretical Postgres btree index limit (~2700 bytes/row) that this approach hits if a recursive destroy ever targets >70 sessions in one call.

fregataa (approved):

PR #11433, opened 2026-04-30
fix(BA-5916): route v1 CLI `-v <uuid>` mounts into `mount_ids`

jopemachine:
## Summary
- v1 CLI `-v <vfolder-uuid>` was silently dropped: `prepare_mount_arg` put every source into the legacy name-keyed `mounts` config field, but the new sokovan path on the manager (`registry.py:_mount_entries_from_creation_config`) only consumes UUID-keyed `mount_ids` / `mount_id_map` and explicitly skips string-keyed `mounts`. The session was created without ever mounting the vfolder, with no warning anywhere.
- `prepare_mount_arg` now detects UUID-shaped sources and routes them to `mount_ids` / `mount_id_map`, leaving non-UUID names in the legacy buckets. Both v1 CLI session create/start callsites forward the new buckets to `ComputeSession.get_or_create`.
- The model-serving create callsite keeps merging UUIDs back into the name list because `Service.create` does its own `/folders` lookup to resolve names↔ids — no behavior change there.
- Unit tests cover the UUID-bucket / name-bucket split.

Verified e2e: `./backend.ai session create python -v <uuid> --scaling-group default --enqueue-only` now produces a session with `vfolder_mounts = [<uuid>]` instead of an empty list.

Out of scope: digit-leading UUIDs still fail at the `MountExpression` lark grammar (CNAME-first rule), tracked separately. Server-side resolution of name-keyed `mounts` to UUIDs in the new sokovan path is also a separate issue.

Resolves BA-5916.

### DIFF
--- a/changes/BA-5916.fix.md
+++ b/changes/BA-5916.fix.md
@@ -1,0 +1,1 @@
+Route UUID-shaped vfolder sources from the v1 CLI `-v` flag into `mount_ids` / `mount_id_map` so the manager actually mounts them; previously they landed silently in the legacy name-keyed `mounts` field that the new sokovan path skips.
--- a/src/ai/backend/client/cli/service.py
+++ b/src/ai/backend/client/cli/service.py
@@ -321,7 +321,15 @@ def create(
 
     """
     envs = prepare_env_arg(env)
-    mount, mount_map, mount_options = prepare_mount_arg(mount, escape=True)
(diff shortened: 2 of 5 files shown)


jopemachine:
Wrong direction — v1 CLI `-v` is name-only by design; the real bug is the manager's new sokovan path silently dropping name-keyed mounts. Re-opening with a server-side fix.

PR #11434, opened 2026-04-30
fix(BA-5916): resolve legacy name-keyed mounts in sokovan enqueue

jopemachine:
## Summary
- v1 CLI `-v <vfolder-name>` puts the name into `creation_config["mounts"]`, but the new sokovan path's `_mount_entries_from_creation_config` builder is intentionally pure (no DB I/O) and only consumes UUID-keyed `mount_ids` / `mount_id_map`. Name-keyed entries were silently dropped — the session was created without the requested vfolder, no error or warning anywhere.
- Add a separate async resolver `_resolve_legacy_name_mounts` invoked from `_enqueue_session_via_sokovan` before the pure builder. It queries `query_accessible_vfolders` against the user scope, merges resolved UUIDs into `mount_ids` / `mount_id_map` / `mount_options`, and keeps the original builder unchanged.
- `name/subpath` syntax (e.g. `-v vf/subdir:/dst`) is rejected explicitly: `MountInfoEntry` has no subpath field, so silently dropping the subpath would mount the wrong directory. Subpath mounts must use the UUID-keyed surface, where storage proxy handles `vfsubpath` separately.

Verified e2e:

| input | result |
| --- | --- |
| `-v abc` | session has `vfolder_mounts: [<uuid>]` ✅ |
| `-v abc:/home/work/data` | mounted with destination ✅ |
| `-v doesnotexist` | 400 with clear error ✅ |
| `-v abc/sub:/dst` | 400 with subpath-not-supported error ✅ |

Resolves BA-5916.


src/ai/backend/manager/repositories/vfolder/repository.py
```diff
@@ -420,6 +420,27 @@ async def list_accessible_vfolders(
...
+        No access scoping — callers (e.g. session-create paths) are
+        responsible for validating user access against the resolved id
+        in their own action flow.
+        """
+        async with self._db.begin_readonly_session() as session:
+            row_id = await session.scalar(
+                sa.select(vfolders.c.id).where(
+                    sa.and_(
+                        vfolders.c.name == name,
+                        vfolders.c.status.not_in(HARD_DELETED_VFOLDER_STATUSES),
```
jopemachine: I don't think this status check needed in this action

src/ai/backend/manager/api/rest/session/handler.py
```diff
@@ -399,6 +468,7 @@ async def create_from_params(
             params.session_name,
         )
         architecture = params.architecture or DEFAULT_IMAGE_ARCH
+        domain_name = params.domain or request["user"]["domain_name"]
```
jopemachine: Why is it changed?

changes/11434.fix.md
jopemachine, line 1: Rename this news fragment

src/ai/backend/manager/api/rest/session/handler.py
```diff
@@ -247,6 +250,69 @@ def __init__(
         self._vfolder = vfolder
         self._config_provider = config_provider
 
+    async def _resolve_legacy_name_mounts(
+        self,
+        creation_config: dict[str, Any],
```
jopemachine: It seems more natural for this function to accept only mounts as an argument, and for the merging logic to be handled by the caller.

jopemachine: Also, mount_map also seems to be name-based, but handling for that appears to be missing.

src/ai/backend/manager/services/vfolder/actions/resolve_by_name.py
```diff
@@ -0,0 +1,42 @@
...
+from dataclasses import dataclass
+from typing import override
+
+from ai.backend.manager.actions.action.base import BaseActionResult
+from ai.backend.manager.actions.types import ActionOperationType
+from ai.backend.manager.services.vfolder.actions.base import VFolderAction
+
+
+@dataclass
+class ResolveVFolderIdByNameAction(VFolderAction):
```
jopemachine: This action exists solely to support the legacy API based on names, and should be removed once the legacy API is deprecated. Please note this briefly in a comment.

fregataa (approved):

PR #11436, opened 2026-04-30
fix(BA-5917): restore resource group auto-selection in sokovan enqueue

jopemachine:
## Summary
- Add ``SchedulerRepository.pick_default_resource_group`` — picks the first RG from the owner's allowlist when the caller omits ``--scaling-group``. Empty allowlist raises ``InvalidAPIParameters("No accessible scaling group available")``, matching the legacy ``ScalingGroupResolver`` (deleted in #11250).
- Inject ``SchedulerRepository`` into ``SessionService``, ``ModelServingService``, and ``AgentRegistry``. Each draft-construction site calls the resolver when the user omits the scaling group, then constructs the draft with the picked name. ``SchedulingController.enqueue_session_from_draft`` and the preparer chain are completely unchanged from ``main``.

## Out of scope
- ``PublicPrivateFilterRule`` (also deleted in #11250) is **not** restored here. User-provided private RGs for non-private sessions (and vice versa) are not validated. Tracked separately so this PR stays focused on the auto-selection regression.

Without this, ``backend.ai session create python -r cpu=1 -r mem=512m`` (no ``-q/--scaling-group``) returns HTTP 400 ``SessionSpec fields not resolved: scope.resource_group_name``.

Resolves BA-5917.


src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
```diff
@@ -1632,6 +1634,29 @@ async def fetch_session_spec_contexts(
...
+        domain_name: str,
+        project_id: ProjectID,
+    ) -> ResourceGroupName:
+        """Auto-select the first resource group from the owner's allowlist.
+
+        Used by the service layer when the caller omits ``--scaling-group``.
+        ``is_private`` filtering against ``session_type`` is intentionally
+        not done here — that belongs in a dedicated capability validator
+        (legacy ``PublicPrivateFilterRule`` deleted in PR #11250),
+        which still needs to be restored separately.
```
seedspirit: I think it's not necessary to include comments in the service layer or the repository layer specifying which caller uses this.

tests/unit/manager/repositories/scheduler/test_resource_group_auto_select.py
```diff
@@ -0,0 +1,61 @@
...
+
+from ai.backend.common.identifier.project import ProjectID
+from ai.backend.common.types import AccessKey
+from ai.backend.manager.models.scaling_group import ScalingGroupOpts
+from ai.backend.manager.models.utils import ExtendedAsyncSAEngine
+from ai.backend.manager.repositories.scheduler.db_source.db_source import ScheduleDBSource
+from ai.backend.manager.repositories.scheduler.types.session_creation import AllowedScalingGroup
+
+
+async def test_pick_default_resource_group_returns_allowed_rg(
```
seedspirit: nitpick: It would be best if you could update this to comply with the test code conventions (e.g., using Test Classes, making full use of fixtures, etc.).

HyeockJinKim (dismissed):

seedspirit (dismissed):

HyeockJinKim (approved):

seedspirit (approved):

PR #11444, opened 2026-04-30
fix(BA-5923): require deployment-creation fields on `DeploymentRevisionPreset`

jopemachine:
## Summary

Make `DeploymentRevisionPreset` carry every value `CreateDeploymentInput` needs at the persistence boundary, so callers like `model_card.deploy` that delegate to the preset can no longer end up with a deployment that silently fell back to system defaults.

### Field changes — `CreateDeploymentRevisionPresetInput`

Required iff the underlying row column is `nullable=False`:

| Field | Row column | DTO before | DTO after |
|---|---|---|---|
| `cluster_mode` | `nullable=False` | `str \| None = None` | `str` |
| `cluster_size` | `nullable=False` | `int \| None = None` (ge=1) | `int` (ge=1) |
| `replica_count` | now `nullable=False` (was `True`) | `int \| None = None` (ge=0) | `int` (ge=0) |
| `deployment_strategy` | now `nullable=False` (was `True`) | `DeploymentStrategyInput \| None = None` | `DeploymentStrategyInput` |

### DB migration — `ba5923b1f4a7`

Backfills the legacy nullable columns with the system defaults that `_apply_deployment_level_preset` was silently applying, then promotes them to `NOT NULL`:

- `replica_count` → `1`
- `deployment_strategy` → `'ROLLING'`
- `deployment_strategy_spec` → `'{}'::jsonb`

`server_default`s are kept on the columns so concurrent INSERTs without these fields still get a sensible value during the upgrade window. Verified locally with `upgrade → downgrade → re-upgrade` against rows holding every NULL/non-NULL combination.

### Other touched code

- `manager/models/deployment_revision_preset/row.py` — `replica_count` / `deployment_strategy` / `deployment_strategy_spec` go `nullable=False` with the matching `server_default`.
- `manager/data/deployment_revision_preset/types.py` — same three fields lose their `Optional[...]` shapes on `DeploymentRevisionPresetData`.
- `manager/repositories/deployment_revision_preset/creators.py` — `DeploymentRevisionPresetCreatorSpec` mirrors the DTO.
- `manager/api/adapters/deployment_revision_preset/adapter.py` — the create path uses a non-null `_convert_required_strategy_input` helper; the nullable variant remains for the update path.
- `manager/api/gql/deployment/types/revision_preset.py` — GQL input mirrors the DTO change.
- `manager/sokovan/deployment/deployment_controller.py` — `_apply_deployment_level_preset` drops the dead `replica_count is not None` check, and `_build_policy_from_preset` drops the dead `deployment_strategy is None` branch (return type narrows to `DeploymentPolicyConfig`).
- `tests/component/deployment_revision_preset/test_deployment_revision_preset_crud.py` — pass the newly-required fields explicitly in each case.
- `tests/unit/manager/sokovan/deployment/` — new test asserting `DeploymentController.create_deployment` reaches `repository.create_endpoint` with every preset-driven value populated when handed a preset-only `NewDeploymentCreator`, plus updates to the existing `_apply_deployment_level_preset` tests for the new contract.

Resolves BA-5923.

----
📚 Documentation preview 📚: https://sorna--11444.org.readthedocs.build/en/11444/

----
📚 Documentation preview 📚: https://sorna-ko--11444.org.readthedocs.build/ko/11444/


src/ai/backend/manager/models/alembic/versions/ba5923b1f4a7_make_revision_preset_deployment_fields_not_null.py
```diff
@@ -0,0 +1,103 @@
...
+        "replica_count",
+        existing_type=sa.Integer(),
+        nullable=False,
+        server_default=None,
+    )
+    op.alter_column(
+        "deployment_revision_presets",
+        "replica_count",
+        existing_type=sa.Integer(),
+        nullable=True,
```
seedspirit: Is there a reason why the "replica count" column appears twice?

src/ai/backend/manager/api/adapters/deployment_revision_preset/adapter.py
```diff
@@ -558,6 +558,13 @@ def _convert_strategy_input(
         """Convert DeploymentStrategyInput to (strategy, strategy_spec dict)."""
         if strategy_input is None:
             return None, None
+        return DeploymentRevisionPresetAdapter._convert_required_strategy_input(strategy_input)
```
fregataa:
```suggestion
        return self._convert_required_strategy_input(strategy_input)
```

jopemachine: `_convert_required_strategy_input` is static method, not instance method

fregataa: then please change it into instance method

seedspirit (approved):

fregataa (approved):

PR #11446, opened 2026-04-30
fix(BA-5922): normalize legacy string start_command in stored model definitions

### LINKED ISSUE #11441: Migrate str based startup command to list[str]
## Summary

- 

## Steps to Reproduce

1. 
1. 
1. 

## Expected Behavior

- 

## Actual Behavior

- 

## Logs/Errors

- 

JIRA Issue: BA-5922

seedspirit:
## Summary

- PR #11402 (BA-5891) narrowed `ModelServiceConfig.start_command` from `str | list[str]` to `list[str] | None`. Pre-existing rows that were created when the string form was still accepted now fail Pydantic validation when SQLAlchemy materializes the JSONB columns through `PydanticColumn(ModelDefinition[Draft])`, 500ing every reader (`myDeployments`/`adminDeployments`, scheduler, the `admin_refresh_deployment_revisions` repair path itself, ...).
- Adds Alembic migration `8c1f7d3a9e2b` that wraps any legacy string `start_command` value as a one-item list across the three columns holding `ModelDefinition`-shaped JSONB:
  - `deployment_revisions.model_definition`
  - `deployment_revision_presets.model_definition`
  - `runtime_variants.default_model_definition`
- Data-only migration; `downgrade()` is a no-op because the prior runtime accepted both forms and reverting would be ambiguous for rows already created in the new list form.

## Test plan

- [ ] Seed a deployment_revision / preset / runtime_variant row with the legacy string `start_command` form, run `alembic upgrade head`, confirm the row materializes via `PydanticColumn(ModelDefinition[Draft])` without 500ing.
- [ ] Verify rows that already store `start_command` as a list are left untouched.
- [ ] Verify `runtime_variants` rows with `default_model_definition.models = None` (the empty `custom` seed) are skipped without error.
- [ ] `pants check` / `pants lint` on the new migration file.

Closes #11441.

### DIFF
--- a/src/ai/backend/manager/models/alembic/versions/8c1f7d3a9e2b_normalize_legacy_string_start_command.py
+++ b/src/ai/backend/manager/models/alembic/versions/8c1f7d3a9e2b_normalize_legacy_string_start_command.py
@@ -1,23 +1,8 @@
 """normalize legacy string start_command in stored model definitions
 
-BA-5891 / PR #11402 narrowed ``ModelServiceConfig.start_command`` from
-``str | list[str]`` to ``list[str] | None``. Existing rows that were
-created when the string form was still accepted now fail Pydantic
-validation when SQLAlchemy materializes the JSONB columns through
-``PydanticColumn(ModelDefinition[Draft])``, which 500s every reader
-(``myDeployments``/``adminDeployments``, scheduler, the
-``admin_refresh_deployment_revisions`` repair path itself, ...).
-
-This migration wraps legacy string values as a one-item list so the
-new schema accepts the rows without otherwise changing the stored
-command value.
-
-Three columns hold ``ModelDefinition``-shaped JSONB and need rewriting:
-
-- ``deployment_revisions.model_definition``
-- ``deployment_revision_presets.model_definition``
-- ``runtime_variants.default_model_definition`` (Draft variant; ``models``
-  may be ``None`` for the empty seed of the ``custom`` runtime variant)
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/models/alembic/versions/8c1f7d3a9e2b_normalize_legacy_string_start_command.py
jopemachine, line 1: Is there a particular reason for implementing verbose validation logic instead of leveraging the type system for validation?

seedspirit: Since it is stored as JSONB, it is difficult to validate it using a strict type system.

HyeockJinKim (approved):

src/ai/backend/manager/models/alembic/versions/8c1f7d3a9e2b_normalize_legacy_string_start_command.py
```diff
@@ -0,0 +1,59 @@
...
+    for row_id, model_definition in rows:
+        changed = False
+        for model in model_definition.get("models") or []:
+            try:
+                start_command = model["service"]["start_command"]
+            except (KeyError, TypeError):
+                continue
+            if isinstance(start_command, str):
+                model["service"]["start_command"] = [start_command]
+                changed = True
```
rapsealk, line 41: Try with `for-except`

PR #11456, opened 2026-05-01
feat(BA-5378): add sub-entity RBAC element types for split permissions

fregataa:
## Summary
- Adds `vfolder:data`, `session:app`, and `user:name` as new `RBACElementType` values so entity metadata access can be granted without implicit access to internal data, app endpoints, or user identity discovery.
- Mirrors the new values to `RBACElementTypeDTO` (keeps GQL `RBACElementTypeGQL` schema in sync via `gql_enum`) and adds matching `EntityType` members so `RBACElementType.to_entity_type()` bridge conversions stay valid.
- Bundles the enum-only sub-tasks of two epics plus BA-5611; creator/data-migration follow-ups are tracked separately.

## Test plan
- [x] `pants test tests/unit/manager/api/gql/rbac/test_rbac_element_type_sync.py` — enum-sync guard passes
- [x] `pants fmt fix lint check` — no issues
- [x] `pants test --changed-since=HEAD --changed-dependents=transitive` — all transitively affected tests pass
- [x] CI green

Resolves BA-5378
Resolves BA-5381
Resolves BA-5611

----
📚 Documentation preview 📚: https://sorna--11456.org.readthedocs.build/en/11456/

----
📚 Documentation preview 📚: https://sorna-ko--11456.org.readthedocs.build/ko/11456/


fixtures/manager/example-roles.json
jopemachine, line 1: How about including `description` field to more clearly illustrate the roles and responsibilities of each role?

jopemachine, line 1: The news fragment mentions only administrators, but the fixture’s entity_id appears to include regular users as well.

jopemachine (approved):

PR #11457, opened 2026-05-01
feat(BA-5380): seed RBAC permissions for vfolder:data and session:app

fregataa:
## Summary
- Forward-only Alembic migrations that backfill `vfolder:data` (CRUD) and `session:app` (read) permissions for owner/admin roles in domain/project/user scopes.
- Migrates existing `vfolder_permissions` invitations to per-entity `vfolder:data` grants using the entity-as-scope pattern (`permissions(scope_type='vfolder', scope_id=vfolder_id, ...)`); mount perms map `ro→{read}`, `rw→{read,update}`, `wd→{read,update,hard-delete}` (no soft-delete on data).
- Sub-entity rows in `association_scopes_entities` are intentionally omitted: the resolver walks parent vfolder/session edges via `permission_entity_type`, and the existing `uq_scope_id_entity_id` constraint would conflict with parent edges anyway. Downgrade is a no-op to protect operator-managed grants.

## Test plan
- [x] `pants fmt fix lint check` on the two new migration files
- [x] Local DB upgrade: 102 rows seeded (vfolder:data 82 + session:app 20) on a 33-role / 1-invitation dataset
- [x] Verify entity-as-scope tuples (`vfolder`, `model_deployment` scopes) excluded from broad seed — invitee gets exactly mount-perm-mapped ops, no over-grant
- [x] Downgrade is no-op (data preserved); re-upgrade is idempotent (no duplicate rows via `ON CONFLICT DO NOTHING`)
- [x] CI green

Resolves BA-5380
Resolves BA-5383

### DIFF
--- a/src/ai/backend/manager/models/alembic/versions/3632aad9d5d9_migrate_session_app_to_rbac.py
+++ b/src/ai/backend/manager/models/alembic/versions/3632aad9d5d9_migrate_session_app_to_rbac.py
@@ -1,0 +1,98 @@
+"""migrate_session_app_to_rbac
+
+Revision ID: 3632aad9d5d9
+Revises: 6e5a7a62a687
+Create Date: 2026-05-01 00:00:01.000000
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.engine import Connection
+
+# revision identifiers, used by Alembic.
+revision = "3632aad9d5d9"
+down_revision = "6e5a7a62a687"
+branch_labels = None
+depends_on = None
+
+# Part of: 26.5.0
+
+# Constants
+MEMBER_ROLE_SUFFIX = "member"
+ENTITY_TYPE = "session:app"
+
+# Org-hierarchy scope types. Other scope_type values in `permissions`
+# (e.g. 'vfolder', 'model_deployment') are entity-as-scope grants for
+# specific entities; they must be excluded from the entity-type seed.
+ORG_SCOPE_TYPES = ["domain", "project", "user"]
+
+# session:app is owner-only and read-only: app endpoints expose live
+# state, so write/delete operations on the sub-entity itself are not
+# meaningful.
+OWNER_OPERATIONS = ["read"]
+
+
+def _seed_entity_type_permissions(db_conn: Connection) -> None:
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/models/alembic/versions/6e5a7a62a687_migrate_vfolder_data_to_rbac.py
```diff
@@ -0,0 +1,212 @@
...
+        if not rows:
+            break
+
+        last_id = rows[-1].row_id
+
+        values_list: list[dict[str, str]] = []
+        for row in rows:
+            ops = MOUNT_PERMISSION_TO_OPERATIONS.get(row.mount_permission)
+            if not ops:
+                # Unknown mount permission value — skip silently rather than
```
jopemachine: Let's log some warning message rather then silent failure

seedspirit (dismissed):

src/ai/backend/manager/models/alembic/versions/3632aad9d5d9_migrate_session_app_to_rbac.py
```diff
@@ -0,0 +1,132 @@
...
+revision = "3632aad9d5d9"
+down_revision = "6e5a7a62a687"
+branch_labels = None
+depends_on = None
+
+# Part of: 26.5.0
+
+# Constants
+MEMBER_ROLE_PATTERN = "%member"
+ENTITY_TYPE = "session:app"
```
seedspirit: The entity type enum is defined as `SESSION_APP_SERVICE = "session:app_service"`. Does it matter if there is a mismatch?

jopemachine: Also, vfolder:data does not appear to exist in the EntityType (or anywhere in the codebase).

fregataa: they are added in https://github.com/lablup/backend.ai/pull/11456

jopemachine (approved):

PR #11462, opened 2026-05-03
fix(BA-5927): flatten resourceSlots to plain list with filter/order_by

seedspirit:
## Summary

- Querying `resourceSlots` on `DeploymentRevision` or `DeploymentRevisionPreset` failed with `'AllocatedResourceSlotNode' object has no attribute 'id'`. The DTO carries only `slot_name` and `quantity` — there is no `id` for `PydanticNodeMixin.from_pydantic` to read.
- Root cause: `AllocatedResourceSlot` was modeled as a Relay Node, but Relay Node IDs must be globally unique and slot names are not.
- Fix: convert `AllocatedResourceSlotGQL` to a plain `@gql_pydantic_type` and have `resourceSlots` return `list[AllocatedResourceSlotGQL]` directly. Drop the unused Edge / Connection / `resolve_nodes`.
- Re-expose `filter` and `orderBy` arguments on the field (the lower layers already support them) — only cursor pagination is dropped, since slot count per parent is bounded.
- Extract the fetch limit (`10000`) to a named constant `RESOURCE_SLOTS_FETCH_LIMIT` so it is no longer a magic number duplicated across resolvers.

## Schema change (breaking)

`resourceSlots` on `DeploymentRevision` and `DeploymentRevisionPreset` is now `[AllocatedResourceSlot!]!` instead of an `AllocatedResourceSlotConnection`. Existing client queries that wrap it in `edges { node { ... } }` must drop that wrapping. The field accepts optional `filter: AllocatedResourceSlotFilter` and `orderBy: [AllocatedResourceSlotOrderBy!]` arguments. Announced via `changes/11462.breaking.md`.

## Test plan

- [x] `pants fmt`, `pants fix`, `pants lint` clean.
- [x] Schema import smoke test (`schema OK`).
- [x] Existing unit test `tests/unit/manager/api/gql/deployment/test_revision_preset_types.py` still passes.
- [x] Live verification on the running server — preset side, end-to-end:
  - baseline (no args) returns the flat list with no `id`-attribute error
  - `orderBy` on `SLOT_NAME` and `QUANTITY` (ASC/DESC) all sort correctly
  - `filter` with `equals`, `contains`, `in`, `endsWith`, plus `AND` / `OR` composition all match
  - `filter + orderBy` combined produces the filtered, ordered subset
- [x] Revision side schema accepts the new `filter` / `orderBy` arguments (Strawberry validation passes); the adapter helpers (`_convert_allocated_slot_filter`, the order resolvers) are shared with the preset path.

Resolves BA-5927

### DIFF
--- a/src/ai/backend/manager/api/gql/deployment/types/revision.py
+++ b/src/ai/backend/manager/api/gql/deployment/types/revision.py
@@ -138,6 +138,7 @@ from ai.backend.manager.api.gql_legacy.scaling_group import ScalingGroupNode
 from ai.backend.manager.api.gql_legacy.vfolder import VirtualFolderNode
 
 from .resource_slot import (
+    RESOURCE_SLOTS_FETCH_LIMIT,
     AllocatedResourceSlotFilterGQL,
     AllocatedResourceSlotGQL,
     AllocatedResourceSlotOrderByGQL,
@@ -523,7 +524,7 @@ class ModelRevision(PydanticNodeMixin[RevisionNodeDTO]):
             input=SearchAllocatedResourceSlotsInput(
                 filter=filter.to_pydantic() if filter else None,
                 order=[o.to_pydantic() for o in order_by] if order_by else None,
-                limit=10000,
+                limit=RESOURCE_SLOTS_FETCH_LIMIT,
             ),
         )
         return [AllocatedResourceSlotGQL.from_pydantic(item) for item in payload.items]
--- a/src/ai/backend/manager/api/gql/deployment/types/revision.py
+++ b/src/ai/backend/manager/api/gql/deployment/types/revision.py
@@ -10,7 +10,7 @@ from uuid import UUID
 
 import strawberry
 from strawberry import ID, Info
-from strawberry.relay import Connection, Edge, NodeID, PageInfo
+from strawberry.relay import Connection, Edge, NodeID
 from strawberry.scalars import JSON
 
 from ai.backend.common.config import (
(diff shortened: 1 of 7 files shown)


fregataa (commented):
I think `AllocatedResourceSlotNodeGQL` should not be a node type because ID value should be globally unique
Let's change AllocatedResourceSlotNodeGQL to a normal GQL type rather than relay node

src/ai/backend/manager/api/gql/deployment/types/revision.py
```diff
@@ -515,44 +513,20 @@ async def resource_slots(
...
-            count=payload.total_count,
-            edges=edges,
-            page_info=PageInfo(
-                has_next_page=payload.has_next_page,
-                has_previous_page=payload.has_previous_page,
-                start_cursor=edges[0].cursor if edges else None,
-                end_cursor=edges[-1].cursor if edges else None,
+                filter=filter.to_pydantic() if filter else None,
+                order=[o.to_pydantic() for o in order_by] if order_by else None,
+                limit=10000,
```
jopemachine: Let's extract magic numbe into reusable named constants.

jopemachine:
What I meant was that there could be other cases where values exceeding the maximum are passed to the limit field in SearchXXXInput. For reusability, I was suggesting defining a shared FETCH_LIMIT constant.

jopemachine (approved):

fregataa (approved):

PR #11463, opened 2026-05-03
feat(BA-5921): launch model service from runtime-variant defaults and preset args

seedspirit:
## Summary
- Tokenize runtime-variant preset ARGS values so each flag and value lands as its own argv token instead of a single `"key value"` string.
- Append resolved preset ARGS to `service.start_command` in the deployment draft builder; the kernel-level `startup_command` no longer absorbs them.
- Resolve `{model_path}` placeholders inside `ModelConfigDraft.to_resolved` so the revision snapshot carries fully-resolved argv with no template syntax reaching the kernel.
- Seed runtime-variant baselines (vllm, sglang, huggingface-tgi, modular-max) with their canonical launcher `start_command` using the `{model_path}` placeholder.
- Rename the agent's `_populate_missing_model_service_start_commands` to `_apply_image_cmd_fallback`; behavior unchanged, used only by the `cmd` variant which intentionally inherits from the image's `Config.Cmd`.

## Test plan
- [x] Unit: `pants test tests/unit/common/test_config.py tests/unit/manager/sokovan/deployment/test_deployment_draft_builder.py tests/unit/agent/test_model_service_start_command.py`
- [x] Live e2e — vllm placeholder substitution: container ran `vllm serve /models` and reached healthy
- [x] Live e2e — sglang named flag substitution: container received `python -m sglang.launch_server --model-path /models`
- [x] Live e2e — custom variant + vfolder `model-definition.yaml`: snapshot pass-through, kernel ran user-supplied command
- [x] Image CMD fallback (`cmd` variant) preserved by `_apply_image_cmd_fallback`

### DIFF
--- a/changes/11463.feature.md
+++ b/changes/11463.feature.md
@@ -1,0 +1,1 @@
+Launch model services from runtime-variant default start_command and deployment preset ARGS in inference deployments
--- a/fixtures/manager/example-runtime-variants.json
+++ b/fixtures/manager/example-runtime-variants.json
@@ -8,6 +8,7 @@
           {
             "name": "vllm-model",
             "service": {
+              "start_command": ["vllm", "serve", "{model_path}"],
               "port": 8000,
               "health_check": {
                 "path": "/health",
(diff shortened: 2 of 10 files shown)


jopemachine:
Shouldn’t the semantics of the {model_path} placeholder be explicitly documented somewhere?

fregataa (approved):

PR #11469, opened 2026-05-04
fix(BA-5939): add UUID filter to ImageV2Filter

fregataa:
## Summary
- Add `id: UUIDFilter` to `ImageFilterInputDTO` and the `ImageV2FilterGQL` input so callers can locate a specific image by UUID via the v2 GraphQL/REST search APIs.
- Wire the new field into `ImageAdapter._convert_filter`, reusing the existing `ImageConditions.by_ids` for both `equals` and `in_` paths.
- Add v2 admin search component tests (the v2 path was previously uncovered) — tests register `V2ImageHandler` alongside the existing v1 routes and assert `id.equals` / `id.in_` filtering plus the superadmin-required permission gate.

## Test plan
- [x] `pants test tests/component/image::` (covers new `test_image_v2_search.py`)
- [x] Manual verification via `./bai admin image search` once GraphQL schema docs are regenerated (separate environment due to local Strawberry/Federation issue)

Resolves BA-5939

----
📚 Documentation preview 📚: https://sorna--11469.org.readthedocs.build/en/11469/

----
📚 Documentation preview 📚: https://sorna-ko--11469.org.readthedocs.build/ko/11469/

### DIFF
--- a/changes/11469.fix.md
+++ b/changes/11469.fix.md
@@ -1,0 +1,1 @@
+Add a UUID filter (`id`) to `ImageV2Filter` so callers can locate a specific image by UUID via the v2 image search APIs.
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -7349,6 +7349,9 @@ input ImageV2Filter
   name: StringFilter = null
   architecture: StringFilter = null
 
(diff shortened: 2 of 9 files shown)


jopemachine:
We already provide a single-entity query like image_v2, so is there a need for a separate id filter?

seedspirit (approved):

PR #11471, opened 2026-05-04
feat(BA-5943): add option to delete associated vfolder when deleting model card

jopemachine:
## 📚 Stacked PRs

This PR is stacked on top of another. **Merge in order:**

1. 👉 **#11471** — `feat(BA-5943): add option to delete associated vfolder when deleting model card` ← you are here
2. ⬇️ #11474 — `refactor(BA-5946): bulk-delete model cards in a single DB transaction`

#11474 carries the bulk transactional follow-up; merge this PR first so GitHub retargets #11474 onto `main`.

## Summary
- Add `DeleteModelCardOptions` DTO with a `delete_associated_vfolder` flag, exposed through REST v2 and the Strawberry GraphQL mutations (`adminDeleteModelCardV2`, `adminDeleteModelCardsV2`).
- When the option is set, `ModelCardAdapter.delete` / `bulk_delete` chains a `DeleteVFolderV2Action` after each model card deletion so the underlying VFolder is moved to trash in the same call.
- Fixes the multi-step flow where admins had to delete the model card and then separately trash the model folder from the Data tab.

Resolves BA-5943.

----
📚 Documentation preview 📚: https://sorna--11471.org.readthedocs.build/en/11471/

----
📚 Documentation preview 📚: https://sorna-ko--11471.org.readthedocs.build/ko/11471/


src/ai/backend/common/dto/manager/v2/model_card/request.py
```diff
@@ -107,10 +107,38 @@ class SearchModelCardsInput(BaseRequestModel):
...
+            "If true, also soft-delete (move to trash) the model VFolder "
+            "associated with the deleted model card."
+        ),
+    )
+
+
+class DeleteModelCardsOptions(BaseRequestModel):
+    """Options for bulk model card delete operation."""
+
+    delete_associated_folder: bool = Field(
```
jopemachine: delete_associated_vfolder

src/ai/backend/manager/api/adapters/model_card/adapter.py
```diff
@@ -369,18 +371,34 @@ async def update(
...
+        card_id: UUID,
+        options: DeleteModelCardOptions,
+    ) -> DeleteModelCardPayload:
         result = await self._processors.model_card.delete.wait_for_complete(
             DeleteModelCardAction(id=card_id)
         )
+        if options.delete_associated_vfolder:
+            await self._processors.vfolder.delete_v2.wait_for_complete(
+                DeleteVFolderV2Action(vfolder_id=result.model_card.vfolder_id)
+            )
```
seedspirit: Rather than having the adapter layer call different processors, wouldn’t it be better to design the workflow so that the service layer and repository layer receive the options and process them in a single database transaction?

src/ai/backend/manager/api/adapters/model_card/adapter.py
```diff
@@ -369,18 +371,34 @@ async def update(
...
         """Delete multiple model cards by ID."""
         for card_id in input.ids:
-            await self._processors.model_card.delete.wait_for_complete(
+            result = await self._processors.model_card.delete.wait_for_complete(
                 DeleteModelCardAction(id=card_id)
             )
+            if options.delete_associated_vfolder:
+                await self._processors.vfolder.delete_v2.wait_for_complete(
+                    DeleteVFolderV2Action(vfolder_id=result.model_card.vfolder_id)
+                )
```
seedspirit: Even for bulk operations, rather than iterating through a for loop to delete each model card individually, it seems more efficient—both in terms of performance and for ensuring rollback by keeping everything within a single transaction—to create a bulk delete action and have the repository layer or similar component delete the vfolder as well.

src/ai/backend/manager/api/adapters/model_card/adapter.py
jopemachine: There is a concern that, without introducing a separate mechanism for sharing this logic, the vfolder trash handling will be duplicated within the ModelCard service. However, since the duplicated code is relatively small and unlikely to change frequently, I will proceed with duplication and handle it at the database source level.

src/ai/backend/manager/api/adapters/model_card/adapter.py
jopemachine: It doesn’t seem necessary to apply the bulk_delete pattern within this PR.
I’ll address it separately through a dedicated issue and PR.

src/ai/backend/manager/repositories/model_card/db_source/db_source.py
```diff
@@ -163,16 +170,58 @@ async def _apply_min_resource_change(
...
+        mount_sessions = await get_sessions_by_mounted_folder(
+            session, VFolderID.from_row(vfolder_row)
+        )
+        if mount_sessions:
+            session_ids = [str(session_id) for session_id in mount_sessions]
+            raise VFolderDeletionNotAllowed(
+                "Cannot delete the vfolder. "
+                f"The vfolder(id: {vfolder_row.id}) is mounted on sessions(ids: {session_ids})."
+            )
+        vfolder_row.status = VFolderOperationStatus.DELETE_PENDING
```
seedspirit, line 223: no need to flush here?

jopemachine: It appears that this will be called automatically once the deletion is complete. 
Why would it need to be called manually?

seedspirit: I mentioned that while it seems likely that anything assigned to the ORM instance will be persisted to the database when the session ends, it would be better to do so explicitly.

jopemachine: I think it should be called only when necessary.
There doesn’t seem to be a need to redundantly use flush everywhere.

seedspirit (approved):

PR #11478, opened 2026-05-04
fix(BA-5949): preserve raw PromQL label matchers in preset template render

seedspirit:
## Summary

`MetricPreset.render()` calls `template.format(...)` on the user-supplied PromQL template. PromQL label matchers (`{job="api"}`, `{mode!="idle"}`, ...) share the `{...}` delimiter with Python `str.format`, so any preset whose template was pasted from Grafana — or simply contains literal label matchers — crashes at execute time with `ValueError: expected ':' after conversion specifier` or similar cryptic errors.

This PR makes the render path tolerate raw PromQL pastes regardless of how the user wrote braces, wraps remaining failures in a domain exception, and adds save-time validation for foreign template variables.

### Render path — symmetric brace handling

`MetricPreset.render()` runs each `{X}` block of the template through a context-aware preprocessor before `.format()`. The preprocessor inspects the original characters around the match (`already_wrapped` = preceded by `{` AND followed by `}`) and decides:

| `{X}` content | already wrapped | Action |
|---|---|---|
| Raw matcher (e.g. `job="api"`) | No (1-brace `{X}`) | Escape to `{{X}}` so `.format` keeps it literal |
| Raw matcher | Yes (2-brace `{{X}}`) | Leave alone — already user-escaped |
| `{window}` / `{group_by}` | — | Leave alone (no PromQL braces required) |
| `{labels}` | No (bare 1-brace) | Auto-wrap to `{{{labels}}}` so `.format` produces `{value}` |
| `{labels}` | Yes (3-brace `{{{labels}}}`) | Leave alone — seeded convention |

Net effect: all of `metric{job="api"}`, `metric{{job="api"}}`, `metric{labels}`, and `metric{{{labels}}}` render to a single PromQL `{value}` matcher. No DB migration; the existing `{{{labels}}}` seeded convention keeps working.

### Failure path

`.format()`'s `ValueError`/`KeyError`/`IndexError` are wrapped into a new `InvalidMetricPresetTemplate` domain exception so the manager never leaks raw stdlib errors to the API layer.

### Save path

New `validate_query_template()` helper, wired into Pydantic `field_validator` on `query_template` for both v1 and v2 create/update DTOs. Rejects:
- Foreign template variables (`$ident`, `${ident}` — typically Grafana `$__rate_interval`, `$service`)
- Templates that fail a dry-run render (orphan/nested braces)

## Test plan

Unit tests added/updated:
- [x] `tests/unit/common/clients/prometheus/test_preset.py` — raw matcher pass-through (1-brace), user-escaped 2-brace pass-through, bare `{labels}` auto-wrap, render error wrap, `validate_query_template` accepts/rejects.
- [x] `tests/unit/common/dto/manager/v2/prometheus_query_preset/test_request.py` — DTO validator rejects foreign vars and malformed braces, skips when `query_template=None` on modify.

Manual verification against local manager (`./bai prometheus-query-definition ...`):
- [x] **A. Create reject** — `$__rate_interval`, `$service`, `${region}`, orphan `{`, nested `{a{b}c}` all rejected with `InvalidMetricPresetTemplate`.
- [x] **B. Create accept** — raw `{mode!="idle"}` (1-brace), user-escaped `{{mode!="idle"}}` (2-brace), classic `{{{labels}}}`, multi-matcher + regex, bare `{labels}` all saved.
- [x] **C. Execute regression** — preset with raw `{mode!="idle"}` no longer crashes; query reaches Prometheus. 2-brace user-escaped form returns 200. Bare `{labels}` substitutes with `{key="value"}`. Seeded `container_gauge` still works.
- [x] **D. Modify** — Grafana var rejected on update; non-template field updates skip validation; raw matcher update accepted.
- [x] **E. DB-injected bypass** — manually `UPDATE`-ed templates with orphan/nested braces surface as `InvalidMetricPresetTemplate` server-side rather than raw `ValueError`. Empty `{}` and unknown `{wrong}` placeholders pass render and are evaluated by Prometheus as expected.

## Out of scope

- During E2E verification, the v2 REST endpoint was observed to serialize `BackendAIError` responses with status `HTTP/1.1 -1`, which causes clients to surface a transport-level parse error instead of the structured domain error. The exception is raised correctly server-side; only the HTTP response mapping is broken. Will file as a separate issue.

### DIFF
--- a/src/ai/backend/common/clients/prometheus/preset.py
+++ b/src/ai/backend/common/clients/prometheus/preset.py
@@ -1,7 +1,39 @@
+import re
 from collections.abc import Mapping, Set
 from dataclasses import dataclass, field
 from enum import StrEnum
 from typing import Self
+
+from ai.backend.common.exception import InvalidMetricPresetTemplate
+
+_PLACEHOLDER_NAMES = frozenset({"labels", "window", "group_by"})
+_BRACE_BLOCK_RE = re.compile(r"\{([^{}]*)\}")
+# Matches `$ident` / `${ident}` — foreign template syntax (Grafana, shell, etc.)
+# that Backend.AI does not substitute and Prometheus would reject.
+_UNSUPPORTED_TEMPLATE_VAR_RE = re.compile(r"\$\{[^}]+\}|\$[A-Za-z_][A-Za-z0-9_]*")
+
+
+def validate_query_template(template: str) -> None:
+    """Reject templates with foreign variables or malformed braces."""
+    unsupported_vars = _UNSUPPORTED_TEMPLATE_VAR_RE.findall(template)
+    if unsupported_vars:
+        raise InvalidMetricPresetTemplate(
+            f"Unsupported template variables: {unsupported_vars}. "
+            "Use placeholders {labels}, {window}, {group_by} or literal PromQL values."
+        )
+    # Check for malformed braces by attempting to render with dummy values.
+    MetricPreset(template=template).render()
+
+
+def _escape_non_placeholders(template: str) -> str:
+    # PromQL label matchers `{key=...}` collide with str.format placeholders;
+    # escape any non-placeholder `{...}` block so .format treats it as literal.
+    def repl(match: re.Match[str]) -> str:
+        if match.group(1) in _PLACEHOLDER_NAMES:
+            return match.group(0)
+        return "{{" + match.group(1) + "}}"
+
+    return _BRACE_BLOCK_RE.sub(repl, template)
 
 
 class LabelOperator(StrEnum):
@@ -34,7 +66,7 @@ def _escape_label_value(value: str) -> str:
 
 @dataclass(frozen=True)
 class MetricPreset:
-    """PromQL query preset with template and injectable values."""
+    """PromQL query preset with template (placeholders: {labels}, {window}, {group_by})."""
 
     # PromQL template (placeholders: {labels}, {window}, {group_by})
     template: str
@@ -54,8 +86,13 @@ class MetricPreset:
             f'{key}{value.operator}"{_escape_label_value(value.value)}"'
             for key, value in self.labels.items()
         )
-        return self.template.format(
-            labels=label_str,
-            window=self.window,
-            group_by=",".join(sorted(self.group_by)),  # sorted for consistency
-        )
+        try:
+            return _escape_non_placeholders(self.template).format(
+                labels=label_str,
+                window=self.window,
+                group_by=",".join(sorted(self.group_by)),
+            )
+        except (ValueError, KeyError, IndexError) as e:
+            raise InvalidMetricPresetTemplate(
+                f"Failed to render PromQL template ({type(e).__name__}: {e}): {self.template!r}"
+            ) from e
--- a/src/ai/backend/common/clients/prometheus/preset.py
+++ b/src/ai/backend/common/clients/prometheus/preset.py
@@ -8,21 +8,24 @@ from ai.backend.common.exception import InvalidMetricPresetTemplate
 
 _PLACEHOLDER_NAMES = frozenset({"labels", "window", "group_by"})
 _BRACE_BLOCK_RE = re.compile(r"\{([^{}]*)\}")
-# Matches `$ident` / `${ident}` — foreign template syntax (Grafana, shell, etc.)
-# that Backend.AI does not substitute and Prometheus would reject.
+# `$ident` / `${ident}` — foreign templating syntax (Grafana, shell, etc.)
+# that Backend.AI does not substitute and is almost always unintended in PromQL.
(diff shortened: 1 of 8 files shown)


src/ai/backend/common/clients/prometheus/preset.py
```diff
@@ -1,8 +1,48 @@
...
+def validate_query_template(template: str) -> None:
+    """Reject templates with foreign variables or malformed braces."""
+    unsupported_vars = _UNSUPPORTED_TEMPLATE_VAR_RE.findall(template)
+    if unsupported_vars:
+        raise InvalidMetricPresetTemplate(
+            f"Unsupported template variables: {unsupported_vars}. "
+            "Use placeholders {labels}, {window}, {group_by} or literal PromQL values."
+        )
+    # Check for malformed braces by attempting to render with dummy values.
+    MetricPreset(template=template).render()
```
jopemachine: Why not returning this strnig value instead of None?
It seems to be useful for test

src/ai/backend/common/clients/prometheus/preset.py
```diff
@@ -1,8 +1,48 @@
...
+_UNSUPPORTED_TEMPLATE_VAR_RE = re.compile(r"\$\{[^}]+\}|\$[A-Za-z_][A-Za-z0-9_]*")
+
+
+def validate_query_template(template: str) -> None:
+    """Reject templates with foreign variables or malformed braces."""
+    unsupported_vars = _UNSUPPORTED_TEMPLATE_VAR_RE.findall(template)
+    if unsupported_vars:
+        raise InvalidMetricPresetTemplate(
+            f"Unsupported template variables: {unsupported_vars}. "
+            "Use placeholders {labels}, {window}, {group_by} or literal PromQL values."
```
jopemachine: Nit: we can use `_PLACEHOLDER_NAMES`

fregataa (dismissed):
Let's add a new `prometheus preset` type

src/ai/backend/common/exception.py
```diff
@@ -1069,6 +1069,20 @@ def error_code(self) -> ErrorCode:
...
+
+    error_type = "https://api.backend.ai/probs/invalid-metric-preset-template"
+    error_title = "Invalid metric preset template."
+
+    def error_code(self) -> ErrorCode:
+        return ErrorCode(
+            domain=ErrorDomain.METRIC,
+            operation=ErrorOperation.READ,
+            error_detail=ErrorDetail.INVALID_PARAMETERS,
+        )
```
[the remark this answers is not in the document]
seedspirit: Good catch — fixed in 24e378f1d. Mixed in `web.HTTPBadRequest` so DTO/render failures now surface as a proper 400 response (verified locally: client sees structured `InvalidRequestError(400, ...)` instead of `HTTP/1.1 -1`).

src/ai/backend/common/clients/prometheus/preset.py
```diff
@@ -1,8 +1,48 @@
...
 from dataclasses import dataclass, field
 from enum import StrEnum
 from typing import Self
 
+from ai.backend.common.exception import InvalidMetricPresetTemplate
+
+_PLACEHOLDER_NAMES = frozenset({"labels", "window", "group_by"})
+_BRACE_BLOCK_RE = re.compile(r"\{([^{}]*)\}")
+# Matches `$ident` / `${ident}` — foreign template syntax (Grafana, shell, etc.)
+# that Backend.AI does not substitute and Prometheus would reject.
```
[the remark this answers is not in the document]
seedspirit: Reworded the comment in 24e378f1d to focus on "foreign templating syntax that Backend.AI doesn't substitute," since `$` itself is valid inside PromQL string literals. Tightening the regex to skip quoted strings is possible but seems unnecessary in practice — `$ident` outside quotes is almost always unintentional.

src/ai/backend/common/clients/prometheus/preset.py
seedspirit: Done in 24e378f1d — error message now lists placeholders built from `_PLACEHOLDER_NAMES` (sorted) instead of hard-coding them.

src/ai/backend/common/clients/prometheus/preset.py
seedspirit: Done in 24e378f1d — `validate_query_template()` now returns the dry-run rendered string. The Pydantic field validators still discard the return and pass through the original template (we don't want to store the rendered form), but the helper is now usable directly in tests for inspection.

seedspirit: Misread your intent earlier — addressed properly in 72223c9a4. `test_accepts_valid_template` now compares `validate_query_template()`'s dry-run output against an expected string, so the render path is actually exercised per case rather than just confirming no exception.

jopemachine (approved):

PR #11479, opened 2026-05-04
feat(BA-5948): cascade model card cleanup on vfolder delete-forever/purge

fregataa:
## Summary

- Single `purge` endpoint (REST + GQL, single + bulk) accepts a `PurgeVFolderOptions` body whose `cascade_model_card` flag controls model-card cleanup.
- With `cascade_model_card=false` (the default) or no body, purge is rejected with `VFolderHasLinkedModelCard` when any model card still references the vfolder.
- With `cascade_model_card=true`, the linked model card row(s) are deleted atomically alongside the vfolder data; the response payload still carries the purged vfolder's id.
- Repository: `delete_vfolders_forever` performs the atomic lookup/cleanup with partial-success semantics; the legacy purger-based `purge_vfolder` translates the FK violation into `VFolderHasLinkedModelCard` so older callers get a meaningful business error.

## Test plan
- [x] `purge` with no body / `cascade_model_card=false` against a model-card-linked vfolder → 400 `VFolderHasLinkedModelCard`.
- [x] `purge` with `cascade_model_card=true` → linked model card row(s) and vfolder data both removed; response payload carries the vfolder id.
- [x] Bulk `purge` honors the same flag for every vfolder in the batch.

Resolves BA-5948

----
📚 Documentation preview 📚: https://sorna--11479.org.readthedocs.build/en/11479/

----
📚 Documentation preview 📚: https://sorna-ko--11479.org.readthedocs.build/ko/11479/


docs/manager/graphql-reference/supergraph.graphql
```diff
@@ -11306,7 +11311,7 @@ type Mutation
   deleteVfolderV2(vfolderId: UUID!): DeleteVFolderV2Payload! @join__field(graph: STRAWBERRY)
 
   """Added in 26.4.2. Permanently purge a virtual folder."""
-  purgeVfolderV2(vfolderId: UUID!): PurgeVFolderV2Payload! @join__field(graph: STRAWBERRY)
+  purgeVfolderV2(vfolderId: UUID!, cascadeModelCard: Boolean! = false): PurgeVFolderV2Payload! @join__field(graph: STRAWBERRY)
```
jopemachine: Considering that additional options may be introduced in the future, it would be better to define an Options type and include cascadeModelCard within it.

src/ai/backend/manager/services/vfolder/actions/vfolder_v2.py
```diff
@@ -55,9 +55,15 @@ def target_entity_id(self) -> str:
 
 @dataclass
 class PurgeVFolderV2Action(VFolderSingleEntityAction):
-    """Permanently purge a vfolder by ID with RBAC enforcement."""
+    """Permanently purge a vfolder by ID with RBAC enforcement.
```
jopemachine, line 58: Wouldn't passing the Purger object to the Action be more aligned with the existing conventions?

src/ai/backend/manager/services/vfolder/services/vfolder.py
```diff
@@ -720,7 +725,9 @@ async def force_delete(
         vfolder_data = await self._vfolder_repository.get_by_id_validated(
             action.vfolder_uuid, user.id, user.domain_name
         )
-        await self._vfolder_repository.delete_vfolders_forever([action.vfolder_uuid])
+        result = await self._vfolder_repository.delete_vfolders_forever([action.vfolder_uuid])
+        if result.failures:
+            raise result.failures[0].exception
```
jopemachine, line 730: It would be better to have a more robust approach than just throwing the first error.
(ExceptionGroup or something)

src/ai/backend/client/v2/domains_v2/vfolder.py
```diff
@@ -133,11 +134,18 @@ async def delete(self, vfolder_id: UUID) -> DeleteVFolderPayload:
             response_model=DeleteVFolderPayload,
         )
 
-    async def purge(self, vfolder_id: UUID) -> PurgeVFolderPayload:
-        """Permanently delete a vfolder."""
+    async def purge(
+        self,
+        vfolder_id: UUID,
+        request: PurgeVFolderInput | None = None,
```
seedspirit, line 140: Shouldn't it be set up to accept only the `PurgeVFolderInput`, rather than accepting both the `PurgeVFolderInput` and the `vfolder_id` separately? Also it seems `PurgeVFolderInput` contains vfolder id

src/ai/backend/manager/api/adapters/vfolder/adapter.py
```diff
@@ -433,9 +434,12 @@ async def restore(self, vfolder_id: UUID) -> RestoreVFolderPayload:
...
         return RestoreVFolderPayload(id=vfolder_id)
 
-    async def purge(self, vfolder_id: UUID) -> PurgeVFolderPayload:
-        """Permanently delete a vfolder. RBAC enforced."""
-        action = PurgeVFolderV2Action(vfolder_id=vfolder_id)
+    async def purge(self, vfolder_id: UUID, input: PurgeVFolderInput) -> PurgeVFolderPayload:
+        """Permanently delete a vfolder, optionally cascading linked model cards."""
+        action = PurgeVFolderV2Action(
+            vfolder_id=vfolder_id,
+            cascade_model_card=input.cascade_model_card,
```
seedspirit: As @jopemachine mentioned, it seems better to introduce Options class here as more options could be added when purge vfolder

src/ai/backend/manager/repositories/vfolder/types.py
```diff
@@ -16,11 +18,29 @@
...
 __all__ = (
+    "BulkVFolderPurgeFailure",
+    "BulkVFolderPurgeResult",
     "ProjectVFolderSearchScope",
     "UserVFolderSearchScope",
 )
 
 
+@dataclass(frozen=True)
+class BulkVFolderPurgeFailure:
```
seedspirit: nitpick: Since this refers to a single failure, it would be more intuitive to name this object `VFolderPurgeFailure` and refer to a collection of these `VFolderPurgeFailure` objects as `BulkVFolderPurgeResult`.

src/ai/backend/manager/services/vfolder/services/vfolder.py
fregataa: - this service always requests to delete a single vfolder
- a client expects to get an exception directly about the single vfolder

I'd rather impl another bulk purge API and service not to break expected behavior of this legacy service

src/ai/backend/manager/services/vfolder/actions/vfolder_v2.py
fregataa: 1. it should be updater, not a purger
2. the vfolder repository should determine whether to change a status of a given vfolder or not depending on the number of model cards of the vfolder

docs/manager/graphql-reference/supergraph.graphql
```diff
@@ -13044,13 +13049,23 @@ type PurgeUserV2Payload
   success: Boolean!
 }
 
+"""Added in 26.4.2. Optional behavior flags for vfolder purge operations."""
```
jopemachine: Was this type added in version 26.4.2?

src/ai/backend/manager/api/adapters/vfolder/adapter.py
```diff
@@ -526,9 +530,16 @@ async def bulk_delete(self, input: BulkDeleteVFoldersInput) -> BulkDeleteVFolder
         return BulkDeleteVFoldersPayload(deleted_count=len(input.ids))
 
     async def bulk_purge(self, input: BulkPurgeVFoldersInput) -> BulkPurgeVFoldersPayload:
```
jopemachine, line 533: Wouldn't there be cases where cascade should be applied to some vfolders but not to others?

src/ai/backend/manager/repositories/vfolder/repository.py
```diff
@@ -628,30 +650,103 @@ async def restore_vfolders_from_trash(self, vfolder_ids: list[uuid.UUID]) -> lis
...
+                    model_card_rows=[],
+                ),
+            )
+            if card_row is not None:
+                record.model_card_rows.append(card_row)
+        return list(grouped.values())
+
     @vfolder_repository_resilience.apply()
-    async def delete_vfolders_forever(self, vfolder_ids: list[uuid.UUID]) -> list[VFolderData]:
+    async def delete_vfolders_forever(
```
jopemachine, line 688: Just a question: do delete_vfolder_forever and purge mean the same thing? If so, wouldn't it be better to standardize on purge?

fregataa: we decided that:
- delete_vfolder_forever deletes real data of vfolders and change status of a vfolder to delete-complete
- purge deletes db records of vfolders

but it has broken

src/ai/backend/manager/repositories/vfolder/types.py
```diff
@@ -16,11 +18,29 @@
...
+
+    vfolder_id: UUID
+    exception: BackendAIError
+
+
+@dataclass
+class BulkVFolderPurgeResult:
+    """Partial-success result of ``delete_vfolders_forever``."""
+
+    succeeded: list[VFolderData] = field(default_factory=list)
```
jopemachine, line 40: Do we need to expose not only the UUID but also the entire data of the purged vfolder?

jopemachine (approved):

PR #11480, opened 2026-05-04
feat(BA-5947): expose model_cards connection on VFolderGQL

fregataa:
## Summary
- Add a reverse-direction `modelCards` connection on `VFolderGQL` (a vfolder may back multiple cards since the model card unique constraint is on `(name, domain, project)`, not on `vfolder`).
- The nested resolver accepts the full Relay argument set (`filter`, `orderBy`, `before/after`, `first/last`, `limit/offset`) and is nullable per BA-5956.

## Test plan
- [x] `pants check` / `pants lint` clean
- [x] Schema dump regenerated; `modelCards` field nullable, no leakage of `vfolder_id` into the public `ModelCardV2Filter`

Resolves BA-5947


src/ai/backend/manager/api/gql/vfolder_v2/types/node.py
```diff
@@ -64,6 +74,31 @@ class VFolderGQL(PydanticNodeMixin[VFolderNode]):
...
 
+    @gql_added_field(
+        BackendAIGQLMeta(
+            added_version=NEXT_RELEASE_VERSION,
+            description="Model cards backed by this vfolder, sorted most-recently-created first.",
+        )
+    )  # type: ignore[misc]
+    async def model_cards(
+        self,
+        info: Info[StrawberryGQLContext],
```
jopemachine: How about sorting by most recently created, and instead of adding comments, accepting filter and order options?

seedspirit: It appears that this is already implemented within the DataLoader.
```python
ModelCardOrders.created_at(ascending=False),
ModelCardOrders.id(ascending=False),
```

jopemachine (approved):

PR #11488, opened 2026-05-06
fix(BA-5955): add shmem to image_min_mem when filling default resources

fregataa:
## Summary
- `ComputeKernelResourcesRule` filled missing `mem` from the raw image minimum, but `ResourceLimitRule` validates against `image_min_mem + shmem`. Sessions created without explicit resources were rejected with `kernel_specs[0] resource request is smaller than the image minimum`.
- Mirror the validator's shmem accounting in the preparer via a new `_apply_shmem_to_mem_min` helper so default-filled `mem` clears the validator.
- Add a regression test that asserts default-filled `mem` is at least `image_min_mem + shmem`, and tighten the existing fill test to check the exact value.

## Test plan
- [x] `pants test tests/unit/manager/sokovan/scheduling_controller/preparers/test_compute_kernel_resources_rule.py`
- [x] `pants test tests/unit/manager/sokovan/scheduling_controller/validators/test_session_spec_rules.py`

Resolves BA-5955

### DIFF
--- a/changes/11488.fix.md
+++ b/changes/11488.fix.md
@@ -1,0 +1,1 @@
+Fix session creation rejecting server-filled resource defaults by adding shmem to the image memory minimum during default fill, matching the validator's accounting.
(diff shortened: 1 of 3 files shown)


ironAiken2 (approved):
LGTM

PR #11505, opened 2026-05-07
fix(BA-5967): make ResourceSlot.to_humanized() resilient to unknown slots

fregataa:
## Summary
- Resolve each slot independently in `ResourceSlot.to_humanized()` via per-slot `try/except KeyError`, falling back to `_guess_slot_type()` for keys missing from the supplied `slot_types` map.
- Eliminates a class of 500 Internal Server Errors triggered when error-message humanizers (e.g. `ResourceLimitRule._humanize_slots`) ran against partially-populated slot type metadata — a missing display-metadata entry no longer clobbers the underlying user-facing 4xx.

## Test plan
- [x] New unit tests in `tests/unit/common/test_types.py` cover the per-slot fallback and the empty-`slot_types` case.
- [x] `pants fmt`, `pants fix`, `pants lint --changed-since=origin/main` pass locally.
- [x] Existing callers (`ResourceLimitRule._humanize_slots`, scheduler predicates) continue to work — verified `tests/unit/common/test_types.py` and `tests/unit/manager/sokovan/scheduling_controller/validators/test_session_spec_rules.py`.

Resolves BA-5967

### DIFF
--- a/src/ai/backend/common/types.py
+++ b/src/ai/backend/common/types.py
@@ -1223,14 +1223,17 @@ class ResourceSlot(UserDict[str, Decimal]):
         return cls(data)
 
     def to_humanized(self, slot_types: Mapping[str, Any]) -> Mapping[str, str]:
-        try:
-            return {
-                k: type(self)._humanize_value(Decimal(v), slot_types[k])
-                for k, v in self.data.items()
-                if v is not None
-            }
-        except KeyError as e:
-            raise ValueError(f"Unknown slot type: {e.args[0]!r}") from e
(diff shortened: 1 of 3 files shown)


src/ai/backend/common/types.py
```diff
@@ -1223,14 +1223,17 @@ def from_user_input(
         return cls(data)
 
     def to_humanized(self, slot_types: Mapping[str, Any]) -> Mapping[str, str]:
```
jopemachine, line 1225: Could we change the parameter type to `Mapping[SlotName, SlotTypes]`?

PR #11507, opened 2026-05-07
fix(BA-5974): remove deployment name uniqueness constraint

fregataa:
## Summary
- Drop the partial unique index `ix_endpoints_unique_name_when_active` on `(name, domain, project)` and the application-level `_check_endpoint_name_exists` pre-check, so deployment names are no longer unique within a project.
- The previous constraint scope (`domain, project`) did not match the `my_deployments` list scope (`created_user`), so a deployment created by another user in the same project blocked the create yet stayed invisible to the caller — leaving the user unable to choose the name and unable to locate the conflicting record.
- Remove the now-unused `DeploymentNameAlreadyExists` exception and the duplicate-name test. Routing keys on `route_id`, and the inference session name template `f"{endpoint.name}-{route_id}"` already disambiguates downstream.

## Test plan
- [x] Apply the new alembic migration on a database that holds the existing index and verify `\d endpoints` no longer lists `ix_endpoints_unique_name_when_active`.
- [x] Create two deployments with the same name in the same project (different and same `created_user`) and confirm both succeed.
- [x] Confirm `my_deployments` and project-scoped list queries return both rows as expected.
- [x] Downgrade the migration and verify the index is recreated with predicate `lifecycle_stage NOT IN ('destroying', 'destroyed')`.

Resolves BA-5974

### DIFF
--- a/changes/11507.fix.md
+++ b/changes/11507.fix.md
@@ -1,0 +1,1 @@
+Allow deployment names to be reused within a project so a hidden record from another user no longer blocks creation.
--- a/src/ai/backend/common/exception.py
+++ b/src/ai/backend/common/exception.py
@@ -818,18 +818,6 @@ class ModelRevisionNotFound(BackendAIError, web.HTTPNotFound):
         )
 
 
-class DeploymentNameAlreadyExists(BackendAIError, web.HTTPConflict):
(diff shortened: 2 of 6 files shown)


jopemachine:
Would it be better to remove the unique constraint entirely rather than applying it per scope?

seedspirit (approved):

PR #11509, opened 2026-05-07
refactor(BA-5976): generalize filter DTOs and adapter boolean-clause conversion

fregataa:
## Summary

- Centralize the recursive `AND` / `OR` / `NOT` clauses on a shared `BaseFilter` in `common/dto/manager/v2/common.py` and have every v2 entity filter (38 classes across 31 files) inherit from it, eliminating the per-class duplication and the matching `model_rebuild()` calls.
- Add `convert_and` / `convert_or` / `convert_not` to `BaseFilterAdapter`. The OR helper AND-groups each non-empty sub-filter before OR-combining (`(A AND B) OR (C AND D)` rather than `A OR B OR C OR D`); the NOT helper negates each sub independently.
- Migrate all 25 adapters with AND/OR/NOT handling to the helpers — this **also retroactively applies the BA-5975 grouping fix to the 24 adapters where it had not yet been ported**.
- Add unit tests for the three helpers under `tests/unit/manager/api/adapters/test_filter_helper.py`.

Stacked on #11506 (BA-5975); the three BA-5975 commits will fall out of this PR's range once that one merges.

## Test plan

- [ ] `pants check --changed-since=origin/main`
- [ ] `pants test tests/unit/manager/api/adapters/::`
- [ ] CI green
- [ ] Spot-check a non-deployment adapter (e.g., user, vfolder) end-to-end with `OR=[{a,b}, {c,d}]` to confirm the BA-5975 grouping fix landed correctly

Resolves BA-5976


fregataa:
Closed this PR since it is not planned

PR #11512, opened 2026-05-08
fix: base modify_endpoint on latest revision instead of current

HyeockJinKim:
## Summary
- Switch the legacy `modify_endpoint` flow to merge user overrides on top of the **latest** revision (highest `revision_number`) rather than the **current/serving** revision. Adds `DeploymentRepository.get_latest_revision()` mirroring `get_current_revision()`.
- Fixes a real failure mode during initial deployment: `current_revision` is `NULL` until the first rollout's `deploying_revision → current_revision` swap completes, so any modify issued in that window previously raised `DeploymentRevisionNotFound`. With this change, modify uses the latest revision row and works through that window.
- Also drops the orphan `read_for_legacy_model_service_revision` reader path along with `load_legacy_model_service_revision_read_bundle` and `LegacyRevisionModifyReadBundle` (zero callers; would have applied a different merge precedence than the live service path).

## Test plan
- [x] `pants check --changed-since=HEAD~1 --changed-dependents=transitive` — mypy clean over 2945 files
- [x] `pants test --changed-since=HEAD~1` — 6 affected test modules pass
- [x] New repository unit tests added (and pre-existing fixture gap fixed by including `KeyPairRow` / `KeyPairResourcePolicyRow`):
  - `get_latest_revision` returns the highest `revision_number`
  - `get_latest_revision` ignores `EndpointRow.current_revision` (latest can differ from current)
  - `get_latest_revision` raises `EndpointNotFound` when no revisions exist
- [x] Live verification on local manager (after applying pending alembic head `46e007d9b237`):
  - On an endpoint with `current_revision = NULL` and one revision row, `ModifyEndpoint` (legacy GQL mutation) succeeds and creates v2 layered on v1 (`environ` override applied) — the exact case that previously raised `DeploymentRevisionNotFound`.
  - Second `ModifyEndpoint` creates v3 with `environ = {"BAZ":"QUX","FOO":"BAR"}`, confirming v3 is built on **latest=v2** so the prior modify's changes are preserved (under the old `current` base, `FOO=BAR` would have been dropped).


src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -2589,6 +2571,26 @@ async def get_current_revision_spec(
...
+
+        Unlike :meth:`get_current_revision`, this does not consult
+        ``EndpointRow.current_revision``: it returns the most recently
+        created revision for the endpoint regardless of activation state.
+
+        Args:
+            endpoint_id: ID of the deployment endpoint
+
+        Raises:
+            EndpointNotFound: If no revisions exist for the endpoint.
```
seedspirit: Is it okay to raise `EndpointNotFound` Error here? I think No available revision error or such errors will be more adequate

PR #11514, opened 2026-05-08
refactor(BA-5978): introduce BackendAISchema for per-domain validation errors

jopemachine:
## Stacked PRs

This PR is part of a stack:

1. **#11514 — `fix/BA-5978-pydantic-validation-domain-errors`** ← *this PR*
   Foundational machinery: `BackendAISchema`, `BackendAISchemaValidationFailed`, `build_validation_error` override hook, and the four model overrides (`ModelDefinition`, `ModelDefinitionDraft`, `SessionSpec`, `DeploymentConfigInput`). Also promotes the foundational base classes so anything that already inherits them participates automatically.
2. **#11554 — `feat/BA-6003-bulk-basemodel-migration`** *(stacked on top)*
   Mechanical migration: every remaining `class Foo(BaseModel)` subclass under `src/ai/backend/` switched to `BackendAISchema`, the matching `except ValidationError` → `except BackendAISchemaValidationFailed` swaps, and the seven CLI `_build_dto` helpers routed through `model_validate`. Merge *after* this PR.

---

## Summary

Make Pydantic validation failures surface as a domain-specific `BackendAIError` (HTTP 400) instead of a raw `pydantic.ValidationError` (which leaks out as 500). The mapping is driven by the *model class* itself — not by repeated `try / except` re-wraps at every call site.

Resolves [BA-5978](https://lablup.atlassian.net/browse/BA-5978).

## Approach

Introduce `BackendAISchema`, a project-wide Pydantic base whose `model_validate` / `model_validate_json` classmethods catch `pydantic.ValidationError` and call an overridable `build_validation_error()` classmethod to produce the `BackendAIError` instance to raise.

```python
class BackendAISchema(BaseModel):
    @classmethod
    def build_validation_error(cls, info: SchemaValidationFailureInfo) -> BackendAIError:
        # Default — generic 400.
        return BackendAISchemaValidationFailed(
            extra_msg=info.summary,
            extra_data={"errors": info.errors},
        )
    ...  # validate classmethods call cls.build_validation_error(info)
```

Subclasses override the classmethod to inject their own 400, e.g.:

```python
class SessionSpec(_SpecBaseModel):
    @override
    @classmethod
    def build_validation_error(cls, info: SchemaValidationFailureInfo) -> BackendAIError:
        missing = [cls._format_loc(tuple(err["loc"])) for err in info.errors]
        return IncompleteSessionSpec(
            extra_msg=f"SessionSpec fields not resolved: {', '.join(missing)}",
            extra_data={"missing": missing},
        )
```

`SchemaValidationFailureInfo` is a `@dataclass(frozen=True)` of `summary: str` + `errors: list[pydantic_core.ErrorDetails]`. Subclasses see only this stable shape — no dependency on pydantic's `ValidationError` API.

### Naming note (`Schema`, not `Model`)

Originally named `BackendAIModel` / `BackendAIModelValidationFailed` / `ModelValidationFailureInfo`. Renamed to `BackendAISchema` / `BackendAISchemaValidationFailed` / `SchemaValidationFailureInfo` to avoid confusion with LLM-model terminology in client-facing surfaces. The exception's `error_type` URL and `error_title` were updated to match (`schema-validation-failed`).

### Draft → strict via `model_validate`

`ModelHealthCheckDraft.to_resolved()`, `ModelServiceConfigDraft.to_resolved()`, `ModelConfigDraft.to_resolved()`, and `ModelDefinitionDraft.to_resolved()` now route through `model_validate(self.model_dump(exclude_none=True, ...))` against their strict counterparts. Manual `if self.x is None: raise ValueError(...)` checks are dropped — the strict type's required-field declarations are the single source of truth, and failures surface as the override's domain-specific 400 (e.g. `ModelDefinitionValidationError`).

### Constructor path

`__init__` is intentionally **not** overridden. Pydantic v2 invokes nested models' `__init__` from inside the outer validator, and converting `ValidationError` to a non-`ValidationError` there breaks `loc`-path aggregation (outer field prefix is lost, fail-fast aborts sibling errors). Direct `Model(field=...)` construction therefore still raises stock `pydantic.ValidationError`; switch call sites to `Model.model_validate({...})` to opt into the override path. See `scripts/probe_loc_aggregation.py` for the empirical demonstration.

## What lands in this PR

* **`BackendAISchema`** (`common/types.py`) — Pydantic base with the overridable hook.
* **`BackendAISchemaValidationFailed`** (`common/exception.py`) — the default HTTP 400 raised when the override isn't customized.
* **`SchemaValidationFailureInfo`** (`common/types.py`) — dataclass passed to the override.
* **Foundational base classes promoted to `BackendAISchema`** so anything that already inherits one of them gets the auto-conversion for free:
  `BaseRequestModel`, `BaseFieldModel`, `BaseResponseModel`, `BaseConfigSchema`, `BaseConfigModel` (in `common/`), `_OptionsBaseModel`, `_DraftBaseModel`, `_SpecBaseModel`, `BaseAgentRequestModel`, `BaseAgentResponseModel`, plus the `test/tester` variants and `ai.backend.logging`'s `BaseConfigModel` / `BaseConfigValidationContext`.
* **Concrete overrides applied** at the four models that previously needed call-site `try / except` wraps:

  | Model | `build_validation_error` raises |
  |---|---|
  | `ModelDefinition` (`common/config.py`) | `ModelDefinitionValidationError` |
  | `ModelDefinitionDraft` (`common/config.py`) | `ModelDefinitionValidationError` |
  | `SessionSpec` (`manager/data/session/spec.py`) | `IncompleteSessionSpec` |
  | `DeploymentConfigInput` (`manager/repositories/deployment/storage_source/storage_source.py`) | `InvalidAPIParameters("Invalid deployment config: …")` |

* **`ModelDefinitionValidationError` moved** from `agent/errors/agent.py` to `common/exception.py` so the model — which lives in `common.config` — can construct it without an upward-layer import. (Agent's `__init__.py` re-export and the agent-flavored `/agent/...` `error_type` URL segment go away with the move.)
* **Draft → strict refactor** for `ModelHealthCheckDraft`, `ModelServiceConfigDraft`, `ModelConfigDraft`, `ModelDefinitionDraft` (see "Draft → strict via `model_validate`" above).
* **Call-site `try / except` wraps removed** at:
  * `agent/agent.py` (was: `BackendAISchemaValidationFailed → ModelDefinitionValidationError`)
  * `manager/services/model_card/service.py` (was: `→ ModelCardParseError` for the ModelDefinition validate)
  * `manager/sokovan/scheduling_controller/preparers/session_spec_preparer.py` (was: `→ IncompleteSessionSpec`)
  * `manager/repositories/deployment/storage_source/storage_source.py` (was: `→ InvalidAPIParameters` at two sites)

## Notes

* `PydanticJWTValidator.validate` in `common/typed_validators.py` keeps its existing `try / except → JWTPayloadValidationError` — the validator is generic over any `BackendAISchema` subclass and the JWT context is preserved at the validator level, not the model level. (And `PydanticJWTValidator` itself currently has only one caller.)
* Bulk mechanical migration of every remaining direct `class Foo(BaseModel)` subclass (~120 classes) to `BackendAISchema`, along with the matching `except ValidationError` swaps, is split into a stacked follow-up PR: **#11554 / [BA-6003](https://lablup.atlassian.net/browse/BA-6003)**. This PR only includes the foundational machinery and the four model overrides above.

## Test plan

- [x] `pants lint --changed-since=…`
- [x] `pants check --changed-since=…`
- [x] `tests/unit/manager/sokovan/scheduling_controller/preparers/test_session_spec_preparer.py` (still asserts `IncompleteSessionSpec` thanks to the `SessionSpec` override)
- [x] `tests/unit/manager/models/scaling_group/test_row.py` (asserts `BackendAISchemaValidationFailed` from the default override)
- [x] Manual smoke: `POST /v2/domains` with `{}` against the live mgr — confirmed `BackendAISchema.model_validate` raises `BackendAISchemaValidationFailed` with structured `extra_data.errors`, `BodyParam` then rewraps as `InvalidAPIParameters` (4xx) at the HTTP boundary. Log line includes the per-field `loc`/`type`/`msg` info.

[BA-5978]: https://lablup.atlassian.net/browse/BA-5978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


changes/11514.fix.md
```diff
@@ -0,0 +1 @@
+Convert Pydantic validation failures at service-layer call sites into domain-specific 400 errors with structured field-level messages.
```
jopemachine: Rename this file

tests/unit/common/test_exception.py
```diff
@@ -111,3 +138,121 @@ async def error_handler(_request: web.Request) -> web.Response:
...
+            # Integer index uses bracket notation
+            (("tags", 0), None, "tags[0]"),
+            # Mix of nested fields and index
+            (("users", 2, "email"), None, "users[2].email"),
+            # Prefix is prepended
+            (("name",), "body", "body.name"),
+            (("address", "city"), "body", "body.address.city"),
+            (("tags", 0), "body", "body.tags[0]"),
+            # Leading int with prefix still uses bracket notation
+            ((0, "name"), "body", "body[0].name"),
```
jopemachine: Create a type for readibility

src/ai/backend/common/exception.py
```diff
@@ -406,6 +410,14 @@ def error_code(self) -> ErrorCode:
...
         raise NotImplementedError("Subclasses must implement error_code() method.")
 
+    @classmethod
+    def from_pydantic(cls, exc: ValidationError) -> Self:
+        """Build an instance from a Pydantic ``ValidationError``,
+        surfacing the original multi-line message as ``extra_msg`` and
+        the raw per-field error list (``loc``/``msg``/``type``/``input``/
+        ``url``) as ``extra_data["errors"]`` — no extra reformatting."""
+        return cls(extra_msg=str(exc), extra_data={"errors": exc.errors()})
+
```
HyeockJinKim: Why is this included in `BackendAIError`? There's no reason to include it in `BackendAIError`.

HyeockJinKim: The name "from_pydantic" is also an error value, which feels a bit awkward.

jopemachine: This PR is not ready for PR review

src/ai/backend/common/exception.py
```diff
@@ -10,6 +11,8 @@
 
 from .json import dump_json
 
+log = logging.getLogger(__spec__.name)
```
jopemachine: Why is it added?

src/ai/backend/common/exception.py
```diff
@@ -443,6 +444,51 @@ def error_code(self) -> ErrorCode:
...
+
+    error_type = "https://api.backend.ai/probs/model-definition-validation-failed"
+    error_title = "Model definition validation failed."
+
+    def error_code(self) -> ErrorCode:
+        return ErrorCode(
+            domain=ErrorDomain.MODEL_SERVICE,
+            operation=ErrorOperation.ACCESS,
+            error_detail=ErrorDetail.INVALID_PARAMETERS,
+        )
```
HyeockJinKim, line 489: Why is this in Access?

HyeockJinKim (dismissed):

HyeockJinKim (approved):

PR #11515, opened 2026-05-08
fix(BA-5967): pre-validate slot types in sokovan validator chain

fregataa:
## Summary
- Adds `ImageSlotTypeRule` and `RequestedSlotTypeRule` to the Sokovan SessionSpec validator chain so slot keys absent from the cluster's `known_slot_types` are caught with a clear 4xx **before** reaching `ResourceLimitRule`.
- Without this, `ResourceLimitRule` calls `ResourceSlot.to_humanized()` to format an error message, and a missing slot key turns the original 4xx into a 500 (the bug reported in BA-5967).
- Both rules no-op when `known_slot_types` is empty, matching the existing raw-fallback path that `ResourceLimitRule._humanize_slots` uses on freshly bootstrapped clusters.

## Test plan
- [x] `pants test tests/unit/manager/sokovan/scheduling_controller/validators/test_session_spec_rules.py` — new `TestImageSlotTypeRule` and `TestRequestedSlotTypeRule` cover known-slot pass, mismatched-slot rejection, and empty-known-slot-types skip.
- [x] `pants fmt`/`fix`/`lint`/`check` on changed files.

### DIFF
--- a/src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
@@ -6,6 +6,7 @@ import logging
 from collections import defaultdict
 from collections.abc import AsyncIterator, Mapping, Sequence
 from contextlib import asynccontextmanager as actxmgr
+from dataclasses import dataclass
 from datetime import datetime
 from decimal import Decimal
 from typing import TYPE_CHECKING, Any, cast
@@ -187,6 +188,21 @@ def _create_resource_slot_from_policy(
     return ResourceSlot.from_policy(resource_policy_map, cast(Mapping[str, Any], known_slot_types))
 
 
+@dataclass(frozen=True)
+class _ScalingGroupWithSlotInventory:
+    """Scaling group bundled with the slot inventory served by its agents.
+
+    ``active_slot_types`` maps each slot name served by a non-terminated
+    agent in this scaling group to its registered :class:`SlotTypes`
+    unit. The validator chain consults this map both for membership
+    (reject requests for slots the RG does not provide) and for unit
+    metadata (humanize values during error formatting).
+    """
+
+    sg_row: ScalingGroupRow
+    active_slot_types: Mapping[SlotName, SlotTypes]
+
+
 class ScheduleDBSource:
     """
     Database source for schedule-related operations.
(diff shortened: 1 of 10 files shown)


src/ai/backend/manager/sokovan/scheduling_controller/validators/session_spec_base.py
```diff
@@ -57,6 +57,7 @@ class SessionSpecValidationContext:
     keypair_resource_policy: KeyPairResourcePolicyData | None = None
     image_infos: Mapping[ImageID, ImageInfo] = field(default_factory=dict)
     known_slot_types: Mapping[SlotName, SlotTypes] = field(default_factory=dict)
```
jopemachine, line 59: Is known_slot_types fetched from etcd still being used?
Are we checking both the DB values and etcd values?

src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
```diff
@@ -187,6 +188,19 @@ def _create_resource_slot_from_policy(
...
+class _ScalingGroupWithAgentResources:
+    """Scaling group bundled with its non-terminated agents.
+
+    Each :class:`AgentRow` in ``non_terminated_agents`` carries an
+    eagerly-loaded ``agent_resource_rows`` list, so the caller can derive
+    per-RG slot inventory inline without an additional round-trip.
+    """
+
+    sg_row: ScalingGroupRow
+    non_terminated_agents: tuple[AgentRow, ...]
```
seedspirit: It seems better to return the data rather than the row.

jopemachine: I think it seems awkward to enforce the use of a Data type even within the db_source internals.

src/ai/backend/manager/sokovan/scheduling_controller/validators/session_spec_base.py
fregataa: I created a separate issue to replace etcd-sourced slot types query #11532. This branch job is adding a pre-validator

src/ai/backend/manager/sokovan/scheduling_controller/validators/requested_slot_type_rule.py
```diff
@@ -0,0 +1,60 @@
...
+        for idx, kernel in enumerate(spec.kernel_specs):
+            unknown = sorted({
+                entry.resource_type
+                for entry in kernel.execution_spec.resources
+                if entry.resource_type not in rg_slot_types
+            })
+            if unknown:
+                raise InvalidAPIParameters(
+                    extra_msg=(
+                        f"kernel_specs[{idx}]: the request asks for resource "
```
jopemachine, line 54: Why not including kernel row id instead of kernel_specs[{idx}]?

jopemachine (approved):

Issue #11520, opened 2026-05-08
Legacy `mounts` field rejects UUID-shaped strings (BA-5916 follow-up)


jopemachine:
## Summary

The deprecated creation_config["mounts"] legacy field on REST v1 POST /session/_/create and /_/create-from-template is currently name-only. Passing a UUID-shaped string (e.g. via the v1 CLI -v <vfolder-uuid>:/dst) raises VFolderNotFound because _resolve_legacy_name_mounts treats every entry as a vfolder NAME and routes it through resolve_vfolder_ids_by_names. Before the sokovan refactor (commit 8321c79aa^), the old _build_mount_requests_from_spec accepted both names and UUIDs in the legacy mounts list.

## Reproduction

- POST /session/_/create with config={"mounts": ["<vfolder-uuid>"]} → VFolderNotFound.
- ./backend.ai session create -v <vfolder-uuid>:/dst hits the same path because the v1 CLI puts every -v source into the legacy mounts field without a UUID/name split.

## Root cause

- src/ai/backend/manager/api/rest/session/handler.py:_resolve_legacy_name_mounts adds every mounts / mount_map key to names_to_resolve — UUID-shaped strings end up in the name lookup.
- _merge_resolved_legacy_mounts only routes name-resolved UUIDs into mount_ids; UUID-string inputs that skip the resolver are silently dropped.

## Acceptance Criteria

- _resolve_legacy_name_mounts skips UUID-shaped keys so they bypass the name resolver.
- _merge_resolved_legacy_mounts routes UUID-shaped entries from mounts / mount_map / mount_options into the UUID-keyed mount_ids / mount_id_map / mount_options buckets.
- Existing name-resolution behaviour (BA-5916) is preserved for non-UUID entries.
- Subpath syntax (name/subdir) remains rejected — full subpath support requires a separate, larger change that plumbs vfsubpath through MountInfoEntry / VFolderMountRequest.
- Unit tests cover: bare name, bare UUID-string, mixed name+UUID, name+subpath rejection.

JIRA Issue: BA-5980

PR #11527, opened 2026-05-08
feat: Restore per-mount source-subpath on the UUID-keyed mount contract

### LINKED ISSUE #11526: Restore per-mount source-subpath on the UUID-keyed mount contract
## Summary

The new UUID-keyed mount contract introduced in #11434 (and refined in #11520) carries no per-mount source-subpath. There is currently no way to express **"mount the `<X>` subdirectory of vfolder `<Y>` at container path `<Z>`"** through any session-creation API surface, even though the lower-level mount machinery still supports it.

#11520 acknowledges this deferral explicitly:

> Subpath syntax (`name/subdir`) remains rejected — full subpath support requires a separate, larger change that plumbs `vfsubpath` through `MountInfoEntry` / `VFolderMountRequest`.

This issue is that follow-up.

## Where the capability disappeared

- `src/ai/backend/manager/api/rest/session/handler.py:327-333` — rejects every `mounts` entry containing `/`.
- `src/ai/backend/common/types.py:655-684` — `MountInfoEntry` carries only `vfolder_id`, `mount_destination`, `mount_perm`. No source-subpath slot.
- `src/ai/backend/manager/registry.py:_mount_entries_from_creation_config` — projects `creation_config["mount_ids"]` / `mount_id_map` / `mount_options` into `MountInfoEntry` tuples; no slot to populate even if the wire format gained one.
- `src/ai/backend/manager/repositories/scheduler/db_source/db_source.py:1572-1580` — reconstitutes `VFolderMountRequest(ref=UUID(...))` from `MountInfoEntry`. The string-form `ref` (which `prepare_vfolder_mounts` knows how to split into name + subpath) is no longer reachable from this path.

## Where the capability still exists, just unreachable

`src/ai/backend/manager/models/vfolder/row.py:925-940` (`prepare_vfolder_mounts`) still handles `name/subpath` strings:

```python
for req in mount_requests:
    key = req.ref
    if isinstance(key, uuid.UUID):
        requested_vfolder_ids.add(key)
        continue
    name, _, subpath = key.partition("/")
    requested_vfolder_subpaths[key] = os.path.normpath(subpath)
```

When `ref` is a `UUID`, the subpath is forced to `"."` (mount root). The infrastructure to mount a subpath is intact — only the API-level addressability was removed.

## Impact

Any caller that previously mounted a subdirectory of an existing vfolder at a specific container destination has to either:
1. Replace each subdir with its own dedicated vfolder (costly migration; changes the user-visible vfolder layout).
2. Mount the whole parent vfolder and live with deeper container-visible paths than before.

Neither preserves backwards compatibility for callers that built on the previous behavior. The pattern was common — typical layouts include `<parent>/.shared/`, `<parent>/.scratch/<task_id>/`, `<parent>/inputs/...` mounted at distinct container paths. With #11434's contract there is no way to express these without rearchitecting the parent vfolder into N separate vfolders per workflow.

## Proposed fix

Re-introduce a per-mount source subpath on the contract:

1. Add `source_subpath: str | None = None` to `MountInfoEntry` (or equivalent name; carries the subdir relative to the vfolder root).
2. Plumb it through `_mount_entries_from_creation_config`. Wire it via either:
   - A new `mount_id_subpaths: Mapping[UUID-str, str]` field on `creation_config`, or
   - A richer dict shape inside `mount_id_map` (e.g. `{uuid: {"dst": "/x", "subpath": ".shared"}}`).
   The first form is simpler and avoids reshaping an existing field.
3. Pass `source_subpath` through to `VFolderMountRequest` (which already accepts `str` `ref` with a `/` and parses it). One option: when `source_subpath` is set, build the `ref` as `f"{name}/{source_subpath}"` against the resolved vfolder name. Another: extend `VFolderMountRequest` directly with a `subpath: str | None` field. The latter is cleaner since it removes the string-format ambiguity and makes the data path explicit.
4. Soften the rejection at `_resolve_legacy_name_mounts:327-333` so legacy `<name>/<subpath>` entries on the deprecated `mounts` field are accepted again, resolved to `(uuid, subpath)`, and forwarded through the new field. (This is optional but lets pre-existing v1 callers continue working without rewrite.)

## Acceptance criteria

- `MountInfoEntry` (or its equivalent) has a `source_subpath` field; `None` means "mount the vfolder root" (current behavior).
- `POST /session/_/create` with a UUID-keyed `mount_ids` plus a per-mount `source_subpath` results in a session where the container destination is bind-mounted to `<vfolder>/<source_subpath>` rather than the vfolder root.
- The legacy `mounts: ["<name>/<subpath>"]` form on the same endpoint stops raising `InvalidAPIParameters(400)` and produces an equivalent mount.
- `prepare_vfolder_mounts` is exercised end-to-end with a non-`"."` subpath via the modern UUID-keyed surface (existing tests cover the legacy string surface).
- No change to `MountPermission`, mount destination defaulting, or storage-proxy `vfsubpath` semantics.

## Related

- #11434 — original contract change (removed the legacy subpath addressability).
- #11520 — closed follow-up that explicitly deferred this work.
- #6661 — sibling request for subfolder mounts in the deployment / service path; the proposed fix here would close most of that as a side effect.

rapsealk:
## Summary

Closes #11526.

The UUID-keyed mount contract introduced in #11434 / #11520 dropped the per-mount source-subpath, leaving callers no way to express "mount the `<X>` subdirectory of vfolder `<Y>`" through any session-creation surface. This PR restores that capability end-to-end:

- `MountInfoEntry` gains a `source_subpath: str | None = None` field. `None` (the default) keeps the historical "mount the vfolder root" behavior, and pre-existing persisted rows deserialize cleanly through Pydantic's default.
- `VFolderMountRequest` gains a parallel `subpath: str | None = None` so the typed UUID-keyed surface can express a subpath without falling back to the legacy string-form `ref="name/subpath"`.
- `CreationConfigV7` exposes `mount_id_subpaths: dict[UUID, str] | None` (with a `mountIdSubpaths` camelCase alias). `AgentRegistry._mount_entries_from_creation_config` reads it with the same UUID/UUID-string-key polymorphism already used for `mount_id_map` / `mount_options` and projects the value onto each emitted `MountInfoEntry.source_subpath`.
- The scheduler reconstruction in `SchedulerDBSource` rebuilds `VFolderMountRequest` with the persisted subpath, and `prepare_vfolder_mounts` honors a UUID-keyed request's explicit subpath instead of unconditionally forcing `requested_vfolder_subpaths[uuid] = "."`.
- The REST handler's blanket rejection of `mounts=["name/subdir"]` is replaced with split-and-resolve: the bare name resolves to a UUID via the existing resolver, the subpath is captured into a new `name -> subpath` map, and `_merge_resolved_legacy_mounts` forwards it onto `mount_id_subpaths`.

## Acceptance criteria

- [x] `MountInfoEntry` (or its equivalent) has a `source_subpath` field; `None` means "mount the vfolder root" (current behavior).
- [x] `POST /session/_/create` with a UUID-keyed `mount_ids` plus a per-mount `source_subpath` results in a session where the container destination is bind-mounted to `<vfolder>/<source_subpath>` rather than the vfolder root. (Wire field `mount_id_subpaths` plumbed through to `MountInfoEntry.source_subpath`, which feeds `VFolderMountRequest.subpath` and into `prepare_vfolder_mounts.requested_vfolder_subpaths`.)
- [x] The legacy `mounts: ["<name>/<subpath>"]` form on the same endpoint stops raising `InvalidAPIParameters(400)` and produces an equivalent mount.
- [x] `prepare_vfolder_mounts` is exercised end-to-end with a non-`"."` subpath via the modern UUID-keyed surface (covered by registry-projection unit tests; the lower-layer subpath path is unchanged for the legacy string surface).
- [x] No change to `MountPermission`, mount destination defaulting, or storage-proxy `vfsubpath` semantics.

## Test plan

- [x] `pants fmt --changed-since=origin/main`
- [x] `pants fix --changed-since=origin/main`
- [x] `pants check --changed-since=origin/main`
- [x] `pants test --changed-since=origin/main`
- [x] `pants test tests/unit/common/dto/manager/session/ tests/unit/manager/services/session/ tests/unit/manager/sokovan/scheduling_controller/test_enqueue_session_from_draft.py`
- [ ] Manual end-to-end verification with a live server (subpath mount via `mount_id_subpaths` and via `mounts=["name/subdir"]`) — defer until SDK ergonomics land.

## Out of scope

- Client/SDK changes to expose `mount_id_subpaths` from `ComputeSession.get_or_create`. The wire format is enough for now; SDK plumbing will follow in a separate PR.
- Storage-proxy changes — `vfsubpath` is already supported there.

### DIFF
--- a/src/ai/backend/common/dto/manager/session/types.py
+++ b/src/ai/backend/common/dto/manager/session/types.py
@@ -431,6 +431,16 @@ class CreationConfigV7(BaseFieldModel):
         default=None,
         validation_alias=AliasChoices("mount_id_map", "mountIdMap"),
     )
+    mount_id_subpaths: dict[UUID, str] | None = Field(
+        default=None,
+        validation_alias=AliasChoices("mount_id_subpaths", "mountIdSubpaths"),
+        description=(
+            "Per-mount source-subpath keyed by vfolder UUID. ``None`` (the default) "
+            "and missing entries mean 'mount the vfolder root'; a non-empty value "
+            "mounts that subdirectory of the vfolder at the corresponding "
+            "``mount_id_map`` destination."
+        ),
+    )
     mount_options: dict[str, MountOption] | None = Field(
         default=None,
         validation_alias=AliasChoices("mount_options", "mountOptions"),
--- a/src/ai/backend/manager/api/rest/session/handler.py
+++ b/src/ai/backend/manager/api/rest/session/handler.py
@@ -328,9 +328,7 @@ def _merge_resolved_legacy_mounts(
         except (ValueError, TypeError):
             continue
     merged_mount_id_map: dict[Any, str] = dict(creation_config.get("mount_id_map") or {})
-    merged_mount_id_subpaths: dict[Any, str] = dict(
-        creation_config.get("mount_id_subpaths") or {}
-    )
+    merged_mount_id_subpaths: dict[Any, str] = dict(creation_config.get("mount_id_subpaths") or {})
     merged_mount_options: dict[Any, Any] = dict(creation_config.get("mount_options") or {})
     legacy_mount_map = creation_config.get("mount_map") or {}
     legacy_mount_options = creation_config.get("mount_options") or {}
--- a/src/ai/backend/manager/api/rest/session/handler.py
+++ b/src/ai/backend/manager/api/rest/session/handler.py
@@ -282,9 +282,25 @@ def _route_legacy_uuid_mounts(creation_config: dict[str, Any]) -> dict[str, Any]
     return next_config
 
 
+def _split_legacy_name_subpath(raw: str) -> tuple[str, str | None]:
+    """Split a legacy ``mounts``-style entry of the form ``"name"`` or
+    ``"name/subpath"`` into a ``(name, subpath)`` pair.
+
+    Returns ``subpath=None`` when the entry has no ``/``; a non-empty
+    string otherwise (the ``name/`` form with an empty trailing segment
+    is treated as no-subpath, mirroring the historical behavior of
+    :func:`prepare_vfolder_mounts`'s string-form ``ref`` parser).
+    """
+    name, sep, subpath = raw.partition("/")
+    if not sep or not subpath:
+        return name, None
+    return name, subpath
+
+
 def _merge_resolved_legacy_mounts(
     creation_config: dict[str, Any],
     name_to_id: dict[str, UUID],
+    name_to_subpath: Mapping[str, str] | None = None,
 ) -> dict[str, Any]:
     """Merge a ``name → UUID`` resolution into the UUID-keyed buckets of
     ``creation_config`` and drop the now-resolved name-keyed legacy keys.
@@ -293,9 +309,16 @@ def _merge_resolved_legacy_mounts(
     and entries from the name-keyed ``mount_map`` / ``mount_options``
     dicts are re-keyed onto the resolved UUIDs without overwriting an
     explicit UUID-keyed entry the caller already supplied.
+
+    ``name_to_subpath`` carries any per-name source-subpath extracted from
+    legacy ``mounts=["name/subdir"]`` entries; the resolved subpath is
+    forwarded onto ``mount_id_subpaths`` keyed by the corresponding UUID,
+    again without overwriting an existing UUID-keyed value.
     """
     if not name_to_id:
         return creation_config
+
+    name_to_subpath = name_to_subpath or {}
 
     merged_mount_ids: list[Any] = list(creation_config.get("mount_ids") or [])
     existing_uuid_set: set[UUID] = set()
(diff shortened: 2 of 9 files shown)


src/ai/backend/common/dto/manager/session/types.py
```diff
@@ -431,6 +431,16 @@ class CreationConfigV7(BaseFieldModel):
...
+    mount_id_subpaths: dict[UUID, str] | None = Field(
+        default=None,
+        validation_alias=AliasChoices("mount_id_subpaths", "mountIdSubpaths"),
+        description=(
+            "Per-mount source-subpath keyed by vfolder UUID. ``None`` (the default) "
+            "and missing entries mean 'mount the vfolder root'; a non-empty value "
+            "mounts that subdirectory of the vfolder at the corresponding "
+            "``mount_id_map`` destination."
+        ),
+    )
```
HyeockJinKim, line 443: I don't see any reason to add this to the schema. I don't plan to expand this API any further.

src/ai/backend/manager/api/rest/session/handler.py
```diff
@@ -282,9 +282,25 @@ def _route_legacy_uuid_mounts(creation_config: dict[str, Any]) -> dict[str, Any]
...
+    name, sep, subpath = raw.partition("/")
+    if not sep or not subpath:
+        return name, None
+    return name, subpath
+
+
 def _merge_resolved_legacy_mounts(
     creation_config: dict[str, Any],
     name_to_id: dict[str, UUID],
+    name_to_subpath: Mapping[str, str] | None = None,
```
HyeockJinKim, line 303: It seems odd to keep expanding the map by name.

src/ai/backend/common/dto/manager/session/types.py
rapsealk: As we dropped support for folder names and now only support folder IDs, using a path like `b64bb966-3d81-45b7-8e78-1cb1a7a144b0/path/to/somewhere` feels unnatural.

rapsealk:
Superseded by #11582.

PR #11534, opened 2026-05-09
feat(BA-5985): scan routes via BatchQuerier with traffic-status filter

fregataa:
## Summary
- Replaces `DeploymentRepository.get_routes_by_statuses` with `search_route_datas(*, querier: BatchQuerier)` so route coordinator scans funnel through the standard `execute_batch_querier` plumbing.
- `RouteTargetStatuses` gains a `traffic: list[RouteTrafficStatus] | None` axis; handlers that set it constrain the scan to listed traffic statuses, others pass `None` and the predicate is skipped.
- Both coordinator call sites (lifecycle handler dispatch and observer) build conditions explicitly. Pagination is `NoPagination` — bounded scans are left for a follow-up.

## Test plan
- [x] `pants check src/ai/backend/manager::`
- [x] `pants test tests/unit/manager/sokovan/deployment/route::`

Resolves BA-5985

----
📚 Documentation preview 📚: https://sorna--11534.org.readthedocs.build/en/11534/

----
📚 Documentation preview 📚: https://sorna-ko--11534.org.readthedocs.build/ko/11534/


src/ai/backend/manager/models/deployment_revision/row.py
```diff
@@ -139,6 +139,14 @@ class DeploymentRevisionRow(Base):  # type: ignore[misc]
...
         "model_definition", PydanticColumn(ModelDefinition), nullable=True
     )
+    # Set at insert time from ``model_definition.is_health_check_enabled()``.
+    health_check_enabled: Mapped[bool] = mapped_column(
+        "health_check_enabled",
+        sa.Boolean(),
+        nullable=False,
+        default=False,
+        server_default=sa.false(),
+    )
```
HyeockJinKim: I don't think we should end up in a situation where the same value is accessed from multiple places.

HyeockJinKim (approved):

PR #11539, opened 2026-05-11
fix(BA-5988): raise IncompleteModelDefinitionError for missing model definition fields

jopemachine:
Resolves BA-5988.

## Summary
- Replace four `ValueError` raises in `ModelHealthCheckDraft`, `ModelServiceConfigDraft`, and `ModelConfigDraft`'s `to_resolved()` with a new `IncompleteModelDefinitionError` (subclass of `BackendAIError` + `web.HTTPBadRequest`) that returns `MODEL_DEPLOYMENT / REQUEST / INVALID_PARAMETERS`.
- New error message names the missing field and enumerates every merge-chain source that could have supplied the value (request input, revision preset, runtime variant baseline, model-definition.yaml on the vfolder, model mount destination). `extra_data={"field", "sources"}` makes it programmatically actionable.
- Motivation: callers hitting `"ModelConfig.name is required"` had no way to tell whether the request omitted the field or whether the DB-side defaults (preset, runtime variant) were empty.

Jira: https://lablup.atlassian.net/browse/BA-5988

## Test plan
- [x] `pants fix lint check src/ai/backend/common/config.py src/ai/backend/common/exception.py`
- [x] `pants test tests/unit/common/test_config.py`
- [x] Pre-push hook (full lint + check + relevant test suite on origin/main..HEAD)
- [ ] Verify the API error response shape against a deployment revision request that omits a required field

### DIFF
--- a/src/ai/backend/common/config.py
+++ b/src/ai/backend/common/config.py
@@ -539,7 +539,7 @@ class ModelHealthCheckDraft(BaseConfigModel):
                     "request input",
                     "revision preset",
                     "runtime variant's default model definition",
-                    "model-definition.yaml on the model vfolder",
+                    "model definition file on the model vfolder",
                 ),
             )
         # Drop unset (None) fields so the strict type's ``Field(default=...)``
@@ -567,7 +567,7 @@ class ModelServiceConfigDraft(BaseConfigModel):
                     "request input",
                     "revision preset",
                     "runtime variant's default model definition",
-                    "model-definition.yaml on the model vfolder",
+                    "model definition file on the model vfolder",
                 ),
             )
         # Drop unset (None) scalars so the strict type's ``Field(default=...)``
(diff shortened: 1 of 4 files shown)


src/ai/backend/common/config.py
```diff
@@ -572,9 +588,26 @@ class ModelConfigDraft(BaseConfigModel):
...
         if self.model_path is None:
-            raise ValueError("ModelConfig.model_path is required")
+            raise IncompleteModelDefinitionError(
+                field="ModelConfig.model_path",
+                sources=(
+                    "request input",
+                    "revision preset",
+                    "runtime variant's default model definition",
+                    "model definition file on the model vfolder",
+                    "model mount destination",
```
seedspirit, line 608: Shouldn't we omit this from the guidance message since filling model mount destination in `model_path` is the default behavior non-custom variants or hidden fallback behavior in server?

PR #11570, opened 2026-05-13
fix(BA-6007): populate live_stat stats.max/avg/rate from Prometheus samples

seedspirit:
## Summary

- `LegacyLiveStatConverter._convert_metric_samples` previously dropped `ValueType.MAX`, `AVG`, and `RATE` samples, so `compute_container.live_stat` returned the placeholder `"0"` for `stats.max` / `stats.avg` / `stats.rate` after BA-5744 + BA-5878 migrated the live-stat pipeline to Prometheus.
- Wire `maxes[-1] -> stats.max`, `avgs[-1] -> stats.avg`, and `rates[-1] -> stats.rate` so the window-stats infrastructure built in BA-5878 actually reaches the GraphQL response.
- Keep the existing `net_rx` / `net_tx` hack-multiply on `currents[-1]` (its template magnitude bug is tracked separately).

## Test plan

- [x] `pants test tests/unit/manager/api/gql_legacy/test_stat_converter.py` — added `TestWindowStats` covering MAX/AVG population, RATE → `stats.rate` for counter-rate metrics (`io_read`), and verifying the `net_rx` hack-multiply still wins.
- [x] `pants fmt / lint / check --changed-since=HEAD`
- [ ] Manually verify against a live kernel that `stats.max` / `stats.avg` are populated in `compute_container.live_stat`.

## Out of scope

- `cpu_util.stats.max` / `stats.avg` will still appear as cumulative msec (two MAX series collide — separate dedup fix).
- `net_rx` / `net_tx` `stats.rate` continues to use the hack-multiply path (separate magnitude fix for `_RATE_TEMPLATE`).

Resolves BA-6007

### DIFF
--- a/changes/11570.fix.md
+++ b/changes/11570.fix.md
@@ -1,0 +1,1 @@
+Fix `compute_container.live_stat` returning placeholder "0" for `stats.max`, `stats.avg`, and `stats.rate`
--- a/src/ai/backend/manager/api/gql_legacy/stat_converter.py
+++ b/src/ai/backend/manager/api/gql_legacy/stat_converter.py
@@ -64,8 +64,6 @@ class LegacyLiveStatConverter:
     RATE/DIFF metrics the same `(name, CURRENT)` tuple appears twice;
     `currents[0]` is the raw gauge sample, `currents[-1]` is the
     rate/diff query result.
-
-    `stats.max` / `stats.avg` are not populated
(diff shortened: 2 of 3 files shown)


seedspirit:
Close because it resolved in https://github.com/lablup/backend.ai/pull/11604

PR #11575, opened 2026-05-13
fix(BA-6005): split multi-node inference session into main + sub kernels

seedspirit:
## Summary

- Fix a bug where `DeploymentSessionDraftBuilder` produced a single `role="main"` group with `replica_count=cluster_size` for multi-node INFERENCE sessions, resulting in N kernels all stored with `cluster_role="main"` after expansion.
- Mirror the legacy `registry.py` Shape (a) layout: one `main` kernel + (`cluster_size` - 1) `sub` kernels.
- Resolves the `TooManyKernelsFound` error raised from `SessionRow.main_kernel`, which previously broke the `computeSessionNodeResult` GraphQL query on multi-node sessions.

## Test plan

- [x] Unit tests for `_resolve_kernel_groups` (single-node + parametrized 2/3/5 multi-node cases)
- [ ] Live verification: create a revision with `cluster_size >= 2`, run `add_revision`, enqueue a session, and confirm `computeSessionNodeResult` returns successfully
- [ ] Regression check for single-node (`cluster_size=1`) revisions

Resolves BA-6005

### DIFF
--- a/src/ai/backend/manager/services/session/service.py
+++ b/src/ai/backend/manager/services/session/service.py
@@ -68,6 +68,7 @@ from ai.backend.manager.data.session.options import (
     SessionHandlerOptions,
 )
 from ai.backend.manager.data.session.types import SessionStatus
+from ai.backend.manager.defs import DEFAULT_ROLE
 from ai.backend.manager.errors.common import (
     InternalServerError,
     ServiceUnavailable,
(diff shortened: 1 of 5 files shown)


src/ai/backend/manager/services/session/service.py
```diff
@@ -1599,6 +1600,21 @@ async def enqueue_session(self, action: EnqueueSessionAction) -> EnqueueSessionA
...
+                resources=resource_entries,
+                resource_opts=resource_opts,
+                environ=environ,
+                mounts=mount_entries,
+                startup_command=startup_command,
+                bootstrap_script=bootstrap_script,
+                starts_at=starts_at,
+                batch_timeout_sec=batch_timeout_sec,
+            ),
+        )
```
seedspirit, line 1617: Currently, methods such as `_resolve_kernel_groups` are duplicated in the process of creating a `SessionSpecDraft`. It would be best to refactor the code by introducing a `SessionSpecDraftBuilder` class, implementing a `build` class method within it, and encapsulating validation and kernel group resolution (as is currently done) internally to eliminate duplication.

HyeockJinKim (approved):

PR #11578, opened 2026-05-13
feat(BA-6017): add WARMING_UP route status between PROVISIONING and RUNNING

fregataa:
## Summary

- Adds a new `WARMING_UP` value to `RouteStatus` between `PROVISIONING` and `RUNNING`. The provisioning handler now lands routes in `WARMING_UP` (session enqueued, replica info pending) and a new `WarmingUpRouteHandler` graduates them to `RUNNING` on the first healthy probe — or immediately when the endpoint has no health check configured. `RunningRouteHandler` is narrowed to session-liveness monitoring only.
- Splits the lifecycle transition timestamps in Valkey: `warming_up_at` anchors the health-probe `initial_delay` window (since model loading happens during `WARMING_UP`), while `running_at` records the moment the route first becomes eligible for traffic. Both marks are written via batched APIs typed by a private `_RouteTimestampField` enum.
- Audits and updates every direct `RouteStatus` enum comparison touched by this change (rolling-update classifier, healthy-route count, terminatable-status set, route eviction targets, liveness mapping, delete-route guard) and adds `RouteStatus.pre_running_statuses()` / extends `active_route_statuses()` so future call sites use named helpers instead of hard-coded sets.

## Test plan

- [x] Unit: new `TestCheckWarmingUpRoutes` covers health-disabled / probe-pass / probe-missing / dead-session / missing-revision / mixed paths
- [x] Unit: new `TestWarmingUpRouteHandler` verifies static declarations and `execute` delegation
- [x] Unit: existing `test_initial_delay.py` updated to drive `warming_up_at`-anchored behaviour
- [x] Unit: full route handler/executor suite (`tests/unit/manager/sokovan/deployment/route::`) passes
- [x] mypy + ruff (lint/format) pass on the changed surface
- [ ] Manual verification on a live deployment: provision a route with `health_check_enabled=true` and confirm it sits in `WARMING_UP` until the first probe succeeds, then transitions to `RUNNING`
- [ ] Manual verification with `health_check_enabled=false`: confirm the route promotes to `RUNNING` immediately after the session goes live

Resolves BA-6017

----
📚 Documentation preview 📚: https://sorna--11578.org.readthedocs.build/en/11578/

----
📚 Documentation preview 📚: https://sorna-ko--11578.org.readthedocs.build/ko/11578/

### DIFF
--- a/changes/11578.feature.md
+++ b/changes/11578.feature.md
@@ -1,0 +1,1 @@
+Add a `WARMING_UP` route status between `PROVISIONING` and `RUNNING` so model-serving routes only graduate to `RUNNING` after the first healthy probe (or immediately when the endpoint has no health check configured).
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -14760,6 +14760,7 @@ enum ReplicaStatus
   @join__type(graph: STRAWBERRY)
 {
   PROVISIONING @join__enumValue(graph: STRAWBERRY)
+  WARMING_UP @join__enumValue(graph: STRAWBERRY)
   RUNNING @join__enumValue(graph: STRAWBERRY)
   TERMINATING @join__enumValue(graph: STRAWBERRY)
   TERMINATED @join__enumValue(graph: STRAWBERRY)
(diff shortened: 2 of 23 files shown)


fregataa:
Closed this PR since #11602 is introduced

PR #11580, opened 2026-05-13
feat(BA-6006): Add resource slot validation when creating deployment revision

seedspirit:
## Summary

- Add `DeploymentRevisionValidator` chain (`sokovan/deployment/validators/`) with `RequiredResourceSlotRule` mirroring PR #11556's session-side rule, so deployment revision creation fails fast with `InvalidAPIParameters` when the finalized revision spec omits a globally required resource slot.
- Two validator entry points handle the different finalized shapes per path: `validate(spec: DeploymentRevisionCreatorSpec, ctx)` for v2 `add_revision`, and `validate_legacy_revision_spec(spec: ModelRevisionSpec, ctx)` for `build_creator_from_legacy_draft` / `resolve_legacy_revision_spec`. Both entries are abstract on the rule base class so new rules must wire both paths.
- New `fetch_revision_required_slot_names` on `DeploymentRepository` / `DeploymentDBSource` reads `resource_slot_types.required`; the controller assembles `DeploymentRevisionValidationContext` once per call site.
- Intentional dual-layer defense: this PR adds revision-creation-time enforcement (prevents dead-revision / orphan-endpoint cases). Session enqueue-time enforcement via `scheduling_controller._spec_validator` (added in #11556) stays in place to cover non-revision session paths.

## Test plan

- [ ] Unit tests in `tests/unit/manager/sokovan/deployment/test_revision_validators.py` pass.
- [ ] v2 `add_revision`: 4xx returned and no row written when a required slot is missing.
- [ ] Legacy create flow (`build_creator_from_legacy_draft`): 4xx returned before endpoint creation when a required slot is missing.
- [ ] `resolve_legacy_revision_spec`: 4xx returned before returning the spec when a required slot is missing.
- [ ] Normal request with all required slots present succeeds end-to-end on all three paths.

Resolves BA-6006


src/ai/backend/manager/sokovan/deployment/validators/required_resource_slot_rule.py
```diff
@@ -0,0 +1,72 @@
...
+        self._check_resource_slots(spec.resource_slots, context.required_slot_names)
+
+    @override
+    def validate_legacy_revision_spec(
+        self,
+        spec: ModelRevisionSpec,
+        context: DeploymentRevisionValidationContext,
+    ) -> None:
+        self._check_resource_slots(
+            ResourceSlot(spec.resource_spec.resource_slots),
```
HyeockJinKim, line 54: I would like the type conversion to occur at the time of the initial input.

jopemachine (dismissed):

src/ai/backend/manager/sokovan/deployment/validators/required_resource_slot_rule.py
seedspirit: Since the type of `resource_slots` in the existing `ResourceSpec` is `Mapping[str, Any]`, modifying this part as well would likely make the scope too broad.

jopemachine (approved):

HyeockJinKim (approved):

PR #11604, opened 2026-05-14
refactor(BA-6039): rework container live-stat with explicit per-query pipeline

seedspirit:
## Summary

- Replace the single-query / positional-merge container live-stat pipeline in `common/clients/prometheus/` with **six explicit PromQL queries** (`instant`, `rate_current`, `max`, `rate_max`, `avg`, `rate_avg`).
- Replace `KernelLiveStatBatchResult.entries: dict[KernelId, KernelLiveStatEntry(values=list[MetricValue])]` with `by_kernel: dict[KernelId, KernelLiveStatValues]`, partitioned per query (`instant_current` / `instant_capacity` / `rate_current` / `max` / `avg`).
- Move capacity-sentinel synthesis from the common layer to `LegacyLiveStatConverter`, where it belongs as a legacy-API concern.
- Drop the metric-classification constants (`STATS_*_GAUGE_METRICS`, accelerator suffix patterns) and the unused `ValueType.{MAX,AVG,RATE}` enum values.

### Observable behavior changes

- `stats.max` and `stats.avg` are now populated in the GQL live-stat response (previously documented as "not populated").
- `stats.rate` no longer multiplied by `UTILIZATION_METRIC_INTERVAL` in the converter — the dedicated `rate_current` query produces the per-second magnitude directly, so the legacy `current_hook=stats.rate|diff` undo-hack is removed.

## Test plan

- [x] `pants check` clean
- [x] `pants test` for `tests/unit/common/clients/prometheus/`, `tests/unit/manager/api/gql_legacy/`, `tests/unit/manager/services/utilization_metric/` — all green
- [x] Live verification: GQL `compute_container_list { items { live_stat } }` against a RUNNING session — capacity-sentinel, `stats.max`, `stats.avg`, `pct` all populated as expected
- [ ] Frontend regression check on WebUI live-stat panel (deferred to reviewer)

Resolves BA-6039


jopemachine:
CI is broken

src/ai/backend/common/clients/prometheus/client.py
```diff
@@ -82,15 +81,19 @@ async def fetch_container_metric(
...
+        # max/rate_max and avg/rate_avg are split: gauge metrics can be aggregated
+        # directly, but cumulative counters (cpu_util/net_rx/net_tx) need rate() first.
+        return KernelLiveStatBatchResult.from_responses(
+            instant=await self._query_instant(queries.instant),
+            rate_current=await self._query_instant(queries.rate_current),
+            max=await self._query_instant(queries.max),
+            rate_max=await self._query_instant(queries.rate_max),
+            avg=await self._query_instant(queries.avg),
+            rate_avg=await self._query_instant(queries.rate_avg),
+        )
```
fregataa, line 111:
```suggestion
        instant = await self._query_instant(queries.instant)
        return KernelLiveStatBatchResult.from_responses(
            instant=instant,
            ....
        )
```

fregataa (changes requested):
`FixedQueryBuilder` seems to be a mixture of (query builder + stat query builder). it should be divided into different interfaces

fregataa (approved):

PR #11605, opened 2026-05-14
fix(BA-6037): expose missing replica/state fields on Deployment GQL/DTO

fregataa:
## Summary
- Add `status`, `traffic_status`, `health_status` fields to `ReplicaNode` DTO and `ModelReplica` GQL — `ReplicaStatusFilter`/`TrafficStatusFilter` already referenced these but the node itself didn't expose them, leaving clients unable to read the values they filter by.
- Mirror the three fields on `ModelReplicaData` domain type so the route → replica conversion in the service layer passes them through.
- Centralize manager → common enum conversion in private adapter helpers (`_to_common_route_status`, `_to_common_route_traffic_status`, `_to_common_route_health_status`) using exhaustive `match`/`case` so future enum additions surface as mypy errors.
- Expose `replica_ids` on GQL `ReplicaState` (already present in the DTO `ReplicaStateInfo`).

## Test plan
- [ ] `pants check` / `pants lint` / `pants test` on changed targets
- [ ] Manual GraphQL query against ModelReplica confirms new `status`, `traffic_status`, `health_status` fields are populated
- [ ] Existing `ReplicaStatusFilter` / `TrafficStatusFilter` filters still match the returned node values

Resolves BA-6037

----
📚 Documentation preview 📚: https://sorna--11605.org.readthedocs.build/en/11605/

----
📚 Documentation preview 📚: https://sorna-ko--11605.org.readthedocs.build/ko/11605/

### DIFF
--- a/src/ai/backend/manager/api/gql/deployment/types/deployment.py
+++ b/src/ai/backend/manager/api/gql/deployment/types/deployment.py
@@ -215,12 +215,6 @@ class DeploymentStrategyGQL:
 )
 class ReplicaState:
     desired_replica_count: int
-    replica_ids: list[UUID] = gql_added_field(
-        BackendAIGQLMeta(
-            added_version=NEXT_RELEASE_VERSION,
-            description="IDs of the replicas currently tracked for this deployment.",
-        )
-    )
 
 
 @gql_pydantic_type(
--- a/src/ai/backend/manager/api/gql/deployment/types/deployment.py
+++ b/src/ai/backend/manager/api/gql/deployment/types/deployment.py
@@ -215,6 +215,12 @@ class DeploymentStrategyGQL:
 )
 class ReplicaState:
     desired_replica_count: int
+    replica_ids: list[UUID] = gql_added_field(
+        BackendAIGQLMeta(
+            added_version=NEXT_RELEASE_VERSION,
+            description="IDs of the replicas currently tracked for this deployment.",
+        )
+    )
 
 
 @gql_pydantic_type(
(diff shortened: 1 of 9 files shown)


src/ai/backend/manager/api/gql/deployment/types/deployment.py
```diff
@@ -215,6 +215,12 @@ class DeploymentStrategyGQL:
 )
 class ReplicaState:
     desired_replica_count: int
+    replica_ids: list[UUID] = gql_added_field(
```
jopemachine: Would it be okay not to expose the replica connection?

fregataa: I would like to expose data first

src/ai/backend/manager/api/gql/deployment/types/replica.py
```diff
@@ -111,6 +113,15 @@
     name="TrafficStatus",
 )
 
+ReplicaHealthStatus: type[RouteHealthStatus] = gql_enum(
```
seedspirit: nitpick
```suggestion
ReplicaHealthStatusGQL: type[RouteHealthStatus] = gql_enum(
```

src/ai/backend/manager/api/gql/deployment/types/deployment.py
HyeockJinKim: For values where it's hard to know exactly how many there will be, let's just treat them as "connection" for now.

HyeockJinKim (dismissed):

seedspirit (dismissed):

HyeockJinKim (approved):

PR #11623, opened 2026-05-14
feat(BA-6044,BA-6048): add UUID id column to domains and scaling_groups

fregataa:
## Summary

- Add UUID `id` column as the new primary key on `domains` and `scaling_groups`; demote `name` to a UNIQUE constraint so existing `domain_name` / `scaling_group` FK references continue to function unchanged.
- Add `DomainID` and `ResourceGroupID` `NewType` definitions alongside the existing `DomainName` / `ResourceGroupName` identifiers.
- PostgreSQL ties FK dependencies to the primary-key index, so each Alembic migration temporarily drops and recreates the 10 (domains) / 7 (scaling_groups) referencing FK constraints with identical column, target column, and ON UPDATE/DELETE rules.

## Test plan

- [x] `pants fmt fix lint` on changed files passes
- [x] `pants check` (mypy + visibility) on changed files passes
- [x] `./py -m alembic upgrade head` applies both new migrations cleanly
- [x] `./py -m alembic downgrade` reverses both migrations back to the previous head
- [x] After upgrade, existing rows have `id` backfilled by `uuid_generate_v4()`
- [x] `name` UNIQUE constraint rejects duplicate inserts after the PK swap
- [x] All 17 FK constraints retain their original columns / targets / ON UPDATE/DELETE rules

This PR is the first of four planned per entity under epic BA-6043. Follow-up issues will migrate API identifiers (BA-6045, BA-6049), FK column types (BA-6046, BA-6050), and RBAC data/logic (BA-6047, BA-6051).

Resolves BA-6044
Resolves BA-6048

### DIFF
--- a/changes/11623.feature.md
+++ b/changes/11623.feature.md
@@ -1,0 +1,1 @@
+Add UUID `id` primary key to `domains` and `scaling_groups` tables and demote `name` to a UNIQUE constraint, plus introduce `DomainID` and `ResourceGroupID` identifier types.
--- a/src/ai/backend/common/identifier/domain.py
+++ b/src/ai/backend/common/identifier/domain.py
@@ -1,6 +1,11 @@
+import uuid
 from typing import NewType
 
-__all__ = ("DomainName",)
+__all__ = (
+    "DomainID",
+    "DomainName",
+)
 
 
+DomainID = NewType("DomainID", uuid.UUID)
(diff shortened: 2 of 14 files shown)


jopemachine:
CI is broken

seedspirit (approved):

PR #11625, opened 2026-05-14
enhance(BA-6053): preserve str start_command form in model definitions

### LINKED ISSUE #11624: Allow direct passing of str start_command in model definition
## Motivation

This is a quality-of-life issue, rather than a technical error.

- [Teams thread](https://teams.microsoft.com/l/message/19:8cb54d23c68b4b468992376ae23b2925@thread.skype/1778795274728?tenantId=13c6a44d-9b52-4b9e-aa34-0513ee7131f2&groupId=ddd9fbf6-726c-44f8-b727-34940c950181&parentMessageId=1778778177048&teamName=sales&channelName=General&createdTime=1778795274728)
- When copy-and-pasting web examples (e.g., vLLM recipes in [https://recipes.vllm.ai/openai/gpt-oss-120b](https://recipes.vllm.ai/openai/gpt-oss-120b) ) into the `start_command` field via our WebUI, it is cumbersome to manually remove the backslashes because they are literally preserved via `shlex.split()` and **breaks up the command execution** as “\n” are passed as regular `execv()` argument tokens.

## Required Features

- Pass-through the original (multi-line) str value all the way to the kernel runner’s service launch routine.

## Impact

- Existing model defintions with list-style `start_command` won’t be affected.
- New or existing model definitions with str-style `start_command` will be affected. The string value will be interpreted as a normal shell prompt, instead of being converted to `list[str]` via `shlex.split()`.
- The kernel runner’s own `${{ ... `}} and `{ ... `} variable expansion still works.

## Testing Scenarios

- Copy-and-pasting vLLM recipes for a new model-definition using the custom variant should work AS-IS.

JIRA Issue: BA-6053

achimnol:
resolves #11624 (BA-6053)

## Summary

Stop coercing string `start_command` values into argv lists via `shlex.split` in the API and config validators. The kernel runner already wraps `str` commands as `[shell, "-c", str]` at exec time (`src/ai/backend/kernel/service.py:118-122`), so passing the original string through gives users full shell semantics — line continuations, `$VAR` expansion, pipes — and removes the need to manually strip backslashes from copy-pasted multi-line vendor recipes (e.g. the vLLM recipes the issue cites).

## What changed

- **`common/config.py`**: dropped `_normalize_start_command` and the `_coerce_start_command` field validators on `ModelServiceConfig` / `ModelServiceConfigDraft`; widened `start_command` to `str | list[str] | None`. The trafaret schema `model_definition_iv` no longer routes strings through `shlex.split` either.
- **`{model_path}` substitution** in `ModelConfigDraft.to_resolved` now dispatches on type — per-token `.replace` for the list form, in-place `.replace` for the string form.
- **`ModelDefinition.with_args_appended`** appends preset ARGS as separate argv tokens for the list form, and shell-quotes them via `shlex.join` and concatenates with a single space for the string form so the same shell that runs the user's script parses them.
- **DTOs** in `common/dto/manager/v2/deployment/{request,types}.py` are widened to `str | list[str] | None`.
- **GraphQL types** (`ModelServiceConfigGQL`, `ModelServiceConfigInputGQL`) expose the field as `JSON | None` — same pattern already used for `ModelMetadata.version` since GraphQL cannot natively represent a scalar/list union.

## Why pass-through (and not, e.g., shell-aware line-continuation stripping)?

Two reasonable options were on the table:

1. Pre-strip `\<newline>` before `shlex.split` so multi-line strings still become argv lists.
2. Keep the original string and let the kernel runner shell-eval it.

Option 1 fixes the immediate copy-paste paper cut but does not help users who actually want `$VAR`, pipes, `&&`, or globs in their command — which is the bigger ask in the vendor recipes the issue links. Option 2 (this PR) hands those semantics over to the user explicitly and relies on the kernel runner's already-existing `[shell, "-c", str]` wrap, so the change is concentrated in validators rather than runtime.

Trade-off: a string `start_command` is now eval'd by `bash` inside the container, so anything that flows into it from external input is shell-evaluated. The list form remains argv-only (no shell) and is the default for the variant baselines; users only opt into shell-eval by writing a string.

## Scope

Per the issue: existing model definitions with list-style `start_command` are unaffected. The previously released backfill migration `8c1f7d3a9e2b` (which wrapped legacy string values as one-item lists for the older narrowed schema) is intentionally left untouched.

## Checklist

- [x] Mention to the original issue (#11624 / BA-6053)
- [x] API server-client counterparts — DTOs and GQL types widened in lockstep
- [x] Test case(s):
  - `test_to_resolved_preserves_str_start_command` — string survives validation and `{model_path}` substitution unchanged
  - `test_model_service_config_accepts_str_start_command` — strict type accepts the string form too
  - `test_appends_args_to_str_start_command_with_shell_quoting` — `with_args_appended` shell-quotes appended args
- [x] Documentation — field `description=` updated on `ModelServiceConfig`, the DTO, and both GraphQL types
- [ ] Milestone metadata specifying the target backport version *(set after review)*
- [ ] Installer updates *(none required — no schema change)*
- [ ] Update of end-to-end CLI integration tests in `ai.backend.test` *(N/A — covered by existing model-definition flow tests)*

## Test plan

- [x] `pants check` on the 5 changed Python files (mypy clean)
- [x] `pants fmt` / `pants fix` / `pants lint` clean
- [x] `pants test tests/unit/common/test_config.py` passes
- [x] `pants test tests/unit/agent/test_model_service_start_command.py` passes (no regression on the agent-side parser)
- [ ] Manual: paste a multi-line vLLM recipe (with trailing `\`) into the WebUI's `start_command` field, deploy, and verify the service starts and serves requests

----
📚 Documentation preview 📚: https://sorna--11625.org.readthedocs.build/en/11625/

----
📚 Documentation preview 📚: https://sorna-ko--11625.org.readthedocs.build/ko/11625/


achimnol:
Superseded by #11648.

PR #11632, opened 2026-05-15
fix(BA-6035): split RouteHealthRecord into ReplicaProbeTarget + ReplicaHealthStatus

HyeockJinKim:
## Summary

- Split monolithic `RouteHealthRecord` Valkey type into two focused types:
  - `ReplicaProbeTarget` (probe config, 3600s TTL) — written once at WARMING_UP entry
  - `ReplicaHealthStatus` (health result, 120s TTL) — written by observer after each probe; key expiry signals DEGRADED
- Replace revision-per-cycle DB queries with `health_check: PydanticColumn(ModelHealthCheck)` on `RoutingRow`, backfilled via Alembic migration
- Refactor `DeploymentInfo` to use direct `current_revision`/`deploying_revision` fields instead of `model_revisions` list
- Add `ReplicaProbeTargetSyncHandler` (RUNNING + HEALTHY/UNHEALTHY/DEGRADED, 60s cycle) for TTL refresh and Valkey recovery
- Fix `AppProxySyncRouteHandler` to target `RUNNING + ACTIVE` only (health filter removed; LB handles routing)
- Remove `RouteHealthRecord` class and 8 legacy Valkey methods

## Test plan

- [ ] Unit tests for `ReplicaProbeTarget` / `ReplicaHealthStatus` serialization
- [ ] Unit tests for `_register_route_probe_targets`, `sync_route_probe_targets`
- [ ] Unit tests for `check_warming_up_health` (DB-based timeout via `get_db_now()`)
- [ ] Unit tests for `check_route_health` (TTL 3-way classification: None=DEGRADED)
- [ ] Unit tests for `RouteHealthObserver` (batch write, no `within_initial_delay`)
- [ ] Alembic migration verified against local DB with representative test data

Resolves BA-6035


changes/11632.enhance.md
```diff
@@ -0,0 +1 @@
+Split route health Valkey record into ReplicaProbeTarget (probe config) and ReplicaHealthStatus (TTL-based result), removing initial_delay from Valkey and switching to DB-based timeout in check_warming_up_health.
```
seedspirit, line 1: The PR title says "fix," but the news fragment says "enhance," so I think we need to standardize this.

src/ai/backend/manager/services/deployment/service.py
```diff
@@ -257,10 +243,14 @@ def _convert_deployment_info_to_data(info: DeploymentInfo) -> ModelDeploymentDat
             updated_at=info.metadata.created_at or datetime.now(UTC),
         ),
         network_access=info.network,
-        revision_history_ids=[info.current_revision_id] if info.current_revision_id else [],
+        revision_history_ids=[info.current_revision.id]
```
seedspirit, line 246: Are you currently using this field? It seems you should check whether you’ll be affected by the change in its meaning. If it’s not being used, it might be a good idea to delete it.

PR #11638, opened 2026-05-15
fix(BA-6015): persist resolved model-definition path when set None from request

seedspirit:
## Problem
When a request omits `model_definition_path`, the model service revision was persisted with the request's empty / `None` value instead of the model-definition file actually resolved from the model vfolder. The resolved path was used at runtime but never written back, so the stored revision did not reflect what was really mounted.

## Fix
The vfolder scan now records the matched candidate path (`model-definition.yaml` / `.yml`, or an explicit override) back into the merged draft, so the resolved `model_definition_path` is persisted on the revision instead of the empty input value.

## How (mechanism)
To do this cleanly, mount identity (model vfolder, definition path, destination, extra mounts, subpath) is folded into `RevisionDraft` and participates in the merge chain, instead of being threaded as a separate sidecar `mounts` argument through `add_revision` and the draft readers:

- New `FetchedModelDefinition` / `FetchedConfigFile` types carry the matched filename out of the storage source so the resolved path can flow back into the draft.
- `_merge_mounts` uses upper-priority semantics with `None`-fallback for `model_definition_path` / `vfolder_subpath`; every other mount field is sourced from the same `request_draft.mounts`, so the merge cannot tangle values.
- `ModelRevisionCreator.to_draft` → `to_draft_with_extra_mount`: the creator derives the full `MountMetadata` from `self.mounts` and only takes the permission-resolved `extra_mounts`.

## Test plan
- [ ] `pants test tests/unit/manager/data/deployment:: tests/unit/manager/sokovan/deployment::`
- [ ] v2 add-revision with no `model_definition_path`: resolved `model-definition.yml` path is persisted on the revision
- [ ] Explicit `model_definition_path` override on modify takes precedence over the resolved/stored path
- [ ] Legacy ModifyEndpoint: untouched mount fields (extra mounts, destination, subpath) survive a replica-only modify

Resolves BA-6015

### DIFF
--- a/src/ai/backend/manager/data/deployment/creator.py
+++ b/src/ai/backend/manager/data/deployment/creator.py
@@ -8,6 +8,7 @@ from ai.backend.common.data.model_deployment.types import DeploymentStrategy
 from ai.backend.common.identifier.deployment_preset import DeploymentPresetID
 from ai.backend.common.identifier.image import ImageID
 from ai.backend.common.identifier.vfolder import VFolderUUID
+from ai.backend.common.types import MountInfoEntry
 from ai.backend.manager.data.deployment.types import (
     DeploymentMetadata,
     DeploymentNetworkSpec,
@@ -17,6 +18,7 @@ from ai.backend.manager.data.deployment.types import (
     ModelRevisionSpec,
     ModelRevisionSpecDraft,
     MountInfo,
+    MountMetadata,
     ReplicaSpec,
     ResourceSpec,
     RevisionDraft,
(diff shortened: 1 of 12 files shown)


src/ai/backend/manager/data/deployment/creator.py
```diff
@@ -65,7 +67,7 @@ class ModelRevisionCreator:
     revision_preset_id: DeploymentPresetID | None = None
     preset_values: list[PresetValueData] = field(default_factory=list)
 
-    def to_draft(self) -> RevisionDraft:
+    def to_draft_with_extra_mount(self, extra_mounts: list[MountInfoEntry]) -> RevisionDraft:
```
seedspirit, line 70: Used in DeploymentController. This has been updated to include mount information in `RevisionDraft`. Structural improvements are needed going forward.

HyeockJinKim (approved):

PR #11647, opened 2026-05-18
feat(BA-6063): add UserID type and ResolveUserIDByAccessKey action

fregataa:
## Summary
- Add `UserID = NewType("UserID", UUID)` under `common/identifier/user.py` so handler/service signatures can carry the keypair owner's UUID with static-type discrimination (mirrors the existing `common/identifier/vfolder.py` pattern).
- Add `ResolveUserIDByAccessKeyAction` on the auth service plus a backing `get_user_id_by_access_key` in the auth repository / db_source (selects `keypairs.c.user`, raises `AccessKeyNotFound` when missing) so any handler or service can resolve an access_key to its owning user_uuid without depending on the keypairs table directly.
- Register the new action processor on `AuthProcessors` and add the spec to `supported_actions()`. Pure additive — no existing call site is modified.

## Test plan
- [x] `pants fmt fix lint check` on the touched files (passes locally)
- [x] No behavior change to existing REST v1/v2, GraphQL, agent RPC, or events surfaces

Resolves BA-6063

### DIFF
--- a/src/ai/backend/manager/repositories/auth/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/auth/db_source/db_source.py
@@ -17,6 +17,7 @@ from ai.backend.common.metrics.metric import DomainType, LayerType
 from ai.backend.common.resilience.policies.metrics import MetricArgs, MetricPolicy
 from ai.backend.common.resilience.policies.retry import BackoffStrategy, RetryArgs, RetryPolicy
 from ai.backend.common.resilience.resilience import Resilience
+from ai.backend.common.types import AccessKey
 from ai.backend.manager.data.auth.login_session_types import LoginHistoryData, LoginSessionData
 from ai.backend.manager.data.auth.types import GroupMembershipData, UserData
 from ai.backend.manager.data.common.types import SearchResult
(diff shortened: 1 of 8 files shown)


src/ai/backend/manager/repositories/auth/db_source/db_source.py
```diff
@@ -292,6 +294,16 @@ async def fetch_user_info_by_access_key(self, access_key: str) -> tuple[str, Use
                 raise ValueError("Unknown owner access key")
             return row.domain_name, row.role
 
+    @auth_db_source_resilience.apply()
+    async def fetch_user_id_by_access_key(self, access_key: str) -> UserID:
```
jopemachine: Why not using AccessKey type?

src/ai/backend/manager/repositories/auth/db_source/db_source.py
```diff
@@ -292,6 +295,16 @@ async def fetch_user_info_by_access_key(self, access_key: str) -> tuple[str, Use
                 raise ValueError("Unknown owner access key")
             return row.domain_name, row.role
 
+    @auth_db_source_resilience.apply()
+    async def fetch_user_id_by_access_key(self, access_key: AccessKey) -> UserID:
```
jopemachine, line 299: Just a question, is there a reason this file uses the Core API and a direct connection instead of the ORM and db_session?

fregataa: we used to recommend to use ORM API but it doesn't matter recently because we isolate db related codes into repository or db source layers

seedspirit: Nevertheless, using an ORM API seems preferable for type safety and IDE support.

jopemachine (approved):

PR #11670, opened 2026-05-18
fix(BA-6091): surface reads_vfolder_config_files and default_model_definition in runtime variant CRUD

jopemachine:
## Summary

- `runtime_variants.reads_vfolder_config_files` and `default_model_definition` were added after the v2 CRUD stack shipped, but never wired into the API. Result: `CREATE` blew up on `NotNullViolation`, `UPDATE` silently dropped both fields, and `GET`/`SEARCH` responses omitted them.
- This PR wires both columns through every layer (REST v2 DTOs, repository specs, adapter, GraphQL, SDK, CLI) and adds round-trip component tests.

## Changes

**DTOs (`common/dto/manager/v2/runtime_variant/`)**
- `CreateRuntimeVariantInput`: `reads_vfolder_config_files: bool = False`, `default_model_definition: ModelDefinitionInput | None = None` (reuses the deployment v2 input shape).
- `UpdateRuntimeVariantInput`: both fields default to `None` for the `OptionalState`/`leave unchanged` semantics.
- `RuntimeVariantNode`: both fields exposed (`default_model_definition` as `dict[str, Any]` for JSON-friendly output).

**Repository specs**
- `RuntimeVariantCreatorSpec.build_row()` writes both columns — fixes the `NotNullViolation` on `INSERT`.
- `RuntimeVariantUpdaterSpec` routes both columns through `OptionalState`.

**Adapter**
- `create()` defaults a missing draft to an empty `ModelDefinitionDraft()` (mirrors the seed for the `custom` variant).
- `update()` uses `OptionalState` for the two new fields so omitting them leaves the row untouched.
- `_data_to_node()` serializes the draft via `model_dump()`.

**GraphQL**
- `RuntimeVariantGQL`, `CreateRuntimeVariantInputGQL`, `UpdateRuntimeVariantInputGQL` add the two fields via `gql_added_field(NEXT_RELEASE_VERSION)`, reusing `ModelDefinitionInputGQL` from deployment.

**CLI**
- `./bai admin runtime-variant create` and `update` accept the new keys in their JSON body, including the nested `default_model_definition`.

**Tests**
- New `TestRuntimeVariantExtendedFields` covers: empty-draft default on create, round-trip with both fields, update of both fields, and update preserving existing values when omitted.

## Test plan

- [ ] CI: pants lint / check / test
- [ ] Manual: `./bai admin runtime-variant create '{"name":"x","reads_vfolder_config_files":false}'` succeeds (no longer raises `NotNullViolation`)
- [ ] Manual: `./bai admin runtime-variant create` with a `default_model_definition` payload round-trips through `./bai admin runtime-variant get <id>`
- [ ] Manual: `./bai admin runtime-variant update <id> '{"reads_vfolder_config_files":true}'` flips the gate without touching the draft
- [ ] Existing seed variants (vllm/sglang/nim/tgi/cmd/custom) still behave correctly at deployment-revision time


jopemachine:
This PR is being closed because the feature was intentionally excluded.

PR #11686, opened 2026-05-18
fix(BA-6105): convert BraceStyleAdapter f-string log calls in 25.15-LTS-scope files

jopemachine:
## Summary
- Broader sweep of BA-6099 covering 153 `BraceStyleAdapter` f-string log call sites across 33 files that exist on the `25.15` LTS branch (accelerator, agent, appproxy, common, kernel, manager, storage).
- Since `BraceStyleAdapter.__str__` always re-runs `.format()` on the resolved message, an f-string interpolation containing literal `{` / `}` characters would raise inside the logger and swallow the message — this is the same root cause as BA-6099.
- Switch each call to brace-style format with separate positional arguments so formatting happens lazily and any runtime brace characters cannot break the logger.
- Target milestone `25.15` (LTS). Forward-port to `26.4` / `26.5` / `main`. Low-risk mechanical conversion improving customer LTS observability.

Resolves BA-6105.


seedspirit (dismissed):

src/ai/backend/common/identity.py
seedspirit: It appears you are not currently using BraceStyleLogger, so please check this part.

seedspirit:
There seem to be quite a few points from the Copilot reviews that we should incorporate, so could you please take a look?

HyeockJinKim (approved):

PR #11691, opened 2026-05-19
fix(BA-6107): skip slot-type checks for zero-valued resource slots

fregataa:
## Summary
- `ImageSlotTypeRule` and `RequestedSlotTypeRule` now skip slots whose requested quantity (or image-declared minimum) is `<= 0`, matching the behavior of `ResourceLimitRule`.
- Per-kernel validation errors in both rules are aggregated and raised together instead of failing on the first offending `kernel_spec`.
- Added unit tests covering both zero-value skip cases.

## Test plan
- [x] `pants test tests/unit/manager/sokovan/scheduling_controller/validators/test_session_spec_rules.py`
- [x] `pants check` / `pants lint` on changed files

Resolves BA-6107

### DIFF
--- a/src/ai/backend/manager/sokovan/scheduling_controller/validators/image_slot_type_rule.py
+++ b/src/ai/backend/manager/sokovan/scheduling_controller/validators/image_slot_type_rule.py
@@ -14,8 +14,11 @@ would otherwise let the session reach the scheduler only to fail there.
 
 from __future__ import annotations
 
+from decimal import Decimal
+
 from ai.backend.manager.data.session.spec import SessionSpec
 from ai.backend.manager.errors.api import InvalidAPIParameters
+from ai.backend.manager.sokovan.scheduling_controller.resource_parse import image_min_slots
 from ai.backend.manager.sokovan.scheduling_controller.validators.session_spec_base import (
     SessionSpecValidationContext,
     SessionSpecValidatorRule,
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/sokovan/scheduling_controller/validators/image_slot_type_rule.py
```diff
@@ -41,23 +44,26 @@ def validate(
...
         for idx, kernel in enumerate(spec.kernel_specs):
             image_info = context.image_infos.get(kernel.execution_spec.image_id)
             if image_info is None:
                 continue
+            min_slots = image_min_slots(image_info)
             unknown = sorted(
                 slot_name
                 for slot_name in image_info.resource_spec
                 if slot_name not in rg_slot_types
+                and min_slots.get(slot_name, Decimal(0)) > Decimal(0)
```
seedspirit, line 57: If `slot_name` isn't included in `min_slots`, it seems like it would be treated as "known" even if the slot type isn't specified in `rg`. Would that be okay?

seedspirit: Actually, no—I think I misunderstood the logic. For the sake of readability, how about changing the logic as shown below?

```python
for slot_name, slot_spec in image_info.resource_spec.items():
    if slot_name in rg_slot_types:
        continue
    if image_min_for(slot_spec) <= 0:
        continue
    errors.append(...)
```

jopemachine (approved):

PR #11694, opened 2026-05-19
feat(BA-6095): add scoped audit log GraphQL schema stub

fregataa:
## Summary
- Add `scopedAuditLogsV2` root field with a stub resolver that validates the scope target (via Pydantic) and then raises `NotImplementedError`.
- Add `AuditLogScope` input with category-separated lists (`entity`, `triggeredUser`); leaves are OR'd and the combined set must be non-empty (Pydantic `model_validator`).
- Add `EntityTypeScope` as a shared `(RBACElementType, id)` primitive under `common/dto/manager/v2/rbac` and `api/gql/rbac` so it can be reused outside audit log contexts.

## Test plan
- [x] `pants check` / `pants lint` pass on changed files
- [x] GraphQL introspection exposes `scopedAuditLogsV2`, `AuditLogScope`, `EntityTypeScope`

Resolves BA-6095

----
📚 Documentation preview 📚: https://sorna--11694.org.readthedocs.build/en/11694/

----
📚 Documentation preview 📚: https://sorna-ko--11694.org.readthedocs.build/ko/11694/

### DIFF
--- a/src/ai/backend/manager/api/gql/audit_log/types/filter.py
+++ b/src/ai/backend/manager/api/gql/audit_log/types/filter.py
@@ -42,6 +42,7 @@ class AuditLogStatusFilterGQL(PydanticInputMixin[AuditLogStatusFilter]):
 )
 class AuditLogFilterGQL(PydanticInputMixin[AuditLogFilter]):
     entity_type: StringFilter | None = None
+    entity_id: StringFilter | None = None
     operation: StringFilter | None = None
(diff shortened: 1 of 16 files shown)


src/ai/backend/manager/api/gql/audit_log/types/filter.py
seedspirit, line 49: It seems it need to verify whether it should be changed to UUIDFilter.

jopemachine (approved):

PR #11695, opened 2026-05-19
feat(BA-6100): generalize bulk action to mixed RBAC element types

fregataa:
## Summary
- `BaseBulkAction` now carries `element_refs: list[RBACElementRef]` (was `entity_ids: list[str]`), so one bulk action can mix RBAC element types.
- `BulkActionProcessor` runs every validator against the original action, aggregates all denials across the chain, and raises `PermissionDeniedError` with each denied entity paired with its `deny_reason`. Any denial fails the whole action.
- `BulkActionRBACValidator` builds one `PermissionResolutionKey` per ref using the ref's own `element_type`, so heterogeneous types resolve in a single bulk repository call.

Resolves BA-6100

### DIFF
--- a/src/ai/backend/manager/actions/action/bulk.py
+++ b/src/ai/backend/manager/actions/action/bulk.py
@@ -1,30 +1,18 @@
-from abc import ABC, abstractmethod
+from abc import abstractmethod
 from collections.abc import Sequence
 from typing import override
 
 from ai.backend.manager.data.permission.types import RBACElementRef
-from ai.backend.manager.repositories.base.types import SearchScope
 
 from .base import BaseAction, BaseActionResult
+from .types import ActionTarget
 
 
-class BulkActionTarget(ABC):
-    @abstractmethod
-    def to_rbac_element_ref(self) -> RBACElementRef:
-        raise NotImplementedError
-
-
-class SearchableBulkActionTarget(BulkActionTarget):
-    @abstractmethod
-    def to_search_scope(self) -> SearchScope:
-        raise NotImplementedError
-
-
-class BaseBulkAction[TTarget: BulkActionTarget](BaseAction):
-    """Bulk action over a sequence of :class:`BulkActionTarget`.
+class BaseBulkAction[TTarget: ActionTarget](BaseAction):
+    """Bulk action over a sequence of :class:`ActionTarget`.
 
     Parametrize the target type to require a richer contract — e.g.
-    ``BaseBulkAction[SearchableBulkActionTarget]`` for a scoped search.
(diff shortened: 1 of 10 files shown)


src/ai/backend/manager/actions/processor/bulk.py
```diff
@@ -69,44 +43,24 @@ def __init__(
...
+            validation = await validator.validate(action, action_trigger_meta)
+            if validation.denied_entities:
+                denied_entities.extend(validation.denied_entities)
+
+        if denied_entities:
+            err_msg = f"Bulk action denied for {len(denied_entities)} entity(ies): " + ", ".join(
+                f"{d.entity_ref.to_str()} ({d.deny_reason})" for d in denied_entities
             )
-            filtered_action = self._filter_by_validation(filtered_action, validation)
+            raise PermissionDeniedError(err_msg)
```
jopemachine, line 58: Could there be cases where operations on allowed_entities should still proceed even if denied_entities exist?

tests/unit/manager/actions/test_bulk_processor.py
```diff
@@ -25,14 +25,20 @@
...
-    @override
-    def typed_entity_ids(self) -> list[str]:
-        return list(self.entity_ids)
+def _ref(element_type: RBACElementType, element_id: str) -> RBACElementRef:
+    return RBACElementRef(element_type=element_type, element_id=element_id)
+
+
+_REF_A = _ref(RBACElementType.SESSION, "a")
+_REF_B = _ref(RBACElementType.SESSION, "b")
+_REF_C = _ref(RBACElementType.SESSION, "c")
```
jopemachine: These look better injected as fixtures.

src/ai/backend/manager/actions/action/bulk.py
```diff
@@ -1,33 +1,17 @@
...
-    Bulk actions intentionally carry **only** ``entity_ids``. User context
-    (user id, role) flows through ``current_user()``, not the action, so
-    ``BulkActionProcessor`` can reconstruct a filtered action by calling
-    ``type(action)(entity_ids=...)`` directly — no ``__init__`` override or
-    factory hook is required. Subclasses that try to add required fields
-    break that constructor call and will fail fast at runtime, which is
-    intentional.
-    """
+class BaseBulkAction(BaseAction):
+    """Bulk action over a list of ``RBACElementRef`` (may mix element types)."""
```
jopemachine: Could you add a single example of an Action where element types are mixed in the comment?

jopemachine (dismissed):

jopemachine (dismissed):

jopemachine (approved):

PR #11699, opened 2026-05-19
fix(BA-6102): accept legacy VFolderMount shape in MountInfoEntry

seedspirit:
## Summary

`deployment_revisions.extra_mounts` rows persisted before commit 8321c79aa still carry the legacy `VFolderMount` shape (`{name, vfid, vfsubpath, host_path, kernel_path, mount_perm, usage_mode}`). When `get_deployments_by_ids` loads them, `PydanticListColumn.process_result_value` fails on missing `vfolder_id` and aborts the whole `sync_appproxy` batch — taking healthy deployments down with it.

`MountInfoEntry` now accepts the load-bearing legacy keys as validation aliases:
- `vfid` → `vfolder_id` (`field_validator` strips the `"<quota_scope_id>/"` prefix)
- `vfsubpath` → `subpath`
- `kernel_path` → `mount_destination` (pre-existing)

Surplus keys (`name`, `host_path`, `usage_mode`) drop through Pydantic's `extra="ignore"` — already re-derived at session creation. No data backfill.

## Test plan

- [x] Parametrized legacy-shape decoding tests in `tests/unit/common/test_types.py`.
- [x] Integration test that raw-`UPDATE`s a legacy-shape JSONB and verifies `get_deployments_by_ids` hydrates it.
- [ ] Manual: legacy-shape row in local DB → `sync_appproxy` cycle no longer aborts.

Resolves BA-6102

### DIFF
--- a/src/ai/backend/common/types.py
+++ b/src/ai/backend/common/types.py
@@ -772,9 +772,7 @@ class MountInfoEntry(BackendAISchema):
     def _strip_quota_scope_prefix(cls, value: Any) -> Any:
         # VFolderID strings may be in the form "<quota_scope_id>/<uuid>" or just "<uuid>"
         if isinstance(value, str) and "/" in value:
-            # Use rpartition in case there are multiple slashes
-            _, _, folder_part = value.rpartition("/")
-            return folder_part
+            return VFolderID.from_str(value).folder_id
         return value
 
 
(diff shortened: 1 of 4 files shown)


src/ai/backend/common/types.py
```diff
@@ -753,13 +754,28 @@ class MountInfoEntry(BackendAISchema):
...
+    )
+
+    @field_validator("vfolder_id", mode="before")
+    @classmethod
+    def _strip_quota_scope_prefix(cls, value: Any) -> Any:
+        # VFolderID strings may be in the form "<quota_scope_id>/<uuid>" or just "<uuid>"
+        if isinstance(value, str) and "/" in value:
+            # Use rpartition in case there are multiple slashes
+            _, _, folder_part = value.rpartition("/")
+            return folder_part
```
fregataa: how about this

```suggestion
    VFolderID.from_str(value).folder_id
```

fregataa (approved):

PR #11731, opened 2026-05-21
feat(BA-6098): expose scoped audit-log search via REST v2, SDK, and CLI

fregataa:
## Summary
- Add the non-admin scoped audit-log search across the full v2 stack on top of the adapter introduced in BA-6096.
- New `POST /v2/audit-logs/scoped-search` (`auth_required`) — accepts `ScopedSearchAuditLogsInput` in the body and delegates to `AuditLogAdapter.scoped_search`. Existing `POST /v2/audit-logs/search` (`superadmin_required`) is unchanged.
- New SDK method `V2AuditLogClient.scoped_search()`, and new CLI `./bai audit-log scoped-search` with repeatable `--scope <type>:<id>` flags (`triggered_user:<uuid>` → actor scope, any other `<type>` → entity scope via `RBACElementTypeDTO`).
- Component tests covering admin endpoint unchanged, scoped-search auth, empty-scope rejection at the API boundary, attribute filter, and both pagination modes.

### URL pattern note
The scoped audit-log query is a multi-item OR-union (entity list + triggered_user list), so it cannot be expressed via the `/v2/{entity}/{scope_type}/{scope_id}/search` nested-path pattern. The Jira ticket explicitly specifies the scope as a body field, so the endpoint is `/v2/audit-logs/scoped-search`.

## Test plan
- [x] Live REST/SDK/CLI smoke test against running manager (admin keypair):
  - `./bai audit-log scoped-search --scope triggered_user:<uuid> --limit 2` returns rows.
  - `./bai audit-log scoped-search --scope agent:i-MacBook-Air.local --scope triggered_user:<uuid> --status error` honors OR-union scope and `--status` filter.
  - Existing `./bai audit-log search` (admin) still works.
  - Unknown scope type / invalid UUID / missing `--scope` produce clean CLI errors.
- [x] `pants fmt`, `pants fix`, `pants lint --changed-since=origin/main` clean.
- [x] CI runs `pants check` and the new `tests/component/audit_log/test_scoped_search_rest.py`.

Resolves BA-6098

### DIFF
--- a/src/ai/backend/client/cli/v2/audit_log/commands.py
+++ b/src/ai/backend/client/cli/v2/audit_log/commands.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import uuid
 
 import click
 
@@ -12,6 +13,55 @@ from ai.backend.client.cli.v2.helpers import (
     parse_order_options,
     print_result,
 )
+from ai.backend.common.dto.manager.v2.rbac.types import (
+    EntityTypeScope,
+    RBACElementTypeDTO,
+    UUIDScope,
+)
+
+_TRIGGERED_USER_KEYWORD = "triggered_user"
+
+
+def _parse_scope_items(
+    raw_scopes: tuple[str, ...],
+) -> tuple[list[EntityTypeScope], list[UUIDScope]]:
+    """Parse ``--scope`` values of form ``<type>:<id>`` into entity / triggered_user buckets."""
+    entity: list[EntityTypeScope] = []
+    triggered_user: list[UUIDScope] = []
+    for raw in raw_scopes:
+        if ":" not in raw:
+            raise click.BadParameter(
+                f"--scope must be in '<type>:<id>' form (got: {raw!r})",
+                param_hint="--scope",
+            )
+        type_part, id_part = raw.split(":", 1)
+        type_part = type_part.strip()
+        id_part = id_part.strip()
+        if not type_part or not id_part:
+            raise click.BadParameter(
+                f"--scope requires both type and id (got: {raw!r})",
+                param_hint="--scope",
+            )
+        if type_part == _TRIGGERED_USER_KEYWORD:
+            try:
+                triggered_user.append(UUIDScope(value=uuid.UUID(id_part)))
+            except ValueError as exc:
+                raise click.BadParameter(
+                    f"--scope triggered_user requires a UUID id (got: {id_part!r})",
+                    param_hint="--scope",
+                ) from exc
+            continue
+        try:
+            element_type = RBACElementTypeDTO(type_part)
+        except ValueError as exc:
+            valid = ", ".join(sorted(t.value for t in RBACElementTypeDTO))
+            raise click.BadParameter(
+                f"unknown scope type {type_part!r}; "
+                f"expected one of: {_TRIGGERED_USER_KEYWORD}, {valid}",
+                param_hint="--scope",
+            ) from exc
+        entity.append(EntityTypeScope(entity_type=element_type, entity_id=id_part))
+    return entity, triggered_user
 
 
 @click.group(name="audit-log")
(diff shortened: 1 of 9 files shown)


src/ai/backend/manager/api/rest/v2/audit_log/registry.py
```diff
@@ -26,5 +26,11 @@ def register_v2_audit_log_routes(
         handler.admin_search_audit_logs,
         middlewares=[superadmin_required],
     )
+    registry.add(
+        "POST",
+        "/scoped-search",
```
jopemachine, line 31: What about making scoped-search the default and exposing it as the /search endpoint (or /scope/search for consistency)?

Then my and admin searches could use explicit prefixes like /my/search and /admin/search.

src/ai/backend/client/cli/v2/audit_log/commands.py
```diff
@@ -12,6 +13,55 @@
...
+from ai.backend.common.dto.manager.v2.rbac.types import (
+    EntityTypeScope,
+    RBACElementTypeDTO,
+    UUIDScope,
+)
+
+_TRIGGERED_USER_KEYWORD = "triggered_user"
+
+
+def _parse_scope_items(
```
jopemachine: This logic seems likely to be reused across the CLI. How about extracting it into a common utility method?

src/ai/backend/manager/api/rest/v2/audit_log/registry.py
fregataa: Thanks — agree the prefix-based scheme reads cleaner. Out of scope here though: BA-6098 explicitly requires the existing `POST /v2/audit-logs/search` (superadmin) to remain unchanged. Restructuring would touch every v2 entity and break clients, so it belongs in a separate convention proposal.

src/ai/backend/client/cli/v2/audit_log/commands.py
fregataa: Good call — refactored. Split into two layers:

- `ScopeArgType` (Click `ParamType` in `client/cli/v2/helpers.py`) validates the `<type>:<id>` syntax only and returns a generic `ScopeArg` dataclass with both fields as plain strings. Reusable by any future CLI that wants this shape.
- `_to_audit_log_scope_buckets()` stays in `audit_log/commands.py` as a private helper that maps the parsed args into `EntityTypeScope` / `UUIDScope` buckets (this is audit-log-specific because the bucket shape comes from `AuditLogScope` and the `triggered_user` keyword is its own convention).

The reasoning for not pushing the bucketing into `helpers`: the return type is tied to `AuditLogScope`'s shape (entity list + actor UUID list) and there's no second user of that shape today, so abstracting it now would be premature. The generic syntax parser is the part that's safe to share.

src/ai/backend/client/cli/v2/audit_log/commands.py
```diff
@@ -109,3 +152,121 @@ async def _run() -> None:
...
+)
+@click.option(
+    "--operation",
+    type=str,
+    default=None,
+    help="Filter by operation (contains).",
+)
+@click.option(
+    "--status",
+    type=click.Choice(["success", "error", "unknown", "running"], case_sensitive=False),
```
jopemachine, line 189: Ideally, it would be better to move AuditLogStatus to an appropriate location under common and reuse it here as well.

src/ai/backend/client/cli/v2/audit_log/commands.py
```diff
@@ -109,3 +152,121 @@ async def _run() -> None:
...
+@click.option(
+    "--scope",
+    "scopes",
+    type=ScopeArgType(),
+    multiple=True,
+    required=True,
+    help=(
+        "Scope item in '<type>:<id>' form. "
+        f"Use '{_TRIGGERED_USER_KEYWORD}:<user-uuid>' to scope by actor; "
+        "otherwise '<entity_type>:<entity_id>' (e.g., 'vfolder:<uuid>', 'project:<uuid>'). "
```
jopemachine, line 167: Since they are semantically slightly different, and if there may be room for more scopes to be added later, what do you think about accepting entity_scope and trigger_by_scope as separate arguments?

src/ai/backend/client/cli/v2/audit_log/commands.py
```diff
@@ -12,6 +13,48 @@
     parse_order_options,
     print_result,
 )
+from ai.backend.client.cli.v2.types import ScopeArg, ScopeArgType
+from ai.backend.common.dto.manager.v2.rbac.types import (
+    EntityTypeScope,
+    RBACElementTypeDTO,
+    UUIDScope,
+)
```
HyeockJinKim, line 21: Since these types keep getting included in the CLI header, the issue where CLI calls take too long is reoccurring.

src/ai/backend/client/cli/v2/audit_log/commands.py
```diff
@@ -12,6 +13,48 @@
...
+            element_type = RBACElementTypeDTO(scope.type)
+        except ValueError as exc:
+            valid = ", ".join(sorted(t.value for t in RBACElementTypeDTO))
+            raise click.BadParameter(
+                f"unknown scope type {scope.type!r}; "
+                f"expected one of: {_TRIGGERED_USER_KEYWORD}, {valid}",
+                param_hint="--scope",
+            ) from exc
+        entity.append(EntityTypeScope(entity_type=element_type, entity_id=scope.id))
+    return entity, triggered_user
```
HyeockJinKim, line 57: We keep running into issues where conversion logic is being defined as global functions, so I think we need to sort this out.

PR #11734, opened 2026-05-21
fix(BA-6042): detect Sentinel failover via ROLE health check

HyeockJinKim:
## Summary
- Add unified `check_health()` on `AbstractValkeyClient`; Sentinel implementation runs `ROLE` on the connected node and raises `ValkeyRoleMismatchError` when the node is no longer master, so failover is detected on the next health check via a single round-trip instead of relying on `Sentinel.discover_master()`.
- `MonitoringValkeyClient` monitor loop now calls `check_health()` (replacing the prior `_check_ping` + `need_reconnect` combo); role mismatch is added to `reconnectable_exceptions` so failover triggers immediate reconnection while existing failure-threshold and reconnect-event semantics are preserved.
- `ping()` / `need_reconnect()` remain on the abstract interface for external callers (health checker, tests); only the monitor loop migrates.

## Test plan
- [x] `pants fmt :: && pants fix :: && pants lint --changed-since=origin/main` clean
- [x] `pants test tests/unit/common/clients/valkey_client::` all pass
- [ ] CI: full `pants check` / `pants test` green
- [ ] Manual: observe monitor-loop reconnect on simulated Sentinel failover (separate follow-up — no real-sentinel integration harness in repo yet)

Resolves BA-6042


HyeockJinKim:
### Manual verification with real Sentinel cluster

Brought up `redis-master + 2 replicas + 3 sentinels` (bitnami latest, quorum 2, down-after 2s, failover-timeout 10s) on host network, with `--bind 127.0.0.1` and `--replica-announce-ip 127.0.0.1` so the IPs sentinel reports are reachable from the host.

Drove the full `MonitoringValkeyClient` lifecycle via `create_valkey_client(...)` against the real sentinels (`monitor_interval=1.0s` for fast detection), then issued `SENTINEL FAILOVER mymaster` from outside the process.

Observed log line from the monitor loop:
```
Health check signaled reconnect: Valkey Role Mismatch
(Connected node role is b'slave', expected 'master' (service='mymaster')),
reconnecting immediately...
```

Timeline:
- `t≈0s` — connected, `check_health()` OK on master `127.0.0.1:16379`
- `t≈0s` — `SENTINEL FAILOVER mymaster` -> `OK`
- `t≈13s` — `check_health()` sees `b'slave'` from ROLE → `ValkeyRoleMismatchError` → caught by `reconnectable_exceptions` → `_reconnect()` → `Sentinel.discover_master()` returns the new master → both operation and monitor clients reconnect to `127.0.0.1:16381`
- Post-reconnect `check_health()` passes again on the new master

Also confirmed by direct `redis-cli` that the originally-elected master node was actually demoted (`ROLE` returned `slave`) — so the ROLE-based detection is exercising the genuine failover signal, not a stale cached connection state.

PR #11736, opened 2026-05-21
fix(BA-6138): use SESSION entity_type for SearchKernelsAction RBAC

fregataa:
## Summary
- `SearchKernelsAction` used `EntityType.SESSION_KERNEL`, which has no corresponding `RBACElementType`, causing `RBACTypeConversionError` when resolving the `kernels` field under a session in GraphQL (e.g. `projectSessionsV2.edges.node.kernels`).
- Switch to `EntityType.SESSION` so the RBAC check matches existing session-level role bindings — kernel access is governed by the same permission as the owning session.

## Test plan
- [x] Restart manager and query `projectSessionsV2 { edges { node { kernels { ... } } } }` as a regular user — expect 200, no `permission_generic_internal-error`.

Resolves BA-6138

### DIFF
--- a/src/ai/backend/manager/services/session/actions/search_kernel.py
+++ b/src/ai/backend/manager/services/session/actions/search_kernel.py
@@ -27,7 +27,7 @@ class SearchKernelsAction(SessionScopeAction):
     @override
     @classmethod
     def entity_type(cls) -> EntityType:
-        return EntityType.SESSION_KERNEL
+        return EntityType.SESSION
 
     @override
     def entity_id(self) -> str | None:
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/services/session/actions/search_kernel.py
```diff
@@ -27,7 +27,7 @@ class SearchKernelsAction(SessionScopeAction):
     @override
     @classmethod
     def entity_type(cls) -> EntityType:
-        return EntityType.SESSION_KERNEL
```
jopemachine, line 30: SESSION_KERNEL no longer seems to be used anywhere. Is there any reason to keep it around?

jopemachine (approved):

PR #11766, opened 2026-05-22
feat(BA-6155): add metadata-driven chunk-store upload session engine

jopemachine:
## 📚 Stacked PRs

This PR is part of a 5-PR stack implementing BA-3974 (epic: BA-6153). **Merge in order:**

1. 👉 **#11766** — `feat(BA-6155): add Valkey-backed chunk-store upload session engine` ← you are here
2. ⬇️ #11767 — `fix(BA-6156): rewire TUS PATCH/HEAD to chunk-based store` (**actual user-visible fix**)
3. ⬇️ #11768 — `test(BA-6157): add multi-proxy NFS race regression test`
4. ⬇️ #11769 — `feat(BA-6158): support TUS Checksum extension`
5. ⬇️ #11770 — `feat(BA-6159): add /upload/status endpoint + progress headers`

## Summary

Introduces the concurrency-safe TUS upload session engine. Per-session metadata
is the source of truth in **Valkey** (keyed by session id) and is guarded by a
**per-session distributed lock** produced by an injected `DistributedLockFactory`.
Chunk payload bytes live as individual files under `chunks/chunk_<offset>.dat`
on the shared filesystem; they are content-addressed by `(offset, sha256)` and
idempotent, so the heavy write path needs no coordination at all. Only the small
metadata read-modify-write window is serialized — and across Storage Proxy
replicas, since the lock is distributed.

> **Note:** This PR combines two originally stacked issues — the pure
> session-state model (BA-6154) and the on-disk storage engine (BA-6155). They
> live in the same package and the model has no consumer on its own, so splitting
> them across PRs hurt reviewability. Both issues are resolved here.

## Architecture

The handler (PR #11767) only adapts the request body; **all chunk file I/O and
metadata management lives in the engine** (this PR).

```mermaid
flowchart TB
    classDef pr66 fill:#e6f7ff,stroke:#1890ff,color:#000
    classDef pr67 fill:#fff7e6,stroke:#fa8c16,color:#000
    classDef redis fill:#fff1f0,stroke:#cf1322,color:#000
    classDef disk fill:#f6ffed,stroke:#52c41a,color:#000

    subgraph L1["PATCH handler (PR #11767)"]
        REQ["request.content"]:::pr67
        ADP["TusChunkUploadStreamReader"]:::pr67
        HND["tus_upload_part"]:::pr67
        REQ --> ADP --> HND
    end

    subgraph L2["Upload engine (PR #11766)"]
        WTC["write_temp_chunk"]:::pr66
        CC["commit_chunk"]:::pr66
        FIN["assemble + cleanup"]:::pr66
    end

    subgraph L3R["Valkey · source of truth"]
        STATE["tus.upload.session:id"]:::redis
        LOCK["tus.upload.lock:id"]:::redis
    end

    subgraph L3F["shared FS · chunk payloads"]
        TMP["chunk_off.tok.tmp"]:::disk
        DAT["chunk_off.dat"]:::disk
        OUT["assembled file"]:::disk
    end

    HND -->|"1. drain body"| WTC
    HND -->|"2. promote temp"| CC
    HND -->|"3. on final commit"| FIN
    WTC -->|"lock-free write"| TMP
    CC -->|"acquire"| LOCK
    CC -->|"get / set"| STATE
    CC -->|"atomic rename"| DAT
    FIN -->|"concat in order"| OUT
    FIN -->|"mark COMPLETED"| STATE
    FIN -->|"unlink chunks"| DAT
```

### Data model

```mermaid
classDiagram
    class TusUploadSession {
        +ensure_initialized()
        +read_state() TusSessionState
        +open_temp_chunk(offset) Path
        +write_temp_chunk(offset, reader) WrittenChunk
        +commit_chunk(offset, path, length, sha256) ChunkCommitResult
        +assemble(target)
        +cleanup()
    }
    class TusSessionState {
        +session_id
        +total_size
        +committed_chunks : ChunkMetadata[]
        +status : UploadStatus
        +committed_offset
        +find_at_offset(offset)
        +append_chunk(chunk)
    }
    class ChunkMetadata {
        +offset
        +length
        +sha256
    }
    class WrittenChunk {
        +path
        +length
        +sha256
    }
    class ChunkCommitResult {
        +state : TusSessionState
        +committed : bool
        +is_final_commit : bool
    }
    class TusUploadSessionArgs {
        +session_dir
        +session_id
        +total_size
        +valkey_client : ValkeyTusClient
        +lock_factory : DistributedLockFactory
    }
    TusSessionState o-- ChunkMetadata
    TusUploadSession ..> TusSessionState
    TusUploadSession ..> ChunkCommitResult
    TusUploadSession ..> WrittenChunk
    TusUploadSession ..> TusUploadSessionArgs
```

### One PATCH lifecycle

The expensive body write is lock-free; only the short metadata read-modify-write
is serialized via the distributed lock, so concurrent replicas never corrupt
each other.

```mermaid
sequenceDiagram
    autonumber
    participant H as tus_upload_part
    participant S as TusUploadSession
    participant V as Valkey
    participant D as shared FS

    H->>S: ensure_initialized()
    S->>V: acquire lock + SET state if missing
    H->>S: write_temp_chunk(offset, reader)
    loop async for chunk in reader.read()
        S->>D: write chunk_off.tok.tmp (lock-free) + sha256
    end
    S-->>H: WrittenChunk(path, length, sha256)
    H->>S: commit_chunk(offset, path, length, sha256)
    Note over S,V: distributed lock window (short)
    alt status == COMPLETED
        S-->>H: no-op (late duplicate from another replica)
    else duplicate (same offset, length, sha256)
        S-->>H: committed=False (idempotent)
    else conflict (same offset, different content)
        S-->>H: ChunkConflictError 409
    else new chunk
        S->>D: atomic rename tmp to chunk_off.dat
        S->>V: update TusSessionState with new record
        S-->>H: ChunkCommitResult(is_final_commit?)
    end
    opt is_final_commit
        H->>S: assemble(target)
        S->>D: concat chunks in offset order
        S->>V: mark status=COMPLETED
        H->>S: cleanup()
        S->>D: unlink chunk files (state kept for TTL)
    end
    H-->>H: 204 No Content
```

## Locking & HA semantics

**Session state is NOT held in memory and NOT on the shared filesystem.**
`TusUploadSession` is stateless in-process — every operation reads/writes the
TusSessionState in Valkey, and chunk payload files are content-addressed.

This makes it safe for the production topology of **multiple storage-proxy
processes/replicas** sharing an NFS mount: no in-process state to synchronize,
no dependency on filesystem lock semantics, no dependency on NFS attribute-cache
coherence.

How replicas coordinate:

- **Source of truth = Valkey**, fetched per request (no in-process cache).
- **Per-session distributed lock** (RedisLock via `DistributedLockFactory`)
  serializes the metadata read-modify-write window. The factory pattern mirrors
  the manager's `DistributedLockFactory` — the lock backend lives in
  `RootContext.tus_lock_factory` and is injected into the engine.
- **Atomic rename** (`Path.replace`) promotes a temp chunk file
  (`chunk_<offset>.<token>.tmp`) to its canonical name (`chunk_<offset>.dat`).
  Readers see old-or-new, never a partial file.
- **Per-offset content-addressed chunk files** — duplicates are idempotent
  no-ops, mismatched content surfaces as a 409 `ChunkConflictError`.
- **Stateless JWT upload tokens** — any PATCH may land on any replica with no
  session affinity, and still resolves to the same Valkey key and same chunk
  file path.

Stale or abandoned sessions are reclaimed by the Valkey state's TTL (24h), so
there is no separate GC sweep.

Resolves BA-6154, BA-6155.

### DIFF
--- a/changes/11766.feature.md
+++ b/changes/11766.feature.md
@@ -1,0 +1,1 @@
+Introduce `TusUploadSession`, a concurrency-safe upload engine for resumable (TUS) uploads that keeps session metadata and a per-session lock in Valkey (via `ValkeyTusClient`) while storing chunk payloads on the shared filesystem, so multiple storage-proxy replicas can accept chunks without corruption and without relying on filesystem lock semantics.
--- a/src/ai/backend/common/clients/valkey_client/valkey_tus/client.py
+++ b/src/ai/backend/common/clients/valkey_client/valkey_tus/client.py
@@ -1,0 +1,116 @@
+"""
+Valkey client for TUS resumable-upload sessions.
+
+Stores per-session upload metadata (the source of truth) on the Glide-based
+valkey connection so that multiple Storage Proxy replicas share one view of an
+upload's progress without relying on shared-filesystem semantics. This client is
+metadata-only; the per-session distributed lock that serializes the
+read-modify-write window is a separate :class:`DistributedLockFactory` resource
+(see :mod:`ai.backend.storage.services.upload.tus_session`). The payload bytes
+stay on the shared filesystem; only this small metadata lives in Valkey.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Final, Self
+
+from glide import ExpirySet, ExpiryType
+
+from ai.backend.common.clients.valkey_client.client import (
+    AbstractValkeyClient,
+    create_valkey_client,
+)
+from ai.backend.common.exception import BackendAIError
+from ai.backend.common.metrics.metric import DomainType, LayerType
+from ai.backend.common.resilience import (
+    BackoffStrategy,
+    MetricArgs,
+    MetricPolicy,
+    Resilience,
+    RetryArgs,
+    RetryPolicy,
+)
+from ai.backend.common.types import TusSessionId, ValkeyTarget
+from ai.backend.logging import BraceStyleAdapter
+
+log = BraceStyleAdapter(logging.getLogger(__spec__.name))
+
+valkey_tus_resilience = Resilience(
+    policies=[
+        MetricPolicy(MetricArgs(domain=DomainType.VALKEY, layer=LayerType.VALKEY_TUS)),
+        RetryPolicy(
+            RetryArgs(
+                max_retries=3,
+                retry_delay=0.1,
+                backoff_strategy=BackoffStrategy.FIXED,
+                non_retryable_exceptions=(BackendAIError,),
+            )
+        ),
+    ]
+)
+
+_STATE_KEY_PREFIX: Final = "tus.upload.session"  # tus.upload.session:{session_id}
+
+# Stale (never-completed / abandoned) sessions are reclaimed by this TTL, so no
+# separate GC sweep is needed.
+_DEFAULT_STATE_TTL_SECONDS: Final = 24 * 60 * 60
+
+
+class ValkeyTusClient:
+    """Valkey-backed metadata store for TUS uploads."""
+
+    _client: AbstractValkeyClient
+
+    def __init__(self, client: AbstractValkeyClient) -> None:
+        self._client = client
+
+    @classmethod
+    async def create(
+        cls,
+        valkey_target: ValkeyTarget,
+        *,
+        db_id: int,
+        human_readable_name: str,
+    ) -> Self:
+        client = create_valkey_client(
+            valkey_target=valkey_target,
+            db_id=db_id,
+            human_readable_name=human_readable_name,
+        )
+        await client.connect()
+        return cls(client=client)
+
+    @valkey_tus_resilience.apply()
+    async def close(self) -> None:
+        await self._client.disconnect()
+
+    @staticmethod
+    def _state_key(session_id: TusSessionId) -> str:
+        return f"{_STATE_KEY_PREFIX}:{session_id}"
+
+    @valkey_tus_resilience.apply()
+    async def get_session_state(self, session_id: TusSessionId) -> bytes | None:
+        """Return the raw serialized session state, or ``None`` if absent."""
+        async with self._client.client() as conn:
+            return await conn.get(self._state_key(session_id))
+
+    @valkey_tus_resilience.apply()
+    async def set_session_state(
+        self,
+        session_id: TusSessionId,
+        payload: str | bytes,
+        *,
+        ttl_seconds: int = _DEFAULT_STATE_TTL_SECONDS,
+    ) -> None:
+        async with self._client.client() as conn:
+            await conn.set(
+                self._state_key(session_id),
+                payload,
+                expiry=ExpirySet(ExpiryType.SEC, ttl_seconds),
+            )
+
+    @valkey_tus_resilience.apply()
+    async def delete_session_state(self, session_id: TusSessionId) -> None:
+        async with self._client.client() as conn:
+            await conn.delete([self._state_key(session_id)])
--- a/src/ai/backend/storage/server.py
+++ b/src/ai/backend/storage/server.py
@@ -26,11 +26,13 @@ from aiohttp import web
 from aiohttp.typedefs import Middleware
 from setproctitle import setproctitle
 
+from ai.backend.common import redis_helper
 from ai.backend.common.bgtask.bgtask import BackgroundTaskManager
 from ai.backend.common.clients.valkey_client.valkey_artifact.client import (
     ValkeyArtifactDownloadTrackingClient,
 )
 from ai.backend.common.clients.valkey_client.valkey_bgtask.client import ValkeyBgtaskClient
+from ai.backend.common.clients.valkey_client.valkey_tus import ValkeyTusClient
 from ai.backend.common.clients.valkey_client.valkey_volume_stats import ValkeyVolumeStatsClient
 from ai.backend.common.config import (
     ConfigurationError,
@@ -42,6 +44,8 @@ from ai.backend.common.defs import (
     REDIS_BGTASK_DB,
     REDIS_STATISTICS_DB,
     REDIS_STREAM_DB,
+    REDIS_STREAM_LOCK,
+    REDIS_TUS_DB,
     RedisRole,
 )
 from ai.backend.common.etcd import AsyncEtcd
@@ -111,6 +115,7 @@ from .plugin import (
     StorageManagerWebappPluginContext,
     StoragePluginContext,
 )
+from .services.upload.lock import create_tus_lock_factory
 from .storages.storage_pool import StoragePool
 from .volumes.noop import init_noop_volume
 from .volumes.pool import VolumePool
@@ -644,6 +649,24 @@ async def server_main(
         )
         storage_init_stack.push_async_callback(valkey_artifact_client.close)
 
+        # TUS upload session metadata store (Valkey) and the per-session lock
+        # factory, kept as separate resources. The lock backend is a dedicated
+        # redis-for-lock connection (redis.asyncio, required by RedisLock),
+        # provisioned and owned here; the factory closes over it.
+        valkey_tus_client = await ValkeyTusClient.create(
+            valkey_target=valkey_target,
+            db_id=REDIS_TUS_DB,
+            human_readable_name=f"storage-proxy-tus-uploader-{pidx}",
+        )
+        storage_init_stack.push_async_callback(valkey_tus_client.close)
+        tus_lock_redis = redis_helper.get_redis_object_for_lock(
+            redis_config.to_redis_profile_target().profile_target(RedisRole.STREAM_LOCK),
+            name=f"storage-proxy-tus-lock-{pidx}",
+            db=REDIS_STREAM_LOCK,
+        )
+        storage_init_stack.push_async_callback(tus_lock_redis.close)
+        tus_lock_factory = create_tus_lock_factory(tus_lock_redis)
+
         # Initialize health probe
         health_probe = HealthProbe(options=HealthProbeOptions(check_interval=60))
         # Liveness-registered: also surfaced in readiness — connection-stuck
@@ -724,6 +747,8 @@ async def server_main(
             },
             manager_client_pool=manager_client_pool,
             valkey_artifact_client=valkey_artifact_client,
+            valkey_tus_client=valkey_tus_client,
+            tus_lock_factory=tus_lock_factory,
             health_probe=health_probe,
             volume_stats_observer=volume_stats_observer,
             volume_stats_state=volume_stats_state,
--- a/src/ai/backend/storage/services/upload/tus_session.py
+++ b/src/ai/backend/storage/services/upload/tus_session.py
@@ -1,0 +1,301 @@
+"""Valkey-backed TUS upload session.
+
+Session metadata lives in Valkey (via :class:`ValkeyTusClient`), guarded by a
+per-session distributed lock from an injected :class:`DistributedLockFactory`.
+Chunk payload bytes live on the shared filesystem, content-addressed by
+``(offset, sha256)``::
+
+    <session_dir>/
+      chunks/
+        chunk_<offset>.dat
+        chunk_<offset>.<token>.tmp
+"""
+
+from __future__ import annotations
+
+import asyncio
+import hashlib
+import logging
+import os
+from pathlib import Path
+
+import aiofiles
+import aiofiles.os
+
+from ai.backend.common.clients.valkey_client.valkey_tus import ValkeyTusClient
+from ai.backend.common.lock import AbstractDistributedLock, DistributedLockFactory
+from ai.backend.common.types import StreamReader, TusSessionId
+from ai.backend.logging import BraceStyleAdapter
+from ai.backend.storage.errors.upload import (
+    ChunkConflictError,
+    UploadSessionCorruptedError,
+)
+
+from .lock import LOCK_KEY_PREFIX, LOCK_LIFETIME_SECONDS
+from .types import (
+    ChunkCommitResult,
+    ChunkMetadata,
+    TusSessionState,
+    TusUploadSessionArgs,
+    UploadStatus,
+    WrittenChunk,
+)
+from .utils import (
+    CHUNKS_DIRNAME,
+    committed_chunk_path,
+    staging_path,
+    temp_chunk_path,
+    unlink_quietly,
+)
+
+log = BraceStyleAdapter(logging.getLogger(__spec__.name))
+
+
+class TusUploadSession:
+    """
+    A TUS upload session whose metadata lives in Valkey and whose chunk payloads
+    live under ``session_dir`` on the shared filesystem.
+
+    Instances are cheap to construct; all state lives in Valkey (metadata) and on
+    disk (chunk bytes). The metadata read-modify-write window is serialized across
+    replicas by a per-session distributed lock obtained from the injected
+    :class:`DistributedLockFactory`.
+    """
+
+    _session_dir: Path
+    _session_id: TusSessionId
+    _total_size: int
+    _valkey: ValkeyTusClient
+    _lock_factory: DistributedLockFactory
+
+    def __init__(self, args: TusUploadSessionArgs) -> None:
+        self._session_dir = args.session_dir
+        self._session_id = args.session_id
+        self._total_size = args.total_size
+        self._valkey = args.valkey_client
+        self._lock_factory = args.lock_factory
+
+    @property
+    def session_dir(self) -> Path:
+        return self._session_dir
+
+    @property
+    def session_id(self) -> TusSessionId:
+        return self._session_id
+
+    @property
+    def chunks_dir(self) -> Path:
+        return self._session_dir / CHUNKS_DIRNAME
+
+    @property
+    def _lock_key(self) -> str:
+        return f"{LOCK_KEY_PREFIX}:{self._session_id}"
+
+    def _lock(self) -> AbstractDistributedLock:
+        return self._lock_factory(self._lock_key, LOCK_LIFETIME_SECONDS)
+
+    async def _load_state(self) -> TusSessionState | None:
+        raw = await self._valkey.get_session_state(self._session_id)
+        if raw is None:
+            return None
+        return TusSessionState.model_validate_json(raw)
+
+    async def _store_state(self, state: TusSessionState) -> None:
+        await self._valkey.set_session_state(self._session_id, state.model_dump_json())
+
+    async def ensure_initialized(self) -> None:
+        """Create an initial Valkey state for this session if missing.
+
+        If the state is absent but ``chunks_dir`` still holds files, those are
+        leftovers from a prior session whose Valkey state expired by TTL. The
+        new state would not know about them and could silently mix their bytes
+        into the assembled output, so drop them before starting fresh.
+        """
+        async with self._lock():
+            if await self._load_state() is not None:
+                return
+            await self._delete_chunk_files()
+            await self._store_state(TusSessionState.empty(self._session_id, self._total_size))
+
+    async def exists(self) -> bool:
+        """Return True iff a Valkey state currently exists for this session.
+
+        Valkey is the source of truth — a missing state means the session was
+        never registered or its TTL elapsed, regardless of what is on disk.
+        """
+        return (await self._load_state()) is not None
+
+    async def read_state(self) -> TusSessionState:
+        state = await self._load_state()
+        if state is None:
+            return TusSessionState.empty(self._session_id, self._total_size)
+        return state
+
+    def open_temp_chunk(self, offset: int) -> Path:
+        """
+        Allocate a unique temp file path under chunks/ for a chunk at
+        ``offset``. The caller writes payload bytes into the returned path and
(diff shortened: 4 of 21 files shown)


changes/11766.feature.md
jopemachine, line 1: Rename to feature type

src/ai/backend/storage/services/upload/tus_session.py
```diff
@@ -0,0 +1,454 @@
...
+        return self.offset + self.length
+
+
+class SessionState(BackendAISchema):
+    model_config = ConfigDict(frozen=True, populate_by_name=True)
+
+    session_id: str = Field(alias="session")
+    total_size: int
+    received: tuple[ChunkRecord, ...] = ()
+    status: str = "pending"
```
jopemachine: Create StrEnum instead of raw string

src/ai/backend/storage/services/upload/tus_session.py
```diff
@@ -0,0 +1,454 @@
...
+from pydantic import ConfigDict, Field, field_validator
+
+from ai.backend.common.exception import BackendAIError
+from ai.backend.common.types import BackendAISchema, SchemaValidationFailureInfo, StreamReader
+from ai.backend.storage.errors.upload import (
+    ChunkConflictError,
+    UploadSessionCorruptedError,
+)
+
+INFO_FILENAME = "info.json"
```
jopemachine: _INFO_FILENAME

jopemachine: Done — privatized the module constants (`_INFO_FILENAME`, `_LOCK_FILENAME`, `_CHUNKS_DIRNAME`, `_CHUNK_FILENAME_FMT`, `_TEMP_SUFFIX_BYTES`); tests now use `session.info_path` instead of importing the constant.

changes/11766.feature.md
jopemachine: Done — renamed `changes/11766.enhance.md` → `changes/11766.feature.md`.

src/ai/backend/storage/services/upload/tus_session.py
jopemachine: Done — added an `UploadStatus(StrEnum)` (PENDING/COMPLETED); `SessionState.status` and all comparisons now use it instead of raw strings.

src/ai/backend/storage/services/upload/tus_session.py
```diff
@@ -0,0 +1,460 @@
...
+
+class TusUploadSession:
+    """
+    A TUS upload session backed by a chunks-per-offset directory layout.
+
+    Instances are cheap to construct; all state lives on disk under ``root``.
+    """
+
+    def __init__(self, root: Path, session_id: str, total_size: int) -> None:
+        self._root = root
```
jopemachine: _root -> _session_dir

src/ai/backend/storage/services/upload/tus_session.py
```diff
@@ -0,0 +1,460 @@
...
+        self._session_id = session_id
+        self._total_size = total_size
+
+    @property
+    def root(self) -> Path:
```
jopemachine: session_dir

src/ai/backend/storage/services/upload/tus_session.py
```diff
@@ -0,0 +1,460 @@
...
+
+
+class TusUploadSession:
+    """
+    A TUS upload session backed by a chunks-per-offset directory layout.
+
+    Instances are cheap to construct; all state lives on disk under ``root``.
+    """
+
+    def __init__(self, root: Path, session_id: str, total_size: int) -> None:
```
jopemachine: Make TusUploadSessionArgs type

src/ai/backend/storage/services/upload/tus_session.py
```diff
@@ -0,0 +1,460 @@
...
+
+    @property
+    def end(self) -> int:
+        return self.offset + self.length
+
+
+class SessionState(BackendAISchema):
+    model_config = ConfigDict(frozen=True, populate_by_name=True)
+
+    session_id: str = Field(alias="session")
```
jopemachine: Why this alias needed?

src/ai/backend/storage/services/upload/tus_session.py
jopemachine: Done — renamed the `root` property to `session_dir` (and updated the test usages).

src/ai/backend/storage/services/upload/tus_session.py
jopemachine: Done — renamed the `_root` attribute to `_session_dir` throughout the class.

src/ai/backend/storage/services/upload/tus_session.py
jopemachine: Done — added a frozen `TusUploadSessionArgs` dataclass (`session_dir` / `session_id` / `total_size`); `TusUploadSession.__init__` now takes it, and all call sites (PATCH/HEAD/status handlers + tests) construct it explicitly.

src/ai/backend/common/clients/valkey_client/valkey_tus/client.py
```diff
@@ -0,0 +1,147 @@
...
+    @staticmethod
+    def _state_key(session_id: str) -> str:
+        return f"{_STATE_KEY_PREFIX}:{session_id}"
+
+    @staticmethod
+    def _lock_key(session_id: str) -> str:
+        return f"{_LOCK_KEY_PREFIX}:{session_id}"
+
+    @valkey_tus_resilience.apply()
+    async def get_state(self, session_id: str) -> bytes | None:
```
jopemachine: get_session_state

src/ai/backend/storage/services/upload/tus_session.py
```diff
@@ -0,0 +1,410 @@
...
+    @property
+    def chunks_dir(self) -> Path:
+        return self._session_dir / _CHUNKS_DIRNAME
+
+    async def _load_state(self) -> SessionState | None:
+        raw = await self._valkey.get_state(self._session_id)
+        if raw is None:
+            return None
+        # BackendAISchema.model_validate_json maps a pydantic ValidationError to
+        # UploadSessionCorruptedError via SessionState.build_validation_error.
```
jopemachine: Remove useless comment

src/ai/backend/storage/services/upload/tus_session.py
```diff
@@ -0,0 +1,410 @@
...
+        path to :meth:`commit_chunk`.
+        """
+        self.chunks_dir.mkdir(parents=True, exist_ok=True)
+        token = _random_token()
+        return _TempChunkFile(
+            path=self.chunks_dir / f"chunk_{offset}.{token}.tmp",
+            offset=offset,
+        )
+
+    async def write_temp_chunk(
```
jopemachine: write_temp_chunk seems unused

src/ai/backend/storage/services/upload/tus_session.py
jopemachine: Removed the alias. It was a carryover from the old on-disk `info.json` format that used a `session` key. The session metadata now lives in Valkey, a store we fully control with no migration concern, so the field is just `session_id` and serializes as `session_id`. (`populate_by_name` dropped too.)

src/ai/backend/common/clients/valkey_client/valkey_tus/client.py
jopemachine: Renamed `get_state`→`get_session_state` and `set_state`→`set_session_state` (mirroring `valkey_session.get_session_data`). Also dropped the unused `delete_state` (cleanup keeps the completed marker for the TTL to reclaim, so nothing deletes it).

src/ai/backend/storage/services/upload/tus_session.py
jopemachine: It is used — the PATCH handler (`tus_upload_part` in #11767) drains the request body through `write_temp_chunk` before `commit_chunk`, and it has `TestWriteTempChunk` coverage. It reads as unused from #11766 alone because the handler that calls it lands in the next PR of the stack.

src/ai/backend/storage/services/upload/tus_session.py
jopemachine: Removed.

src/ai/backend/storage/server.py
```diff
@@ -644,6 +649,24 @@ async def server_main(
...
         storage_init_stack.push_async_callback(valkey_artifact_client.close)
 
+        # TUS upload session metadata store (Valkey) and the per-session lock
+        # factory, kept as separate resources. The lock backend is a dedicated
+        # redis-for-lock connection (redis.asyncio, required by RedisLock),
+        # provisioned and owned here; the factory closes over it.
+        valkey_tus_client = await ValkeyTusClient.create(
+            valkey_target=valkey_target,
+            db_id=REDIS_TUS_DB,
+            human_readable_name=f"storage-proxy-tus-{pidx}",
```
jopemachine: storage-proxy-tus-uploader-{pidx}

src/ai/backend/storage/services/upload/tus_session.py
```diff
@@ -0,0 +1,456 @@
...
+
+# Random suffix on `chunks/chunk_<offset>.<token>.tmp` so concurrent writers
+# from different storage-proxy replicas never share a temp path.
+def _random_token() -> str:
+    return secrets.token_hex(_TEMP_SUFFIX_BYTES)
+
+
+# Best-effort delete used in rollback/cleanup paths where the file may have
+# already been removed by a concurrent worker or never created.
+def _unlink_quietly(path: Path) -> None:
```
jopemachine:
```
def _unlink_quietly(path: Path) -> None:
     """
     Best-effort delete used in rollback/cleanup paths where the file may have
     already been removed by a concurrent worker or never created.
     """
```

tests/unit/storage/services/upload/test_tus_session.py
```diff
@@ -0,0 +1,174 @@
...
+        )
+        assert state.find_at_offset(500) is None
+
+
+class TestMissingRanges:
+    @pytest.mark.parametrize(
+        ("total_size", "records", "expected"),
+        [
+            # Empty session: the whole declared size is missing.
+            (1000, [], [(0, 1000)]),
```
jopemachine: Introduce type

jopemachine: Introduced typed case dataclasses (`_CommittedOffsetCase`, `_MissingRangesCase`, `_ProgressPercentCase`) and replaced all three parametrize blocks' raw tuples with them.

src/ai/backend/storage/server.py
jopemachine: Renamed to `storage-proxy-tus-uploader-{pidx}` for clarity.

src/ai/backend/storage/services/upload/tus_session.py
jopemachine: Moved the inline `#` comments to proper docstrings on `_unlink_quietly`, and consistently on the two sibling helpers (`_TempChunkFile`, `_random_token`).

src/ai/backend/storage/services/upload/tus_session.py
```diff
@@ -0,0 +1,476 @@
...
+    async def _load_state(self) -> SessionState | None:
+        raw = await self._valkey.get_session_state(self._session_id)
+        if raw is None:
+            return None
+        return SessionState.model_validate_json(raw)
+
+    async def _store_state(self, state: SessionState) -> None:
+        await self._valkey.set_session_state(self._session_id, state.model_dump_json())
+
+    async def ensure_initialized(self) -> SessionState:
```
jopemachine: Returning type is not used anywhere

tests/unit/storage/services/upload/test_tus_session.py
```diff
@@ -0,0 +1,210 @@
...
+from ai.backend.common.types import TusSessionId
+from ai.backend.storage.errors.upload import UploadSessionCorruptedError
+from ai.backend.storage.services.upload.tus_session import (
+    ChunkMetadata,
+    SessionState,
+    UploadStatus,
+)
+
+
+def _rec(offset: int, length: int, sha256: str = "x") -> ChunkMetadata:
```
jopemachine: Don't abbreviate

src/ai/backend/storage/services/upload/tus_session.py
jopemachine: Changed `ensure_initialized` to return `None` — all production callers discard the value (`await session.ensure_initialized()` only). Updated the two tests in `test_tus_upload_session.py` to verify the persisted state via a follow-up `read_state()` instead of the (now-removed) return.

tests/unit/storage/services/upload/test_tus_session.py
jopemachine: Renamed `_rec` → `_make_chunk_metadata` (29 occurrences in test_tus_session.py).

src/ai/backend/storage/services/upload/tus_session.py
```diff
@@ -0,0 +1,261 @@
...
+        async with self._lock():
+            state = await self._load_state()
+            if state is None:
+                raise UploadSessionCorruptedError("no session metadata to assemble")
+            if state.status == UploadStatus.COMPLETED:
+                return
+            if state.committed_offset < state.total_size:
+                raise UploadSessionCorruptedError("cannot assemble: not all chunks received")
+            await asyncio.to_thread(self._assemble_chunks_to_disk, state, target_path)
+            await self._store_state(state.with_status(UploadStatus.COMPLETED))
```
jopemachine: How about just creating a method like `complete_session_state` instead of `with_status`, `with_record`?

jopemachine: Refactored `SessionState` to be mutable rather than frozen, so the with_\* builders are no longer needed.

- Dropped `ConfigDict(frozen=True)` on `SessionState` (`ChunkMetadata` stays frozen as a value type).
- Switched `received` from `tuple[ChunkMetadata, ...]` to `list[ChunkMetadata]` so mutation is a plain `.append()` + `.sort()`.
- Inlined the mutations at their callsites:
  - `commit_chunk`: `state.received.append(record); state.received.sort(key=lambda r: r.offset)` then `_store_state(state)`.
  - `assemble`: `state.status = UploadStatus.COMPLETED` then `_store_state(state)`.
- Dropped the unused `updated_at` field (0 readers across the whole stack) and the now-redundant `TestWithModifiers` test class (sort is still covered by `test_validate_sorts_received_by_offset`; the COMPLETED transition is covered by the engine integration tests).

src/ai/backend/storage/services/upload/types.py
```diff
@@ -0,0 +1,121 @@
...
+    offset: int
+    length: int
+    sha256: str
+
+    @property
+    def end(self) -> int:
+        return self.offset + self.length
+
+
+class SessionState(BackendAISchema):
```
jopemachine: -> TusSessionState

src/ai/backend/storage/services/upload/tus_session.py
```diff
@@ -0,0 +1,295 @@
...
+
+        If the state is absent but ``chunks_dir`` still holds files, those are
+        leftovers from a prior session whose Valkey state expired by TTL. The
+        new state would not know about them and could silently mix their bytes
+        into the assembled output, so drop them before starting fresh.
+        """
+        async with self._lock():
+            if await self._load_state() is not None:
+                return
+            await asyncio.to_thread(self._delete_chunk_files)
```
fregataa: Why is `to_thread()` used?

src/ai/backend/storage/services/upload/tus_session.py
```diff
@@ -0,0 +1,295 @@
...
+    _total_size: int
+    _valkey: ValkeyTusClient
+    _lock_factory: DistributedLockFactory
+
+    def __init__(self, args: TusUploadSessionArgs) -> None:
+        self._session_dir = args.session_dir
+        self._session_id = args.session_id
+        self._total_size = args.total_size
+        self._valkey = args.valkey_client
+        self._lock_factory = args.lock_factory
```
fregataa, line 76: Why is `lock_factory` passed as a parameter?

src/ai/backend/storage/services/upload/tus_session.py
```diff
@@ -0,0 +1,295 @@
...
+    UploadStatus,
+    WrittenChunk,
+)
+from .utils import (
+    CHUNKS_DIRNAME,
+    committed_chunk_path,
+    staging_path,
+    temp_chunk_path,
+    unlink_quietly,
+)
```
fregataa, line 49: They don't need to be in a separate module

src/ai/backend/storage/services/upload/tus_session.py
```diff
@@ -0,0 +1,295 @@
...
+        async with self._lock():
+            state = await self._load_state()
+            if state is None:
+                raise UploadSessionCorruptedError("no session metadata to assemble")
+            if state.status == UploadStatus.COMPLETED:
+                return
+            if state.committed_offset < state.total_size:
+                raise UploadSessionCorruptedError("cannot assemble: not all chunks received")
+            await asyncio.to_thread(self._assemble_chunks_to_disk, state, target_path)
+            state.status = UploadStatus.COMPLETED
```
fregataa: let's add `set_complete()` method

fregataa (commented):
1. Does a client of tus-session need to call `write_temp_chunk()`, `commit_chunk()` and `assemble()` separately?
2. Are all chunk saved in valkey till complete upload?
3. Is `commit_chunk()` called in `write_temp_chunk()`? The PR description and codes are different

jopemachine:
> 1. Does a client of tus-session need to call `write_temp_chunk()`, `commit_chunk()` and `assemble()` separately?
> 2. Are all chunk saved in valkey till complete upload?
> 3. Is `commit_chunk()` called in `write_temp_chunk()`? The PR description and codes are different

1. Yes please see #11767.
2. No only metadata (TusSessionState) were stored in redis, file chunks should be stored in the storage.
3. PR description was outdated. I updated them.

src/ai/backend/storage/services/upload/tus_session.py
jopemachine: I'd prefer not to centralize all utility methods and types into a single file.

src/ai/backend/storage/services/upload/tus_session.py
jopemachine: Because metadata updates in Redis must be protected by a lock to prevent concurrent reads and writes from multiple Storage Proxy instances.

src/ai/backend/storage/services/upload/tus_session.py
jopemachine: If we process everything on a single thread, storage proxy requests may be blocked when storage operations take a long time, especially with NFS-backed storage.

For consistency, this changes the implementation to use aiofiles.

src/ai/backend/storage/services/upload/tus_session.py
```diff
@@ -0,0 +1,295 @@
...
+        - Same ``(offset, length, sha256)`` already present → idempotent;
+          ``chunk_path`` is removed and ``committed=False``.
+        - Conflicting record at the same offset → ``ChunkConflictError``;
+          ``chunk_path`` is removed.
+        - Already-completed session → idempotent no-op (a late duplicate that
+          raced completion on another replica); ``chunk_path`` is removed.
+        - Otherwise the chunk file is renamed into place and the record added.
+        """
+        target = committed_chunk_path(self.chunks_dir, offset)
+        async with self._lock():
```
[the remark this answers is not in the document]
jopemachine: All real call sites flow `open_temp_chunk → write_temp_chunk → commit_chunk`, and `open_temp_chunk` already runs `chunks_dir.mkdir(parents=True, exist_ok=True)`, so by the time `commit_chunk` calls `replace()` the directory is guaranteed to exist. Adding a defensive `makedirs` here would cost an extra round-trip per PATCH on NFS without catching any reachable bug, so leaving as-is.

tests/unit/storage/services/upload/test_tus_upload_session.py
```diff
@@ -0,0 +1,360 @@
...
+
+
+async def _commit(session: TusUploadSession, chunk: _Chunk) -> ChunkCommitResult:
+    return await session.commit_chunk(
+        offset=chunk.offset,
+        chunk_path=_write_temp_chunk(session, chunk),
+        length=chunk.length,
+        sha256=chunk.sha256,
+    )
+
```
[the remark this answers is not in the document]
jopemachine: The intent of these concurrency tests is to model multi-replica racing inside the per-session lock window (the lock-free temp-write path is exercised separately by `TestWriteTempChunk`). Using a sync `_write_temp_chunk` helper keeps the focus on the metadata read-modify-write under contention; verifying parallel temp-write throughput is a separate concern.

jopemachine:
This PR has become stale, and I don't believe it is necessary at this time. If it is determined to be needed in the future, I will revisit and implement it then.

PR #11767, opened 2026-05-22
fix(BA-6156): rewire TUS PATCH/HEAD to chunk-based store

jopemachine:
## 📚 Stacked PRs

This PR is part of a 5-PR stack implementing BA-3974 (epic: BA-6153). **Merge in order:**

1. ⬇️ #11766 — `feat(BA-6155): add metadata-driven chunk-store upload session engine`
2. 👉 **#11767** — `fix(BA-6156): rewire TUS PATCH/HEAD to chunk-based store` (**actual user-visible fix**) ← you are here
3. ⬇️ #11768 — `test(BA-6157): add multi-proxy NFS race regression test`
4. ⬇️ #11769 — `feat(BA-6158): support TUS Checksum extension`
5. ⬇️ #11770 — `feat(BA-6159): add /upload/status endpoint + progress headers`

## Summary
- Replace the legacy single-file append in \`tus_upload_part\` with \`TusUploadSession\`-backed per-offset chunk commits. Concurrent PATCH requests from multiple Storage Proxy replicas on a shared NFS mount no longer corrupt each other.
- \`tus_check_session\` now returns \`Upload-Offset\` from \`SessionState.committed_offset\` instead of a raw \`stat()\` of the temp file.
- \`prepare_upload\` (vfs) creates the session directory instead of touching an empty file, matching the new on-disk layout.
- Handler tests rewritten to drive the real \`TusUploadSession\` through \`tmp_path\`.

Wire format on \`/upload\` (PATCH/HEAD/OPTIONS) stays TUS 1.0.0-compatible; existing tus-js-client based callers continue to work unchanged.

Resolves BA-6156. Part of epic BA-6153 (implements BA-3974).


src/ai/backend/storage/api/client.py
```diff
@@ -361,61 +393,106 @@ class Params(TypedDict):
...
+                upload_length=total_size,
+            )
     return web.Response(status=HTTPStatus.NO_CONTENT, headers=headers)
 
 
+def _resolve_tus_upload_session_dir(volume: AbstractVolume, token_data: UploadTokenData) -> Path:
+    return volume.mangle_vfpath(token_data["vfid"]) / ".upload" / token_data["session"]
+
+
+_TUS_HEADER_LIST = "Tus-Resumable, Upload-Length, Upload-Metadata, Upload-Offset, Content-Type"
```
jopemachine, line 473: Inline this

jopemachine: Inlined `volume.mangle_vfpath(token_data["vfid"]) / ".upload" / token_data["session"]` at the two call sites (tus_check_session, tus_upload_part) and removed the helper. Same one-liner is also used in the `tus_upload_status` handler in #11770, so the inline applied there too.

src/ai/backend/storage/api/client.py
```diff
@@ -329,13 +333,44 @@ class Params(TypedDict):
     ) as params:
         token_data = params["token"]
         async with ctx.get_volume(token_data["volume"]) as volume:
-            headers = await prepare_tus_session_headers(request, token_data, volume)
+            session_dir = (
+                volume.mangle_vfpath(token_data["vfid"]) / ".upload" / token_data["session"]
+            )
+            if not session_dir.exists():
+                raise web.HTTPNotFound(
```
jopemachine: Why is it session not found

src/ai/backend/storage/volumes/vfs/__init__.py
```diff
@@ -624,8 +624,8 @@ async def prepare_upload(self, vfid: VFolderID) -> str:
         def _create_target() -> None:
             upload_base_path = vfpath / ".upload"
             upload_base_path.mkdir(exist_ok=True)
-            upload_target_path = upload_base_path / session_id
-            upload_target_path.touch()
+            session_dir = upload_base_path / session_id
+            session_dir.mkdir(parents=True, exist_ok=True)
```
jopemachine, line 628: Why did you patch vfs volume?

src/ai/backend/storage/api/client.py
jopemachine: Right — the `session_dir.exists()` check was a leftover from the pre-Valkey design where the session *was* the directory. In the new model the source of truth is the Valkey state, not the filesystem, so this commit (`fix(BA-6156): authoritatively use Valkey state for session existence`) replaces both HEAD and PATCH guards with `await session.exists()` against Valkey, and has `create_upload_session` write an empty `SessionState` to Valkey right after `prepare_upload`. PATCH no longer calls `ensure_initialized` — the existence check has already verified the state, and a TTL-expired state correctly 404s instead of being silently revived. The `tus_upload_status` handler in #11770 gets the same treatment for consistency.

jopemachine:
This PR has become stale, and I don't believe it is necessary at this time. If it is determined to be needed in the future, I will revisit and implement it then.

PR #11787, opened 2026-05-26
docs: Source the advertised API version from `client.config.API_VERSION`

### LINKED ISSUE #11786: Update advertised API version from v6.20220615 to v9.20250722
## Problem

`docs/index.rst:4` and `docs/install/install-from-package/install-manager.rst:263`
still advertise **API v6.20220615** as the latest. The actual current version is
**v9.20250722** — see:

- `src/ai/backend/client/config.py:31` — `API_VERSION = (9, "20250722")`
- `src/ai/backend/manager/api/rest/app.py:47-62` — `LATEST_REV_DATES` ends at `9: "20250722"`, `LATEST_API_VERSION` evaluates to `"v9.20250722"`.

The stale strings give new readers (and the install guide's sample `curl` output)
a misleading view of what a fresh deployment serves, and they have already drifted
~4 years.

## Proposal

Rather than hand-edit the strings each release, source the value once and let
docs reference it:

1. In `docs/conf.py`, import `API_VERSION` from `ai.backend.client.config` and
   expose it as the Sphinx substitution `|api_version|` via `rst_prolog`.
   (`docs/conf.py:26` already inserts `src/` into `sys.path`, so the import
   resolves at build time.)
2. Replace the hard-coded version string in `docs/index.rst` with `|api_version|`.
3. Replace the `code-block:: console` sample in `install-manager.rst` with a
   `parsed-literal::` block using `|api_version|` and `|release|` (the latter
   already exists via `conf.py:73`).
4. Update `docs/locales/ko/LC_MESSAGES/index.po` to track the new placeholder-based
   msgid; long-term, regenerate via `make gettext`.

## Out of scope

- Sweeping audit for older revision strings (`v7.2023*`, `v8.2024*`) elsewhere
  in `docs/`.
- v2 REST/GQL reference content for v9-only fields.

## Acceptance

- [ ] `grep -rn "v9\.20250722\|v6\.20220615" docs/` returns no hard-coded hits in
      the touched pages.
- [ ] Local `make html` renders the version inline correctly on the homepage and
      the install guide.
- [ ] Korean `.po` entry uses the `|api_version|` placeholder (not a literal
      version string).

rapsealk:
## Summary

- Wire `ai.backend.client.config.API_VERSION` into `docs/conf.py` as the Sphinx
  substitution `|api_version|` so the rendered API version stays in sync with
  the client's source of truth.
- Replace the hard-coded `v6.20220615` string in `docs/index.rst` with
  `|api_version|`.
- Switch the install guide's sample `curl` block from `code-block:: console`
  to `parsed-literal::` so `|api_version|` and `|release|` expand inside it.
- Refresh the Korean `.po` entry to track the new placeholder-based msgid
  (drop stale `#, fuzzy`).

Resolves #11786.

## Test plan

- [ ] `make html` in `docs/` renders `v9.20250722` on the homepage and inside
      the `curl bai-m1:8081` example.
- [ ] No `grep -rn "v6\.20220615" docs/` hits remain.
- [ ] Korean `.po` `msgid` for the homepage version line contains `|api_version|`
      (not a literal version string).

----
📚 Documentation preview 📚: https://sorna--11787.org.readthedocs.build/en/11787/

----
📚 Documentation preview 📚: https://sorna-ko--11787.org.readthedocs.build/ko/11787/


seedspirit (dismissed):
It seems to be okay, @HyeockJinKim  Could you take a look?

docs/index.rst
```diff
@@ -1,7 +1,7 @@
 Backend.AI Documentation
 ========================
 
-**Latest API version: v6.20220615**
+**Latest API version:** |api_version|
```
HyeockJinKim: Could you suggest a name that isn't used as often as this one? Also, please upload it using the version update method described in `release.sh` as a reference.

HyeockJinKim (changes requested):

rapsealk:
### Known limitation: the substitution is sourced from the client, but both render sites describe the manager's value

Flagging this rather than widening the PR — it wants a source change, so I've left it as a follow-up.

`docs/conf.py` imports `API_VERSION` from `ai.backend.client.config`, but the two places the substitution renders both describe the **manager's** advertised value:

- `docs/index.rst` — "Latest API version"
- `docs/install/install-from-package/install-manager.rst` — verbatim the response body of `GET /`, which returns `LATEST_API_VERSION` from `src/ai/backend/manager/api/rest/app.py` (derived from `LATEST_REV_DATES`)

These are not the same constant by design. `src/ai/backend/client/session.py` negotiates `min(server_version, client_version)` and warns *"The server API version is higher than the client"* — the client's value is its supported ceiling, and the code expects the server to lead.

They have diverged before:

| commit | date | client `API_VERSION` | manager `LATEST_API_VERSION` |
|---|---|---|---|
| `43bc1d186` | 2022-06-23 | `(6, '20220315')` | `'v6.20220615'` |
| `b6c4181d7` | 2024-02-28 | `(8, '20240615')` | `'v8.20240315'` |
| `714c1191c` | 2024-02-28 | `(8, '20240315')` | `'v8.20240315'` |

Both constants read `v9.20250722` today, so the rendered output is correct as of this PR. The risk is a release window where the manager bumps first: the docs would then advertise a version the server does not return.

Sourcing `LATEST_API_VERSION` is the correct fix, but importing it into `docs/conf.py` pulls in the manager web stack (`aiohttp_cors`, `BootstrapConfig`, metrics), so it wants `LATEST_REV_DATES` hoisted into a dependency-free module first — a source change, hence the follow-up.

A test asserting the two constants are equal would be the *wrong* fix: divergence is architecturally legitimate, so such a test would fail the next time the server ships an API version ahead of the client.

### On sourcing this through the release script

`scripts/release.sh` updates the **release** version — it writes `VERSION` and freezes `NEXT_RELEASE_VERSION` via `scripts/freeze_release_version.py`. The docs already use that path: `|release|` reads the `VERSION` file, which is what the manager version in the `curl` sample now comes from.

The API version is a separate constant on its own cadence. Routing it through the release script would make releases responsible for a value they don't own, and it would go stale the moment `API_VERSION` changes outside a release. Reading it at docs build time keeps it correct without that coupling — but happy to reconsider if you had a different mechanism in mind.

### Update: the duplication is wider than client-vs-manager

Grepping the API version across source turns up four hard-coded copies, not two:

| file | form |
|---|---|
| `src/ai/backend/client/config.py` | `API_VERSION = (9, "20250722")` |
| `src/ai/backend/manager/api/rest/app.py` | `LATEST_REV_DATES` → `LATEST_API_VERSION` |
| `src/ai/backend/client/cli/v2/helpers.py` | `"api_version": "v9.20250722"` |
| `src/ai/backend/client/v2/config.py` | `api_version: str = "v9.20250722"` |

(Two further hits under `src/ai/backend/common/dto/manager/` are docstring examples, not copies.)

A plain-text `API_VERSION` file beside the root `VERSION` file, read by all of them, would collapse the four copies to one and let `docs/conf.py` read a file instead of importing the client package — the same shape `|release|` already uses. The trade-off to weigh: a shared file makes client/manager divergence impossible, which `session.py`'s `min(server, client)` negotiation deliberately permits, so shipping a server API ahead of the client would no longer be expressible.

HyeockJinKim (approved):

PR #11813, opened 2026-05-27
feat(BA-6134): gate session allocation through agent_resources reserved

HyeockJinKim:
## Summary
- Add a `reserved` column to `agent_resources` and gate resource allocation at the table level: scheduling reserves each kernel's requested slots on its agent at the SCHEDULED transition and enforces `reserved + used <= capacity` atomically (per-kernel, idempotent via the PENDING→SCHEDULED rowcount gate). The RUNNING transition moves the hold from `reserved` to `used`; free/terminate/cancel release reserved/used via a shared helper.
- `allocate_sessions` becomes a pure resource gate (reserve + assign agents in one batch transaction, returns scheduled session ids); the coordinator owns the session status transition (BEP-1030). Remove dead allocation methods, the legacy `ScheduleHandler`/`SokovanHandler` abstractions, and the legacy `status_data.scheduler` failure bookkeeping.
- Add repository-layer tests (real DB) covering reservation, idempotency, capacity gate, batch atomicity, reserved-only release, the full lifecycle, and concurrency/cleanup invariants (no overcommit, no leak, consistency with live allocations).

## Test plan
- [ ] `pants test tests/unit/manager/repositories/scheduler::` (reserved allocation + concurrency)
- [ ] `pants test tests/unit/manager/sokovan/scheduler::` (handler / lifecycle)
- [ ] Live verify via `./bai` CLI: create session → reserved increments at SCHEDULED, moves to used at RUNNING, returns to 0 on termination; near-capacity concurrent requests rejected by the table gate

Resolves BA-6134


fregataa (approved):

src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
```diff
@@ -2368,29 +2204,139 @@ async def _allocate_kernel_resources(
                     ResourceAllocationRow.used_at.is_(None),
                 )
                 .values(used=s.quantity, used_at=sa.func.now())
+                .returning(ResourceAllocationRow.requested)
             )
```
fregataa, line 2208: select with update might be useful here?

PR #11835, opened 2026-05-27
feat(BA-6172): add enabled flag to resource_slot_types for admin opt-out

seedspirit:
## Summary

- Adds `enabled` boolean column (default `true`) to `resource_slot_types`. When `enabled=false`, image-side validators (`ImageSlotTypeRule`, `ResourceLimitRule`) skip the slot, so vendor images (e.g. NGC PyTorch declaring `cuda.device: 1`) still pass on clusters that don't serve that slot.
- Plumbs the policy through a new `SlotTypePolicy(enabled, required)` value object into `SessionSpecContextFetch` and `SessionSpecValidationContext`.
- Replaces the `KeyError`-derived `ValueError` raised by `ResourceSlot.to_humanized` / `from_user_input` / `from_policy` with a typed `UnknownResourceSlotType` (`BackendAIError`, 400).
- Fixes a latent `BinarySize.__format__` error in the shmem-vs-memory rejection message by using `!s` string coercion.

## Test plan

- [ ] ` ./py -m alembic -c alembic.ini upgrade head` applies cleanly; `resource_slot_types.enabled` defaults to `true` on existing rows.
- [ ] Roundtrip: `downgrade -1` then `upgrade head` succeed.
- [ ] Mark a vendor-only slot as disabled (`UPDATE resource_slot_types SET enabled = false WHERE slot_name = 'cuda.device'`) and create a session with an NGC-style image declaring that slot — should succeed instead of 500.
- [ ] Re-enable the slot, confirm normal validation resumes.
- [ ] `pants test --changed-since=main --changed-dependents=transitive` green.

Resolves BA-6172


src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
```diff
@@ -1503,7 +1509,10 @@ async def fetch_session_spec_contexts(
             rg_defaults = None
             resource_group_allow_fractional = False
             known_slot_types: Mapping[SlotName, SlotTypes] = {}
-            required_slot_names = await self._fetch_required_slot_names(db_sess)
+            slot_type_policy = SlotTypePolicy(
+                enabled=await self._fetch_enabled_slot_names(db_sess),
+                required=await self._fetch_required_slot_names(db_sess),
+            )
```
fregataa: We can query all slot types at once and filter them

src/ai/backend/common/types.py
```diff
@@ -1324,7 +1327,7 @@ def to_humanized(self, slot_types: Mapping[str, Any]) -> Mapping[str, str]:
                 if v is not None
             }
         except KeyError as e:
-            raise ValueError(f"Unknown slot type: {e.args[0]!r}") from e
+            raise UnknownResourceSlotType(extra_msg=f"Unknown slot type: {e.args[0]!r}") from e
```
fregataa: I think it is not proper to raise an error in a humanize function when a key doesn't exist

fregataa (changes requested):

src/ai/backend/common/types.py
```diff
@@ -1313,18 +1314,17 @@ def from_user_input(
...
-                k: type(self)._humanize_value(Decimal(v), slot_types[k])
-                for k, v in self.data.items()
-                if v is not None
-            }
-        except KeyError as e:
-            raise ValueError(f"Unknown slot type: {e.args[0]!r}") from e
+        result: dict[str, str] = {}
+        for k, v in self.data.items():
+            slot_type = slot_types.get(k, type(self)._guess_slot_type(k))
+            result[k] = type(self)._humanize_value(Decimal(v), slot_type)
```
fregataa: Is `type(self)` needed here?

src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
```diff
@@ -342,12 +343,17 @@ async def _fetch_scaling_group_with_slot_inventory(
...
-    async def _fetch_required_slot_names(self, db_sess: SASession) -> frozenset[SlotName]:
-        stmt = sa.select(ResourceSlotTypeRow.slot_name).where(
-            ResourceSlotTypeRow.required.is_(True)
+    async def _fetch_slot_type_policy(self, db_sess: SASession) -> SlotTypePolicy:
+        stmt = sa.select(
+            ResourceSlotTypeRow.slot_name,
+            ResourceSlotTypeRow.enabled,
+            ResourceSlotTypeRow.required,
+        )
+        rows = (await db_sess.execute(stmt)).all()
```
fregataa, line 357: Let's add condition here, enabled is true or required is true

fregataa (approved):

PR #11838, opened 2026-05-27
fix(BA-6205,BA-6206): allow empty email and null domain for VFolder invitations

fregataa:
## Summary
- Drop falsy guards on `inviter_email` / `requester_email` / `user.domain_name` across `invite`, `accept`, `reject`, `update`, `list`, `list-sent`, and `leave` flows so accounts whose `User.email` is an empty string (and `User.domain_name` may be NULL) can participate.
- `get_by_id_validated` short-circuits when the requester is the vfolder owner, so an empty `domain_name` no longer blocks an owner from acting on their own vfolder.
- `get_user_by_email` no longer drops users with `domain_name = NULL`; return type widened to `tuple[uuid.UUID, str | None] | None`.

Resolves BA-6205, BA-6206

## Test plan
- [x] Added `TestEmptyEmailAccountInvitationScenarios` covering send / accept / reject / update / list / list-sent / leave with empty email and NULL domain.
- [x] `pants lint --changed-since=origin/main` passes locally.
- [x] `pants check` / `pants test` for the touched files pass locally.

### DIFF
--- a/src/ai/backend/manager/repositories/vfolder/repository.py
+++ b/src/ai/backend/manager/repositories/vfolder/repository.py
@@ -190,8 +190,6 @@ class VfolderRepository:
             if not vfolder_row:
                 raise VFolderNotFound()
 
-            # Owner can always act on their own vfolder regardless of domain.
-            # Required for SSO-created accounts whose domain_name may be NULL or empty.
             if vfolder_row.user == user_id:
                 return self._vfolder_row_to_data(vfolder_row)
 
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/repositories/vfolder/repository.py
```diff
@@ -1236,17 +1239,16 @@ async def check_user_has_vfolder_permission(
             return (count or 0) > 0
 
     @vfolder_repository_resilience.apply()
-    async def get_user_by_email(self, email: str) -> tuple[uuid.UUID, str] | None:
+    async def get_user_by_email(self, email: str) -> tuple[uuid.UUID, str | None] | None:
```
jopemachine, line 1242: It would be better to define a dedicated return type instead of using tuple[uuid.UUID, str | None].

jopemachine (approved):

PR #11841, opened 2026-05-27
feat(BA-6175): add role_preset and role_permission_preset tables

fregataa:
## Summary
- Add `role_presets` table (id, name, scope_type, auto_apply, timestamps) with a partial unique index ensuring at most one `auto_apply=true` preset per scope_type.
- Add `role_permission_presets` table holding `(entity_type, operation)` pairs per preset, with `ON DELETE CASCADE` FK to `role_presets`.

## Test plan
- [x] `alembic upgrade head` creates both tables on a fresh DB
- [x] `alembic downgrade -1` drops them cleanly

Resolves BA-6175

### DIFF
--- a/src/ai/backend/manager/models/rbac_models/role_permission_preset/row.py
+++ b/src/ai/backend/manager/models/rbac_models/role_permission_preset/row.py
@@ -1,11 +1,12 @@
 from __future__ import annotations
 
-import uuid
 from datetime import datetime
 
 import sqlalchemy as sa
 from sqlalchemy.orm import Mapped, mapped_column
 
+from ai.backend.common.identifier.role_permission_preset import RolePermissionPresetID
+from ai.backend.common.identifier.role_preset import RolePresetID
 from ai.backend.manager.data.permission.types import (
     EntityType,
     OperationType,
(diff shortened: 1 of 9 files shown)


src/ai/backend/manager/models/rbac_models/role_permission_preset/row.py
```diff
@@ -0,0 +1,48 @@
...
+    __table_args__ = (
+        sa.UniqueConstraint(
+            "role_preset_id",
+            "entity_type",
+            "operation",
+            name="uq_role_permission_presets_preset_entity_op",
+        ),
+    )
+
+    id: Mapped[uuid.UUID] = mapped_column(
```
jopemachine: As with the other types, it would be better to provide a dedicated RolePermissionPresetID type as well.

src/ai/backend/manager/models/rbac_models/role_preset/row.py
```diff
@@ -0,0 +1,54 @@
...
+    __tablename__ = "role_presets"
+    __table_args__ = (
+        sa.Index(
+            "ix_role_presets_scope_type_auto_apply",
+            "scope_type",
+            postgresql_where=sa.text("auto_apply IS TRUE"),
+        ),
+    )
+
+    id: Mapped[uuid.UUID] = mapped_column(
```
jopemachine: As with the other types, it would be better to provide a dedicated RolePermissionID type as well.

src/ai/backend/manager/models/rbac_models/role_preset/row.py
```diff
@@ -0,0 +1,54 @@
...
+        "id", GUID, primary_key=True, server_default=sa.text("uuid_generate_v4()")
+    )
+    name: Mapped[str] = mapped_column("name", sa.String(64), nullable=False)
+    scope_type: Mapped[ScopeType] = mapped_column(
+        "scope_type", StrEnumType(ScopeType, length=32), nullable=False
+    )
+    # If true, this preset is auto-applied when a scope of `scope_type` is created.
+    auto_apply: Mapped[bool] = mapped_column(
+        "auto_apply", sa.Boolean, nullable=False, server_default=sa.false()
+    )
```
HyeockJinKim: Should we need this flag?

jopemachine (approved):

PR #11842, opened 2026-05-27
feat(BA-6176): define role preset v2 DTOs

fregataa:
## Summary
- Add Pydantic v2 DTOs for role preset CRUD under `common/dto/manager/v2/role_preset/` (Create / Update / Delete / Search inputs and payloads, plus filter and order).
- Add Add / Remove permission inputs and payloads for managing a preset's permission set; permission entries are resolved separately (GQL data loader or REST endpoint) and not embedded in `RolePresetNode`.
- Introduce typed `RolePresetID` under `common/identifier/role_preset.py` following the existing per-entity identifier convention.

Resolves BA-6176


src/ai/backend/common/dto/manager/v2/role_preset/response.py
```diff
@@ -0,0 +1,81 @@
...
+
+    Permission entries are resolved separately (e.g., via a GraphQL `permissions` field
+    backed by a data loader, or a dedicated REST endpoint) and are not embedded here.
+    """
+
+    id: RolePresetID = Field(description="Role preset UUID.")
+    name: str = Field(description="Role preset name.")
+    scope_type: RBACElementTypeDTO = Field(description="Scope type this preset targets.")
+    auto_apply: bool = Field(description="If true, this preset is auto-applied at scope creation.")
+    auto_assign: bool = Field(
```
jopemachine, line 35: auto_apply seems similar to the meeting's is_default, but auto_assign is new to me. Could you explain the role of this field?

src/ai/backend/common/dto/manager/v2/role_preset/response.py
```diff
@@ -0,0 +1,81 @@
...
+    role_preset: RolePresetNode = Field(description="Updated role preset.")
+
+
+class DeleteRolePresetPayload(BaseResponseModel):
+    """Payload for role preset deletion."""
+
+    id: RolePresetID = Field(description="UUID of the deleted role preset.")
+
+
+class AddRolePresetPermissionsPayload(BaseResponseModel):
```
jopemachine: Does the AddRolePresetPermissions mutation add records to role_permission_preset?

Since RolePresetNode does not seem to directly contain information from role_permission_preset, shouldn’t this return a role_permission_preset node rather than a role_preset node?

jopemachine: Also, RolePresetPermissions and RolePermissionPreset seem to be used interchangeably, which looks easy to confuse.

src/ai/backend/common/dto/manager/v2/role_preset/request.py
```diff
@@ -57,20 +53,36 @@ class CreateRolePresetInput(BaseRequestModel):
...
     auto_assign: bool | None = Field(
         default=None,
         description="Updated default value for the `auto_assign` flag of instantiated roles.",
     )
 
 
+class DeleteRolePresetInput(BaseRequestModel):
+    """Input for soft-deleting a role preset (sets ``deleted = true``)."""
+
+    id: RolePresetID = Field(description="Role preset UUID to soft-delete.")
```
jopemachine: (Though this is still a draft PR,) in the case of soft delete, it is still an update mutation, so it should return the Node data.

HyeockJinKim (approved):

Issue #11814, opened 2026-05-27
RBAC scope-entity permission model consolidation & validation


HyeockJinKim:
## Background

RBAC scope_entities (association_scopes_entities) and role_permission (permissions) are implemented per BEP-1048, but gaps remain:

1. scope_entities cannot express the maximum **delegable** permission per entity. Relations are distinguished only by relation_type (AUTO/REF), and REF's read-only meaning is not enforced as a permission cap.

2. Permission resolution does not clip scope-derived permissions by the relation's cap, so a permission granted at a scope can exceed what the relation should allow for the entity (e.g., a REF relation should only yield read-only).

3. RBAC scope_entities/role_permission repository access (querier/creator) is not consolidated into the repositories/ops layer.

## Resolution Model: 1-hop direct-scope query (no scope walk)

The Permission-bitmask-based permission check resolves permissions by querying only the scopes the target entity is directly associated with in scope_entities (1-hop). The recursive scope walk is removed.

- Premise: RBAC entity creation paths associate each entity directly with all of its effective scopes (primary + additional scope refs), so multi-hop traversal over scope-to-scope edges is redundant.

- Consequence: cap clipping reduces to a single-edge bitwise AND (granted permission AND the edge's permission_cap). No running-min accumulation along a multi-hop descent path is needed.

## Cap Semantics (design principle)

- cap is the \***ceiling on delegation through an edge**\*: it bounds how much of a scope's permission flows to the entity through that scope-entity edge. It is applied at \***resolution time**\* as a bitwise AND of the edge's permission_cap with the permission resolved via that scope (single-edge clip under the 1-hop model). A NULL cap means "no ceiling" (full).

- cap does \***not**\* reject or reduce a permission whose resolution does not cross the capped edge. The union/additive model has no deny; cap only bounds the permission flowing through its own edge. A permission resolved via one scope edge is never reduced by another edge's cap.

- \***Entity-scope direct grants bypass cap.**\* A permission whose scope is the target entity itself (scope_type/scope_id == the entity, i.e. the entity-as-scope pattern) — e.g. the read/write grant created by a VFolder invitation — is matched directly (Layer 1) and is not clipped by any edge cap. The ref edge's cap only stops the invitee's own scope permissions from applying beyond the cap (Layer 2).

## Goals

- Introduce a permission (cap) concept on scope_entities and reconcile AUTO/REF into the cap.

[...]

PR #11843, opened 2026-05-28
feat(BA-6191): add ScopeCreatorSpec and ScopePurgerSpec to Ops layer

fregataa:
## Summary

- Add `ScopeCreatorSpec` / `ScopePurgerSpec` as composite coordinators for scope provisioning and teardown. Leaf sub-specs stay single-table.
- New `RBACWriterOps` class owns `create_scope` / `purge_scope`. Provisioning inserts the scope row, snapshots a role + its permissions per active preset matching the new scope, and writes role-to-scope plus parent-scope association rows in a single transaction.
- Teardown runs the inverse: scope-bound permissions, then association rows, then the scope row. Roles are intentionally left alone — their lifecycle is independent of the scope's.

## Test plan

- [x] CI typecheck + lint

Resolves BA-6191


fregataa:
Closed since we planned to refactor this part after #11814

PR #11845, opened 2026-05-28
feat(BA-6225): expand and rename Prometheus query presets

### LINKED ISSUE #11836: Add and rename prometheus fixtures
## Motivation

- 

## Objective

- 

## Details

1. 
1. 
1. 

## Impact

- 

JIRA Issue: BA-6225

seedspirit:
## Summary
- Rename the 8 example Prometheus query presets to verbose, aggregation-explicit names (e.g. `container_gauge` → `Per-Kernel Resource Metric — Instant Value (sum)`) and add avg/max/min variants for container and vLLM metrics, growing the set from 8 to 22.
- Add migration `7af18070fdef` that renames/deletes the seeded container presets **by name** (the prior seed used `uuid_generate_v4()`, so ids differ per environment) and idempotently seeds all new container + vLLM presets and their categories.
- Drop the `container_rate` (`sum(rate)/5.0`) preset whose normalization doesn't compose with avg/min/max.

## Test plan
- [x] Migration idempotency: re-running on already-seeded DB is a no-op (counts + `updated_at` unchanged)
- [x] Production simulation: old seed rows with auto-generated UUIDs → rename in place, `container_rate` deleted, new rows inserted
- [x] Full chain re-applied after rebase onto new main heads (`338bc3284f20`, `bee1c0de01a1`) — single linear alembic head
- [x] `pants lint/check` (ruff/mypy/visibility) pass
- [x] CLI `prometheus-query-definition execute` against a live session — all 8 container presets return correct, aggregation-consistent values; verified dynamic CPU load tracking

Resolves #11836 (BA-6225)

### DIFF
--- a/src/ai/backend/manager/models/alembic/versions/7af18070fdef_expand_prometheus_query_presets.py
+++ b/src/ai/backend/manager/models/alembic/versions/7af18070fdef_expand_prometheus_query_presets.py
@@ -1,7 +1,7 @@
 """expand_prometheus_query_presets
 
 Revision ID: 7af18070fdef
-Revises: bee1c0de01a1
+Revises: 0113c63f3261
 Create Date: 2026-05-27 00:00:00.000000
 
 # Part of: 26.3.0 (main)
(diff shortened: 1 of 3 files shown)


jopemachine (dismissed):

src/ai/backend/manager/models/alembic/versions/7af18070fdef_expand_prometheus_query_presets.py
```diff
@@ -0,0 +1,393 @@
+"""expand_prometheus_query_presets
+
+Revision ID: 7af18070fdef
+Revises: ed42bc179b91
+Create Date: 2026-05-27 00:00:00.000000
+
+# Part of: 26.3.0 (main)
```
fregataa, line 7: wrong version

fregataa (approved):

PR #11848, opened 2026-05-28
feat(BA-6230): add guard hooks to Updater and BulkUpdater partial primitive

fregataa:
## Summary
- `UpdaterSpec` gains abstract `guard_condition`, `not_found_error`, `on_guard_failure` hooks. `execute_updater` AND-merges the guard into the WHERE clause and, on a zero-row UPDATE, re-queries the row to differentiate `not_found_error` (raised when set) from `on_guard_failure` (raised when set, otherwise the existing row is returned).
- `BatchUpdaterSpec` gains an abstract `guard_condition` hook. `execute_batch_updater` only AND-merges the guard; the return type and `updated_count` semantics are unchanged.
- New `BulkUpdater` (sequence of `Updater`), `BulkUpdaterPartialResult`, and `execute_bulk_updater_partial` run each updater inside its own savepoint so failures are isolated. Exposed on `WriteOps` as `bulk_update_partial`.
- Every existing `UpdaterSpec` / `BatchUpdaterSpec` subclass implements the new abstract methods returning `None` — no behaviour change for existing specs.

Resolves BA-6230

### DIFF
--- a/src/ai/backend/manager/repositories/base/updater.py
+++ b/src/ai/backend/manager/repositories/base/updater.py
@@ -4,13 +4,14 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, TypeVar, cast
 from uuid import UUID
 
 import sqlalchemy as sa
 from sqlalchemy.engine import CursorResult
 
+from ai.backend.common.exception import BackendAIError
 from ai.backend.manager.errors.repository import UnsupportedCompositePrimaryKeyError
 from ai.backend.manager.models.base import Base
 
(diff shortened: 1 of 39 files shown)


src/ai/backend/manager/repositories/base/updater.py
```diff
@@ -58,6 +64,32 @@ def integrity_error_checks(self) -> Sequence[IntegrityErrorCheck]:
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def on_guard_failure(self) -> BackendAIError | None:
+        """Error raised when the row exists but the guard rejects the update.
+
+        Implementations return ``None`` to make the executor return the
+        existing row data unchanged instead of raising.
+        """
+        raise NotImplementedError
```
HyeockJinKim, line 91: It looks like there's some verbose code in there. What is this?

PR #11859, opened 2026-05-29
fix(BA-6240): look up streaming session by user_uuid instead of access_key

fregataa:
## Summary
- `StreamDBSource.get_streaming_session()` filtered by `access_key`, so users with multiple keypairs (including inactive ones) failed to connect to a session's apps when the request's access_key differed from the session's owner access_key.
- Replace the access_key filter with `user_uuid + status=RUNNING + session_name`. All kernels (not only main) are eager-loaded for the returned session.
- Stream actions (`GetStreamingSession`, `ExecuteInStream`, `StartServiceInStream`, `InterruptInStream`, `RestartInStream`) now carry `user_uuid` instead of `access_key`; handlers pass `user_ctx.user_uuid` directly.

## Test plan
- [x] New unit tests in `tests/unit/manager/repositories/stream/test_stream_repository.py` cover: RUNNING session lookup by user_uuid, main kernel loading, isolation from other users, TERMINATED sessions excluded, unknown name not found.
- [x] Manual: webui terminal app connects successfully for a user whose session was created with a now-inactive keypair.

Resolves BA-6240

### DIFF
--- a/changes/11859.fix.md
+++ b/changes/11859.fix.md
@@ -1,0 +1,1 @@
+Fix session app connection failure for users with multiple keypairs by looking up streaming sessions by user ID instead of access key.
--- a/src/ai/backend/manager/api/rest/stream/handler.py
+++ b/src/ai/backend/manager/api/rest/stream/handler.py
@@ -31,15 +31,16 @@ from aiohttp import web
 from aiotools import apartial
 
 from ai.backend.common.api_handlers import PathParam, QueryParam
+from ai.backend.common.contexts.user import current_user
 from ai.backend.common.dto.manager.stream.request import SessionNamePath, StreamProxyRequest
 from ai.backend.common.dto.manager.stream.response import StreamAppItem
-from ai.backend.common.exception import BackendAIError
+from ai.backend.common.exception import BackendAIError, UnreachableError
 from ai.backend.common.json import dump_json, load_json
-from ai.backend.common.types import AccessKey, KernelId, SessionId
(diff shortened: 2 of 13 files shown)


jopemachine:
Does this action need user_id parameter?
Are there any cases where the user obtained from current_user would be insufficient?
Also CI is broken

fregataa:
> Does this action need user_id parameter? Are there any cases where the user obtained from current_user would be insufficient? Also CI is broken

I don't want to change too many codes here because the stream API should be legacy since it identify session by name rather than id.
We can replace UserContext by current user in API handler functions but it will make too many change

jopemachine:
> > Does this action need user_id parameter? Are there any cases where the user obtained from current_user would be insufficient? Also CI is broken
> 
> I don't want to change too many codes here because the stream API should be legacy since it identify session by name rather than id. We can replace UserContext by current user in API handler functions but it will make too many change

I think Claude is increasingly getting sidetracked by incorrect patterns, outdated code, or dead code.

If possible, I’d prefer to align this with the correct pattern while we are already touching this area. Calling current_user does not seem like a particularly large change to me.

Also, CI is still failing.

fregataa:
> I think Claude is increasingly getting sidetracked by incorrect patterns, outdated code, or dead code.
> 
> If possible, I’d prefer to align this with the correct pattern while we are already touching this area. Calling current_user does not seem like a particularly large change to me.
> 
> Also, CI is still failing.

I changed the handler signatures

HyeockJinKim (approved):

PR #11863, opened 2026-05-29
feat(BA-6242): give custom runtime variant a default health check

seedspirit:
## Summary
- Seed the `custom` runtime variant's `default_model_definition` with a baseline `health_check` (path `/health`, `interval` 10s, `max_retries` 10, `initial_delay` 1800s) via alembic migration `ed42bc179b91`.
- Previously `custom` held `{"models": null}`, so deployments that omitted `health_check` fell back to the 60s schema default and could be marked unhealthy while large models load (up to ~30 min). 1800s matches the prebuilt variants.
- Merge stays field-wise: a user's `model-definition.yaml` / request still overrides each field, so this is purely a fallback layer. Operator-customised rows are left untouched (WHERE-guarded UPDATE, idempotent).
- Kept both fixtures (`fixtures/manager/` and `src/ai/backend/install/fixtures/`) in sync with the migration.

## Test plan
- [x] `pants fmt/lint/check` on the migration
- [x] Real alembic run on a throwaway DB: `0113c63f3261 → ed42bc179b91` seeds the def; `downgrade -1` reverts to `{"models": null}`; re-`upgrade head` is idempotent
- [x] Both fixtures parse as valid JSON
- [ ] CI

Resolves BA-6242

----
📚 Documentation preview 📚: https://sorna--11863.org.readthedocs.build/en/11863/

----
📚 Documentation preview 📚: https://sorna-ko--11863.org.readthedocs.build/ko/11863/

### DIFF
--- a/fixtures/manager/example-runtime-variants.json
+++ b/fixtures/manager/example-runtime-variants.json
@@ -12,6 +12,7 @@
               "start_command": ["vllm", "serve", "{model_path}"],
               "port": 8000,
               "health_check": {
+                "enable": false,
                 "path": "/health",
                 "interval": 10.0,
                 "max_retries": 10,
@@ -33,6 +34,7 @@
             "service": {
               "port": 8000,
               "health_check": {
+                "enable": false,
                 "path": "/v1/health/ready",
                 "interval": 10.0,
                 "max_retries": 10,
@@ -70,6 +72,7 @@
               "start_command": ["text-generation-launcher", "--model-id", "{model_path}"],
               "port": 3000,
               "health_check": {
+                "enable": false,
                 "path": "/info",
                 "interval": 10.0,
                 "max_retries": 10,
@@ -98,6 +101,7 @@
               ],
               "port": 9001,
               "health_check": {
+                "enable": false,
                 "path": "/health",
                 "interval": 10.0,
                 "max_retries": 10,
@@ -120,6 +124,7 @@
               "start_command": ["max", "serve", "--model", "{model_path}"],
               "port": 8000,
               "health_check": {
+                "enable": false,
                 "path": "/health",
                 "interval": 10.0,
                 "max_retries": 10,
@@ -140,6 +145,7 @@
             "name": "custom-model",
             "service": {
               "health_check": {
+                "enable": false,
                 "path": "/health",
                 "interval": 10.0,
                 "max_retries": 10,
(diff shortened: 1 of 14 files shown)


fregataa (commented):
health check field is nullable and when it is null then no health check runs

src/ai/backend/common/dto/manager/v2/deployment/request.py
```diff
@@ -122,6 +122,7 @@
 
 
 class ModelHealthCheckInput(BaseRequestModel):
+    enable: bool | None = None
```
fregataa: this should not be nullable if it is used as creation input

src/ai/backend/manager/api/gql/deployment/types/revision.py
```diff
@@ -862,6 +871,13 @@ class PreStartActionInputGQL(PydanticInputMixin[PreStartActionDTO]):
     name="ModelHealthCheckInput",
 )
 class ModelHealthCheckInputGQL(PydanticInputMixin[ModelHealthCheckInputDTO]):
+    enable: bool | None = gql_field(
```
fregataa: This is a creation input and boolean field does not need to be nullable
```suggestion
    enable: bool = gql_added_field(
```

src/ai/backend/common/config.py
```diff
@@ -650,6 +658,34 @@ def to_resolved(self) -> ModelDefinition:
...
+            ):
+                new_health_check = service.health_check.model_copy(update={"enable": True})
+                new_service = service.model_copy(update={"health_check": new_health_check})
+                model = model.model_copy(update={"service": new_service})
+                changed = True
+            new_models.append(model)
+        if not changed:
+            return draft
+        return draft.model_copy(update={"models": new_models})
+
```
fregataa, line 682: How is this draft resolved?

fixtures/manager/example-runtime-variants.json
```diff
@@ -12,6 +12,7 @@
               "start_command": ["vllm", "serve", "{model_path}"],
               "port": 8000,
               "health_check": {
+                "enable": false,
```
HyeockJinKim: Why is the "enable" value set to false here?

HyeockJinKim (changes requested):

HyeockJinKim (approved):

PR #11868, opened 2026-05-29
fix(BA-6241): pick deterministic active keypair for sokovan deployment sessions

fregataa:
## Summary
- Sokovan deployment built `sessions.access_key` from an unfiltered `users ⋈ keypairs` JOIN, so the persisted access_key was non-deterministic and could already be inactive at session creation time (observed: all recent INFERENCE sessions of a superadmin pinned to one `is_active=false` key).
- Extracted a `_resolve_user_and_active_access_key` helper used for both `created_user` and `session_owner` lookups in `DeploymentDBSource.fetch_deployment_context`. It filters `keypairs.is_active = TRUE` and orders by `(access_key == users.main_access_key) DESC, created_at DESC, access_key ASC LIMIT 1`, returning a private `_DeploymentUserResolution` dataclass.
- Distinguishes "user missing" from "user without an active keypair" by falling back to a UserRow-only check before raising.

## Test plan
- [x] CI: existing unit suite under `tests/unit/manager/sokovan/deployment/route/executor/` still passes.
- [x] Manual: on a deployment owner with multiple keypairs (one inactive, one active matching `users.main_access_key`), spawn a new replica and confirm the new `sessions.access_key` row resolves to the main active key.
- [x] Manual: deactivate the chosen keypair, spawn another replica, and confirm it now selects the next active key deterministically (and never picks the inactive one).

Resolves BA-6241

### DIFF
--- a/src/ai/backend/manager/errors/deployment.py
+++ b/src/ai/backend/manager/errors/deployment.py
@@ -59,7 +59,7 @@ class NoActiveKeypairForDeployment(ObjectNotFound):
 
     def error_code(self) -> ErrorCode:
         return ErrorCode(
-            domain=ErrorDomain.MODEL_SERVICE,
+            domain=ErrorDomain.MODEL_DEPLOYMENT,
             operation=ErrorOperation.READ,
             error_detail=ErrorDetail.NOT_FOUND,
         )
--- a/src/ai/backend/manager/errors/deployment.py
+++ b/src/ai/backend/manager/errors/deployment.py
@@ -54,6 +54,17 @@ class UserNotFoundInDeployment(ObjectNotFound):
         )
 
 
+class NoActiveKeypairForDeployment(ObjectNotFound):
+    object_name = "active keypair for deployment user"
+
+    def error_code(self) -> ErrorCode:
+        return ErrorCode(
+            domain=ErrorDomain.MODEL_SERVICE,
+            operation=ErrorOperation.READ,
+            error_detail=ErrorDetail.NOT_FOUND,
+        )
+
+
 class DeploymentHasNoTargetRevision(BackendAIError, web.HTTPBadRequest):
     error_type = "https://api.backend.ai/probs/deployment-has-no-target-revision"
     error_title = "Deployment has no target revision."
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -1977,6 +1987,44 @@ async def delete_routes_by_route_ids(
...
+        )
+        active_stmt = (
+            sa.select(UserRow, keypairs.c.access_key)
+            .select_from(sa.join(UserRow, keypairs, UserRow.uuid == keypairs.c.user))
+            .where(UserRow.uuid == user_uuid)
+            .where(keypairs.c.is_active.is_(True))
+            .order_by(
+                is_main_access_key.desc(),
+                keypairs.c.created_at.desc(),
+                keypairs.c.access_key.asc(),
```
jopemachine, line 2093: It seems lexicographic ordering of access keys is unnecessary here.

src/ai/backend/manager/errors/deployment.py
```diff
@@ -54,6 +54,17 @@ def error_code(self) -> ErrorCode:
         )
 
 
+class NoActiveKeypairForDeployment(ObjectNotFound):
+    object_name = "active keypair for deployment user"
+
+    def error_code(self) -> ErrorCode:
+        return ErrorCode(
+            domain=ErrorDomain.MODEL_SERVICE,
```
seedspirit: nitpick:

```suggestion
            domain=ErrorDomain.MODEL_DEPLOYMENT,
```

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
seedspirit: @fregataa  Is there reason for choosing lexicographic ordering?

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -204,6 +206,14 @@ class EndpointWithRoutesRawData:
     route_rows: list[RoutingRow]
 
 
+@dataclass(frozen=True)
+class _DeploymentUserResolution:
+    """Pairs a deployment user with the active keypair chosen for it."""
+
+    user: UserRow
+    access_key: AccessKey
```
seedspirit, line 216: Is the reason for accepting only one access key that the current deployment logic is configured to accept only a single access key? I’m asking because I’m curious why only one access key is accepted, even though a single user can have multiple access keys.

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
fregataa: It is almost impossible to have identical creation time in multiple records but I want this function work deterministic

HyeockJinKim (dismissed):

HyeockJinKim (approved):

PR #11886, opened 2026-06-01
fix(vfolder): fallback system role creation on invitation acceptance

wali-reheman:
When accepting a vfolder invitation for a user without a SYSTEM role, the lookup now creates one as a fallback inside the transaction instead of failing with ObjectNotFound. This handles SSO-created accounts that may not have an RBAC role yet. Includes unit tests for both the existing-role path and the fallback-creation path.

### DIFF
--- a/src/ai/backend/manager/repositories/vfolder/repository.py
+++ b/src/ai/backend/manager/repositories/vfolder/repository.py
@@ -139,6 +139,11 @@ from ai.backend.manager.repositories.base.rbac.revoker import (
     execute_rbac_revoker,
 )
 from ai.backend.manager.repositories.base.updater import Updater, execute_updater
+from ai.backend.manager.repositories.permission_controller.role_manager import (
+    RoleManager,
+    UserSystemRoleSpec,
+)
+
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/repositories/vfolder/repository.py
```diff
@@ -1148,9 +1156,13 @@ async def _get_user_role_id(self, session: SASession, user_id: uuid.UUID) -> uui
...
-        if result is None:
-            raise ObjectNotFound(object_name="user system role", extra_msg=str(user_id))
-        return result
+        if result is not None:
+            return result
+
+        # Fallback: create a system role for this user and return its id
+        spec = UserSystemRoleSpec(user_id=user_id)
+        role = await self._role_manager.create_system_role(session, spec)
+        return role.id
```
HyeockJinKim, line 1165: This approach doesn't seem appropriate. It would make more sense for the operation to fail if it can't retrieve the data from `get`; creating and passing the user's role when it can't be retrieved is very tricky.

HyeockJinKim (changes requested):

PR #11891, opened 2026-06-01
feat(BA-6254): add superadmin ensure-system-role repository/service/action

fregataa:
## Summary
- Add `ensure_user_system_roles(specs)` across repository → service → action → processor: an idempotent superadmin operation that guarantees each given user has a SYSTEM role and its user-role mapping.
- Creates the role (with permissions) and mapping when missing; reuses the existing role and leaves the mapping untouched on repeat calls (no duplicate / unique-constraint violation). Accepts a batch of `UserSystemRoleSpec` and returns `RoleMappingData` per user.
- Scope is USER (this issue covers repository/service/action only — GQL/adapter, REST/SDK/CLI, and the `_get_user_role_id` error change are sibling issues BA-6255/6256/6257).

## Test plan
- [x] `pants fmt / fix / lint / check` pass on changed files
- [x] Repository tests (`test_ensure_user_system_roles.py`, real DB via `with_tables`): create-when-missing grants role + permissions + mapping; idempotent reuse on repeated calls; batch with mixed existing/missing users; empty input is a no-op
- [ ] CI: full type check + test suite

Resolves BA-6254

### DIFF
--- a/changes/11891.feature.md
+++ b/changes/11891.feature.md
@@ -1,0 +1,1 @@
+Add a superadmin operation that ensures a user has an RBAC system role, creating it (with permissions and the user-role mapping) when missing.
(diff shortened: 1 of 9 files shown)


fregataa:
Closed this PR since it is not planned anymore

PR #11909, opened 2026-06-03
fix(BA-6270): recalculate appproxy worker available_slots on restart

jopemachine:
## Summary

- App Proxy Worker computed `available_slots` only at first registration (`Worker.create`). On restart with a changed `port_range`, `update_worker` updated `port_range` but left `available_slots` pinned to the original value.
- `available_slots` is the admission-control quota checked in `pick_worker` and `add_circuit` (`available_slots - occupied_slots`), while the actual port pool is derived from `port_range`. The two diverged: ports were physically free but the stale quota blocked new circuit creation once `occupied_slots` reached the old value (e.g. stuck at 100 while `port_range` 10501–10800 has 300 ports).
- Extracted the slot calculation into the private `Worker._calculate_available_slots()` and exposed `Worker.refresh_available_slots()`, which `update_worker` (and `create`) now call, so a restart with an updated `port_range` / `frontend_mode` is reflected.

## Why the stale value blocked port creation

`available_slots` acts as a cached capacity ceiling. Both admission gates filter/reject workers when `available_slots - occupied_slots <= 0`, so the code never reaches the `port_range`-based pool computation. When `port_range` was expanded but `available_slots` stayed at the old (smaller) value, legitimate allocations were rejected despite free ports.

## Note on shrinking port_range

When `port_range` is reduced such that the new `available_slots < occupied_slots`, the worker is naturally excluded by the capacity gates. Existing circuits whose ports fall outside the new range are **not** force-removed by this PR (cleanup/migration of out-of-range ports is out of scope and tracked separately, as noted in BA-6270). New allocation already uses `port_range` directly, so out-of-range existing ports do not block new allocations.

## Test plan

- [x] `pants check` on changed source files
- [x] New unit tests in `tests/unit/appproxy/coordinator/test_worker_available_slots.py` (port-mode span, recompute on range change, wildcard unlimited, missing-config errors, `create()` integration)
- [ ] CI green
- [ ] Live verification: restart appproxy worker with an expanded `port_range`, confirm DB `available_slots` updates and new ports can be allocated past the previous ceiling

Resolves BA-6270

### DIFF
--- a/src/ai/backend/appproxy/coordinator/models/worker.py
+++ b/src/ai/backend/appproxy/coordinator/models/worker.py
@@ -254,6 +254,20 @@ class Worker(Base, BaseMixin):  # type: ignore[misc]
                 return port_range[1] - port_range[0] + 1
         raise MissingFrontendConfigError(f"Unsupported frontend mode: {frontend_mode}")
 
+    def refresh_available_slots(self) -> None:
+        """
+        Recompute ``available_slots`` from the worker's current frontend
+        configuration. Call this after mutating ``frontend_mode`` / ``port_range``
+        / ``wildcard_domain`` (e.g. when a worker re-registers on restart) so the
+        slot quota tracks the live port range instead of staying pinned to the
+        value computed at the worker's first registration (BA-6270).
+        """
+        self.available_slots = self.calculate_available_slots(
+            self.frontend_mode,
+            port_range=self.port_range,
+            wildcard_domain=self.wildcard_domain,
+        )
+
     @property
     def use_tls(self) -> bool:
         return self.tls_listen or self.tls_advertised
--- a/src/ai/backend/appproxy/coordinator/models/worker.py
+++ b/src/ai/backend/appproxy/coordinator/models/worker.py
@@ -223,36 +223,26 @@ class Worker(Base, BaseMixin):  # type: ignore[misc]
 
         return w
 
-    @staticmethod
-    def _calculate_available_slots(
-        frontend_mode: FrontendMode,
-        *,
-        port_range: tuple[int, int] | None = None,
-        wildcard_domain: str | None = None,
-    ) -> int:
-        """Number of available slots derived from the frontend configuration."""
-        match frontend_mode:
+    def _calculate_available_slots(self) -> int:
+        """Number of available slots derived from the current frontend configuration."""
+        match self.frontend_mode:
             case FrontendMode.WILDCARD_DOMAIN:
-                if not wildcard_domain:
+                if not self.wildcard_domain:
                     raise MissingFrontendConfigError(
                         "Wildcard domain is required for WILDCARD_DOMAIN frontend mode"
                     )
                 return -1
             case FrontendMode.PORT:
-                if not port_range:
+                if not self.port_range:
                     raise MissingFrontendConfigError(
                         "Port range is required for PORT frontend mode"
                     )
-                return port_range[1] - port_range[0] + 1
-        assert_never(frontend_mode)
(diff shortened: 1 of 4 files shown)


src/ai/backend/appproxy/coordinator/models/worker.py
```diff
@@ -219,21 +219,40 @@ def create(
...
+
+        return w
+
+    @staticmethod
+    def _calculate_available_slots(
+        frontend_mode: FrontendMode,
+        *,
+        port_range: tuple[int, int] | None = None,
+        wildcard_domain: str | None = None,
+    ) -> int:
```
fregataa: this method should be an instance method rather than a static method

src/ai/backend/appproxy/coordinator/models/worker.py
```diff
@@ -219,21 +219,40 @@ def create(
...
             case FrontendMode.PORT:
                 if not port_range:
                     raise MissingFrontendConfigError(
                         "Port range is required for PORT frontend mode"
                     )
-                w.available_slots = port_range[1] - port_range[0] + 1
-
-        return w
+                return port_range[1] - port_range[0] + 1
+        raise MissingFrontendConfigError(f"Unsupported frontend mode: {frontend_mode}")
```
fregataa: Is this reachable?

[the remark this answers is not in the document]
jopemachine: Replaced the dead branch with `typing.assert_never` so it is statically verified as unreachable instead of raising a misleading `MissingFrontendConfigError`. FrontendMode only has WILDCARD_DOMAIN and PORT, both already matched.

tests/unit/appproxy/coordinator/test_worker_available_slots.py
```diff
@@ -0,0 +1,105 @@
...
+                expected_available_slots=1,
+            ),
+        ],
+        ids=["expand", "shrink", "unchanged", "single-port"],
+    )
+    def test_refresh_after_port_range_change(self, case: RangeChangeCase) -> None:
+        worker = make_port_worker(case.initial)
+        worker.port_range = case.updated
+        worker.refresh_available_slots()
+        assert worker.available_slots == case.expected_available_slots
```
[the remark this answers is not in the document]
jopemachine: Added `TestCreate.test_create_initializes_available_slots` asserting that `Worker.create()` initializes `available_slots` correctly.

src/ai/backend/appproxy/coordinator/models/worker.py
```diff
@@ -219,21 +219,40 @@ def create(
...
         w.occupied_slots = 0
+        w.refresh_available_slots()
+
+        return w
+
+    @staticmethod
+    def _calculate_available_slots(
+        frontend_mode: FrontendMode,
+        *,
+        port_range: tuple[int, int] | None = None,
```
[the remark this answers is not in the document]
jopemachine: Updated the PR description to reflect the actual API: private `_calculate_available_slots()` + public `refresh_available_slots()`.

src/ai/backend/appproxy/coordinator/models/worker.py
jopemachine: Done — converted to an instance method that reads frontend_mode / port_range / wildcard_domain from self.

src/ai/backend/appproxy/coordinator/models/worker.py
jopemachine: Replaced by UnreachableError

fregataa (approved):

PR #11918, opened 2026-06-04
feat(BA-6189): add role update CLI command and --auto-assign flag

fregataa:
## Summary
- Add `--auto-assign/--no-auto-assign` flag to `bai rbac role create`, exposing the `Role.auto_assign` field already wired through the v2 DTO/SDK.
- Add a new `bai rbac role update` command supporting `--name`, `--description`, `--status`, and `--auto-assign`, with unset options mapped to no-change semantics (`SENTINEL` for description, `None` for status/auto_assign).
- Route both commands through the shared `run_async` helper for consistent SDK error handling.

Scope note: the `role_option` (role_ids/overwrite) on user-join portion of this story is deferred — it depends on the service-layer apply logic (BA-6190) which is not yet implemented, consistent with how the sibling GQL story (BA-6187) shipped only the `auto_assign` half.

## Test plan
- [x] `bai rbac role create --name x --auto-assign` then `bai rbac role get <id>` shows `auto_assign: true`
- [x] `bai rbac role update <id> --no-auto-assign` toggles it back to false
- [x] `bai rbac role update <id> --name y` changes only the name (description/status/auto_assign unchanged)

Resolves BA-6189

### DIFF
--- a/src/ai/backend/client/cli/v2/rbac/role.py
+++ b/src/ai/backend/client/cli/v2/rbac/role.py
@@ -16,6 +16,7 @@ from ai.backend.client.cli.v2.helpers import (
     load_v2_config,
     parse_order_options,
     print_result,
+    run_async,
 )
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


src/ai/backend/client/cli/v2/rbac/role.py
```diff
@@ -191,21 +192,70 @@ async def _run() -> None:
...
-    asyncio.run(_run())
+    run_async(_run)
+
+
+@role.command()
+@click.argument("role_id", type=click.UUID)
+@click.option("--name", default=None, help="Updated role name.")
+@click.option(
+    "--status",
+    type=click.Choice(["active", "inactive", "deleted"], case_sensitive=False),
```
jopemachine, line 223: Why not click.Choice([*RoleStatus])?

fregataa: There is no such convention in v2 CLI because it requires to import a common module on a top of this file and it is not allowed in any CLI files

jopemachine (approved):

PR #11919, opened 2026-06-04
chore: replace Pants visibility lint with standalone Rust visibility-checker

achimnol:
## Summary

- Adds a standalone Rust CLI (`tools/visibility-checker/`) that checks first-party Python import reachability under `src/ai/backend/` against the repo's `visibility_private_component()` BUILD-file rules, including **transitive** import paths that Pants' built-in visibility linter misses
- Wires the checker into `pants lint` via a new Pants plugin (`tools/pants-plugins/visibility_checker/`); Pants' own experimental visibility backend is kept loaded (so BUILD macros still parse) but its checks are disabled in favour of the new tool
- Ships prebuilt static binaries for Linux x86-64, Linux aarch64, and macOS aarch64 as **Git LFS** objects so developers and CI do not need a Rust/Cargo toolchain; the plugin selects the right binary at lint time based on the host platform

## Test plan

- [ ] `pants lint --changed-since=origin/main` passes on a Linux x86-64 machine
- [ ] Verify the checker reports known violations (e.g. a scratch import crossing a `visibility_private_component` boundary) and exits non-zero
- [ ] Verify `pants lint` skips the checker when `--visibility-checker-skip` flag is set
- [ ] Confirm `git lfs pull` fetches the three binaries on a fresh clone


achimnol:
It turns out that visibility check itself is not a bottleneck but the prior dependency inference step (which can be separately invoked by `pants peek`) is the bottleneck.
Closing this issue as there seems to be no real gain.

PR #11903, opened 2026-06-02
feat(BA-6267): add replica field to SessionV2 GraphQL type

jopemachine:
## Summary

Adds a `replica` field to the `SessionV2` GraphQL type so the Session detail page can surface **Deployment** information for inference sessions. The field resolves to the `ModelReplica` (DB: `RoutingRow`) served by the session; clients follow `replica.revision` / the deployment chain to reach the Deployment.

> **Stacked on #11921** (`execute_bulk_querier`). This PR targets `feat/bulk-querier`; review/merge that first. The bulk-querier infrastructure itself lives there.

## Changes

- **`SessionV2GQL.replica`** — resolves the `ModelReplica` for a session via a session-scoped DataLoader (`replica_by_session_loader`), avoiding N+1 across a session list.
- **Deployment side** — the adapter builds one by-session `Querier` per session and resolves them through a generic `bulk_query_routes` action → service → repository → `execute_bulk_querier`.
- **RBAC** — `BulkQueryRoutesAction` is a USER-scope action run through `ScopeActionProcessor` with the `rbac.scope` validator; `UserRouteSearchScope` filters routes to the requester's own (`RoutingRow.session_owner == user_uuid`).

## Notes

- A session maps to at most one replica → the field returns a single `ModelReplica | None`.
- **USER-scope limitation (follow-up):** a requester only sees replicas for sessions they own; superadmin / project-admin viewing others' sessions would currently get `null`. Deferred pending the scoping decision.

## Verification

- [ ] `./dev restart mgr && ./dev restart web`, then query `session { replica { id revision { id } } }` for an inference session and a plain session.

Jira: BA-6267

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11921, opened 2026-06-04
feat(BA-6274): add execute_bulk_querier with partitioned results and scopes

jopemachine:
## Summary

Adds `execute_bulk_querier` to the repository base layer — the **bulk counterpart of `execute_querier`**. It resolves many single-row queriers in one round-trip per `(row_class, lookup column)` group and **partitions** the outcome into successes and typed failures, with optional RBAC scope filtering.

This is split out of the SessionV2 `replica` work (#11903) as standalone, reusable infrastructure.

## Changes

- **`Querier.lookup_column`** — optional ORM attribute to match a 1:1 alternate key instead of the primary key (e.g. `RoutingRow.session`). `execute_querier` honors it too.
- **`execute_bulk_querier(db_sess, queriers, scopes=())`** — groups queriers by lookup column into one `WHERE col IN (...)` per group; returns `BulkQuerierResult(successes: list[QuerierResult], failures: list[QuerierFailureResult])`, preserving input order.
- **`scopes`** — applied exactly as in `execute_batch_querier` (existence checks validated once; `to_condition()`s OR-combined and AND-merged). An out-of-scope row is treated as not found, so there is no cross-scope leakage.
- New types: `QuerierFailureResult`, `BulkQuerierResult`, `BulkQuerierFailureReason`, exported from `repositories.base`.

## Tests

`test_querier.py` covers PK lookup, alternate-key lookup, partial (found + missing), input ordering, and scope filtering.

Jira: BA-6274

### DIFF
--- a/changes/11921.feature.md
+++ b/changes/11921.feature.md
@@ -1,0 +1,1 @@
+Add `execute_bulk_querier`, the bulk counterpart of `execute_querier`, which resolves many single-row queriers in one query per lookup-column group and partitions the outcome into successes and typed failures, with optional RBAC scope filtering
--- a/src/ai/backend/manager/repositories/base/__init__.py
+++ b/src/ai/backend/manager/repositories/base/__init__.py
@@ -58,9 +58,13 @@ from .purger import (
 from .querier import (
     BatchQuerier,
     BatchQuerierResult,
+    BulkQuerierFailureReason,
+    BulkQuerierResult,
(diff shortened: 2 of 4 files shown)


jopemachine:
Closing — the execute_bulk_querier infra is no longer needed for the SessionV2 replica feature, which will use a direct sessions.replica_id column instead.

PR #11925, opened 2026-06-04
feat(BA-6267): expose deployment info on SessionV2 via replica_id

### LINKED ISSUE #11901: Expose deployment (replica/route) info on SessionV2 GraphQL type via session-scoped dataloader
## Objective

The Session detail page needs to show Deployment information for inference sessions. To support this, add a replica field to the SessionV2 GraphQL type (SessionV2GQL) that resolves to the ModelReplica corresponding to the session, allowing the client to follow the chain to its Deployment.

## Background

- ModelReplica (GQL type) corresponds to RoutingRow (DB), which has a nullable session FK to sessions.id and an endpoint FK to the deployment (EndpointRow).
- Existing infra: SessionV2GQL lives in api/gql/session/types.py; ModelReplica type and replica_loader (keyed by replica/route id) and the route->session_v2 resolver already exist in api/gql/deployment/. This issue is the reverse direction (session -> route -> deployment).
- Resolving the field per-session would cause an N+1 query problem when listing many sessions, so a DataLoader that bulk-queries RoutingRow by a list of session_ids is required.

## Scope

- Add a repository/adapter method to batch-load RoutingRow records by a list of session ids (batch_load_replicas_by_session_ids).
- Register a new DataLoader keyed by SessionId that returns the ModelReplica (or None) for each session.
- Add a replica resolver field on SessionV2GQL that uses the new dataloader; return None for non-inference sessions.
- Decide and document multiplicity: a session may map to multiple RoutingRow records; clarify whether to return a single replica or a list.

## Acceptance Criteria

- SessionV2GQL exposes a replica field that resolves to the corresponding ModelReplica, from which the Deployment can be reached.
- Querying the replica field across N sessions issues a single bulk query (no N+1) via the new DataLoader.
- Non-inference sessions (no associated route) return null without error.
- Verified against the live server via GraphQL query for both inference and non-inference sessions.

JIRA Issue: BA-6267

jopemachine:
Resolves #11901 (BA-6267)

## Summary

Surface the model deployment a session serves on the `SessionV2` GraphQL type, so the Session detail page can reach Deployment info. The traversal added is **session → replica → deployment**, kept deliberately simple: a nullable `sessions.replica_id` column batch-loaded through the existing replica DataLoader — no search, scope, or RBAC plumbing.

Supersedes the earlier `execute_bulk_querier` + scope-action approach (PRs #11921 / #11903, now closed), which hit an RBAC model mismatch (`DEPLOYMENT_REPLICA` has no `RBACElementType`) and all-or-nothing bulk validation.

## Changes

**Read path — `SessionV2.replica`**
- Alembic migration `169cb5f48658` adds a nullable `sessions.replica_id` column (chained onto the current head `eb9d9c018e85`; marked `Part of: 26.6.0`).
- `SessionRow.replica_id` — plain `GUID` column with **no DB FK** (avoids a circular `routings`↔`sessions` FK and the resulting SQLAlchemy relationship ambiguity; stale ids resolve to null gracefully).
- Threaded `replica_id` through `SessionData` → `SessionNode` DTO → `SessionV2GQL`.
- `SessionV2GQL.replica` resolves `ModelReplica` via the existing `replica_loader(replica_id)`; `null` for non-inference sessions.

**Population — `sessions.replica_id` (now in scope)**
- Migration backfill (idempotent): set `replica_id` from the earliest route bound to each session (`routings.session = sessions.id`).
- The two `db_source` methods that bind a route to a session now also set `sessions.replica_id = route_id` in the same transaction, gated on the bind actually happening (`rowcount > 0`) and first-bind-wins (`replica_id IS NULL`) to match the backfill. Dead `update_route_session` path removed.

**Reach deployment — `ModelReplica.deployment`**
- `ModelReplica` gains `deploymentId` (ID) and a `deployment` resolver that loads the owning `ModelDeployment` via the existing `deployment_loader`, completing the `session → replica → deployment` chain.

## Verification

Verified end-to-end on a live manager: created an inference deployment via `./bai deployment create`; the scheduler bound a route to a freshly-enqueued session and the `db_source` hook populated `sessions.replica_id` at bind time (before the session turned healthy). A GraphQL `sessionV2 { replicaId replica { deploymentId deployment { id metadata { name } } } }` query resolves the full chain to the owning deployment; non-inference sessions return `replica: null`. `alembic upgrade head` applies cleanly as a single linear head.

----
📚 Documentation preview 📚: https://sorna--11925.org.readthedocs.build/en/11925/

----
📚 Documentation preview 📚: https://sorna-ko--11925.org.readthedocs.build/ko/11925/


jopemachine:
Removed `DeploymentDBSource.update_route_session` (the single-route variant) in 2f75ba90d — it was **dead code** (no callers in `src/` or `tests/`; `grep -rn 'update_route_session\b'` matched only its own definition).

The live route→session binding path is the bulk `update_route_sessions` (called from `sokovan/deployment/route/executor.py` during `provision_routes`), which retains the `sessions.replica_id` mirror. So nothing functional is lost.

src/ai/backend/manager/models/alembic/versions/169cb5f48658_add_session_replica_id.py
```diff
@@ -0,0 +1,51 @@
...
+                ORDER BY routings.created_at ASC, routings.id ASC
+                LIMIT 1
+            )
+            WHERE sessions.replica_id IS NULL
+              AND EXISTS (
+                SELECT 1 FROM routings WHERE routings.session = sessions.id
+              )
+            """
+        )
+    )
```
HyeockJinKim: Instead of updating based on the session, we only update the sessions that haven't been resolved in the routing.

src/ai/backend/manager/api/gql/session/types.py
```diff
@@ -319,6 +322,9 @@ class SessionV2GQL(PydanticNodeMixin[SessionNode]):
     domain_name: str
     user_id: ID
     project_id: ID
+    replica_id: ID | None = gql_field(
```
fregataa: gql_added_field

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -1974,7 +1959,20 @@ async def update_route_sessions(
...
                     .values(session=session_id)
                 )
-                await db_sess.execute(query)
+                result = await db_sess.execute(query)
+                # Mirror the binding onto the session only when the route was
+                # actually bound this call (the `session IS NULL` guard above
+                # held). The `replica_id IS NULL` guard keeps first-bind-wins,
+                # matching the migration backfill, so a session never flips to a
+                # route serving another.
+                if cast(CursorResult[Any], result).rowcount > 0:
```
fregataa: 1. is there any test for this code?
2. this cast looks weird

fregataa (changes requested):

seedspirit (approved):

fregataa (approved):

PR #11939, opened 2026-06-05
feat: Mount one vfolder at multiple subpaths in a single session

### LINKED ISSUE #11936: Allow mounting the same vfolder at multiple subpaths in a single session
> **Reopened (2026-06-07):** Previously closed as a duplicate of #10599. That was a miscommunication with @HyeockJinKim — #10599 (the v2 session REST API) did **not** address this, and the last-write-wins collapse remains. Tracking the real fix here; see PR #11954 (supersedes the un-reopenable #11939).

---

## Summary

The session-creation mount API cannot mount **the same vfolder at more than one subpath** in a single session. A request that references one vfolder UUID several times — each with a distinct `subpath`, `mount_destination`, and `permission` — is silently collapsed to a single mount (last-write-wins per UUID), so all but one of the requested subpath mounts are dropped without error.

The downstream data model already supports multiple non-overlapping subpath mounts of one vfolder; the limitation is purely an input-collapsing step in `prepare_vfolder_mounts`.

## Use case

A workload needs several subdirectories of **one** shared vfolder mounted at different container paths with different permissions in the same session — e.g. a dataset vfolder exposing `inputs/` read-only at `/work/in1` and `/work/in2`, plus `outputs/` read-write at `/work/out`. Today this requires creating a separate vfolder per mount point, even though they are subdirectories of the same vfolder.

## Current behavior (manager @ `26.4.4rc6`)

`prepare_vfolder_mounts` accepts a genuine list of requests but immediately re-buckets UUID-referenced ones into UUID-keyed dicts:

`src/ai/backend/manager/models/vfolder/row.py:938`
```python
for req in mount_requests:                       # mount_requests: Sequence[VFolderMountRequest]
    if isinstance(req.ref, uuid.UUID):
        if req.dst_path is not None:
            requested_mount_map[req.ref] = req.dst_path        # dict[UUID, str]  -> last wins
        requested_mount_options[req.ref] = req.options          # dict[UUID, ...] -> last wins
```
and derives exactly one subpath per UUID:

`src/ai/backend/manager/models/vfolder/row.py:970`
```python
for vfolder_uuid, value in requested_mount_map.items():
    requested_vfolder_subpaths[vfolder_uuid] = _normalize_mount_subpath(...)   # one per UUID
```

So passing
```python
[VFolderMountRequest(ref=uuidA, dst_path="/work/in",  options=VFolderMountOptions(subpath="inputs",  permission=RO)),
 VFolderMountRequest(ref=uuidA, dst_path="/work/out", options=VFolderMountOptions(subpath="outputs", permission=RW))]
```
yields a single mount for `uuidA` (the second overwrites the first).

Every input surface funnels through this same collapse:
- **UUID-keyed creation config** — `mount_id_map: dict[UUID, str]` and `mount_options: dict[str, MountOption]` (`src/ai/backend/common/dto/manager/session/types.py:447,451`, `CreationConfigV7`) are one-entry-per-UUID by construction; `_mount_entries_from_creation_config` reads a single `subpath` from `mount_options[uuid]` (`src/ai/backend/manager/registry.py:242,252`).
- **Legacy `name/subpath` form** — resolved with `subpath_by_name.setdefault(name, subpath)` (`src/ai/backend/manager/api/rest/session/handler.py:486`), i.e. first-wins, one subpath per vfolder name.

## The data model already supports it

The resolved mounts are held as a **list**, and duplicate detection keys on the **`(vfid, subpath)` pair**, rejecting only *overlapping* subpaths — distinct non-overlapping subpaths of the same vfolder are explicitly allowed:

`src/ai/backend/manager/models/vfolder/row.py:2000`
```python
def is_mount_duplicate(folder_id, subpath, mounts) -> bool:
    for mount in mounts:
        if mount.vfid != folder_id:
            continue
        if subpath.is_relative_to(mount.vfsubpath) or mount.vfsubpath.is_relative_to(subpath):
            return True
    return False
```

`VFolderMountRequest` / `VFolderMountOptions` (`src/ai/backend/common/types.py:797,789`) and `MountInfoEntry` (`:732`) all carry a per-entry `subpath`, and the scheduler builds one `VFolderMountRequest` per entry with no dedup (`src/ai/backend/manager/repositories/scheduler/db_source/db_source.py:1634`). Only the dict re-bucketing in `prepare_vfolder_mounts` discards the extra entries.

## Proposed change

In `prepare_vfolder_mounts`, stop collapsing UUID-referenced requests into `dict[UUID, …]`. Iterate the request list and emit one `VFolderMount` per request, using the existing `is_mount_duplicate` `(vfid, subpath)` check to reject only genuinely overlapping mounts. The internal list-based representation and the storage-proxy `vfsubpath` resolution downstream already handle the rest; this is a localized change to the request-bucketing step.

## Context (prior PRs)

- #9149 (BA-2983) introduced the UUID-keyed `subpath` surface — one subpath per vfolder.
- #11250 dropped the legacy `vfname/subpath` shape during the sokovan refactor.
- #11434 (BA-5916) made `vfname/subpath` an explicit 400.
- #11582 (BA-6022) restored `vfname/subpath`, re-routed onto the same one-subpath-per-vfolder surface.

None of these reinstated the ability to mount one vfolder at multiple subpaths simultaneously, which is the gap this issue tracks.

## Affected version

Verified on `26.4.4rc6` (tree `26.4.4rc6-21-g22e5149a3`).

rapsealk:
> **Reopened (2026-06-07):** Previously closed as a duplicate of #10599 — that was a miscommunication; #10599 did not fix this and the collapse remains. This PR additionally raises `VFolderNotFound` for a requested-but-inaccessible vfolder UUID (previously silently dropped), addressing the earlier review feedback.

---

Resolves #11936.

## Problem

`prepare_vfolder_mounts` could not mount the **same vfolder at more than one subpath** in a single session. It re-bucketed UUID-referenced mount requests into per-UUID maps (`requested_mount_map`, `requested_vfolder_subpaths[uuid]`, `requested_mount_options[uuid]`), so when a vfolder UUID appeared several times — each with a distinct `subpath`/`mount_destination`/`permission` — only the last entry survived. The extra subpath mounts were dropped silently, with no error.

The downstream model already supported it: resolved mounts are a list and `is_mount_duplicate` keys on the `(vfid, vfsubpath)` pair, rejecting only *overlapping* subpaths. The collapse was solely in the request-bucketing step.

## Change

`prepare_vfolder_mounts` now keys each UUID-referenced request by its **index** in `mount_requests` instead of the bare vfolder UUID (`uuid_request_indices` maps each vfolder UUID to the request keys that target it). Each request keeps its own subpath, destination, and options through to the mount loop, so one vfolder resolves to one `VFolderMount` per request. Overlapping `(vfid, subpath)` pairs are still collapsed by the existing `is_mount_duplicate` guard, and the final kernel-path overlap check is unchanged. The redundant `requested_mount_map`-based accessible-vfolder query condition (already covered by `requested_vfolder_ids`) and the now-dead name-keyed option copies were removed. The name-referenced surface (`name/subpath`) and dot-folder automount paths are unchanged.

## Tests

Added to `tests/unit/manager/models/test_vfolder_prepare_mounts.py`:
- `test_same_vfolder_multiple_subpaths_yield_distinct_mounts` — one vfolder, two subpaths → two mounts with the correct `vfsubpath`/`host_path`/`kernel_path`.
- `test_same_vfolder_duplicate_subpath_is_deduplicated` — identical `(vfolder, subpath)` twice → one mount.

Verified locally against a real Postgres (testcontainers):
- `test_vfolder_prepare_mounts.py` — 7 passed
- `test_resolve_vfolder_mounts_rule.py` — 2 passed
- `test_legacy_extra_mounts_hydration.py` — 1 passed
- `test_session_lifecycle_service.py` (mount/legacy/subpath) — 5 passed
- `ruff check` / `ruff format` — clean (repo pre-commit hook)


rapsealk:
Duplicates to #10599.

rapsealk:
Superseded by #11954. This PR cannot be reopened — its head branch was deleted on close, so GitHub blocks the reopen (`state cannot be changed … branch was force-pushed or recreated`). The same branch `feat/multisubpath-vfolder-mounts` and work (plus the `VFolderNotFound` fix from the earlier review) continue in #11954.

Issue #11634, opened 2026-05-15
Add lamma cpp in runtime variant fixture


seedspirit:
## Motivation

Add lamma cpp in runtime variant fixture

## Objective

[Clearly state the purpose or goal of this task.]

- 

## Details

[Outline the specific steps or changes needed to complete this task.]

1. 
1. 
1. 

## Impact

[Describe how this task may indirectly influence other services, if at all.]

- 

JIRA Issue: BA-6058

PR #11940, opened 2026-06-05
feat(BA-6058): add llama-cpp runtime variant to example fixtures

seedspirit:
## Summary
- Add a `llama-cpp` runtime variant to the manager and installer `example-runtime-variants` fixtures.
- `start_command` selects the first `.gguf` under the model mount (`/models`) and forwards preset args to `llama-server` via `"$@"`; the service listens on port 8080.

## Notes
- The variant assumes a Backend.AI `llama.cpp` image where `llama-server` is on `PATH` and `LD_LIBRARY_PATH=/app` (the stock `ghcr.io/ggml-org/llama.cpp` binary relies on `WORKINGDIR /app`, which Backend.AI bypasses), mirroring how the `vllm` variant assumes `vllm` is on `PATH`.
- Fixtures seed fresh installs only; exposing the variant on an existing deployment needs a separate alembic seed migration.

## Test plan
- [ ] Runtime-variant fixture loads / seeds without error.
- [ ] Create a model service with `runtime_variant=llama-cpp` + a llama.cpp image and a `.gguf` vfolder → `llama-server` boots and `/health` returns 200.

Resolves BA-6058 #11634

### DIFF
--- a/fixtures/manager/example-runtime-variants.json
+++ b/fixtures/manager/example-runtime-variants.json
@@ -131,6 +131,33 @@
       }
     },
     {
+      "name": "llama-cpp",
+      "description": "llama.cpp",
+      "reads_vfolder_config_files": false,
+      "default_model_definition": {
+        "models": [
+          {
+            "name": "llamacpp-model",
+            "service": {
+              "start_command": [
+                "sh",
+                "-c",
+                "llama-server -m \"$(ls /models/*.gguf | head -n1)\" \"$@\"",
+                "llama-cpp"
+              ],
+              "port": 8080,
+              "health_check": {
+                "path": "/health",
+                "interval": 10.0,
+                "max_retries": 30,
+                "initial_delay": 60.0
+              }
+            }
+          }
+        ]
+      }
+    },
+    {
       "name": "custom",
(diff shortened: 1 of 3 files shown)


fixtures/manager/example-runtime-variants.json
```diff
@@ -130,6 +130,33 @@
...
+      "reads_vfolder_config_files": false,
+      "default_model_definition": {
+        "models": [
+          {
+            "name": "llamacpp-model",
+            "service": {
+              "start_command": [
+                "sh",
+                "-c",
+                "llama-server -m \"$(ls /models/*.gguf | head -n1)\" \"$@\"",
```
seedspirit, line 145: `sh -c` is needed because llama.cpp takes a single .gguf file. Unlike vllm/sglang which receive the mounted vfolder directory via `{model_path}`. So the command shell-globs the model mount to pick the first `.gguf`

jopemachine: It concerns me that the behavior appears to be arbitrary when more than one .gguf file is present.

fregataa (approved):

PR #11947, opened 2026-06-05
feat(BA-6283): add permission cap enum and scope_entities cap column

fregataa:
## Summary
- Introduce a `PermissionCap` IntEnum (`READ_ONLY`/`READ_WRITE`/`WRITE_DELETE`) with a cumulative cap → `OperationType` mapping and a `cap_operations()` helper, representing the maximum assignable permission (ceiling) on a scope-entity edge.
- Add an `IntEnumType` column decorator (SMALLINT-backed) so IntEnum columns store their ordinal value, preserving ordering in the database.
- Add a NOT NULL `permission` column to `association_scopes_entities` (server default `WRITE_DELETE`) and reflect it in `AssociationScopesEntitiesData`.
- Alembic migration backfills the cap from `relation_type` (`REF` → `READ_ONLY`, `AUTO` → `WRITE_DELETE`).

Switching permission resolution to the cap column and any write-path handling are out of scope (follow-ups).

## Test plan
- [x] Alembic upgrade/downgrade roundtrip verified against local DB
- [x] Backfill verified: `auto` rows → `30`, `ref` rows → `10`, 0 mismatches
- [x] CI: lint / type check / tests pass

Resolves BA-6283

### DIFF
--- a/src/ai/backend/common/data/permission/types.py
+++ b/src/ai/backend/common/data/permission/types.py
@@ -538,3 +538,39 @@ def owner_operations(entity_type: RBACElementType) -> frozenset[OperationType]:
 def member_operations(entity_type: RBACElementType) -> frozenset[OperationType]:
     """Operations granted to a *member* role on the given entity type."""
     return _MEMBER_OPS_OVERRIDES.get(entity_type, _DEFAULT_MEMBER_OPS)
+
+
+# ---------------------------------------------------------------------------
+# Permission cap (ceiling) on scope-entity edges
+#
+# A cap expresses the *maximum* permission assignable through a scope-entity
+# edge, generalizing the binary auto/ref relation into ordered levels. Caps are
+# cumulative ceilings: a higher cap permits a superset of the operations a lower
+# cap permits, so the integer ordering of :class:`PermissionCap` is a faithful
+# proxy for the subset ordering of :func:`cap_operations`. The integer gaps
+# (10/20/30) leave room to insert intermediate levels without renumbering.
+# ---------------------------------------------------------------------------
+
+
+class PermissionCap(enum.IntEnum):
+    READ_ONLY = 10
+    READ_WRITE = 20
+    WRITE_DELETE = 30
+
+
+_READ_WRITE_OPS: frozenset[OperationType] = frozenset({
+    OperationType.CREATE,
+    OperationType.READ,
+    OperationType.UPDATE,
+})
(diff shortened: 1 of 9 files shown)


src/ai/backend/common/data/permission/types.py
```diff
@@ -538,3 +538,39 @@ def owner_operations(entity_type: RBACElementType) -> frozenset[OperationType]:
...
+# cap permits, so the integer ordering of :class:`PermissionCap` is a faithful
+# proxy for the subset ordering of :func:`cap_operations`. The integer gaps
+# (10/20/30) leave room to insert intermediate levels without renumbering.
+# ---------------------------------------------------------------------------
+
+
+class PermissionCap(enum.IntEnum):
+    READ_ONLY = 10
+    READ_WRITE = 20
+    WRITE_DELETE = 30
```
HyeockJinKim: If we're going to use an `int` enum, I was thinking we could use a flag-based approach.

src/ai/backend/common/data/permission/types.py
```diff
@@ -538,3 +545,26 @@ def owner_operations(entity_type: RBACElementType) -> frozenset[OperationType]:
...
+    * is one mask a subset of another?     ``(a & ~b) == Permission.NONE``
+    * intersect / union two masks          ``a & b`` / ``a | b``
+    """
+
+    NONE = 0
+    READ = 1
+    UPDATE = 2
+    CREATE = 4
+    SOFT_DELETE = 8
+    HARD_DELETE = 16
```
HyeockJinKim: Can't it be written as `1 << n`?

jopemachine (dismissed):

src/ai/backend/manager/models/alembic/versions/d69322160c90_add_permission_cap_column_to_.py
```diff
@@ -0,0 +1,44 @@
...
+    # relation default.
+    op.add_column(
+        "association_scopes_entities",
+        sa.Column(
+            "permission_cap",
+            sa.SmallInteger(),
+            server_default=sa.text("31"),
+            nullable=False,
+        ),
+    )
```
HyeockJinKim: Why is the default value for the server 31?

jopemachine (approved):

HyeockJinKim (approved):

PR #11948, opened 2026-06-06
fix(BA-3974): coordinate TUS upload offsets via Valkey + per-session lease

jopemachine:
## Summary

Multi-instance storage-proxy behind a load balancer returned spurious `409 Conflict` on spec-compliant sequential TUS PATCH requests because per-instance NFS attribute caches diverged on the upload temp file's size — the caller's `Upload-Offset` check failed against a stale `stat()` result that lagged the actual on-disk state.

This PR makes **Valkey the single source of truth for the per-session committed offset** and **serializes the destructive write section with a TTL'd per-session lease that a background watcher keeps alive across long body drains**.

### `common/clients/valkey_client/valkey_tus/` (new sub-package)

- `ValkeyTusClient` — atomic coordinator for the offset + lease pair:
  - `initialize_offset` — session creation.
  - `try_load_offset(session_id, holder_token)` — single MULTI/EXEC: `SET lease NX EX` + `GET offset`. Returns the canonical committed offset, or raises `TusLeaseHeldError` / `TusSessionNotFoundError`.
  - `advance_offset(session_id, holder_token, length)` — single Lua step: lease-ownership CAS + `INCRBY` + `EXPIRE` + `DEL`. Returns the new offset, or raises `TusLeaseLostError` when the lease was reclaimed mid-write.
  - `extend_lease(session_id, holder_token)` — single Lua step: CAS-EXPIRE. Returns `True` while we still own the lease, `False` once it is reclaimed.
  - `watch_lease(session_id, holder_token)` — async loop that calls `extend_lease` every `_DEFAULT_LEASE_HEARTBEAT_SECONDS` (10 s) until cancelled or the lease is lost. Spawned as a background task by the PATCH handler.
  - `release_lease(session_id, holder_token)` — idempotent CAS-DEL for cleanup paths (drain failure, precondition mismatch).
  - `get_offset` for HEAD.
- `valkey_tus/exceptions.py` — `TusLeaseHeldError` (409), `TusLeaseLostError` (409), `TusSessionNotFoundError` (404). All `BackendAIError` subclasses; storage handler lets them propagate as already-typed HTTP responses.
- `TusSessionId = NewType("TusSessionId", str)` — boundary-cast at call sites.
- New `REDIS_TUS_DB = 8`, `RedisRole.TUS`, `LayerType.VALKEY_TUS`.

### `storage/`

- `RootContext.valkey_tus_client` wired up in `server.py` and `migration.py`.
- `create_upload_session` (manager API) initializes the per-session counter to 0.
- `tus_upload_part` (client API):
  - `try_load_offset` → spawn `watch_lease` as a background task → drain body via `aiofiles.open` → cancel and `await` the watcher → `advance_offset`.
  - Failures along the way call `release_lease` so the lease does not linger.
  - `_drain_into_upload_file` uses `mode="w+b"` on the first PATCH (`start_offset == 0`) and `mode="r+b"` on continuations. A continuation with a vanished staging file (NFS GC, manual rm, etc.) surfaces as `TusSessionNotFoundError` instead of silently zero-padding the prefix via `truncate`.
  - Truncates to the Valkey-canonical offset before writing, streams the body in, `flush` + `fsync`.
  - No `stat()` calls on the offset path; no staging file (each byte hits NFS exactly once).
- `prepare_tus_session_headers` reads `Upload-Offset` from Valkey.
- `UploadOffsetMismatchError` stays as the only TUS-protocol-level error in storage. Lease/session domain errors live next to the client in `common`.

### Tests

- `tests/unit/common/clients/valkey_client/test_valkey_tus_client.py` — sibling pattern (real Valkey via `redis_container` fixture). Exercises the offset lifecycle: `initialize_offset` → `try_load_offset` → `advance_offset`.
- `tests/unit/storage/api/test_tus_upload.py` — mock surface updated for `try_load_offset` / `advance_offset` / `extend_lease` / `release_lease`.

## Why the lease + watcher

Per-instance retries, LB round-robin, or response-loss-induced re-PATCH can route two PATCH requests at the same offset to two different storage-proxy instances concurrently. With a single shared upload file, file writes are not commutative and `ftruncate(known_good_offset)` is destructive — both need serialization. `INCRBY` alone is atomic but does not order disk writes. The lease is the smallest serialization that covers this.

The 30 s lease TTL would expire mid-write for any chunk larger than (network bandwidth × 30 s). The WebUI's `getOptimalChunkSize` ramps from 15 MB up to 200 MB for files ≥ 5 GB, so without renewal a single chunk over a typical WAN would trip `TusLeaseLostError`. The watcher refreshes the lease every 10 s for as long as the PATCH is in flight, so an arbitrarily slow PATCH stays valid as long as the storage-proxy itself is alive. If the lease is ever reclaimed (e.g. the storage-proxy stalled), the watcher exits on its own and the eventual `advance_offset` surfaces the conflict via `TusLeaseLostError` → client HEAD-resumes.

## Round-trip budget

| Phase | Round-trips | Mechanism |
|-------|-------------|-----------|
| `try_load_offset` | 1 | Batch (`SET NX EX` + `GET`) |
| `watch_lease` (per heartbeat) | 1 every 10 s while the PATCH is in flight | Lua CAS-EXPIRE |
| `advance_offset` | 1 | Lua (CAS + `INCRBY` + `EXPIRE` + `DEL`) |
| `release_lease` | 1 (only on cleanup paths) | Lua CAS-DEL |
| **Hot path per PATCH** | **2 + N heartbeats** (N = ceil(drain_duration / 10s)) | — |

## Test plan

- [x] `pants fmt / fix / lint / check` — clean for all touched source files.
- [x] `pants test tests/unit/storage/api/test_tus_upload.py` — green.
- [x] `pants test tests/unit/common/clients/valkey_client/test_valkey_tus_client.py` — green (real Valkey).
- [x] BA-3974 reproduction rig (2 storage-proxy + NFS + nginx LB): **29/30 PATCH → 409** on `main` baseline.
- [x] Same rig with this patch: **30/30 PATCH → 204, SHA-256 match**.
- [x] **Slow-PATCH watcher verification**: a single 50 s PATCH (5 MB body trickled across 50 s, well past the 30 s lease TTL) succeeded with `Upload-Offset: 5242880`, `sha_match=True`. Storage-proxy logs show `extend_lease` firing at t+10s/t+20s/t+30s/t+40s. Without the watcher this would have hit `TusLeaseLostError`.
- [ ] CI green.

## Out of scope

- BA-6155 chunk-per-file engine (parallel work; separate PR).
- Disk-side orphan upload-file reaper (recommended but independent).


src/ai/backend/common/clients/valkey_client/valkey_tus/client.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+        await client.connect()
+        return cls(client=client)
+
+    @valkey_tus_resilience.apply()
+    async def close(self) -> None:
+        await self._client.disconnect()
+
+    @staticmethod
+    def _key(session_id: TusSessionId) -> str:
```
jopemachine: offset_key

jopemachine:
CI seems broken on main

src/ai/backend/common/clients/valkey_client/valkey_tus/client.py
```diff
@@ -0,0 +1,167 @@
...
+            )
+
+    @valkey_tus_resilience.apply()
+    async def get_offset(self, session_id: TusSessionId) -> int | None:
+        async with self._client.client() as conn:
+            raw = await conn.get(self._offset_key(session_id))
+        if raw is None:
+            return None
+        decoded = raw.decode() if isinstance(raw, bytes) else raw
+        return int(decoded)
```
HyeockJinKim: 1. Rather than providing `acquire` and `release` separately, wouldn’t it be better to pass `acquire + offset` to `AsyncIterator` (and raise an exception if `acquire` fails)? Also, I’d like this to be guaranteed in a single request via a Lua script. (Multiple requests would cause a performance bottleneck.)
2. It seems better to combine the `advance_offset` with the query and provide them together. (Considering this, the `AsyncIterator` does seem a bit ambiguous.)

HyeockJinKim (dismissed):
It doesn't seem to be a major issue. If the offset isn't updated in Redis after it's uploaded, the same offset might be entered again. Please just add a test to verify that there are no file write issues when the previous offset is received.

HyeockJinKim (approved):

PR #11953, opened 2026-06-06
fix(BA-6287): prevent garbage collection of fire-and-forget asyncio tasks

fregataa:
## Summary
- The event loop only keeps weak references to tasks, so a fire-and-forget `asyncio.create_task()` whose return value is not retained may be garbage collected mid-execution.
- `runner.types.Runner`: track observer/run tasks in a `set` and discard them on completion, keeping a strong reference while they run.
- `bgtask.BackgroundTaskManager`: move local cron startup out of `__init__` into an async `create()` factory and accept a `BackgroundTaskManagerArgs` dataclass; the cron's task runners already hold strong references internally, so no dangling wrapper task remains.
- Re-enable the `RUF006` (asyncio-dangling-task) lint rule.

## Test plan
- [x] `pants test tests/unit/common/bgtask::`
- [x] `pants test tests/component::` (background_task_manager fixture)
- [x] CI lint/check passes with RUF006 enabled

Resolves BA-6287

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -147,7 +147,6 @@ ignore = [
     # RUF
     "RUF003", # ambiguous-unicode-character-comment - intentional use of special characters (e.g., × for multiplication)
     "RUF005", # collection-literal-concatenation - explicit concatenation is sometimes clearer
-    "RUF006", # asyncio-dangling-task - fire-and-forget tasks are intentional in some cases
     "RUF009", # function-call-in-dataclass-default - cast() in dataclass defaults is intentional
     "RUF012", # mutable-class-default - existing codebase uses mutable defaults intentionally
     "RUF015", # unnecessary-iterable-allocation-for-first-element - [0] is clearer than next(iter())
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -845,7 +845,6 @@ class AbstractAgent[
     # Health monitoring tracking
     _active_pulls: dict[str, PullTaskInfo]  # key: image canonical name
     _active_creates: dict[KernelId, CreateTaskInfo]
-    _model_service_health_tasks: set[asyncio.Task[Any]]
 
     @contextmanager
     def track_pull(self, image: str) -> Generator[bool, None, None]:
@@ -929,7 +928,6 @@ class AbstractAgent[
         # Initialize health monitoring tracking maps
         self._active_pulls = {}
         self._active_creates = {}
-        self._model_service_health_tasks = set()
         self._sync_container_lifecycle_observer = SyncContainerLifecycleObserver.instance()
         self._clean_kernel_registry_task = asyncio.create_task(self._clean_kernel_registry_loop())
 
@@ -990,7 +988,7 @@ class AbstractAgent[
             human_readable_name="agent.kernel_presence",
             db_id=REDIS_LIVE_DB,
         )
-        self.background_task_manager = await BackgroundTaskManager.create(
+        self.background_task_manager = BackgroundTaskManager(
             BackgroundTaskManagerArgs(
                 event_producer=self.event_producer,
                 valkey_client=self.valkey_bgtask_client,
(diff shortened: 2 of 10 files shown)


src/ai/backend/common/runner/types.py
```diff
@@ -28,10 +29,21 @@ class Runner:
...
     _resources: Sequence[AbstractResource]
     _closed_event: asyncio.Event
+    _tasks: set[asyncio.Task[None]]
 
     def __init__(self, resources: Sequence[AbstractResource]) -> None:
         self._resources = resources
         self._closed_event = asyncio.Event()
+        self._tasks = set()
+
+    def _track(self, coro: Coroutine[Any, Any, None]) -> asyncio.Task[None]:
```
seedspirit: Is it necessary to return the task?

src/ai/backend/common/bgtask/bgtask.py
```diff
@@ -266,39 +276,38 @@ class BackgroundTaskManager:
...
         ])
-        self._task_registry = task_registry or BackgroundTaskHandlerRegistry()
-        self._local_cron = LocalCron([
-            BgtaskHeartbeatTask(self),
-            BgtaskRetryTask(self),
+        self._task_registry = args.task_registry or BackgroundTaskHandlerRegistry()
+        self._local_cron = None
+
+    @classmethod
+    async def create(cls, args: BackgroundTaskManagerArgs) -> Self:
```
seedspirit: Is the reason for having a separate initialization classmethod because of `await local_cron.start()`?

fregataa: yes

src/ai/backend/common/bgtask/bgtask.py
```diff
@@ -266,39 +276,38 @@ class BackgroundTaskManager:
...
+        """Construct a manager and start its local cron in an async context."""
+        instance = cls(args)
+        local_cron = LocalCron([
+            BgtaskHeartbeatTask(instance),
+            BgtaskRetryTask(instance),
         ])
-        asyncio.create_task(self._local_cron.start())
+        await local_cron.start()
```
HyeockJinKim: The local cron job itself is created in the constructor, and only the start operation is separated.

src/ai/backend/common/bgtask/bgtask.py
```diff
@@ -266,39 +276,38 @@ class BackgroundTaskManager:
...
+        # No setter because we don't want this to be changed after initialization
+        instance._local_cron = local_cron
+        return instance
```
HyeockJinKim: It seems a bit odd to set this up in a `classmethod` rather than in the constructor.

pyproject.toml
```diff
@@ -147,7 +147,6 @@ ignore = [
     # RUF
     "RUF003", # ambiguous-unicode-character-comment - intentional use of special characters (e.g., × for multiplication)
     "RUF005", # collection-literal-concatenation - explicit concatenation is sometimes clearer
-    "RUF006", # asyncio-dangling-task - fire-and-forget tasks are intentional in some cases
```
HyeockJinKim, line 150: 👍

jopemachine (approved):

src/ai/backend/agent/agent.py
```diff
@@ -1551,11 +1555,13 @@ async def _clean_kernel_registry_loop(self) -> None:
...
                         ", ".join(map(str, dangling_kernel_ids)),
                     )
-                    asyncio.create_task(
+                    cleanup_task = asyncio.create_task(
                         asyncio.wait_for(
                             self.clean_kernel_objects(dangling_kernel_ids), timeout=30.0
                         )
                     )
+                    cleanup_tasks.add(cleanup_task)
+                    cleanup_task.add_done_callback(cleanup_tasks.discard)
```
HyeockJinKim, line 1564: Is garbage collection guaranteed even in the event of an exception?

Issue #11962, opened 2026-06-08
Expose per-mount metadata (subpath, alias, permission) on GraphQL Session nodes


rapsealk:
## Summary

A vfolder mount on a session carries per-mount metadata beyond the vfolder identity — **subpath**, **alias** (custom mount destination), and an effective **permission**. This data is persisted on the session but is **not exposed through any GraphQL Session schema**, so the WebUI cannot render it.

## Background

When mounting a vfolder, the caller may set:

| Concept | Stored field (`VFolderMount`) | Meaning |
|---|---|---|
| subpath | `vfsubpath` (`PurePosixPath`, default `.`) | which subdirectory inside the vfolder is mounted |
| alias | `kernel_path` (`PurePosixPath`, default `/home/work/{name}`) | the mount destination path inside the container |
| permission | `mount_perm` | per-mount `ro`/`rw`/`wd` override |

These are stored in the `sessions.vfolder_mounts` JSON column (`StructuredJSONObjectListColumn(VFolderMount)`, `src/ai/backend/manager/models/session/row.py:512`).

## The gap (current state)

Nothing in GraphQL exposes the per-mount `subpath`/`alias`/`permission`:

- **Legacy `ComputeSession.mounts`** (`gql_legacy/session.py:1022`) → deduped vfolder **names** only.
- **Legacy `ComputeSession.vfolder_mounts` / `ComputeSessionNode.vfolder_mounts`** (`:1023`, `:249`) → vfolder **UUIDs** only (already marked for deprecation, `:379`).
- **Legacy `vfolder_nodes` connection** (`:256`, resolver `:471`) → resolves the vfolder **entity** by `folder_id` and **discards** all per-mount fields.
- **New `SessionV2GQL`** (`api/gql/session/types.py:305`) → **no mount field at all**; explicit `# TODO: Add vfolder_mounts dynamic field` at `types.py:437`.

## Key finding — the data already flows to the adapter

No DB/loader change is required. `SessionData.vfolder_mounts: list[VFolderMountData] | None` (`manager/data/session/types.py:175`) is already populated (`models/session/row.py:733`) and is in hand inside the v2 adapter `_session_data_to_node()` (`api/adapters/session/adapter.py:901`) — it is simply dropped before reaching the schema. The work is confined to the **DTO → GraphQL output** path.

## Proposed output shape

A per-mount object mapped from `VFolderMountData`:

| Field | Source | Notes |
|---|---|---|
[...]

PR #11966, opened 2026-06-08
feat: Expose per-mount subpath, destination, and permission via the new SessionV2 mounts field

### LINKED ISSUE #11963: Expose per-mount subpath/alias/permission on the SessionV2 GraphQL node
Part of the per-mount metadata epic: #11962

## Goal

Expose per-mount **subpath**, **alias** (mount destination), and **permission** on the new Strawberry `SessionV2GQL` node, replacing the `# TODO: Add vfolder_mounts dynamic field` placeholder at `src/ai/backend/manager/api/gql/session/types.py:437`.

This is the v2-first path mandated by project rules (`api/gql/CLAUDE.md`). The data is already available in the v2 adapter, so no DB/loader/query change is needed.

## Changes

1. **DTO** — `src/ai/backend/common/dto/manager/v2/session/response.py`
   - Add `class SessionMount(BaseResponseModel)` with: `vfolder_id: UUID`, `name: str`, `subpath: str | None`, `mount_destination: str`, `permission: str`, `usage_mode: str`.
   - Add `mounts: list[SessionMount]` to `SessionNode`.

2. **Adapter** — `src/ai/backend/manager/api/adapters/session/adapter.py`, `_session_data_to_node()` (~line 901)
   - Populate `mounts=[SessionMount(...) for m in (data.vfolder_mounts or [])]`, mapping `vfid.folder_id`, `name`, `str(vfsubpath)` → `None` when `"."`, `str(kernel_path)`, `mount_perm.value`, `usage_mode.value`. Data is already present here.

3. **GQL type** — `src/ai/backend/manager/api/gql/session/types.py`
   - Add `SessionMountGQL` via `@gql_pydantic_type(model=SessionMount, name="SessionMount")` (use `name="SessionMountV2"` if supergraph composition collides — see federation caveat in `api/gql/CLAUDE.md`).
   - Replace the TODO at `:437` with a `mounts` field on `SessionV2GQL` using `gql_added_field(BackendAIGQLMeta(added_version=NEXT_RELEASE_VERSION, ...))` returning `list[SessionMountGQL]`. Plain field (data rides on the already-loaded DTO via `session_loader`).
   - **Optional follow-up:** add a `vfolder` resolver on `SessionMountGQL` → `VFolderGQL | None` (v2 type at `api/gql/vfolder_v2/types/node.py`) via `info.context.data_loaders` + `strawberry.lazy()`, so the UI can drill into the folder.

4. **Schema** — run `scripts/generate-graphql-schema.sh` and confirm supergraph composition succeeds.

5. **Tests** — `tests/unit/manager/...`
   - Adapter test: `SessionData` with a non-default `vfsubpath` + custom `kernel_path` + `ro` perm yields the expected `SessionNode.mounts`, incl. `"."` → `null` normalization.
   - GQL resolver test for the `mounts` field shape.

## Verification

Restart manager (`./dev restart mgr`) and query `SessionV2 { mounts { vfolderId name subpath mountDestination permission usageMode } }`. Cross-check against a session created with explicit subpath + custom mount path.

## Out of scope

- Legacy `ComputeSession` schema (separate sub-issue).
- Changing how mounts are stored or created.

rapsealk:
## Summary

Adds a `mounts` field to the Strawberry `SessionV2` GraphQL node, replacing the `# TODO: Add vfolder_mounts dynamic field` placeholder. Each entry is a `SessionMount` carrying mount-specific metadata, with folder-level attributes reachable through an associated `vfolder` node — the same vfolder may be mounted more than once under different subpaths.

The mount data already reaches the v2 adapter via `SessionData.vfolder_mounts`, so **no DB, loader, or query change was required**.

### `SessionMount` fields

| GraphQL field | Source | Notes |
|---|---|---|
| `vfolderId` | `vfid.folder_id` | |
| `subpath` | `vfsubpath` | `"."` (vfolder root) normalizes to `null` |
| `mountDestination` | `kernel_path` | Mount alias inside the container |
| `permission` | `mount_perm` | `ro` / `rw` / `wd` |
| `vfolder` | `vfolder_loader` | Resolver for folder-level attributes (name, usage mode, ownership) |

## Test plan

- [x] `pants fmt` / `fix` / `lint` / `check` (mypy) pass
- [x] `pants test tests/unit/manager/api/adapters/test_session_adapter.py` passes
- [x] GraphQL schema regenerated (`v2-schema.graphql`, `supergraph.graphql`)
- [x] Unit tests cover the adapter mapping: no mounts, full mount mapping, and `"."` → `null` subpath normalization
- [ ] Live verification pending

Part of #11962
Resolves #11963


src/ai/backend/manager/api/gql/session/types.py
```diff
@@ -295,6 +296,27 @@ class SessionV2NetworkInfoGQL:
...
+    vfolder_id: UUID = gql_field(description="UUID of the mounted virtual folder.")
+    name: str = gql_field(description="Name of the mounted virtual folder.")
+    subpath: str | None = gql_field(
+        description="Subpath within the vfolder that is mounted. Null when the vfolder root is mounted."
+    )
+    mount_destination: str = gql_field(
+        description="Mount destination (alias) path inside the session container."
+    )
+    permission: str = gql_field(description="Effective mount permission (e.g., 'ro', 'rw').")
+    usage_mode: str = gql_field(description="Usage mode of the mounted virtual folder.")
```
jopemachine: It would be better to add a vfolder node field so that mount-specific information (e.g., subpath, permission, etc.) and VFolder information can be queried separately in a structured manner.

src/ai/backend/manager/api/gql/session/types.py
```diff
@@ -295,6 +296,27 @@ class SessionV2NetworkInfoGQL:
...
+class SessionMountGQL:
+    vfolder_id: UUID = gql_field(description="UUID of the mounted virtual folder.")
+    name: str = gql_field(description="Name of the mounted virtual folder.")
+    subpath: str | None = gql_field(
+        description="Subpath within the vfolder that is mounted. Null when the vfolder root is mounted."
+    )
+    mount_destination: str = gql_field(
+        description="Mount destination (alias) path inside the session container."
+    )
+    permission: str = gql_field(description="Effective mount permission (e.g., 'ro', 'rw').")
```
jopemachine: Since the VFolderMountPermissionGQL, VFolderUsageModeGQL type already exists, it would be better to use them

src/ai/backend/manager/api/gql/session/types.py
```diff
@@ -295,6 +296,27 @@ class SessionV2NetworkInfoGQL:
     network_id: strawberry.auto
 
 
+@gql_pydantic_type(
+    BackendAIGQLMeta(
+        added_version=NEXT_RELEASE_VERSION,
+        description="A single virtual folder mount on a session.",
+    ),
+    model=SessionMount,
```
jopemachine: Nit: SessionMountDTO

src/ai/backend/manager/api/gql/session/types.py
jopemachine: There used to be a convention that Node types were only attached to root types, but that convention does not seem applicable in this case.

Since the same vfolder can be mounted multiple times under different subpaths, it makes more sense for the mount itself to be associated with a vfolder node. This more accurately reflects the relationship between mounts and vfolders.

src/ai/backend/manager/api/gql/session/types.py
rapsealk: Renamed `SessionMount` → `SessionMountDTO`.

src/ai/backend/manager/api/gql/session/types.py
rapsealk: Done — `permission` now uses the existing `VFolderMountPermission` GQL enum (and `VFolderPermissionField` on the DTO side) instead of a bare string. `usage_mode` was dropped from the mount type itself (see the vfolder-node thread); it is now reachable via `vfolder.metadata.usageMode`.

src/ai/backend/manager/api/gql/session/types.py
rapsealk: Good call — restructured accordingly. Each mount now exposes a `vfolder: VFolder` node (the v2 Strawberry `VFolderGQL`, resolved through the existing `vfolder_loader`), so folder-level information can be queried separately in a structured manner. As you noted, the same vfolder can be mounted multiple times under different subpaths, so the relationship lives on the mount itself.

The `SessionMount` type now carries only mount-specific fields — `vfolderId`, `subpath`, `mountDestination`, `permission` — while `name`/`usageMode` moved to `vfolder.metadata`:

```graphql
type SessionMount {
  vfolderId: UUID!
  subpath: String
  mountDestination: String!
  permission: VFolderMountPermission!
  vfolder: VFolder
}
```

jopemachine:
Could you update this PR to the latest state? @rapsealk

jopemachine (dismissed):

jopemachine (dismissed):

seedspirit:
@rapsealk  Please resolve conflict

Issue #11965, opened 2026-06-08
v2 client: four more session/infra methods send query params in the request body


rapsealk:
## Summary

A follow-up audit of #11956 / #11958 found **four more v2 client methods** that send their parameters in the request **body** while the manager handler reads them from the **query string** (`QueryParam`). These were missed by the earlier `request=`-shaped greps because they use the lower-level `self._client._request(..., json=...)` / `self._client.download(json=...)` shapes.

## Affected methods

| Client method | Shape | Manager handler / binding | Route |
|---|---|---|---|
| `SessionClient.download_single` (`client/v2/domains/session.py`) | `.download(json=...)` | `download_single(query: QueryParam[DownloadSingleRequest])` (`rest/session/handler.py:1230`) — reads `file` | POST `/session/{name}/download_single` |
| `InfraClient.get_watcher_status` (`infra.py`) | `_request("GET", json=...)` | `QueryParam[WatcherStatusQuery]` (`rest/resource/handler.py:233`) — `agent_id` | GET `/resource/watcher` |
| `InfraClient.get_usage_per_period` (`infra.py`) | `_request("GET", json=...)` | `QueryParam[UsagePerPeriodQuery]` (`handler.py:194`) — `start_date`/`end_date` | GET `/resource/usage/period` |
| `InfraClient.get_usage_per_month` (`infra.py`) | `_request("GET", json=...)` | `QueryParam[UsagePerMonthQuery]` (`handler.py:171`) — `month`, **`group_ids: list`** | GET `/resource/usage/month` |

Because several fields are **required**, these requests currently fail outright (422 / missing param) rather than just dropping an optional filter.

## Client signing fragility (also fixed here)

`client/v2/base_client.py::_request` builds the signed `rel_url` from raw, unencoded params (`"&".join(f"{k}={v}")`) while aiohttp re-encodes the wire query string. They diverge for any value with special characters (e.g. `download_single`'s `file` path), breaking HMAC auth. The manager signs `request.raw_path` (`rest/middleware/auth.py:389`), so the client must produce a byte-identical encoded query string (and encode list values as repeated keys).

## Fix plan

- Make `base_client._request`/`download` build the URL+query via `yarl` so the signed `rel_url` matches the wire and list values encode as repeated keys.
- Route the four client methods through query params instead of the body.
- Regression tests asserting params (not body).

## Dependency

[...]

PR #11970, opened 2026-06-08
fix: Parse list-typed query parameters in manager `QueryParam`

### LINKED ISSUE #11969: Manager QueryParam.from_query cannot parse list-typed query parameters
## Summary

`common/api_handlers.py::QueryParam.from_query` validates the request model directly from the aiohttp multidict (`model_validate(request.query)`). Because a multidict collapses to a single value per key under the plain `Mapping` protocol, a **list-typed query field** (e.g. `group_ids: list[uuid.UUID]`) cannot be parsed: repeated query params (`?group_ids=a&group_ids=b`) fail schema validation, and a single value validates a bare `str` against a `list` type and also fails. (Verified empirically.)

This blocks any GET endpoint whose query DTO has a list field — surfaced by `GET /resource/usage/month` (`UsagePerMonthQuery.group_ids`).

## Proposed fix

Make `QueryParam.from_query` list-aware:
- Keep scalar fields on their current first-value semantics (no behavior change).
- For list/tuple/set-typed fields, collect all values for the matching key via `MultiMapping.getall()`, honoring `validation_alias` / `AliasChoices`.

This is a shared change in `common/api_handlers.py` used by every GET endpoint, so it must preserve existing scalar behavior exactly and be covered by tests.

## Context

Split out from #11965 (client-side fixes). The client SDK side — encoding list values as repeated query keys and signing them — is handled there; this issue covers the manager-side parsing so the values are actually consumed.

rapsealk:
## Summary

`common/api_handlers.py::QueryParam.from_query` validated the request model directly from the aiohttp multi-dict (`model_validate(request.query)`). A multi-dict collapses to a single value per key under the plain `Mapping` protocol, so a **list-typed query field** (e.g. `UsagePerMonthQuery.group_ids: list[uuid.UUID]`) could not be parsed: repeated query params (`?group_ids=a&group_ids=b`) failed schema validation, and a single value validated a bare `str` against a `list` type and also failed. (Verified empirically.)

## Changes

- Add `_query_to_mapping`: scalar fields keep their existing first-value behavior; list/tuple/set-typed fields collect all values via `MultiMapping.getall()`, with query keys resolved through `validation_alias` / `AliasChoices`.
- `QueryParam.from_query` now routes through that helper.

This is a shared change in `common/api_handlers.py` used by every GET endpoint, so it preserves existing scalar behavior exactly.

## Tests

- `tests/unit/common/test_api_handlers.py` — end-to-end (real aiohttp client through `api_handler`): repeated query params parse into a list field, and a single value parses into a one-element list.

## Test plan

- [x] `pants test tests/unit/common/test_api_handlers.py`
- [x] `pants --changed-since=origin/main fmt lint check`
- [x] Pre-push hook ran the transitive dependent test sweep (every GET endpoint's handler path) — green.

## Why split out

This is the manager-side half of #11965. The client SDK half (encoding list values as repeated query keys + signing) is in its own PR. This change is independently useful and must merge **before** the client PR's `get_usage_per_month` works end-to-end.

Closes #11969


src/ai/backend/common/api_handlers.py
```diff
@@ -114,6 +117,45 @@ def from_body(self, json_body: str) -> Self:
...
+    """Resolve the query-string keys a field accepts, honoring validation aliases."""
+    alias = field_info.validation_alias
+    if isinstance(alias, str):
+        return [alias]
+    if isinstance(alias, AliasChoices):
+        return [choice for choice in alias.choices if isinstance(choice, str)]
+    return [name]
+
+
+def _query_to_mapping(model: type[BaseRequestModel], query: MultiMapping[str]) -> dict[str, Any]:
```
seedspirit: Is this the only place where this logic is used? If so, keeping it as an internal method within `QueryParam` rather than as a module might be a good way to prevent misuse. It seems like `HeaderParam` also uses MultiMapping, so it might be worth checking that as well.

seedspirit:
It would be helpful if you could include the issue number in the PR title

src/ai/backend/common/api_handlers.py
rapsealk: Good call — done in 6b1e190. `QueryParam.from_query` was the only caller, so I moved `_to_mapping` plus the `_is_sequence_annotation` / `_candidate_query_keys` helpers into `QueryParam` as private (static) methods, so they can't be reused/misused elsewhere.

On `HeaderParam`: I checked, and its only model in the codebase is `ExportFilenameHeader`, whose sole field (`filename: str | None`) is scalar — no list-typed header field exists today. So I intentionally left `HeaderParam.from_header` on the plain `model_validate` path rather than generalize speculatively. If a list-typed header field is ever added, the same encapsulated approach can be lifted into `HeaderParam`.

rapsealk:
> It would be helpful if you could include the issue number in the PR title

This isn't tracked by a Jira ticket (e.g. `BA-XXXX`), just a GitHub issue, so I've kept the title prefix-only per our convention and skipped the number. The issue is still linked via `Closes #11969` in the PR body, so it'll auto-close on merge. Happy to add it back if you'd prefer the number in the title regardless.

src/ai/backend/common/api_handlers.py
```diff
@@ -174,9 +135,47 @@ def parsed(self) -> TRequestModel:
...
+        for name, field_info in self._model.model_fields.items():
+            if not self._is_sequence_annotation(field_info.annotation):
+                continue
+            for key in self._candidate_query_keys(name, field_info):
+                if key in query:
+                    data[key] = query.getall(key)
+                    break
+        return data
+
+    @staticmethod
```
seedspirit: Wouldn't a regular function be better than a static method?

rapsealk: Agreed — `@staticmethod` wasn't pulling its weight here, since both helpers are stateless and don't touch `self`/`cls`. Inlined them as nested functions inside `_to_mapping` (their only caller). That drops the decorator per your suggestion and keeps them fully encapsulated (not callable from outside at all), which also covers the "prevent misuse" point from the earlier thread. Done in the latest commit.

seedspirit (approved):

PR #11972, opened 2026-06-08
chore(BA-6289): announce HA Sentinel master/replica by hostname for host access

jopemachine:
Resolves BA-6289

## Summary

The HA development halfstack Sentinels failed to serve a host-reachable master address on macOS, so the manager/agent could not connect to Valkey/Redis through Sentinel:

```
glide_shared.exceptions.ClosingError:
  Connection error: Standalone(Received error for address `172.18.0.3:9500`: timed out)
```

**Root cause:** `configs/redis/sentinel.conf` sets `resolve-hostnames yes` (so `sentinel monitor mymaster node01 9500` is accepted) but does **not** set `announce-hostnames yes`. With the default (`no`), Sentinel resolves `node01` to the container's Docker bridge IP and **returns that IP** (e.g. `172.18.0.3`) to clients via `get-master-addr-by-name`.

- On **Linux** the Docker bridge subnet is routable from the host, so the host-side manager/agent reached the internal IP directly — masking the gap. The `/etc/hosts` entries `install-dev.sh` prints were effectively unused for the Redis/Valkey path.
- On **macOS** (Docker Desktop runs the engine in a VM) the bridge subnet is **not** routable from the host; only published ports on `localhost` work. The returned internal IP is unreachable → timeout.

This is a pre-existing gap (applies to Redis too), surfaced while verifying the Valkey dev-halfstack migration (#11900 / BA-6263).

## Changes

- `configs/redis/sentinel.conf`: add `sentinel announce-hostnames yes`. Sentinel now advertises the master/replica by hostname (`node01`-`node03`), which the host maps to `127.0.0.1` via `/etc/hosts` and reaches through the published ports.
- `scripts/install-dev.sh`: the `/etc/hosts` `node01-03` entries are now stated as **REQUIRED** (previously a `[NOTE]`), with the rationale, since clients now receive hostnames.

## Test plan

- [x] Regenerated the Sentinel configs from the committed template (same substitution as `install-dev.sh`); all three contain `announce-hostnames yes`.
- [x] `SENTINEL get-master-addr-by-name mymaster` returns a **hostname** (`node01`/`node02`), not the Docker-internal IP.
- [x] On macOS, manager + agent connect to the master via Sentinel (`ValkeyClient for master at node0x:950x`), agent registers `ALIVE`, API healthy.
- [x] Session lifecycle on the HA Sentinel cluster: enqueue → RUNNING → terminate.
- [x] Failover: killed the master → Sentinel promoted a replica (~2s, data preserved), manager/agent glide clients auto-reconnected to the new master (`Monitor failure threshold (3) → Reconnecting`), post-failover ops succeeded.

### DIFF
--- a/configs/redis/sentinel.conf
+++ b/configs/redis/sentinel.conf
@@ -2,6 +2,9 @@ port REDIS_SENTINEL_SELF_PORT
 requirepass develove
 
 sentinel resolve-hostnames yes
+# Uncommented by install-dev.sh on macOS only (BA-6289), where bridge IPs are not
+# routable from the host; reaches master/replicas by hostname via /etc/hosts.
+# sentinel announce-hostnames yes
 sentinel announce-ip REDIS_SENTINEL_SELF_HOST
 sentinel announce-port REDIS_SENTINEL_SELF_PORT
 sentinel auth-pass mymaster develove
--- a/changes/11972.fix.md
+++ b/changes/11972.fix.md
@@ -1,0 +1,1 @@
+Fix the HA development halfstack being unusable on macOS by enabling `sentinel announce-hostnames yes` on macOS only, so Sentinel advertises the master/replicas by hostname (`node01`-`node03`) reachable through the `/etc/hosts` entries; Linux keeps announcing the directly-routable bridge IPs and needs no `/etc/hosts` changes.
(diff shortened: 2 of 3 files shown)


configs/redis/sentinel.conf
```diff
@@ -2,6 +2,9 @@ port REDIS_SENTINEL_SELF_PORT
 requirepass develove
 
 sentinel resolve-hostnames yes
+# Uncommented by install-dev.sh on macOS only (BA-6289), where bridge IPs are not
```
fregataa, line 5: Let's not leave any issue number here

fregataa (approved):

PR #12000, opened 2026-06-09
feat(BA-6326): add `required` flag to runtime variant presets

seedspirit:
## Summary
- Add a `required` boolean column to `runtime_variant_presets` (NOT NULL, server default `false`) so a preset param can be marked mandatory.
- Thread `required` through the full stack: ORM row + `to_data`, `RuntimeVariantPresetData`, creator/updater specs, v2 DTO request/response, GQL node + Create/Update inputs, and the adapter.
- Idempotent Alembic up/down migration.

This is a prerequisite for a separate task that validates required runtime-variant preset params are supplied on a deployment revision.

Semantics note for the follow-up: a preset is treated as mandatory only when `required=True` **and** it has no `default_value`.

## Test plan
- [x] Repository round-trip: create `required=True` persists; omit → `false`; update toggles true↔false
- [x] `pants lint check` clean; affected manager + common test packages pass
- [ ] CI green

Resolves BA-6326

----
📚 Documentation preview 📚: https://sorna--12000.org.readthedocs.build/en/12000/

----
📚 Documentation preview 📚: https://sorna-ko--12000.org.readthedocs.build/ko/12000/


src/ai/backend/manager/models/alembic/versions/a3f1c7e2b9d4_add_required_to_runtime_variant_presets.py
```diff
@@ -0,0 +1,36 @@
...
+            sa.Column("required", sa.Boolean(), nullable=False, server_default=sa.false()),
+        )
+
+
+def downgrade() -> None:
+    conn = op.get_bind()
+    inspector = sa.inspect(conn)
+    columns = [c["name"] for c in inspector.get_columns("runtime_variant_presets")]
+    if "required" in columns:
+        op.drop_column("runtime_variant_presets", "required")
```
HyeockJinKim: Why run the if statement when this condition is definitely not true?

HyeockJinKim (approved):

PR #12011, opened 2026-06-09
test(BA-6345): align pytest resolve aiohttp with production to fix aioresponses

jopemachine:
## Summary

`tests/unit/storage-proxy/artifacts/test_service.py::TestHuggingFaceDownloadStep::test_probe_head_raises_on_http_error` (and any other `aioresponses`-mocked test) fails with:

```
TypeError: ClientResponse.__init__() missing 1 required keyword-only argument: 'stream_writer'
```

### Root cause

The **pytest resolve** pulled **aiohttp 3.14.0** transitively through `pytest-aiohttp`, while the **production resolve** pins `aiohttp~=3.13.3` (resolved 3.13.5). aiohttp 3.14 made `stream_writer` a required keyword-only argument of `ClientResponse.__init__`, but `aioresponses` 0.7.8 (the latest release on PyPI) does not pass it — so every `aioresponses`-mocked request raises the error above.

Two problems compounded:
1. The test environment ran a **different aiohttp major** (3.14) than production (3.13), which is a latent risk on its own.
2. Pants had **cached a stale success** for this test (produced before the aiohttp bump), so `main` CI never re-ran it. Any PR touching a transitive dependency of the test invalidates that cache and surfaces the failure.

This is split out of #11995 (BA-6333) since it is an independent, pre-existing issue on `main`.

## Changes

- Pin `aiohttp~=3.13.3` in `tools/pytest-requirements.txt` so the pytest resolve matches the production resolve.
- Regenerate `tools/pytest.lock` (aiohttp `3.14.0` → `3.13.5`).
- The test keeps using `aioresponses` as intended — no test-code workaround.

## Verification

```
pants test tests/unit/storage-proxy/artifacts/test_service.py   # ✓ all pass
```

### DIFF
--- a/changes/12011.test.md
+++ b/changes/12011.test.md
@@ -1,0 +1,1 @@
+Pin `aiohttp~=3.13.3` in the pytest resolve so it matches the production resolve; otherwise `pytest-aiohttp` floats aiohttp to 3.14.x, whose `ClientResponse` requires a `stream_writer` argument that `aioresponses` does not pass, breaking mocked HTTP tests.
--- a/tools/pytest-requirements.txt
+++ b/tools/pytest-requirements.txt
@@ -6,4 +6,8 @@ pytest-custom_exit_code>=0.3.0
 pytest-dependency>=0.6.0
 pytest-mock>=3.14.0
 pytest-timeout>=2.3.1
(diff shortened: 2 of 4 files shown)


jopemachine:
Tracking: BA-6345 (https://lablup.atlassian.net/browse/BA-6345)

fregataa (approved):

PR #12015, opened 2026-06-09
fix(BA-6256): raise 5xx when user system role is missing

fregataa:
## Summary
- `VfolderRepository._get_user_role_id` now raises `UserSystemRoleNotProvisioned` (5xx) instead of `ObjectNotFound` (404). A missing RBAC SYSTEM role is a server-side data-integrity condition (e.g. legacy/externally-provisioned accounts), to be remediated via the superadmin ensure-system-role API — not a client error.
- Consolidated duplicated invitation state-machine tests into `test_vfolder_management.py` and removed `TestVFolderInvitationStateMachine` from `test_vfolder_sharing.py`.
- Added a `user_system_role` fixture that provisions the SYSTEM role (RoleRow + UserRoleRow + scope + owner permissions), un-xfailed the accept-invitation test, and added a test asserting the 5xx error when the invitee has no system role.

## Test plan
- [x] `pants test tests/component/vfolder::` passes
- [x] `test_invitee_accepts_invitation` passes (no longer xfail)
- [x] `test_accept_fails_when_invitee_has_no_system_role` asserts 5xx

Resolves BA-6256

### DIFF
--- a/tests/component/vfolder/test_vfolder_sharing.py
+++ b/tests/component/vfolder/test_vfolder_sharing.py
@@ -11,11 +11,6 @@ from ai.backend.client.exceptions import BackendAPIError
 from ai.backend.client.v2.registry import BackendAIClientRegistry
 from ai.backend.common.dto.manager.field import VFolderPermissionField
 from ai.backend.common.dto.manager.vfolder import (
-    AcceptInvitationReq,
-    DeleteInvitationReq,
-    InviteVFolderReq,
-    InviteVFolderResponse,
-    ListInvitationsResponse,
     ListSharedVFoldersQuery,
     ListSharedVFoldersResponse,
     MessageResponse,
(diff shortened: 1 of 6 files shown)


tests/component/vfolder/test_vfolder_sharing.py
```diff
@@ -81,188 +75,6 @@ async def test_share_accept_access_unshare(
         assert user_email in unshare_result.unshared_emails
 
 
-class TestVFolderInvitationStateMachine:
-    """Invitation state transitions: verifies all PENDING → ACCEPTED/REJECTED/CANCELED
-    paths and the list_invitations query for pending invitations."""
```
fregataa, line 86: This test is duplicate of `tests/component/vfolder/test_vfolder_management.py:TestVFolderInviteAcceptReject`

seedspirit (approved):

PR #12016, opened 2026-06-10
feat(BA-6343): honor max_retries and interval in route health checks

HyeockJinKim:
## Summary
- The route health-check loop now honors `ModelHealthCheck.max_retries`, `interval`, `max_wait_time`, and `expected_status_code`, which were stored but previously ignored (only `initial_delay` was enforced).
- The health observer accumulates a per-route `consecutive_failures` counter in Valkey (reset on a passing probe) and throttles each route to its configured `interval`; the running-route handler keeps a route HEALTHY until the count reaches `max_retries`, then marks it UNHEALTHY. A missing status stays DEGRADED.
- Warming-up / `initial_delay` behavior is unchanged (already guaranteed by stage separation).

## Test plan
- [x] Unit tests: config judgment methods, observer interval throttle + failure count, RUNNING classification (within-budget HEALTHY / exhausted UNHEALTHY), Valkey round-trip + per-route TTL
- [ ] Live: deploy a model whose health endpoint starts failing after it reaches RUNNING; confirm the route stays HEALTHY for `max_retries` probes then flips to UNHEALTHY

Resolves BA-6343

### DIFF
--- a/src/ai/backend/common/config.py
+++ b/src/ai/backend/common/config.py
@@ -211,6 +211,32 @@ class ModelHealthCheck(BaseConfigModel):
         ge=0,
     )
 
+    def is_retry_exhausted(self, consecutive_failures: int) -> bool:
+        """Whether consecutive failed probes have reached ``max_retries``.
+
+        The health observer accumulates ``consecutive_failures`` (reset to 0
+        on a passing probe); the consuming handler calls this to decide when
+        a route should transition to UNHEALTHY.
+        """
+        return consecutive_failures >= self.max_retries
+
+    def is_probe_due(self, last_check: int, now: int) -> bool:
+        """Whether ``interval`` has elapsed since the last probe.
+
+        ``last_check`` / ``now`` are Unix-second timestamps. Used by the
+        observer to throttle per-route probing to the configured interval.
+        """
+        return now - last_check >= self.interval
+
+    def health_status_ttl_sec(self) -> int:
+        """Per-route TTL for the cached health status in Valkey.
+
+        Key expiry signals DEGRADED (no recent check), so the TTL must stay
+        comfortably above the probe ``interval`` — normal probing refreshes
+        it every ``interval`` with margin for scheduling jitter.
+        """
+        return max(120, int(self.interval * 3))
+
 
 class ModelServiceConfig(BaseConfigModel):
     pre_start_actions: list[PreStartAction] = Field(
--- a/src/ai/backend/common/config.py
+++ b/src/ai/backend/common/config.py
@@ -481,6 +481,18 @@ class ModelDefinition(BaseConfigModel):
                 return model.service.health_check
         return None
 
+    def health_check_setting(self) -> ModelHealthCheck | None:
+        """The configured health check, preserved even when disabled.
+
+        Unlike :meth:`health_check_config`, this does not gate on ``enable``,
+        so a caller that persists the setting keeps the ``enable`` flag intact
+        and consumers decide based on it rather than on presence alone.
+        """
+        for model in self.models:
+            if model.service and model.service.health_check:
+                return model.service.health_check
+        return None
+
     def with_args_appended(self, args: list[str]) -> ModelDefinition:
         """Return a copy with ``args`` appended to each model's
         ``service.start_command`` as separate argv tokens.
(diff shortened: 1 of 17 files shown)


src/ai/backend/manager/sokovan/deployment/route/handlers/observer/health_check.py
```diff
@@ -53,53 +71,81 @@ async def observe(self, routes: Sequence[RouteData]) -> RouteObservationResult:
...
+            self._valkey_schedule.get_redis_time(),
+        )
+
+        # Select routes that have a probe target and are due for a probe (interval throttle).
+        # A probe target only exists when health_check is set (see _build_probe_target),
+        # so health_check is non-None for every selected route.
+        plans: list[_ProbePlan] = []
+        for route in routes:
+            target = probe_targets.get(route.route_id)
+            if target is None or route.health_check is None:
```
seedspirit: How about
```suggestion
            if target is None or route.health_check is None or route.health_check.enable is False:
```

src/ai/backend/common/config.py
```diff
@@ -211,6 +211,32 @@ class ModelHealthCheck(BaseConfigModel):
...
+        return now - last_check >= self.interval
+
+    def health_status_ttl_sec(self) -> int:
+        """Per-route TTL for the cached health status in Valkey.
+
+        Key expiry signals DEGRADED (no recent check), so the TTL must stay
+        comfortably above the probe ``interval`` — normal probing refreshes
+        it every ``interval`` with margin for scheduling jitter.
+        """
+        return max(120, int(self.interval * 3))
```
seedspirit, line 238: Could you explain why 3 mulitiplication is needed?

src/ai/backend/manager/sokovan/deployment/route/executor.py
```diff
@@ -480,12 +481,20 @@ async def check_route_health(self, routes: Sequence[RouteData]) -> RouteExecutio
 
             if status.healthy:
                 successes.append(route)
+            elif route.health_check is not None and not route.health_check.is_retry_exhausted(
+                status.consecutive_failures
+            ):
```
fregataa: How is it handled when route.health_check is None?

src/ai/backend/manager/sokovan/deployment/route/handlers/observer/health_check.py
```diff
@@ -53,53 +71,81 @@ async def observe(self, routes: Sequence[RouteData]) -> RouteObservationResult:
...
-            for (route, _target), is_healthy in zip(checkable, results, strict=False)
+            ReplicaHealthResult(
+                replica_id=plan.route_id,
+                healthy=is_healthy,
+                consecutive_failures=0 if is_healthy else plan.prev_failures + 1,
+                ttl_sec=plan.health_check.health_status_ttl_sec(),
+            )
+            for plan, is_healthy in zip(plans, results, strict=False)
         ]
         await self._valkey_schedule.record_route_health_statuses_batch(health_results)
```
fregataa, line 129: Is it possible to show health check failure reason or status code to clients?

fregataa (commented):
Is this correct?
- coordinator's route health check: Records route health check in batch
- executor's health check: Determine route's status transition by recorded health check

src/ai/backend/manager/sokovan/deployment/route/handlers/observer/health_check.py
HyeockJinKim: Oh... This should be based on "enable" instead of "None." I'll go ahead and change it.

src/ai/backend/manager/sokovan/deployment/route/handlers/observer/health_check.py
HyeockJinKim: It appears that the health check component of the retry count can be provided separately.

src/ai/backend/manager/sokovan/deployment/route/executor.py
HyeockJinKim: This situation only occurs when the health check is not "none."

jopemachine: Then how about including `route.health_check is not None` check into `is_retry_exhausted` for readability?

src/ai/backend/common/config.py
jopemachine: I agree that this value and the multiplier feel too arbitrary.

PR #12022, opened 2026-06-10
fix(BA-6346): create GraphQL DataLoaders per request instead of per worker

fregataa:
## Summary
- The Strawberry v2 `DataLoaders` instance was constructed once in `setup_api()` and shared across all requests, so each loader's `cache_map` persisted for the worker's lifetime. This caused cross-request cache reuse (stale values, unbounded memory growth, and a potential authorization bypass if a check were ever placed inside `load_fn`), plus cross-request batch coalescing.
- Construct `DataLoaders` per GraphQL operation instead: in `handle_gql_strawberry` for HTTP requests and in `SubscriptionExecutor._make_gql_context` for WS subscriptions. Caching and batching are now scoped to a single operation, which is the standard DataLoader usage pattern.
- Add component regression tests (`tests/component/dataloader/`) that exercise the real `/admin/gql/strawberry` endpoint with a real image adapter/service/repository stack: an image renamed or deleted between two requests must be visible to the second request. With a shared loader cache, both tests fail deterministically.

## Test plan
- [x] `pants test tests/component/dataloader::` — per-request cache isolation tests pass on this branch
- [x] Characterization tests from branch `test/dataloader-cross-request-batching` (shared-loader cache reuse, in-load_fn authz bypass via cache hit, cross-request batch coalescing) flip to `XPASS(strict)` on this fix, confirming all three flaws are resolved
- [x] `pants fmt` / `fix` / `lint` / `check` pass

Resolves BA-6346

### DIFF
--- a/tests/component/dataloader/conftest.py
+++ b/tests/component/dataloader/conftest.py
@@ -1,0 +1,141 @@
+"""Fixtures serving the real ``/admin/gql/strawberry`` endpoint backed by the
+real image stack (adapter → service → repository → DB), so DataLoader caching
+behaviour is observable end-to-end without mocking the data path.
+"""
+
+from __future__ import annotations
+
+import uuid
+from collections.abc import AsyncIterator
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+import sqlalchemy as sa
+from sqlalchemy.ext.asyncio.engine import AsyncEngine as SAEngine
+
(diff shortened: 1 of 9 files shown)


tests/component/dataloader/conftest.py
HyeockJinKim, line 1: The image implementation is a bit overkill for regression testing. Could you please write a test by adding a query for regression purposes?

HyeockJinKim (approved):

PR #12024, opened 2026-06-10
feat(BA-6269): expose storage-proxy usage fields on v2 VFolder GraphQL node

jopemachine:
## Summary
- Add `VFolderUsageInfoGQL` (`numFiles`, `usedBytes`) and expose `usage` on the Strawberry v2 `VFolderGQL` node as an **on-demand resolver field**: selecting it triggers a `GetVFolderUsageAction` that measures usage live through the storage proxy. This is a slow, backend-dependent operation (e.g., a full directory walk on vfs), so the cost is paid only when the client explicitly selects the field — including per-node in search results.
- Expose the configured quota limits as a `quota` field group (`VFolderQuotaInfo`: `maxSize`, `maxFiles`) on the `VFolder` node, following the existing `metadata` / `accessControl` / `ownership` nested-type convention. Quota limits are configuration read from the DB, so they stay on the node itself, separate from the live-measured `usage`.
- **BREAKING:** Remove the `usage` field from the REST v2 `VFolderNode` response (get/search and the `vfolder` payloads of create/update/clone). It was populated from the `vfolders.num_files` / `cur_size` DB columns, which are dead (no code path has ever written them), so the field effectively returned stale values. The field was nullable with `default=None`, so existing SDK clients parse responses without errors in both version-skew directions; consumers needing live statistics should select the new GraphQL `usage` field instead.
- The adapter only triggers the action and adapts the result to the DTO; the orchestration (vfolder lookup, unmanaged-folder short-circuit, storage-proxy call) lives in the service layer with the same single-entity RBAC validation as `get_v2`.
- Regenerate the GraphQL v2 schema and supergraph.

Resolves BA-6269.

----
📚 Documentation preview 📚: https://sorna--12024.org.readthedocs.build/en/12024/

----
📚 Documentation preview 📚: https://sorna-ko--12024.org.readthedocs.build/ko/12024/


src/ai/backend/manager/services/vfolder/actions/get_live_usage.py
```diff
@@ -0,0 +1,51 @@
...
+    def target_element(self) -> RBACElementRef:
+        return RBACElementRef(
+            element_type=RBACElementType.VFOLDER,
+            element_id=str(self.vfolder_uuid),
+        )
+
+
+@dataclass
+class GetVFolderLiveUsageActionResult(BaseSingleEntityActionResult):
+    vfolder_id: uuid.UUID
```
jopemachine: vfolder_uuid

src/ai/backend/manager/services/vfolder/processors/vfolder.py
```diff
@@ -328,6 +340,7 @@ def supported_actions(self) -> list[ActionSpec]:
             CreateVFolderV2Action.spec(),
             CreateUploadSessionV2Action.spec(),
             GetVFolderV2Action.spec(),
+            GetVFolderLiveUsageAction.spec(),
```
jopemachine: GetVFolderUsageAction

docs/manager/graphql-reference/supergraph.graphql
```diff
@@ -21038,6 +20868,19 @@ type VFolder implements Node
   """Path for unmanaged virtual folders."""
   unmanagedPath: String
 
+  """
+  Added in UNRELEASED. Capacity quota limit in bytes with human-readable display. Null if unlimited.
+  """
+  maxSize: BinarySizeInfo
+
+  """Added in UNRELEASED. File-count quota for the folder."""
+  maxFiles: Int!
```
HyeockJinKim: Please group the fields together.

jopemachine: Grouped them into a new `VFolderQuotaInfo` nested type (`quota { maxSize, maxFiles }`), following the existing `metadata`/`ownership`/`usage` convention. Applied across the DTO, adapter, and GQL layers in 8ff029028.

HyeockJinKim (dismissed):

HyeockJinKim (approved):

PR #12127, opened 2026-06-10
docs: add BEP-1053 for the ROUTER frontend mode (LLM inference gateway)

achimnol:
## Summary

Adds **BEP-1053: ROUTER Frontend Mode — LLM Inference Gateway Integration**, proposing the Backend.AI Manager and AppProxy changes required to run [Continuum Router](https://github.com/lablup/continuum-router) as a new kind of AppProxy worker. This is the Backend.AI-side half of the design; the worker-side half is the design document in lablup/continuum-router#748.

### What it proposes

- **`FrontendMode.ROUTER`** — a worker holds a single binding address for its sole endpoint; all slot management (subdomain/port allocation) is skipped; circuits are created with `port = None` and `subdomain = None`; `get_endpoint_url()` returns the worker's one advertised base URL. The request's `model` name, gated by an API key, becomes the addressing key.
- **Key–model–service mappings managed by the Manager** — a published model maps to one or more deployment endpoints (A/B testing, cross-resource-group balancing) with split ratios; router API keys (the `EndpointTokenRow` pattern lifted from endpoint scope to model scope) carry the allowed-model set. Users receive (base URL, model name, API key) instead of a per-deployment URL + endpoint JWT.
- **Coordinator desired-state persistence + propagation reusing the existing protocol** — `router_models`/`router_api_keys` tables with a per-authority revision, manager-scope `/v2/routers/{authority}/…` APIs, a worker-scope `router-config` snapshot, and five new Pub/Sub events following the existing full-state-per-entity/first-ack conventions (`initialize_legacy_circuit` pattern, set-diff reconcile as in `reconcile_traefik_etcd_state`). Mutations are proactive-deploy: persist → broadcast → wait ≤15 s for the first worker ack, succeeding even on timeout. Key material never transits Redis (key events are id-only notifications).

### Compatibility

All changes are additive; stock WILDCARD/PORT workers, existing circuits, endpoint tokens, and the Traefik path are untouched. A ROUTER worker cannot register against a coordinator predating the proposal (intended version gate).

### Notes

- JIRA reference in the BEP header is TBD.


achimnol:
Merging the first version. I'll add a follow-up PR.

PR #12128, opened 2026-06-10
feat(BA-6349): add Permission bitmask column to permissions table with backfill

fregataa:
## Summary
- Add a `Permission` IntFlag bitmask column to the `permissions` table, coexisting with the legacy per-operation `operation` StrEnum column (grant-side analog of the cap column added to `scope_entities`).
- Introduce `Permission.from_operation()`: CRUD operations map to their bit; `GRANT_*` operations have no dedicated bit and map to `NONE` (grant authority stays on the `operation` column during the transition).
- Wire the new field through `PermissionRow` / `PermissionData` / `PermissionCreator` and every write path (`PermissionCreatorSpec`, role manager, RBAC granter, vfolder owner-grant insert).
- Alembic migration adds the column, backfills each row's single `operation` into the corresponding bit, then sets `NOT NULL`; downgrade drops the column.

Switching permission resolution to the new column, API exposure, and row de-duplication / `operation` removal are separate follow-ups.

## Test plan
- [x] `pants fmt / fix / lint / check` (mypy) pass for affected packages
- [x] Affected permission_controller / base.rbac / service unit suites pass; updated 25 inline `PermissionRow`/`PermissionData` test constructions
- [x] Migration verified against local DB (2,427 rows): backfill maps create→4, read→1, update→2, soft-delete→8, hard-delete→16, `grant:*`→0; no NULLs; column is `smallint NOT NULL`
- [x] `downgrade` drops the column cleanly and re-`upgrade` succeeds

Resolves BA-6349

### DIFF
--- a/src/ai/backend/manager/models/alembic/versions/c6648c039bd4_add_permission_bitmask_column_to_.py
+++ b/src/ai/backend/manager/models/alembic/versions/c6648c039bd4_add_permission_bitmask_column_to_.py
@@ -1,7 +1,7 @@
 """add permission bitmask column to permissions
 
 Revision ID: c6648c039bd4
-Revises: a3f1c7e2b9d4
+Revises: e7b3a1f9c2d4
 Create Date: 2026-06-11 01:51:27.442017
 
 """
@@ -11,8 +11,8 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = "c6648c039bd4"
-down_revision = "a3f1c7e2b9d4"
-# Part of: 26.6.0
+down_revision = "e7b3a1f9c2d4"
+
 branch_labels = None
 depends_on = None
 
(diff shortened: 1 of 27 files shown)


src/ai/backend/manager/models/alembic/versions/c6648c039bd4_add_permission_bitmask_column_to_.py
```diff
@@ -0,0 +1,44 @@
...
+""")
+
+
+def upgrade() -> None:
+    op.add_column(
+        "permissions",
+        sa.Column("permission", sa.SmallInteger(), nullable=True),
+    )
+    op.execute(_BACKFILL_PERMISSION_FROM_OPERATION)
+    op.alter_column("permissions", "permission", nullable=False)
```
HyeockJinKim, line 40: It looks like the `row` column remains unchanged—shouldn't the `UniqueConstraint` be modified?

fregataa: I plan to delete this when `permission` column is deleted

HyeockJinKim (approved):

PR #12137, opened 2026-06-11
feat(BA-6455): add single-entity action processor, validator, monitor framework

fregataa:
## Summary
- Build the per-type execution framework on the pure-ABC `BaseSingleEntityAction`, independent of the legacy `BaseAction` hierarchy (`src/ai/backend/manager/actions/single_entity/`).
- `SingleEntityActionProcessor` runs validators, then executes the action within a monitor lifecycle (`prepare` → func → `done`), capturing status/timing/error.
- Adds the type-specific ABCs `SingleEntityActionValidator` and `SingleEntityActionMonitor` (as packages) plus self-contained `SingleEntityActionResultMeta` / `SingleEntityActionProcessResult` (non-optional `entity_id`).

Resolves BA-6455

### DIFF
--- a/src/ai/backend/manager/actions/single_entity/processor.py
+++ b/src/ai/backend/manager/actions/single_entity/processor.py
@@ -1,0 +1,105 @@
+import logging
+import uuid
+from collections.abc import Awaitable, Callable, Sequence
+from datetime import UTC, datetime
+
+from ai.backend.common.exception import BackendAIError, ErrorCode
+from ai.backend.logging.utils import BraceStyleAdapter
+from ai.backend.manager.actions.action import BaseActionTriggerMeta
+from ai.backend.manager.actions.single_entity.base import BaseSingleEntityAction
+from ai.backend.manager.actions.single_entity.monitor import SingleEntityActionMonitor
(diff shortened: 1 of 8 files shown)


src/ai/backend/manager/actions/single_entity/processor.py
```diff
@@ -0,0 +1,105 @@
...
+)
+from ai.backend.manager.actions.single_entity.validator import SingleEntityActionValidator
+from ai.backend.manager.actions.types import OperationStatus
+
+__all__ = ("SingleEntityActionProcessor",)
+
+log = BraceStyleAdapter(logging.getLogger(__spec__.name))
+
+
+class SingleEntityActionProcessor[TAction: BaseSingleEntityAction, TResult]:
```
jopemachine, line 23: Are we deprecating the existing SingleEntityActionProcessor and using this new type instead?
In that case, it would be better to mark the existing type as deprecated in its documentation/comments.

fregataa: it is not fully interchangeable yet since not all validators are implemented in this new framework. So it is not good to add such comment.

HyeockJinKim (approved):

PR #12162, opened 2026-06-12
feat(BA-6472): make agent utilization metric interval configurable

fregataa:
## Summary
- Expose the utilization metric collection interval as agent config (`agent.utilization-metric.interval`), replacing the hardcoded `UTILIZATION_METRIC_INTERVAL` constant.
- Derive the Valkey expiration (TTL) of stored utilization values from the configured interval (`interval * 4`), so the TTL scales when the interval changes — applied uniformly to both agent-level and per-process metric stores.
- The three collection tasks and the compute-plugin measurement timeout now read the configured interval, injected via `local_config`.

## Test plan
- [x] Agent starts with default config; utilization metrics collect at the default interval and Valkey keys carry the derived TTL.
- [x] Changing `agent.utilization-metric.interval` changes both the collection cadence and the Valkey TTL.

Resolves BA-6472

### DIFF
--- a/src/ai/backend/agent/config/unified.py
+++ b/src/ai/backend/agent/config/unified.py
@@ -131,7 +131,7 @@ class UtilizationMetricConfig(BaseConfigSchema):
                 "Time interval in seconds between utilization metric collection cycles "
                 "(per-node, per-container, and per-process statistics)."
             ),
-            added_version="26.7.0",
+            added_version="26.4.4",
             example=ConfigExample(local="30.0", prod="30.0"),
         ),
     ]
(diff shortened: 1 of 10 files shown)


src/ai/backend/agent/config/unified.py
```diff
@@ -1192,6 +1207,22 @@ class OverridableAgentConfig(BaseConfigSchema):
...
+            default_factory=UtilizationMetricConfig,
+            validation_alias=AliasChoices("utilization-metric", "utilization_metric"),
+            serialization_alias="utilization-metric",
+        ),
+        BackendAIConfigMeta(
+            description=(
+                "Configuration for utilization metric collection. Controls the collection "
+                "interval and, derived from it, the Valkey expiration of stored utilization values."
+            ),
+            added_version="26.4.4",
```
HyeockJinKim, line 1222: The version doesn't match the value above; please update it.

HyeockJinKim (approved):

PR #12168, opened 2026-06-12
fix(BA-6454): propagate session commit bgtask failures to clients

jopemachine:
## Summary
- Raise domain exceptions (`SessionNotFound`, `ContainerRegistryNotFound`, `BgtaskFailedError`) from `CommitSessionHandler.execute()` instead of returning a success-typed result carrying an `error_message`, so the bgtask framework emits `bgtask_failed` to clients instead of `bgtask_done`. Previously the WebUI showed image commit as successful even when it had failed.
- Resolve the base image with `alive_only=False` inside the background task, matching the service-layer pre-validation, so committing a session whose base image was deleted no longer fails with `ImageNotFound` mid-task.
- Record the actual task result status and message in the task store (`finish_subtask`): failed/cancelled results are now recorded as `FAILURE` with the failure reason instead of being unconditionally marked `SUCCESS`. Added `BgtaskStatus.to_task_status()` and `TaskResult.last_message()` to support this.

## Test plan
- [ ] Existing bgtask unit tests pass (`tests/unit/manager/bgtask`, `tests/unit/common/bgtask`)
- [ ] Live verification: commit a session whose base (customized) image row is deleted → task succeeds
- [ ] Live verification: force a commit failure → client SSE stream receives `bgtask_failed` and WebUI shows the failure

Resolves BA-6454 ([Jira](https://lablup.atlassian.net/browse/BA-6454))

### DIFF
--- a/src/ai/backend/manager/bgtask/tasks/commit_session.py
+++ b/src/ai/backend/manager/bgtask/tasks/commit_session.py
@@ -30,6 +30,8 @@ from ai.backend.common.types import AgentId, ImageRegistry, SessionId
 from ai.backend.logging import BraceStyleAdapter
 from ai.backend.manager.bgtask.types import ManagerBgtaskName
 from ai.backend.manager.data.image.types import ImageIdentifier
+from ai.backend.manager.errors.image import ContainerRegistryNotFound
+from ai.backend.manager.errors.kernel import SessionNotFound
 
 if TYPE_CHECKING:
     from ai.backend.common.events.fetcher import EventFetcher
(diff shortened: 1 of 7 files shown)


src/ai/backend/manager/bgtask/tasks/commit_session.py
```diff
@@ -254,9 +254,9 @@ async def execute(self, manifest: CommitSessionManifest) -> CommitSessionResult:
             log.info("Session commit completed successfully. Image ID: {}", result_image_id)
             return CommitSessionResult(image_id=result_image_id)
 
-        except Exception as e:
+        except Exception:
             log.exception("Failed to commit session {}", manifest.session_id)
-            return CommitSessionResult(error_message=str(e))
+            raise
```
HyeockJinKim, line 259: It seems like the message is different from before. Does this actually match the current assumption regarding the BGTask Manager?

HyeockJinKim (approved):

PR #12191, opened 2026-06-15
fix(BA-6489): translate legacy inference env vars into model-service CLI args

seedspirit:
## Summary
- The runtime-variant baseline `start_command` (e.g. `vllm serve {model_path}`) overrides the image's `start-*.sh` wrapper that used to read legacy env vars (`VLLM_EXTRA_ARGS`, `VLLM_TP_SIZE`, ...) and splice them into the server command line, so those args silently stopped applying.
- New `LegacyInferenceEnvArgs.to_cli_args` reproduces that env→argv step on the agent: maps known `VLLM_*` / `SGLANG_*` keys to flags and `shlex`-splits the raw `*_EXTRA_ARGS` string.
- The agent appends the result to each model service's `start_command` right after the image-cmd fallback.

## Test plan
- [x] `pants test tests/unit/agent/test_legacy_inference_env.py` (to_cli_args + append behavior)
- [x] Live e2e on local dev stack: deployed a model service with `VLLM_EXTRA_ARGS`/`VLLM_TP_SIZE`; `docker exec ps` showed the args appended to the container's model-service argv and `/health` returned 200.

## Known limitation (pending design decision)
- Args whose value contains literal `{}` (e.g. vLLM `--hf-overrides '{...}'` JSON) currently break in the krunner's `ServiceArgumentInterpolator`, which treats every `{...}` token as a `{var}` template (`format_map`) and raises `unmatched '{' in format spec`. Quoting does not help — `start_command` is a template string and literal JSON collides with it. Handling for brace-containing args is being decided separately (tolerant interpolator vs. out-of-band literal-arg delivery).

Resolves BA-6489

### DIFF
--- a/changes/12191.fix.md
+++ b/changes/12191.fix.md
@@ -1,0 +1,1 @@
+Restore legacy inference env vars (`VLLM_EXTRA_ARGS`, `VLLM_TP_SIZE`, ...) by translating them into model-service CLI args on the agent
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -67,7 +67,7 @@ from ai.backend.agent.errors import (
 )
 from ai.backend.agent.etcd import AgentEtcdClientView
 from ai.backend.agent.health.heartbeat import HeartbeatTask
-from ai.backend.agent.legacy_inference_env import LegacyInferenceEnvArgs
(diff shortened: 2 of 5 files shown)


HyeockJinKim (approved):
The code isn't exactly clean, but since it doesn't seem to be a major issue, I'll go ahead and merge it.

PR #12193, opened 2026-06-15
fix(BA-6490): preserve baseline service port when adding a revision with health check

jopemachine:
## Summary

Resolves [BA-6490](https://lablup.atlassian.net/browse/BA-6490).

Adding a model revision with a health check enabled failed with a spurious validation error:

```
Schema validation failed. (1 validation error for ModelServiceConfig
port
  Field required [type=missing, input_value={'health_check': ModelHea...200, initial_delay=1.0)}, input_type=dict])
```

## Root cause

The model-definition draft merge (`_pick` in `common/config.py`) decided overrides purely from `model_fields_set`. The GraphQL layer materializes an *omitted* optional input field (e.g. `port`) as an explicit `None` and puts it in `model_fields_set`. So when a client sent a partial `service` block (`{ health_check: { enable: true, ... } }`) without a `port`, the merge overrode the runtime-variant baseline port (e.g. vllm `8000`) with `None`, and `to_resolved()` then failed strict `port` validation.

When the health check was disabled the client omitted the whole `service` block, so the baseline port survived — which is why the error only appeared with a health check enabled.

## Fix

Change `_pick` so a field overrides the base **only when it is both set and non-`None`**. An unset field, or one explicitly set to `None`, now means "no change" and keeps the lower-priority value.

This matches the non-`None` override semantics already used by the `RevisionDraft` scalar merge and the `model_path` inline merge, and it covers explicit-null inputs from any client (GraphQL and REST v2) — rather than relying on a GQL-layer `UNSET` default. The `health_check: null → disabled` normalization keeps working because `from_file_payload` produces a concrete (non-`None`) disabled draft, not a bare `None`.

## Tests

- `tests/unit/common/test_config.py`: `test_explicit_none_port_does_not_clobber_baseline` — an override draft with `port=None` keeps the base port through merge + `to_resolved()`.
- `tests/unit/manager/api/gql/deployment/test_revision_input_conversion.py`: full GQL input → draft → merge path; health check enabled with no port preserves the baseline port, explicit port still overrides.

## Verification

Reproduced and verified end-to-end against a local manager via `./bai gql --v2 addModelRevision` (forcing the vllm variant, whose baseline ships `port=8000`):
- before: `ModelServiceConfig.port field required` (`backendai_parsing_invalid-parameters`);
- after: the port error is gone and the model definition resolves with the baseline port preserved.

----
📚 Documentation preview 📚: https://sorna--12193.org.readthedocs.build/en/12193/

[BA-6490]: https://lablup.atlassian.net/browse/BA-6490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

----
📚 Documentation preview 📚: https://sorna-ko--12193.org.readthedocs.build/ko/12193/


docs/manager/graphql-reference/supergraph.graphql
```diff
@@ -9484,16 +9484,16 @@ input ModelConfigInput
...
+  modelPath: String
 
   """Configuration for the model service."""
-  service: ModelServiceConfigInput = null
+  service: ModelServiceConfigInput
 
   """Metadata about the model."""
-  metadata: ModelMetadataInput = null
+  metadata: ModelMetadataInput
 }
```
HyeockJinKim, line 9497: Why did you remove the "optional" part?

PR #12194, opened 2026-06-15
fix(BA-6479): apply scope-level auto-assign roles on user creation

fregataa:
## Summary
- A newly created user assigned to a project/domain at creation time was not granted that scope's `auto_assign` roles — only its own user-scope preset roles were provisioned (project/domain auto-assign worked only on the existing-user join/update path).
- Add `RBACWriteOps.add_users_to_scope()` bundling the user↔scope binding (`association_scopes_entities`, idempotent) with `auto_assign` role mapping.
- Grant the new user's initial domain/project auto-assign roles from the `create_user` service flow via `UserRepository.assign_initial_scope_roles()`, plus a dedicated `assign_user_to_model_store()` path for the domain's model-store project membership and role mapping.

## Test plan
- [x] `pants test tests/component/user/test_user_crud.py` — new `TestUserCreateAutoAssignRoles` verifies project- and model-store-scope auto_assign role mappings end-to-end (real server + DB)
- [x] `pants test tests/unit/manager/services/::` — service-level create_user delegation
- [x] Manual: create a user into a project/domain that has an `auto_assign=true` role; confirm `user_roles` mapping is created

Resolves BA-6479

### DIFF
--- a/src/ai/backend/manager/repositories/ops/rbac/provider.py
+++ b/src/ai/backend/manager/repositories/ops/rbac/provider.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import uuid
 from collections.abc import AsyncIterator, Collection, Sequence
 from contextlib import asynccontextmanager
 
@@ -10,6 +9,7 @@ from sqlalchemy.dialects.postgresql import insert as pg_insert
 from sqlalchemy.ext.asyncio import AsyncSession as SASession
 
 from ai.backend.common.data.permission.types import RelationType
+from ai.backend.common.identifier.user import UserID
 from ai.backend.manager.data.permission.id import ScopeId
 from ai.backend.manager.data.permission.types import EntityType
 from ai.backend.manager.models.base import Base
@@ -44,7 +44,7 @@ class RBACWriteOps(WriteOps):
     async def add_users_to_scope(
         self,
         scope_id: ScopeId,
-        user_ids: Collection[uuid.UUID],
+        user_ids: Collection[UserID],
     ) -> None:
         """Bind users to a scope and grant the scope's ``auto_assign`` roles.
 
--- a/src/ai/backend/manager/repositories/ops/rbac/provider.py
+++ b/src/ai/backend/manager/repositories/ops/rbac/provider.py
@@ -2,20 +2,37 @@
 
 from __future__ import annotations
 
-from collections.abc import AsyncIterator, Sequence
+import uuid
+from collections.abc import AsyncIterator, Collection, Sequence
 from contextlib import asynccontextmanager
 
+from sqlalchemy.dialects.postgresql import insert as pg_insert
+from sqlalchemy.ext.asyncio import AsyncSession as SASession
+
+from ai.backend.common.data.permission.types import RelationType
+from ai.backend.manager.data.permission.id import ScopeId
+from ai.backend.manager.data.permission.types import EntityType
 from ai.backend.manager.models.base import Base
+from ai.backend.manager.models.rbac_models.association_scopes_entities import (
+    AssociationScopesEntitiesRow,
+)
 from ai.backend.manager.repositories.base.rbac.entity_creator import (
     RBACBulkEntityCreatorResult,
     RBACEntityCreator,
     execute_rbac_entity_creators,
 )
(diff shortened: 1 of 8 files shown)


src/ai/backend/manager/repositories/user/repository.py
```diff
@@ -106,6 +106,18 @@ async def assign_project_membership(self, user_uuid: UUID, project_id: ProjectID
         """Add a user to a project, mapping the user to the project's member role."""
         await self._db_source.assign_project_membership(user_uuid, project_id)
 
+    @user_repository_resilience.apply()
+    async def assign_users_to_scope(
+        self, user_uuid: UUID, domain_name: str | None, project_ids: Collection[ProjectID]
```
jopemachine: Why not using UserID?

src/ai/backend/manager/repositories/ops/rbac/provider.py
```diff
@@ -2,28 +2,75 @@
...
+            {
+                "scope_type": scope_id.scope_type,
+                "scope_id": scope_id.scope_id,
+                "entity_type": EntityType.USER,
+                "entity_id": str(user_id),
+                "relation_type": RelationType.AUTO,
+                "permission_cap": None,
+            }
+            for user_id in user_ids
+        ]
```
HyeockJinKim, line 69: It seems better to define it as a method that takes a scope and entities.

src/ai/backend/manager/repositories/user/repository.py
fregataa: what UserID?

src/ai/backend/manager/repositories/user/db_source/db_source.py
```diff
@@ -368,6 +370,62 @@ async def _create_single_user_with_keypair_and_groups(
 
         return UserCreateResultData(created_user, kp_data)
 
+    async def assign_users_to_scope(
+        self,
+        user_uuid: UUID,
+        domain_name: str | None,
+        project_ids: Collection[ProjectID],
```
seedspirit, line 378: Is there a reason why domains aren't accepted as a collection? I'm curious why only projects are accepted as a collection, while domains are accepted as a single item or none at all.

fregataa: we don't have any requirement to assign single user to multiple domains yet. Plus I plan to implement new user creation API and deprecate this

src/ai/backend/manager/repositories/user/db_source/db_source.py
```diff
@@ -368,6 +371,62 @@ async def _create_single_user_with_keypair_and_groups(
 
         return UserCreateResultData(created_user, kp_data)
 
+    async def assign_users_to_scope(
+        self,
+        user_uuid: UserID,
+        domain_name: str | None,
```
jopemachine, line 377: Is there a reason this takes domain_name as an argument, though domain_id had already been introduced?

src/ai/backend/manager/repositories/user/repository.py
fregataa: updated this to use UserID

src/ai/backend/manager/repositories/user/db_source/db_source.py
fregataa: API still gets a domain name only and I don't want to introduce a domain id resolver here, which is out of fix task

src/ai/backend/manager/repositories/ops/rbac/provider.py
```diff
@@ -2,28 +2,75 @@
...
 
 
 class RBACWriteOps(WriteOps):
     """Base write ops plus RBAC scope-associated creation."""
 
+    _role_manager: RoleManager
+
+    def __init__(self, sess: SASession) -> None:
+        super().__init__(sess)
+        self._role_manager = RoleManager()
```
jopemachine, line 35: It feels awkward that an object named RoleManager is instantiated inside the constructor when RBACWriteOps is created.

fregataa: it is a naming mistake

jopemachine (approved):

PR #12203, opened 2026-06-16
fix(BA-6487): prune dead-PID process metrics each collection cycle

fregataa:
## Summary
  - Prune per-process utilization metrics for PIDs no longer alive on each collection cycle, routed through lazy_remove_process_metric (the same cleanup path used at container destroy), so the in-memory map / per-container Redis blob / exported Prometheus series
  stay bounded instead of growing for the container's lifetime. A container with no live-process listing in a cycle has all its tracked PIDs pruned.
  - Add an config flag (default true) to turn off node/container/per-process metric collection entirely.
  - Add Prometheus duration histograms for stat collection: `backendai_stage_duration_sec` (per-segment: docker_top/gather/serialize/redis_write) and `backendai_stat_task_duration_sec` (whole-cycle, by scope).

## Test plan
  - [x] Per-process Prometheus series for dead PIDs are removed after UTILIZATION_METRIC_DETENTION.
  - [x] Disabling enable_process_metric skips per-process collection while node/container metrics keep flowing.

Resolves BA-6487

### DIFF
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -1349,6 +1349,9 @@ class AbstractAgent[
 
     @_observe_stat_task(stat_scope=StatScope.NODE)
     async def collect_node_stat(self, resource_scaling_factors: Mapping[SlotName, Decimal]) -> None:
+        if not self.local_config.agent.utilization_metric.node.enable:
+            log.debug("skipping node stat collection as it's disabled in config")
+            return
         if self.local_config.debug.log_stats:
             log.debug("collecting node statistics")
         async with asyncio.timeout(STAT_COLLECTION_TIMEOUT):
@@ -1356,6 +1359,9 @@ class AbstractAgent[
 
     @_observe_stat_task(stat_scope=StatScope.CONTAINER)
     async def collect_container_stat(self) -> None:
+        if not self.local_config.agent.utilization_metric.container.enable:
+            log.debug("skipping container stat collection as it's disabled in config")
+            return
         if self.local_config.debug.log_stats:
             log.debug("collecting container statistics")
         container_ids: set[ContainerId] = set()
@@ -1368,7 +1374,7 @@ class AbstractAgent[
 
     @_observe_stat_task(stat_scope=StatScope.PROCESS)
     async def collect_process_stat(self) -> None:
-        if not self.local_config.agent.utilization_metric.enable_process_metric:
+        if not self.local_config.agent.utilization_metric.process.enable:
             log.debug("skipping process stat collection as it's disabled in config")
             return
         if self.local_config.debug.log_stats:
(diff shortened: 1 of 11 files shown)


jopemachine:
PR description said "Containers with no live-process listing this cycle are left untouched to avoid pruning on a transient empty/failed listing", but I don't see such check.

src/ai/backend/agent/stats.py
```diff
@@ -342,6 +343,20 @@ def to_serializable_dict(self) -> MetricValue:
...
+async def _packb_many_in_executor(
+    loop: asyncio.AbstractEventLoop,
+    mapping: dict[str, Any],
+) -> dict[str, bytes]:
+    """
+    msgpack-pack each value of ``mapping`` off the event loop in a single hop.
+    """
+
+    def _pack(m: dict[str, Any]) -> dict[str, bytes]:
+        return {key: msgpack.packb(value) for key, value in m.items()}
```
HyeockJinKim: I'm not sure if the msgpack operation takes long enough to warrant splitting it into separate threads.

src/ai/backend/agent/config/unified.py
```diff
@@ -135,6 +135,24 @@ class UtilizationMetricConfig(BaseConfigSchema):
...
+            description=(
+                "Whether to collect per-process utilization metrics for each container. "
+                "When disabled, the agent skips listing container processes and reporting "
+                "per-process statistics, while per-node and per-container metrics are "
+                "still collected."
+            ),
+            added_version="26.4.4",
+            example=ConfigExample(local="true", prod="true"),
+        ),
+    ]
```
HyeockJinKim: Rather than just separating the collection of process metrics, I think it would be better to set the "enable" value for each metric individually.

src/ai/backend/agent/agent.py
```diff
@@ -1349,13 +1357,19 @@ async def collect_logs(
...
         if self.local_config.debug.log_stats:
             log.debug("collecting node statistics")
         async with asyncio.timeout(STAT_COLLECTION_TIMEOUT):
             await self.stat_ctx.collect_node_stat(resource_scaling_factors)
 
     @_observe_stat_task(stat_scope=StatScope.CONTAINER)
     async def collect_container_stat(self) -> None:
+        if not self.local_config.agent.utilization_metric.container.enable:
+            log.debug("skipping container stat collection as it's disabled in config")
+            return
```
seedspirit, line 1372: Since these metrics are tied to features like IdleChecker and deployment auto-scaling, I'm not sure if it would be a good idea to provide an option to disable their collection.

fregataa: this is enabled by default. I think It is good for debugging

src/ai/backend/common/metrics/metric.py
```diff
@@ -778,23 +781,76 @@ def to_prometheus_singleprocess(self) -> str:
...
+            ):
+                await do_docker_call()
+        """
+        start = time.perf_counter()
+        try:
+            yield
+        finally:
+            self._stage_duration_sec.labels(
+                stage=stage.value, upper_layer=upper_layer.value
+            ).observe(time.perf_counter() - start)
```
HyeockJinKim: We'll collect the "count" here as well and use "success" as the label.

HyeockJinKim (approved):

PR #12206, opened 2026-06-16
fix(BA-6506): parametrize app/storage base domains

### LINKED ISSUE #12205: Remove hardcoded internal hostnames from the installer's wildcard-domain path
## Motivation

- The installer built service hostnames from a fixed prefix plus Lablup-internal domain suffixes, so internal infrastructure names were committed in a public repository.
- Those baked-in suffixes tied the wildcard-domain install path to one specific deployment, so anyone else installing with wildcard domains got the wrong hostnames.
- A help example also exposed an internal collector IP, and the installer carried dead code (an unused passphrase helper, a duplicate shell lookup, an empty module) that added noise and maintenance cost.

## Objective

-   Let operators supply the app and storage base domains explicitly so the installer no longer embeds internal hostnames, while leaving IP-based installs unchanged.

## Details

1. Add two installer options for the app base domain and the storage base domain, used in place of the previously hardcoded internal suffixes.
1. When the new options are not set, fall back to the public-facing address so IP-based installs behave exactly as before.
1. Remove the internal collector IP from the help example.
1. Delete the dead installer code (unused passphrase helper, duplicate shell lookup, empty module).

## Impact

- No internal infrastructure hostnames or IPs remain in the public repo's installer.
- Wildcard-domain installs work for any deployment, not just the original one, via explicit domain options.
- IP-based installs are unaffected because the new options default to the existing public-facing address behavior.
- Less installer code to maintain after removing the dead paths.

JIRA Issue: BA-6506

devyubin:
resolves #12205 (BA-6506)

Replaces the hardcoded Lablup-internal hostnames (`<prefix>.app.backend.ai`,`<prefix>.public.isla-sorna.backend.ai`) that the installer built from `--fqdn-prefix` with two injected options, `--app-base-domain` and `--storage-base-domain`. Those strings put internal infrastructure names in a public repo and tied the wildcard-domain install path to one deployment. With the new options unset the hostnames fall back to `--public-facing-address`, so IP-based installs are untouched. The branch also drops an internal collector IP from a help example and removes dead installer code: an unused passphrase
helper, a duplicate SHELL lookup, and an empty module.

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/install/context.py
```diff
@@ -1359,7 +1349,6 @@ async def configure_client(self) -> None:
         with self.resource_path(
             "ai.backend.install.fixtures", "example-keypairs.json"
         ) as keypair_path:
-            current_shell = os.environ.get("SHELL", "sh")
```
seedspirit, line 1465: Did you delete it because it was dead code?

seedspirit (commented):
It looks fine overall, @Yaminyam  could you please take a look?

src/ai/backend/install/context.py
devyubin: Yes`current_shell` was assigned twice with the same `os.environ.get("SHELL", "sh")`, once in the keypairs block and once in the users block, but only the users block reads it (`if current_shell == "fish"`). The keypairs-block copy was never used before that second assignment, so removing it changes nothing.

Yaminyam (dismissed):

seedspirit (approved):

PR #12211, opened 2026-06-16
feat(BA-6499): resolve model folder mount permission per deploy entry point

HyeockJinKim:
## Summary
- Freeze the **model vfolder mount permission** on each deployment revision at write time instead of hard-coding `READ_ONLY` at draft-build time.
- `vfolder deploy` / `model-card deploy` mount the model **read-only**; `deployment create` / `revision add` adopt the requester's **own effective permission** on the model vfolder when unspecified, or a caller-supplied permission capped by it (**rejected fail-fast** when it exceeds).
- Adds a nullable `model_mount_perm` column to `deployment_revisions` (NULL → `READ_ONLY` fallback for legacy rows), threaded through the creator/draft/merge pipeline. Extra-mount permission handling is unchanged.

## Test plan
- [x] Unit: `MountPermission.exceeds`, controller `_ground_model_mount_perm` (inherit / forced-RO / cap / fail-fast), draft builder uses stored perm with RO fallback, refresh preserves `model_mount_perm`.
- [x] Updated all affected construction sites (data types, creator spec, row).
- [x] Live (`./bai`, local manager): deployment create w/o perm on an RW-owned vfolder → `model_mount_perm=rw`; with `ro` → `ro`; with `wd` (exceeds rw) → `VFolderPermissionError`. Verified the persisted column in DB.
- [ ] CI green.

> Note: the pre-existing `test_generate_token` failure is an environmental `aioresponses`/`aiohttp` incompatibility unrelated to this change (no HTTP-client or dependency changes in this diff).

Resolves BA-6499


src/ai/backend/manager/api/rest/service/handler.py
```diff
@@ -714,6 +715,7 @@ async def _to_model_revision(
                 model_vfolder_id=validation_result.model_vfolder_id,
                 model_definition_path=validation_result.model_definition_path,
                 model_mount_destination=params.config.model_mount_destination,
+                model_mount_perm=MountPermission.READ_ONLY,
```
seedspirit, line 718: Is this the code corresponding to the "NULL → READ_ONLY fallback for legacy rows" mentioned in the PR body?

src/ai/backend/manager/data/deployment/creator.py
```diff
@@ -65,15 +70,23 @@ class ModelRevisionCreator:
     execution: ExecutionSpec | None = None
     model_definition: ModelDefinitionDraft | None = None
     revision_preset_id: DeploymentPresetID | None = None
-    runtime_variant_preset_values: list[RuntimeVariantPresetValueData] = field(default_factory=list)
+    preset_values: list[PresetValueData] = field(default_factory=list)
 
-    def to_draft_with_extra_mount(self, extra_mounts: list[MountInfoEntry]) -> RevisionDraft:
+    def to_draft_with_extra_mount(
```
seedspirit, line 75: Since this no longer involves only operations related to extra mounts, I think we should rename the method.

src/ai/backend/manager/data/deployment/types.py
```diff
@@ -787,6 +793,9 @@ def _merge_mounts(
         else lower.model_definition_path,
         model_mount_destination=upper.model_mount_destination,
         extra_mounts=list(upper.extra_mounts),
+        model_mount_perm=upper.model_mount_perm
+        if upper.model_mount_perm is not None
+        else lower.model_mount_perm,
```
seedspirit, line 799: If "None" is set in the user request, how does the merge ultimately work? It seems like it would be ignored if set to "None," so in order for it not to be ignored, it would need to be resolved to the requester's own effective permissions first.

seedspirit: It looks like the deployment controller is already resolving this. I confirmed that even if "none" is set in the user request, a non-null value is passed during the merge.

fregataa (dismissed):

PR #12241, opened 2026-06-16
docs(BA-6523): rewrite BEP-1052 for the rank-based implementation

jopemachine:
Rewrites BEP-1052 to match the current implementation.

- Drops the **AppConfigPolicy / scope_sources** design entirely — merge is now driven by a per-`name` **`rank`** column (next-value on insert, modeled on `DeploymentRevisionPreset`); fragment participation is decided by scope matching, not a policy allow-list.
- **Removes all code blocks** (Python / GraphQL SDL / JSON) — keeps ideas only, concisely (2056 → 294 lines).
- Updates the data model, merge semantics, GraphQL/REST exposure, caching, and user scenarios to the rank-based, policy-free design.

Jira: BA-6523

### DIFF
--- a/proposals/BEP-1052-scoped-app-config-redesign.md
+++ b/proposals/BEP-1052-scoped-app-config-redesign.md
@@ -203,17 +203,7 @@ defaults. Each raw fragment's `config` follows the same rule.
 
 ---
 
-## 4. Caching
-
-The merged-view read path is fronted by a Valkey cache keyed per
-`(user, name)`. Writes invalidate the affected scope so stale merges are
-not served; cache failures fall through to the database transparently
-(never break a request). The cache stores the merged `config`; raw
-per-scope reads are uncached.
-
----
-
-## 5. Client integration (WebUI)
+## 4. Client integration (WebUI)
 
 - **Before login**, the WebUI fetches `public` documents (theme,
   branding) anonymously so the shell can render.
@@ -225,7 +215,7 @@ per-scope reads are uncached.
 
 ---
 
-## 6. User scenarios
+## 5. User scenarios
 
 - **Pre-login public config** — anonymous read of `public` `theme`.
 - **Bootstrap after login** — read merged `AppConfig`s in one round of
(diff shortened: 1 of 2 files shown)


proposals/BEP-1052-scoped-app-config-redesign.md
```diff
@@ -11,2045 +11,199 @@ Implemented-Version:
...
+  wins). Assigned on create (see §2).
+- `config` — schema-less JSON payload.
+- `created_at` / `updated_at`.
+
+### Scope-ID convention
+
+| `scope_type` | `scope_id`              | Meaning of `config`              |
+|--------------|-------------------------|----------------------------------|
+| `public`     | literal `"public"`      | pre-login value of the document  |
+| `domain`     | `domain_id`             | the domain's value / default     |
```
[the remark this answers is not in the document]
jopemachine: `domains` has both a `name` PK and a separate unique `id` (GUID) column — see `models/domain/row.py:99`. This table intentionally uses the domain `id` (a stable, rename-safe UUID), not the name. The premise that domains are name-only is wrong, so there is no mismatch. Resolving.

proposals/BEP-1052-scoped-app-config-redesign.md
```diff
@@ -11,2045 +11,199 @@ Implemented-Version:
...
-Write mutations split into two paths with distinct rules. All
-bulk-only.
+### Read variants
 
-**Admin path** — `adminBulkCreateAppConfigFragments`,
-`adminBulkUpdateAppConfigFragments`. Admin regardless of each item's
-`key.scopeType`:
-
-| Operation                                  | Anonymous | User | Admin |
-|--------------------------------------------|-----------|------|-------|
```
fregataa, line 1070: Is admin queries excluded here? They all seem to be resolved app config queriers

fregataa (approved):
It seems good to announce that 404 can be returned if no config found

jopemachine:
> It seems good to announce that 404 can be returned if no config found

I think it belongs to API implementation detail. Let's keep only the core idea of the BEP

proposals/BEP-1052-scoped-app-config-redesign.md
jopemachine: admin is not scope type.

PR #12242, opened 2026-06-17
refactor(BA-6524): clean up the TUI installer internals

### LINKED ISSUE #12243: Clean up duplicated TUI installer context code and fix the non-interactive PACKAGE path
## Motivation

- The DEVELOP and PACKAGE install contexts duplicated their setup logic almost line for line and had drifted apart. Values present in DEVELOP but missing in PACKAGE (such as the app-proxy secret and address) caused real PACKAGE install failures, and the same kind of bug kept coming back.
- One context module carried `# type: ignore` suppressions, which breaks the repo's rule against suppressing type checks.
- Dead code was still in the tree: a no-op module and an unused binary-fetch function, along with their imports and call sites.
- Non-interactive `--mode PACKAGE` stalled at the package-type submenu and never started the install, while DEVELOP started right away.
- The command runner swallowed `CancelledError`, so an install aborted with Ctrl-C continued silently into the next step.

## Objective

- Remove the structural duplication and dead code in the installer, restore type checking in the context layer, and fix the non-interactive PACKAGE start and the cancel handling so an aborted install actually stops.

## Details

1. Merge the two duplicated context setup methods into one shared builder, with each context passing only its real differences (install type, base path, local-proxy port, loopback alias, optional harbor/sftp).
1. Pass the optional harbor/sftp overrides through a typed structure instead of an untyped dict, so wrong field names or types are caught statically.
1. Move the duplicated mock-accelerator setup into the base context.
1. Remove the dead module and the unused function, including their imports and calls.
1. Replace the type suppressions in the context module by typing the loaded config document explicitly, leaving only the one unrelated suppression that has no real check to restore.
1. Make non-interactive PACKAGE auto-select its single release option and start the install, matching DEVELOP.
1. Make the command runner reap the child process (terminate, bounded wait, then kill) and re-raise on cancellation, since cancellation here is always a global abort and nothing wraps these calls in a per-call timeout.

## Impact

- Removes the drift that was the root cause of the recurring PACKAGE install bugs, rather than patching each symptom.
- Non-interactive PACKAGE installs, the real path for setting up a fresh server, now start and proceed.
- An aborted install stops instead of running on into the next step.
- Restores static type checking in the context layer and deletes dead code.

JIRA Issue: BA-6524

devyubin:
resolves #12243 (BA-6524)

Clean up the single-host TUI installer internals: merge the duplicated `DevContext`/`PackageContext` install-info builders into a single shared helper, remove the dead `python.py`/`pkg.py` modules, and drop 59 of 60 `# type: ignore` comments in `context.py` by typing the loaded TOML documents. Along the way, fix `--non-interactive --mode PACKAGE` never starting the install and `run_exec` swallowing `CancelledError` (an aborted install no longer continues silently).

- [ ] Test case(s) to:
- [ ] Documentation


src/ai/backend/install/context.py
```diff
@@ -1105,34 +1220,34 @@ async def configure_appproxy(self) -> None:
...
-            data = tomlkit.load(fp)
-            data["db"]["type"] = "postgresql"  # type: ignore[index]
-            data["db"]["name"] = "appproxy"  # type: ignore[index]
-            data["db"]["user"] = "appproxy"  # type: ignore[index]
-            data["db"]["password"] = "develove"  # type: ignore[index]
-            data["db"]["pool_size"] = 8  # type: ignore[index]
-            data["db"]["max_overflow"] = 64  # type: ignore[index]
-            data["db"]["addr"]["host"] = halfstack.postgres_addr.face.host  # type: ignore[index]
-            data["db"]["addr"]["port"] = halfstack.postgres_addr.face.port  # type: ignore[index]
+            data: Any = tomlkit.load(fp)
```
seedspirit: Setting it to "Any" and removing "type: ignore" doesn't seem like a meaningful code change.

src/ai/backend/install/context.py
```diff
@@ -956,12 +1071,12 @@ async def configure_webserver(self) -> None:
         )
 
         with conf_path.open("r") as fp:
-            data = tomlkit.load(fp)
+            data: Any = tomlkit.load(fp)
```
seedspirit: Same here

src/ai/backend/install/context.py
```diff
@@ -81,6 +80,21 @@
 )
 
 
+class _ExtraServiceKwargs(TypedDict, total=False):
```
seedspirit: How about creating objects that wrap the `harbor` and `sftp_agent` options, respectively, and making them members?

changes/12242.enhance.md
```diff
@@ -0,0 +1 @@
+Clean up the single-host TUI installer internals: merge the duplicated `DevContext`/`PackageContext` install-info builders into a single shared helper, remove the dead `python.py`/`pkg.py` modules, and drop 59 of 60 `# type: ignore` comments in `context.py` by typing the loaded TOML documents. Along the way, fix `--non-interactive --mode PACKAGE` never starting the install and `run_exec` swallowing `CancelledError` (an aborted install no longer continues silently).
```
seedspirit: This is too long - please just give me the highlights.

src/ai/backend/install/context.py
```diff
@@ -114,9 +128,105 @@ def __init__(
...
+            Accelerator.CUDA_MIG_MOCK: "configs/accelerator/cuda-mock-mig.toml",
+            Accelerator.ROCM_MOCK: "configs/accelerator/rocm-mock.toml",
+        }
+
+        src = mapping.get(accelerator)
+        if not src:
+            return
+
+        dst = Path("mock-accelerator.toml")
+        print(f"[Installer] Copying accelerator config: {src} -> {dst}")
```
seedspirit: Is `print` enough? Are you not using `log` or anything else?

devyubin: The installer sends all output to `self.log` (Textual SetupLog), but this part was using `print` so it wasn't showing up in the TUI installer log window... I changed it to `self.log_header(...)`

seedspirit (approved):

PR #12246, opened 2026-06-17
doc(BA-6525): rewrite BEP-1052 for the three-table scoped app config model

jopemachine:
## Summary
- Rewrite BEP-1052 (Scoped App Config Redesign) around three tables: `app_config_definitions`, `app_config_fragments`, and `app_config_allow_list`.
- `app_config_definitions`: an explicitly admin-managed table of registered `config_name`s; no fragment may exist for an unregistered name.
- `app_config_fragments`: the per-scope JSON values keyed by `(scope_type, scope_id, config_name)`, each carrying a `rank`. Reads are the hot path — a single-table merge of the applicable fragments ordered by `rank`, with no join and no permission lookup.
- `app_config_allow_list`: one row per `(config_name, scope_type)` that gates every fragment write (admin and regular mutations alike). It carries no `rank` and is never consulted on read; admins alone manage the allow-list and `app_config_definitions`.
- Scopes are `public` / `domain` / `user`; overridability is decided by whether a `(config_name, user)` allow-list row exists.

Resolves BA-6525.

### DIFF
--- a/proposals/BEP-1052-scoped-app-config-redesign.md
+++ b/proposals/BEP-1052-scoped-app-config-redesign.md
@@ -49,8 +49,10 @@ Three scopes cover the use cases (`public` for the pre-login shell):
 - Where the admin has granted it, users persist their own settings on
   the server (language, recently used sessions, visible/ordered table
   columns, experimental-feature toggles).
-- The same scope may publish several independently-managed documents
-  loaded by different parts of the WebUI.
+- A single `config_name` can hold a different value at each scope: an
+  admin manages the `public` / `domain` value, and the user changes only
+  their own `user`-scope fragment. This one mechanism — per-scope
+  fragments combined by the merge — covers every use case above.
 
 ## Design Principles
 
--- a/proposals/BEP-1052-scoped-app-config-redesign.md
+++ b/proposals/BEP-1052-scoped-app-config-redesign.md
@@ -24,9 +24,11 @@ query with **no joins and no permission lookup**.
 
 **Write authorization is set up ahead of time.** Every document name is
 **explicitly registered** in `app_config_keys`, and `app_config_allow_list`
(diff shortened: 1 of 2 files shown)


proposals/BEP-1052-scoped-app-config-redesign.md
```diff
@@ -12,135 +12,205 @@ Implemented-Version:
...
+- **Explicitly registered names.** Every config name lives in
+  `app_config_definitions`; fragments and allow-list entries reference it by
+  foreign key. No fragment may exist for an unregistered `config_name`.
+- **Reads are join-free and unconditional.** The merge reads
+  `app_config_fragments` alone, ordering the existing fragments by
+  `rank`. No allow-list or policy is consulted at read time — the hot
+  path stays a single indexed scan.
+- **Allow-list = the write gate for every fragment.** `app_config_allow_list`
+  holds **one record per `(config_name, scope_type)`**; a fragment at
+  that scope may be created/updated/purged **only if** the record exists
```
fregataa, line 77: It seems good to allow purge fragments even tho the allow-list does not exist

fregataa (approved):
Is there any section about admin API and user API definition? not a detail spec.

proposals/BEP-1052-scoped-app-config-redesign.md
jopemachine: The allow-list is a record that determines whether a given scope has permission to mutate its app configuration.

If a scope is not allow-listed, it should not be able to delete fragments. Otherwise, a user could arbitrarily remove domain-level configurations that were set by an administrator.

jopemachine:
> Is there any section about admin API and user API definition? not a detail spec.

If I recall correctly, the previous discussion was that API specifications and data models should be kept out of the BEP.

Let's discuss the API specification separately in the PR I'll be requesting a review for shortly.

PR #12254, opened 2026-06-17
ci(BA-6526): add workflow to notify and close stale PRs

fregataa:
## Summary
- Add `actions/stale@v9` workflow that marks PRs stale after 14 days of inactivity and closes them after 14 more days, posting a comment at each step.
- Stale/close thresholds are surfaced as constants in a top-level `env` block; comment text references them automatically.
- Issues are excluded; only PRs are processed. Runs daily via cron and supports manual `workflow_dispatch`.

You can check specs about the stale action here (https://github.com/actions/stale#all-options).

Resolves BA-6526

### DIFF
--- a/.github/workflows/stale-pr.yml
+++ b/.github/workflows/stale-pr.yml
@@ -1,0 +1,38 @@
+name: 'Close Stale PRs'
+
+on:
+  schedule:
+    # Run daily at 00:00 UTC
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+
+permissions:
+  pull-requests: write
+
+env:
+  # Days of inactivity before a PR is marked stale and notified.
+  DAYS_BEFORE_STALE: 7
+  # Days of further inactivity after being marked stale before the PR is closed.
+  DAYS_BEFORE_CLOSE: 7
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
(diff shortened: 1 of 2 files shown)


HyeockJinKim (commented):
Let's test this out in a different repo first, then merge it.

.github/workflows/stale-pr.yml
```diff
@@ -0,0 +1,38 @@
...
+  DAYS_BEFORE_CLOSE: 7
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          days-before-stale: ${{ env.DAYS_BEFORE_STALE }}
+          days-before-close: ${{ env.DAYS_BEFORE_CLOSE }}
```
HyeockJinKim, line 28: Is there a difference between "stale" and "close"?

.github/workflows/stale-pr.yml
```diff
@@ -0,0 +1,38 @@
...
+  workflow_dispatch:
+
+permissions:
+  pull-requests: write
+
+env:
+  # Days of inactivity before a PR is marked stale and notified.
+  DAYS_BEFORE_STALE: 7
+  # Days of further inactivity after being marked stale before the PR is closed.
+  DAYS_BEFORE_CLOSE: 7
```
HyeockJinKim: After the "stale" status is applied, I think there should be a grace period before closing the case.

HyeockJinKim (approved):

PR #12273, opened 2026-06-17
fix(BA-6539): guarantee unique app proxy subdomain per worker via row lock

fregataa:
## Summary
- Concurrent circuit creation could allocate duplicate subdomains (or ports) on the same worker, because the dedupe check ran against an in-memory snapshot with no isolation.
- Lock the worker row with `SELECT ... FOR UPDATE` while allocating a circuit, so concurrent allocations on the same worker serialize and the taken-subdomain / acquired-port checks stay consistent.

## Test plan
- [x] CI: `pants check` / `pants test` pass

Resolves BA-6539


src/ai/backend/appproxy/coordinator/models/circuit.py
```diff
@@ -106,7 +107,18 @@ class Circuit(Base, BaseMixin):  # type: ignore[misc]
...
+        # so a subdomain must be unique per worker. The rule is scoped to the
+        # wildcard frontend mode via a partial index.
+        sa.Index(
+            "uq_circuits_worker_subdomain",
+            "worker",
+            "subdomain",
+            unique=True,
+            postgresql_where=sa.text("frontend_mode = 'WILDCARD_DOMAIN'"),
+        ),
+    )
```
HyeockJinKim: I'd like to use an app proxy while minimizing database interactions as much as possible, with the intention of removing it later. Is it possible to do this without a migration, using `WITH FOR UPDATE` or similar?

src/ai/backend/appproxy/coordinator/models/subdomain.py
```diff
@@ -0,0 +1,90 @@
...
+_MAX_RETRIES_FOR_UNIQUE_SUBDOMAIN = 20
+_LETTER_DIGIT_HYPHEN_CHARS = frozenset("abcdefghijklmnopqrstuvwxyz0123456789-")
+_HYPHEN_RUN = re.compile(r"-{2,}")
+# Length reserved for a "-<8 hex>" uniqueness suffix so the suffixed form still
+# fits within the DNS label limit.
+_UNIQUE_SUFFIX_LENGTH = 9
+_FALLBACK_PREFIX = "app"
+
+
+class SubdomainGenerator:
```
HyeockJinKim, line 24: The work tends to be too extensive to be classified as a "fix." Is it possible to keep "fix" tasks simple?

HyeockJinKim (approved):

HyeockJinKim:
It doesn't seem ideal from a maintenance perspective, but since we plan to remove the database-related components from the app proxy, I'll accept this as a necessary compromise.

PR #12276, opened 2026-06-18
fix(BA-6536): normalize app proxy subdomains from deployment names

fregataa:
## Summary
- Deployment names are passed verbatim as app proxy wildcard subdomains; a name with a dot (e.g. `glm-5.2-fp8`) or non-ASCII characters yielded an invalid DNS label.
- Add `SubdomainGenerator` (new `Subdomain` type in `common.types`) that folds a name into a single DNS label — lowercase, replace `.`/invalid chars with `-`, Punycode (`xn--`) for non-ASCII, 63-char cap — and returns a collision-free variant against subdomains already taken on the worker.
- Apply it in `add_circuit` for wildcard-domain mode.

## Test plan
- [x] CI: `pants check` / `pants test`
- [x] `tests/unit/appproxy/coordinator/test_subdomain.py`
- [x] Deploy a model with a dotted/non-ASCII name and confirm the issued endpoint URL is a single valid wildcard label

Resolves BA-6536

### DIFF
--- a/src/ai/backend/appproxy/coordinator/models/subdomain.py
+++ b/src/ai/backend/appproxy/coordinator/models/subdomain.py
@@ -41,11 +41,13 @@ class SubdomainGenerator:
             # candidate is empty when the preferred == "." or ""
             candidate = self._fallback_subdomain()
 
+        candidate = self._trim(candidate)
+
         if candidate in taken:
             final = self._find_unique_subdomain(candidate, taken)
         else:
             final = candidate
-        return self._trim(final)
+        return final
 
     def _find_unique_subdomain(self, base: Subdomain, taken: Container[Subdomain]) -> Subdomain:
         """Find a unique subdomain by appending a suffix to ``base``."""
--- a/src/ai/backend/appproxy/coordinator/models/subdomain.py
+++ b/src/ai/backend/appproxy/coordinator/models/subdomain.py
@@ -36,16 +36,30 @@ class SubdomainGenerator:
             candidate = self._fallback_subdomain()
         else:
             candidate = self._normalize(preferred)
+
         if not candidate:
+            # candidate is empty when the preferred == "." or ""
             candidate = self._fallback_subdomain()
-        if candidate not in taken:
-            return candidate
-        trimmed = candidate[: _DNS_LABEL_MAX_LENGTH - _UNIQUE_SUFFIX_LENGTH].strip("-")
-        prefix = trimmed or _FALLBACK_PREFIX
+
+        if candidate in taken:
+            final = self._find_unique_subdomain(candidate, taken)
+        else:
+            final = candidate
+        return self._trim(final)
+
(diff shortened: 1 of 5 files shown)


src/ai/backend/appproxy/coordinator/models/worker.py
```diff
@@ -436,17 +438,12 @@ async def add_circuit(
     circuit_params: dict[str, Any] = {}
 
     if worker.frontend_mode == FrontendMode.WILDCARD_DOMAIN:
-        acquired_subdomains = [c.subdomain for c in worker.circuits]
-        if _requested_subdomain := preferred_subdomain:
-            subdomain = _requested_subdomain
-        else:
-            sub_id = str(uuid.uuid4()).split("-")[0]
-            subdomain = f"app-{sub_id}"
-
```
HyeockJinKim, line 445: There doesn't seem to be a reason to create a Generator every time.

src/ai/backend/appproxy/coordinator/models/subdomain.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def generate_subdomain(
+        self, preferred: Subdomain | None, taken: Container[Subdomain]
+    ) -> Subdomain:
+        if preferred is None:
+            candidate = self._fallback_subdomain()
+        else:
+            candidate = self._normalize(preferred)
+        if not candidate:
+            candidate = self._fallback_subdomain()
```
HyeockJinKim, line 42: Could the "if not candidate" condition be triggered at the end here?

src/ai/backend/appproxy/coordinator/models/subdomain.py
```diff
@@ -0,0 +1,83 @@
...
+        if candidate not in taken:
+            return candidate
+        trimmed = candidate[: _DNS_LABEL_MAX_LENGTH - _UNIQUE_SUFFIX_LENGTH].strip("-")
+        prefix = trimmed or _FALLBACK_PREFIX
```
HyeockJinKim: Could you make sure the `trim` part runs here, rather than in the `fallback` or `normalize` sections? (It would be great if this were guaranteed to apply universally.)

src/ai/backend/appproxy/coordinator/models/subdomain.py
```diff
@@ -0,0 +1,97 @@
...
+
+        if not candidate:
+            # candidate is empty when the preferred == "." or ""
+            candidate = self._fallback_subdomain()
+
+        if candidate in taken:
+            final = self._find_unique_subdomain(candidate, taken)
+        else:
+            final = candidate
+        return self._trim(final)
```
HyeockJinKim: It seems like we should perform the `trim` operation first. If the results of the `trim` operation overlap, how about removing one character at a time and replacing it with a random value?
As it stands now, it looks like the values will overlap after the `trim` operation.

HyeockJinKim (approved):

PR #12277, opened 2026-06-18
fix(BA-6341): accept session UUID in session REST path parameters

jopemachine:
## Summary

Session REST handlers read `request.match_info["session_name"]` and passed it straight into name-keyed actions, so a **UUID (row_id)** in the `{session_name}` path segment was rejected with `Session(name=<UUID>) does not exist`. This regressed the legacy (24.09) behavior where the `{session_name}` segment accepted either a session **name or its id** — an API breaking change introduced during the handler-class migration (BA-4723).

Reported on 26.4.4 for `POST /func/session/{UUID}/shutdown-service` and `.../start-service`, but it affects every session handler that takes the `{session_name}` path parameter.

## Changes

- New **pure id→name** action `resolve_session_name` (service + repository `get_session_name(session_id)`), returning the canonical session name for a session id.
- New `SessionHandler._resolve_session_name(session_name_or_id)` helper: resolves a UUID-shaped reference to its real name via the processor, and passes a plain name through unchanged. The UUID/name discrimination lives at the API boundary, not in the repository.
- All ~19 session handlers (`shutdown-service`, `start-service`, `get_info`, `destroy`, `execute`, `rename`, `commit`, `logs`, `files`, `dependency-graph`, …) now normalize the path reference through this helper before downstream lookups.

## Verification

- Unit tests: `resolve_session_name` service (success + `SessionNotFound`).
- Integration (live local cluster, legacy `GET /session/{ref}` via v1 SDK against a RUNNING session):
  - by name → `200`, by UUID → `200`, identical session identity (same `containerId`/`image`/`creationTime`).
  - bogus UUID → correctly rejected.
  - On `main` (without this fix) the UUID reference reproduces `Session does not exist`.

## Notes

- `match_sessions`' existing name-or-id discrimination in the repository is left untouched here; it can be removed once all callers go through the adapter/handler normalization.

### DIFF
--- a/changes/12277.fix.md
+++ b/changes/12277.fix.md
@@ -1,0 +1,1 @@
+Restore acceptance of a session UUID (in addition to the session name) in the `{session_name}` path of session REST endpoints (e.g. `start-service`, `shutdown-service`), normalizing the reference to the canonical session name at the adapter boundary.
--- a/src/ai/backend/manager/api/rest/session/handler.py
+++ b/src/ai/backend/manager/api/rest/session/handler.py
@@ -85,6 +85,7 @@ from ai.backend.common.identifier.session import SessionID
 from ai.backend.common.types import (
     AgentId,
     KernelId,
+    SessionId,
 )
 from ai.backend.logging import BraceStyleAdapter
 from ai.backend.manager.api.utils import undefined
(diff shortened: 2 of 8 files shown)


jopemachine:
Wrong layer: the REST session endpoints already accept UUID via `match_sessions`. The actual BA-6341 bug is in the **streaming** endpoints (used by the desktop/Electron app proxy), fixed in a new PR.

PR #12278, opened 2026-06-18
feat(BA-6503): make CSP nonce injection opt-in via the 'nonce' keyword

HyeockJinKim:
## Summary
- CSP nonce is now injected only into directives whose config source list contains the `'nonce'` keyword (replaced in place with `'nonce-<random>'`), instead of always nonce-ing `script-src`/`style-src`.
- A directive that also lists `'unsafe-inline'` keeps that source and drops the nonce (a browser ignores `'unsafe-inline'` once a nonce is present); other unsafe-* keywords (`'unsafe-eval'` etc.) are orthogonal and coexist with the nonce.
- Updated CSP config descriptions/examples (`unified.py`, regenerated `sample.toml`) and the hand-maintained `sample.conf` with the opt-in usage guide.

## Test plan
- [ ] `pants test tests/unit/webserver/test_security_policy.py` (nonce keyword opt-in, no-keyword skip, unsafe-inline drops nonce + warning, unsafe-eval keeps nonce, any-directive keyword)
- [ ] Live: set `script-src = ["'self'", "'nonce'"]`, restart webserver, `curl -I` shows `script-src 'self' 'nonce-...'`
- [ ] Live: set `style-src = ["'self'", "'unsafe-inline'", "'nonce'"]`, header has no nonce + warning logged

> Note: behavior change — deployments relying on the old implicit nonce must now add `'nonce'` to opt in. The feature (25.12.0) is unreleased, so blast radius is minimal.

Resolves BA-6503

### DIFF
--- a/src/ai/backend/web/security.py
+++ b/src/ai/backend/web/security.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 import secrets
 from collections.abc import Callable, Iterable, Mapping
 from contextvars import ContextVar
@@ -7,6 +8,10 @@ from typing import Self
 
 from aiohttp import web
 from aiohttp.typedefs import Handler
+
+from ai.backend.logging import BraceStyleAdapter
+
+log = BraceStyleAdapter(logging.getLogger(__spec__.name))
 
 type RequestPolicy = Callable[[web.Request], None]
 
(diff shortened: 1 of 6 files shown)


seedspirit (approved):

src/ai/backend/web/security.py
```diff
@@ -135,6 +153,21 @@ def policy(response: web.StreamResponse) -> web.StreamResponse:
     return policy
 
 
+def _resolve_csp_sources(directive: str, sources: list[str], nonce: str) -> list[str]:
+    if not any(_is_nonce_keyword(source) for source in sources):
+        return sources
```
fregataa, line 158: how about declaring a new var here
`has_nonce_keyword = any(_is_nonce_keyword(source) for source in sources)`

fregataa (approved):

PR #12323, opened 2026-06-20
fix(BA-6570): order scale-in drain by termination priority

HyeockJinKim:
## Summary
- Scale-in selected drain targets by `created_at` only, so it could keep unhealthy or still-provisioning replicas while terminating healthy, serving ones.
- Drain candidates now match the **live set** (`PROVISIONING OR (RUNNING & ACTIVE)`) that the deficit counts, removing the live/drain asymmetry that left PROVISIONING replicas alive during scale-in.
- Candidates are ordered by termination priority: not-yet-serving PROVISIONING routes first, then RUNNING by health (`UNHEALTHY < DEGRADED < NOT_CHECKED < HEALTHY`), ties broken oldest-first.
- Removes the unused `RouteInfo.termination_priority` property and its priority dict (dead code) in favor of shared SQL predicates/`CASE` helpers in the db source.

## Test plan
- [x] `pants fmt :: / fix :: / lint` pass
- [x] `pants test tests/unit/manager/repositories/replica_group::` passes
- [ ] DB-backed unit tests cover: health priority beats created_at; oldest-first tie-break; PROVISIONING drained before RUNNING
- [ ] Live: scale a deployment down with mixed UNHEALTHY / PROVISIONING replicas and confirm non-serving / unhealthy replicas terminate first

Resolves BA-6570

### DIFF
--- a/src/ai/backend/manager/repositories/replica_group/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/replica_group/db_source/db_source.py
@@ -19,6 +19,7 @@ from ai.backend.manager.data.deployment.types import (
     DeploymentHandlerOptions,
     ReplicaGroupHandlerCategory,
     ReplicaGroupLifecycle,
+    RouteHealthStatus,
     RouteStatus,
     RouteSubStatus,
     RouteTrafficStatus,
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/repositories/replica_group/db_source/db_source.py
```diff
@@ -520,15 +543,16 @@ async def _build_drain_updater(
...
+            # Candidates are the live set (same as the deficit count): not-yet-serving
+            # PROVISIONING routes drain before serving RUNNING routes, RUNNING by health.
             query = (
                 sa.select(RoutingRow.id)
                 .where(
                     RoutingRow.replica_group_id == drain.replica_group_id,
                     RoutingRow.revision == drain.revision_id,
-                    RoutingRow.status == RouteStatus.RUNNING,
-                    RoutingRow.traffic_status == RouteTrafficStatus.ACTIVE,
+                    _live_predicate(),
```
fregataa, line 553: is `RoutingRow.traffic_status == RouteTrafficStatus.INACTIVE` not scale-in target?

fregataa (approved):

PR #12331, opened 2026-06-21
docs: refine BEP-1053 ROUTER frontend mode after design review

achimnol:
Refines [BEP-1053: ROUTER Frontend Mode](https://github.com/lablup/backend.ai/blob/bep/refine-1053/proposals/BEP-1053-router-frontend-mode.md) across **two design-review rounds** (2026-06-21, 2026-07-15) and restructures it into the segmented BEP format so a first-time reader can follow intent → key decisions → detailed trade-offs.

## Document restructure

The monolithic BEP is now a main overview + four sub-documents (per the BEP segmentation guide), with mermaid diagrams throughout:

| Document | Contents |
|---|---|
| [main](https://github.com/lablup/backend.ai/blob/bep/refine-1053/proposals/BEP-1053-router-frontend-mode.md) | Motivation & design intent, architecture at a glance, Decision Log (15 entries), Open Questions, phased plan |
| [architecture](https://github.com/lablup/backend.ai/blob/bep/refine-1053/proposals/BEP-1053/architecture.md) | Hierarchical routing model, per-level weight normalization, KV-cache tier scoping, per-replica-group instance lifecycle + in-memory vs subprocess trade-off, HA |
| [coordinator](https://github.com/lablup/backend.ai/blob/bep/refine-1053/proposals/BEP-1053/coordinator.md) | `FrontendMode.ROUTER`, registration validation, per-node liveness, desired-state tables, `/v2/routers/*`, snapshot + events, failure model, key custody |
| [manager](https://github.com/lablup/backend.ai/blob/bep/refine-1053/proposals/BEP-1053/manager.md) | Publication / model API key entities, Manager-side data model, RBAC shrink reconciler, BEP-1049 interplay |
| [migration](https://github.com/lablup/backend.ai/blob/bep/refine-1053/proposals/BEP-1053/migration.md) | Adding a ROUTER surface to an existing AppProxy deployment, version gates, rollback |

## Round 1 (2026-06-21): resolved the four original open questions

- **Placement & HA**: endpoint placement via existing scaling-group routing; per-node liveness (ephemeral `node_id`, valkey liveness set, derived `nodes`) exposed via REST + Prometheus + fleet view.
- **Keys & visibility**: explicit-at-issuance / RBAC-validated / shrink-only visibility with a periodic shrink reconciler; two-tier revocation + opt-in strict all-live-nodes confirmation; **hash-only key custody** (plain SHA-256, plaintext shown once).
- **Data model & propagation**: per-authority publication scope (multi-authority = caller composition); single per-authority `revision` + conditional snapshot polling; first-class **aliases** gated per key; atomic key rotation; `control_mode` to prevent a BEP-1049 dual-writer conflict; per-node best-effort `rate_limit`; `ratio` as relative weight (`0` = drain); authority discovery via `list_workers()`.

## Round 2 (2026-07-15): hierarchy, fail-fast, and the Manager side

- **Hierarchical (chained) routing adopted; the flattened weighted pool is superseded.** Realized *inside* the ROUTER worker as per-replica-group router instances, so continuum-router's replica-level KV machinery (prefix-aware routing, KV index, disaggregated serving) is preserved. Instance substrate (in-memory scopes vs subprocess pool) is recorded as an Open Question with a recommendation (in-memory first, behind a substrate-agnostic seam).
- **Replica-group id / `traffic_weight` propagate over the existing circuit wire** (`replica_groups` on the circuit payload, `replica_group_id` per route) — folds the BA-6233 "weight never reaches AppProxy" follow-up into this proposal; rollout ramps become ordinary route updates.
- **Colocation fail-fast**: registration rejects mixing a ROUTER worker with stock inference workers on one coordinator (review request); the migration document covers the dedicated-coordinator path for existing deployments.
- **Manager-side data model specified**: normalized `model_publications` / `model_publication_names` / `model_publication_endpoints` / `model_api_keys`, CASCADE backstops + service-layer lifecycle hooks — vs the coordinator's deliberately wire-shaped JSONB mirror (rationale recorded).
- **Deterministic mixed-`node_id` liveness** (`nodes` = liveness-set count + legacy counter) so rolling upgrades of stock workers stay correct; coordinator schema recorded as freely evolvable (BEP-1005 has no work plan).
- All four review threads answered in-line; set-operator symbols replaced per review.

## Pairing / cross-repo status

The worker-side counterpart is **lablup/continuum-router#748** (`docs/en/architecture/appproxy-worker.md`). The shipped router implementation (epic lablup/continuum-router#804) implements the now-superseded flattened selection; its rework to the hierarchical model — and the doc re-sync — is tracked on #804.

https://claude.ai/code/session_01DRHPAAYiQxQyMRkzic4Dda
https://claude.ai/code/session_01P38DWyutrMv8kuC8XKjyux


proposals/BEP-1053-router-frontend-mode.md
```diff
@@ -204,32 +242,121 @@ class FrontendMode(enum.StrEnum):
...
+  `ScalingGroupProxyTarget` (scaling group → coordinator `addr` + `api_token`).
+  Point **every** scaling group whose deployments should sit behind one model
+  surface at the **same** coordinator — the one hosting the ROUTER authority.
+  This is what lets a single router serve **multiple scaling groups** (the
+  primary "one endpoint for many deployments across scaling groups" use case).
+- Within that coordinator, workers are distinguished only by `protocol` and
+  `accepted_traffics`. The ROUTER worker is the inference/HTTP worker; with the
+  unbounded-slot treatment above, `pick_worker()` selects it for inference
+  circuits.
+- **Operational constraint (documented, not enforced):** do **not** colocate a
```
jopemachine: Could we fail fast instead of silently misrouting? For example, when an app proxy worker is registered, could we raise an error if a ROUTER and an app proxy worker type that can receive legacy inference traffic are registered at the same time?

proposals/BEP-1053-router-frontend-mode.md
```diff
@@ -240,27 +367,40 @@ Two new tables, scoped per worker authority, with a per-authority monotonic
 router_models:
     id UUID PK
     worker_authority str  (indexed)
-    model_name str
+    model_name str        # primary name; the API/identity key for the publication
+    aliases JSONB         # ["alias-a", …] additional names that route identically
     mappings JSONB        # [{"endpoint_id": UUID, "ratio": float}]
```
jopemachine: Is there a reason to keep this in a JSONB column instead of normalizing it into a separate table?

proposals/BEP-1053-router-frontend-mode.md
```diff
@@ -285,17 +430,34 @@ ack timeout the call still **succeeds** (the persisted state is
...
 The router worker pulls this at registration (full in-memory state restore)
 and on its reconcile timer, reconciling by **set-diff** — the same idiom as
-`reconcile_traefik_etcd_state`, moved worker-side. The opaque `revision`
-makes an unchanged poll a cheap no-op.
+`reconcile_traefik_etcd_state`, moved worker-side. A **single per-authority
+`revision`** covers **both** `router_models` and `router_api_keys`, so a poll
+returns the combined full state whenever anything changed and the worker
+set-diffs the whole thing. The worker passes its last-applied revision via
+`?known_revision=`; an unchanged revision yields a cheap `304`. (An explicit
+query param is used rather than HTTP `ETag`/`If-None-Match` because the
```
jopemachine: For extensibility, I think it would be good to introduce a middleware layer and align the implementation with it.

proposals/BEP-1053-router-frontend-mode.md
```diff
@@ -118,9 +118,31 @@ used consistently throughout:
...
-  expects an API key.
+  generated by the Manager, carrying an explicit, RBAC-validated allowed-model
+  set, and presented to the ROUTER-mode worker the way any OpenAI-compatible
+  SDK expects an API key. Custody is hash-only (§9).
+- **Model publication** (or just *publication*) — the *new* first-class entity
+  that makes a deployment reachable by model name through a router. One
+  publication binds **a primary model name (plus optional aliases)** to **one
+  or more deployment endpoints** (each with a split `ratio`) on **one router
+  authority**, and carries a `control_mode`. It is the object users
+  create/edit/delete; it is stored Manager-side (source of truth) and mirrored
```
jopemachine: Will the same table be added to the manager database with the same schema as well?

proposals/BEP-1053-router-frontend-mode.md
achimnol: Agreed. Adopted as-is in the second review round (2026-07-15). Clarified that:
- Registration now **rejects with an explicit error** any combination that would make one coordinator host both a ROUTER worker and a stock worker whose `accepted_traffics` include `inference`, in either registration order. Interactive stock workers still colocate freely.

Doc relocation:
- Since the doc was restructured in 4d3f1ef68, this now lives in `BEP-1053/coordinator.md` ("`FrontendMode.ROUTER` and registration") with a Decision Log entry in the main doc, and `BEP-1053/migration.md` documents the operational path for existing deployments (a dedicated coordinator for the ROUTER authority; rollback steps included).

proposals/BEP-1053-router-frontend-mode.md
achimnol: Reasoning for keeping aliases as an embedded column rather than a child table, on the coordinator specifically:

- **Every read is whole-entity.** The only consumers (the snapshot builder and the model-updated event) always emit the full publication. A child table would add a join/aggregate to every snapshot and event build, only to reassemble the exact array we stored.
- **Every write is whole-entity.** Mutations arrive as full-state upserts (`PUT` with complete state, last-writer-wins), so a column makes an upsert a single-row write; a child table turns each upsert into a delete-and-reinsert of child rows inside a transaction, for the same end state.
- **Separation of concern.** The heavy burden of validation (e.g., uniqueness of alias) is already handled by the Manager.
  - `model_publication_names` stores primary + aliases as rows with an `is_primary` flag and `UNIQUE` (authority, name) (`BEP-1053/manager.md`, 4d3f1ef68).
- This coordinator-internal design can be updated whenever we want in the future, though, without impacting the wire protocol. Even we proceed with BEP-1005 (merging the coordinator into the manager), the coordinator-specific mirror could be migrated as a "cache" layer.

proposals/BEP-1053-router-frontend-mode.md
achimnol: Yes on persistence, no on "same schema".

The difference is intentional (see the sibling JSONB thread for the coordinator half). The stack splits state three ways:
- The **Manager** owns the normalized source of truth and validation (`model_publications` + `model_publication_names` with `UNIQUE (authority, name)` as a DB constraint `model_publication_endpoints` with CASCADE-backstop FKs + `model_api_keys`, specified in `BEP-1053/manager.md` since 4d3f1ef68).
- The **coordinator** owns a wire-shaped denormalized mirror whose only job is entity-granular replay to stateless router nodes.
- The **router** holds everything in memory only (its key store is explicitly never persisted) and rebuilds from the coordinator's snapshot on (re-)registration.

proposals/BEP-1053-router-frontend-mode.md
achimnol: Recorded the trade-off as an Open Question (`BEP-1053/coordinator.md` "Snapshot endpoint" since 4d3f1ef68).

The catch with a *generic* conditional-GET middleware: it can only compare after the handler has already built the response body, so it saves response bytes but not the DB read. Skipping the read requires a per-route "current version" resolver — which is the `known_revision` check under another name, plus indirection.

Alternative way:
- Adopt the standard **`ETag`/`If-None-Match` header form**, implemented as a small handler-level helper (the authority `revision` as a strong ETag, early 304 before the state query) rather than a blanket middleware.
  - It keeps standard HTTP semantics and lets future snapshot-style endpoints reuse the helper, without pretending the version check can be route-agnostic. The router-side client change is a one-line header swap, which we've confirmed is acceptable.

Checks on ecosystem compatibility:
- The `ETag` header design is compatible with any client. None of the public LLM APIs (OpenAI, OpenRouter, Anthropic) use HTTP conditional caching on their data planes (they cache by request-body keys such as prompt caching, OpenRouter's `X-OpenRouter-Cache-*`) and our ETag use would sit only on the private coordinator↔worker control-plane endpoint, which the end-user clients never touch.

PR #12333, opened 2026-06-22
feat: Expose per-host permissions from the vfolder host list

rapsealk:
## Summary

`GET /folders/_/hosts` returned only `allowed` — the list of host *names* a user may use — while discarding the per-host permission sets that `list_hosts` already computes internally as a `VFolderHostPermissionMap`. As a result, a client cannot tell apart hosts where the user may **create** a vfolder (`create-vfolder`) from hosts where they only hold read/mount permissions (`mount-in-session`, `download-file`, …). The host list therefore advertises hosts that fail later when a create is attempted against them.

This adds an `allowed_permissions` mapping (host → permission names) to the response, sourced from the permission map the service already builds — so clients can mark or hide non-writable hosts up front.

## Changes

- `ListHostsActionResult` (`services/vfolder/actions/storage_ops.py`): add `allowed_permissions: dict[str, list[str]]`.
- `VFolderService.list_hosts` (`services/vfolder/services/vfolder.py`): populate it from `allowed_hosts.to_json()` (the same filtered `VFolderHostPermissionMap` used for `allowed`).
- `ListHostsResponse` (`common/dto/manager/vfolder/response.py`): add `allowed_permissions`, defaulting to `{}` so existing clients parse unchanged.
- `VFolderHandler.list_hosts` (`api/rest/vfolder/handler.py`): pass the field through.
- Tests: service-level assertion that a create-capable host and a read-only host are distinguished; DTO round-trip + backward-compat (absent field → `{}`).

## Compatibility

Additive and backward-compatible: `allowed` is unchanged and `allowed_permissions` is optional with an empty-mapping default. Its keys mirror `allowed` exactly. The per-host permission data was already exposed via the V2 GraphQL storage-host-permissions query; this closes the gap for the REST host-list endpoint.

### DIFF
--- a/src/ai/backend/common/dto/manager/vfolder/response.py
+++ b/src/ai/backend/common/dto/manager/vfolder/response.py
@@ -303,7 +303,7 @@ class ListHostsResponse(BaseResponseModel):
             "by host name (e.g. `create-vfolder`, `mount-in-session`). Lets "
             "clients distinguish hosts the user may create folders on from "
             "read-only hosts, since `allowed` carries only host names. "
-            "Added in 26.4.4."
+            "Added in 26.4.5."
         ),
     )
 
--- a/src/ai/backend/common/dto/manager/vfolder/response.py
+++ b/src/ai/backend/common/dto/manager/vfolder/response.py
@@ -296,6 +296,16 @@ class ListHostsResponse(BaseResponseModel):
     volume_info: dict[str, VolumeInfoDTO] = Field(
         default_factory=dict, description="Volume info per host"
     )
+    allowed_permissions: dict[str, list[str]] = Field(
+        default_factory=dict,
+        description=(
+            "Permissions the requesting user holds on each allowed host, keyed "
+            "by host name (e.g. `create-vfolder`, `mount-in-session`). Lets "
+            "clients distinguish hosts the user may create folders on from "
+            "read-only hosts, since `allowed` carries only host names. "
+            "Added in 26.4.4."
+        ),
+    )
 
 
 class ListAllHostsResponse(BaseResponseModel):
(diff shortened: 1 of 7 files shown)


src/ai/backend/manager/services/vfolder/services/vfolder.py
```diff
@@ -1061,6 +1061,7 @@ async def list_hosts(self, action: ListHostsAction) -> ListHostsActionResult:
             default=default_host,
             allowed=sorted(allowed_hosts),
             volume_info=volume_info,
+            allowed_permissions=allowed_hosts.to_json(),
```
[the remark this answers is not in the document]
rapsealk: Good catch — fixed in c34ff84f2. `allowed_permissions` is now built with sorted hosts and sorted permission names instead of `to_json()`'s set ordering, so the REST response is deterministic. Added a test asserting each permission list is sorted.

src/ai/backend/common/dto/manager/vfolder/response.py
```diff
@@ -296,6 +296,16 @@ class ListHostsResponse(BaseResponseModel):
...
         default_factory=dict, description="Volume info per host"
     )
+    allowed_permissions: dict[str, list[str]] = Field(
+        default_factory=dict,
+        description=(
+            "Permissions the requesting user holds on each allowed host, keyed "
+            "by host name (e.g. `create-vfolder`, `mount-in-session`). Lets "
+            "clients distinguish hosts the user may create folders on from "
+            "read-only hosts, since `allowed` carries only host names. "
+            "Added in 26.4.4."
```
fregataa: it should be 26.4.5

rapsealk: Thanks — corrected to 26.4.5 (since 26.4.4 is already in RC). Updated in the latest push.

PR #12344, opened 2026-06-22
fix(BA-6576): return 4xx not 500 for invalid enqueue resource quantity

jopemachine:
## Summary
- The session enqueue API returned an unhandled **500** when a resource entry `quantity` was not a parseable decimal (e.g. `"4g"`): `ResourceSlotEntry.to_resource_slot` called bare `Decimal(quantity)`, raising `decimal.InvalidOperation`, which does not inherit `BackendAIError` and so fell through the action processor as an "Unexpected error".
- Route the conversion through `ResourceSlot.from_user_input` — the same tolerant parser already used by the legacy (`AgentRegistry.create_session`, `registry.py:496`) and sokovan (`scheduling_controller.py:380`) enqueue paths — so human-readable sizes such as `"4g"` for memory slots are accepted, and genuinely invalid values raise `InvalidResourceSlotQuantity` (**4xx**).
- Scope is limited to the session enqueue scheduler path: `to_resource_slot`'s only callers are `repositories/scheduler/creators.py:73,163`, so deployment / preset / resource-policy paths (which already tolerate `"4g"` by storing it as a string) are untouched — no regression.

## Test plan
- [x] `tests/unit/common/test_types.py::TestResourceSlotEntryToResourceSlot` — plain decimals, `"4g"` accepted, non-decimal & negative → 4xx, empty input
- [x] `pants lint` (ruff) on changed files
- [x] Regression: `sokovan/scheduler`, session & deployment adapter, session v2 DTO tests pass
- [ ] Live: enqueue with `mem: "4g"` → 200, `mem: "abc"` → 400 (manual check)

Resolves BA-6576

### DIFF
--- a/src/ai/backend/common/types.py
+++ b/src/ai/backend/common/types.py
@@ -1398,8 +1398,20 @@ class ResourceSlotEntry(BackendAISchema):
         Transitional helper: repository layer still writes ``ResourceSlot``
         to the DB column, so the final boundary converts back. New
         in-memory pipelines should stay on the entry-list form.
+
+        Quantities are parsed with the same tolerance as user input
+        (:meth:`ResourceSlot.from_user_input`), so human-readable sizes such
+        as ``"4g"`` for memory slots are accepted just like the legacy
+        enqueue path. A non-parseable quantity is rejected with a 4xx
+        :class:`InvalidResourceSlotQuantity` instead of letting
+        ``decimal.InvalidOperation`` propagate as an unhandled 500.
         """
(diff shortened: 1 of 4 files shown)


src/ai/backend/common/types.py
fregataa: How about changing this method name to `inputs_to_resource_slot()` because:
- it is weird to use `ResourceSlot.from_user_input()` in a general parsing method
- To clarify that this method is only used in user-input parsing

seedspirit (approved):

fregataa (approved):

PR #12345, opened 2026-06-22
feat: make the TUI installer log selectable and copyable

devyubin:
## Summary

- The installer's setup log was a Textual `RichLog`, a container widget that cannot be selected, so a failed install left no way to grab the log text for a bug report.
- Switch `SetupLog` from `RichLog` to `Log`, a selectable plain-text widget. Rich renderables are rendered to plain text before write, so the visible output is unchanged.
- Bump `textual` `~=0.79.1` to `~=6.2` (lockfile regenerated). 6.2.1 is the newest release that still satisfies the `rich<14` pin and the first to ship text selection.
- Add `Ctrl+C` to copy the drag-selection, or the whole log when nothing is selected (also works on the report tab after install). It writes via OSC52 and falls back to a native clipboard tool (`pbcopy`, `wl-copy`, `xclip`, `xsel`) for terminals that drop OSC52.
- Detach install-subprocess stdin in `run_exec` so children stop stealing terminal input, which keeps the copy key working for the whole install.
CleanShot 2026-06-22 at 14 09 08@2x

## Test plan

**Unit / static**
- [x] `pants check` (mypy) clean; the lock resolves to `textual 6.2.1`.
- [x] `pants fmt / fix / lint` pass.

**Manual**
- [x] Headless `App.run_test()` pilot: drag-select then `Ctrl+C` copies the selection; with no selection it copies the whole log; report-tab copy works; `pbcopy` receives the text.
- [x] Real terminal (Terminal.app / iTerm2). The VSCode integrated terminal eats `Ctrl+C` before the app sees it, so it can't test the keybinding.


python.lock.metadata
```diff
@@ -103,7 +103,7 @@
     "temporenc~=0.1.0",
     "tenacity>=9.0",
     "testcontainers[minio,postgres,redis]~=4.13.3",
-    "textual~=0.79.1",
+    "textual~=6.2",
```
seedspirit, line 106: The version number has gone up quite a bit all at once—is there a reason for such a sudden increase?

devyubin: We need Textual's text-selection API for the copyable log, and it first shipped in 0.86, so 0.79.x was a non-starter. The number looks like a big leap mostly because Textual bumps its major version quickly after 1.0 (it's on 8.x now), so 0.79 -> 6.2 is less change than the gap suggests. We capped at 6.2.1 because 7.0+ requires rich>=14.2 while we pin rich~=13.6, so it's the newest release that still fits our rich pin.

seedspirit (approved):

PR #12346, opened 2026-06-22
feat(BA-6261): add idle checker persistence foundation

seedspirit:
## Summary
- Add `ABCColumn` + `ABCColumnPayload`: a reusable, domain-agnostic polymorphic JSONB column (write → JSONB dict, load → typed object). Modeled on `PydanticColumn`.
- Add `idle_checkers` / `idle_checker_bindings` ORM rows + Alembic migration (BEP-1054): FK `idle_checker_id → idle_checkers.id` ON DELETE CASCADE, unique `(idle_checker_id, scope_type, scope_id)`, index `(scope_type, scope_id)`.
- Add `IdleCheckerSpecABC` payload base; `idle_checkers.spec` is typed via `ABCColumn(IdleCheckerSpecABC)`. Its `write`/`load` are intentionally **stubbed** (`NotImplementedError`) — polymorphic serialization and the concrete per-`checker_type` specs land in a follow-up, so the `spec` column is not runtime-functional yet.

## Notes / open questions
- **`scope_id` type**: currently `UUID`. BEP-1054 describes scope as a polymorphic string key (domain name / project id / resource group name); only project ids are UUIDs, so domain / resource-group names wouldn't fit a UUID. Should this be `str` instead? Leaving for review.
- `idle_checkers.name` is intentionally **not** unique.
- The data-conversion layer (`to_data` / `IdleCheckerData`) is deferred to a follow-up.

## Test plan
- [ ] `ABCColumn` unit tests pass (write → dict, load → typed subtype, round-trip)
- [ ] `alembic upgrade head` creates both tables / `downgrade` drops them (offline SQL verified; live run pending — local dev DB is behind head)
- [ ] CI green

Resolves BA-6261

### DIFF
--- a/src/ai/backend/manager/models/idle_checker/row.py
+++ b/src/ai/backend/manager/models/idle_checker/row.py
@@ -1,0 +1,73 @@
+from __future__ import annotations
+
+import uuid
+from datetime import datetime
+
+import sqlalchemy as sa
+from sqlalchemy.orm import Mapped, mapped_column
+
+from ai.backend.manager.models.base import GUID, ABCColumn, Base
+from ai.backend.manager.models.idle_checker.spec import IdleCheckerSpecABC
+
+
+class IdleCheckerRow(Base):  # type: ignore[misc]
+    __tablename__ = "idle_checkers"
+
+    id: Mapped[uuid.UUID] = mapped_column(
+        "id", GUID, primary_key=True, server_default=sa.text("uuid_generate_v4()")
+    )
+    name: Mapped[str] = mapped_column("name", sa.String(length=128), nullable=False)
+    description: Mapped[str | None] = mapped_column("description", sa.Text, nullable=True)
+    checker_type: Mapped[str] = mapped_column("checker_type", sa.String(length=64), nullable=False)
+    spec: Mapped[IdleCheckerSpecABC] = mapped_column(
+        "spec", ABCColumn(IdleCheckerSpecABC), nullable=False
+    )
+    created_at: Mapped[datetime] = mapped_column(
+        "created_at", sa.DateTime(timezone=True), nullable=False, server_default=sa.func.now()
+    )
+    modified_at: Mapped[datetime] = mapped_column(
+        "modified_at",
+        sa.DateTime(timezone=True),
+        nullable=False,
+        server_default=sa.func.now(),
+        onupdate=sa.func.now(),
+    )
+
+
+class IdleCheckerBindingRow(Base):  # type: ignore[misc]
+    __tablename__ = "idle_checker_bindings"
+    __table_args__ = (
+        sa.ForeignKeyConstraint(
+            ["idle_checker_id"],
+            ["idle_checkers.id"],
+            name="fk_idle_checker_bindings_idle_checker_id",
+            ondelete="CASCADE",
+        ),
+        sa.UniqueConstraint(
+            "idle_checker_id",
+            "scope_type",
+            "scope_id",
+            name="uq_idle_checker_bindings_checker_scope",
+        ),
+        sa.Index("ix_idle_checker_bindings_scope", "scope_type", "scope_id"),
+    )
+
+    id: Mapped[uuid.UUID] = mapped_column(
+        "id", GUID, primary_key=True, server_default=sa.text("uuid_generate_v4()")
+    )
+    scope_type: Mapped[str] = mapped_column("scope_type", sa.String(length=64), nullable=False)
+    scope_id: Mapped[uuid.UUID] = mapped_column("scope_id", GUID, nullable=False)
+    idle_checker_id: Mapped[uuid.UUID] = mapped_column("idle_checker_id", GUID, nullable=False)
+    enabled: Mapped[bool] = mapped_column(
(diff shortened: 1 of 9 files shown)


src/ai/backend/manager/models/idle_checker/row.py
```diff
@@ -0,0 +1,73 @@
...
+            name="uq_idle_checker_bindings_checker_scope",
+        ),
+        sa.Index("ix_idle_checker_bindings_scope", "scope_type", "scope_id"),
+    )
+
+    id: Mapped[uuid.UUID] = mapped_column(
+        "id", GUID, primary_key=True, server_default=sa.text("uuid_generate_v4()")
+    )
+    scope_type: Mapped[str] = mapped_column("scope_type", sa.String(length=64), nullable=False)
+    scope_id: Mapped[uuid.UUID] = mapped_column("scope_id", GUID, nullable=False)
```
seedspirit, line 63: I set the scope ID to a UUID (since both the scaling group and the domain now have an ID column). Will this cause any major issues? @fregataa

src/ai/backend/manager/models/idle_checker/spec.py
```diff
@@ -0,0 +1,23 @@
+from __future__ import annotations
+
+from typing import Any
+
+from ai.backend.manager.models.base import ABCColumnPayload
+
+
+class IdleCheckerSpecABC(ABCColumnPayload):
```
seedspirit, line 8: It was originally called `IdleCheckerABC` in BEP, but since it seems to be more closely related to the ABC for the Spec than to the behavior of IdleChecker itself, I changed the name.

src/ai/backend/manager/models/idle_checker/row.py
```diff
@@ -0,0 +1,73 @@
...
+from datetime import datetime
+
+import sqlalchemy as sa
+from sqlalchemy.orm import Mapped, mapped_column
+
+from ai.backend.manager.models.base import GUID, ABCColumn, Base
+from ai.backend.manager.models.idle_checker.spec import IdleCheckerSpecABC
+
+
+class IdleCheckerRow(Base):  # type: ignore[misc]
```
jopemachine, line 15: How about IdleCheckerDefinition? (like AppConfigDefinition)

src/ai/backend/manager/models/idle_checker/row.py
```diff
@@ -0,0 +1,73 @@
...
+    __tablename__ = "idle_checkers"
+
+    id: Mapped[uuid.UUID] = mapped_column(
```
jopemachine: Let's add a dedicated ID new type for this.

src/ai/backend/manager/models/idle_checker/spec.py
```diff
@@ -0,0 +1,23 @@
...
+
+    Declares the :class:`ABCColumnPayload` write/load contract but leaves it
+    unimplemented for now. The polymorphic serialization (dispatching by
+    ``checker_type``) and the concrete per-``checker_type`` specs land in a
+    follow-up; this PR only establishes the payload base type.
+    """
+
+    def write(self) -> dict[str, Any]:
+        raise NotImplementedError
+
```
fregataa: the name of this function is a bit out. How about serialize() since the result of this function is used?

src/ai/backend/manager/models/idle_checker/row.py
fregataa: no, there won't be major issue

src/ai/backend/manager/models/idle_checker/row.py
```diff
@@ -0,0 +1,76 @@
...
+
+class IdleCheckerRow(Base):  # type: ignore[misc]
+    __tablename__ = "idle_checkers"
+
+    id: Mapped[IdleCheckerID] = mapped_column(
+        "id", GUID(IdleCheckerID), primary_key=True, server_default=sa.text("uuid_generate_v4()")
+    )
+    name: Mapped[str] = mapped_column("name", sa.String(length=128), nullable=False)
+    description: Mapped[str | None] = mapped_column("description", sa.Text, nullable=True)
+    checker_type: Mapped[str] = mapped_column("checker_type", sa.String(length=64), nullable=False)
```
jopemachine: Is checker_type a value that cannot be represented using StrEnum?

src/ai/backend/manager/models/idle_checker/row.py
```diff
@@ -0,0 +1,76 @@
...
+    spec: Mapped[IdleCheckerSpecABC] = mapped_column(
+        "spec", ABCColumn(IdleCheckerSpecABC), nullable=False
+    )
+    created_at: Mapped[datetime] = mapped_column(
+        "created_at", sa.DateTime(timezone=True), nullable=False, server_default=sa.func.now()
+    )
+    modified_at: Mapped[datetime] = mapped_column(
```
jopemachine: We decided to standardize on updated_at.

jopemachine (dismissed):

jopemachine (approved):

HyeockJinKim (approved):

PR #12365, opened 2026-06-23
feat(BA-6593): add app_config_definition config_name resolver (RBAC groundwork)

jopemachine:
Gates the AppConfigFragment scoped search (BA-6556) on the user's access to the corresponding `app_config_definition` (BEP-1052, epic BA-5781). Two pieces, both delivered here.

## 1. Object-level RBAC registration

On definition create, register the definition as an RBAC object so object-level READ on `(APP_CONFIG_DEFINITION, definition_id)` is resolvable through the scope chain.

- `AppConfigDefinitionGlobalScopeAssociationSpec` (`DependentCreatorSpec`) writes an **AUTO** `association_scopes_entities` row in the same write transaction as the definition.
- Definitions are **system-global**, so the association is anchored at the **GLOBAL scope** (`scope_id=GLOBAL_SCOPE_ID`).

⚠️ **Design note:** `ScopeType.GLOBAL` / `GLOBAL_SCOPE_ID` are marked *deprecated* in the RBAC scope hierarchy, but GLOBAL is the only functional anchor for a system-global entity today (artifact RBAC uses the same). The action layer expresses GLOBAL as `scope_id=""`, whereas RBAC data rows (permissions/associations) use `GLOBAL_SCOPE_ID="global"` — this PR uses `"global"` so the bulk scope-chain query (exact `(scope_type, scope_id)` match) resolves. If we'd rather not anchor on a deprecated GLOBAL scope, that's the decision to make here.

## 2. config_name resolver

The bulk RBAC validator resolves a flat `(element_type, entity_id)` key — **no FK/alternate-key resolution** — and keys `entity_id` by the definition **UUID**. The fragment scoped search only holds a `config_name`, so it must first resolve it.

- `repository.by_config_name(config_name)` (+ db_source) → `AppConfigDefinitionData`, raising `AppConfigDefinitionNotFound`.
- `ResolveAppConfigDefinitionByConfigName` service action + processor. Direction is **`config_name` → `id`** (what the gate needs).

## Tests
- Real-DB: `by_config_name` (hit/miss); `create` writes the GLOBAL/AUTO association.
- Mocked: `resolve_by_config_name` service path (hit/miss).
- `AssociationScopesEntitiesRow` added to the definition / allow-list `with_tables`.

## Follow-ups (same issue BA-6593)
- **Backfill** associations for pre-existing definitions (data migration).
- Grant READ on `app_config_definition` to the relevant roles (policy/seed).
- The fragment-side gate (build `RBACElementRef(APP_CONFIG_DEFINITION, resolved_id)` + wire `validators.rbac.bulk`) lands in **BA-6556 (REST v2)**.

Jira: BA-6593

### DIFF
--- a/src/ai/backend/common/data/permission/types.py
+++ b/src/ai/backend/common/data/permission/types.py
@@ -359,6 +359,10 @@ class ScopeType(enum.StrEnum):
     NOTIFICATION_CHANNEL = "notification_channel"
     KEYPAIR = "keypair"
     KEYPAIR_RESOURCE_POLICY = "keypair_resource_policy"
+    # An app config definition acts as a scope so per-definition READ can gate the
+    # AppConfigFragment scoped search (BA-6593). Other app_config element types become
+    # scopes only when an operation actually needs it.
+    APP_CONFIG_DEFINITION = "app_config_definition"
 
     def to_element(self) -> RBACElementType:
         from ai.backend.common.exception import RBACTypeConversionError
--- a/changes/12365.feature.md
+++ b/changes/12365.feature.md
@@ -1,0 +1,1 @@
+Add an app_config_definition resolver that looks up a registered config name and returns its definition (and id), so callers holding only a config name can authorize against the definition.
--- a/changes/12365.feature.md
+++ b/changes/12365.feature.md
@@ -1,1 +1,1 @@
-Add an app_config_definition resolver that looks up a registered config name and returns its definition (and id), so callers holding only a config name can authorize against the definition.
+Register each app_config_definition as a GLOBAL-scope RBAC object on create and add a config-name resolver, so callers holding only a config name can authorize against the definition.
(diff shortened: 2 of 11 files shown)


src/ai/backend/common/data/permission/types.py
```diff
@@ -359,6 +359,10 @@ class ScopeType(enum.StrEnum):
     NOTIFICATION_CHANNEL = "notification_channel"
     KEYPAIR = "keypair"
     KEYPAIR_RESOURCE_POLICY = "keypair_resource_policy"
+    # An app config definition acts as a scope so per-definition READ can gate the
+    # AppConfigFragment scoped search (BA-6593). Other app_config element types become
+    # scopes only when an operation actually needs it.
```
jopemachine, line 364: Remove this

jopemachine:
Closing: this PR's resolver (config_name→id) and the `ScopeType.APP_CONFIG_DEFINITION` addition were both groundwork for RBAC-gating the AppConfigFragment scoped search. We've decided the scoped search stays on `BaseBulkAction[SearchableActionTarget]` **without** wiring an RBAC validator (RBAC is opt-in at the processor), so neither change is needed. Continuing in the app_config_fragment PR stack instead.

PR #12367, opened 2026-06-23
fix: Scope group vfolders by `group_id` in the REST folder list

### LINKED ISSUE #12363: REST GET /folders silently ignores group_id (scope not applied in VFolderService.list)
## Summary

The REST endpoint `GET /folders` (vfolder list) accepts a `group_id` query
parameter but **silently ignores it** — the returned set is identical with or
without `group_id`. Verified on **26.4.4**.

## Details

`api/rest/vfolder/handler.py::list_folders` resolves `group_id` into a scope:

```python
group_id = params.group_id
if group_id is not None:
    scope_type = ScopeType.PROJECT
    scope_id = str(group_id)
else:
    scope_type = ScopeType.USER
    scope_id = str(owner_user_uuid)
result = await self._vfolder.list_vfolder.wait_for_complete(
    ListVFolderAction(user_uuid=owner_user_uuid, _scope_type=scope_type, _scope_id=scope_id)
)
```

…but `services/vfolder/services/vfolder.py::VFolderService.list` never uses the
scope. It calls `list_accessible_vfolders(user_id, user_role, domain_name,
allowed_vfolder_types)` with **no** scope filter, and the action's
`_scope_type`/`_scope_id` are only echoed back into the result:

```python
return ListVFolderActionResult(
    user_uuid=action.user_uuid,
    vfolders=vfolders,
    _scope_type=action.scope_type(),   # echoed, never used to filter
    _scope_id=action.scope_id(),
)
```

So the result is always the wide `query_accessible_vfolders` union (user-owned +
invited + every MODEL_STORE project in the domain + all-domain groups for admins
+ all group vfolders for superadmins), regardless of `group_id`.

## Reproduction (26.4.4)

Signed `GET /folders` as both a superadmin and a `role: user` account:

```
/folders  ==  /folders?group_id=<projectA>  ==  /folders?group_id=<model-store>
```

…return identical result sets every time.

## Impact

Clients cannot scope the REST vfolder list to a single project. This looks like
an incomplete RBAC refactor: the handler wires the scope but the service still
calls the legacy union. Downstream UIs that use this endpoint surface every
MODEL_STORE / cross-project group folder. (The `vfolder_nodes` GraphQL
connection with `scope_id: "project:<id>"` *does* scope correctly, and the WebUI
relies on it — but the REST `group_id` param remains misleading.)

## Suggested fix

Either honor the scope in `VFolderService.list` (filter group vfolders by the
PROJECT scope when `_scope_type == PROJECT`, restrict to the user when USER), or
remove the dead `group_id` param and document `vfolder_nodes` as the scoped path.

Found while debugging a downstream client (Backend.AI FastTrack) that showed
many empty, duplicate-named project folders on its data-folder page.

rapsealk:
Fixes #12363.

## Problem

`GET /folders` accepts a `group_id` query parameter, and the REST handler resolves it into a **PROJECT** scope on `ListVFolderAction` (`api/rest/vfolder/handler.py` — `group_id is not None → ScopeType.PROJECT`). But `VFolderService.list` never consumed that scope: it called `list_accessible_vfolders` without it and only echoed `_scope_type`/`_scope_id` back in the result. The listing therefore always returned the full accessible union — every `MODEL_STORE` project, every domain project for admins, and all group vfolders for superadmins — regardless of the requested scope. **`group_id` was a silent no-op.**

Verified empirically on a 26.4.4 manager: `/folders` returns byte-identical results to `/folders?group_id=<project>` for both a superadmin and a regular user.

## Why it matters downstream

- FastTrack's folder pickers pass `group_id` and had to ship a client-side scope filter to compensate (lablup/backend.ai-fasttrack#4634).
- FastTrack resolves pipeline mount names → UUIDs through this listing at launch time; the missing scope semantics were part of the failure surface behind a customer incident where a batch task ran without its requested folder (lablup/backend.ai-fasttrack#5016).

## Change

Thread the action's PROJECT scope through the layers that already exist for it:

- `VFolderService.list`: when `action.scope_type() == ScopeType.PROJECT`, pass `project_id=UUID(action.scope_id())` to the repository.
- `VfolderRepository.list_accessible_vfolders(project_id=...)`: forwards `extra_vf_group_conds=(VFolderRow.group == project_id)` into the existing `query_accessible_vfolders` hook.

Scope semantics: **project-owned vfolders are restricted to the requested project; user-owned and invited vfolders are unaffected** — so the default (no `group_id`) listing is byte-for-byte unchanged, and a scoped listing still includes the caller's personal folders exactly as the WebUI expects.

## Status

Rebased onto current `main` (2026-07-15) — the branch had gone stale and was auto-closed; the REST-handler contract was re-verified after the `api/rest/` reorganization (`handler.py:338-351` still resolves `group_id` → PROJECT scope), and `tests/unit/manager/services/vfolder/test_vfolder_crud_service.py` (including the three new scope tests) passes via pants.


src/ai/backend/manager/repositories/vfolder/repository.py
```diff
@@ -415,9 +416,13 @@ async def list_accessible_vfolders(
         domain_name: str,
         allowed_vfolder_types: list[str],
         extra_conditions: sa.sql.elements.ColumnElement[bool] | None = None,
+        group_scope_id: UUID | None = None,
```
seedspirit: does group mean project? If so, It would be better to change name to `project_scope_id`

src/ai/backend/manager/services/vfolder/services/vfolder.py
```diff
@@ -579,12 +581,22 @@ async def list(self, action: ListVFolderAction) -> ListVFolderActionResult:
...
+        if action.scope_type() == ScopeType.PROJECT:
+            group_scope_id = UUID(action.scope_id())
+
         # Use repository to get accessible vfolders
         vfolder_list_result = await self._vfolder_repository.list_accessible_vfolders(
             user_id=action.user_uuid,
             user_role=user_role,
             domain_name=user_domain_name,
             allowed_vfolder_types=list(allowed_vfolder_types),
+            group_scope_id=group_scope_id,
```
seedspirit: Does `list_accessible_vfolders` accept `group_scope_id` as an argument? I checked, and it doesn't seem to exist as a separate parameter, but this needs to be verified.

seedspirit (changes requested):

src/ai/backend/manager/repositories/vfolder/repository.py
rapsealk: Yes — `group` is the legacy term for project here. Renamed the parameter to `project_id` in 936766667. (You suggested `project_scope_id`; the maintainer preferred the shorter `project_id`, which also matches the existing `project_id` usage in these tests.)

src/ai/backend/manager/services/vfolder/services/vfolder.py
rapsealk: It does — this PR adds the parameter to `list_accessible_vfolders` itself (see the same diff, `repositories/vfolder/repository.py`), now named `project_id` after 936766667. The service just passes it through to that new parameter.

rapsealk:
Reopened — this went stale, but the bug it fixes just bit us downstream in FastTrack: a customer pipeline ran its batch command without a requested folder because name→UUID mount resolution relies on this listing's scope semantics (lablup/backend.ai-fasttrack#5016). The `group_id` no-op also forced a client-side workaround in the FastTrack folder pickers (backend.ai-fasttrack#4634).

The head branch was restored as-is; it likely needs a rebase onto current `main` before review — will follow up. Keeping #12363 open as the tracking issue.

rapsealk:
Rebased onto current `main` and reframed the description. The service/repository seams moved slightly since June (REST handler now lives under `api/rest/vfolder/`) but the contract this PR relies on is unchanged — re-verified that the handler still resolves `group_id` into a PROJECT scope that `VFolderService.list` previously dropped. Unit tests for the vfolder CRUD service pass on the rebased branch. Ready for review.

HyeockJinKim (commented):
We do not plan to add any more fields to the legacy API unless they are required. Please use the v2 API.

PR #12373, opened 2026-06-23
fix(BA-6595): backfill custom runtime variant service block when missing

seedspirit:
## Problem
`f3a8c1d05e64` / `ed42bc179b91` only backfilled the default `port` when the `custom` runtime variant's `models[0].service` was already a JSON object. Databases whose `custom` baseline has **no `service` block** (or no `models[0]`) were left without a port, so model-service deployments still fail with `port: Field required` (`port` is required; `health_check` is optional).

## Fix
A new idempotent migration converges the `custom` baseline for every remaining state:

| State | Action |
|---|---|
| `models[0]` null / `[]` / absent | seed full definition |
| `models[0]` present, `service` null/absent | create full service block |
| `service` present, `port` null/absent | set default port (8080) |

Existing non-default ports and custom `health_check`s are preserved (no clobber); re-running is a no-op.

## Migration chain (per `models/alembic/README.md` backport strategy)
- `c7e1a9d4f6b2` (d) — inserted after `9fc9e6bfe695`; the **same revision id is reused on the 26.4 backport** so the chains don't diverge.
- `2d6443ac0d4a` repointed `9fc9e6bfe695 → c7e1a9d4f6b2`.
- `e3b8d2a1c5f7` (d') — idempotent head duplicate for DBs already upgraded past the insertion point.

Single head after: `e3b8d2a1c5f7`.

## Verification
- SQL logic tested against the local DB across all states + idempotency + no-clobber, rolled back (nothing persisted).
- `alembic heads` single; `pants fmt/lint/check` clean.

## Backport
26.4 backport: see linked PR.

### DIFF
--- a/changes/12373.fix.md
+++ b/changes/12373.fix.md
@@ -1,0 +1,1 @@
+Backfill a default service block (with port) for the `custom` runtime variant baseline when it was seeded without one, so model-service deployments no longer fail with `port: Field required`.
--- a/src/ai/backend/manager/models/alembic/versions/2d6443ac0d4a_create_app_config_allow_list_table.py
+++ b/src/ai/backend/manager/models/alembic/versions/2d6443ac0d4a_create_app_config_allow_list_table.py
@@ -8,7 +8,7 @@ foreign key into ``app_config_definitions`` so an entry may only reference a
 registered config name.
 
 Revision ID: 2d6443ac0d4a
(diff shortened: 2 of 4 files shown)


seedspirit:
26.4 backport: #12374

fregataa (approved):

PR #12396, opened 2026-06-24
refactor(BA-6608): move QueryCondition/QueryOrder to models layer to break models→repositories imports

fregataa:
## Summary
- Move the `QueryCondition` / `QueryOrder` SQLAlchemy type aliases from `repositories/base` down to a new leaf module `models/query_types.py` (DB layer), so that the 83 `models/*/{conditions,orders}.py` modules no longer import upward into the repositories layer.
- Make `models.query_types` the single source of truth: all upper-layer consumers (api adapters, repositories, services, sokovan, tests) import the aliases directly from there; `repositories/base` no longer re-exports them.
- Removes the `models → repositories` layer-violating import edges (84 → 0, excluding the `vfolder/row.py` purger case tracked separately as BA-6616), which inflate the dependency SCC and the transitive `pants test` / `pants check` re-run scope.

This is part of epic BA-6607 (remove manager cross-layer import violations). Pure import-relocation — no behavior change.

## Test plan
- [x] `pants --changed-since=origin/main fmt fix lint` — clean
- [x] `pants check` (mypy) — passes
- [x] `pants test` model condition unit tests — pass (integration tests require halfstack/Docker)
- [x] CI green

Resolves BA-6608


HyeockJinKim (commented):
1. Is there a better name than “query_types”?
2. I’m also thinking about moving the “repositories/base” section.

seedspirit (approved):

PR #12397, opened 2026-06-24
fix(BA-6597): preserve container-registry API error category in metrics

seedspirit:
## Summary
- Registry-side API errors (e.g. Harbor `FORBIDDEN`) during rescan were raised as a bare `RuntimeError`, so the action processor caught them in its generic `except BaseException` branch and tagged them `backendai_generic_internal-error` in `backendai_action_count_total` — masking the real failure category (forbidden vs genuine internal error).
- Add `PassthroughError.from_http_status()` in `common/exception.py`: a standard HTTP-status → `ErrorDetail` mapping (4xx + 5xx; unmapped → `INTERNAL_ERROR`) that preserves the upstream status code.
- Raise `PassthroughError.from_http_status(...)` from the Harbor/GitLab/GitHub scanners (`fetch_repositories`, Harbor untag) instead of bare `RuntimeError`, so the failure category survives into metrics and audit logs.

## Test plan
- [ ] `pants test tests/unit/common/test_exception.py` — `from_http_status` mapping (403→forbidden, 503→unavailable, unmapped→internal, status preserved)
- [ ] CI: container_registry unit + component suites (no regression)

Resolves BA-6597

### DIFF
--- a/src/ai/backend/manager/container_registry/github.py
+++ b/src/ai/backend/manager/container_registry/github.py
@@ -4,6 +4,7 @@ from typing import override
 
 import aiohttp
 
+from ai.backend.common.exception import ErrorDomain, ErrorOperation, PassthroughError
 from ai.backend.logging import BraceStyleAdapter
 from ai.backend.manager.exceptions import ContainerRegistryProjectEmpty
 
@@ -51,6 +52,9 @@ class GitHubRegistry(BaseContainerRegistry):
                     else:
                         break
                 else:
-                    raise RuntimeError(
-                        f"Failed to fetch repositories! {response.status} error occured."
+                    raise PassthroughError.from_http_status(
+                        response.status,
+                        domain=ErrorDomain.CONTAINER_REGISTRY,
+                        operation=ErrorOperation.LIST,
+                        error_message=f"Failed to fetch repositories! {response.status} error occurred.",
                     )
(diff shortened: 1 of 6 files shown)


src/ai/backend/manager/container_registry/harbor.py
```diff
@@ -186,7 +187,12 @@ async def untag(
                     if (
                         e.status != 404
                     ):  #  404 means image is already removed from harbor so we can just safely ignore the exception
-                        raise RuntimeError(f"Failed to untag {image}: {e.message}") from e
+                        raise PassthroughError.from_http_status(
+                            e.status,
+                            domain=ErrorDomain.CONTAINER_REGISTRY,
+                            operation=ErrorOperation.SOFT_DELETE,
```
jopemachine: untag is a hard deletion operation that actually removes the image from the container registry.

src/ai/backend/manager/container_registry/github.py
```diff
@@ -51,6 +52,9 @@ async def fetch_repositories(self, sess: aiohttp.ClientSession) -> AsyncIterator
                     else:
                         break
                 else:
-                    raise RuntimeError(
-                        f"Failed to fetch repositories! {response.status} error occured."
+                    raise PassthroughError.from_http_status(
+                        response.status,
+                        domain=ErrorDomain.CONTAINER_REGISTRY,
+                        operation=ErrorOperation.LIST,
```
jopemachine: Wouldn't UPDATE be more appropriate than LIST?

fregataa (commented):
Let's set a target milestone to 26.4

fregataa (approved):

PR #12398, opened 2026-06-24
feat(BA-6581): wire IdleCheckerRepository into idle-check reconciler source

seedspirit:
## Summary
- Add `IdleCheckerRepository` (DB source + result types) that fetches, per reconciler tick, the RUNNING sessions and the scope-bound idle checkers applicable to each, and wire it through orchestration DI + the reconciler stage factory so `IdleCheckSource` returns real data instead of a placeholder.
- Model the checker spec as a single `IdleCheckerSpec` (`type` discriminator + optional per-kind sub-specs `session_lifetime` / `network` / `utilization`) stored in `idle_checkers.spec` JSONB via `PydanticColumn`; a `model_validator` enforces the matching sub-spec. Each spec declares its `target_session_types`.
- Move the spec + `CheckerType` out of the ORM layer to `common/data/idle_checker/types.py` (pure Pydantic); `ScopeType` / `ScopeRef` stay in `manager/data`.
- Batch read (bindings are few, sessions many): read bindings first, then select candidate sessions with one `(scope_id = ? AND session_type IN <checker targets>)` term per binding — using the new `sessions.resource_group_id` / `domain_id` columns, no joins — and attach the matching checkers per session in memory.

No DB migration: `idle_checkers` is unreleased and the JSONB column is unchanged.

## Notes
- `repositories/conftest.py` eagerly registers all models so mapper configuration is independent of Pants test sharding.

## Test plan
- [x] `pants fmt / fix / lint / check / test` on changed targets (full `--changed-since` green)
- [x] Repository tests cover scope-binding resolution and per-checker session-type filtering

Resolves BA-6581


src/ai/backend/common/data/idle_checker/types.py
```diff
@@ -0,0 +1,52 @@
...
+    UTILIZATION = "utilization"
+
+
+class SessionLifetimeSpec(BackendAISchema):
+    """Config for ``CheckerType.SESSION_LIFETIME``.
+
+    Concrete fields (e.g. max lifetime) land with the checker-logic stories.
+    """
+
+    type: Literal[CheckerType.SESSION_LIFETIME] = CheckerType.SESSION_LIFETIME
```
HyeockJinKim: Why is "Literal" used here?

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
```diff
@@ -0,0 +1,142 @@
...
+                SessionRow.id.label("session_id"),
+                SessionRow.created_at.label("session_created_at"),
+                SessionRow.starts_at.label("session_starts_at"),
+                ScalingGroupRow.id.label("resource_group_id"),
+                SessionRow.group_id.label("project_id"),
+                DomainRow.id.label("domain_id"),
+            )
+            .select_from(SessionRow)
+            .join(DomainRow, SessionRow.domain_name == DomainRow.name)
+            .join(ScalingGroupRow, SessionRow.scaling_group_name == ScalingGroupRow.name)
```
HyeockJinKim: Is there a reason to use a JOIN like this?

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
```diff
@@ -0,0 +1,142 @@
...
+
+        async with self._db.begin_readonly_session() as db_sess:
+            binding_rows = (await db_sess.execute(binding_query)).fetchall()
+            if not binding_rows:
+                return IdleCheckSnapshot(
+                    session_views_by_id={},
+                    bindings_by_scope={},
+                    checkers_by_id={},
+                )
+            session_rows = (await db_sess.execute(session_query)).fetchall()
```
HyeockJinKim: Please also consider whether it might be better to use a batch query here.

src/ai/backend/manager/sokovan/idle_check/types.py
```diff
@@ -40,12 +41,12 @@ class IdleCheckTargetStatuses(BaseReconcilerTargetStatuses):
...
 @dataclass
 class IdleCheckReconcileInfo(BaseReconcilerInfo):
-    session_ids: Sequence[SessionId]
+    snapshot: IdleCheckSnapshot
     current_time: datetime
 
     @override
     def entity_ids(self) -> Sequence[UUID]:
-        return self.session_ids
+        return list(self.snapshot.session_views_by_id.keys())
```
HyeockJinKim: We're doing batch processing, and while I'm not sure if it would be best to manage this part using a dictionary, I think it would be a good idea to consider the storage structure.

src/ai/backend/manager/repositories/idle_checker/types.py
```diff
@@ -0,0 +1,45 @@
...
+class IdleCheckSnapshot:
+    """Normalized idle-check source data.
+
+    Session views carry their applicable scopes, scopes reference their checker bindings, and
+    checker specs are loaded once by id.
+    """
+
+    session_views_by_id: Mapping[SessionId, IdleCheckSessionView]
+    bindings_by_scope: Mapping[ScopeRef, Sequence[IdleCheckerBindingRef]]
+    checkers_by_id: Mapping[IdleCheckerID, IdleCheckerRef]
```
HyeockJinKim: Please review whether all of these maps are necessary.

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
seedspirit: This is to look up the ID in Scope. Since the current session row only contains the domain name and resource group name, a join is required to determine their IDs.

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
seedspirit: I am currently retrieving all bindings and session rows at once, without using N+1. Are you suggesting I use a batch querier?

src/ai/backend/manager/repositories/idle_checker/types.py
seedspirit: Although I had previously included various types of maps for the convenience of external I/O, I’ve since changed this because I believe that responsibility lies with the handler or idle checker, not with the repository and source that use this data class.

src/ai/backend/manager/data/idle_checker/types.py
```diff
@@ -1,11 +1,33 @@
...
+    scope_id: uuid.UUID
+
+
+@dataclass(frozen=True)
+class IdleCheckSession:
+    """Session fields needed to evaluate idle checkers."""
+
+    session_id: SessionId
+    created_at: datetime
+    starts_at: datetime | None
```
jopemachine, line 15: Nit: started_at

jopemachine: Although using the past tense would be more consistent, if it's already named starts_at, please leave it as is.

src/ai/backend/manager/repositories/idle_checker/types.py
jopemachine, line 1: If possible, I don't want to comment on naming, but wasn't it the convention to use the Data suffix for these data pattern types?

Is there a reason for omitting the Data suffix here?

I think it would be clearer if the type's purpose could be identified from its name alone.

jopemachine (dismissed):

src/ai/backend/manager/data/idle_checker/types.py
seedspirit: Since it was set to "starts_at" in the session row, I adjusted it accordingly.

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
```diff
@@ -0,0 +1,145 @@
...
+            .select_from(SessionRow)
+            .join(DomainRow, SessionRow.domain_name == DomainRow.name)
+            .join(ScalingGroupRow, SessionRow.scaling_group_name == ScalingGroupRow.name)
+            .where(
+                sa.and_(
+                    # The caller (idle-check stage) owns which statuses are idle-check targets;
+                    # today that is RUNNING only. Kept as a parameter so the policy stays in the
+                    # stage's target_statuses rather than baked into this read.
+                    SessionRow.status.in_(session_statuses),
+                    SessionRow.session_type != SessionTypes.INFERENCE,
```
fregataa: This condition (session_type) also can be a reconcile target

fregataa (dismissed):

src/ai/backend/manager/sokovan/stages/idle_check.py
```diff
@@ -36,7 +40,13 @@ def build_idle_check_stage() -> ReconcilerStageRegistration:
...
         kind=IdleCheckKind.SESSION,
         target_statuses=IdleCheckTargetStatuses(
-            session_statuses=frozenset({SessionStatus.RUNNING})
+            session_statuses=frozenset({SessionStatus.RUNNING}),
+            # Inference sessions follow the deployment/routing lifecycle, not idle termination.
+            session_types=frozenset({
+                SessionTypes.INTERACTIVE,
+                SessionTypes.BATCH,
+                SessionTypes.SYSTEM,
+            }),
```
HyeockJinKim: Why is the system session targeted?

fregataa (dismissed):

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
```diff
@@ -0,0 +1,143 @@
...
+                        IdleCheckerBindingRow.scope_type == ScopeType.PROJECT.value,
+                        IdleCheckerBindingRow.scope_id == SessionRow.group_id,
+                    ),
+                    sa.and_(
+                        IdleCheckerBindingRow.scope_type == ScopeType.DOMAIN.value,
+                        IdleCheckerBindingRow.scope_id == SessionRow.domain_id,
+                    ),
+                ),
+            )
+        )
```
HyeockJinKim: I'm a little disappointed with the query binding part—wouldn't it make more sense to query based on IdleChecker or the session?

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
```diff
@@ -0,0 +1,175 @@
...
+                IdleCheckerRow.spec,  # typed: PydanticColumn loads it to an IdleCheckerSpec
+            )
+            .join(IdleCheckerRow, IdleCheckerBindingRow.idle_checker_id == IdleCheckerRow.id)
+            .where(IdleCheckerBindingRow.enabled == sa.true())
+            .order_by(
+                IdleCheckerBindingRow.scope_type,
+                IdleCheckerBindingRow.scope_id,
+                IdleCheckerBindingRow.created_at,
+                IdleCheckerBindingRow.idle_checker_id,
+            )
```
HyeockJinKim: Is there a specific reason why "order_by" has to be included this way?

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
```diff
@@ -0,0 +1,175 @@
...
+        )
+
+        querier = BatchQuerier(pagination=NoPagination())
+        async with self._ops.read_ops() as r:
+            binding_rows = (await r.batch_query_in_global(binding_query, querier)).rows
```
HyeockJinKim: It feels really unnatural to put the entire query in the `select` clause and leave the `Querier` empty.

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
```diff
@@ -0,0 +1,175 @@
...
+                .where(
+                    sa.and_(
+                        SessionRow.status.in_(session_statuses),
+                        sa.or_(*candidate_conditions),
+                    )
+                )
+                .order_by(SessionRow.created_at, SessionRow.id)
+            )
+            session_rows: list[_CandidateSession] = []
+            for row in (await r.batch_query_in_global(session_query, querier)).rows:
```
HyeockJinKim: It's the same here, too.

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
```diff
@@ -0,0 +1,175 @@
...
+class _CandidateSession:
+    """Typed view of one session_query row (fields follow the SELECT order)."""
+
+    id: SessionId
+    created_at: datetime
+    starts_at: datetime | None
+    session_type: SessionTypes
+    resource_group_id: ResourceGroupID
+    group_id: UUID
+    domain_id: DomainID
```
HyeockJinKim: I do wonder if a dataclass is really necessary just for internal values. Also, I'm not sure if reading this value from a row based on name mapping is a good constraint.

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
```diff
@@ -0,0 +1,175 @@
...
+
+class IdleCheckerDBSource:
+    _ops: DBOpsProvider
+
+    def __init__(self, ops_provider: DBOpsProvider) -> None:
+        self._ops = ops_provider
+
+    async def fetch_idle_check_batch(
+        self,
+        session_statuses: Collection[SessionStatus],
```
HyeockJinKim: It seems like the right way to do this is through a query.

HyeockJinKim (changes requested):

src/ai/backend/manager/repositories/idle_checker/repository.py
```diff
@@ -0,0 +1,86 @@
...
+            idle_check_candidates.append((bound_checker.scope, target_session_types))
+
+        session_querier = BatchQuerier(
+            pagination=NoPagination(),
+            conditions=[
+                SessionConditions.by_statuses(session_statuses),
+                SessionConditions.by_idle_check_candidates(idle_check_candidates),
+            ],
+        )
+        candidate_sessions = await self._db_source.fetch_candidate_sessions(session_querier)
```
seedspirit, line 65: I wonder it's okay to fetch sessions in another transaction

src/ai/backend/common/data/idle_checker/types.py
```diff
@@ -0,0 +1,76 @@
...
+
+    ``type`` selects which sub-config is valid; exactly one sub-config field must
+    be present for the chosen type (validated below). Stored via ``PydanticColumn``.
+    """
+
+    type: CheckerType = Field(description="Idle checker kind; selects the sub-config.")
+    target_session_types: frozenset[SessionTypes] = Field(
+        default=frozenset({SessionTypes.INTERACTIVE, SessionTypes.BATCH}),
+        description="Session types this checker applies to; other types are skipped.",
+    )
```
HyeockJinKim: Since these fields are different in nature, wouldn't it be better to manage them separately?

src/ai/backend/manager/data/idle_checker/types.py
```diff
@@ -1,11 +1,33 @@
...
+    PROJECT = "project"
+    DOMAIN = "domain"
+
+
+@dataclass(frozen=True)
+class ScopeRef:
+    """A concrete scope a binding targets (UUID-identified)."""
+
+    scope_type: ScopeType
+    scope_id: uuid.UUID
```
HyeockJinKim: ScopeID

src/ai/backend/manager/data/idle_checker/types.py
```diff
@@ -1,11 +1,33 @@
...
-    SESSION_LIFETIME = "session_lifetime"
-    NETWORK_TIMEOUT = "network_timeout"
-    UTILIZATION = "utilization"
+class ScopeType(enum.StrEnum):
+    """Scope a binding attaches a checker to. Declared most- to least-specific."""
+
+    RESOURCE_GROUP = "resource_group"
+    PROJECT = "project"
+    DOMAIN = "domain"
+
```
HyeockJinKim: It seems like a very bad idea to have a separate scope type with the same name.

src/ai/backend/manager/repositories/idle_checker/repository.py
fregataa: Can we do this in a single transaction?

fregataa (commented):
I lost context of a reason why Idle checker spec changed from an abc to a plain data. There is no abc anymore?

src/ai/backend/manager/repositories/idle_checker/repository.py
seedspirit: Before the change, this was handled within a single transaction, but it seems difficult to implement this pattern using a querier. The querier would need to be created in the repository layer and passed to the DB source, but doing so would result in separate transactions.

HyeockJinKim (approved):

PR #12418, opened 2026-06-25
feat(BA-6551): expose single-string model service command on v2 API

seedspirit:
## Summary
- Add a `command: String` field and deprecate `start_command: [String!]` across the **deployment-revision** and **revision-preset** v2 stacks (DTO + GraphQL). The output `shell` field becomes nullable (default `/bin/bash` retained).
- Temporary adapter bridge folds the input `command` into the str `start_command` form so the existing config validator wraps it (`shell -c`). Marked `FIXME`, to be removed once BA-6550 resolves `command` directly in the validator.
- Read-side exposure of `command` and the full round-trip test are deferred to a follow-up gated on BA-6550.

## Test plan
- [ ] CI `pants check`/`test` on `common` + `manager` (local mypy/test blocked by an empty pants `venv_cache`)
- [ ] `add_revision` / preset create with `command` only → revision persisted with resolved string `start_command`
- [ ] `add_revision` / preset create with deprecated `start_command` (list) only → still works (back-compat)
- [ ] GraphQL SDL: `command: String` on input+output; `start_command` carries `deprecation_reason`; output `shell` nullable

Resolves BA-6551

----
📚 Documentation preview 📚: https://sorna--12418.org.readthedocs.build/en/12418/

----
📚 Documentation preview 📚: https://sorna-ko--12418.org.readthedocs.build/ko/12418/

### DIFF
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -10319,11 +10319,14 @@ type ModelServiceConfig
   """
   preStartActions: [PreStartAction!]!
 
-  """Command to start the model service."""
-  startCommand: [String!]
+  """Single-string command to start the model service."""
+  command: String
+
+  """Deprecated. Command to start the model service. Use `command` instead."""
+  startCommand: [String!] @deprecated(reason: "Use `command` instead.")
 
   """Shell configured for the model service."""
-  shell: String!
+  shell: String
 
   """Port number for the model service."""
   port: Int!
@@ -10343,8 +10346,11 @@ input ModelServiceConfigInput
   """
   preStartActions: [PreStartActionInput!] = null
 
+  """Added in 26.7.0. Single-string command to start the model service."""
+  command: String = null
+
   """Command to start the model service."""
-  startCommand: [String!] = null
+  startCommand: [String!] = null @deprecated(reason: "Use `command` instead.")
 
   """Shell configured for the model service."""
   shell: String = null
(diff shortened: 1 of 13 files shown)


docs/manager/graphql-reference/supergraph.graphql
```diff
@@ -10319,11 +10319,14 @@ type ModelServiceConfig
   """
   preStartActions: [PreStartAction!]!
 
-  """Command to start the model service."""
-  startCommand: [String!]
+  """Single-string command to start the model service."""
+  command: String
+
+  """Deprecated. Command to start the model service. Use `command` instead."""
```
fregataa: Please find how to add "Deprecated since ..." here

tests/unit/manager/api/gql/deployment/test_revision_preset_types.py
```diff
@@ -197,11 +197,11 @@ def test_rejects_model_without_service(self) -> None:
...
-    def test_rejects_service_without_start_command(self) -> None:
-        # ``start_command`` is a required field; omitting it must be rejected.
-        payload: dict[str, Any] = {"port": 8080}
-        with pytest.raises(ValidationError):
-            PresetModelServiceConfigInput(**payload)
+    def test_accepts_service_with_command_only(self) -> None:
+        # ``start_command`` is now optional (deprecated); ``command`` replaces it.
+        service = PresetModelServiceConfigInput(port=8080, command="python server.py")
+        assert service.command == "python server.py"
+        assert service.start_command is None
```
fregataa, line 204: Good. How about adding a test that start_command still can be used only when command is not set and start_command is set?

docs/manager/graphql-reference/supergraph.graphql
```diff
@@ -10319,11 +10319,14 @@ type ModelServiceConfig
...
+  startCommand: [String!] @deprecated(reason: "Use `command` instead.")
```
jopemachine, line 10328: I think it would be better to explicitly add a note saying, `Do not use command and start_command together` (or that start_command will be ignored because command takes precedence).

jopemachine (approved):

fregataa (approved):

PR #12434, opened 2026-06-26
feat(BA-6600): define dry-run schedule action and input/output types

fregataa:
## Summary
- `data/session/dry_run.py`: `KernelDryRunSpec` (per-kernel `requested_slots` + `architecture`) and `KernelDryRunResult` (`spec` + `schedulable` + `required_reduction` — the minimum per-slot reduction to fit on the best-fitting node).
- `services/session/actions/dry_run_schedule.py`: `DryRunScheduleAction` (frozen, `EntityType.SESSION`, `GET`, carries `kernels` / `cluster_mode` / `resource_group_id`) and `DryRunScheduleActionResult`.

## Test plan
- [x] `pants check` / `pants test` pass in CI.

Resolves BA-6600

### DIFF
--- a/src/ai/backend/manager/services/session/actions/dry_run_schedule.py
+++ b/src/ai/backend/manager/services/session/actions/dry_run_schedule.py
@@ -1,0 +1,50 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import override
+
+from ai.backend.common.data.permission.types import EntityType
+from ai.backend.common.identifier.resource_group import ResourceGroupID
+from ai.backend.common.types import ClusterMode
+from ai.backend.manager.actions.action import BaseAction, BaseActionResult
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/services/session/actions/dry_run_schedule.py
```diff
@@ -0,0 +1,50 @@
...
+    @override
+    @classmethod
+    def operation_type(cls) -> ActionOperationType:
+        return ActionOperationType.GET
+
+
+@dataclass(frozen=True)
+class DryRunScheduleActionResult(BaseActionResult):
+    # Per-kernel results, in the same order as the action's ``kernels``.
+    kernels: list[KernelDryRunResult]
```
HyeockJinKim, line 46: Rather than the kernel, we need information such as which resources can be allocated where, or how scarce resources might be.

PR #12439, opened 2026-06-26
refactor(BA-6632): move image scanning out of the models layer into the image repository

fregataa:
## Summary
- Move the container-registry scan/untag logic out of `models/image/row.py` so the **models layer no longer imports the `container_registry` package**.
- `scan_single_image` / `rescan_images` (and the registry loading/filtering helpers) now live on `ImageDBSource`.
- The registry `untag` external call lives on `ImageRepository`, which fetches the image + registry row first, then calls the Harbor registry directly.
- `cli/image_impl` uses `ImageDBSource` for rescan and inlines the Harbor untag.

## Notes
- Pure relocation of existing logic — no new scanning behavior. The broader container-registry refactor (extracting external-access Client classes and deprecating `BaseContainerRegistry`) is tracked separately in BA-6634.

## Test plan
- [x] CI: `pants check` (mypy)
- [x] CI: `pants test` — image rescan / untag paths (repository, DB source, CLI) behave unchanged

Resolves BA-6632

### DIFF
--- a/changes/12439.enhance.md
+++ b/changes/12439.enhance.md
@@ -1,0 +1,1 @@
+Move container-registry image scanning out of `models/image/row.py` into the image repository layer (scan/rescan on the DB source, the registry untag on the repository) so the models layer no longer imports the `container_registry` package.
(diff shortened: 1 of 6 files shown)


jopemachine:
Strictly speaking, the rescan logic would be better placed on the client side. However, I don't think it necessarily needs to be addressed in this PR.

jopemachine (approved):

PR #12445, opened 2026-06-29
feat(BA-6550): represent model service command as string

seedspirit:
## Summary
- Refactor model service `start_command` to use a single internal command string while preserving deprecated list input compatibility.
- Keep `/bin/bash` as the default shell, allow explicit null/empty shell for direct argv execution, and update kernel-side command resolution accordingly.
- Add an idempotent Alembic data migration for stored model definitions in revisions, revision presets, and runtime variant defaults.

## Test plan
- [x] `pants fmt --changed-since=HEAD`
- [x] `pants fix --changed-since=HEAD`
- [x] `pants lint --changed-since=HEAD`
- [x] `pants check --changed-since=HEAD`
- [x] `pants test --changed-since=HEAD`
- [x] Push hook: lint/check/test on `origin/main..HEAD@603d0918f`
- [x] Live manager/web API verification for explicit `command`, deprecated list `start_command`, vLLM default model definition, and revision preset storage

Resolves BA-6550

----
📚 Documentation preview 📚: https://sorna--12445.org.readthedocs.build/en/12445/

----
📚 Documentation preview 📚: https://sorna-ko--12445.org.readthedocs.build/ko/12445/

### DIFF
--- a/src/ai/backend/common/config.py
+++ b/src/ai/backend/common/config.py
@@ -150,14 +150,16 @@ def _resolve_start_command(service: Any) -> Any:
     if not isinstance(service, dict):
         return service
 
-    result = {k: v for k, v in service.items() if k not in ("command", "start-command")}
+    result = {
+        k: v for k, v in service.items() if k not in ("command", "start_command", "start-command")
+    }
     # Override deprecated ``start_command`` with the new ``command`` when supplied.
     start_command = service.get("command")
     if start_command is None:
         start_command = service.get("start_command", service.get("start-command"))
     if isinstance(start_command, list):
         start_command = shlex.join(start_command)
-    if start_command is not None or "start-command" in service:
+    if start_command is not None or "start_command" in service or "start-command" in service:
         result["start_command"] = start_command
     return result
 
--- a/src/ai/backend/common/config.py
+++ b/src/ai/backend/common/config.py
@@ -20,6 +20,7 @@ from pydantic import (
 from . import validators as tx
 from .etcd import AsyncEtcd, ConfigScopes
 from .exception import BackendAIError, ConfigurationError, ModelDefinitionValidationError
+from .model_service_start_command_compat import resolve_model_service_start_command
 from .types import BackendAISchema, RedisHelperConfig, SchemaValidationFailureInfo
 
 __all__ = (
@@ -146,24 +147,6 @@ agent_selector_config_iv = t.Dict({}) | agent_selector_globalconfig_iv
 DEFAULT_SHELL = "/bin/bash"
 
 
-def _resolve_start_command(service: Any) -> Any:
-    if not isinstance(service, dict):
-        return service
-
-    result = {
-        k: v for k, v in service.items() if k not in ("command", "start_command", "start-command")
-    }
-    # Override deprecated ``start_command`` with the new ``command`` when supplied.
-    start_command = service.get("command")
-    if start_command is None:
-        start_command = service.get("start_command", service.get("start-command"))
-    if isinstance(start_command, list):
-        start_command = shlex.join(start_command)
-    if start_command is not None or "start_command" in service or "start-command" in service:
-        result["start_command"] = start_command
-    return result
-
-
 class PreStartAction(BaseConfigModel):
     action: str = Field(
         description="The name of the pre-start action to execute.",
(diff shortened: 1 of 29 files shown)


src/ai/backend/manager/api/adapters/deployment_revision_preset/adapter.py
```diff
@@ -143,10 +144,20 @@ def _model_health_check_to_dto(check: ModelHealthCheck) -> ModelHealthCheckInfoD
     )
 
 
+def _legacy_start_command_to_dto(start_command: str | None) -> list[str] | None:
+    if start_command is None:
+        return None
+    try:
+        return shlex.split(start_command)
+    except ValueError:
```
fregataa: In which case can ValueError be raised?

fregataa (commented):
This seems to be a feature job rather than an enhancement or refactoring since API schema changes.
And is this correct?
- API allows array typed `start_command` and `shell` is used only when `start_command` is set
- Agent and Kernel also allows `start_command` and `shell` is used only when `start_command` is set
- All Agent, Kernel and API uses `command` if it is set

seedspirit:
@fregataa Yes, that's correct with a few details:

- The list-form `start_command` is kept for backward compatibility, but it is deprecated and will be removed in the future.
- Since both the deprecated `start_command` and the new `command` can currently be set, `command` takes precedence when both are provided.
- `shell`, `start_command`, and `command` are independent configuration fields.
- When `shell` is set, the command string is executed as `shell -c command`.
- When `shell` is `null`, the command string is split with `shlex.split()` and executed directly as argv.

src/ai/backend/manager/api/adapters/deployment_revision_preset/adapter.py
seedspirit: `shlex.split()` can raise `ValueError` for malformed shell-like strings, such as an unclosed quote. This is a best-effort conversion from the new `command` string to the deprecated `start_command: list[str]`, so when splitting fails we preserve the original command as a single list item.

src/ai/backend/manager/api/adapters/deployment/adapter.py
```diff
@@ -338,6 +342,15 @@ def _tristate_from_input[T](value: T | Sentinel | None) -> TriState[T]:
...
 
 
+def _model_definition_to_dto(model_definition: ModelDefinition) -> ModelDefinitionInfoDTO:
+    data = model_definition.model_dump(by_alias=False)
+    for model in data.get("models") or []:
+        service = model.get("service")
+        if service is not None:
+            model["service"] = normalize_model_service_command_response(service)
+    return ModelDefinitionInfoDTO.model_validate(data)
+
```
fregataa: It doesn't need to be dumped into a dictionary here

src/ai/backend/common/model_service_start_command_compat.py
```diff
@@ -0,0 +1,44 @@
...
+    if command is None:
+        command = service.get("start_command", service.get("start-command"))
+        if isinstance(command, list) and all(isinstance(item, str) for item in command):
+            command = shlex.join(command)
+    if command is not None or "start_command" in service or "start-command" in service:
+        result["start_command"] = command
+    return result
+
+
+def normalize_model_service_command_response(service: Mapping[str, Any]) -> dict[str, Any]:
```
fregataa: It is better to set typed parameters to this function, rather than allowing a dictionary

src/ai/backend/common/config.py
```diff
@@ -280,8 +268,8 @@ class ModelServiceConfig(BaseConfigModel):
 
     @model_validator(mode="before")
     @classmethod
-    def _wrap_str_start_command(cls, data: Any) -> Any:
-        return _wrap_str_start_command_into_argv(data)
+    def _resolve_start_command(cls, data: Any) -> Any:
+        return resolve_model_service_start_command(data)
```
fregataa, line 272: How about changing this model_validator mode to after? then we don't need all `data` here

seedspirit: Since this also includes the process of converting currently deprecated input aliases (command, start-command, and list-type start_command) to start_command: str, it seems correct for it to run as a "before" action.

fregataa (approved):

PR #12448, opened 2026-06-29
feat(BA-6600): return per-requirement selection failures with suggestions

fregataa:
## Summary
- Make `select_agents_for_batch_requirements()` evaluate **every** requirement and return `BatchAgentSelectionResult` (successes + per-requirement failures) instead of aborting on the first unplaceable kernel. Real scheduling re-raises an aggregated `BatchAgentSelectionFailedError` to keep its all-or-nothing semantics.
- Add `build_suggestion() -> Suggestion` to every `AgentSelectionError` so a dry-run can surface structured remediation (reduce resource/container, change arch, revise designated agents) without parsing messages.
- Give `NoCompatibleAgentError` structured architecture context and add `InsufficientResourcesError.deficit()` (per-slot shortage); `NoAvailableAgentError` computes the node-axis MIN reduction.

## Test plan
- [x] `pants test tests/unit/manager/sokovan/scheduler::`
- [x] New `test_build_suggestion.py` covers deficit, arch/reduce/container/designated/none suggestions, node-axis MIN
- [x] Updated selector/provisioner/scheduler tests reflect the new return type

Resolves BA-6600


src/ai/backend/manager/sokovan/scheduler/provisioner/selectors/exceptions.py
```diff
@@ -60,6 +73,11 @@ def error_code(self) -> ErrorCode:
             error_detail=ErrorDetail.UNAVAILABLE,
         )
 
+    def build_suggestion(self) -> Suggestion:
+        # The resource group has no candidate agents at all; nothing the caller
+        # can adjust on the request itself fixes this.
+        return Suggestion(kind=SuggestionKind.NONE)
```
HyeockJinKim: Please add `@override`.

src/ai/backend/manager/sokovan/scheduler/provisioner/selectors/exceptions.py
```diff
@@ -101,6 +119,45 @@ def error_code(self) -> ErrorCode:
...
+                required_reduction=reduction,
+            )
+        if reduction is not None:
+            return Suggestion(
+                kind=SuggestionKind.REDUCE_RESOURCE,
+                required_reduction=reduction,
+            )
+        if any(isinstance(err, ContainerLimitExceededError) for err in self._agent_errors.values()):
+            # No node is short on slots; the blocker is the per-agent container limit.
+            return Suggestion(kind=SuggestionKind.REDUCE_CONTAINER)
```
HyeockJinKim: Instead of handling specific errors, should we use a specific base exception?

src/ai/backend/manager/sokovan/scheduler/provisioner/selectors/selector.py
```diff
@@ -115,6 +116,22 @@ class AgentSelection:
     selected_agent: AgentInfo
 
 
+@dataclass
+class RequirementSelectionFailure:
+    """A requirement that could not be placed, paired with its structured reason."""
+
+    resource_requirement: ResourceRequirements
+    error: AgentSelectionError
```
HyeockJinKim: Would it be better to include the exception itself in the dataclass?

src/ai/backend/manager/sokovan/scheduler/provisioner/selectors/exceptions.py
```diff
@@ -147,18 +204,75 @@ def _format_reason_lines(reasons: Iterable[str]) -> str:
...
+
+    Carries the structured architecture context so a dry-run can suggest which
+    architectures the request could target instead.
+    """
 
     error_type = "https://api.backend.ai/probs/no-compatible-agents"
     error_title = "No agents meet the resource requirements."
 
+    _required_architecture: str
+    _available_architectures: Sequence[str]
```
HyeockJinKim: It appears that the _required_architecture value is saved but not used later on.

src/ai/backend/manager/sokovan/scheduler/provisioner/selectors/selector.py
```diff
@@ -278,34 +295,44 @@ async def select_agents_for_batch_requirements(
...
+                    designated_agent_ids,
+                )
+            except (NoAvailableAgentError, NoCompatibleAgentError) as e:
+                failures.append(
+                    RequirementSelectionFailure(
+                        resource_requirement=resource_req,
+                        error=e,
+                    )
+                )
+                continue
```
HyeockJinKim, line 305: If you're only dealing with `resource_req` anyway, it seems like you should be able to tell which resource is in short supply without having to iterate through the data. What behavior were you intending?

fregataa: Each candidate node has different remaining slots, so the shortage (requested − remaining) differs per node — it can't be derived from `resource_req` alone. And when a kernel fails mid-batch, the per-node remaining shifts as earlier kernels' placements are already accumulated, so the excess for a later kernel can differ from an earlier one. We therefore iterate the candidate nodes to pick the one needing the smallest reduction.

jopemachine (approved):

PR #12449, opened 2026-06-30
fix(BA-6635): retry image rescan result commits

seedspirit:
## Summary
- Retry the container registry image rescan commit phase with the repository resilience retry decorator.
- Preserve pending image updates across retry attempts and emit progress only after a successful DB commit.
- Add a unit test covering serialization failure retry during rescan result commit.

## Test plan
- [x] pants fmt ::
- [x] pants fix ::
- [x] pants lint --changed-since=origin/main
- [x] pants fmt --changed-since=HEAD~1
- [x] pants fix --changed-since=HEAD~1
- [x] pants lint --changed-since=HEAD~1
- [x] pants check --changed-since=HEAD~1
- [x] pants test --changed-since=HEAD~1

Resolves BA-6635

### DIFF
--- a/src/ai/backend/manager/container_registry/base.py
+++ b/src/ai/backend/manager/container_registry/base.py
@@ -30,11 +30,14 @@ from ai.backend.common.docker import (
 )
 from ai.backend.common.docker import login as registry_login
 from ai.backend.common.exception import (
+    BackendAIError,
     InvalidImageName,
     InvalidImageTag,
     ProjectMismatchWithCanonical,
 )
 from ai.backend.common.json import read_json
+from ai.backend.common.resilience.policies.retry import BackoffStrategy, RetryArgs, RetryPolicy
+from ai.backend.common.resilience.resilience import Resilience
 from ai.backend.common.types import SlotName, SSLContextType
 from ai.backend.common.utils import join_non_empty
 from ai.backend.logging import BraceStyleAdapter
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/container_registry/base.py
```diff
@@ -176,97 +191,121 @@ def _determine_additional_image_scopes(
...
+                if (reporter := progress_reporter.get()) is not None:
+                    await reporter.update(1, message=progress_msg)
+        return scanned_images
+
+    @commit_rescan_result_resilience.apply()
+    async def _commit_rescan_result_once(
+        self,
+        original_updates: dict[ImageIdentifier, dict[str, Any]],
+        image_identifiers: list[tuple[str, str]],
+    ) -> tuple[list[ImageData], list[tuple[str, str]]]:
```
jopemachine: I think it'd be better to define a new dedicated type instead of using a complex composite type.

src/ai/backend/manager/container_registry/base.py
```diff
@@ -176,97 +191,121 @@ def _determine_additional_image_scopes(
...
+        scanned_images: list[ImageData] = []
+        progress_events: list[tuple[str, str]] = []
```
jopemachine: Aggregating the logs doesn't seem like a good approach.

Wouldn't it be better to handle logging and reporter.update separately?

Even if we decide to aggregate logs, collecting raw strings such as "warning" and "info" and then dispatching them through a switch statement seems quite fragile.

seedspirit: Since retry logic has been implemented, it seems more appropriate to aggregate the logs rather than log them each time, and display them all at once when the operation succeeds. I’ve changed "warning" and "info" to enums for now.

seedspirit:
Resolved by https://github.com/lablup/backend.ai/pull/12464

PR #12453, opened 2026-06-30
test: register minimal ORM cluster so isolated manager tests configure mappers

fregataa:
## Summary
- Manager unit tests run as isolated Pants targets that import only a subset of the ORM graph. SQLAlchemy's `configure_mappers()` then fails to resolve string-based relationships whose targets are imported only under `TYPE_CHECKING` (e.g. `DomainRow -> "ScalingGroupForDomainRow"`), so any test that instantiates or queries a Row errors at setup.
- Each affected isolated test now imports the **minimal set of ORM rows** needed to close the string-relationship graph that its own imports pull in — typically just `AgentRow` plus one `ScalingGroupFor*Row` (the routing test also needs `EndpointRow`). Registering those rows lets `configure_mappers()` succeed even in an isolated target. The rows are held live in a module-level `_ORM_CLUSTER` tuple.
- The set is minimal by construction and verified per file: every retained row is load-bearing (removing it makes the isolated target fail), and tests that never trigger mapper configuration (pure-value or fully mocked) need no cluster at all.
- `conftest.py`: `domain_factory` imports the `domains` table lazily so `DomainRow` is not registered on the global registry for tests that never touch a domain (`pyproject.toml` exempts the lazy import from `PLC0415`).

## Test plan
- [x] Every affected manager unit test passes when run as an isolated Pants target (the original failure condition)
- [x] `pants lint` / `fmt` / `fix` / visibility / mypy on changed files pass


HyeockJinKim (changes requested):
Don't import models in testutils package

tests/unit/manager/repositories/runtime_variant_preset/test_repository.py
```diff
@@ -11,9 +11,73 @@
...
+from ai.backend.manager.models.endpoint import (
+    EndpointAutoScalingRuleRow,
+    EndpointRow,
+    EndpointTokenRow,
+)
+from ai.backend.manager.models.fair_share import (
+    DomainFairShareRow,
+    ProjectFairShareRow,
+    UserFairShareRow,
+)
```
HyeockJinKim: The "fair share" value does not appear to be necessary in the runtime variant preset.

HyeockJinKim (approved):

PR #12455, opened 2026-06-30
fix(BA-6645): re-trigger image pulling for sessions stuck in PREPARING

jopemachine:
## Summary

Re-trigger the idempotent image pull (`check_and_pull`) for sessions stuck in `PREPARING` by widening `CheckPreconditionLifecycleHandler`'s query to cover them.

## Background

When a session transitions `SCHEDULED -> PREPARING`, `CheckPreconditionLifecycleHandler` sends `check_and_pull` exactly once. The handler previously targeted only `SCHEDULED` sessions with `SCHEDULED` kernels; once a session moves to `PREPARING` (kernels typically advancing to `PULLING`) it is no longer returned by `get_sessions_for_handler`, so `check_and_pull` is never re-sent.

If the resulting `ImagePullFinished` (or an intermediate kernel lifecycle) event is lost in transit, the session can remain stuck in `PREPARING` — there is no active reconciliation for the `PREPARING`/`PULLING` zone (unlike `RUNNING`, covered by `SWEEP_STALE_KERNELS`). With this change the handler re-queries such sessions on subsequent scheduler ticks and re-triggers the idempotent pull; when the image already exists the agent re-emits `ImagePullFinished("Image already exists")`, unsticking the session.

## Changes

- `target_statuses`: `[SCHEDULED]` -> `[SCHEDULED, PREPARING]`.
- `target_kernel_statuses`: `[SCHEDULED]` -> `[SCHEDULED, PULLING]` (so a `PREPARING` session whose kernels already advanced to `PULLING` is actually matched by the `session-status AND has-matching-kernel` filter).

## Notes for review

- The handler's success transition re-applies `session/kernel = PREPARING`; please confirm re-processing an in-progress `PREPARING` session does not regress kernels that legitimately advanced (e.g. to `PREPARED`).
- `check_and_pull` is expected to be idempotent; re-triggering for healthy in-progress sessions should be a no-op.

## Related

- BA-6645

### DIFF
--- a/src/ai/backend/manager/sokovan/scheduler/handlers/lifecycle/check_precondition.py
+++ b/src/ai/backend/manager/sokovan/scheduler/handlers/lifecycle/check_precondition.py
@@ -65,8 +65,14 @@ class CheckPreconditionLifecycleHandler(SessionLifecycleHandler):
     @classmethod
     @override
     def target_kernel_statuses(cls) -> list[KernelStatus] | None:
-        """Include sessions where kernels are in SCHEDULED status."""
-        return [KernelStatus.SCHEDULED]
+        """Include sessions with kernels in SCHEDULED or PULLING status.
+
+        PULLING is included so that a session stuck in PREPARING whose
+        kernels have already advanced to PULLING (but whose pull-completion
+        event was lost) is still picked up and has its image pull
+        re-triggered.
+        """
+        return [KernelStatus.SCHEDULED, KernelStatus.PULLING]
 
     @classmethod
     @override
--- a/src/ai/backend/manager/sokovan/scheduler/handlers/lifecycle/check_precondition.py
+++ b/src/ai/backend/manager/sokovan/scheduler/handlers/lifecycle/check_precondition.py
@@ -145,17 +145,22 @@ class CheckPreconditionLifecycleHandler(SessionLifecycleHandler):
             sessions_for_pull_data.image_configs,
         )
 
-        # Mark all sessions as success for status transition
+        # Report re-triggered PREPARING sessions as skipped so the coordinator
+        # does not re-apply the PREPARING transition (avoids racing against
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/sokovan/scheduler/handlers/lifecycle/check_precondition.py
```diff
@@ -52,14 +52,26 @@ def name(cls) -> str:
...
     def target_kernel_statuses(cls) -> list[KernelStatus] | None:
-        """Include sessions where kernels are in SCHEDULED status."""
-        return [KernelStatus.SCHEDULED]
+        """Include sessions with kernels in SCHEDULED or PULLING status.
+
+        PULLING covers kernels whose pull already finished on the agent but
+        were never updated to PREPARED because the ImagePullFinished event
+        was lost; re-triggering makes the agent re-emit the completion event.
+        """
+        return [KernelStatus.SCHEDULED, KernelStatus.PULLING]
```
HyeockJinKim: "PREPARING" is missing.

src/ai/backend/manager/sokovan/scheduler/handlers/lifecycle/check_precondition.py
```diff
@@ -132,17 +144,22 @@ async def execute(
...
+                session_id=session_info.identity.id,
+                from_status=from_status,
+                reason="passed-preconditions",
+                creation_id=session_info.identity.creation_id,
+                access_key=AccessKey(session_info.metadata.access_key),
             )
+            if from_status == SessionStatus.PREPARING:
+                result.skipped.append(transition_info)
+            else:
+                result.successes.append(transition_info)
```
HyeockJinKim, line 164: I don't think the `from_status` value is important; rather, the response from the agent is what matters.

HyeockJinKim (approved):

PR #12457, opened 2026-07-01
feat(BA-2325): Allow v2 client registries to borrow a shared aiohttp session

### LINKED ISSUE #5820: Add connection pooling support to synchronous API session

rapsealk:
Resolves #5820 (BA-2325). Supersedes #5841.

## Problem

The v2 async client opens **and closes** a fresh `aiohttp.ClientSession` for every registry: `BackendAIClientRegistry.create` builds a session (`base_client.py:86`) and `registry.close()` tears it down (`base_client.py:98`). Callers that make repeated Manager calls — especially synchronous callers wrapping the client in `async_to_sync`, and any long-lived service — therefore pay a new TCP + TLS handshake on every request and cannot pool connections.

(The original BA-2325 framing — pooling the *synchronous* `Session` — was a dead end: that path is CLI-only and has no long-lived consumer to pool. See #5841.)

## Change

A minimal session-injection seam on the v2 async client:

- `BackendAIAuthClient` / `BackendAIAnonymousClient` gain `owns_session` (default `True`). `close()` closes the session only when owned.
- `BackendAIClientRegistry.create` / `V2ClientRegistry.create` accept an optional `session=`. When given, the clients **borrow** it (`owns_session=False`), so `close()` leaves the shared pool open. When omitted, behavior is unchanged (each client creates and owns its own session).

Because the aiohttp session is auth-agnostic transport (HMAC signing is per-request, `base_client.py:112`), a single shared session can serve **all** credentials against a given endpoint — one connection pool, not one-per-keypair.

## Caller pattern (out of scope here, for downstream)

Callers own the pool lifecycle. Recommended: one shared session **per event loop** (keyed by the running loop) so sessions created under `async_to_sync` are never used across loops, then pass it as `session=` and drop the per-call `registry.close()`; close the pool on worker shutdown.

## Tests

- `close()` closes an owned session but leaves an injected one open (auth + anon clients).
- `create(..., session=shared)` shares one session across both clients and preserves it across `registry.close()` (both registries).

## Checklist
- [x] News fragment (`changes/5820.enhance.md`)
- [x] `pants fmt/fix/lint/check/test` pass on changed files
- [ ] Draft — downstream per-loop pool wiring to follow separately

----
📚 Documentation preview 📚: https://sorna--12457.org.readthedocs.build/en/12457/

----
📚 Documentation preview 📚: https://sorna-ko--12457.org.readthedocs.build/ko/12457/

### DIFF
--- a/changes/5820.enhance.md
+++ b/changes/5820.enhance.md
@@ -1,0 +1,1 @@
+Allow the v2 async client registries (`BackendAIClientRegistry`, `V2ClientRegistry`) to borrow a caller-owned `aiohttp.ClientSession` so repeated Manager calls can reuse a shared connection pool; `close()` leaves an injected session open.
--- a/src/ai/backend/client/v2/base_client.py
+++ b/src/ai/backend/client/v2/base_client.py
@@ -61,21 +61,30 @@ class BackendAIAuthClient:
 
     Prefer ``create()`` for production use; ``__init__`` accepts a pre-built
     session so tests can inject a mock directly.
+
+    Pass ``owns_session=False`` when injecting a caller-owned session (e.g. a
+    shared connection pool): ``close()`` will then leave the session open so the
+    caller can keep reusing it. Sessions built by ``create()`` are owned and
+    closed normally.
     """
 
     _config: ClientConfig
     _auth: AuthStrategy
     _session: aiohttp.ClientSession
+    _owns_session: bool
 
     def __init__(
         self,
         config: ClientConfig,
         auth: AuthStrategy,
         session: aiohttp.ClientSession,
+        *,
+        owns_session: bool = True,
     ) -> None:
         self._config = config
         self._auth = auth
         self._session = session
+        self._owns_session = owns_session
 
     @classmethod
     async def create(
@@ -95,7 +104,8 @@ class BackendAIAuthClient:
         return self._config
 
     async def close(self) -> None:
-        await self._session.close()
+        if self._owns_session:
+            await self._session.close()
 
     def _build_url(self, path: str) -> str:
         base = str(self._config.endpoint).rstrip("/")
@@ -378,14 +388,18 @@ class BackendAIAnonymousClient:
 
     _config: ClientConfig
     _session: aiohttp.ClientSession
+    _owns_session: bool
 
     def __init__(
         self,
         config: ClientConfig,
         session: aiohttp.ClientSession,
+        *,
+        owns_session: bool = True,
     ) -> None:
         self._config = config
         self._session = session
+        self._owns_session = owns_session
 
     @classmethod
     async def create(
@@ -396,7 +410,8 @@ class BackendAIAnonymousClient:
         return cls(config, session)
 
     async def close(self) -> None:
-        await self._session.close()
+        if self._owns_session:
+            await self._session.close()
 
     def _build_url(self, path: str) -> str:
         base = str(self._config.endpoint).rstrip("/")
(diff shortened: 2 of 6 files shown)


rapsealk:
### Note for the downstream per-loop pool consumer (follow-up, out of scope here)

This PR only adds the **seam** (`session=` injection + non-owning `close()`). The actual pooling win lands in the caller, which owns the shared session's lifecycle. Recommended pattern — one session **per event loop**, keyed by the running loop, so sessions created under `async_to_sync` are never used across loops:

```python
import asyncio
import aiohttp
from ai.backend.client.v2.auth import HMACAuth
from ai.backend.client.v2.base_client import _create_aiohttp_session  # or a local factory
from ai.backend.client.v2.registry import BackendAIClientRegistry

_shared: dict[asyncio.AbstractEventLoop, aiohttp.ClientSession] = {}

def _shared_session(config) -> aiohttp.ClientSession:
    loop = asyncio.get_running_loop()
    sess = _shared.get(loop)
    if sess is not None and not sess.closed:
        return sess
    sess = _create_aiohttp_session(config)   # bound to *this* loop
    _shared[loop] = sess
    return sess

async def call(config, access_key, secret_key):
    auth = HMACAuth(access_key=access_key, secret_key=secret_key)
    registry = await BackendAIClientRegistry.create(
        config, auth, session=_shared_session(config),
    )
    # NO registry.close() — the shared pool lives on
    return await registry.<domain>.<op>(...)
```

Checklist for whoever wires this up:

- **Loop affinity is mandatory.** Key the cache by the running loop (as above). Do **not** cache a single global session — `async_to_sync` may run calls on different loops, and an aiohttp session is not safe across loops.
- **Drop the per-call `registry.close()`** at the `async_to_sync` call sites — the injected session is non-owning now, so `close()` is a no-op on it, but removing it makes the intent explicit.
- **One pool per endpoint is enough.** The session is auth-agnostic transport (HMAC signing is per-request), so a single shared session serves all keypairs against a given endpoint. Don't key the pool by access key.
- **Close on shutdown.** Register a worker-teardown hook that closes every session in the cache (avoids `ResourceWarning`).
- **`V2ClientRegistry` takes the same `session=`** for the `/v2/` GQL/REST path.

A precedent for the per-loop pattern already exists in the downstream proxy layer (a `dict[event_loop, aiohttp.ClientSession]` guarded by `not session.closed`); reuse that shape rather than inventing a new one.

rapsealk:
Closing as not planned.

This was pooling infrastructure ahead of a validated consumer — YAGNI. Two reasons:

1. **Premise unmeasured.** No data that per-call TCP/TLS handshakes to the Manager are material (QPS unknown; the internal hop may not even be TLS).
2. **Documented consumer likely doesn't work.** The per-loop pool pattern (from [this earlier comment](https://github.com/lablup/backend.ai/pull/12457#issuecomment-4850757352)) comes from the proxy's *persistent* ASGI loop. The target `async_to_sync` sites run in a *sync* context where asgiref generally creates and closes a fresh loop per call — so the pool would never hit, and would leak sessions bound to dead loops. **I retract that comment's pattern.**

If pooling is ever warranted, it needs a measurement first and a design matching the sync runtime (e.g. the existing `ResilientSession`/urllib3 path, or a dedicated persistent-loop thread), not this seam. Tracked context in #5820 (also closed).

PR #9946, opened 2026-03-12
fix(BA-5033): register AssociationScopesEntitiesRow table in image CLI rescan

jopemachine:
Resolves #9937 (BA-5033).

## Summary
- Fix `'NoneType' object has no attribute 'set'` error during image rescan in CLI paths
- Call `ensure_all_tables_registered()` in `connect_database()` to ensure all model tables are imported before the first ORM query triggers `configure_mappers()`
- Lazy imports in the container registry factory caused models like `AssociationScopesEntitiesRow` to be loaded after `configure_mappers()` had already run, leaving their mapper attributes uninitialized

## Test plan
- [x] Run `mgr image rescan cr.backend.ai` and verify image scan completes without errors

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12468, opened 2026-07-02
fix(BA-6653): register image CLI ORM tables needed for mapper initialization

### LINKED ISSUE #12466: CLI image rescan/alias fails with mapper init error due to unregistered ORM tables
## Objective

The manager CLI image rescan (and alias) path does not go through the normal server bootstrap, so only a subset of ORM Row classes get imported. When SQLAlchemy configures its mappers on the first ORM query, string-based relationships fail to resolve, raising InvalidRequestError. This surfaces during install-dev.sh at the image registry scan step.

```
When initializing mapper Mapper[GroupRow(groups)], expression 'ScalingGroupForProjectRow' failed to locate a name ('ScalingGroupForProjectRow').
```

## Background

Same class of failure was previously patched in BA-5033 (#9946) by manually importing AssociationScopesEntitiesRow. That one-off approach does not scale: a newly added GroupRow -> ScalingGroupForProjectRow relationship now triggers the same error. The proper fix is to register all ORM tables up front via ensure_all_tables_registered() (added in BA-3644), replacing the single-class manual import in both rescan_images() and alias().

## Acceptance Criteria

- ./backend.ai mgr image rescan and image alias run without mapper InvalidRequestError.
- install-dev.sh completes the image registry scan step cleanly.
- The single-class manual import workaround is replaced with ensure_all_tables_registered().

JIRA Issue: BA-6653

jopemachine:
Resolves #12466 (BA-6653)

## Summary
- The image CLI commands (`list`, `inspect`, `forget`, `purge`, `set-resource-limit`, `rescan`, `alias`, `dealias`, `validate-image-alias`, `validate-image-canonical`) do not go through the normal server bootstrap, so ORM `Row` classes referenced only via string-based relationships are never imported. SQLAlchemy then fails to configure its mappers on each command's first ORM query, e.g. `Mapper[GroupRow(groups)]` cannot locate `'ScalingGroupForProjectRow'`. This surfaced during `install-dev.sh` at the image registry scan step.
- Extends the earlier BA-5033 fix (#9946), which guarded only `rescan` with `AssociationScopesEntitiesRow`. A `GroupRow -> ScalingGroupForProjectRow` relationship now goes unresolved on this path, and it affects every image subcommand that issues an ORM query — not just `rescan`.
- Registers the minimal ORM cluster (`AgentRow`, `AssociationScopesEntitiesRow`, `ScalingGroupForProjectRow`) inside each command handler, following the existing `_ =`/ORM-cluster registration pattern (cf. #12453).

## Verification
Confirmed against the live CLI on the current tree (before the fix), all failing with the identical `ScalingGroupForProjectRow` mapper error:
- `mgr image list` -> fails at `ImageRow.list`
- `mgr image inspect <name> <arch>` -> fails
- `mgr image dealias <alias>` -> fails
- `mgr image validate-image-alias <alias>` -> fails

After the fix, `configure_mappers()` succeeds once a handler's cluster imports run before its first query.

## Why no 26.4 backport is needed
This is a **main-only regression introduced by the import-graph change in BA-6632** ("move image scanning out of the models layer into the image repository"), not a defect in the ORM relationship itself.

The same `GroupRow -> "ScalingGroupForProjectRow"` relationship exists on both branches, but `image_impl` reaches the image scan through different imports:

| Branch | rescan import in `image_impl` | Effect on mapper config |
| --- | --- | --- |
| **26.4** | `from ...models.image import rescan_images` | that chain transitively loads `scaling_group`, so `ScalingGroupForProjectRow` is registered before `GroupRow` is configured -> OK |
| **main** | `from ...repositories.image.db_source import ImageDBSource` | that chain no longer loads `scaling_group.row`, so `GroupRow` configuration cannot resolve the string target -> fails |

Verified on a live 26.4 environment (v26.4.6): `mgr image list` runs cleanly, `configure_mappers()` succeeds, and `GroupRow`'s `sgroup_for_groups_rows` relationship resolves. The bug does not reproduce there, so no backport to 26.4 is required.

## Notes
- Deliberately does **not** use `ensure_all_tables_registered()`. Measured cost of importing every model on this CLI path is ~0.2-0.3s of extra startup per invocation (register+configure roughly doubles), and it does not reduce fragility enough to justify that. The minimal set was determined empirically by iteratively resolving each unresolved relationship name until `configure_mappers()` succeeds.

## Test plan
- [x] Live CLI: the four commands above fail before the fix, succeed (past mapper init) after
- [x] `configure_mappers()` succeeds in a fresh process once the cluster imports run before the first query
- [x] Confirmed on live 26.4 (v26.4.6) that the bug does not reproduce -> no backport needed
- [x] `pants fmt` / `pants lint` / `pants check` pass on the changed file

### DIFF
--- a/src/ai/backend/manager/cli/image_impl.py
+++ b/src/ai/backend/manager/cli/image_impl.py
@@ -27,7 +27,19 @@ from .context import CLIContext, redis_ctx
 log = BraceStyleAdapter(logging.getLogger(__spec__.name))
 
 
+def _register_image_cli_orm_cluster() -> None:
+    """Register ORM rows reachable only via string relationships so the CLI can configure mappers (kept minimal, not all models)."""
+    from ai.backend.manager.models.agent.row import AgentRow
+    from ai.backend.manager.models.rbac_models.association_scopes_entities import (
+        AssociationScopesEntitiesRow,
+    )
+    from ai.backend.manager.models.scaling_group.row import ScalingGroupForProjectRow
+
+    _ = (AgentRow, AssociationScopesEntitiesRow, ScalingGroupForProjectRow)
+
+
 async def list_images(cli_ctx: CLIContext, short: bool, installed_only: bool) -> None:
+    _register_image_cli_orm_cluster()
     # Connect to postgreSQL DB
     bootstrap_config = await cli_ctx.get_bootstrap_config()
     async with (
@@ -80,6 +92,7 @@ async def list_images(cli_ctx: CLIContext, short: bool, installed_only: bool) ->
 
 
 async def inspect_image(cli_ctx: CLIContext, canonical_or_alias: str, architecture: str) -> None:
+    _register_image_cli_orm_cluster()
     bootstrap_config = await cli_ctx.get_bootstrap_config()
     async with (
         connect_database(bootstrap_config.db) as db,
@@ -101,6 +114,7 @@ async def inspect_image(cli_ctx: CLIContext, canonical_or_alias: str, architectu
 
 
 async def forget_image(cli_ctx: CLIContext, canonical_or_alias: str, architecture: str) -> None:
+    _register_image_cli_orm_cluster()
     bootstrap_config = await cli_ctx.get_bootstrap_config()
     async with (
         connect_database(bootstrap_config.db) as db,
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/cli/image_impl.py
```diff
@@ -27,7 +27,34 @@
...
+    stay unregistered and mapper initialization fails, e.g.::
+
+        When initializing mapper Mapper[GroupRow(groups)], expression
+        'ScalingGroupForProjectRow' failed to locate a name.
+
+    We import only the specific rows required to close the relationship graph rather than
+    every model on purpose: importing all models would add ~0.2-0.3s to each CLI
+    invocation (see the discussion on PR #9946). Add a row here if a new relationship
+    makes mapper configuration fail on the image CLI path.
+    """
```
seedspirit: nitpick: I think it would be best to keep only the important parts of the comments and tidy them up.

HyeockJinKim (approved):

fregataa (approved):

PR #12481, opened 2026-07-02
fix(BA-6646): recover sessions stuck in CREATING by re-sending idempotent create_kernel

jopemachine:
## Summary

- A session could remain stuck in `CREATING` indefinitely when the `KernelStarted` gate event was lost in transit, since nothing re-queried the CREATING zone (same class of gap as BA-6645 / #12455 for PREPARING).
- Make the agent's `create_kernel` idempotent for non-restarting calls: if the kernel is already running, it re-emits the `KernelStarted` anycast/broadcast events (reconstructing the creation info from the kernel registry) instead of creating a duplicate container; if a creation is already in flight (or the kernel exists in the registry but has not reached RUNNING yet), the call is skipped and returns `None` instead of raising `ResourceError`.
- Include `CREATING` in `StartSessionsLifecycleHandler.target_statuses()` / `target_kernel_statuses()` so the periodic START sweep re-sends the now-idempotent `create_kernels` RPC for stuck sessions, recovering them without manual intervention.

Note: the Jira issue suggested a read-only reconciliation via `check_creating`/`check_running` RPCs. This PR instead re-sends `create_kernels` after making it idempotent, which reuses the original event path (including the full `creation_info` payload) and also covers the manager-side `WHERE status IN (PREPARED, CREATING)` guard for duplicate events.

## Test plan

- [x] `tests/unit/agent/test_create_kernel_idempotency.py`: re-emit for a RUNNING kernel, skip for a registered-but-not-running kernel, skip (no raise) for an in-flight creation
- [x] `test_lifecycle_handlers.py` SC-ST-006: `CREATING` included in the handler's target statuses
- [x] `pants fmt/fix/lint/check` and affected unit tests pass locally
- [ ] Live verification: kill the KernelStarted event delivery and confirm a CREATING-stuck session recovers within one sweep interval

Resolves BA-6646

### DIFF
--- a/changes/12481.fix.md
+++ b/changes/12481.fix.md
@@ -1,0 +1,1 @@
+Recover sessions stuck in `CREATING` due to a lost `KernelStarted` event by making the agent's `create_kernel` idempotent (re-emitting the event for an already-running kernel and skipping in-flight creations) and re-sending it from the periodic session-start sweep.
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -2522,6 +2522,57 @@ class AbstractAgent[
             service.start_command = [*service.start_command, *extra_args]
         return models
 
+    async def _replay_kernel_started_event(
+        self,
+        kernel_obj: AbstractKernel,
+        kernel_config: KernelCreationConfig,
+    ) -> KernelCreationResult:
+        """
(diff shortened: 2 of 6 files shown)


jopemachine:
This work also needs to make create_kernel idempotent on the agent side, so it makes sense to address it as part of the agent refactoring.

PR #12495, opened 2026-07-02
refactor(BA-6676): separate vfolder-mount resolution from spec-context fetch

fregataa:
## Summary
- Split vfolder-mount resolution out of `SchedulerRepository.fetch_session_spec_contexts` into a dedicated `resolve_vfolder_mounts_by_role` method.
- `SessionSpecContextFetch` no longer carries `vfolder_mounts_by_role`; the fetch signature drops the vfolder-only `storage_manager` / `allowed_vfolder_types` args.
- The enqueue path resolves vfolder mounts in a separate call and merges the result into the preparation context — behavior preserved.
- Lets resource-only callers (e.g. a scheduler dry-run) run the preparer chain without the storage-manager RPC and etcd vfolder-types lookup (the slowest part of the fetch).

## Test plan
- [x] `pants fmt/fix/lint/check`
- [x] `test_enqueue_session_from_draft`, `test_automount_vfolder_resolution`, `test_owner_resource_group_access`

Resolves BA-6676

### DIFF
--- a/src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
@@ -1509,11 +1509,8 @@ class ScheduleDBSource:
     async def fetch_session_spec_contexts(
         self,
         draft: SessionSpecDraft,
-        *,
-        storage_manager: StorageSessionManager,
-        allowed_vfolder_types: list[str],
     ) -> SessionSpecContextFetch:
-        """Batch-fetch every DB read the draft-based preparer and validator
+        """Batch-fetch the DB reads the draft-based preparer and validator
         rules depend on, inside a single readonly transaction.
 
         Returns raw fetched data as :class:`SessionSpecContextFetch` — the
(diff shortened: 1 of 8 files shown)


src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
```diff
@@ -1706,14 +1658,91 @@ async def fetch_session_spec_contexts(
...
+        """
+        user_uuid = draft.identity.user_uuid
+        domain_name = str(draft.scope.domain_name) if draft.scope.domain_name else None
+        project_id = draft.scope.project_id
+        access_key = draft.identity.access_key
+        kernel_specs = tuple(draft.options.kernel_groups or ())
+
+        vfolder_mounts_by_role: dict[str, tuple[VFolderMount, ...]] = {}
+        if domain_name is None or user_uuid is None:
+            return vfolder_mounts_by_role
```
seedspirit, line 1692: Is it okay to just return an empty dict without throwing an error?

fregataa: This preserves the existing behavior: the previous `fetch_session_spec_contexts` guarded the vfolder loop with `if domain_name is not None and user_uuid is not None`, so a null domain/user already yielded an empty `{}` (no error, `_fetch_vfolder_mounts` not called). In a normal enqueue both are always set, so this is just the defensive fallback. Raising here would be a behavior change, out of scope for this refactor.

seedspirit (approved):

PR #12507, opened 2026-07-02
fix(BA-6678): add pool_pre_ping/pool_recycle to appproxy-coordinator DB engine

Yaminyam:
## Summary
The appproxy-coordinator SQLAlchemy engine is created **without** `pool_pre_ping` (and without `pool_recycle`), unlike the manager and account-manager engines. This ports the robust-connection handling the manager received in #1991 to the coordinator.

## Problem
When a pooled Postgres connection is dropped underneath SQLAlchemy (server-side termination, failover, network blip, or `idle_in_transaction_session_timeout`), asyncpg transparently reconnects the socket with a clean no-transaction state, but SQLAlchemy's pooled connection keeps stale transaction state. The next query is then issued as a nested SAVEPOINT, which the fresh connection rejects:

```
asyncpg.exceptions.NoActiveSQLTransactionError: SAVEPOINT can only be used in transaction blocks
```

Without `pool_pre_ping` the poisoned connection is never revalidated at checkout, so it stays in the pool and every subsequent request reusing it fails the same way — the coordinator is effectively down until a process restart, with no automatic recovery. Observed as a ~24h outage in a production deployment where every `PATCH /api/worker/{id}` heartbeat returned 500, causing workers to be marked stale and app/SSH endpoint creation to fail with `AppLaunchError: Worker not available`.

The first error is already the SAVEPOINT error with **no preceding SQLSTATE 40001 serialization failure** — the trigger is a dropped connection, not a serialization conflict. (`pgsql_connect_opts` sets `idle_in_transaction_session_timeout = 60s`, one concrete way a pooled connection gets terminated server-side.)

## Fix
- Add `pool_recycle` and `pool_pre_ping` fields to appproxy-coordinator `DBConfig` (previously only `pool_size` / `max_overflow` existed).
- Pass both to `create_async_engine`, mirroring manager (#1991) and account-manager.
- `pool_pre_ping` defaults to `True` here so the resilience is on by default. (manager/account-manager default it to `False`; consider flipping those or enabling in deployment config in a follow-up so they benefit too.)

## Notes
- Secondary hardening (not in this PR): `execute_with_txn_retry` reuses one caller-provided connection across retry attempts, so a poisoned connection cannot recover mid-loop. Running each attempt on a fresh transaction would harden the retry path further. Aggravator, not root cause.

Jira: BA-6678

### DIFF
--- a/src/ai/backend/appproxy/coordinator/config.py
+++ b/src/ai/backend/appproxy/coordinator/config.py
@@ -29,7 +29,12 @@ from ai.backend.common.configs import (
     PyroscopeConfig,
     ServiceDiscoveryConfig,
 )
-from ai.backend.common.meta import BackendAIConfigMeta, CompositeType, ConfigExample
+from ai.backend.common.meta import (
+    NEXT_RELEASE_VERSION,
+    BackendAIConfigMeta,
+    CompositeType,
+    ConfigExample,
+)
 from ai.backend.common.types import ServiceDiscoveryType
 from ai.backend.logging import LogLevel
 from ai.backend.logging.config import LoggingConfig
@@ -174,7 +179,7 @@ class DBConfig(BaseSchema):
                 "Useful for handling database connections closed by the server after inactivity "
                 "or by network equipment with idle timeouts."
             ),
-            added_version="26.5.0",
+            added_version=NEXT_RELEASE_VERSION,
             example=ConfigExample(local="-1", prod="3600"),
         ),
     ]
@@ -189,7 +194,7 @@ class DBConfig(BaseSchema):
                 "used in transaction blocks' errors after a Postgres connection drop or failover. "
                 "Adds a small overhead per checkout but is recommended for production."
             ),
-            added_version="26.5.0",
+            added_version=NEXT_RELEASE_VERSION,
             example=ConfigExample(local="true", prod="true"),
         ),
     ]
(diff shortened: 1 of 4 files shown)


src/ai/backend/appproxy/coordinator/config.py
```diff
@@ -164,6 +164,35 @@ class DBConfig(BaseSchema):
...
+        float,
+        Field(default=-1, ge=-1),
+        BackendAIConfigMeta(
+            description=(
+                "Maximum lifetime of a connection in seconds before it's recycled. "
+                "Set to -1 to disable connection recycling. "
+                "Useful for handling database connections closed by the server after inactivity "
+                "or by network equipment with idle timeouts."
+            ),
+            added_version="26.5.0",
```
fregataa: The version annotation is wrong. Please use a `NEXT_RELEASE_VERSION` constant

src/ai/backend/appproxy/coordinator/config.py
```diff
@@ -164,6 +164,35 @@ class DBConfig(BaseSchema):
...
+        Field(default=True),
+        BackendAIConfigMeta(
+            description=(
+                "Whether to test connections with a lightweight ping before using them. "
+                "Detects stale or disconnected connections at pool checkout and transparently "
+                "reconnects them with clean transaction state, preventing 'SAVEPOINT can only be "
+                "used in transaction blocks' errors after a Postgres connection drop or failover. "
+                "Adds a small overhead per checkout but is recommended for production."
+            ),
+            added_version="26.5.0",
```
fregataa: Same here

fregataa (changes requested):

fregataa (approved):

HyeockJinKim (approved):

PR #12517, opened 2026-07-03
feat(BA-6701): expose allow-list rank on the v2 API surface

jopemachine:
## Summary

Exposes the allow-list `rank` (introduced in #12516) on the v2 API surface. Pure API-layer change — no schema or behavior change beyond the new field:

- **DTO** (`common/dto/manager/v2/app_config_allow_list`): optional `rank` on `CreateAppConfigAllowListInput` (defaults to the scope type's default rank when omitted), required `rank` on `AppConfigAllowListNode`, and a `RANK` order field.
- **Adapter**: maps `rank` between the DTO and the creator spec / domain data, and handles the `RANK` ordering.
- **GraphQL**: `rank` on the `AppConfigAllowList` node and `CreateAppConfigAllowListInput` (version-tagged with `NEXT_RELEASE_VERSION`), `RANK` in `AppConfigAllowListOrderField`; supergraph / v2 schema regenerated.
- **CLI**: `./bai admin app-config-allow-list create --rank`.
- **New `update` operation (rank only)** — with rank on the allow list, re-ordering the merge needs a post-create adjustment (purge now cascades to fragments, so purge+recreate is not a substitute): `AppConfigAllowListUpdaterSpec` + service action/processor, `PATCH /v2/app-config-allow-list/{id}`, `adminUpdateAppConfigAllowList` GQL mutation, SDK `admin_update`, and `./bai admin app-config-allow-list update <id> --rank`. The identity pair (`config_name`, `scope_type`) stays immutable.
- The SDK create/get/search need no code change (they share the DTOs).

Verified against the live server: REST/GQL/CLI create with default (user=300) and explicit (`--rank 250`) ranks, `rank` ordering in both directions, and rank update via both the CLI (`update --rank 275`) and the GQL mutation.

## 📚 Stacked PRs

Part of the AppConfigFragment / AppConfig stack under BEP-1052 (epic BA-5781). **Merge in order:**

1. ✅ #12306 — `feat(BA-6552): app_config_fragments DB model and Alembic migration`
2. ✅ #12307 — `feat(BA-6553): repository layer`
3. ✅ #12403 — `refactor(BA-6619): consolidate AppConfigScopeType into common.data`
4. ✅ #12405 — `refactor(BA-6620): ExistsQuerier + AppConfigAllowList.exists`
5. ✅ #12358 — `feat(BA-6554): AppConfigFragment service layer`
6. #12516 — `feat(BA-6628): move fragment rank to the allow list with scope defaults` (replaces #12429)
[...]

PR #12429, opened 2026-06-26
feat(BA-6628): conditional-bulk repository primitives

jopemachine:
## Summary

Generic, domain-agnostic repository primitives for **gated bulk writes**, split out of #12426 (BA-6626) so app_config_fragment — and future domains — build on them. No app_config-specific code.

- `repositories/base`: `ConditionalCreator` / `ConditionalUpdater` / `ConditionalPurger` and `BulkConditional{Creator,Updater,Purger}` — a write spec paired with an `only_if` existence gate.
- `WriteOps.bulk_conditional_create_partial` / `_update_partial` / `_purge_partial` — each item runs in its own savepoint for **partial success** (gate-rejected / missing / failed items reported per-item; the rest proceed). Implemented directly in `WriteOps` with private `_update_one`/`_purge_one` row helpers (not delegating to the standalone `execute_*` functions).
- Repository errors `ConditionalWriteNotAllowed` / `ConditionalWriteTargetNotFound`.
- Ops-provider unit tests.

## 📚 Stacked PRs

Part of the AppConfigFragment / AppConfig stack under BEP-1052 (epic BA-5781). **Merge in order:**

1. ✅ #12306 — `feat(BA-6552): app_config_fragments DB model and Alembic migration`
2. ✅ #12307 — `feat(BA-6553): repository layer`
3. ✅ #12403 — `refactor(BA-6619): consolidate AppConfigScopeType into common.data`
4. ✅ #12405 — `refactor(BA-6620): ExistsQuerier + AppConfigAllowList.exists`
5. ✅ #12358 — `feat(BA-6554): AppConfigFragment service layer`
6. 👉 **#12429 — `feat(BA-6628): conditional-bulk repository primitives`** ← you are here
7. #12426 — `feat(BA-6626): app_config_fragment bulk repository layer`
8. #12401 — `feat(BA-6618): AppConfigFragment bulk CRUD service layer`
9. #12359 — `feat(BA-6555): app_config service layer`
10. #12377 — `feat(BA-6556): AppConfig REST v2 API`

PR #12426, opened 2026-06-26
feat(BA-6626): app_config_fragment bulk repository layer

jopemachine:
## Summary

Repository-layer `app_config_fragment` bulk operations, split out of #12401 (BA-6618) to keep that PR reviewable. No API/service surface here — the service layer stacks on top. Built on the generic partial bulk ops already in `WriteOps` (`bulk_create_partial` / `bulk_update_partial` / `bulk_purge_partial`) — no new base primitives.

- `bulk_create(creators: Sequence[Creator[AppConfigFragmentRow]])` — **no write-gate**: the FK to the allow-list is the gate, so an item with no allow-list row for its `(config_name, scope_type)` fails per item as `AppConfigFragmentWriteNotAllowed` (via the spec's integrity checks), and a duplicate natural key fails likewise. Each insert runs in its own savepoint for **partial success**.
- `bulk_update(updaters)` / `bulk_purge(purgers)` — likewise no gate: a fragment row exists only while its allow-list entry does (FK with cascade, #12518), so an existing fragment is always writable at its own scope. Missing targets are reported as per-item not-found failures (the generic partial ops silently skip a missing PK).
- The `AppConfigFragmentBulkWriteResult` data type (`succeeded` + `failed[index, message]`).
- Repository tests (real DB): partial create with a not-allow-listed item, partial update/purge with a missing target.

## 📚 Stacked PRs

Part of the AppConfigFragment / AppConfig stack under BEP-1052 (epic BA-5781). **Merge in order:**

1. ✅ #12306 — `feat(BA-6552): app_config_fragments DB model and Alembic migration`
2. ✅ #12307 — `feat(BA-6553): repository layer`
3. ✅ #12403 — `refactor(BA-6619): consolidate AppConfigScopeType into common.data`
4. ✅ #12405 — `refactor(BA-6620): ExistsQuerier + AppConfigAllowList.exists`
5. ✅ #12358 — `feat(BA-6554): AppConfigFragment service layer`
6. ✅ #12516 — `feat(BA-6628): move fragment rank to the allow list with scope defaults` (replaces #12429)
7. ✅ #12517 — `feat(BA-6628): expose allow-list rank on the v2 API surface`
8. ✅ #12518 — `feat(BA-6628): cascade fragment deletion from the allow list`
9. 👉 **#12426 — `feat(BA-6626): app_config_fragment bulk repository layer`** ← you are here
10. #12401 — `feat(BA-6618): AppConfigFragment bulk CRUD service layer`
[...]

PR #12405, opened 2026-06-25
refactor(BA-6620): ExistsQuerier ops primitive + AppConfigAllowList.exists

jopemachine:
## Summary

Adds the existence-check repository primitive used by the AppConfig fragment write-gate, split out of #12358 to keep that PR small:

- **`ExistsQuerier`** dataclass + `execute_exists` in `repositories/base` — `SELECT EXISTS(SELECT 1 FROM table WHERE ...)`, no count/fetch.
- **`ReadOps.exists`** on the ops provider (so it is available inside `write_ops()` too — `WriteOps` extends `ReadOps`).
- **`AppConfigAllowListRepository.exists`** (+ db_source) built on `ExistsQuerier`.
- Real-DB repository tests for `exists`.

The fragment write-gate (#12358) consumes this; here it lands as reusable infrastructure at the base of the stack.

## 📚 Stacked PRs

Part of the AppConfigFragment / AppConfig stack under BEP-1052 (epic BA-5781). **Merge in order:**

1. ✅ #12306 — `feat(BA-6552): app_config_fragments DB model and Alembic migration`
2. ✅ #12307 — `feat(BA-6553): repository layer`
3. #12403 — `refactor(BA-6619): consolidate AppConfigScopeType into common.data`
4. 👉 **#12405 — `refactor(BA-6620): ExistsQuerier + AppConfigAllowList.exists`** ← you are here
5. #12358 — `feat(BA-6554): AppConfigFragment service layer`
6. #12401 — `feat(BA-6618): AppConfigFragment bulk CRUD service layer`
7. #12359 — `feat(BA-6555): app_config service layer`
8. #12377 — `feat(BA-6556): AppConfig REST v2 API`

PR #12518, opened 2026-07-03
feat(BA-6704): cascade app config subtree deletion from the definition

jopemachine:
## Summary

Ties fragments to their allow-list entry with a composite FK and lets deletion cascade the whole subtree `definition → allow_list → fragment` (BEP-1052 review feedback):

- **Composite FK with `ON DELETE CASCADE`.** `app_config_fragments (config_name, scope_type)` references `app_config_allow_list`, so purging an allow-list entry drops every fragment written under it — a revoked value disappears from the merge in the same statement, with no orphaned fragments to clean up separately.
- **Cascade from the definition, with exactly two FKs.** `app_config_allow_list.config_name → app_config_definitions` changes from `NO ACTION` to `ON DELETE CASCADE`, and the **redundant direct** `app_config_fragments.config_name → app_config_definitions` FK is **dropped** (not made cascade). A fragment already requires a matching allow-list entry (composite FK, both columns `NOT NULL`) whose `config_name` is registered, so registration is guaranteed transitively — the direct FK added no integrity and only blocked the definition delete (a `NO ACTION` check firing before the allow-list cascade cleared the fragments). Result: retiring a config name (`DELETE` its definition row) cascades cleanly through `allow_list` down to the fragments.
- **Write-gate cleanup.** With the composite FK in place, a fragment row can only exist while its allow-list entry does, so the `only_if` gates on fragment **update/purge** are provably dead and are removed (from the actions, service, repository, and db_source). The **create** gate stays — it surfaces the domain error (`AppConfigFragmentWriteNotAllowed`) instead of a raw FK violation.
- **Single migration `a560420476b6`**: deletes orphaned fragments, adds the composite allow-list FK, swaps the allow-list→definition FK to `CASCADE`, and drops the redundant direct fragment→definition FK. Constraint names are discovered via the inspector (they differ between a migration-built DB and a metadata-built dev DB). Upgrade/downgrade round-trip and the live `definition → allow_list → fragment` cascade were verified against a local DB; the end state is exactly two FKs.
- **BEP-1052 updated**: cascade semantics on allow-list purge **and** on definition retirement, the transitive-registration model (fragments need no direct definition FK), the corrected read model (the merge *must* join the allow-list for `rank` — the "reads never consult the allow-list" wording was stale), and the create-only write gate.

## 📚 Stacked PRs

Part of the AppConfigFragment / AppConfig stack under BEP-1052 (epic BA-5781). **Merge in order:**

1. ✅ #12306 — `feat(BA-6552): app_config_fragments DB model and Alembic migration`
2. ✅ #12307 — `feat(BA-6553): repository layer`
3. ✅ #12403 — `refactor(BA-6619): consolidate AppConfigScopeType into common.data`
4. ✅ #12405 — `refactor(BA-6620): ExistsQuerier + AppConfigAllowList.exists`
5. ✅ #12358 — `feat(BA-6554): AppConfigFragment service layer`
6. #12516 — `feat(BA-6628): move fragment rank to the allow list with scope defaults` (replaces #12429)
7. #12517 — `feat(BA-6628): expose allow-list rank on the v2 API surface`
8. 👉 **#12518 — `feat(BA-6704): cascade app config subtree deletion from the definition`** ← you are here
9. #12426 — `feat(BA-6626): app_config_fragment bulk repository layer`
10. #12401 — `feat(BA-6618): AppConfigFragment bulk CRUD service layer`
11. #12359 — `feat(BA-6555): app_config service layer (merge engine)`
12. #12377 — `feat(BA-6556): AppConfig REST v2 API`

----
📚 Documentation preview 📚: https://sorna--12518.org.readthedocs.build/en/12518/

----
📚 Documentation preview 📚: https://sorna-ko--12518.org.readthedocs.build/ko/12518/


src/ai/backend/manager/repositories/app_config_fragment/db_source/db_source.py
```diff
@@ -73,7 +73,8 @@ async def create(
         only_if: ExistsQuerier[AppConfigAllowListRow],
```
jopemachine: I know that now we have a foreign key constraint between the AllowList and AppConfigFragment tables, the only_if check is technically no longer necessary.

However, passing only_if makes the behavior more explicit and easier to understand at a glance.

Should we keep the ExistsQuerier for the sake of future extensibility and explicitness, or would you prefer to remove it? @HyeockJinKim

src/ai/backend/manager/models/app_config_fragment/row.py
```diff
@@ -20,8 +20,12 @@
...
     """One scoped app config fragment — a single JSON document at ``(config_name, scope_type, scope_id)``.
 
-    The fragment carries no merge priority of its own — its rank is the ``rank`` of
-    the allow-list entry for its ``(config_name, scope_type)``.
+    A fragment hangs off its ``app_config_allow_list`` entry (FK on
+    ``(config_name, scope_type)``, ``ON DELETE CASCADE``): it exists only while the
+    entry does, and its merge priority is the entry's ``rank`` — the fragment row
+    carries no rank of its own. No direct FK to ``app_config_definitions`` is
+    needed: the required allow-list entry already references a registered
+    ``config_name``.
```
seedspirit: Aside from the comments related to "rank," I think you can figure it out just by looking at the row. I don't think those comments are really necessary.

proposals/BEP-1052-scoped-app-config-redesign.md
```diff
@@ -66,18 +66,23 @@ Three scopes cover the use cases (`public` for the pre-login shell):
...
+  no permission or policy is evaluated at read time.
+- **Allow-list = the write gate and the merge order.** `app_config_allow_list`
   holds **one record per `(config_name, scope_type)`**; a fragment at
-  that scope may be created/updated/purged **only if** the record exists
-  — through the admin mutations and the regular ones alike. What sets
-  admins apart is that they alone manage the allow-list (and the
-  `app_config_definitions`) itself. It governs **writes only** — never reads.
+  that scope may be created **only if** the record exists — through the
+  admin mutations and the regular ones alike. Fragments reference their
+  entry by FK (`ON DELETE CASCADE`), so removing an entry removes its
```
seedspirit, line 83: Since there are many implementation details, it seems best to just outline the general direction here.

src/ai/backend/manager/repositories/app_config_fragment/db_source/db_source.py
HyeockJinKim: It does seem like we don't need ExistsQuerier.

HyeockJinKim (approved):

PR #12528, opened 2026-07-03
feat(BA-6605): implement virtual scope DB ops

fregataa:
## Summary
- Add read/write DB ops for the virtual scope ownership layer on the RBAC ops surface (`RBACReadOps` / `RBACWriteOps` / `RBACOpsProvider`).
- The virtual scope is hidden from callers: every method takes `ScopeRef` / `EntityRef` and resolves the backing `VirtualScopeID` internally. `VirtualScopeID` surfaces only in `VirtualScopeReach` read results.
- Writes: `create/delete_scope` (node), `bind/unbind_scope` (inbound edge), `add/remove_entity_members` (outbound edge, mixed entity types). Idempotent `ON CONFLICT DO NOTHING`, consistent with `add_users_to_scope`.
- Reads: bidirectional (`scope -> VS -> entity` and reverse) returning each hop's `permission_cap` so callers can clip.
- `VirtualScopeNotFound` (HTTP 500): a missing VS is a server-side invariant violation since every owner scope is created with its VS.

## Test plan
- [x] `pants check` / `test` (CI)

Resolves BA-6605

### DIFF
--- a/src/ai/backend/common/data/permission/virtual_scope.py
+++ b/src/ai/backend/common/data/permission/virtual_scope.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 
 from ai.backend.common.data.permission.types import Permission
-from ai.backend.common.entity.types import EntityRef, EntityType, ScopeRef, ScopeType
+from ai.backend.common.entity.types import EntityType, ScopeType
 from ai.backend.common.identifier.entity import EntityID
 from ai.backend.common.identifier.scope import ScopeID
 from ai.backend.common.identifier.virtual_scope import VirtualScopeID
@@ -12,7 +12,6 @@ __all__ = (
     "VirtualScopeData",
     "ScopeBindingData",
     "EntityMembershipData",
-    "VirtualScopeReach",
 )
 
 
@@ -53,19 +52,3 @@ class EntityMembershipData:
     entity_type: EntityType
     entity_id: EntityID
     permission_cap: Permission | None
-
-
-@dataclass(frozen=True)
-class VirtualScopeReach:
-    """One resolved ``scope -> virtual_scope -> entity`` path.
-
-    Carries both hop caps *unclipped*; the caller clips the effective permission by
-    bitwise-ANDing ``binding_cap`` and ``membership_cap`` (``None`` means no ceiling
-    at that hop).
-    """
-
-    virtual_scope_id: VirtualScopeID
-    scope: ScopeRef
-    entity: EntityRef
-    binding_cap: Permission | None  # real scope -> virtual_scope hop
-    membership_cap: Permission | None  # virtual_scope -> entity hop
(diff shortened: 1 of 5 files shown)


src/ai/backend/common/data/permission/virtual_scope.py
```diff
@@ -52,3 +53,19 @@ class EntityMembershipData:
     entity_type: EntityType
     entity_id: EntityID
     permission_cap: Permission | None
+
+
+@dataclass(frozen=True)
+class VirtualScopeReach:
```
HyeockJinKim: What does “Reach” mean?

src/ai/backend/manager/repositories/ops/rbac/provider.py
```diff
@@ -72,6 +159,158 @@ async def add_users_to_scope(
...
+        """Create multiple real scope entities and their virtual scope nodes.
+
+        Each real scope row and its related virtual scope node are created together
+        inside one nested transaction (savepoint), so an item's row and node commit
+        or roll back as a unit.
+        """
+        results: list[CreatorResult[TRow]] = []
+        for creation in creations:
+            async with self.savepoint():
+                result = await execute_creator(self._sess, creation.creator)
```
HyeockJinKim: It seems best to use the value of `self.execute_creator`.

src/ai/backend/manager/repositories/ops/rbac/provider.py
```diff
@@ -72,6 +161,147 @@ async def add_users_to_scope(
...
+    ) -> BatchPurgerResult:
+        """Delete the real scope entities matched by ``deletion.purger`` and their virtual
+        scope nodes.
+
+        The real scope rows are purged atomically via the batch purge (all-or-nothing), then
+        the virtual scope nodes for ``deletion.scopes`` are dropped (FK ``ON DELETE CASCADE``
+        removes their edges).
+        """
+        result = await self.batch_purge(deletion.purger)
+        await self._delete_virtual_scopes(deletion.scopes)
```
HyeockJinKim, line 220: How about ensuring that `delete` is set to `cascade` in the constraint?

src/ai/backend/manager/repositories/ops/rbac/provider.py
```diff
@@ -72,6 +161,147 @@ async def add_users_to_scope(
...
+        """
+        virtual_scope_id = await self._resolve_virtual_scope_id(owner)
+        stmt = (
+            pg_insert(ScopeBindingRow)
+            .values(
+                virtual_scope_id=virtual_scope_id,
+                scope_type=scope.scope_type,
+                scope_id=scope.scope_id,
+                permission_cap=permission_cap,
+            )
```
HyeockJinKim, line 245: Huh? Was a permission cap added here? Shouldn't permission caps be added at the entity and virtual scope levels?

HyeockJinKim (approved):

PR #12529, opened 2026-07-03
feat: Enable external event-dispatcher plugins to observe session lifecycle results

### LINKED ISSUE #12526: Manager: make the event-dispatcher plugin group usable by external plugins (forward session result events + standard dependency context)
### Main idea

Make the `backendai_event_dispatcher_v20` plugin group a fully usable extension point for **externally installed** (pip/package-metadata) manager plugins that react to scheduler and session lifecycle outcomes. Two gaps currently prevent this, and both can be fixed with small, plugin-agnostic changes.

#### Gap 1 — Session lifecycle result events never reach event-dispatcher plugins

`SessionEventHandler` forwards events to `EventDispatcherPluginContext.handle_event` only for `SessionStartedAnycastEvent` (`src/ai/backend/manager/event_dispatcher/handlers/session.py:103`). All other lifecycle handlers — notably `handle_batch_result` (`session.py:152`, which consumes `SessionSuccessAnycastEvent` / `SessionFailureAnycastEvent` carrying `reason` and `exit_code`) and `handle_session_terminated` — update the DB and invoke webhooks but never call the plugin context.

Since plugins never receive the `EventDispatcher` object itself, they cannot self-subscribe: the forwarding calls in the handlers are their *only* delivery path. As a result, a plugin today can observe that a session started, but not how it ended — which rules out whole categories of plugins (failure analytics, operator notifications, custom post-mortem or resubmission policies reacting to scheduler/batch outcomes).

Proposed change: forward the session result/termination events to `_event_dispatcher_plugin_ctx.handle_event(...)` in the corresponding handlers, mirroring what `handle_session_started` already does. Plugins filter by `isinstance` and ignore what they don't care about, so this is backward-compatible (the intrinsic `ErrorEventDispatcher` early-returns on non-`AgentErrorEvent`).

#### Gap 2 — No standard way for event-dispatcher plugins to obtain manager dependencies

Plugin contexts initialize at the Plugins stage (Stage 4), before repositories and processors exist, so any plugin that needs them must start disabled and be re-initialized later. The only existing mechanism is a per-plugin special case in the dependency composer:

```python
# src/ai/backend/manager/dependencies/composer.py:190
for plugin_instance in plugins.event_dispatcher_plugin_ctx.plugins.values():
    if isinstance(plugin_instance, ErrorEventDispatcher):
        await plugin_instance.init(context={"error_log_repository": ...})
```

This requires core to import each plugin's class — impossible for an out-of-tree plugin discovered via package metadata, and contrary to the plugin architecture (the accelerator plugin group, for comparison, needs no core-side knowledge of individual plugins).

Proposed change: replace the special case with a generic late re-initialization of **all** event-dispatcher plugins using a documented standard context dict (e.g. `etcd`, `config_provider`, `repositories`, `processors`, plus the existing `error_log_repository` key for compatibility). The natural place is inside `ProcessingComposer.compose`, after processors are created and **before** `event_dispatcher.start()` — this guarantees every plugin is fully wired before the first event can be delivered, and is strictly safer than the current Stage 6.5 timing (no events flow before `start()` anyway). Each plugin picks the keys it needs and must tolerate extra ones; the context keys become the stable contract for third-party manager plugins.

#### Expected outcome

- Third-party manager plugins installed via `pip` / `scripts/install-plugin.sh` can react to session scheduling outcomes and act through the service layer (processors), with zero manager source modifications.
- Core stops importing concrete plugin classes; `dependencies/composer.py` gets simpler.
- The intrinsic `ErrorEventDispatcher` keeps working unchanged (it already reads its dependency from the context dict by key).

### Alternative ideas

- Expose the `EventDispatcher` itself in the plugin init context and let plugins `subscribe()`/`consume()` directly. More flexible, but leaks delivery-mode decisions (anycast vs broadcast, consumer-group semantics) into plugins and makes it harder to reason about which events plugins see; the forwarding approach keeps handlers as the single choke point.
- Keep per-plugin re-init wiring in the composer but key it by entrypoint name instead of `isinstance`. Avoids the class import, but core still needs to know each plugin's dependency list, so it does not scale to external plugins.
- Out-of-process integration via session `callback_url` webhooks + REST API. Works without core changes but requires separate deployment/credentials and only covers sessions that register the callback, so it is not a substitute for a proper in-process extension point.

### Anything else?

Both changes are small and have been validated locally (unit tests for the handler forwarding and the generic re-init pass, and an externally installed test plugin is discovered via `scan_entrypoint_from_package_metadata`, receives failure events, and gets the standard context). Happy to submit the PR.

rapsealk:
## Summary

- Forward session lifecycle events (`SessionSuccess/FailureAnycastEvent`, `SessionTerminatedAnycastEvent`, `SessionCancelledAnycastEvent`, `SessionTerminatingAnycastEvent`) to `backendai_event_dispatcher_v20` plugins in `SessionEventHandler`, mirroring the existing `session_started` forwarding — previously plugins could observe that a session started but never how it ended.
- Replace the `isinstance(ErrorEventDispatcher)` special-case re-initialization in `dependencies/composer.py` with a generic late re-init of **all** event-dispatcher plugins inside `ProcessingComposer`, using a standard context dict (`etcd`, `config_provider`, `repositories`, `processors`, plus `error_log_repository` for compatibility). It runs after processors creation and **before** `event_dispatcher.start()`, so every plugin is fully wired before the first event can be delivered.
- Harden the extension point for third-party code: a plugin whose `init` raises is skipped (manager startup is never aborted), and a plugin raising in `handle_event` is logged and suppressed per-plugin — so it can neither leave the message un-acked (which would cause redelivery with duplicate side effects) nor starve sibling plugins.
- Document the plugin lifecycle and the standard context-key contract on `AbstractEventDispatcherPlugin`, where plugin authors will actually find it.
- Together these make the event-dispatcher plugin group a real extension point for externally installed (package-metadata-discovered) plugins: core no longer imports any concrete plugin class.

**Behavioral note:** with per-plugin exception isolation, an exception in the intrinsic `ErrorEventDispatcher`'s `handle_event` (e.g. a transient DB failure while writing an error log) is now logged and dropped instead of leaving the message un-ACKed for auto-claim retry. This trades best-effort error-log telemetry for eliminating poison-message redelivery loops (which previously re-ran the whole handler, including duplicate webhooks).

## Test plan

- [x] `tests/unit/manager/event_dispatcher/handlers/test_session.py::TestEventDispatcherPluginForwarding` — batch success, batch failure, terminated, cancelled, and terminating events reach the plugin context with the original source/event.
- [x] `tests/unit/common/plugin/test_event.py` — a raising plugin does not propagate into the caller and sibling plugins still receive the event.
- [x] `tests/unit/manager/dependencies/processing/test_composer.py` — plugins receive the standard context keys, `init` is awaited before `event_dispatcher.start()`, and a plugin whose `init` raises does not abort composition.
- [x] `pants fmt/fix/lint/check --changed-since=origin/main` clean; existing handler/composer/plugin suites pass.
- [x] Verified locally with an externally pip-installed test plugin: discovered via `scan_entrypoint_from_package_metadata`, receives failure events, and gets the standard context (no core modification).

Resolves #12526

### DIFF
--- a/changes/12529.feature.md
+++ b/changes/12529.feature.md
@@ -1,0 +1,1 @@
+Forward session result and termination events to manager event-dispatcher plugins and initialize them with a standard dependency context, so externally installed plugins can react to session lifecycle outcomes without manager source changes.
--- a/src/ai/backend/common/plugin/event.py
+++ b/src/ai/backend/common/plugin/event.py
@@ -14,6 +14,32 @@ log = BraceStyleAdapter(logging.getLogger(__spec__.name))
 
 
 class AbstractEventDispatcherPlugin(AbstractPlugin, metaclass=ABCMeta):
+    """Base class for event-dispatcher plugins.
+
+    Lifecycle contract for implementations (in-tree or externally installed):
+
+    - ``init(context=None)`` is called once at plugin discovery time, before
+      the host's dependency graph exists. Implementations must tolerate a
+      ``None`` context (typically by staying disabled).
+    - The host calls ``init`` again with a context mapping after its
+      dependencies are ready and **before** any event is delivered, so
+      ``init`` must be safely re-entrant. The manager provides at least the
+      following keys and may add more; implementations must pick the keys
+      they need and tolerate extra ones:
+
+      - ``etcd``: the shared :class:`~ai.backend.common.etcd.AsyncEtcd` client
+      - ``config_provider``: the manager configuration provider
+      - ``repositories``: the manager repositories container
+      - ``processors``: the manager service-layer processors container
+      - ``error_log_repository``: shortcut kept for backward compatibility
+
+    - ``handle_event`` receives every event the host forwards and must filter
+      by event type (``isinstance``) and return quickly for events it does
+      not care about. Exceptions raised here are logged and suppressed by the
+      host so that one plugin cannot disturb event handling or sibling
+      plugins — do not rely on exceptions for control flow.
+    """
+
     @override
     async def init(self, context: Any | None = None) -> None:
         pass
@@ -40,5 +66,15 @@ class EventDispatcherPluginContext(BasePluginContext[AbstractEventDispatcherPlug
         source: AgentId,
         event: AbstractEvent,
     ) -> None:
-        for plugin_instance in self.plugins.values():
-            await plugin_instance.handle_event(source, event)
+        for plugin_name, plugin_instance in self.plugins.items():
+            try:
+                await plugin_instance.handle_event(source, event)
+            except Exception:
+                # A misbehaving plugin must not disturb the surrounding event
+                # handler (which would leave the message un-acked and cause
+                # redelivery) nor starve sibling plugins of the event.
+                log.exception(
+                    "event-dispatcher plugin {} failed to handle event {}",
+                    plugin_name,
+                    event.event_name(),
+                )
--- a/src/ai/backend/manager/dependencies/composer.py
+++ b/src/ai/backend/manager/dependencies/composer.py
@@ -8,7 +8,6 @@ from typing import override
 
 from ai.backend.common.dependencies import DependencyComposer, DependencyStack
 from ai.backend.logging.types import LogLevel
-from ai.backend.manager.plugin.error_monitor import ErrorEventDispatcher
 from ai.backend.manager.plugin.monitor import ManagerErrorPluginContext, ManagerStatsPluginContext
 
 from .agents import AgentsComposer, AgentsInput, AgentsResources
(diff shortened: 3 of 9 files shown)


rapsealk:
Reviewed with an adversarial pass over the full diff (robustness, DI ordering, delivery/ACK semantics, contract quality, tests):

- Plugin exception isolation verified: a raising `handle_event` no longer propagates into the dispatcher, so messages are ACKed (no auto-claim redelivery / duplicate side effects) and sibling plugins still receive the event. `CancelledError` still propagates for clean shutdown.
- A plugin whose `init` raises is skipped with a log instead of aborting manager startup, matching the `BasePluginContext.init` precedent.
- Re-init runs strictly before `event_dispatcher.start()`, so no event can reach a half-wired plugin; the intrinsic `ErrorEventDispatcher` receives the identical repository object it got before.
- The lifecycle + context-key contract is documented on `AbstractEventDispatcherPlugin` where plugin authors will find it.
- Full lifecycle coverage (started/success/failure/terminating/terminated/cancelled), including the cancelled path where the creation-tracker early-return is taken.

Remaining nits are follow-up material (per-plugin failure metrics, `BraceStyleAdapter` in the composer logger).

LGTM

PR #12535, opened 2026-07-04
fix(BA-6695): chunk bulk INSERTs in RBAC backfill migrations

fregataa:
## Summary
- RBAC backfill migrations built a single multi-row `INSERT` whose bind-parameter count (`rows × columns`) exceeded PostgreSQL's Int16 limit of 32767 on large sites, aborting `alembic upgrade` with a `DBAPIError`.
- Chunk the shared `insert_skip_on_conflict` helper (`rbac_models/migration/utils.py`) by column count, covering `2c9000848b6e` and the other migrations that use it; apply the same chunking to the local copies in `d3683e2703ff` and `ad7acfe8aa1c`.
- Reduce `2c9000848b6e` `page_size` to bound the in-memory batch.
- Document the "always chunk bulk INSERTs" rule in the alembic `README.md`.

## Test plan
- [x] Run the affected migrations against a DB seeded with enough users/projects to exceed 32767 bind parameters and confirm the upgrade completes.
- [x] Confirm re-running is idempotent (`ON CONFLICT DO NOTHING`).

Resolves BA-6695

### DIFF
--- a/src/ai/backend/manager/models/alembic/versions/d3683e2703ff_backfill_role_scope_mappings.py
+++ b/src/ai/backend/manager/models/alembic/versions/d3683e2703ff_backfill_role_scope_mappings.py
@@ -95,11 +95,19 @@ def _get_user_roles_table() -> sa.Table:
     )
 
 
+_MAX_BIND_PARAMS = 16000
+
+
 def _insert_skip_on_conflict(
     db_conn: Connection, table: sa.Table, rows: list[dict[str, Any]]
 ) -> None:
-    if rows:
-        stmt = pg_insert(table).values(rows).on_conflict_do_nothing()
+    if not rows:
+        return
+    params_per_row = max(1, len(rows[0]))
+    chunk_size = max(1, _MAX_BIND_PARAMS // params_per_row)
+    for start in range(0, len(rows), chunk_size):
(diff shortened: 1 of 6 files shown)


src/ai/backend/manager/models/alembic/versions/d3683e2703ff_backfill_role_scope_mappings.py
```diff
@@ -95,11 +95,19 @@ def _get_user_roles_table() -> sa.Table:
...
 ) -> None:
-    if rows:
-        stmt = pg_insert(table).values(rows).on_conflict_do_nothing()
+    if not rows:
+        return
+    params_per_row = max(1, len(rows[0]))
+    chunk_size = max(1, _MAX_BIND_PARAMS // params_per_row)
+    for start in range(0, len(rows), chunk_size):
+        chunk = rows[start : start + chunk_size]
+        stmt = pg_insert(table).values(chunk).on_conflict_do_nothing()
```
jopemachine, line 110: It might be worth encapsulating _MAX_BIND_PARAMS and the related logic into a utility function, rather than duplicating the implementation wherever it's needed.

jopemachine (approved):

PR #12543, opened 2026-07-06
fix(BA-6699): route webserver auth paths through HealthyEndpointPool

jopemachine:
## Summary

The web server accepts multiple Manager endpoints via `api.endpoint` for HA. Proxied REST/GraphQL traffic already fails over correctly through `HealthyEndpointPool`, but the **login, logout, edu-launcher token-login, and no-auth password-reset** paths were pinned to `config.api.endpoint[0]`. When the *first* configured endpoint specifically was down, these auth flows failed even though other Managers were healthy.

This PR routes all four paths through `app["manager_pool"]` (`HealthyEndpointPool`), so they select a healthy endpoint and only surface `ManagerConnectionUnavailable` (503) when none is available — consistent with proxied traffic.

Fixes BA-6699.

## Changes

| Path | Before | After |
|------|--------|-------|
| `login_handler` | anonymous `APIConfig` on `endpoint[0]` | endpoint acquired from `manager_pool` |
| `token_login_handler` (edu-launcher) | anonymous `APIConfig` on `endpoint[0]` | same (shared helper) |
| `logout_handler` | `POST endpoint[0]/auth/logout` | endpoint acquired from `manager_pool` |
| `update_password_no_auth` (password reset) | single v2 registry pinned to `endpoint[0]` at startup | one registry per endpoint; handler uses the pool-selected healthy endpoint's registry |

### Notes

- **`_authorize_via_anonymous_session` helper**: `login` and `token_login` shared the same anonymous-session authorize logic (no access/secret key, used to obtain a token), now unified. The request runs *inside* the `async with manager_pool.acquire()` context so the `LeastConnections` strategy's in-flight accounting stays correct.
- **No-auth password reset**: the v2 `ClientConfig` accepts only a single endpoint and has no built-in failover. `no_auth_client_registries_ctx` now builds one no-auth `BackendAIClientRegistry` per configured endpoint at startup (keyed by endpoint string), and the handler picks the registry of a healthy endpoint chosen via `manager_pool` — keeping the existing context-manager lifecycle and persistent-connection reuse, just no longer pinned to `endpoint[0]`.

## Scope

Proxied REST (`web_handler`) and GraphQL/Apollo (`web_handler_with_jwt`) traffic were already HA and are untouched. Only the auth helper calls listed above were pinned to `endpoint[0]`.

## Verification

Unit tests (`tests/unit/webserver/test_server.py`) cover `_authorize_via_anonymous_session` and `no_auth_client_registries_ctx`.

End-to-end, this was verified against a live dev stack: a throwaway web server is configured with a **dead first endpoint** and the **live Manager second** (`api.endpoint = "http://127.0.0.1:19999,http://127.0.0.1:8091"`), then login/logout are driven through the Backend.AI v2 client SDK. The script reads the generated config back to confirm the HA endpoint list actually took effect, and waits until the pool's probe flags the dead endpoint unhealthy. With `endpoint[0]` dead, a successful authenticated login can only have reached `endpoint[1]` — so it proves the auth path is no longer pinned to `endpoint[0]`. Before this fix the same setup fails (login pins to the dead `endpoint[0]` → HTTP 502).

Observed output:

```
endpoint[0] = http://127.0.0.1:19999 (dead), endpoint[1] = http://127.0.0.1:8091 (live)
web server config: endpoint = "http://127.0.0.1:19999,http://127.0.0.1:8091"
pool flagged the dead endpoint[0] as unhealthy; trying auth via the SDK:
  login  -> HTTP 200, authenticated=True
  logout -> sent
PASS: auth succeeded via endpoint[1] — not pinned to endpoint[0].
```

<details>
<summary>Verification script (<code>./py verify_ba6699.py</code>, requires a live Manager on :8091)</summary>

```python
"""BA-6699 verification — the web server's auth paths fail over to a healthy
Manager instead of always hitting ``api.endpoint[0]``.

It boots a throwaway web server whose FIRST Manager endpoint is dead and whose
SECOND is the live Manager, then logs in / out through the Backend.AI v2 client
SDK. A successful login proves the request was routed to the healthy endpoint
rather than pinned to the dead ``endpoint[0]``.

Prerequisite: a live Manager at http://127.0.0.1:8091 (the local dev stack).

Run from the repo root:

    BACKEND_USER=admin@lablup.com BACKEND_PASSWORD=wJalrXUt ./py verify_ba6699.py
"""

from __future__ import annotations

import asyncio
import os
import signal
import socket
import subprocess
import sys
import tempfile
import time
from pathlib import Path

from yarl import URL

from ai.backend.client.v2.auth import NoAuth
from ai.backend.client.v2.config import ClientConfig
from ai.backend.client.v2.registry import BackendAIClientRegistry

LIVE_MANAGER = "http://127.0.0.1:8091"
DEAD_MANAGER = "http://127.0.0.1:19999"  # nothing listens here
HA_ENDPOINT = f"{DEAD_MANAGER},{LIVE_MANAGER}"  # dead first, live second
WEB_PORT = 18092
WEB_URL = f"http://127.0.0.1:{WEB_PORT}"
USER = os.environ.get("BACKEND_USER", "admin@lablup.com")
PASSWORD = os.environ.get("BACKEND_PASSWORD", "wJalrXUt")


def write_ha_config() -> Path:
    """Clone webserver.conf but aim [api].endpoint at [DEAD, LIVE]."""
    conf = Path("webserver.conf").read_text()
    original = 'endpoint = "http://127.0.0.1:8091"'
    if original not in conf:
        raise RuntimeError(f"{original!r} not found in webserver.conf — template changed?")
    conf = conf.replace("port = 8090", f"port = {WEB_PORT}")
    conf = conf.replace("ssl_verify = true", "ssl_verify = false")
    conf = conf.replace(
        original,
        f'endpoint = "{HA_ENDPOINT}"\n'
        "health-check-interval = 1.0\n"
        "health-check-failure-threshold = 1\n"
        "health-check-probe-timeout = 1.0",
    )
    path = Path(tempfile.mkstemp(prefix="ba6699-", suffix=".conf")[1])
    path.write_text(conf)
    return path


def assert_ha_endpoint_written(cfg: Path) -> None:
    """Read the written config back and confirm the [api].endpoint list took effect."""
    api_line = next(
        (ln.strip() for ln in cfg.read_text().splitlines() if DEAD_MANAGER in ln),
        None,
    )
    if api_line is None or HA_ENDPOINT not in api_line:
        raise RuntimeError(f"HA endpoint not present in generated config: {cfg}")
    print(f"web server config: {api_line}")


def wait_until_dead_endpoint_flagged(log: Path, timeout: float = 30.0) -> None:
    """Block until the pool's probe marks the dead endpoint unhealthy."""
    deadline = time.monotonic() + timeout
    while time.monotonic() < deadline:
        if "19999" in log.read_text() and "marked unhealthy" in log.read_text():
            return
        time.sleep(0.5)
    raise TimeoutError("web server did not flag the dead endpoint in time")


async def post(client, path: str, **kw):
    async with client.session.post(client.build_url_raw(path), **kw) as resp:
        body = await (resp.json() if resp.content_type == "application/json" else resp.text())
        return resp.status, body


async def run_auth_checks() -> bool:
    config = ClientConfig(
        endpoint=URL(WEB_URL), endpoint_type="session", skip_ssl_verification=True
    )
    registry = await BackendAIClientRegistry.create(config, NoAuth())
    client = registry._client
    try:
        status, body = await post(
            client, "/server/login", json={"username": USER, "password": PASSWORD}
        )
        authenticated = isinstance(body, dict) and bool(body.get("authenticated"))
        print(f"  login  -> HTTP {status}, authenticated={authenticated}")
        await post(client, "/server/logout")
        print("  logout -> sent")
        return status == 200 and authenticated
    finally:
        # Guard the close so a slow SDK session teardown can't hang the script.
        try:
            await asyncio.wait_for(registry.close(), timeout=10)
        except asyncio.TimeoutError:
            pass


def _kill_group(proc: subprocess.Popen) -> None:
    """Terminate the web server's whole process group promptly (workers included)."""
    try:
        pgid = os.getpgid(proc.pid)
    except ProcessLookupError:
        return
    os.killpg(pgid, signal.SIGTERM)
    try:
        proc.wait(timeout=3)
    except subprocess.TimeoutExpired:
        os.killpg(pgid, signal.SIGKILL)
        proc.wait()


def main() -> int:
    if socket.socket().connect_ex(("127.0.0.1", 8091)) != 0:
        print("FAIL: no live Manager on :8091 — start the dev stack first.")
        return 1

    cfg = write_ha_config()
    log = Path(tempfile.mkstemp(prefix="ba6699-", suffix=".log")[1])
    print(f"endpoint[0] = {DEAD_MANAGER} (dead), endpoint[1] = {LIVE_MANAGER} (live)")
    assert_ha_endpoint_written(cfg)
    with open(log, "w") as log_fp:
        # Own process group so we can kill the web server *and* its workers.
        web = subprocess.Popen(
            [sys.executable, "-m", "ai.backend.web.server", "-f", str(cfg)],
            stdout=log_fp,
            stderr=subprocess.STDOUT,
            start_new_session=True,
        )
    try:
        wait_until_dead_endpoint_flagged(log)
        print("pool flagged the dead endpoint[0] as unhealthy; trying auth via the SDK:")
        ok = asyncio.run(run_auth_checks())
    finally:
        _kill_group(web)
        cfg.unlink(missing_ok=True)
        log.unlink(missing_ok=True)

    print("PASS: auth succeeded via endpoint[1] — not pinned to endpoint[0]." if ok else "FAIL.")
    return 0 if ok else 1


if __name__ == "__main__":
    sys.exit(main())
```

</details>


src/ai/backend/web/server.py
```diff
@@ -336,6 +343,43 @@ async def login_check_handler(request: web.Request) -> web.Response:
...
+    async with manager_pool.acquire() as acquired:
+        anon_api_config = APIConfig(
+            domain=config.api.domain,
+            endpoint=acquired.endpoint,
+            access_key="",
+            secret_key="",  # anonymous session
+            user_agent=user_agent,
+            skip_sslcert_validation=not config.api.ssl_verify,
+        )
+        if not anon_api_config.is_anonymous:
```
seedspirit: nitpick: If `anon_api_config.is_anonymous` is turned off, does the system always throw an error? If so, this might need to be improved in the future.

src/ai/backend/web/server.py
```diff
@@ -336,6 +343,43 @@ async def login_check_handler(request: web.Request) -> web.Response:
...
+            raise InvalidAPIConfigurationError(
+                "Anonymous API configuration is not properly initialized."
+            )
+        async with APISession(config=anon_api_config) as api_session:
+            fill_forwarding_hdrs_to_api_session(request, api_session)
+            if forward_cookies:
+                api_session.aiohttp_session.cookie_jar.update_cookies(request.cookies)
+            return await api_session.User.authorize(username, password, extra_args=extra_args)
```
seedspirit: `if forward_cookies` seems like a new action that wasn't there before—was it added as part of a bug fix this time?

src/ai/backend/web/server.py
```diff
@@ -437,72 +481,68 @@ async def _set_login_history(last_login_attempt: float, login_fail_count: float
...
-            secret_key="",  # anonymous session
-            user_agent=user_agent,
-            skip_sslcert_validation=not config.api.ssl_verify,
+        extra_args = {}
+        extra_keys = set(creds.keys()) ^ {"username", "password"}
+        for extra_key in extra_keys:
+            extra_args[extra_key] = creds[extra_key]
+        auth_result = await _authorize_via_anonymous_session(
+            request,
+            manager_pool,
```
seedspirit: Wouldn't it be better to inject `acquired` instead of `pool`?

src/ai/backend/web/server.py
jopemachine: Since the APIConfig object is created just above, is_anonymous is always True, so this branch is never executed. It would be better to handle the code refactoring as a separate change.

src/ai/backend/web/server.py
jopemachine: It was not a new action.

The login handler did not call update_cookies, while the token login handler did. The forward_cookies flag and its branch were introduced when refactoring the logic into _authorize_via_anonymous_session, in order to preserve the existing behavior.

src/ai/backend/web/server.py
```diff
@@ -939,18 +974,33 @@ async def _shutdown(_app: web.Application) -> None:
...
+    """Build one no-auth v2 client registry per configured Manager endpoint.
+
+    Keyed by the same endpoint strings the manager pool hands out, so the
+    no-auth password reset path can use the registry of a pool-selected
+    healthy endpoint instead of pinning to ``config.api.endpoint[0]``.
+    """
+    registries: dict[str, BackendAIClientRegistry] = {}
     try:
-        yield registry
+        for endpoint in config.api.endpoint:
```
fregataa: Can we use created manager pool here rather than using config?

fregataa (approved):

PR #12559, opened 2026-07-06
fix(BA-6717): apply resource group default_session_options to enqueued sessions

fregataa:
## Summary
- The session enqueue draft pre-filled `handler_options` and `resource_opts` with non-None defaults, so `MergeResourceGroupDefaultsRule` always took the draft branch and the resource group's `default_session_options` were silently ignored.
- Leave both unset (`None`) at draft construction so the merge rule fills them from the resource group defaults; the spec field defaults restore the prior values when neither caller nor resource group provides one.
- Add service-layer regression tests (draft leaves `handler_options`/`resource_opts` unset when the caller omits them; shmem preserved) and strengthen `MergeResourceGroupDefaultsRule` tests to assert per-handler `handler_options` and execution-spec `resource_opts` propagation.

## Test plan
- [x] Set `default_session_options.handler_options` / `resource_opts` on a resource group, enqueue a session without those values, and confirm the session inherits the resource group defaults.
- [x] Enqueue with caller-supplied values and confirm they still override the resource group defaults.
- [x] `pants test` on the added/updated unit tests passes in CI.

Resolves BA-6717

### DIFF
--- a/src/ai/backend/manager/services/session/service.py
+++ b/src/ai/backend/manager/services/session/service.py
@@ -66,7 +66,6 @@ from ai.backend.manager.data.session.draft import (
 from ai.backend.manager.data.session.options import (
     InternalDataExtras,
     ResourceOpts,
-    SessionHandlerOptions,
 )
 from ai.backend.manager.data.session.types import SessionStatus
 from ai.backend.manager.defs import DEFAULT_ROLE
@@ -1584,7 +1583,9 @@ class SessionService:
         resource_opts_payload: dict[str, Any] = {}
         if action.resource.shmem is not None:
             resource_opts_payload["shmem"] = BinarySize.from_str(action.resource.shmem)
-        resource_opts = ResourceOpts.model_validate(resource_opts_payload)
+        resource_opts = (
+            ResourceOpts.model_validate(resource_opts_payload) if resource_opts_payload else None
+        )
 
         mount_entries = tuple(action.mounts or ())
 
@@ -1681,7 +1682,7 @@ class SessionService:
                     ),
                 ),
                 kernel_groups=kernel_groups,
-                handler_options=SessionHandlerOptions(),
+                handler_options=None,
             ),
             internal_data_extras=InternalDataExtras(
                 sudo_session_enabled=False,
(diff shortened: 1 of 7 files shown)


src/ai/backend/manager/services/session/service.py
```diff
@@ -1652,7 +1653,7 @@ async def enqueue_session(self, action: EnqueueSessionAction) -> EnqueueSessionA
                     ),
                 ),
                 kernel_groups=kernel_groups,
-                handler_options=SessionHandlerOptions(),
```
seedspirit, line 1684: It seems there are a few more places in the code where similar issues exist, so we'll need to check them.

https://github.com/lablup/backend.ai/blob/3c31d425673fc389b7d0a05a090c0671c233a23e/src/ai/backend/manager/registry.py#L1170

https://github.com/lablup/backend.ai/blob/3c31d425673fc389b7d0a05a090c0671c233a23e/src/ai/backend/manager/sokovan/deployment/deployment_draft_builder.py#L124

src/ai/backend/manager/services/session/service.py
```diff
@@ -1557,7 +1556,9 @@ async def enqueue_session(self, action: EnqueueSessionAction) -> EnqueueSessionA
         resource_opts_payload: dict[str, Any] = {}
         if action.resource.shmem is not None:
             resource_opts_payload["shmem"] = BinarySize.from_str(action.resource.shmem)
-        resource_opts = ResourceOpts.model_validate(resource_opts_payload)
+        resource_opts = (
+            ResourceOpts.model_validate(resource_opts_payload) if resource_opts_payload else None
+        )
```
seedspirit, line 1588: Same here

https://github.com/lablup/backend.ai/blob/3c31d425673fc389b7d0a05a090c0671c233a23e/src/ai/backend/manager/registry.py#L1022

https://github.com/lablup/backend.ai/blob/3c31d425673fc389b7d0a05a090c0671c233a23e/src/ai/backend/manager/sokovan/deployment/deployment_draft_builder.py#L74

seedspirit (approved):

HyeockJinKim (approved):

PR #12568, opened 2026-07-07
fix(BA-6725): authorize owner_id delegation in session enqueue

jopemachine:
## Summary

Fixes BA-6725. `POST /v2/sessions/enqueue` accepted an `owner_id` field to create a session on behalf of another user, but never authorized the caller. Any authenticated user could set `owner_id` to an arbitrary user's UUID and create a session **owned by that user** (using their access key, domain, and quota) — a privilege-escalation / impersonation hole.

## Root cause

`EnqueueSessionAction.target_element()` / `scope_id()` referenced the **caller** (`user_id`), so the RBAC scope validator only checked the caller against their own USER scope — which always passes. `owner_id` never entered the check and was resolved later in the service with no authorization.

## Fix

When delegating (`owner_id` set), target the RBAC scope check at the **owner's** USER scope, so the caller must hold session-create permission over the owner via the scope chain. Superadmin bypass is unchanged. No other behavior changes — the scope validator itself is untouched.

## Verification (local server, RBAC enforcement on)

| Case | Before | After |
|---|---|---|
| non-admin + `owner_id`=admin | session created as admin (RUNNING) | **403** |
| non-admin, no `owner_id` | own session | own session (unchanged) |
| superadmin delegation | ok | ok |

## Related

Prerequisite for BA-6719 (adding `owner_id` to more v2 REST endpoints) — the same scope-targeting pattern should be reused there instead of copying the current unguarded behavior.


src/ai/backend/manager/actions/validators/rbac/scope.py
```diff
@@ -27,12 +27,29 @@ def __init__(
...
+        # silently acting on behalf of the owner.
+        if (
+            not enforcement_enabled
+            and action.delegated_owner_id() is not None
+            and not user.is_superadmin
+        ):
+            raise NotEnoughPermission(
+                "Delegating to another user via owner_id requires RBAC "
+                "enforcement to be enabled."
+            )
```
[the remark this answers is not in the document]
jopemachine: I'd prefer not to add this logic if possible. If we conclude that it's unavoidable, I'll add tests for it.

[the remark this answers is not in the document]
jopemachine: We decided not to include this logic

tests/unit/manager/actions/validators/test_rbac_validators.py
```diff
@@ -647,3 +654,78 @@ async def test_empty_targets_returns_empty_result(
 
         assert result.allowed_entities == []
         assert result.denied_entities == []
+
+
+def _make_enqueue_action(
+    *,
+    caller_id: uuid.UUID,
+    owner_id: uuid.UUID,
```
jopemachine: UserID

tests/unit/manager/actions/validators/test_rbac_validators.py
```diff
@@ -647,3 +654,78 @@ async def test_empty_targets_returns_empty_result(
...
+        scheduling=SessionSchedulingSpec(),
+        user_id=caller_id,
+        owner_id=owner_id,
+    )
+
+
+@pytest.fixture
+async def delegation_users(
+    db_with_rbac_tables: ExtendedAsyncSAEngine,
+) -> tuple[uuid.UUID, uuid.UUID]:
```
jopemachine: Don't use tuple, create type

tests/unit/manager/actions/validators/test_rbac_validators.py
```diff
@@ -647,3 +654,78 @@ async def test_empty_targets_returns_empty_result(
...
+            entries=[ResourceSlotEntry(resource_type="cpu", quantity="1")],
+        ),
+        scheduling=SessionSchedulingSpec(),
+        user_id=caller_id,
+        owner_id=owner_id,
+    )
+
+
+@pytest.fixture
+async def delegation_users(
```
jopemachine: Please use the test scenario as the fixture name. The name delegation_users doesn't clearly convey what the test is trying to verify.

fregataa (approved):

PR #12587, opened 2026-07-07
feat(BA-6740): add owner_id delegation to vfolder restore

jopemachine:
Implements **BA-6740** (under epic BA-6727).

## Summary

Add an optional `owner_id` (typed `UserID`) query param to `POST /v2/vfolders/{id}/restore` (`bai vfolder restore --owner-id`) so a caller can restore a trashed vfolder **on behalf of another user**. The endpoint is bodyless, so `owner_id` is a query param.

## Behavior / Authorization

The restore service resolves the acting user via `get_user_by_uuid` and validates the vfolder against that user's ownership. When `owner_id` is set it becomes the acting user, so the restore runs in the owner's context. RBAC still authorizes the **caller** against the target vfolder (single-entity `VFOLDER` scope) — this endpoint already has an RBAC target, so no new scaffolding is added.

## Changes

| Layer | Change |
|---|---|
| DTO | `RestoreVFolderQuery.owner_id: UserID \| None` |
| Handler | `QueryParam[RestoreVFolderQuery]` |
| Adapter | use `owner_id` as the acting `user_uuid` when set |
| SDK | `restore(owner_id)` → query param |
| CLI | `bai vfolder restore --owner-id` |

No action/service change — the restore service already parameterizes on the acting `user_uuid`.

----
📚 Documentation preview 📚: https://sorna--12587.org.readthedocs.build/en/12587/

----
📚 Documentation preview 📚: https://sorna-ko--12587.org.readthedocs.build/ko/12587/

### DIFF
--- a/docs/manager/graphql-reference/v2-schema.graphql
+++ b/docs/manager/graphql-reference/v2-schema.graphql
@@ -7959,7 +7959,7 @@ type Mutation {
   """
   Added in 26.4.4. Restore a trashed virtual folder. RBAC enforced via scope chain.
   """
-  restoreVFolder(vfolderId: UUID!, ownerId: UUID = null): RestoreVFolderPayload
+  restoreVFolder(vfolderId: UUID!, options: RestoreVFolderOptions = null): RestoreVFolderPayload
 
   """Added in 26.4.2. Deploy a deployment directly from a model VFolder."""
   deployVfolderV2(vfolderId: UUID!, input: DeployVFolderV2Input!): DeployVFolderV2Payload
@@ -11252,6 +11252,12 @@ Added in 25.15.0. Response payload for artifact restoration operations. Contains
 type RestoreArtifactsPayload {
   """List of restored artifacts."""
   artifacts: [Artifact!]!
+}
+
+"""Added in 26.8.0. Options for restoring a virtual folder from trash."""
+input RestoreVFolderOptions {
+  """Delegated owner user UUID. Restore the vfolder on behalf of this user."""
+  ownerId: UUID = null
 }
 
 """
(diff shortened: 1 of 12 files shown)


jopemachine:
Note on `vfolder_id` typing: the added `owner_id` field uses the domain identifier type (`UserID`), but `vfolder_id` here stays raw `UUID` to match the existing convention across the whole v2 vfolder layer — `VFolderIdPathParam.vfolder_id`, all ~13 adapter methods, and the action `vfolder_uuid` fields are all raw `UUID` today (`VFolderUUID` is currently only used on the `VFolderData.id` data type).

Switching `vfolder_id` to `VFolderUUID` only here would break at the boundary (the path param yields a `UUID`, which is not assignable to the `VFolderUUID` NewType) and would leave this method inconsistent with the rest of the file. Migrating the whole v2 vfolder layer to `VFolderUUID` is worth doing as a separate, dedicated cleanup PR rather than piecemeal in this delegation PR.

docs/manager/graphql-reference/v2-schema.graphql
```diff
@@ -7959,7 +7959,7 @@ type Mutation {
   """
   Added in 26.4.4. Restore a trashed virtual folder. RBAC enforced via scope chain.
   """
-  restoreVFolder(vfolderId: UUID!): RestoreVFolderPayload
+  restoreVFolder(vfolderId: UUID!, ownerId: UUID = null): RestoreVFolderPayload
```
jopemachine: Create RestoreVFolderOptions type

jopemachine:
As the requirements for this feature request have changed significantly from the original design, this work is no longer needed. Therefore, this PR will be closed.

PR #12621, opened 2026-07-07
fix(BA-6726): enforce in-use guard on vfolder purge paths

HyeockJinKim:
## Summary
- Soft delete blocks removing a vfolder mounted by a live session, but **purge had no such guard** — a `ready` / mounted / model-service vfolder could be physically destroyed, causing irreversible data loss and dangling endpoints (hit in QA on a model-service model vfolder).
- Add an in-use guard to **both** purge paths (v1 `purge_vfolder`, v2 `delete_vfolders_forever`): reject when the vfolder is mounted by a live session/kernel, referenced as `model` or `extra_mounts` by an active endpoint, or not in a purgable status. Gated by an explicit `force` flag (default off).
- **v1 schema is unchanged** — `force` is surfaced only on the v2 surface (DTO / GraphQL / action / SDK / CLI `--force`); v1's `force_delete` endpoint stays the escape hatch.
- Centralize every vfolder reference in a purge-guard registry (`repositories/vfolder/purge_guards.py`) and add a schema-introspection **completeness test** so a newly added referencing column (FK or `VFolderMount`/`MountInfoEntry` JSONB) cannot silently bypass the guard. TEMPORARY until the soft references are normalized into FK junction tables.

## Test plan
- [ ] Repository guard tests: live-session mount / live-kernel mount / active-endpoint `model` / active-endpoint `extra_mounts` / non-purgable status are all rejected; `force=True` bypasses each; destroyed endpoint & unreferenced delete-complete vfolder succeed (no regression).
- [ ] Completeness test asserts every FK to `vfolders.id` and every `VFolderMount`/`MountInfoEntry` column is registered (verified it fails when a registration is removed).
- [ ] Service test: `force_delete` forwards `force=True`; `purge_v2` forwards `action.force`.
- [ ] CI: `pants check` + `pants test`.

Resolves BA-6726

----
📚 Documentation preview 📚: https://sorna--12621.org.readthedocs.build/en/12621/

----
📚 Documentation preview 📚: https://sorna-ko--12621.org.readthedocs.build/ko/12621/

### DIFF
--- a/src/ai/backend/manager/repositories/vfolder/purge_guards.py
+++ b/src/ai/backend/manager/repositories/vfolder/purge_guards.py
@@ -1,0 +1,158 @@
+"""Registry of active vfolder references consulted by the purge in-use guard.
+
+``vfolders.id`` is referenced in two forms:
+
+- **Hard FK columns** — discoverable from the schema and (partly) enforced by
+  the DB: ``model_cards.vfolder`` (RESTRICT), ``deployment_revisions.model``
+  (SET NULL), and the ``vfolder_*`` CASCADE child tables.
+- **Soft JSONB references** that embed a vfolder id but are *not* foreign keys,
+  so schema metadata and the DB cannot see them:
+  ``sessions.vfolder_mounts`` / ``kernels.vfolder_mounts`` (``VFolderMount``)
+  and ``deployment_revisions.extra_mounts`` (``MountInfoEntry``).
+
+The purge in-use guard must reject a purge while any *active* reference exists,
+but the DB cannot enforce that for the soft references (and ``SET NULL`` FKs are
+silently nulled rather than blocked). Both categories are therefore checked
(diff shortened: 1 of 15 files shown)


src/ai/backend/manager/repositories/vfolder/purge_guards.py
```diff
@@ -0,0 +1,158 @@
...
+        referrer_ids = await check.query(session, vfolder_row)
+        if referrer_ids:
+            hits.append(
+                VFolderReferenceHit(
+                    source=check.source,
+                    describe=check.describe,
+                    referrer_ids=[str(rid) for rid in referrer_ids],
+                )
+            )
+    return hits
```
seedspirit, line 158: Is there a reason why this was implemented as a regular function rather than a VFolderRepository?

fregataa (approved):
we need vfolder API v3

PR #12624, opened 2026-07-07
feat(BA-6708): add non-nullable resource_group_id column to agents

seedspirit:
## Summary
- Add `agents.resource_group_id` (UUID FK to `scaling_groups.id`, indexed) alongside the legacy name column, backfilled from `agents.scaling_group` and made NOT NULL in the same migration — following the kernel-row pattern (#12557) ahead of the resource group PK swap (BA-6050).
- Heartbeat ingest now resolves the reported scaling group name to its id inside the existing existence-check query (no extra round trip) and writes both columns in the upsert; unknown names are rejected with `ScalingGroupNotFound` as before.
- The legacy `modify_agent` mutation resolves the new scaling group name to its id and updates both columns atomically; a nonexistent name now returns `ok=False` before the agent-side RPC instead of reconfiguring the agent and then failing on the FK.
- The agent-side wire format (`AgentInfo.scaling_group`), TOML config, and etcd sgroup config scopes stay name-based — only the manager ingest resolves.

## Test plan
- [x] `sync_agent_heartbeat` insert path sets `resource_group_id` matching the scaling group name (new assertion)
- [x] heartbeat reporting a different scaling group updates both name and id columns consistently (new test)
- [x] heartbeat with unknown scaling group still raises `ScalingGroupNotFound` (existing test, unchanged)
- [x] migration verified on a scratch DB: backfill joins name→id with zero NULLs, downgrade/re-upgrade clean
- [x] `pants fmt/lint/check` and all affected unit/component test suites pass

Resolves BA-6708


src/ai/backend/manager/api/gql_legacy/agent.py
```diff
@@ -914,7 +915,15 @@ async def mutate(
...
+        scaling_group = data.get("scaling_group")
+        if scaling_group is not None:
+            async with graph_ctx.db.begin_readonly_read_committed() as conn:
+                resource_group_id = await conn.scalar(
+                    sa.select(ScalingGroupRow.id).where(ScalingGroupRow.name == scaling_group)
+                )
+            if resource_group_id is None:
+                return cls(False, f"no such scaling group: {scaling_group}")
+            data["resource_group_id"] = resource_group_id
             await graph_ctx.registry.update_scaling_group(AgentId(id), scaling_group)
```
HyeockJinKim, line 927: It does feel a bit out of place to access and handle the database separately here.

HyeockJinKim (approved):
The names of the scaling groups and resource groups seem a bit inconsistent, but we'll address that later.

PR #12637, opened 2026-07-07
feat(BA-6582): evaluate prepared idle checks into verdicts in the handler

seedspirit:
## Summary
- Add the idle checker contract: `IdleChecker[StateT]` (`prepare`/`check_idle`), `IdleCheckerDependencies` (I/O client bundle placeholder), and `checker_for` static dispatch — concrete implementations register as the checker-logic stories land
- `IdleCheckSource` fetches the repository batch and drives each checker type's `prepare` once over all of its definitions (`CheckerAssignment` batches; checker-owned I/O confined to the Source phase), composing judgment-ready `PreparedTarget`s
- The reconcile handler evaluates prepared checks purely (no DB/Valkey/Prometheus reads) and emits one `IdleVerdict(session_id, checker_ids)` per idle session listing **every** checker that judged it idle, so the terminating checkers are all traceable by the applier

## Test plan
- [x] `pants test tests/unit/manager/sokovan/idle_check::` — per-type batched prepare, per-session composition, and all-checker verdict recording
- [x] `pants lint` / `pants check` on changed files

Resolves BA-6582


src/ai/backend/manager/sokovan/idle_check/preparer.py
```diff
@@ -0,0 +1,74 @@
...
+                definitions[bound.checker.checker_id] = bound.checker
+                sessions_by_checker[bound.checker.checker_id].append(target.session)
+        requests_by_type: defaultdict[CheckerType, list[PrepareRequest]] = defaultdict(list)
+        for checker_id, definition in definitions.items():
+            requests_by_type[definition.checker_type].append(
+                PrepareRequest(
+                    definition=definition,
+                    sessions=sessions_by_checker[checker_id],
+                )
+            )
```
seedspirit: I plan to work on changing the repository's return data in a subsequent PR.

src/ai/backend/manager/sokovan/idle_check/checkers/base.py
```diff
@@ -0,0 +1,51 @@
...
+        self,
+        context: IdleCheckContext,
+        requests: Sequence[PrepareRequest],
+    ) -> Mapping[IdleCheckerID, IdleCheckerState]:
+        """Called once per tick with every definition of this type.
+
+        Batch the I/O across all requests and return one state per definition;
+        capture everything check_idle needs into the states here.
+        """
+        raise NotImplementedError
```
fregataa, line 45: Seems IdleCheckerState type should be generic here

src/ai/backend/manager/sokovan/idle_check/checkers/factory.py
```diff
@@ -0,0 +1,16 @@
...
+
+from ai.backend.common.data.idle_checker.types import CheckerType
+from ai.backend.manager.sokovan.idle_check.checkers.base import IdleChecker
+
+# Concrete checker types register here as their stories land.
+_CHECKERS: Final[Mapping[CheckerType, IdleChecker]] = {}
+
+
+def checker_for(checker_type: CheckerType) -> IdleChecker | None:
+    return _CHECKERS.get(checker_type)
```
fregataa: Is it possible to have multiple checker types can map to the same idle checker or a checker type has no idle checker?

seedspirit: It does assume that there is only one IdleChecker per checker type.

src/ai/backend/manager/sokovan/idle_check/preparer.py
HyeockJinKim: I didn't understand why this was necessary.

src/ai/backend/manager/sokovan/stages/idle_check.py
```diff
@@ -48,7 +49,7 @@ def build_idle_check_stage(
     )
     stage = ReconcilerStage(
         handler=IdleCheckReconcileHandler(),
-        source=IdleCheckSource(idle_checker_repository),
+        source=IdleCheckSource(idle_checker_repository, IdleCheckContext()),
```
HyeockJinKim: Why is IdleCheckContext necessary?

src/ai/backend/manager/sokovan/idle_check/types.py
```diff
@@ -39,14 +39,33 @@ class IdleCheckTargetStatuses(BaseReconcilerTargetStatuses):
...
+    def check_idle(self, session_id: SessionId) -> bool:
+        return self.checker.check_idle(session_id, self.state)
+
+
+@dataclass(frozen=True)
+class PreparedTarget:
+    """One session and its judgment-ready checkers, in resolved order."""
+
+    session_id: SessionId
+    checkers: Sequence[CheckerWithState]
```
HyeockJinKim: There seems to be something wrong with the type definition.

src/ai/backend/manager/sokovan/idle_check/handlers/reconcile.py
```diff
@@ -2,15 +2,24 @@
...
     async def execute(self, reconcile_info: IdleCheckReconcileInfo) -> IdleCheckResult:
-        # Placeholder: idle judgment lands in the checker-logic stories.
-        return IdleCheckResult()
+        idle_session_ids: list[SessionId] = []
+        for target in reconcile_info.targets:
+            # First idle verdict in resolved order wins; at most one termination per session.
+            for checker_with_state in target.checkers:
+                if checker_with_state.check_idle(target.session_id):
+                    idle_session_ids.append(target.session_id)
+                    break
```
fregataa: How about iterating all idle checkers?

src/ai/backend/manager/sokovan/idle_check/checkers/base.py
```diff
@@ -0,0 +1,51 @@
...
+    sessions: Sequence[IdleCheckSession]
+
+
+class IdleChecker[StateT: IdleCheckerState](ABC):
+    """Stateless per-``CheckerType`` judgment behavior over its own state type."""
+
+    @abstractmethod
+    async def prepare(
+        self,
+        context: IdleCheckContext,
```
fregataa: IdleCheckContext can be a private attribute rather than passing through argument. Is there a reason?

src/ai/backend/manager/sokovan/idle_check/checkers/base.py
```diff
@@ -0,0 +1,51 @@
...
+from ai.backend.manager.repositories.idle_checker.types import IdleCheckerDefinitionData
+
+
+@dataclass(frozen=True)
+class IdleCheckContext:
+    """I/O clients used by ``prepare``; filled in by the checker-logic stories."""
+
+
+class IdleCheckerState:
+    """Marker for the state a checker prepares; each checker defines its own shape."""
```
fregataa: Is this the same with idle checker spec?

src/ai/backend/manager/sokovan/stages/idle_check.py
seedspirit: Since ValkeyClient, PrometheusClient, and others are expected to be included in IdleCheckContext, it seems essential.

src/ai/backend/manager/sokovan/idle_check/checkers/base.py
seedspirit: Since ValkeyClient, PrometheusClient, and others are expected to be included in IdleCheckContext, it seems essential need to be injected to each checker

fregataa (dismissed):

src/ai/backend/manager/sokovan/idle_check/checkers/factory.py
```diff
@@ -0,0 +1,16 @@
...
+
+from ai.backend.common.data.idle_checker.types import CheckerType
+from ai.backend.manager.sokovan.idle_check.checkers.base import IdleChecker
+
+# Concrete checker types register here as their stories land.
+_CHECKERS: Final[Mapping[CheckerType, IdleChecker[Any]]] = {}
+
+
+def checker_for(checker_type: CheckerType) -> IdleChecker[Any] | None:
+    return _CHECKERS.get(checker_type)
```
HyeockJinKim: Why is this being classified as a global issue...

src/ai/backend/manager/sokovan/idle_check/handlers/reconcile.py
```diff
@@ -2,15 +2,31 @@
...
 
 
 class IdleCheckReconcileHandler(ReconcilerHandler[IdleCheckReconcileInfo, IdleCheckResult]):
+    """Turns prepared checks into idle verdicts; performs no external I/O."""
+
     @override
     async def execute(self, reconcile_info: IdleCheckReconcileInfo) -> IdleCheckResult:
-        # Placeholder: idle judgment lands in the checker-logic stories.
-        return IdleCheckResult()
+        verdicts: list[IdleVerdict] = []
```
HyeockJinKim: It doesn't seem like a commonly used word, but Claude has been using terms like this a lot lately, so it feels a bit familiar.

src/ai/backend/manager/sokovan/idle_check/checkers/base.py
```diff
@@ -0,0 +1,51 @@
...
+
+        Batch the I/O across all assignments and return one state per definition;
+        capture everything check_idle needs into the states here.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def check_idle(self, session_id: SessionId, state: StateT) -> bool:
+        """Judge one session from prepared state alone; True means idle."""
+        raise NotImplementedError
```
HyeockJinKim: I do think it would be a bit neater to handle it in batches.

src/ai/backend/manager/sokovan/idle_check/source.py
```diff
@@ -16,15 +29,67 @@ class IdleCheckSource(
...
+        for checker_id, definition in definitions.items():
+            assignments_by_type[definition.checker_type].append(
+                CheckerAssignment(
+                    definition=definition,
+                    sessions=sessions_by_checker[checker_id],
+                )
+            )
+        prepared_by_id = await self._prepare_checkers(assignments_by_type)
+        prepared_targets: list[PreparedTarget] = []
+        for target in batch.targets:
```
HyeockJinKim: The process of reading the data is already complicated, and with the additional logic involved, it seems to be causing some confusion.

src/ai/backend/manager/sokovan/idle_check/checkers/base.py
HyeockJinKim: There will likely be only one implementation of the “idle checker” object. (Of course, it will behave differently depending on the spec values it receives.)
In that situation, if a single idle checker directly handles a single session, the upper layers would become too complex. (Or I wonder if batch processing might be possible—for example, reading data from Redis in batches.)

In that case, we’d need to batch processing based on one of these entities. Rather than batching idle checkers based on sessions, I think it’s more appropriate to batch sessions based on idle checkers. (The idle checker is the reference point for handling operations; a session cannot serve as that reference.)

So, wouldn’t it be appropriate to have the IdleChecker handle the batching? In that context, the DataClass you implemented below—where a single session ID is associated with multiple IdleCheckers—feels unnatural.

src/ai/backend/manager/sokovan/idle_check/source.py
```diff
@@ -16,15 +29,67 @@ class IdleCheckSource(
...
+        prepared_by_id: dict[IdleCheckerID, CheckerWithState] = {}
+        for checker_type, assignments in assignments_by_type.items():
+            checker = checker_for(checker_type)
+            if checker is None:
+                continue
+            states = await checker.prepare(self._dependencies, assignments)
+            for checker_id, state in states.items():
+                prepared_by_id[checker_id] = CheckerWithState(
+                    checker_id=checker_id, checker=checker, state=state
+                )
```
HyeockJinKim: It does seem a bit odd to do all the preparation at the reader level.
This is actually business logic that needs to be handled, but it doesn’t seem much different from having the repository handle everything.
Mapping the session to the idle checker seems like business logic.

src/ai/backend/manager/sokovan/idle_check/source.py
```diff
@@ -16,15 +29,67 @@ class IdleCheckSource(
...
 ):
     _repository: IdleCheckerRepository
+    _dependencies: IdleCheckerDependencies
 
-    def __init__(self, repository: IdleCheckerRepository) -> None:
+    def __init__(
+        self, repository: IdleCheckerRepository, dependencies: IdleCheckerDependencies
+    ) -> None:
         self._repository = repository
+        self._dependencies = dependencies
```
HyeockJinKim: There's no need to make the source that complicated.

src/ai/backend/manager/sokovan/idle_check/types.py
```diff
@@ -77,13 +98,21 @@ def policy_resolver(self) -> HandlerPolicyResolver:
         return self.handler_policy
 
 
+@dataclass(frozen=True)
+class IdleVerdict:
+    """One session judged idle and every checker that judged it so."""
+
+    session_id: SessionId
+    checker_ids: Sequence[IdleCheckerID]
```
HyeockJinKim: 1. Could you do something about the term “verdict,” too? lol...
2. I think it would be best to just read the “checkers” and “sessions” from the source code, and then figure out how to filter those values in the handler and pass them to the idlechecker accordingly.
3. If you're referring to the results from Idlechecker, you'll need the message as well. (It needs to be included in the history.)

HyeockJinKim (changes requested):

src/ai/backend/manager/sokovan/idle_check/checkers/base.py
```diff
@@ -0,0 +1,53 @@
...
+    async def prepare(
+        self,
+        assignments: Sequence[CheckerAssignment],
+    ) -> Mapping[IdleCheckerID, Mapping[SessionId, StateT]]:
+        """Called once per tick with every definition of this type.
+
+        Batch the I/O across all assignments (session ids key the Valkey/DB reads) and
+        bake everything ``judge`` needs into per-session states.
+        """
+        raise NotImplementedError
```
HyeockJinKim, line 45: What is this used for?

HyeockJinKim: It seems like a good idea to structure it as a dataclass.

src/ai/backend/manager/sokovan/idle_check/checkers/base.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @abstractmethod
+    def judge(self, session_states: Mapping[SessionId, StateT]) -> Sequence[IdleJudgment]:
+        """Judge one definition's sessions from prepared states alone; no I/O."""
+        raise NotImplementedError
```
HyeockJinKim: Should this be separated into an abstract method?

src/ai/backend/manager/sokovan/idle_check/handlers/reconcile.py
```diff
@@ -1,16 +1,81 @@
...
     async def execute(self, reconcile_info: IdleCheckReconcileInfo) -> IdleCheckResult:
-        # Placeholder: idle judgment lands in the checker-logic stories.
-        return IdleCheckResult()
+        assignments_by_type = self._assignments_by_type(reconcile_info.batch)
+        reasons_by_session: defaultdict[SessionId, list[IdleReason]] = defaultdict(list)
+        for checker_type, assignments in assignments_by_type.items():
+            checker = self._checkers.get(checker_type)
+            if checker is None:
+                continue
+            judgments = await checker.judge(assignments)
```
fregataa, line 44: In which case `judge()` call should be awaited?

seedspirit: Since it involves operations such as Valkey and DB IO, you must use `await`.

jopemachine (approved):

jopemachine:
There is a requested change by @HyeockJinKim, could you check this?

tests/unit/manager/sokovan/idle_check/test_handler.py
```diff
@@ -0,0 +1,320 @@
...
+    async def test_batches_judgments_by_checker_type(
+        self,
+        handler: IdleCheckReconcileHandler,
+        first_session_id: SessionId,
+        second_session_id: SessionId,
+        lifetime_bound: BoundCheckerData,
+        second_lifetime_bound: BoundCheckerData,
+        network_bound: BoundCheckerData,
+        lifetime_checker: FakeChecker,
+        network_checker: FakeChecker,
```
jopemachine, line 164: Rather than creating every individual fixture, I think it would improve readability to inject the test scenario itself as a fixture instead.

This isn't a change that necessarily needs to be made though

HyeockJinKim (approved):

PR #12638, opened 2026-07-07
fix(BA-6769): expose BinarySizeInfo byte count as string field 'expr'

HyeockJinKim:
## Summary
- `BinarySizeInfo` exposed the byte count as `value: int`, which GraphQL maps to the 32-bit `Int` scalar — byte sizes over 2 GiB failed response serialization (`Int cannot represent non 32-bit signed integer value`). Input/DB paths were already fine; only the response scalar was narrowed.
- Replace `value: int` with `expr: str` (exact decimal byte string, mirroring `BinarySizeInput.expr`) so arbitrarily large sizes serialize as strings across GraphQL and REST v2. Adds `BinarySize.to_size_info()` to build the DTO from the source type.
- Also fixes a falsy-zero guard that dropped an explicit `shared_memory` of `0`, and regenerates the GraphQL reference schemas.

**BREAKING**: `BinarySizeInfo.value` (Int) is now `expr` (String). Affects resource policy `maxQuotaScopeSize`, vfolder `usedBytes`/`maxSize`, and resource preset/allocation `sharedMemory`. Clients that read these as numbers must read `expr` as a string (parse to int where needed). WebUI consumers need a coordinated migration.

## Test plan
- [x] `BinarySizeInfo` serializes byte counts over 2 GiB and 2^53 as JSON strings (round-trip)
- [x] `BinarySizeInfoGQL.expr` maps to GraphQL `String`, not `Int`
- [x] `BinarySize.to_size_info()` returns exact `expr` (not humanized) incl. `BinarySize` inputs and `0`
- [x] `shared_memory == 0` is emitted, not dropped as unset
- [x] Input path unchanged: policy create with `expr="1073741824"` still succeeds
- [x] GraphQL reference schemas (`v2-schema.graphql`, `supergraph.graphql`) show `expr: String!`
- [x] `pants fmt/fix/lint/check/test` pass

Resolves BA-6769

----
📚 Documentation preview 📚: https://sorna--12638.org.readthedocs.build/en/12638/

----
📚 Documentation preview 📚: https://sorna-ko--12638.org.readthedocs.build/ko/12638/

### DIFF
--- a/changes/12638.breaking.md
+++ b/changes/12638.breaking.md
@@ -1,0 +1,1 @@
+The GraphQL and REST v2 `BinarySizeInfo` type now returns the exact byte count as a string field `expr` (replacing the integer `value`), lifting the 2 GiB GraphQL `Int` limit that made larger sizes fail to serialize; clients must read `expr` as a string on resource policy `maxQuotaScopeSize`, vfolder `usedBytes`/`maxSize`, and resource preset/allocation `sharedMemory`.
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -2048,8 +2048,10 @@ Added in 26.4.2. Binary size with both raw bytes and human-readable format.
 type BinarySizeInfo
   @join__type(graph: STRAWBERRY)
 {
-  """Size in bytes."""
-  value: Int!
+  """
+  Exact size in bytes as a decimal string (e.g., '1073741824'); accepted as BinarySizeInput.expr.
+  """
+  expr: String!
 
   """Size in human-readable format (e.g., '1g', '512m')."""
   display: String!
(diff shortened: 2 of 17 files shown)


tests/unit/manager/models/conftest.py
```diff
@@ -1,13 +1,29 @@
...
+# relationships across the whole model graph (e.g. KernelRow -> "SessionRow"),
+# resolved at global mapper-configuration time; an isolated pytest batch would
+# otherwise import only a subset, leaving some relationship targets unregistered.
+# Importing every model module (mirrors the alembic env bootstrap) registers them
+# all. Subpackages without Row definitions are skipped.
+_MODEL_SUBPACKAGES_WITHOUT_ROWS = {"alembic", "hasher", "minilang", "rbac"}
+for _module_info in pkgutil.iter_modules(ai.backend.manager.models.__path__):
+    if _module_info.ispkg and _module_info.name in _MODEL_SUBPACKAGES_WITHOUT_ROWS:
+        continue
+    importlib.import_module(f"ai.backend.manager.models.{_module_info.name}")
```
seedspirit: Is this necessary? If there's a problem, is there no other way to solve it not importing whole models package?

HyeockJinKim: I just left this in there for now, figuring I wouldn't worry too much about it until I finished the "relationship" task.

seedspirit (approved):

PR #12655, opened 2026-07-08
refactor(BA-6709): switch scheduler queries to resource_group_id filters

seedspirit:
## Summary
- Replace scaling-group-name based filters/joins in the scheduler path with `resource_group_id`, so the scheduler no longer depends on the name columns (prerequisite for dropping `sessions.scaling_group_name` in the follow-up story).
- `get_all_scaling_groups()` now returns resource group ids; the coordinator loops, `SessionLifecycleHandler`/`KernelLifecycleHandler` `execute()`, and `KernelObserver` ABCs take `ResourceGroupID`. Pending-session fetch, scheduler_opts join, sessions/kernels-for-handler fetches, agent fetch, and the kernel occupancy snapshot all filter by `resource_group_id`.
- External surfaces are unchanged: the provisioner still receives the scaling group **name** via `SchedulingData.scaling_group` (Prometheus `scaling_group` labels and Valkey pending-queue keys keep name), `FairShareObserver` resolves id→name once per observation since fair share records are name-keyed, and existing name-based `SessionConditions`/`KernelConditions` factories used by the API are untouched (new `by_resource_group_id` factories added alongside).

## Test plan
- [x] `pants lint check --changed-since=origin/main --changed-dependents=direct` passes
- [x] `pants test tests/unit/manager/sokovan:: tests/unit/manager/repositories/scheduler:: tests/unit/manager/models:: tests/unit/manager/repositories/fair_share:: tests/unit/manager/repositories/scaling_group::` — 130 targets pass
- [ ] Verify scheduling behavior against the live dev stack (session enqueue → schedule → start)

Resolves BA-6709


src/ai/backend/manager/sokovan/scheduler/handlers/lifecycle/schedule_sessions.py
```diff
@@ -133,7 +134,7 @@ async def execute(
         # Delegate to Provisioner with pre-fetched data
         provision_time = await self._repository.get_db_now()
         schedule_result = await self._provisioner.schedule_scaling_group(
-            scaling_group, scheduling_data, provision_time
+            scheduling_data.scaling_group.name, scheduling_data, provision_time
```
HyeockJinKim: Why is "name" used here?

src/ai/backend/manager/sokovan/scheduler/provisioner/provisioner.py
```diff
@@ -225,9 +225,7 @@ async def schedule_scaling_group(
                 sequencer.name, success_detail=sequencer.success_message()
             ),
         ):
-            sequenced_workloads = await sequencer.sequence(
-                scaling_group, system_snapshot, workloads
-            )
+            sequenced_workloads = await sequencer.sequence(sg_info.name, system_snapshot, workloads)
```
seedspirit, line 228: Intentionally passes the **name** here: the fair-share sequencer resolves factors via `get_user_fair_share_factors_batch(resource_group, ...)`, and the fair share tables key `resource_group` by name. Passing the id would make every factor lookup miss and silently degrade fair-share ordering to the default factor.

src/ai/backend/manager/sokovan/scheduler/provisioner/provisioner.py
```diff
@@ -290,26 +288,26 @@ async def schedule_scaling_group(
...
             allocations=session_allocations,
             failures=scheduling_failures,
         )
-        with self._phase_metrics.measure_phase("scheduler", scaling_group, "allocation"):
+        with self._phase_metrics.measure_phase("scheduler", resource_group_id, "allocation"):
             scheduled_session_ids = await self._allocator.allocate(batch)
 
         failure_ids = [f.session_id for f in scheduling_failures]
-        await self._valkey_schedule.set_pending_queue(scaling_group, failure_ids)
+        await self._valkey_schedule.set_pending_queue(sg_info.name, failure_ids)
```
seedspirit, line 302: Intentionally keeps the **name** as the Valkey key (`pending_queue:{name}`): the only reader is the legacy GQL `session_pending_queue` resolver, whose `resource_group_id` argument is a name string in practice. Switching this key to the uuid breaks that public query unless the GQL side changes in the same release.

src/ai/backend/manager/sokovan/scheduler/handlers/observer/fair_share.py
```diff
@@ -121,26 +122,30 @@ async def observe(
...
+            resource_group_id,
             len(kernels),
         )
         if not kernels:
             log.debug("[FairShareObserver] No kernels to observe, returning early")
             return ObservationResult(observed_count=0)
 
         now = await self._scheduler_repository.get_db_now()
+        # Fair share records and factor queries are keyed by name.
+        resource_group_name = await self._scheduler_repository.get_resource_group_name_by_id(
```
seedspirit, line 142: One id→name resolution per observation tick: fair share usage records and factor/bucket rows store the resource group **name**, so the observer needs the name for all downstream writes and the phase-2 factor query.

seedspirit (commented):
Notes on the spots that intentionally keep the scaling group **name** (everything else in the scheduler path now runs on `resource_group_id`). These will flip to id only when their name-keyed storage/readers migrate.

src/ai/backend/manager/data/sokovan/workload.py
```diff
@@ -89,6 +90,8 @@ class SessionWorkload:
     domain_name: str
     # Scaling group name
     scaling_group: str
+    # Scaling group id
+    resource_group_id: ResourceGroupID
```
HyeockJinKim, line 94: This keeps getting written incorrectly, but I'll just let it slide for now...

HyeockJinKim (approved):

PR #12660, opened 2026-07-08
feat(BA-6781): add triggered_user() context and resolve X-BackendAI-Act-As in auth middleware

jopemachine:
## Summary

Part of **BEP-1058 (Super-Admin User Impersonation)**, issue **BA-6781**.

Adds a **trigger-user** identity context alongside the existing effective-user context, and resolves the `X-BackendAI-Act-As` impersonation header globally in the REST auth middleware.

### (a) Accessor — `common/contexts/user.py`
- New `triggered_user() -> UserData | None` and `with_triggered_user()` over a separate `ContextVar`, mirroring `current_user()`.
- `current_user()` docstring clarified: it is the **effective (scope) subject**, not necessarily the caller.

### (b) Middleware — `manager/api/rest/middleware/auth.py`
Right after authentication (HMAC/JWT/hook), resolve `X-BackendAI-Act-As` (target user UUID). The two identities are modeled directly: `triggered_user` is always the authenticated caller, and `current_user` (effective) defaults to the caller and is **overridden** by the act-as target.

| Case | Behavior |
|------|----------|
| No header | `current_user = triggered_user = authenticated` |
| Header, caller superadmin, valid target | `current_user = target`, `triggered_user = authenticated` |
| Header, caller **not** superadmin | reject — `403 InsufficientPrivilege` |
| Invalid UUID | reject — `400 InvalidAuthParameters` |
| Target not found | reject — `404 UserNotFound` |
| Target exists but misconfigured (no role/domain) | reject — `500 InternalServerError` |

Applied globally so the whole request runs as the target (fail-closed). The target is loaded by UUID through the `UserRow` ORM under `READ COMMITTED`.

### (c) `owner_access_key` — no special handling
This PR does **not** reject a request that carries both `X-BackendAI-Act-As` and `owner_access_key`. Impersonation only overrides the `current_user()` context var, while `owner_access_key`-based paths (session/vfolder/…) derive the requester from `request["user"]`/`request["keypair"]`, so on those paths `owner_access_key` (owner_id) wins and the act-as header has no effect. This is documented in the BEP (§4.4 + Decision Log) and flagged in a PR comment for reviewer visibility.

## Success Criteria
- [x] `triggered_user()` returns the value inside `with_triggered_user(u)`, `None` outside (no `LookupError`)
- [x] Request without the header: behavior unchanged (`current_user() == triggered_user() == authenticated`)
- [x] Regular user sets the header → 403 reject
- [x] Superadmin with a valid UUID → `current_user()` = target, `triggered_user()` = superadmin
- [x] Non-existent UUID → reject

## Tests
- `tests/unit/common/contexts/test_user.py` — `triggered_user()` / `with_triggered_user()` accessor contract.
- `tests/unit/manager/api/test_impersonation.py` — `TestResolveEffectiveUser` (no-header → caller, superadmin → target, reject cases) and `TestSetupUserContext` (current/triggered wiring via the actual context vars).

## Notes
- Downstream RBAC/scoping that reads `current_user()` becomes target-based automatically while impersonating; paths that read `request["user"]`/`request["keypair"]` directly are unaffected (see (c)).
- Follow-ups (separate issues): BA-6782 (event-dispatcher propagation of `triggered_user()`), BA-6783 (audit `acted_as` column).


src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -645,29 +653,110 @@ async def _authenticate_via_hook(
...
+        ).where(users.c.uuid == target_id)
+        result = await conn.execute(query)
+        return result.first()
+
+
+async def _load_target_user_data(db: ExtendedAsyncSAEngine, target_id: uuid.UUID) -> UserData:
+    """Load a target user's ``UserData`` for impersonation. Raises if not found."""
+    row = await execute_with_retry(functools.partial(_query_target_user, db, target_id))
+    if row is None:
+        raise ObjectNotFound(object_name="impersonation target user")
```
jopemachine: UserNotFoundError

src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -66,6 +69,11 @@
 
 log: Final = BraceStyleAdapter(logging.getLogger(__spec__.name))
 
+# Impersonation signal (BEP-1058): a super admin may run a request under a
+# target user's permission context by supplying the target's UUID here.
+ACT_AS_HEADER: Final = "X-BackendAI-Act-As"
+_OWNER_ACCESS_KEY_PARAM: Final = "owner_access_key"
```
HyeockJinKim: Why is _OWNER_ACCESS_KEY_PARAM included?

src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -645,29 +653,110 @@ async def _authenticate_via_hook(
...
+            return False
+        if isinstance(body, dict) and body.get(_OWNER_ACCESS_KEY_PARAM):
+            return True
+    return False
+
+
+async def _resolve_impersonation(
+    request: web.Request,
+    db: ExtendedAsyncSAEngine,
+) -> tuple[UserData | None, UserData | None]:
```
jopemachine: Create a type, not using tuple

src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -645,29 +653,110 @@ async def _authenticate_via_hook(
...
+    if authenticated is None or not authenticated.is_superadmin:
+        raise InsufficientPrivilege(f"Only superadmin may use {ACT_AS_HEADER}")
+    if await _request_has_owner_access_key(request):
+        raise InvalidAuthParameters(
+            f"{ACT_AS_HEADER} cannot be combined with {_OWNER_ACCESS_KEY_PARAM}"
+        )
+    try:
+        target_id = uuid.UUID(raw_target)
+    except ValueError as e:
+        raise InvalidAuthParameters(f"{ACT_AS_HEADER} must be a valid user UUID") from e
```
HyeockJinKim: Rather than this, I had in mind using the existing user information as the basis for processing when `current_user` is passed in, so I don't think the code would be this long. Please take a look at it.

src/ai/backend/manager/api/rest/middleware/auth.py
jopemachine: Don't make these

tests/unit/manager/api/test_impersonation.py
```diff
@@ -0,0 +1,109 @@
...
+        request["is_superadmin"] = role == UserRole.SUPERADMIN
+        request["user"] = {
+            "uuid": uuid.uuid4(),
+            "role": role.value,
+            "domain_name": "default",
+        }
+    return request
+
+
+@pytest.mark.asyncio
```
jopemachine: Don't use @pytest.mark.asyncio

src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -66,6 +68,10 @@
 
 log: Final = BraceStyleAdapter(logging.getLogger(__spec__.name))
 
+# Impersonation signal (BEP-1058): a super admin may run a request under a
```
jopemachine: Don't mention BEP number in comments

src/ai/backend/manager/utils.py
```diff
@@ -20,12 +21,37 @@
...
     target_role: UserRole,
     target_domain: str,
 ) -> bool:
+    # Defense-in-depth: any acting-on-behalf-of delegation (owner_access_key,
+    # owner_user_email) is refused while impersonating, regardless of read site.
+    if is_impersonating():
+        raise InvalidAPIParameters(
+            "Acting on behalf of another user is not allowed while impersonating "
+            "(X-BackendAI-Act-As)."
+        )
```
jopemachine: Why is this required?
You defended all the handlers themselves

tests/unit/common/contexts/test_user.py
```diff
@@ -0,0 +1,62 @@
...
+        domain_name="default",
+    )
+
+
+def test_triggered_user_none_when_unset() -> None:
+    assert triggered_user() is None
+
+
+def test_triggered_user_inside_context() -> None:
+    u = _user()
```
jopemachine: Don't abbreviate

src/ai/backend/manager/api/rest/middleware/auth.py
jopemachine: I removed it

src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -645,29 +649,79 @@ async def _authenticate_via_hook(
...
-            )
-            stack.enter_context(
-                with_log_context_fields({
-                    "user_id": str(user_id),
-                })
+    effective_user: UserData | None
+    trigger_user: UserData | None
+
+
+async def _load_user_data(db: ExtendedAsyncSAEngine, user_id: uuid.UUID) -> UserData:
```
jopemachine: Use UserID

src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -645,29 +649,79 @@ async def _authenticate_via_hook(
...
+    """Load a user's ``UserData`` by UUID (impersonation target). Raises if not found."""
+
+    async def _query() -> Any:
+        async with db.begin_readonly() as conn:
+            query = sa.select(users.c.uuid, users.c.role, users.c.domain_name).where(
```
jopemachine: Use ORM

tests/unit/common/contexts/test_user.py
```diff
@@ -0,0 +1,53 @@
...
+from ai.backend.common.contexts.user import (
+    current_user,
+    triggered_user,
+    with_triggered_user,
+    with_user,
+)
+from ai.backend.common.data.user.types import UserData, UserRole
+
+
+def _user(role: UserRole = UserRole.USER) -> UserData:
```
jopemachine: Don't make this

tests/unit/manager/api/test_impersonation.py
jopemachine, line 1: Use pyrametrize

src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -645,29 +650,75 @@ async def _authenticate_via_hook(
     await valkey_stat.increment_keypair_query_count(access_key)
 
 
-def _setup_user_context(request: web.Request) -> ExitStack:
+@dataclass(frozen=True)
+class RequesterIdentity:
```
jopemachine: Wouldn't it be better to move this type to a common location, such as types.py?

tests/unit/manager/api/test_impersonation.py
```diff
@@ -0,0 +1,105 @@
...
+    assert identity.effective_user is not None and identity.trigger_user is not None
+    assert identity.effective_user.user_id == target_id
+    assert not identity.effective_user.is_superadmin
+    assert identity.effective_user.domain_name == "target-domain"
+    assert identity.trigger_user.user_id == caller_id
+    assert identity.trigger_user.is_superadmin
+
+
+@pytest.mark.parametrize(
+    ("role", "raw_target", "loader_error", "expected"),
```
jopemachine: Create a type for this, not using tuple

jopemachine:
### ⚠️ Behavior note: `X-BackendAI-Act-As` + `owner_access_key` together

Heads-up on the interaction between the new impersonation header and the existing `owner_access_key` (owner_id) delegation, since this PR intentionally does **not** reject the combination.

Impersonation only overrides the `current_user()` **context var** — it does **not** touch `request["user"]` / `request["keypair"]`. But the session-create path derives the requester from the raw auth result, not from `current_user()`:

```python
# api/rest/session/handler.py
ResolveAccessKeyScopeAction(
    requester_access_key=request["keypair"]["access_key"],  # authenticated super admin
    requester_role=request["user"]["role"],                 # super admin
    ...
    owner_access_key=params.owner_access_key,               # UserB
)
```

**So if a super admin sends `X-BackendAI-Act-As: UserA` and `owner_access_key: UserB` to enqueue a session:**
- requester is evaluated as the **super admin** (from `request["user"]`), so the `owner_access_key` delegation passes,
- the session owner becomes **UserB** (owner_access_key wins),
- the `X-BackendAI-Act-As: UserA` header has **no effect** on this path.

In other words, on `request["user"]`/`request["keypair"]`-based paths (session, vfolder, …) **`owner_id` wins and Act-As is silently ignored** — impersonation currently only affects code that reads `current_user()`.

Flagging for visibility. If we'd rather this not be silently ambiguous, the alternative is to reinstate the "reject when both are set" rule (dropped in this PR).

src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -645,29 +649,76 @@ async def _authenticate_via_hook(
...
+async def _setup_user_context(request: web.Request, db: ExtendedAsyncSAEngine) -> ExitStack:
+    stack = ExitStack()
+
+    authenticated_user = _authenticated_user(request)
+    if authenticated_user is not None:
+        # trigger = caller always; effective = caller, overridden by act-as impersonation.
+        effective_user = await _resolve_effective_user(request, db, authenticated_user)
+        stack.enter_context(with_triggered_user(authenticated_user))
+        stack.enter_context(with_user(effective_user))
+        stack.enter_context(with_log_context_fields({"user_id": str(effective_user.user_id)}))
```
HyeockJinKim: I'm not sure if this needs to run asynchronously. It seems like it would be better to separate operations like `_resolve_effective_user`.
It might be worth thinking about how to separate them to better manage the implementation. 
I had included the request right up to the `context setup` phase, but once it gets to the database, it would probably be better to separate it from the front end.

HyeockJinKim (changes requested):

HyeockJinKim (approved):

PR #12665, opened 2026-07-08
fix(BA-6786): strip materialized nulls from stored model definitions

seedspirit:
## Summary
- Stored model definitions were serialized with plain `model_dump()`, materializing unset optional fields as explicit `null` keys; re-parsing marked them "explicitly set", and the draft merge helpers (`model_construct`) compounded this — so `to_resolved()` skipped default injection and `service.shell` resolved to `null`, running start commands without shell wrapping.
- Fix: union `model_fields_set` in the draft merge helpers so unset stays unset; add an `exclude_unset` flag to `PydanticColumn` and apply it to `runtime_variants.default_model_definition` so unset fields stay absent in stored JSON while intentional explicit nulls survive.
- Repair existing rows with a one-time `jsonb_strip_nulls` data migration (`2ec0aa5a19cf`) over `runtime_variants`, `deployment_revision_presets`, and `deployment_revisions` — pre-fix nulls are serializer artifacts by definition; downgrade is a no-op.

## Test plan
- [x] Unit: merge keeps unset shell unset and resolves to `DEFAULT_SHELL`; explicit null shell preserved (`tests/unit/common/test_config.py`)
- [x] Component: full revision-draft chain (variant baseline → model-definition.yaml → request) — layer precedence incl. shell, explicit-null shell, health-check file normalization (`tests/unit/manager/sokovan/deployment/test_revision_draft_reader.py`)
- [x] Column: `PydanticColumn(exclude_unset=True)` round-trip keeps unset unset, preserves explicit nulls, `copy()` retains the flag (`tests/unit/manager/models/test_pydantic_column.py`)
- [ ] Data migration verified against local DB (downgrade → seed bug-era rows → upgrade → verify strip)

Resolves BA-6786

### DIFF
--- a/src/ai/backend/manager/models/runtime_variant/row.py
+++ b/src/ai/backend/manager/models/runtime_variant/row.py
@@ -32,7 +32,7 @@ class RuntimeVariantRow(Base):  # type: ignore[misc]
     )
     default_model_definition: Mapped[ModelDefinitionDraft] = mapped_column(
         "default_model_definition",
-        PydanticColumn(ModelDefinitionDraft),
+        PydanticColumn(ModelDefinitionDraft, exclude_unset=True),
         nullable=False,
     )
 
--- a/tests/unit/manager/models/test_pydantic_column.py
+++ b/tests/unit/manager/models/test_pydantic_column.py
@@ -63,13 +63,21 @@ class TestPydanticColumnExcludeUnset:
         column_with_exclude_unset_false: PydanticColumn[ModelDefinitionDraft],
         dialect: Dialect,
     ) -> None:
-        # Contrast case: without exclude_unset, unset fields are stored as explicit nulls.
+        # Contrast case: without exclude_unset, unset fields are stored as explicit
+        # nulls, and reloading marks them "explicitly set" — the legacy bug behavior.
         draft = ModelDefinitionDraft.model_validate({"models": [{"service": {"port": 8080}}]})
 
         stored = column_with_exclude_unset_false.process_bind_param(draft, dialect)
+        loaded = column_with_exclude_unset_false.process_result_value(stored, dialect)
 
         assert stored is not None
         assert stored["models"][0]["service"]["shell"] is None
+        assert loaded is not None
+        assert loaded.models
+        service = loaded.models[0].service
+        assert service is not None
+        assert "shell" in service.model_fields_set
+        assert service.shell is None
 
     def test_copy_preserves_exclude_unset(
         self, column: PydanticColumn[ModelDefinitionDraft], dialect: Dialect
--- a/tests/unit/manager/models/test_pydantic_column.py
+++ b/tests/unit/manager/models/test_pydantic_column.py
@@ -1,0 +1,67 @@
+"""Unit tests for ``PydanticColumn`` bind/result behavior."""
+
+from __future__ import annotations
+
+import pytest
+from sqlalchemy.engine import Dialect
+from sqlalchemy.engine.default import DefaultDialect
+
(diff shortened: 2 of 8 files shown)


tests/unit/manager/models/test_pydantic_column.py
```diff
@@ -0,0 +1,67 @@
...
+
+class TestPydanticColumnExcludeUnset:
+    @pytest.fixture
+    def dialect(self) -> Dialect:
+        # process_bind_param/process_result_value ignore the dialect.
+        return DefaultDialect()
+
+    @pytest.fixture
+    def column(self) -> PydanticColumn[ModelDefinitionDraft]:
+        return PydanticColumn(ModelDefinitionDraft, exclude_unset=True)
```
jopemachine, line 21: Would it make sense to add a test for the case where exclude_unset is False as well?

src/ai/backend/manager/models/runtime_variant/row.py
```diff
@@ -32,7 +32,7 @@ class RuntimeVariantRow(Base):  # type: ignore[misc]
     )
     default_model_definition: Mapped[ModelDefinitionDraft] = mapped_column(
         "default_model_definition",
-        PydanticColumn(ModelDefinitionDraft),
+        PydanticColumn(ModelDefinitionDraft, exclude_unset=True),
```
jopemachine, line 35: The migration script touches three tables, but exclude_unset=True was only applied to one column.

Was there a specific reason for limiting it to that column?

tests/unit/manager/sokovan/deployment/test_revision_draft_reader.py
```diff
@@ -74,6 +87,270 @@ def test_adds_lowest_priority_model_path_default(self) -> None:
...
+    expected_start_command: str
+    expected_port: int
+    # Resolved shell — layers below may leave it unset, then it resolves to DEFAULT_SHELL.
+    expected_shell: str
+    reads_files: bool
+    yaml_payload: Mapping[str, Any] | None = None
+    request_payload: Mapping[str, Any] | None = None
+
+
+class DraftMergeChain:
```
jopemachine: Is there a reason this was implemented as a class instead of a function?

src/ai/backend/manager/models/runtime_variant/row.py
seedspirit: Thanks for sharp comment. The unset-vs-set distinction only matters for the draft type. `runtime_variants.default_model_definition` is the only column storing a `ModelDefinitionDraft`, whose `model_fields_set` drives the revision merge chain - so its write path needs `exclude_unset=True`.

The other two columns store the resolved strict `ModelDefinition`: after `to_resolved()` every field holds a concrete value, so a `null` in those dumps is the actual resolved value (e.g. `shell: null` = run without shell wrapping), not an unset artifact, and the round-trip is lossless without the flag.

Following the same reasoning (and the Copilot thread above), I've narrowed the migration to strip only `runtime_variants`.

tests/unit/manager/sokovan/deployment/test_revision_draft_reader.py
seedspirit: I did it that way because I thought grouping them into a class would look a little neater, but I've changed it to a function.

tests/unit/manager/models/test_pydantic_column.py
seedspirit: Add `test_default_dump_materializes_unset_fields_as_null`

jopemachine (dismissed):

jopemachine (approved):

PR #12668, opened 2026-07-08
feat(BA-6788): process all workloads in a single tick by priority order in sequencer

fregataa:
## Summary
- `SchedulingSequencer.sequence()` previously kept only the highest-priority workloads (`max(priority)`) and dropped all lower-priority workloads from the scheduling pass.
- Now it groups workloads by priority and concatenates them in descending priority order, passing the full set to the underlying workload sequencer.
- No workload is dropped; priority only determines ordering, so higher-priority sessions are still attempted first while lower-priority ones remain eligible in the same tick.

Resolves BA-6788


changes/12668.breaking.md
HyeockJinKim, line 1: Since this involves a change to a fundamental assumption, I think it should be classified as a "breaking" change rather than an "enhancement."

HyeockJinKim (approved):

PR #12670, opened 2026-07-08
feat(BA-6787): add scope filtering to Role.scopes GraphQL filter

fregataa:
## Summary
- Add `scopeType` and `scopeId` fields to the `EntityFilter` GraphQL input so a role's scopes can be filtered by the **scope side** of the association (e.g. only the projects a role is mapped to).
- Previously `EntityFilter` exposed only `entity_type`/`entity_id`, which `search_role_scopes` already pins to `ROLE`/`<role_id>` — so the filter could not narrow results at all.
- Wire the new fields in the adapter's `_convert_entity_filter` to the existing `EntityScopeConditions.by_scope_type_*` / `by_scope_id_*` factories (model layer already implemented). New GQL fields are version-tagged via `gql_added_field(NEXT_RELEASE_VERSION)`.

## Test plan
- [x] `pants fmt / fix / lint / check` (incl. mypy) pass
- [x] Live manager (direct API, `./bai gql --v2`) verified:
  - `scopeType {equals/in}` filters by scope type (PROJECT vs USER)
  - `scopeId {equals/startsWith}` filters by scope id
  - `AND` / `NOT` compose correctly; no filter returns all scopes unchanged
- [x] v2-schema & supergraph snapshots regenerated with `Added in` version annotation
- [x] CI green

Resolves BA-6787

----
📚 Documentation preview 📚: https://sorna--12670.org.readthedocs.build/en/12670/

----
📚 Documentation preview 📚: https://sorna-ko--12670.org.readthedocs.build/ko/12670/

### DIFF
--- a/changes/12670.feature.md
+++ b/changes/12670.feature.md
@@ -1,0 +1,1 @@
+Support filtering a role's scopes by scope type and scope id in the `Role.scopes` GraphQL field.
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -7056,6 +7056,16 @@ input EntityFilter
 {
   entityType: RBACElementTypeFilter = null
   entityId: StringFilter = null
+
+  """
+  Added in 26.8.0. Filter by the type of scope the entity is registered in.
+  """
+  scopeType: RBACElementTypeFilter = null
+
+  """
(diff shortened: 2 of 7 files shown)


seedspirit (approved):

ironAiken2 (approved):
LGTM

PR #12673, opened 2026-07-08
refactor(BA-6674): remove outward models -> config/clients imports

fregataa:
## Summary
- Cut the manager models layer's outward imports on `config` and `clients` so it depends only downward (`data`/`common`/`defs`/`errors`).
- Moved health status / DB-connection helpers out of `models/health.py` into `repositories/manager_admin/health.py`.
- `ImageRow.get_min_slot()` now takes `slot_units`, and `ModelServiceHelper.check_extra_mounts()` now takes `allowed_vfolder_types`, instead of a `LegacyEtcdLoader`.
- Removed the `models/storage.py` backward-compat re-export of `StorageSessionManager`; consumers import it directly from `clients.storage_proxy`, while models-internal usages go through a new `StorageManagerProtocol` and a shared `is_noop_host` helper in `common/defs`.

## Test plan
- [x] `pants fmt` / `pants fix` / `pants lint` (includes visibility rules)
- [x] `pants check` (mypy) — confirms Protocol structural conformance
- [x] `pants test` — unit + component (vfolder, vfs_storage, group, user, deployment)

Resolves BA-6674


HyeockJinKim (approved):
It's a little unclear why it was added as a general function, but let's address that in a future update.

PR #12679, opened 2026-07-08
fix(BA-6794): retain announcement message when disabling

### LINKED ISSUE #12678: Disabling an announcement deletes the message; update_announcement discards the message on the disable path
## Summary

The manager announcement API (`POST /manager/announcement`) accepts both `enabled` and `message`, but when `enabled=false` the service deletes the etcd key and silently discards the provided `message`. There is no way to represent a "disabled but retained" announcement, so disabling permanently loses the message.

## Current behavior

Code: `src/ai/backend/manager/services/manager_admin/service.py`

- Storage: a single etcd key `manager/announcement` holds only the message string. There is no separate `enabled` field — `enabled` is derived from key existence.
- `get_announcement`: key present → `{enabled: true, message: <data>`}; key absent → `{enabled: false, message: ""`}.
- `update_announcement`: if `enabled` → `etcd.put("manager/announcement", message)`; else → `etcd.delete("manager/announcement")` (the `message` argument is ignored on the disable branch).
- Consequence: `POST {enabled: false, message: "..."`} discards the message and removes the announcement. Re-enabling requires retyping the message. The state `enabled=false` with a retained `message` is not representable.

## Why this is a problem

The API surface (`enabled` + `message`) implies a caller can keep a message while disabled, but the storage model conflates `enabled` with key existence, so that state cannot exist. Accepting `message` on the disable path and silently dropping it is misleading and causes data loss.

## Expected (choose one product direction)

- (a) Store `enabled` and `message` separately so disabling retains the message (draft preserved across enable/disable), or
- (b) If "disable = remove announcement" is the intended product behavior, drop `message` from the disable path / document that it is ignored so the API is not misleading.

## Reproduction

1. Set an announcement with a non-empty message (enabled).
1. `POST /manager/announcement` with `{enabled: false, message: "keep me"`}.
1. `GET /manager/announcement` → `{enabled: false, message: ""`} — the message is gone.

## Frontend context (informational)

Surfaced while building the WebUI announcement edit modal (webui PR #8067 / FR-3223); the WebUI sends both fields and the discarded message shows up as "content disappears when disabling."

JIRA Issue: BA-6794

agatha197:
resolves #12678 (BA-6794)

## Problem

`POST /manager/announcement` accepts both `enabled` and `message`, but the announcement was stored as a **single** etcd key (`manager/announcement`) holding only the message, with `enabled` derived from key existence. Disabling (`enabled=false`) **deleted** the key, so the message was discarded. There was no way to represent a "disabled but retained" announcement, and disabling permanently lost the text even though the API accepts a `message` on that path.

## Change

Store the message and the enabled flag under **separate** etcd keys:

- `manager/announcement` — the message text
- `manager/announcement_enabled` — `"true"` / `"false"`

`update_announcement` now writes the message only when one is provided, so a disable request without a message preserves the stored text and merely flips the flag. Enabling still rejects an empty message; passing an explicit empty message clears the stored text. Announcements written before the flag existed (message key present, no flag key) are treated as enabled for backward compatibility. `get_announcement` reads both keys and returns `enabled`/`message` independently.

## Before / after

- **Before:** `POST {enabled: false, message: "x"}` → key deleted → `GET` returns `{enabled: false, message: ""}` (message lost).
- **After:** same request → `GET` returns `{enabled: false, message: "x"}` (message retained, just hidden).

## Tests

Updated `tests/component/operations/test_operations.py` and `tests/integration/operations/test_operations.py` to assert the message is retained on disable and cleared only on an explicit empty message.

## Frontend note

The WebUI announcement banner (lablup/webui) currently renders based on message presence only; a follow-up will gate it on `enabled` so a disabled-but-retained announcement stays hidden. Related: FR-3223.

**Checklist:**

- [x] Test case(s) demonstrating the before/after difference


src/ai/backend/manager/services/manager_admin/service.py
```diff
@@ -28,6 +28,14 @@
 
 __all__ = ("ManagerAdminService",)
 
+# etcd keys backing the system announcement. The message and the enabled flag
+# are stored under separate keys so that disabling an announcement hides it
+# without discarding the message (re-enabling does not require retyping). The
+# flag key is a sibling of — not a child of — the message key to avoid
+# parent/child ambiguity when reading the message key.
+_ANNOUNCEMENT_MESSAGE_KEY = "manager/announcement"
+_ANNOUNCEMENT_ENABLED_KEY = "manager/announcement_enabled"
```
HyeockJinKim: Does a key need to be added?

agatha197: Good catch — dropped the extra key. It is now stored as a single JSON object `{"enabled": bool, "message": str}` in the existing `manager/announcement` key (f9a397f57), so no new key is added and get/update stay atomic (one read / one write). Legacy bare-string values are still read as enabled for backward compatibility.

HyeockJinKim: Adding `announcement_enabled` as a key does not seem to be a valid justification.
It seems more appropriate to read the existing value and distinguish subpaths in the format of `announcement/enabled` and `announcement/message`.

HyeockJinKim: Since this seems like an area where migration would be possible without any issues, please take a look at it.

src/ai/backend/manager/services/manager_admin/service.py
```diff
@@ -82,23 +91,43 @@ async def update_status(
...
+            data = json.loads(raw)
+        except (ValueError, TypeError):
+            data = None
+        if isinstance(data, dict) and "enabled" in data and "message" in data:
+            return GetAnnouncementActionResult(
+                enabled=bool(data["enabled"]), message=str(data["message"])
+            )
+        # Legacy value: the key used to hold a bare message string. Treat its
+        # presence as an enabled announcement for backward compatibility.
+        return GetAnnouncementActionResult(enabled=True, message=raw)
```
HyeockJinKim: That wasn't what I meant—I was saying that reading the announcement is set as the fallback. As you mentioned, the key is used to separate "enabled" and "message," and if there's no value at that location, the system simply reads the "read" as a fallback.
After that, "write" is used to separate the keys.

HyeockJinKim (changes requested):

src/ai/backend/manager/services/manager_admin/service.py
agatha197: Applied as suggested (8aa959518). The message and enabled flag are now stored under separate subpaths — `manager/announcement/message` and `manager/announcement/enabled` — written together in a single `put_prefix` transaction.

On read, when the split keys are absent we fall back to the legacy bare `manager/announcement` key (treating its presence as enabled), exactly as you described. The first write migrates to the split layout and drops the legacy key so stale data doesn't linger.

Behavior is unchanged: disable retains the stored message, enable still rejects an empty message, and an explicit empty message clears the text. fmt/lint/mypy and the component/integration tests pass.

HyeockJinKim (approved):
I made the correction.

PR #12680, opened 2026-07-09
fix(BA-6792): prevent manager event-propagator memory leaks (WIP)

jopemachine:
## Summary

Fixes two related memory leaks in the manager's **event-propagator** path that
cause steady RSS growth over days (both scale with the *cumulative* number of
sessions / background-task subscriptions ever handled, not concurrent count).

Jira: BA-6792

### Leak A — bgtask SSE handler blocks forever on client disconnect
`manager/api/rest/events/handler.py` · `push_background_task_events`

`async for event in propagator.receive(...)` only wakes on an event, so when a
client disconnects **before** the task's terminal event arrives (tab close,
navigation, network blip, or subscribing to an already-finished task), the
handler blocks forever. Each abandoned stream permanently leaks an asyncio
task, the registered `WithCachePropagator` (+ its queue), and a metric series.

**Fix:** race `receive()` against `resp.wait()` via a `close_on_disconnect()`
task that closes the propagator on disconnect (same pattern the sibling
`push_session_events` already uses), and close the propagator in `finally`.

### Leak B — event-propagator alias metric series never removed
`common/metrics/metric.py` · `common/metrics/safe.py`

`observe_propagator_unregistered` called only `.dec()` on the
`backendai_event_propagator_alias_count` gauge, whose `alias_id` label is a
session/kernel/bgtask id. `prometheus_client` retains every distinct label
tuple forever, so RSS grows linearly with the total aliases ever registered.

**Fix:** call `.remove()` after `.dec()`; add a safe `SafeGauge.remove()` that
tolerates a missing series without tripping the metrics circuit breaker.
(Note: under Prometheus multiprocess mode `.remove()` frees the in-process
child — the dominant heap cost — but cannot shrink the append-only mmap `.db`
files; fully eliminating that requires dropping the high-cardinality `alias_id`
label, left as a follow-up.)

## Verification

- **Leak A** reproduced live: opening + immediately closing 1000 distinct bgtask
  SSE streams left `propagator_count=1000` and `async_task_count` elevated
  indefinitely on the running manager; after the fix both drain to baseline
  within the SSE ping interval.
- **Leak B** quantified with tracemalloc + memray: 100k register/unregister
  cycles retained 100k series (~71 MB heap, ~714 B/alias); with the fix, 0.
  Allocations attribute to `prometheus_client` `labels()`.

### DIFF
--- a/changes/12680.fix.md
+++ b/changes/12680.fix.md
@@ -1,0 +1,1 @@
+Fix two manager memory leaks in the event-propagator path: the background-task event stream (`GET /events/background-task`) no longer leaks an asyncio task and a registered event propagator when a client disconnects before the task finishes, and per-alias event-propagator metric series are now released on unregister instead of accumulating for the process lifetime.
--- a/src/ai/backend/common/metrics/metric.py
+++ b/src/ai/backend/common/metrics/metric.py
@@ -714,7 +714,17 @@ class EventPropagatorMetricObserver:
         self._propagator_count.dec()
         self._propagator_unregistration_count.inc()
         for domain, alias_id in aliases:
+            # alias_id is a session/kernel/bgtask id (unbounded cardinality). A
+            # bare .dec() leaves the series in the registry forever, so RSS grows
+            # linearly with the cumulative number of aliases ever registered.
+            # .dec() first (keeps the multiprocess livesum correct), then remove
+            # the series to free the in-process child object.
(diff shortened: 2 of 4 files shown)


jopemachine:
Split into per-leak PRs: #12686 (BA-6795, Leak A) and #12687 (BA-6796, Leak B) under epic BA-6792. Closing this combined PR.

PR #12686, opened 2026-07-09
fix(BA-6795): detect client disconnect in bgtask SSE event handler

jopemachine:
## Summary

Fixes a manager memory leak: the background-task SSE event handler blocks
forever when a client disconnects before the task's terminal event, leaking an
asyncio task, the registered event propagator (and its queue), and a per-alias
metric series — one leak per abandoned stream.

Jira: BA-6795 (part of epic BA-6792)

### Root cause
`manager/api/rest/events/handler.py` · `push_background_task_events`

`async for event in propagator.receive(...)` only wakes on an event, so a client
disconnect before the terminal (close) event — tab close, navigation, network
blip, or subscribing to an already-finished task — is never detected and the
handler blocks indefinitely.

### Fix
Race `receive()` against `resp.wait()` via a `close_on_disconnect()` task that
closes the propagator on disconnect (the same pattern `push_session_events`
already uses), and close the propagator in `finally`. Skip the trailing
`server_close` send when the client is already gone.

### Verification
Reproduced live: opening + immediately closing 1000 distinct bgtask SSE streams
left `propagator_count=1000` and `async_task_count` elevated indefinitely on the
running manager; after the fix both drain to baseline within the SSE ping
interval.

### DIFF
--- a/src/ai/backend/manager/api/rest/events/handler.py
+++ b/src/ai/backend/manager/api/rest/events/handler.py
@@ -209,6 +209,20 @@ class EventsHandler:
             self.event_hub.register_event_propagator(
                 propagator, [(EventDomain.BGTASK, str(task_id))]
             )
+
+            async def close_on_disconnect() -> None:
+                # resp.wait() returns once the SSE connection is closed by the
+                # client. Closing the propagator then unblocks receive() below,
+                # so the handler exits and the finally: cleanup runs. Without
+                # this, receive() blocks forever when a client disconnects
+                # before the terminal (close) event arrives, leaking the task,
+                # the registered propagator, and its per-alias metric series.
+                try:
+                    await resp.wait()
+                finally:
+                    await propagator.close()
+
+            disconnect_task = asyncio.create_task(close_on_disconnect())
             try:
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/api/rest/events/handler.py
```diff
@@ -230,9 +244,19 @@ async def push_background_task_events(
...
+                    await resp.send(dump_json_str({}), event="server_close")
             finally:
                 self.event_hub.unregister_event_propagator(propagator.id())
+                await propagator.close()
+                if not disconnect_task.done():
+                    disconnect_task.cancel()
+                    try:
+                        await disconnect_task
+                    except asyncio.CancelledError:
+                        pass
```
HyeockJinKim, line 259: I don't quite understand—doesn't aiohttp call `cancel` when the connection is lost?

jopemachine: See `handler_cancellation` option which is false by default
https://docs.aiohttp.org/en/stable/web_advanced.html

fregataa (dismissed):

fregataa (approved):

PR #12687, opened 2026-07-09
fix(BA-6796): remove event-propagator alias metric series on unregister

jopemachine:
## Summary

Fixes a manager memory leak where per-alias event-propagator metric series are
never released, so RSS grows linearly with the cumulative number of aliases
(session / kernel / bgtask ids) ever registered.

Jira: BA-6796 (part of epic BA-6792)

### Root cause
`common/metrics/metric.py` · `common/metrics/safe.py`

`observe_propagator_unregistered` called only `.dec()` on the
`backendai_event_propagator_alias_count` gauge, whose `alias_id` label is a
session/kernel/bgtask id. `prometheus_client` retains every distinct label
tuple forever, so the series is never released.

### Fix
Call `.remove()` after `.dec()`; add a safe `SafeGauge.remove()` that tolerates
a missing series without tripping the metrics circuit breaker. (Under Prometheus
multiprocess mode `.remove()` frees the in-process child — the dominant heap
cost — but cannot shrink the append-only mmap `.db` files; fully eliminating
that requires dropping the high-cardinality `alias_id` label, tracked separately.)

### Verification
Quantified with tracemalloc + memray: 100k register/unregister cycles retained
100k series (~71 MB heap, ~714 B/alias); with the fix, 0. Allocations attribute
to `prometheus_client` `labels()`.

### DIFF
--- a/changes/12687.fix.md
+++ b/changes/12687.fix.md
@@ -1,0 +1,1 @@
+Fix a manager memory leak where per-alias event-propagator metric series were retained for the process lifetime; they are now released when the propagator is unregistered.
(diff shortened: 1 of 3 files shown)


jopemachine:

PR #12694, opened 2026-07-09
fix(BA-6802): drop unbounded operation_name label from GraphQL request metrics

jopemachine:
## Summary
- `GraphQLMetricObserver` labeled `backendai_graphql_request_count` and `backendai_graphql_request_duration_sec` with `operation_name` — the client-supplied GraphQL operation name (`info.operation.name.value`). Every distinct name created new Counter/Histogram time series that were never removed, letting a client grow manager memory without bound.
- Verified on a live worker (Linux + memray): 1000 distinct operation names grew exposed series 1,296 → 14,296 (+13/op) and were never reclaimed while idle.
- Remove `operation_name` from the label set of both metrics; the schema-bounded `field_name`/`parent_type` labels remain. `operation_name` is still recorded on the OTEL span.

Resolves BA-6802.


jopemachine:

PR #12695, opened 2026-07-09
fix(BA-6803): evict completed retriable background tasks from `_ongoing_tasks`

jopemachine:
## Summary
- `BackgroundTaskManager.start_retriable()` and `_retry_bgtask()` add tasks to the in-memory `_ongoing_tasks` dict, but the only pop is on the legacy `start()` path. Completed retriable tasks lingered forever, and because `SingleBgtask`/`ParallelBgtask.retriable()` always return `True`, `do_heartbeat()` kept re-registering them to Valkey on every cycle.
- Verified on a live worker (Linux + memray): 206 completed `rescan_gpu_alloc_maps` tasks left 206 `bgtask:meta:*` keys whose TTL was continually refreshed by heartbeat; the count never decreased.
- Add `_evict_when_done()`, which removes a task from `_ongoing_tasks` once all of its async tasks complete, and call it from both retriable entry points (handles single and parallel).

Resolves BA-6803.

### DIFF
--- a/changes/12695.fix.md
+++ b/changes/12695.fix.md
@@ -1,0 +1,1 @@
+Fix a manager memory leak where completed retriable background tasks were never removed from the in-flight task registry and kept being re-registered to Valkey by the heartbeat loop.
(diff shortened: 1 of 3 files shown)


jopemachine:

PR #12696, opened 2026-07-09
fix(BA-6804): drop per-kernel keys from stream defaultdicts on termination

jopemachine:
## Summary
- The four per-kernel stream defaultdicts (`stream_pty_handlers`, `stream_execute_handlers`, `stream_proxy_handlers`, `stream_stdin_socks`) never deleted their outer keys. `handle_kernel_terminating()` also used a bare `stream_stdin_socks[stream_key]` access, so `defaultdict.__missing__` inserted an empty key on every kernel termination — including kernels that never opened a stream — growing the dicts unbounded across the session lifecycle.
- Verified on a live worker (Linux + memray): after real session create→terminate cycles, `stream/handler.py` and `defaultdict.__missing__` frames remained in the memray `--leaks` (un-freed) set.
- Use `.get()` in `handle_kernel_terminating` and pop all four per-kernel keys after cancelling the handlers.

Resolves BA-6804.

### DIFF
--- a/changes/12696.fix.md
+++ b/changes/12696.fix.md
@@ -1,0 +1,1 @@
+Fix a manager memory leak where the per-kernel stream handler dictionaries grew without bound as compute sessions were created and terminated.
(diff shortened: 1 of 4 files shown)


jopemachine:

PR #12703, opened 2026-07-09
doc(BA-6801): add BEP-1059 for agent↔resource group source-of-truth redesign

HyeockJinKim:
## Summary
- Add **BEP-1059** proposing to make the manager DB the single source of truth for an agent's resource-group membership (Epic **BA-6800**, follow-up to a production scheduling incident where sessions stayed pinned to an agent whose resource-group assumption broke at deploy time).
- Stops the heartbeat from overwriting `agents.scaling_group` / `resource_group_id` (config becomes a first-registration seed only), removes the compensation-debt `update_scaling_group` RPC + `agent.toml` rewrite, and adds a **service-layer group-change orchestration path** (v1 `ModifyAgent` hardcoded `TERMINATE`+`force`; new v2 API with a policy enum + `force`).
- Replaces agent config `scaling-group` with a **nullable** `initial_resource_group_name` (name-based, resolved to id at INSERT; **fail-fast breaking change**) and removes the dormant agent etcd `SGROUP` scope.
- Adds a **default resource group** (`scaling_groups.is_default`, at most one): first registration falls back to the default when the config name is null or unresolvable (an unresolvable explicit name still falls back but logs a warning), so event-based registration stays reliable while topology remains DB-authoritative.
- Reserves BEP number **1059** in the registry.

## Test plan
- [ ] Proposal document only — no runtime change. Reviewed for cross-section consistency and correct rendering.

Resolves BA-6801

### DIFF
--- a/proposals/BEP-1059-agent-resource-group-source-of-truth.md
+++ b/proposals/BEP-1059-agent-resource-group-source-of-truth.md
@@ -1,0 +1,211 @@
+---
+Author: HyeokJin Kim (hyeokjin@lablup.com)
+Status: Draft
+Created: 2026-07-09
+Created-Version:
+Target-Version:
+Implemented-Version:
+---
+
+# Agent↔Resource Group Registration: Make the Manager DB the Source of Truth
+
+## Related Issues
+
+- Epic **BA-6800**, this BEP **BA-6801**
+- Must stay consistent with the in-flight `scaling_group` name → `resource_group_id`
+  migration: BA-6706, BA-6780, BA-6050/6051, BA-6644/6710/6713
+- Background: a production scheduling incident (sessions stayed pinned to an agent
+  whose resource-group assumption broke at deploy time)
+
+## 1. Goal
+
+**Move the source of truth for an agent's resource-group membership from its local config
+to the manager DB.**
+
+Today the agent's local `scaling-group` config is effectively the source of truth. The
+agent reports `scaling_group` in every heartbeat, and the manager's upsert overwrites
+`agents.scaling_group` / `agents.resource_group_id` on **every** beat
+(`ON CONFLICT DO UPDATE`). So an agent that boots with a stale/default config **silently
+moves itself to the wrong resource group** on the next heartbeat, orphaning the sessions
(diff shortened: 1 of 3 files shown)


proposals/BEP-1059-agent-resource-group-source-of-truth.md
```diff
@@ -0,0 +1,211 @@
...
+## 2. Current State & Scope, by Area
+
+For each area, separate **✅ what already exists** from **➕ what to add**.
+
+### 2.1 Agent (heartbeat report)
+
+| | Item |
+|---|---|
+| ✅ | Every heartbeat sends the local config value (`self.local_config.agent.scaling_group`) in `AgentInfo` — `agent/agent.py:1277` |
+| ➕ | Remove agent config `scaling-group`, replace with new `initial_resource_group_name` (first-registration seed only). The agent keeps reporting (compat), but the reported value no longer changes the DB group |
```
fregataa, line 47: How about something like "resource group name hint"?

fregataa (commented):
Does the v2 "group change" API update a resource group of the target agent immediately or after finishing cleanup?

proposals/BEP-1059-agent-resource-group-source-of-truth.md
```diff
@@ -0,0 +1,228 @@
...
+  omits a group currently just picks the first allowed candidate, `registry.py:2022`). Out of
+  scope for this BEP; possible follow-up.
+- Other implementation details, such as the exact endpoint / GraphQL naming of the new v2 API,
+  are settled in the implementation PR.
+
+> Items decided and moved to the Decision Summary / body: ModifyAgent default policy
+> (terminate), `force`-unset behavior, config swap (name-based + fail-fast breaking, resolved
+> to id at INSERT → independent of the id migration), reschedule scope (keeps the resource
+> group + `RESCHEDULING` follows the preemption work).
+
```
fregataa, line 225: Those items seems not really questions. And a question about how a network should be handled when rescheduling a multi-kernel session can be added

fregataa (approved):

PR #12716, opened 2026-07-09
feat(BA-6815): expose ContainerRegistry.images nested connection

jopemachine:
## Summary

Adds a nested `images: ImageV2Connection` paginated field to the `ContainerRegistryV2` GraphQL type, so the one-to-many registry → images relationship can be traversed directly on the node.

### Why

`images.registry_id` is a `NOT NULL` FK to `container_registries.id` (one registry : many images). The forward direction `ImageV2.registry` was added in BA-6812. The reverse was reachable only through the standalone root query `containerRegistryImagesV2(scope: { registryId })` — not as a nested field — so the relationship was not graph-navigable. This exposes it where it belongs, on the node.

## Changes

- `ContainerRegistryV2` gains an async `images` connection resolver (filter, orderBy, cursor + offset pagination) scoped to the registry via `ImageConditions.by_registry_id(self.id)`, reusing `admin_search_images_gql` — the same logic as the existing `containerRegistryImagesV2` root query and the `ImageV2.aliases` nested-connection pattern.
- Regenerated `v2-schema.graphql` and `supergraph.graphql`.

## Example

```graphql
query {
  # (whatever query returns a ContainerRegistryV2 node)
  ... on ContainerRegistryV2 {
    registryName
    images(first: 20) { count edges { node { id identity { canonicalName } } } }
  }
}
```

## Verification

- `pants lint / check (mypy) / fix` pass. No circular import (image/types.py does not import container_registry at runtime).
- `dump-gql-schema --v2` and `generate-supergraph` build the full schema with the new connection field.

----
📚 Documentation preview 📚: https://sorna--12716.org.readthedocs.build/en/12716/

----
📚 Documentation preview 📚: https://sorna-ko--12716.org.readthedocs.build/ko/12716/

### DIFF
--- a/changes/12716.feature.md
+++ b/changes/12716.feature.md
@@ -1,0 +1,1 @@
+Add a nested `images: ImageV2Connection` paginated field to the `ContainerRegistryV2` GraphQL type, making the one-to-many registry->images relationship graph-navigable directly on the node instead of only via the standalone `containerRegistryImagesV2` root query.
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -3382,6 +3382,11 @@ type ContainerRegistryV2 implements Node
 
   """Extra metadata for the container registry"""
   extra: JSON
+
+  """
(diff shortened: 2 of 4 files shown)


jopemachine:
We decided to tackle this after adding RBAC support for the container registry.

PR #12744, opened 2026-07-10
feat(BA-1982): add `--app-proxy-frontend=traefik` option to install-dev.sh

Yaminyam:
Resolves #5319 (BA-1982)

## Summary

Automates the manual traefik-mode App Proxy setup (`docs/app-proxy/traefik-integration.md`) behind a new installer option:

```
./scripts/install-dev.sh --app-proxy-frontend=traefik
```

**Approach: Traefik runs as a halfstack compose service** (not a host binary), gated behind the `traefik` compose profile so default installs are untouched.

## Changes

| File | Change |
|---|---|
| `docker-compose.halfstack-main.yml` | New `backendai-half-traefik` service (`traefik:v3.7`, `profiles: ["traefik"]`): publishes the API entrypoint `8080` (the worker reconciles routes against `http://127.0.0.1:{api_port}`) and `10205-10300` portproxy entrypoints; healthcheck via `traefik healthcheck --ping` |
| `configs/traefik/traefik.halfstack.yml` | Completed with `experimental.localPlugins` (appproxy Traefik plugin wasm + go) and `ping: {}`; marker-directory placeholder substituted by the installer |
| `scripts/install-dev.sh` | Option parsing/validation + usage docs (`--app-proxy-impl` kept as a legacy alias — it was parsed but never consumed before); Traefik asset preparation **before** `compose up` (static config copy + marker path substitution, plugin download from `backend.ai-appproxy-worker-traefik` releases with a `gh` CLI fallback while that repository is private); `--profile traefik` via the existing `HALFSTACK_PROFILE_ARGS` mechanism; coordinator/worker TOML generation; run-guide note. Aborts when combined with `--configure-ha` (left as the existing separate TODO) |
| `scripts/delete-dev.sh` | `--profile traefik` on compose down + removal of the generated `traefik.halfstack.current.yml` |

### Generated configuration

Coordinator: `enable_traefik = true` + `[proxy_coordinator.traefik.etcd]` (`namespace = "traefik"`, halfstack etcd port). Worker: `frontend_mode = "traefik"` + `[proxy_worker.traefik]` (`api_port`, sub-mode `port`, marker directory) + `[proxy_worker.traefik.port_proxy]` (`port_range = [10205, 10300]` — note the tuple form; the string form shown in sample.toml does not validate).

[...]

PR #12749, opened 2026-07-10
feat(BA-1982): add traefik app-proxy frontend mode to the TUI installer

### LINKED ISSUE #5319: Add traefik-mode installation of App Proxy in `install-dev.sh`
Currently it requires manual configuration (`/docs/app-proxy/traefik-integration.md`) to use Traefik after running `install-dev.sh`.

Let’s add an option `--app-proxy-frontend=traefik`.

JIRA Issue: BA-1982

Yaminyam:
Resolves #5319 (BA-1982)

## Summary

Automates the manual traefik-mode App Proxy setup (`docs/app-proxy/traefik-integration.md`) in the **TUI installer**:

```
./backend.ai install --frontend-mode=traefik
```

Traefik runs as a **halfstack compose service** gated behind the `traefik` compose profile (same pattern as the observability/storage profiles), so default installs are untouched. Supersedes #12744 (same feature originally targeted at `install-dev.sh`, which is planned for deprecation).

## Changes

| File | Change |
|---|---|
| `docker-compose.halfstack-main.yml` | New `backendai-half-traefik` service (`traefik:v3.7`, `profiles: ["traefik"]`): API entrypoint published at `8080` (the worker reconciles routes against `http://127.0.0.1:{api_port}`), `10205-10300` portproxy entrypoints, healthcheck via `traefik healthcheck --ping` |
| `configs/traefik/traefik.halfstack.yml` | Completed with `experimental.localPlugins` (appproxy Traefik plugin wasm + go) and `ping: {}`; marker-directory placeholder substituted by the installer |
| `src/ai/backend/install/configs/traefik.yml` | New resource symlink (same pattern as the other bundled configs) |
| `src/ai/backend/install/types.py` | `FrontendMode.TRAEFIK` (CLI choices derive from the enum) + `ServiceConfig.frontend_mode` for the install report |
| `src/ai/backend/install/context.py` | `prepare_traefik_assets()` before `compose up` (static config copy + marker path substitution + plugin download with `gh` CLI fallback while the plugin repo is private); `--profile traefik`; `configure_appproxy` traefik branches for the coordinator (`enable_traefik` + `[proxy_coordinator.traefik.etcd]`, namespace `traefik`) and the HTTP worker (`[proxy_worker.traefik]`, port sub-mode, `port_range = [10205, 10300]`) |
| `src/ai/backend/install/app.py` | Traefik-specific instructions in the install report (App-Proxy Workers tab) |
| `src/ai/backend/install/cli.py` | Help text |

### Design notes

- **Marker directory**: bind-mounted into the container at the **same absolute path as on the host**, because the coordinator hands the worker's marker path verbatim to the Traefik plugin (`circuit.py` → `lastusedmarkerpath`).
- **TCP worker** keeps the port-based frontend: Traefik's etcd `rootKey` is per worker *authority*, and the bundled static config serves the HTTP worker (`worker-1`). Extending to the TCP worker needs a second rootKey/entrypoint set — left out of scope.
- Both `DevContext` and `PackageContext` run the same `install_halfstack`/`configure_appproxy` paths, so both install modes are covered.
- The plugin repository is currently **private**: anonymous `curl` 404s, so the installer falls back to `gh release download` (developer credentials); when the repo goes public the `curl` path takes over.

## Verification

- The exact tomlkit mutations were validated against the actual appproxy pydantic config models — including that `enable_traefik` lands inside `[proxy_coordinator]` (not a subtable) and `port_range` uses the tuple form (see #12747 for the sample.toml bug this uncovered).
- `docker compose config` valid with and without the `traefik` profile.
- Isolated compose run (etcd + traefik with these exact assets): both plugins load, `/ping` + dashboard 200, healthcheck healthy, and a router written under `traefik/worker_worker-1/...` in etcd appears via the etcd provider.
- `pants fmt lint check` clean on all touched Python files.
- Not yet exercised: a full `backend.ai install --frontend-mode=traefik` run with a live session app end-to-end; reviewer validation welcome.


seedspirit (commented):
Overall, the comments are a bit verbose, so please just leave the key points.

docker-compose.halfstack-main.yml
```diff
@@ -268,6 +268,33 @@ services:
       timeout: 3s
       retries: 10
 
+  backendai-half-traefik:
+    # Traefik dataplane for the app-proxy worker's traefik frontend mode.
+    # Enabled only via `install-dev.sh --app-proxy-frontend=traefik`
+    # (docker compose --profile traefik). The static configuration and the
+    # appproxy traefik plugin are prepared by install-dev.sh before `up`.
+    image: traefik:v3.7
```
seedspirit, line 274: Is there a reason why it was set to 3.7?

Yaminyam: This is because 3.7 is the latest minor version among the current stable versions.

Yaminyam: It was the latest stable minor line at implementation time (the plugin only requires Traefik v3.0+ with WASM support), and I verified it empirically with an isolated compose run (both appproxy plugins load, etcd-provider routing works). The `v3.7` minor tag keeps patch updates while avoiding an accidental jump to v3.8/v4.

For reference, the pyinfra fleet installer pins v3.3.7 and one internal cluster runs v3.4.4 in production with the same plugin — so if we prefer aligning the dev halfstack with the production-proven line, I'm happy to change the pin to `v3.4` (or `v3.3`). Let me know which you'd prefer.

seedspirit: The latest working version is best. Version 3.7 seems like the best choice.

Yaminyam:
The plugin repository (`lablup/backend.ai-appproxy-worker-traefik`) is now **public**, which unblocks this PR's install-time download path:

- Verified the exact command the installer runs works anonymously: `curl -fsSL <release-url> | tar xz` → exit 0, extracts the expected `plugins-local/src/backend.ai/...` layout (wasm + go plugin, `.traefik.yml` manifests).
- Pushed b83e5f1 dropping the now-unneeded `gh` CLI fallback (it existed only for the private-repo period) and making a failed download abort the install with a clear error instead of continuing silently.

Ready for review — the earlier v3.7 pin discussion is resolved above.

Yaminyam:
Design update in 52a5449: **split the traffic pattern and the dataplane backend into orthogonal flags.**

`--frontend-mode traefik` conflated two independent axes, which made the wildcard traffic pattern (used by `--fqdn-prefix` public setups) mutually exclusive with the Traefik backend — even though the appproxy worker config supports Traefik's own port/wildcard sub-modes.

- New: `--proxy-backend builtin|traefik` × `--frontend-mode port|wildcard` (full matrix). `--frontend-mode traefik` remains as a deprecated alias for `(port, traefik)`.
- Traefik wildcard sub-mode implemented: worker gets `[proxy_worker.traefik] frontend_mode="wildcard"` + `wildcard_domain` table; the installer swaps the `portproxy_*` entrypoints for the fixed `domainproxy` entrypoint (`:10250`) that the coordinator emits wildcard routers on.
- Bonus fix: the TCP worker config no longer receives `frontend_mode="traefik"` without a traefik table (latent validation failure in the previous shape); it always keeps the builtin frontend as intended.

Verification: both traefik sub-mode tables validate against the real `TraefikConfig` pydantic model; the static-config rewrite preserves the plugin + etcd provider blocks; `docker compose --profile traefik config` passes; pants fmt/lint/check green.

PR #12763, opened 2026-07-10
refactor(BA-6841): store audit-log acted_as as a native UUID column

jopemachine:
resolves BA-6841 (Epic BA-6778 / BEP-1058)

## Summary

The audit-log `acted_as` field is now a native **`UUID`** end to end. It is always a user UUID or `None` (the audit monitor writes `current_user().user_id` or `None`; existing rows were backfilled from `triggered_by`, itself a stringified user UUID), and the v2 response node / filters already exposed it as a `UUID` — but the `audit_logs.acted_as` **DB column** was still `String`.

This stores it as a native `UUID` column so the DB type, the internal data type, and the exposed type all agree, and the `str ↔ UUID` conversions at the boundary go away. `triggered_by` intentionally stays `String` (released and exposed as `str`).

## Changes

| File | Change |
|------|--------|
| `alembic/versions/b13d304bf1fd_…` | **new migration** — `op.alter_column` retype `String → uuid` with `USING acted_as::uuid` (+ downgrade back to text) |
| `models/audit_log/row.py` | column `String` → `GUID`; `to_dataclass` returns `acted_as` directly (no `uuid.UUID(...)` parse) |
| `repositories/audit_log/creators.py` | `build_row` passes the `UUID` directly (drops `str(...)`) |
| `repositories/audit_log/options.py` | `by_acted_as_equals` / `by_acted_as_in` compare against the `UUID`, not `str(value)` |
| `api/rest/export/adapter.py` | `_build_uuid_condition` compares against the `UUID` directly |
| `repositories/export/reports/audit_log.py` | `acted_as` export field type `STRING` → `UUID` (matches `id` / `action_id`) |
| `actions/monitors/audit_log.py`, `api/adapters/audit_log/adapter.py`, `data/audit_log/types.py` | carry `acted_as` as `uuid.UUID` end-to-end |

## Migration safety (verified against a live DB)

The `acted_as::uuid` cast is total: `acted_as` only ever holds a stringified user UUID or `NULL`.

Verified against the local halfstack DB (`audit_logs`, 377 rows):
- `377` total, `239` non-null `acted_as`, **`0` non-UUID** values.
- `ALTER … TYPE uuid USING acted_as::uuid` (upgrade) and `… TYPE varchar USING acted_as::text` (downgrade) both succeed inside a rolled-back transaction — no data loss, DB left untouched.

## No exposed-schema change

The response DTO (`AuditLogNode.acted_as: UUID`) and all filters were already `UUID`, so `openapi.json` / GraphQL dumps are unchanged.

## Tests

- Export-filter compile tests now render under the **PostgreSQL dialect** (the production target), where `GUID` renders hyphenated UUID literals.
- Repository / service / processor tests already exercise `acted_as` as `UUID`.
- Full unit suite + `fmt` / `lint` / `check` run in CI.

### DIFF
--- a/src/ai/backend/manager/models/alembic/versions/b13d304bf1fd_change_audit_logs_acted_as_to_uuid.py
+++ b/src/ai/backend/manager/models/alembic/versions/b13d304bf1fd_change_audit_logs_acted_as_to_uuid.py
@@ -1,0 +1,47 @@
+"""change audit_logs.acted_as to uuid
+
+``acted_as`` is always a user UUID or NULL (the monitor writes ``current_user().user_id``
+or ``None``, and existing rows were backfilled from ``triggered_by`` which is a user UUID).
+Retype the column from ``String`` to native ``UUID`` so the DB type matches the value it
+already holds and the exposed type. ``triggered_by`` intentionally stays ``String``.
+
+Existing text values are valid UUID strings, so the cast is total; NULLs stay NULL.
+
+Revision ID: b13d304bf1fd
+Revises: a3c1d8e5b294
+Create Date: 2026-07-13
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "b13d304bf1fd"
+down_revision = "a3c1d8e5b294"
+# Part of: NEXT_RELEASE_VERSION
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.alter_column(
+        "audit_logs",
+        "acted_as",
+        existing_type=sa.String(),
+        type_=postgresql.UUID(as_uuid=True),
+        existing_nullable=True,
+        postgresql_using="acted_as::uuid",
+    )
+
+
+def downgrade() -> None:
+    op.alter_column(
+        "audit_logs",
+        "acted_as",
+        existing_type=postgresql.UUID(as_uuid=True),
+        type_=sa.String(),
+        existing_nullable=True,
+        postgresql_using="acted_as::text",
+    )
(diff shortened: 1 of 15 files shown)


src/ai/backend/manager/models/alembic/versions/b13d304bf1fd_change_audit_logs_acted_as_to_uuid.py
```diff
@@ -0,0 +1,55 @@
...
+down_revision = "a3c1d8e5b294"
+# Part of: NEXT_RELEASE_VERSION
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    bind = op.get_bind()
+    inspector = sa.inspect(bind)
+    col = {c["name"]: c for c in inspector.get_columns("audit_logs")}.get("acted_as")
```
seedspirit: Is the part where we find "col" really necessary?

jopemachine: Actually, I don't think this part is necessary, but didn't we establish this as our standard pattern for keeping Alembic migration scripts idempotent?

seedspirit: There's no particular pattern here. When I wrote similar code before, I received feedback suggesting that I could just use "get item" and so on.

jopemachine: The guideline I had in mind was the database backport migration guide. Since this migration does not require a backport, I simplified the migration script.

seedspirit (approved):

fregataa (approved):

PR #12769, opened 2026-07-10
fix(BA-6843): register RBAC admin-page GET actions in the action registry

fregataa:
## Summary
- `PROJECT_ADMIN_PAGE` is declared as a valid scope-entity combination (`VALID_SCOPE_ENTITY_COMBINATIONS`) but had **no registered RBAC action**, so `rbacEntityOperationCombinations` and `rbacPermissionMatrix` returned **zero assignable operations** for it — making the permission un-assignable to custom project-scoped roles even though the default project-admin system role carries `PROJECT_ADMIN_PAGE: {READ}`.
- Add `ProjectAdminPageGetRBACAction` (PROJECT scope, GET → READ) as a declaration-only `BaseRBACAction` and register it in `RBAC_ACTION_REGISTRY`.
- All three RBAC matrix queries (`rbacScopeEntityCombinations`, `rbacEntityOperationCombinations`, `rbacPermissionMatrix`) now surface this entity consistently. No processor/service/handler needed — this is a metadata-only pseudo-entity gating admin-page visibility.

Resolves BA-6843

### DIFF
--- a/changes/12769.fix.md
+++ b/changes/12769.fix.md
@@ -1,1 +1,1 @@
-Allow `PROJECT_ADMIN_PAGE` and `DOMAIN_ADMIN_PAGE` permissions to be granted to custom RBAC roles by registering their admin-page actions in the permission matrix.
+Allow the `PROJECT_ADMIN_PAGE` permission to be granted to custom RBAC roles by registering its admin-page action in the permission matrix.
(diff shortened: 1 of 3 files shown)


jopemachine:
1. How about adding a test for this?
2. There is a DOMAIN_ADMIN_PAGE action in the code. Do we not need to add this action as well?

seedspirit (commented):
Is there nowhere else where `ProjectAdminPageGetRBACAction` is used?

ironAiken2 (changes requested):
Please check the comments :)

PR #12797, opened 2026-07-13
feat(BA-6857): add --shell option to deployment revision add CLI

seedspirit:
## Summary
- Add `--shell [SHELL]` / `--no-shell` options to `./bai deployment revision add` so shell wrapping can be controlled without hand-editing the `--config` JSON.
- The option injects `shell` into `model_definition.models[*].service` of the request config; an explicit `null` from `--no-shell` disables shell wrapping.
- If the config has no `model_definition` (or has it as explicit `null`), a single-model entry is created/normalized so the value still merges by index over the baseline.

## Behavior matrix

| Input | Injected value |
|-------|----------------|
| (omitted) | config/baseline untouched |
| `--shell` | `service.shell = "/bin/bash"` (`DEFAULT_SHELL`) |
| `--shell /bin/zsh` | `service.shell = "/bin/zsh"` |
| `--no-shell` | `service.shell = null` (disables shell wrapping) |
| `--shell … --no-shell` | `UsageError` (mutually exclusive) |

## Notes
- Overriding a preset/vfolder **baseline** shell with the explicit `null` depends on the draft-merge fix in #12665; until that lands, `--no-shell` only takes effect when no baseline provides a shell.

## Test plan
- [x] `pants fmt/lint/check` pass on the changed file
- [x] All behavior-matrix cases plus null-container configs (`model_definition: null`, `service: null`) verified via Click `CliRunner`
- [x] Explicit `shell: null` survives `AddRevisionInput` → `to_draft()` with `shell` in `model_fields_set`
- [ ] Live-server verification with a real deployment

Resolves BA-6857

### DIFF
--- a/changes/12797.feature.md
+++ b/changes/12797.feature.md
@@ -1,1 +1,1 @@
-Add `--use-shell` and `--no-use-shell` options to the `deployment revision add` v2 CLI to control model service shell wrapping
+Add `--shell` and `--no-shell` options to the `deployment revision add` v2 CLI to control model service shell wrapping
--- a/changes/12797.feature.md
+++ b/changes/12797.feature.md
@@ -1,0 +1,1 @@
+Add `--use-shell` and `--no-use-shell` options to the `deployment revision add` v2 CLI to control model service shell wrapping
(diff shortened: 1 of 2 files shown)


jopemachine:
Is there a reason you named it use_shell instead of just shell? 
I think use_shell reads like a boolean flag, so the name feels a bit misleading.

seedspirit:
Good point — renamed to `--shell` / `--no-shell` in b090a4b. Bare `--shell` still defaults to `/bin/bash`, and an explicit value (e.g. `--shell /bin/zsh`) overrides it.

fregataa (approved):

PR #12800, opened 2026-07-13
feat: add opt-in memray allocation tracking (agent, manager, storage-proxy, app-proxy)

Yaminyam:
## Summary

Adds an opt-in memray allocation-tracking option across the agent, manager, storage proxy, and app-proxy, backed by a **shared `MemrayConfig`** (`common/configs/memray.py`). Useful for diagnosing memory growth. Mirrors the profiling setup the app-proxy already shipped (`memray.Tracker(follow_fork=True)`).

## Shared config

`MemrayConfig` fields:
- `enabled` (default `false`)
- `output-destination` (default `./memray-output.bin`)
- `native-traces` (default `false`) — also capture native C/C++ stacks (`memray --native`); higher overhead, needs debug symbols to resolve

## Per-component

- **agent** — `[memray]` section on `AgentGlobalConfig`; installs `memray.Tracker(follow_fork=True, native_traces=...)` in `main()` before the service worker forks. `follow_fork=True` is required because `aiotools.start_server()` forks the worker that does the real work. The capture path gets the PID appended (memray refuses to overwrite an existing file).
- **manager** — same wiring; `memray` on `BootstrapConfig` (read in `main()` before the workers fork) and on `ManagerUnifiedConfig` (for the generated sample).
- **storage-proxy** — same wiring; `[memray]` on `StorageProxyUnifiedConfig`, installed in `main()` before the worker forks.
- **app-proxy** (coordinator + worker) — converge onto the shared config via a new `[profiling.memray]` table, **non-breaking**: the old `enable_memray` / `memray_output_destination` keys keep working as deprecated aliases, reconciled by a `model_validator` when the new table is absent (an explicit `[profiling.memray]` wins). This also adds native-traces support and the PID-suffixed capture path to app-proxy.

Disabled by default everywhere; enabling adds runtime and disk overhead (the capture grows for the process lifetime).

## Worker-side capture close (fix)

The forked worker's capture — the only one worth reading, since workers do the real work — used to end up truncated and unparseable (memray reported a 0.0s span out of a 366 MB file):

- memray's `Tracker` writes the capture trailer only from `__exit__`; it has no `__del__` and registers no `atexit` handler.
- We only closed the tracker in the master's `finally`, which never runs in a forked child — and `aiotools` ends children with a bare `os._exit()`, so `atexit` would not run either.

The tracker lifecycle now lives in two helpers next to the Pyroscope profiler: `start_memray_tracker()` for the master and `close_memray_tracker_in_worker()` called from each component's forked shutdown path. `Tracker.__exit__` is not idempotent, so master and worker each close exactly one tracker: their own (the module global is a per-process copy after the fork).

## Notes

- App-proxy uses snake_case TOML keys, so the shared memray table renders its kebab serialization aliases (`output-destination`, `native-traces`); both snake and kebab are accepted on load.
- Regenerating the app-proxy **coordinator** sample also pulled in previously un-regenerated schema fields (`pool_recycle`, `pool_pre_ping`, traefik `client_ip_strategy`) that were already merged — the sample is a generated artifact, so it is committed in sync.

## Verification

- Agent verified end-to-end on an internal GPU agent node (real workload): RPC/metric ports open, Prometheus scraping `/metrics` -> 200, `follow_fork` capturing the worker + kernel-runner forks (hundreds of thousands of allocations), `native-traces=true` path confirmed via capture metadata. Rolled back after testing.
- Worker-side close verified on the same node: after the fix, the worker's PID-suffixed capture parses cleanly with a full time span (previously truncated at the first bad record).
- App-proxy back-compat reconcile covered by a 4-case check (legacy flat keys / new nested table / absent / both-set precedence).
- `pants fmt/lint/check` green across all changed files; component samples regenerated from schema.


jopemachine:
From what I remember, the memray option added to the App Proxy Coordinator was intended to be temporary and is expected to be removed. I think it's worth confirming whether that's still the plan. @HyeockJinKim

PR #12814, opened 2026-07-13
feat(BA-6606): add virtual-scope-chain permission check

fregataa:
## Summary
- Add virtual-scope-chain permission resolution to `PermissionDBSource`, parallel to the existing direct scope-walk check.
- Resolve a user's effective `Permission` on an entity by walking `entity → entity_memberships → scope_bindings → scope`, evaluating the granted bitmask at each resolved scope and clipping each path by **both** hop caps (`granted & scope_cap & entity_cap`; `None` = no ceiling). Contributions are OR-combined.
- Public surface: `check_permission_via_virtual_scope` (single), `check_bulk_permission_via_virtual_scope` (bulk), `resolve_effective_permissions_via_virtual_scope` (key → `Permission` bitmask). Keys are built on the canonical open types (`EntityRef` / `UserID` / `Permission`), not `RBACElementType` / `RelationType`.

## Test plan
- [x] Unit tests (real DB, `with_tables`): permitted / denied (operation mismatch) / multi-level chain / cap clipping at scope→VS hop / cap clipping at VS→entity hop / NULL-cap (no clip) / no-VS fallback / inactive-role denied / bulk mapping / effective-bitmask resolve (both-hop clipping) / user isolation.

## Performance

Benchmarked the **actual** `PermissionDBSource.resolve_effective_permissions_via_virtual_scope()` code path (SQLAlchemy + asyncpg, via `begin_readonly_session_read_committed`) against synthetic, production-scale data on Postgres 16.3 (80 reps after warmup). Two query strategies were compared, each with and without **covering indexes** (`INCLUDE` the selected columns so the em / perm / sb scans go index-only):
- **(1)** the current single 4-join,
- **(3)** a virtual-scope-centric 2-query split (`entity_memberships ⋈ scope_bindings`, then `permissions ⋈ roles ⋈ user_roles`, joined in memory).

Each cell is **p50 / p95 ms**.

### Low fan-out — ~1 resolved row per entity
1M `entity_memberships` / 100k `virtual_scopes` / 100k `scopes` (scope:vs = 1:1).

| batch | (1) 4-join | (1) +covering | (3) split | (3) split +covering |
|------:|:---:|:---:|:---:|:---:|
| 1 | 1.01 / 1.16 | 1.14 / 1.77 | 1.83 / 2.29 | 1.00 / 1.10 |
| 10 | 1.66 / 1.81 | 1.97 / 2.49 | 2.82 / 3.52 | 1.24 / 1.32 |
| 100 | 7.72 / 11.56 | 2.58 / 3.04 | 8.69 / 15.10 | 7.77 / 12.10 |
| 1000 | 17.54 / 19.07 | 17.67 / 19.15 | 30.38 / 34.49 | 30.46 / 33.25 |

### High fan-out — ~10 resolved rows per entity
scope:vs = 10:1, vs:entity = 1:100 (1M entities / 10k `virtual_scopes` / 100k `scopes`; a batch of 1000 yields ~10k joined rows).

| batch | (1) 4-join | (1) +covering | (3) split | (3) split +covering |
|------:|:---:|:---:|:---:|:---:|
| 1 | 1.59 / 1.96 | 1.42 / 1.78 | 1.71 / 2.18 | 2.01 / 2.87 |
| 10 | 3.30 / 4.36 | 3.59 / 5.07 | 6.00 / 7.45 | 6.35 / 8.23 |
| 100 | 10.10 / 15.67 | 9.71 / 13.65 | 21.35 / 22.99 | 17.77 / 22.11 |
| 1000 | 124.33 / 147.52 | **87.30 / 110.66** | 145.26 / 183.75 | 144.31 / 182.66 |

**Findings**
- **The single 4-join is the right default in both regimes.** At batch 1000 it beats the split at low fan-out (17.5 vs 30 ms) and at high fan-out (87 vs 144 ms with covering indexes). `EXPLAIN ANALYZE` confirms the `permissions` CAST join (`perm.scope_id = CAST(sb.scope_id AS varchar)`) uses `ix_permissions_scope_entity` and the `roles`/`user_roles` join uses the `(user_id, role_id)` unique index.
- **The 2-query split does not pay off in the real code path.** A raw-`asyncpg` microbenchmark suggested it wins at high fan-out (35 vs 81 ms at batch 1000), but that gain vanished when measured through the real SQLAlchemy path: the split issues two queries and materializes Q1's `entity × scope` rows plus a large `IN`-list, and that client-side overhead outweighs the DB-side savings. It is slower-or-equal at every size — a useful caution that raw-driver microbenchmarks do not represent the ORM code path.
- **Covering indexes help the 4-join where the query is DB-bound**: low-fan-out mid batches (batch 100: 7.7 → 2.6 ms) and high-fan-out large batches (batch 1000: 124 → 87 ms), with no query change. At batch 1000 low fan-out the per-call session + row-materialization overhead dominates, so the index effect is neutral there.

**Takeaway:** keep the single 4-join and add the covering indexes; the virtual-scope-centric split was validated against the real code and does not help. Follow-ups tracked separately.

Resolves BA-6606

### DIFF
--- a/src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
@@ -13,6 +13,9 @@ from ai.backend.common.data.permission.types import (
     RBACElementType,
     RelationType,
 )
+from ai.backend.common.entity.types import EntityRef
+from ai.backend.common.identifier.entity import EntityID
+from ai.backend.common.identifier.user import UserID
 from ai.backend.logging.utils import BraceStyleAdapter
 from ai.backend.manager.actions.action.rbac_role_invitation import (
     CreateRoleInvitationResult,
@@ -52,11 +55,13 @@ from ai.backend.manager.data.permission.status import (
 from ai.backend.manager.data.permission.types import (
     EntityType,
     OperationType,
+    Permission,
     RBACElementRef,
     ScopeData,
     ScopeListResult,
     ScopeType,
 )
+from ai.backend.manager.data.permission.virtual_scope import VirtualScopePermissionCheckKey
 from ai.backend.manager.data.role_invitation.types import (
     RoleInvitationData,
     RoleInvitationState,
@@ -81,6 +86,8 @@ from ai.backend.manager.models.rbac_models.user_role import UserRoleRow
 from ai.backend.manager.models.role_invitation.row import RoleInvitationRow
 from ai.backend.manager.models.user import UserRow
 from ai.backend.manager.models.utils import ExtendedAsyncSAEngine
+from ai.backend.manager.models.virtual_scope.entity_membership import EntityMembershipRow
+from ai.backend.manager.models.virtual_scope.scope_binding import ScopeBindingRow
 from ai.backend.manager.repositories.base.creator import (
     BulkCreator,
     BulkCreatorResultWithFailures,
@@ -143,6 +150,20 @@ class _PermissionGroupKey:
     user_id: uuid.UUID
     element_type: RBACElementType
     subject_entity_type: RBACElementType
+
+
+@dataclass(frozen=True)
+class _VirtualScopePermissionGroupKey:
+    """Group key for batching virtual-scope-chain resolution inputs.
+
+    Keys sharing the same ``(user_id, entity_type)`` are resolved by a single
+    SQL round-trip differing only in the per-target ``entity_id`` IN-list.
+    ``entity_type`` is the DB-facing :class:`EntityType` enum (converted from the
+    open ``EntityRef.entity_type``) so it binds against the permission columns.
+    """
+
+    user_id: uuid.UUID
+    entity_type: EntityType
 
 
 class PermissionDBSource:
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
```diff
@@ -1161,6 +1182,136 @@ async def resolve_effective_permissions(
         """
         return await self._resolve_permissions_via_direct_scope_walk(keys)
 
+    # ------------------------------------------------ virtual-scope-chain checks
+
+    async def check_permission_via_virtual_scope(
+        self,
+        user_id: UserID,
+        entity: EntityRef,
```
seedspirit: nit: Wouldn't it be better to accept `VirtualScopePermissionCheckKey`?

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
```diff
@@ -1161,6 +1182,136 @@ async def resolve_effective_permissions(
...
+        """Check *permission* on each target key through the virtual-scope chain in one go.
+
+        Returns a mapping from each input key to whether the permission is granted.
+        """
+        if not keys:
+            return {}
+        resolved = await self.resolve_effective_permissions_via_virtual_scope(keys)
+        return {key: bool(resolved.get(key, Permission.NONE) & permission) for key in keys}
+
+    async def resolve_effective_permissions_via_virtual_scope(
```
seedspirit, line 1212: It looks like a public function - are you planning to use it elsewhere in the repository layer?

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
```diff
@@ -1161,6 +1182,136 @@ async def resolve_effective_permissions(
...
+            )
+            .where(
+                sa.and_(
+                    em.c.entity_type == group_key.entity_type,
+                    em.c.entity_id.in_(entity_ids),
+                    user_roles.c.user_id == group_key.user_id,
+                    roles.c.status == RoleStatus.ACTIVE,
+                )
+            )
+        )
```
jopemachine, line 1298: I think using `join_from()` to flatten the join conditions would improve readability.

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
```diff
@@ -1161,6 +1182,136 @@ async def resolve_effective_permissions(
...
+                        # stores its canonical string form. Cast to compare.
+                        perm.c.scope_id == sa.cast(sb.c.scope_id, sa.String),
+                        perm.c.entity_type == group_key.entity_type,
+                    ),
+                )
+                .join(roles, roles.c.id == perm.c.role_id)
+                .join(user_roles, user_roles.c.role_id == roles.c.id)
+            )
+            .where(
+                sa.and_(
```
jopemachine: `sa.and_` seems unnecessary

src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
fregataa: it will be used to check "what permissions do a certain user have on this entity"

jopemachine (approved):

seedspirit (approved):

PR #12821, opened 2026-07-14
feat(BA-6583): Implement session lifetime checker

seedspirit:
## Summary
- Add a reconciler-based session lifetime checker driven only by checker specifications.
- Skip disabled checker definitions and running sessions without a recorded start time.
- Register the checker in the idle-check stage and cover deadline, multi-definition, and repository filtering behavior.

## Test plan
- [x] `pants fmt ::`
- [x] `pants fix ::`
- [x] Push-hook MyPy checks for changed targets
- [x] Push-hook tests for changed targets

Resolves BA-6583

### DIFF
--- a/src/ai/backend/manager/models/session/conditions.py
+++ b/src/ai/backend/manager/models/session/conditions.py
@@ -46,7 +46,7 @@ class SessionConditions:
         return inner
 
     @staticmethod
-    def started() -> QueryCondition:
+    def by_started_at() -> QueryCondition:
         def inner() -> sa.sql.expression.ColumnElement[bool]:
             return SessionRow.starts_at.isnot(None)
 
--- a/src/ai/backend/manager/models/session/conditions.py
+++ b/src/ai/backend/manager/models/session/conditions.py
@@ -46,6 +46,13 @@ class SessionConditions:
         return inner
 
     @staticmethod
+    def started() -> QueryCondition:
+        def inner() -> sa.sql.expression.ColumnElement[bool]:
+            return SessionRow.starts_at.isnot(None)
+
+        return inner
+
+    @staticmethod
     def by_idle_check_candidates(
         candidates: Collection[tuple[ScopeId, Collection[SessionTypes]]],
     ) -> QueryCondition:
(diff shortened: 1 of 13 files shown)


src/ai/backend/manager/models/session/conditions.py
```diff
@@ -45,6 +45,13 @@ def inner() -> sa.sql.expression.ColumnElement[bool]:
 
         return inner
 
+    @staticmethod
+    def started() -> QueryCondition:
```
jopemachine: by_started_at

tests/unit/manager/repositories/idle_checker/test_repository.py
```diff
@@ -53,6 +53,7 @@ class SeededIdleCheckData:
     project_session_id: SessionId
     domain_session_id: SessionId
     inference_session_id: SessionId
+    unstarted_session_id: SessionId
```
jopemachine: Why not not_started_session_id?

src/ai/backend/manager/sokovan/idle_check/checkers/session_lifetime.py
```diff
@@ -0,0 +1,54 @@
...
+        *,
+        current_time: datetime,
+    ) -> Sequence[IdleJudgment]:
+        judgments: list[IdleJudgment] = []
+        for assignment in assignments:
+            lifetime_spec = assignment.definition.spec.session_lifetime
+            if lifetime_spec is None:
+                raise InternalServerError(
+                    f"Session lifetime checker(id: {assignment.definition.checker_id}) received an empty spec."
+                )
```
fregataa: I think it doesn't need to cancel all iterations here

src/ai/backend/manager/sokovan/idle_check/checkers/session_lifetime.py
```diff
@@ -0,0 +1,54 @@
...
+
+class SessionLifetimeChecker(IdleChecker):
+    """Judge sessions solely against each checker definition's lifetime setting."""
+
+    @override
+    async def judge(
+        self,
+        assignments: Sequence[CheckerAssignment],
+        *,
+        current_time: datetime,
```
fregataa: Does all idle checker need current time? and is there any spec that contains time data?

seedspirit: Some checkers may not use it directly, but time-based checkers should share one evaluation timestamp per reconcile tick. 
Utilization will likely need it as the end of its aggregation window and for its grace-period calculation, while the spec defines durations rather than an absolute evaluation time.

fregataa (approved):

PR #12828, opened 2026-07-14
feat(BA-6865): report idle checker deadlines to Valkey Live

seedspirit:
## Summary

- Collect absolute deadlines from idle checker judgments instead of storing time-varying remaining seconds.
- Store checker-specific deadlines in Valkey as UTC ISO 8601 timestamps with batched writes and expiration.
- Wire the Valkey Live client into the idle-check reconciler post-process flow so consumers can derive remaining time when needed.

## Test plan

- [x] Verify active and idle judgments preserve their deadlines for post-processing.
- [x] Verify multiple deadline reports are stored in one batch with checker-specific deadline_at keys.
- [x] Verify deadlines are serialized as UTC ISO 8601 timestamps.
- [x] Verify empty results skip Valkey writes.
- [x] Run changed-scope lint, type checks, and tests.

Resolves BA-6865


HyeockJinKim (changes requested):
Why do you save the "remaining time"?

PR #12841, opened 2026-07-14
feat(BA-6874): wire compute-schedule into the API handler

fregataa:
## Summary
- Wire the compute-schedule (dry-run) GraphQL resolver through the session adapter to the service, so callers can invoke the query end-to-end (resolver → adapter → service → scheduling controller). Previously the resolver was a schema-only stub raising `ServiceUnavailable`.
- Resolve the caller's main access key server-side (in the service, from `user_uuid` via the user repository) instead of passing a hardcoded value from the adapter.
- Extend the response payload to carry the full remediation breadth the internal data type already exposes: `UnschedulableReasonHint.required_container_reduction` / `image_not_found` and `ComputeSchedulePayload.resource_group_reason` (new GraphQL fields use `gql_added_field`).

## Known limitation (draft)
The end-to-end path is not yet functional at runtime: the `compute_schedule` service builds a `SessionSpecDraft` whose `scope.domain_id` / `domain_name` / `project_id` are unset, so the shared `SessionSpec` finalize validation (`session_spec_preparer._finalize`) rejects every request with `IncompleteSessionSpec` before node-fitting runs. These scope fields are consumed only by the enqueue persistence path (`scheduler/creators.py`) and are dead weight for the DB-mutation-free dry-run. Resolving this (populate scope from the user vs. relax the validation for this path) is deferred and tracked separately; this PR lands the wiring + payload surface.

## Test plan
- [x] `pants fmt / fix / lint / check` pass
- [x] GraphQL schema builds; new payload fields (`requiredContainerReduction`, `imageNotFound`, `resourceGroupReason`) exposed in SDL
- [x] Live GQL query reaches resolver → adapter → service → controller (verified request flow; blocked at the scope validation described above)
- [x] End-to-end success for admin and non-admin (blocked on the scope limitation above)

Resolves BA-6874

----
📚 Documentation preview 📚: https://sorna--12841.org.readthedocs.build/en/12841/

----
📚 Documentation preview 📚: https://sorna-ko--12841.org.readthedocs.build/ko/12841/

### DIFF
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -3046,6 +3046,11 @@ type ComputeSchedulePayload
   Per-kernel compute-schedule outcomes, correlated to the request by list index.
   """
   results: [ComputeScheduleKernelResult!]!
+
+  """
+  Reason the resource group as a whole cannot admit the session (e.g. it has no agents). Null when per-kernel results apply.
+  """
+  resourceGroupReason: String
 }
 
 type ComputeSession implements Item
@@ -19284,6 +19289,11 @@ type UnschedulableReasonHint
 
   """Architectures that actually exist among the scheduling targets."""
   availableArchs: [String!]
+
+  """
+  The requested image could not be found. When True, adjusting resource slots cannot make the kernel schedulable.
+  """
+  imageNotFound: Boolean!
 }
 
 type UnsetQuotaScope
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -3046,11 +3046,6 @@ type ComputeSchedulePayload
   Per-kernel compute-schedule outcomes, correlated to the request by list index.
   """
   results: [ComputeScheduleKernelResult!]!
-
-  """
-  Reason the resource group as a whole cannot admit the session (e.g. it has no agents). Null when per-kernel results apply.
-  """
-  resourceGroupReason: String
 }
 
 type ComputeSession implements Item
@@ -19286,14 +19281,6 @@ type UnschedulableReasonHint
 {
   """Subtract these slots to fit the best-fitting node."""
   requiredReduction: [ResourceSlotEntry!]
-
-  """Architectures that actually exist among the scheduling targets."""
-  availableArchs: [String!]
-
-  """
-  The requested image could not be found. When True, adjusting resource slots cannot make the kernel schedulable.
-  """
-  imageNotFound: Boolean!
 }
 
 type UnsetQuotaScope
--- a/src/ai/backend/manager/api/adapters/session/adapter.py
+++ b/src/ai/backend/manager/api/adapters/session/adapter.py
@@ -40,6 +40,12 @@ from ai.backend.common.dto.manager.v2.resource_slot.types import (
     ResourceOptsEntryInfoDTO,
     ResourceOptsInfoDTO,
 )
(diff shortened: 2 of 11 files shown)


src/ai/backend/manager/api/adapters/session/adapter.py
```diff
@@ -317,6 +329,88 @@ async def enqueue(
...
+
+    async def compute_schedule(self, input: ComputeScheduleInput) -> ComputeSchedulePayload:
+        """Probe whether each kernel of a would-be session fits the target
+        resource group's nodes, without provisioning.
+
+        A self-service query: the requesting user is resolved from the request
+        context and passed as ``user_uuid``. The service resolves that user's
+        main access key from it, so the adapter supplies no access key.
+        """
+        cluster_mode = (
```
jopemachine, line 344: Why not just `cluster_mode = input.cluster_mode`?

jopemachine: Oh, there are two separate types

src/ai/backend/manager/api/adapters/session/adapter.py
```diff
@@ -317,6 +329,88 @@ async def enqueue(
...
+            )
+        return ComputeScheduleKernelResultInfo(
+            requested_slots=[
+                ResourceSlotEntryInfo(
+                    resource_type=entry.resource_type,
+                    quantity=Decimal(entry.quantity),
+                )
+                for entry in result.requested_slots
+            ],
+            requested_architecture=result.requested_architecture or "",
```
jopemachine: If possible, I'd like to remove this kind of value injection. The fact that requested_architecture is None seems to correspond to the case where the Image information is empty.

Is it possible that the kernel's image information hasn't been loaded in that case? If that scenario can never actually happen, what do you think about throwing an error instead?"

src/ai/backend/manager/api/gql/scheduler.py
```diff
@@ -230,6 +239,15 @@ class ComputeScheduleKernelResultGQL:
 )
 class ComputeSchedulePayloadGQL:
     results: list[ComputeScheduleKernelResultGQL]
+    resource_group_reason: str | None = gql_added_field(
```
jopemachine: From an extensibility perspective, what do you think about introducing a generic failure type and making resource_group_reason one of its variants?

src/ai/backend/manager/api/adapters/session/adapter.py
fregataa: I changed this to nullable. Clients can request with image id that does not exist.
And this API is about checking whether each kernel can be schedule with given options, multiple kernels can be created with different images. It should be checked one by one rather than just raising single exception

docs/manager/graphql-reference/supergraph.graphql
```diff
@@ -19284,6 +19299,11 @@ type UnschedulableReasonHint
 
   """Architectures that actually exist among the scheduling targets."""
   availableArchs: [String!]
+
+  """
+  The requested image could not be found. When True, adjusting resource slots cannot make the kernel schedulable.
+  """
+  imageNotFound: Boolean!
```
HyeockJinKim: ???

src/ai/backend/manager/api/adapters/session/adapter.py
fregataa: We decided to return "required resource reduction" only in response rather than including other schedule failure reasons. It now raises image not found error

HyeockJinKim (approved):

seedspirit (approved):

PR #12842, opened 2026-07-14
doc(BA-6568): add BEP-1062 Virtual Scope RBAC Ownership Model

HyeockJinKim:
## Summary
- Add `proposals/BEP-1062-virtual-scope-rbac.md` — documents the as-built virtual scope surface (tables/types/DB ops/resolution from BA-6603..6606, 6863, all currently unwired) and defines the target design.
- Core design: replace the recursive scope-walk with a non-recursive two-hop ownership model `scope -> virtual_scope -> entity`; the VS hub bounds connection fan-out and removes BEP-1048's auto/ref edge distinction.
- Covers Scope Creator lifecycle (VS + self scope_binding), hierarchy/association bindings at creation, invitation as a capped membership, action-derived possible/impossible catalog, and the backfill + cutover migration behind a temporary resolution config.
- Register BEP number 1062 in the proposals registry.

## Test plan
- [ ] Docs-only (BEP proposal); no code paths changed. Reviewers confirm the as-built description matches the merged BA-6603..6606/6863 code and that the migration plan resolves the Epic's open questions.

Extends BEP-1048. Part of Epic BA-6567.
Resolves BA-6568

### DIFF
--- a/changes/12842.doc.md
+++ b/changes/12842.doc.md
@@ -1,0 +1,1 @@
+Add BEP-1062 proposing the Virtual Scope RBAC ownership model that replaces the recursive scope-walk with a non-recursive two-hop `scope -> virtual_scope -> entity` resolution, documenting the as-built tables/ops and the migration plan.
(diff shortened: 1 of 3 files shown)


fregataa (commented):
- How should we handle a permission check to scope-type entities themselves? e.g. does this user have a read permission to the project?
- Will we add any API to CRUD virtual scope? or they will be hidden from clients?
- Will we allow a virtual scope without any original scope?
- Will `association_scopes_entities` table be dropped?

fregataa (approved):

PR #12849, opened 2026-07-14
refactor(BA-6880): decouple the session-spec preparer chain from ownership scope

fregataa:
## Summary
- Introduce `SessionResourceSpec` / `SessionResourceSpecDraft` (`SessionSpec` / `SessionSpecDraft` minus scope) as the scope-free type the preparer chain and all 9 rules operate on and finalize into. No preparer rule ever reads or writes ownership scope, yet the finalized `SessionSpec` required it — forcing every caller to supply a complete `domain`/`project`/`resource-group` scope even when irrelevant.
- `SessionSpec` is now composed from a `SessionResourceSpec` + a resolved `SessionScope` via `SessionSpec.from_resource_spec()`. The enqueue path attaches scope after finalize (via `SessionSpecDraft.to_resource_draft()` → prepare → compose), preserving its behavior and scope validation.
- The compute-schedule (dry-run) path consumes `SessionResourceSpec` directly and passes the resource group id explicitly, so it no longer fails finalize purely because `domain`/`project` are unset.
- Replace `SessionMetadata.scaling_group` (name) with `resource_group_id: ResourceGroupID`, and make `NoAgentsInResourceGroupError` carry a `ResourceGroupID`.

## Test plan
- [x] `pants fmt / fix / lint / check` pass (source + tests)
- [x] `pants test` on the affected sokovan suites passes (enforced by the pre-push hook)
- [x] Preparer-rule, preparer-finalize, and selector message-format unit tests updated for the scope-free type and `resource_group_id`

Resolves BA-6880

### DIFF
--- a/src/ai/backend/manager/data/session/draft.py
+++ b/src/ai/backend/manager/data/session/draft.py
@@ -207,6 +207,25 @@ class SessionClassificationDraft(_DraftBaseModel):
     tag: str | None = None
 
 
+class SessionResourceSpecDraft(_DraftBaseModel):
+    """Scope-free draft consumed by the preparer chain.
+
+    Mirrors :class:`~ai.backend.manager.data.session.spec.SessionResourceSpec`
+    (the finalized counterpart) and holds everything the preparer
+    resolves. Ownership scope is carried separately by the caller (see
+    :meth:`SessionSpecDraft.to_resource_draft`), so no rule needs it.
+    """
+
+    identity: SessionIdentityDraft = Field(default_factory=SessionIdentityDraft)
+    classification: SessionClassificationDraft = Field(default_factory=SessionClassificationDraft)
+    network: SessionNetworkDraft = Field(default_factory=SessionNetworkDraft)
+    callback_url: yarl.URL | None = None
+    dependencies: tuple[SessionID, ...] = ()
+    options: SessionOptionsDraft = Field(default_factory=SessionOptionsDraft)
+    kernel_specs: tuple[KernelSpecDraft, ...] = ()
+    internal_data_extras: InternalDataExtras = Field(default_factory=InternalDataExtras)
+
+
 class SessionSpecDraft(_DraftBaseModel):
     """Top-level draft mirroring ``SessionSpec``.
 
@@ -226,3 +245,20 @@ class SessionSpecDraft(_DraftBaseModel):
     options: SessionOptionsDraft = Field(default_factory=SessionOptionsDraft)
     kernel_specs: tuple[KernelSpecDraft, ...] = ()
     internal_data_extras: InternalDataExtras = Field(default_factory=InternalDataExtras)
+
+    def to_resource_draft(self) -> SessionResourceSpecDraft:
+        """Project onto the scope-free draft the preparer chain operates on.
+
+        Scope is dropped here — it is pass-through input the preparer
+        never touches; the caller re-attaches it after finalize.
+        """
+        return SessionResourceSpecDraft(
+            identity=self.identity,
+            classification=self.classification,
+            network=self.network,
+            callback_url=self.callback_url,
+            dependencies=self.dependencies,
+            options=self.options,
+            kernel_specs=self.kernel_specs,
+            internal_data_extras=self.internal_data_extras,
+        )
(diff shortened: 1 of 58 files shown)


src/ai/backend/manager/sokovan/scheduler/provisioner/selectors/exceptions.py
```diff
@@ -76,9 +77,9 @@ class NoAgentsInResourceGroupError(AgentSelectionError):
     error_type = "https://api.backend.ai/probs/no-agents-in-resource-group"
     error_title = "Unavailable : Resource group has no candidate agents."
 
-    _resource_group: str
+    _resource_group: ResourceGroupID
 
-    def __init__(self, resource_group: str) -> None:
+    def __init__(self, resource_group: ResourceGroupID) -> None:
         self._resource_group = resource_group
```
seedspirit: nit: 
```suggestion
    _resource_group_id: ResourceGroupID

    def __init__(self, resource_group_id: ResourceGroupID) -> None:
self._resource_group_id = resource_group_id
```

src/ai/backend/manager/sokovan/scheduling_controller/scheduling_controller.py
```diff
@@ -278,7 +283,9 @@ async def enqueue_session_from_draft(
         with self._metric_observer.measure_phase(
             "scheduling_controller", rg_id, "spec_preparation"
         ):
-            spec = await self._spec_preparer.prepare(draft, prep_ctx)
+            resource_spec = await self._spec_preparer.prepare(draft.to_resource_draft(), prep_ctx)
+            scope = SessionScope.model_validate(draft.scope.model_dump(exclude_none=True))
```
seedspirit, line 287: Aren't all fields in SessionScope non-nullable? The `exclude_none=True` option seems unnecessary.

src/ai/backend/manager/sokovan/scheduling_controller/scheduling_controller.py
```diff
@@ -367,7 +374,8 @@ def _prepare_kernel_data(
 
     async def _compute_schedule(
         self,
-        spec: SessionSpec,
+        spec: SessionResourceSpec,
+        resource_group_id: ResourceGroupID,
```
seedspirit, line 382: Although the current assumption is that we'll only be handling one resource group, if we might need to expand this later, it seems like a good idea to use a data class wrapped in a `scope` or similar.

fregataa: We decided  not to include any scope info but resource group when query compute schedule. It is intentional

src/ai/backend/manager/sokovan/scheduling_controller/scheduling_controller.py
fregataa: I wanted how SessionSpecPreparer.prepare() works stays the same

src/ai/backend/manager/data/session/draft.py
```diff
@@ -207,6 +207,19 @@ class SessionClassificationDraft(_DraftBaseModel):
     tag: str | None = None
 
 
+class SessionResourceSpecDraft(_DraftBaseModel):
```
jopemachine, line 210: Could we reduce the duplication by having SessionSpec inherit from or contain SessionResourceSpecDraft?

src/ai/backend/manager/data/session/spec.py
```diff
@@ -44,6 +44,16 @@
 from ai.backend.manager.errors.kernel import IncompleteSessionSpec
 
 
+def _format_loc(loc: tuple[object, ...]) -> str:
```
jopemachine, line 47: Nit: It might make more sense to move this to common or util.

src/ai/backend/manager/data/session/spec.py
```diff
@@ -152,18 +159,28 @@ class SessionSpec(_SpecBaseModel):
...
+
+    resource_spec: SessionResourceSpec
+    scope: SessionScope
+
+    @classmethod
+    def from_resource_spec(
+        cls, scope: SessionScope, resource_spec: SessionResourceSpec
+    ) -> SessionSpec:
+        """Attach an ownership ``scope`` to a prepared, scope-free spec."""
+        return cls(resource_spec=resource_spec, scope=scope)
```
seedspirit: It seems there's no longer any reason for this classmethod to exist.

seedspirit (approved):

PR #12856, opened 2026-07-15
feat(BA-6886): register APP_CONFIG_FRAGMENT as an RBAC resource type

jopemachine:
## Summary
Register `APP_CONFIG_FRAGMENT` as an RBAC resource type and backfill its write permissions onto existing write-capable roles, so later PRs in the stack can authorize fragment writes through RBAC (BEP-1052).

Split out of #12826 as its base: the entity-type registration and the permission backfill are independent of the per-fragment scope binding, and land together because a backfilled permission is meaningless unless the type is a resource type.

## Changes
- `EntityType._resource_types()` — add `APP_CONFIG_FRAGMENT`, so **new** roles receive its permissions at role creation (10 → 11 types)
- `rbac_models/migration/enums.py` — same registration for the migration-local enum
- `b3d9f7a2c184` — backfill the permissions onto roles that **predate** the entity type

## What the migration backfills
Only `permissions` rows, and only where a fragment can actually live:

| | |
|---|---|
| Scopes touched | `user` and `domain` only — a fragment lives at no other scope. `public` maps to `ScopeType.GLOBAL` and is granted to **nobody**: superadmin reaches it through the `is_superadmin` short-circuit in the RBAC validators, which needs no rows |
| Roles targeted | a user's own `user`-scope role, and `domain` admins **selected by capability** — the role must hold `domain_admin_page` at that same domain (see below) |
| Operations | the full owner operation set; `permission` carries the matching bitmask (`c6648c039bd4` shape), grant operations store 0 |
| Not migrated | the scope **associations** (`association_scopes_entities`) — they are per-fragment and written at fragment-creation time (#12826). The feature is unreleased, so no fragments and no associations exist yet |

`ON CONFLICT DO NOTHING`, so re-running grants nothing twice. `downgrade` is a no-op, following `f2b9a4c7e103`: a `DELETE` by `entity_type` cannot tell backfilled rows from the ones role creation now grants natively, so it would strip permissions that roles created since this revision legitimately hold.

## Selecting admins by capability, not by name
The sibling backfills (`f1a2b3c4d5e6`, `30c8308738ee`, …) select owner roles **negatively**, via `NOT (role_name LIKE '%member')`. That is unsafe: role names carry no enforced convention — `CreateRoleInput.name` is a free string and the caller picks the scope — so any domain-scoped role not named `%member` matches. The `domain-viewer` preset in `test_create_preset_roles.py` (read-only: `VFOLDER:READ`, `SESSION:READ`) is exactly that shape and would have received `hard-delete` and `grant:all` on fragments across the whole domain.

Those grant rows are not inert: the live check matches on `permissions.operation` (`db_source.py`), and the bitmask is the half that is currently unread.

So this migration selects admins positively instead:

```sql
WHERE rs.scope_type = 'user'
   OR EXISTS (SELECT 1 FROM permissions admin_page
              WHERE admin_page.role_id     = rs.role_id
                AND admin_page.entity_type = 'domain_admin_page'
                AND admin_page.scope_type  = rs.scope_type
                AND admin_page.scope_id    = rs.scope_id)
```

- **`domain`** — every domain admin role holds `domain_admin_page` at its own domain by the time this runs, under **both** naming schemes: `3b6297b1bd75` covers `role_domain_%_admin`, `f2b9a4c7e103`/`a3c1d8e5b294` cover `domain-%-admin`, and all three are ancestors of this revision. So no name matching is needed and the `roles` join drops out entirely. Matching `scope_id` too keeps an admin of domain A from being granted on domain B.
- **`user`** — kept broad. A user-scope grant only ever reaches that user's own scope, and writing an own-scope fragment is deliberately **not** admin-only (see `AppConfigScopeType.to_rbac_scope_type`).
- **`role_superadmin`** holds domain-scoped rows but no `domain_admin_page`, so it drops out. That is correct and matches `3b6297b1bd75`: superadmin authority comes from the `is_superadmin` short-circuit, never from these rows.

Verified against a local DB (per `models/alembic/AGENTS.md`, data migrations must be checked against a real DB):

| domain-scope role | outcome |
|---|---|
| `domain-testdom-admin` (runtime naming) | full ops ✅ |
| `role_domain_default_admin` (legacy naming) | full ops ✅ |
| `domain-auditors-readonly` (read-only custom, the escalation) | **nothing** ✅ |
| `role_domain_default_member` | nothing ✅ |
| `role_superadmin` | nothing ✅ |

## Backport
Not needed for 26.4 — the feature does not exist there:

| | `main` | `26.4` |
|---|---|---|
| `app_config_fragment` paths | 36 | **0** |
| `create_app_config_fragments_table` migration | present | **absent** |
| `EntityType.APP_CONFIG_FRAGMENT` | present | **absent** |

26.4's `84d5c6daf8cc` says it directly: the replacement tables "are introduced in a follow-up migration on top of this one". Backporting would fail to import `EntityType.APP_CONFIG_FRAGMENT`, and would grant permissions on a table that does not exist.

Jira: https://lablup.atlassian.net/browse/BA-6886

### DIFF
--- a/src/ai/backend/manager/models/alembic/versions/b3d9f7a2c184_add_app_config_fragment_permissions_to_rbac.py
+++ b/src/ai/backend/manager/models/alembic/versions/b3d9f7a2c184_add_app_config_fragment_permissions_to_rbac.py
@@ -34,21 +34,38 @@ depends_on = None
 
 def upgrade() -> None:
     # Backfill only ``user``/``domain`` scopes (``public`` is superadmin-only, no role) and only
-    # the write path (reads use the allow list, not RBAC); member roles (suffix ``member``) do not
-    # write fragments and are excluded. ``permission`` is the NOT NULL bitmask from c6648c039bd4:
-    # bits are inlined as literals so this migration stays frozen against later Permission enum
-    # changes, and grant operations have no bit (0).
+    # the write path (reads use the allow list, not RBAC). ``permission`` is the NOT NULL bitmask
+    # from c6648c039bd4: bits are inlined as literals so this migration stays frozen against later
+    # Permission enum changes, and grant operations have no bit (0).
+    #
+    # Sibling backfills (f1a2b3c4d5e6, 30c8308738ee, ...) select owner roles negatively, via
+    # ``NOT (role_name LIKE '%member')``. That is unsafe here: role names carry no enforced
+    # convention (``CreateRoleInput.name`` is a free string and the caller picks the scope), so a
+    # custom read-only ``domain``-scoped role matches and would receive fragment ``hard-delete``
+    # and ``grant:all`` over the whole domain. Select admins positively instead — by the
+    # capability each domain admin role provably holds rather than by the shape of its name:
+    #
+    # * ``domain`` — the role must hold ``domain_admin_page`` at that same domain. Every domain
+    #   admin role holds it under both naming schemes by the time this runs: 3b6297b1bd75 covers
+    #   ``role_domain_%_admin``, f2b9a4c7e103/a3c1d8e5b294 cover ``domain-%-admin``, and all three
+    #   are ancestors of this revision. ``domain_admin_page`` is absent from the frozen enum, so
+    #   it is inlined as a literal, as those three migrations do.
+    # * ``user`` — kept broad. A user-scope grant only ever reaches that user's own scope, and
+    #   writing an own-scope fragment is deliberately not admin-only (see
+    #   ``AppConfigScopeType.to_rbac_scope_type``).
+    #
+    # ``role_superadmin`` holds domain-scoped rows but no ``domain_admin_page``, so it drops out.
+    # That is correct and matches 3b6297b1bd75: superadmin authority comes from the
+    # ``user.is_superadmin`` short-circuit in the RBAC validators, never from these rows.
     db_conn = op.get_bind()
     db_conn.execute(
         sa.text("""
             WITH role_scopes AS (
                 SELECT DISTINCT
                     p.role_id,
-                    r.name AS role_name,
                     p.scope_type,
                     p.scope_id
                 FROM permissions p
-                JOIN roles r ON p.role_id = r.id
                 WHERE p.scope_type IN ('user', 'domain')
             ),
             role_operations AS (
@@ -58,7 +75,15 @@ def upgrade() -> None:
                     rs.scope_id,
                     unnest(CAST(:owner_ops AS text[])) AS operation
                 FROM role_scopes rs
-                WHERE NOT (rs.role_name LIKE '%member')
+                WHERE rs.scope_type = 'user'
+                   OR EXISTS (
+                        SELECT 1
+                        FROM permissions admin_page
+                        WHERE admin_page.role_id = rs.role_id
+                          AND admin_page.entity_type = 'domain_admin_page'
+                          AND admin_page.scope_type = rs.scope_type
+                          AND admin_page.scope_id = rs.scope_id
+                   )
             )
             INSERT INTO permissions (
                 role_id, scope_type, scope_id, entity_type, operation, permission
@@ -88,8 +113,11 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
-    db_conn = op.get_bind()
-    db_conn.execute(
-        sa.text("DELETE FROM permissions WHERE entity_type = :entity_type"),
-        {"entity_type": EntityType.APP_CONFIG_FRAGMENT.value},
-    )
+    # No-op, following f2b9a4c7e103. Deleting by ``entity_type`` cannot distinguish the rows
+    # this backfill inserted from the ones granted natively at role creation: current code
+    # already writes APP_CONFIG_FRAGMENT permissions for every new role, because the entity
+    # type is part of ``_resource_types()``. A blanket DELETE would therefore strip permissions
+    # that roles created since this revision legitimately hold, and re-running ``upgrade`` would
+    # not restore them for roles the backfill does not select. Leaving the rows in place is
+    # safe — every row this migration writes is one role creation would have granted anyway.
+    pass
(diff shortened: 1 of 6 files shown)


src/ai/backend/manager/models/alembic/versions/b3d9f7a2c184_add_app_config_fragment_permissions_to_rbac.py
```diff
@@ -0,0 +1,95 @@
...
+        },
+    )
+
+
+def downgrade() -> None:
+    db_conn = op.get_bind()
+    db_conn.execute(
+        sa.text("DELETE FROM permissions WHERE entity_type = :entity_type"),
+        {"entity_type": EntityType.APP_CONFIG_FRAGMENT.value},
+    )
```
fregataa: This is dangerous, let's not delete during downgrade migration

src/ai/backend/manager/models/alembic/versions/b3d9f7a2c184_add_app_config_fragment_permissions_to_rbac.py
```diff
@@ -0,0 +1,95 @@
...
+                WHERE p.scope_type IN ('user', 'domain')
+            ),
+            role_operations AS (
+                SELECT
+                    rs.role_id,
+                    rs.scope_type,
+                    rs.scope_id,
+                    unnest(CAST(:owner_ops AS text[])) AS operation
+                FROM role_scopes rs
+                WHERE NOT (rs.role_name LIKE '%member')
```
fregataa: As copilot reviewed, this can add extra permissions to non-admin roles.
If you want to add permission to 'admins' only, then it would be good to decide by checking whether a role has `PROJECT_ADMIN_PAGE` entity type permission.
There is no `DOMAIN_ADMIN_PAGE` yet, but all domain admin roles are made with `domain-...-admin` name patterns.

src/ai/backend/manager/models/alembic/versions/b3d9f7a2c184_add_app_config_fragment_permissions_to_rbac.py
jopemachine: I see in other PRs that added permission records that those records were removed during the downgrade. Are you suggesting that we leave downgrade as pass here instead?

Is there a particular reason for that?

src/ai/backend/manager/models/alembic/versions/b3d9f7a2c184_add_app_config_fragment_permissions_to_rbac.py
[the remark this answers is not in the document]
jopemachine: Confirmed and fixed in 1f88bb0. Your `domain-viewer` pointer was exact — that preset (`ScopeType.DOMAIN`, `VFOLDER:READ` + `SESSION:READ`) is precisely the shape that slipped through.

I reproduced it against a local DB before fixing: a read-only domain-scoped custom role received `create` / `update` / `hard-delete` **and** `grant:all` on `app_config_fragment` across the whole domain. The grant rows are not inert either — the live check matches on `permissions.operation`, and the bitmask is the half that is currently unread.

Switched to your second suggestion (derive eligibility from existing capabilities) rather than name matching:

```sql
WHERE rs.scope_type = 'user'
   OR EXISTS (SELECT 1 FROM permissions admin_page
              WHERE admin_page.role_id     = rs.role_id
                AND admin_page.entity_type = 'domain_admin_page'
                AND admin_page.scope_type  = rs.scope_type
                AND admin_page.scope_id    = rs.scope_id)
```

The `roles` join drops out entirely. Verified: both admin naming schemes still get the grant; the read-only custom role, member roles, and `role_superadmin` get nothing.

jopemachine: Took this approach in 1f88bb0 — deciding by capability rather than by name is the right call, and I reproduced the escalation against a local DB first to confirm it was real (a read-only domain role got `hard-delete` + `grant:all` domain-wide).

Two notes where the code turned out to differ from the premises, both of which make the fix simpler:

**`DOMAIN_ADMIN_PAGE` does exist** — `common/data/permission/types.py:110` (`EntityType`) and `:427` (`RBACElementType`), granted to domain admins by `DomainData.entity_operations()`. And it is already backfilled for **both** naming schemes by ancestors of this revision:

| migration | covers | ancestor of `b3d9f7a2c184`? |
|---|---|---|
| `3b6297b1bd75` | `role_domain_%_admin` (legacy) | yes |
| `f2b9a4c7e103` / `a3c1d8e5b294` | `domain-%-admin` (runtime) | yes |

So no `domain-...-admin` name pattern is needed — every domain admin role provably holds the capability by the time this runs, and the `roles` join drops out entirely.

**Used `domain_admin_page`, not `PROJECT_ADMIN_PAGE`** — a fragment has no project scope. `AppConfigScopeType` is `PUBLIC` / `DOMAIN` / `USER` only.

**`user` scope kept broad, not admin-only** — `AppConfigScopeType.to_rbac_scope_type()` states it directly: *"writing a fragment is not admin-only — an allow-listed user may write their own `user`-scope fragment"*. A user-scope grant only ever reaches that user's own scope, so it does not carry the domain-wide risk.

**`public` is granted to nobody**, so domain admins cannot write public fragments — the migration only ever writes `user` / `domain` rows. Superadmin reaches public through the `is_superadmin` short-circuit in the RBAC validators, which needs no rows. `role_superadmin` therefore drops out of the backfill too, matching `3b6297b1bd75`.

src/ai/backend/manager/models/alembic/versions/b3d9f7a2c184_add_app_config_fragment_permissions_to_rbac.py
```diff
@@ -0,0 +1,119 @@
+"""add_app_config_fragment_permissions_to_rbac
+
+Backfill ``APP_CONFIG_FRAGMENT`` permissions onto write-capable roles that predate the
+entity type (BEP-1052); new roles receive them at role creation. Only ``permissions`` rows
+are backfilled — the scope associations are written per-fragment at fragment-creation time,
+and no fragments exist yet.
+
+Revision ID: b3d9f7a2c184
+Revises: c4e1a9b73f52
+Create Date: 2026-07-14 00:00:00.000000
```
fregataa, line 10: the timestamp ... 👀

fregataa (approved):

PR #12860, opened 2026-07-15
feat(BA-6882): expose kernel scheduling history via repository/service/adapter/REST v2/SDK/CLI

jopemachine:
Exposes `kernel_scheduling_history` end-to-end through the v2 backend + REST + SDK + CLI, mirroring the existing session/deployment/route slices. GraphQL is intentionally out of scope — BA-6883 adds it on top of the adapter/service built here.

## What was already there

The table, ORM row, domain data types, and an admin-only `search_kernel_history` repository method already existed. Everything above the repository boundary was missing, and the repository had no scoped variant.

## Layers added

| Layer | Added |
|---|---|
| `models/scheduling_history/conditions.py` | Kernel MatchSpec filter variants + string / datetime / result conditions |
| `models/scheduling_history/orders.py` | `KERNEL_ORDER_FIELD_MAP`, default/tiebreaker orders, `resolve_kernel_order` |
| `common/data/permission/types.py` | `EntityType.KERNEL_HISTORY`, `KERNEL_SCOPED_HISTORY` |
| `repositories/scheduling_history/` | `KernelSchedulingHistorySearchScope` + `search_kernel_scoped_history` (repository + db_source) |
| `common/dto/manager/v2/scheduling_history/` | `KernelHistoryFilter` / `Order` / `Node`, `KernelHistoryScopeDTO`, admin + scoped inputs, payload |
| `services/scheduling_history/` | 2 actions, 2 service methods, processor wiring |
| `api/adapters/scheduling_history/adapter.py` | `admin_search_kernel_history`, `kernel_scoped_search`, querier/filter/order/DTO conversion |
| `api/rest/v2/scheduling_history/` | `POST /kernels/search`, `POST /kernels/scoped/search` |
| `client/v2` + `client/cli/v2` | `search_kernel_history` / `kernel_scoped_search`, `bai scheduling-history kernel search / search-scoped` |

## Two deliberate deviations from the existing slices

Both follow the **documented forward direction**, which the existing session/deployment/route code predates.

**1. The scope is a body field, not a path param.** A kernel query is scoped by `session_id`, `kernel_id`, or both, and an empty scope must be rejected — none of which a path param can express. `api/rest/v2/AGENTS.md` names this exact shape as the forward direction (`/v2/{entity}/scoped/search` + scope as a request body field), and `client/v2/AGENTS.md` names the matching `scoped_search(request)` SDK form. Hence the route `/kernels/scoped/search` and the `scoped_` handler prefix.

Non-emptiness is validated once, in the shared DTO `model_validator` (`KernelHistoryScopeDTO`), and defended again in the repository scope.

**2. The scoped route uses `auth_required`, not `superadmin_required`.** `api/rest/v2/AGENTS.md` states that all scoped search routes use `auth_required`, and this story requires a **non-admin** scoped search. Note the existing session/deployment/route scoped routes are all registered `superadmin_required`, which appears to defeat the purpose of a scoped search — left alone as out of scope here.

The CLI follows from (1): `search-scoped` takes `--session-id` / `--kernel-id` rather than a positional ID, and errors cleanly when neither is given.

## Notes

- **`sub_steps` is intentionally absent** from `KernelHistoryNode`: `kernel_scheduling_history` has no such column (the `84b901f69d16` migration added it to deployment/route only). BEP-1061 and BA-6852 both describe kernel as having `sub_steps` — that discrepancy needs resolving on the writer side.
- **Overlaps #12857** (BA-6887) on ~14 lines of `conditions.py`: this branch carries the same `from_phase` → `from_status` correction, since it is the first caller of those factories. Whichever lands second needs a trivial conflict resolution.
- **Live CLI verification is not yet possible.** `kernel_scheduling_history` has no writer, so the table is empty; BA-6852 adds it. Every other acceptance criterion is met in code.


jopemachine:
Superseded by a stack of smaller PRs, one per layer: #12867 (BA-6889), #12868 (BA-6890), #12869 (BA-6891), #12870 (BA-6892), #12866 (BA-6893). Same change, reviewable in ~300-450 line pieces.

PR #12868, opened 2026-07-15
feat(BA-6890): add kernel scheduling-history repository search scope

jopemachine:
## 📚 Stacked PRs

- #12857 — `BA-6887` fix: kernel history conditions reference the real status columns
- #12867 — `BA-6889` models: kernel query conditions and orders
- #12868 — `BA-6890` repository: kernel search scope and scoped search  ← you are here
- #12869 — `BA-6891` DTO + service: kernel v2 DTOs and search actions
- #12870 — `BA-6892` adapter + REST v2: kernel endpoints
- #12866 — `BA-6893` SDK + CLI: kernel commands

## What

`KernelSchedulingHistorySearchScope` plus the scoped repository read path (`search_kernel_scoped_history` on the repository and db_source).

## Why this scope is shaped differently

The session / deployment / route scopes each carry **one required entity id**. A kernel query is scoped by:

- `session_id` — all of that session's kernels, or
- `kernel_id` — one kernel, or
- **both** — intersected.

`to_condition()` filters on each axis given and `AND`s them; `existence_checks` covers the corresponding `SessionRow` / `KernelRow`.

At least one axis is required. An empty scope would silently degenerate into an unscoped, system-wide search — exactly what a scoped search exists to prevent — so it raises `EmptyKernelSchedulingHistoryScope`. The shared DTO validates this too (#12869); the scope defends it again because the scope is what the repository actually trusts, and it is reachable from GraphQL (BA-6883) as well as REST.

## Review notes

- `search_kernel_scoped_history` differs from the existing unscoped method by exactly one argument: `scopes=[scope]` on `execute_batch_querier`.
- Tests cover each axis combination, the intersection, existence checks, and the empty-scope rejection. `tests/AGENTS.md` names "empty scope → error" as a canonical case to cover.

The service actions that build this scope are #12869.

### DIFF
--- a/src/ai/backend/manager/repositories/scheduling_history/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/scheduling_history/db_source/db_source.py
@@ -28,6 +28,7 @@ from ai.backend.manager.repositories.base import (
 )
 from ai.backend.manager.repositories.scheduling_history.types import (
     DeploymentHistorySearchScope,
+    KernelSchedulingHistorySearchScope,
     RouteHistorySearchScope,
     SessionSchedulingHistorySearchScope,
 )
@@ -94,7 +95,7 @@ class SchedulingHistoryDBSource:
                 has_previous_page=result.has_previous_page,
             )
 
-    # ========== Kernel History ==========
+    # ========== Kernel History (Admin) ==========
 
     async def search_kernel_history(
         self,
@@ -109,6 +110,28 @@ class SchedulingHistoryDBSource:
                 query,
                 querier,
             )
+
+            items = [row.KernelSchedulingHistoryRow.to_data() for row in result.rows]
+
+            return KernelSchedulingHistoryListResult(
+                items=items,
+                total_count=result.total_count,
+                has_next_page=result.has_next_page,
+                has_previous_page=result.has_previous_page,
+            )
+
+    # ========== Kernel History (Scoped) ==========
+
+    async def search_kernel_scoped_history(
+        self,
+        querier: BatchQuerier,
+        scope: KernelSchedulingHistorySearchScope,
+    ) -> KernelSchedulingHistoryListResult:
+        """Search kernel scheduling history within scope."""
+        async with self._db.begin_readonly_session() as db_sess:
+            query = sa.select(KernelSchedulingHistoryRow)
+
+            result = await execute_batch_querier(db_sess, query, querier, scopes=[scope])
 
             items = [row.KernelSchedulingHistoryRow.to_data() for row in result.rows]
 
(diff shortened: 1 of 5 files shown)


src/ai/backend/manager/repositories/scheduling_history/types.py
```diff
@@ -62,6 +71,73 @@ def existence_checks(self) -> list[ExistenceCheck[Any]]:
...
+
+@dataclass(frozen=True)
+class KernelSchedulingHistorySearchScope(SearchScope):
+    """Scope for kernel scheduling history search.
+
+    Either axis may be given; when both are, they intersect. At least one is required —
+    an empty scope would degenerate into an unscoped (admin) search.
+    """
+
+    session_id: UUID | None = None
```
jopemachine: SessionID

src/ai/backend/manager/repositories/scheduling_history/types.py
```diff
@@ -62,6 +71,73 @@ def existence_checks(self) -> list[ExistenceCheck[Any]]:
...
+    """Restrict to the kernels of this session."""
+
+    kernel_id: UUID | None = None
```
jopemachine: KernelID

src/ai/backend/manager/repositories/scheduling_history/types.py
jopemachine: Done in 05a2d77 — `session_id` is now `SessionId`.

src/ai/backend/manager/repositories/scheduling_history/types.py
jopemachine: Done in 05a2d77 — `kernel_id` is now `KernelId`.

src/ai/backend/manager/repositories/scheduling_history/db_source/db_source.py
```diff
@@ -119,6 +120,28 @@ async def search_kernel_history(
...
+    async def search_kernel_scoped_history(
+        self,
+        querier: BatchQuerier,
+        scope: KernelSchedulingHistorySearchScope,
+    ) -> KernelSchedulingHistoryListResult:
+        """Search kernel scheduling history within scope."""
+        async with self._db.begin_readonly_session() as db_sess:
+            query = sa.select(KernelSchedulingHistoryRow)
+
+            result = await execute_batch_querier(db_sess, query, querier, scopes=[scope])
```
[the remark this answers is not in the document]
jopemachine: The SQL-rendering and ExistenceCheck tests you flagged are removed in 8332a48 — you were right that they assert nothing observable. The file now keeps only the empty-scope rejection, which is a contract of the type rather than a rendering detail: with both axes unset to_condition() builds sa.and_() with no clauses, which renders empty and lifts the WHERE restriction on a route that is only auth_required.

The real-DB scoped tests you asked for are deliberately not in this PR. Tracking them separately so they land next to the existing with_tables suite in test_scheduling_history_repository.py rather than as a second, thinner harness.

HyeockJinKim (approved):

PR #12871, opened 2026-07-15
feat(BA-6888): register the scope itself as an entity member of its virtual scope on creation

fregataa:
## Summary
- On scope creation, register the scope itself as an `entity_membership` under its own virtual scope so access to the scope object resolves as an ordinary entity through single-path permission resolution — no separate scope-permission concept.
- Adds `RBACWriteOps._insert_scope_self_memberships`, a single idempotent `INSERT … SELECT` from `virtual_scopes` (VS `id` → `virtual_scope_id`, `scope_type`/`scope_id` → `entity_type`/`entity_id`, `permission_cap` NULL), wired into `create_scope` and `bulk_create_scopes`.
- Change is confined to `RBACWriteOps`; resolution and other layers are untouched.

## Test plan
- [x] `pants check` / `pants test` pass in CI

Resolves BA-6888

### DIFF
--- a/src/ai/backend/manager/repositories/ops/rbac/provider.py
+++ b/src/ai/backend/manager/repositories/ops/rbac/provider.py
@@ -102,7 +102,12 @@ class RBACWriteOps(WriteOps):
         return virtual_scope_id
 
     async def _insert_virtual_scopes(self, scopes: Sequence[ScopeRef]) -> None:
-        """Get-or-create the virtual scope nodes for ``scopes`` (idempotent)."""
+        """Create the virtual scope nodes for ``scopes`` and register each scope
+        as an entity member of its own virtual scope (idempotent).
+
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/repositories/ops/rbac/provider.py
```diff
@@ -174,6 +202,7 @@ async def create_scope[TRow: Base](
         """
         result = await self.create(creation.creator)
         await self._insert_virtual_scopes([creation.scope])
+        await self._insert_scope_self_memberships([creation.scope])
```
HyeockJinKim: In my opinion, we should insert entity row in 'insert_virtual_scopes' function.
Don't split this behavior.

HyeockJinKim (approved):

HyeockJinKim (approved):

PR #12878, opened 2026-07-15
feat(BA-6894): implement network timeout idle checker

seedspirit:
## Summary

- Add a reconciler-based `NetworkTimeoutChecker` that returns `ACTIVE`, `IDLE`, or `IDLE_EXPIRED` with the projected expiration time and state-specific message.
- Define `NetworkTimeoutSpec.max_network_inactivity_seconds` and batch-load last-access markers and active app-connection counts from Valkey.
- Wire the Valkey live client through the orchestration dependencies and register the network checker in the idle-check judgment stage.

## Test plan

- [x] `pants fmt ::`
- [x] `pants fix ::`
- [x] Pre-push `lint/check/test` for `origin/main..HEAD`
- [ ] CI green

Resolves BA-6894


src/ai/backend/manager/sokovan/idle_check/checkers/network_timeout.py
```diff
@@ -0,0 +1,115 @@
...
+            # Skip sessions with no timeout configured (0 means "never idle").
+            if network_spec.idle_timeout_seconds == 0:
+                continue
+            idle_timeout_seconds = Decimal(network_spec.idle_timeout_seconds)
+            for session in assignment.sessions:
+                state = states[session.session_id]
+                if state.last_access is None:
+                    continue
+                current_time = Decimal(str(context.current_time.timestamp()))
+                idle_seconds = (current_time - state.last_access).normalize()
```
fregataa: Are current_time and last_access from the same source?

src/ai/backend/manager/sokovan/idle_check/checkers/network_timeout.py
```diff
@@ -0,0 +1,115 @@
...
+        self,
+        sessions: Sequence[IdleCheckSession],
+    ) -> dict[SessionId, _NetworkIdleState]:
+        session_ids = list(dict.fromkeys(session.session_id for session in sessions))
+        last_access_values, active_connection_counts = await asyncio.gather(
+            self._valkey_live.get_multiple_live_data([
+                f"session.{session_id}.last_access" for session_id in session_ids
+            ]),
+            self._valkey_live.count_active_connections_batch(session_ids),
+        )
```
fregataa, line 139: Are they grouped for concurrent query?

src/ai/backend/manager/sokovan/idle_check/checkers/network_timeout.py
seedspirit: "current time" is injected by the idle checker, and "last_access" is the value stored in valkey.

src/ai/backend/manager/sokovan/idle_check/checkers/network_timeout.py
seedspirit: Could have run the queries separately, but I used `gather` because we thought running them simultaneously would yield slightly higher accuracy. It’s not required, so we can run them separately if you prefer.

src/ai/backend/manager/sokovan/idle_check/checkers/network_timeout.py
```diff
@@ -0,0 +1,152 @@
...
+        self,
+        *,
+        checker_id: IdleCheckerID,
+        session_id: SessionId,
+        state: _NetworkIdleState,
+        max_inactivity_seconds: int,
+        current_time: datetime,
+    ) -> IdleJudgment | None:
+        if state.last_access is None:
+            return None
```
seedspirit: If the `last_access` value in valkey has expired (with a TTL set to about one day), it may return `None`. I'm wondering whether it would be best to skip the `judge` in this case.

src/ai/backend/manager/sokovan/idle_check/checkers/network_timeout.py
```diff
@@ -0,0 +1,152 @@
...
+                session_id=session_id,
+                expire_at=current_time + timedelta(seconds=max_inactivity_seconds),
+                status=IdleCheckPhase.ACTIVE,
+                message=(
+                    "Network activity detected: "
+                    f"max_network_inactivity_seconds={max_inactivity_seconds}, "
+                    f"active_connections={state.active_connections}"
+                ),
+            )
+        last_access_at = datetime.fromtimestamp(float(state.last_access), tz=UTC)
```
fregataa: How about setting `_NetworkIdleState.last_access` as float?

jopemachine (approved):

PR #12910, opened 2026-07-16
fix(BA-6914): stop dropping FKs by a name the database may not use

jopemachine:
## 📚 Stacked PRs

- #12910 — `BA-6914` fix: migrations that drop a foreign key by a name the DB may not use  ← you are here
- #12907 — `BA-6913` fix: indexes that only exist on migrated databases
- #12912 — `BA-6915` fix: drop the column default before retyping sessions.result

## Problem

Three migrations create a foreign key under a **hardcoded** name; the models left the same key unnamed and let the metadata naming convention generate one. Same key, same enforcement — different name depending on how the DB was built:

| | migration creates | fresh install has |
|---|---|---|
| `cd067180a8b1` | `fk_kernels_image_id` | `fk_kernels_image_id_images` |
| `c7f2a8e31b04` | `fk_login_sessions_login_client_type_id` | `fk_login_sessions_login_client_type_id_login_client_types` |
| `a3b4c5d6e7f8` | `fk_prometheus_query_presets_category_id` | `fk_prometheus_query_presets_category_id_prometheus_quer_8a51` ¹ |

¹ the convention name exceeds PostgreSQL's 63-char identifier limit, so SQLAlchemy truncates it and appends a hash.

**Not cosmetic:** each downgrade drops the constraint by the hardcoded name and aborts with `UndefinedObjectError` on a fresh install. Whether a downgrade works depends on where the DB came from.

## Fix

**The drop is redundant.** In all three, `drop_constraint` is immediately followed by dropping the very column the key constrains — and `DROP COLUMN` removes the constraint whatever its name. So the three calls come out, and nothing has to agree on a name at all.

That leaves the names diverged, which is fine: they are referenced nowhere outside their own migration files (verified by grep across the tree), and 81 of 82 model FKs are unnamed, so the convention names are the house style.

Also fixed, same assumption: two of the `upgrade()` guards test `fk["name"] not in [...]`. On a database whose key carries the convention name, the guard misses it and `create_foreign_key` adds a **second, redundant** FK on the same column. They now match on the constrained column.

## What this replaces

An earlier revision of this PR converged the names instead — `name=` on the three models plus a migration renaming the constraint on databases built from the convention. Dropping the redundant calls is strictly less: no new migration, no `ALTER` on production databases, no model changes, and the released migrations' `upgrade()` paths are untouched.

## Verification

`DROP COLUMN` removing a differently-named FK, on a real PostgreSQL — a constraint named `weird_name` on the dropped column, `0` foreign keys remaining afterward.

Not re-run since the rewrite: the fresh-install downgrade walls the previous approach was verified against. The change is a strict deletion of a statement whose effect the following `DROP COLUMN` already performs, but the three downgrades are worth re-running before merge.

### DIFF
--- a/changes/12910.fix.md
+++ b/changes/12910.fix.md
@@ -1,0 +1,1 @@
+Fix migration downgrades that aborted on fresh installs by dropping a foreign key under a hardcoded name the database may not use
--- a/src/ai/backend/manager/models/alembic/versions/a3b4c5d6e7f8_add_preset_category_description_rank.py
+++ b/src/ai/backend/manager/models/alembic/versions/a3b4c5d6e7f8_add_preset_category_description_rank.py
@@ -66,11 +66,6 @@ def upgrade() -> None:
 
 def downgrade() -> None:
     op.drop_index("ix_prometheus_query_presets_category_id", table_name="prometheus_query_presets")
-    op.drop_constraint(
-        "fk_prometheus_query_presets_category_id",
-        "prometheus_query_presets",
-        type_="foreignkey",
-    )
     op.drop_column("prometheus_query_presets", "category_id")
     op.drop_column("prometheus_query_presets", "rank")
     op.drop_column("prometheus_query_presets", "description")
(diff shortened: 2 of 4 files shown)


jopemachine:

PR #12912, opened 2026-07-16
fix(BA-6915): drop the column default before retyping sessions.result

jopemachine:
## 📚 Stacked PRs

- #12910 — `BA-6914` fix: migrations that drop a foreign key by a name the DB may not use
- #12907 — `BA-6913` fix: indexes that only exist on migrated databases
- #12912 — `BA-6915` fix: drop the column default before retyping sessions.result  ← you are here

## Problem

`0b1efbb2db84`'s downgrade recreates the plural `sessionresults` type and then:

```sql
ALTER TABLE sessions ALTER COLUMN result TYPE sessionresults USING result::text::sessionresults
```

`sessions.result` carries a default. PostgreSQL cannot cast an existing default across two unrelated enum types, so this aborts before touching a row:

```
DatatypeMismatchError: default for column "result" cannot be cast automatically to type sessionresults
```

The migration guards its DDL with a `try/except` over a list of expected-divergence SQLSTATEs — but `42804` isn't on it, so the error is reraised and **the whole downgrade chain stops here**.

## The correct order already exists in this repo

`dec0deba5893` does the same conversion and succeeds, because it handles the default:

```python
_drop_column_default(conn, table, column)
_alter_column_to_sessionresult(conn, table, column)
_set_column_default_undefined_sessionresult(conn, table, column)
```

`0b1efbb2db84` just omits it. This PR gives it the same order, on both the downgrade and the mirror-image `ALTER` in `upgrade()`.

## Reproduction

On a DB built at head by `mgr schema oneshot`, downgraded with **no skipped revisions**:

1. 150 downgrade steps run cleanly.
2. `dec0deba5893`'s downgrade recreates the singular `sessionresult` and converts `sessions.result` to it.
3. Crossing `0b1efbb2db84`, the `has_singular and not has_plural` branch fires and aborts.

Confirmed by hand that `DROP DEFAULT` → `ALTER COLUMN TYPE` → `SET DEFAULT` completes, leaving `result` as `sessionresults` with default `'UNDEFINED'::sessionresults`.

## Verification

With this fix on top of #12881, #12883, #12887, #12907, #12910, a single straight `alembic downgrade` from head to `d86417a6bf7b` — below the deepest RBAC backfill — now completes:

```
exit=0   steps run: 175   (no skipped revisions)
current: d86417a6bf7b
```

That path previously stopped after 28 steps. Each fix moved it further: 28 → 84 → 150 → **175, target reached**.

### DIFF
--- a/changes/12912.fix.md
+++ b/changes/12912.fix.md
@@ -1,0 +1,1 @@
+Fix the session result enum downgrade aborting when the column default cannot be cast to the recreated type
--- a/src/ai/backend/manager/models/alembic/versions/0b1efbb2db84_fix_sessionresult_enum_type_coexistence.py
+++ b/src/ai/backend/manager/models/alembic/versions/0b1efbb2db84_fix_sessionresult_enum_type_coexistence.py
@@ -41,6 +41,23 @@ def _is_expected_divergence(exc: BaseException) -> bool:
     return _sqlstate_of(exc) in _EXPECTED_DIVERGENCE_SQLSTATES
 
 
+def _retype_result_column(conn: sa.engine.Connection, table: str, target_type: str) -> None:
+    """Change ``<table>.result`` to ``target_type``, carrying its default across.
+
+    PostgreSQL cannot cast an existing column default between two unrelated enum
+    types, so ALTER COLUMN ... TYPE fails outright while a default is attached.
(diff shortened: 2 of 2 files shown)


jopemachine:

PR #12917, opened 2026-07-16
feat(BA-6827): add agent resource-group update service

fregataa:
## Summary
- Add `AgentService.update_resource_group`: change an agent's resource group and handle the sessions still running on it in a single flow.
- The action carries the target `resource_group_id` (caller-resolved), a cleanup policy (`TERMINATE` / `RESCHEDULE`), and `force`.
- `AgentRepository.update_resource_group` atomically gates on the agent's active kernels — it raises `AgentHasConflictingSessions` when they exist and `force` is unset (the admin drains first), otherwise commits the group change (both the id and name columns) and returns those kernels. The service then marks their sessions for termination (graceful → `TERMINATING`, cleanup proceeds asynchronously).
- `RESCHEDULE` is design-only for now (depends on the RESCHEDULING state from the preemption work) and is rejected up front.

## Test plan
- [x] `pants test tests/unit/manager/services/agent/test_service.py` — service unit: no-op commit / drain-first failure / force terminate / reschedule rejected
- [x] `pants test tests/unit/manager/repositories/agent/test_repository.py` — real-DB: no-kernel commit updates both columns; active-kernel force gate
- [x] CI: `pants check` / `pants lint`

Resolves BA-6827


src/ai/backend/manager/repositories/agent/db_source/db_source.py
```diff
@@ -181,6 +185,54 @@ async def update_agent_status(self, updater: Updater[AgentRow]) -> None:
...
+                sa.update(agents)
+                .where(agents.c.id == agent_id)
+                .values(
+                    resource_group_id=resource_group_id,
+                    scaling_group=sa.select(ScalingGroupRow.name)
+                    .where(ScalingGroupRow.id == resource_group_id)
+                    .scalar_subquery(),
+                )
+            )
+        return kernels
```
HyeockJinKim, line 234: If the "force" option is selected, won't the related kernel work be addressed in this PR?

HyeockJinKim (approved):

PR #12925, opened 2026-07-16
feat(BA-6917): add expiry-sweep source fetching expired idle-check deadlines

seedspirit:
## Summary
- Implement the Source path of the BEP-1054 expiry-sweep stage: read `session_idle_checks` rows whose deadline has passed (`expire_at IS NOT NULL AND expire_at <= now()`, matching the partial index) joined to sessions in the target statuses
- `IdleCheckerRepository.fetch_expired_idle_checks` / `IdleCheckerDBSource.fetch_expired_idle_checks` return `ExpiredIdleCheckBatchData` — per-checker expired checks (kept as separate reasons per session) plus a DB-sourced `now` read in the same transaction
- Add `IdleCheckSweepSource` + `IdleCheckSweepReconcileInfo` under `sokovan/idle_check/sweep/` implementing the `ReconcilerSource` contract (`entity_ids()` deduplicates sessions, `now()` is the DB time); Handler/Applier and stage registration follow in a later PR

## Test plan
- [x] `tests/unit/manager/repositories/idle_checker/test_repository.py::TestFetchExpiredIdleChecks` — per-kind seeded fixtures (real DB via `with_tables`): returns each expired check of a running session with stored judgment fields, excludes NULL/future deadlines and non-target-status sessions, `now` covers every returned deadline
- [x] `pants lint/check/test` on the changed files (pre-push hook)

Resolves BA-6917

### DIFF
--- a/src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from collections.abc import Sequence
+from datetime import datetime
 from typing import cast
 
 import sqlalchemy as sa
@@ -22,7 +23,6 @@ from ai.backend.manager.models.session.row import SessionRow
 from ai.backend.manager.repositories.base import BatchQuerier
 from ai.backend.manager.repositories.idle_checker.types import (
     BoundCheckerData,
-    ExpiredIdleCheckBatchData,
     ExpiredIdleCheckData,
     IdleCheckerDefinitionData,
     IdleCheckSessionData,
@@ -35,6 +35,10 @@ class IdleCheckerDBSource:
 
     def __init__(self, ops_provider: DBOpsProvider) -> None:
         self._ops = ops_provider
+
+    async def current_time(self) -> datetime:
+        async with self._ops.read_ops() as r:
+            return await r.current_time()
 
     async def fetch_bound_checkers(
         self,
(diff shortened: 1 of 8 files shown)


src/ai/backend/manager/repositories/idle_checker/types.py
```diff
@@ -57,3 +57,26 @@ class IdleCheckBatchData:
...
+    last_status: str
+    last_message: str
+
+
+@dataclass(frozen=True)
+class ExpiredIdleCheckBatchData:
+    """Sweep input for one reconciler tick.
+
+    ``now`` is the DB time read in the same transaction as the fetch, so every
+    returned check satisfies ``expire_at <= now``.
```
fregataa: This comment is not correct, there is no such comparison condition in fetch_expired_idle_checks(). And will idle checkers use the DB fetched current time to decide whether it is expired or not?

seedspirit: Thanks. The expiration comparison is applied by the repository through `SessionIdleCheckConditions.expired()` (`expire_at <= now()`), rather than using `batch.now`. The returned `now` is the DB transaction timestamp used by the reconciler. I updated the docstring to make this distinction explicit.

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
```diff
@@ -100,3 +106,31 @@ async def fetch_candidate_sessions(
...
+        self,
+        querier: BatchQuerier,
+    ) -> ExpiredIdleCheckBatchData:
+        check_query = sa.select(SessionIdleCheckRow).join(
+            SessionRow, SessionIdleCheckRow.session_id == SessionRow.id
+        )
+        async with self._ops.read_ops() as r:
+            # Same transaction as the fetch, so now >= every returned expire_at.
+            now = await r.current_time()
+            result_rows = (await r.batch_query_in_global(check_query, querier)).rows
```
HyeockJinKim: It's strange that it reads `now` here but uses `sa.func.now` instead of this value.

jopemachine (dismissed):

HyeockJinKim (approved):

PR #12931, opened 2026-07-16
feat(BA-6659): materialize the Project scope's virtual scope on creation

fregataa:
## Summary
- `GroupDBSource.create` now materializes the project's virtual scope node (with its self-membership) in the same transaction as the project row and its RBAC roles, so a project becomes capable of owning child entities through the virtual-scope chain.
- Add `RBACWriteOps.create_virtual_scope`, a public entry point that creates the virtual scope node for a scope entity whose row is created through a separate path (here, `RBACEntityCreator`, which keeps the existing domain→project ownership association).

Resolves BA-6659

### DIFF
--- a/src/ai/backend/manager/repositories/group/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/group/db_source/db_source.py
@@ -16,6 +16,8 @@ from sqlalchemy.engine import CursorResult
 from sqlalchemy.ext.asyncio import AsyncSession as SASession
 
 from ai.backend.common.clients.valkey_client.valkey_stat.client import ValkeyStatClient
+from ai.backend.common.data.entity.types import ScopeRef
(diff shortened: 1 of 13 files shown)


src/ai/backend/manager/repositories/group/db_source/db_source.py
```diff
@@ -180,7 +185,14 @@ async def create(self, creator: Creator[GroupRow]) -> GroupData:
...
 
-            return data
+        async with self._rbac_ops_provider.write_ops() as w:
+            await w.ensure_scope(
+                ScopeRef(
+                    scope_type=VScopeType(RBACElementType.PROJECT.value),
+                    scope_id=data.id,
+                )
+            )
+        return data
```
HyeockJinKim: We'll adjust the timing for that part using ops as well.

HyeockJinKim (approved):

PR #12934, opened 2026-07-17
feat(BA-6923): add ensure_scope op and self/hierarchy scope bindings on creation

fregataa:
## Summary
- Add ensure-virtual-scope RBAC ops that backfill VS structures next to the legacy write path, so an entity can adopt virtual-scope resolution on its own without an all-at-once cutover.
- `_insert_virtual_scopes` now also writes each scope's self `scope_binding`, so an owner resolves its own entities through the single virtual-scope path.
- `create_scope` / `ensure_scope` take an optional `bound_scope` that binds additional scopes to this scope's virtual scope (hierarchy binding). `ensure_scope` backfills the VS node without creating the real scope row.

## Test plan
- [x] `create_scope` materializes the VS node, self entity-membership, and self scope_binding
- [x] `ensure_scope` backfills VS structures without creating the real row, and is idempotent
- [x] `ensure_scope(scope, parent)` adds the hierarchy binding

Resolves BA-6923

### DIFF
--- a/src/ai/backend/manager/repositories/ops/rbac/provider.py
+++ b/src/ai/backend/manager/repositories/ops/rbac/provider.py
@@ -285,18 +285,18 @@ class RBACWriteOps(WriteOps):
     async def create_scope[TRow: Base](
         self,
         creation: ScopeCreation[TRow],
-        parent: ScopeRef | None = None,
+        bound_scope: ScopeRef | None = None,
     ) -> CreatorResult[TRow]:
         """Create the real scope entity via ``creation.creator`` and its virtual scope node.
 
-        ``creation.scope.scope_id`` must match the id the created row carries. When ``parent``
-        is given, the parent scope is bound to this scope's virtual scope so it can reach this
-        scope's entities.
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/repositories/ops/rbac/provider.py
```diff
@@ -409,6 +425,20 @@ async def remove_entity_members(
...
+        self,
+        scope: ScopeRef,
+        parent: ScopeRef | None = None,
+    ) -> None:
+        """Ensure the virtual scope node for an already-created ``scope``. When ``parent`` is
+        given, the parent scope is bound to this scope's virtual scope. Idempotent.
+        """
+        await self._insert_virtual_scopes([scope])
+        if parent is not None:
+            await self.bind_scope(parent, scope, permission_cap=None)
```
HyeockJinKim: It doesn't seem like a good idea to create a virtual scope when the scope hasn't even been created yet.

HyeockJinKim (approved):

PR #12967, opened 2026-07-20
refactor(BA-6927): rename usage_bucket_entries.amount to resource_seconds

jopemachine:
## 📚 Stacked PRs

- #12965 — `BA-6927` fix: stop cross-multiplying usage bucket amounts and durations
- #12967 — `BA-6927` refactor: rename usage_bucket_entries.amount to resource_seconds  ← you are here

## Summary
- Rename `usage_bucket_entries.amount` to `resource_seconds`. The column holds resource-seconds, not a resource amount, and under the old name `amount * duration_seconds` reads like the quantity a caller wants — when both columns are already sums and their product is a cross product, which is exactly the shape of the bug fixed in #12965.
- Mechanical rename across the ORM model, the two repository read paths (`resource_usage_history` and `fair_share`) and the tests, plus an alembic column rename. No behaviour change.
- Split out of #12965 so the correctness fix can be reviewed without the rename churn, and so the rename can be dropped independently if it is not wanted. The table is internal — only the three repository modules read it, and the GraphQL/REST surface exposes the bucket tables' `resource_usage` JSONB rather than these entries.

## Test plan
- [x] `pants test tests/unit/manager/repositories/resource_usage_history/ tests/unit/manager/repositories/retention/`
- [x] `pants fmt/fix/lint` pass on the changeset; `pants check` (mypy) clean
- [x] Confirmed no consumer outside `repositories/{resource_usage_history,fair_share,retention}` references the column

Resolves BA-6927


jopemachine:
Closing: this PR was opened with the wrong head branch. It targeted `feat/BA-6891-kernel-history-dto-service` (unrelated work, tracked in #12869) instead of the rename branch, so its diff is meaningless. Reopening from the correct branch.

PR #12969, opened 2026-07-20
fix: reserve /home/work to prevent duplicate container mount points

hoyajigi:
## Summary
- Setting a vfolder mount alias to exactly `/home/work` passes every alias validation path (`registry.create_session`, sokovan `validate_session_spec`, and `MountNameValidationRule`): the checks only strip the `/home/work/` prefix (with trailing slash), and `RESERVED_VFOLDERS` contains `/home` but not `/home/work`.
- The agent always bind-mounts the scratch directory at `/home/work` (`agent/docker/agent.py`), so such a session passes scheduling and then fails at container creation on the agent with `DockerError(400, "Duplicate mount point: /home/work")` — observed in production on 26.4.3.
- Add `/home/work` to `RESERVED_VFOLDERS` so every validation path rejects the alias up front with `InvalidAPIParameters`.
- `RESERVED_VFOLDERS`/`RESERVED_VFOLDER_PATTERNS` were duplicated between `ai.backend.common.defs` (used by sokovan) and `ai.backend.manager.defs` (used by `verify_vfolder_name`); consolidated onto `common.defs` as the single source of truth and removed the manager copy.

## Test plan
- [x] `pants fmt/fix/lint --changed-since=origin/main` all pass
- [x] Extended `test_vfolder_name_validator` unit test: `verify_vfolder_name("/home/work")` must be rejected while `/home/work/<name>` aliases and a folder literally named `home/work` stay allowed (CI)
- [ ] Manual: create a session with `mount_map` alias `/home/work` → expect `InvalidAPIParameters` at session creation instead of `failed-to-create` at the agent

### DIFF
--- a/changes/12969.fix.md
+++ b/changes/12969.fix.md
@@ -1,0 +1,1 @@
+Reject vfolder mount aliases set to exactly `/home/work`, which collided with the intrinsic scratch mount and made container creation fail with "Duplicate mount point"
--- a/src/ai/backend/common/defs/__init__.py
+++ b/src/ai/backend/common/defs/__init__.py
@@ -34,6 +34,10 @@ RESERVED_VFOLDERS = [
     ".jupyter",
     ".tmux.conf",
     ".ssh",
+    # The agent always bind-mounts the scratch directory at /home/work,
+    # so mounting a vfolder exactly there makes dockerd reject the container
+    # with "Duplicate mount point: /home/work".
+    "/home/work",
     "/bin",
     "/boot",
     "/dev",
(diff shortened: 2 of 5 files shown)


hoyajigi:
Good catch — `sokovan.validate_session_spec()` indeed reads `RESERVED_VFOLDERS` from `ai.backend.common.defs`, which is a duplicate of the list in `ai.backend.manager.defs`. Fixed in 6b6fd83 by consolidating onto the common definition: the reservation is now in `common/defs/__init__.py`, the `manager.defs` copy is removed, and `verify_vfolder_name` imports from `common.defs`, so all validation paths share a single source of truth.

seedspirit (commented):
It doesn't seem to be a big problem, but could @HyeockJinKim take a look at it?

fregataa (approved):

PR #12974, opened 2026-07-20
refactor(BA-6943): authorize scheduling-history scoped searches via RBAC

### LINKED ISSUE #12971: Enforce RBAC on the remaining scoped searches via SearchableActionTarget
## Objective

Convert the scoped search actions that still trust a client-supplied scope to the SearchableActionTarget pattern, so the RBAC scope chain authorizes the caller against the requested scope.

## Background

These actions carry a SearchScope built directly from request input and expose no RBAC element, so validators.rbac never runs. The repository only turns the scope into a WHERE clause, and existence_checks confirms the entity exists rather than that the caller may read it.

Kernel scheduling history was converted in BA-6891 (PR #12869) and is the reference within this domain: SessionKernelHistoryTarget / KernelHistoryTarget implement SearchableActionTarget, the action extends BaseBulkAction[SearchableActionTarget], and the processor is a BulkActionProcessor wired with validators.rbac.bulk. audit_log and app_config_fragment use the same pattern.

## Scope

- scheduling_history: the session, deployment and route scoped searches (kernel is already done).
- Triage the ~18 other actions carrying a client-supplied scope: auth, deployment, domain, dotfile, group, model_card, permission_controller, session, user, vfolder. Convert the ones reachable from non-superadmin routes; record why any are left.

Note: every scheduling-history route other than the kernel scoped one is superadmin_required, so those conversions are hardening rather than an open hole. Split per domain if this exceeds two points.

## Acceptance Criteria

- Converted actions define SearchableActionTarget targets and run through BulkActionProcessor with validators.rbac.bulk.
- A caller requesting a scope they do not own is denied; their own scope succeeds. Both covered by tests.
- Any action deliberately left unconverted is listed with its justification.

JIRA Issue: BA-6943

jopemachine:
Resolves #12971 (BA-6943)

## Summary

- Routes the **session**, **deployment** and **route** scheduling-history scoped searches through the RBAC scope chain. Each defines a `SearchableActionTarget` mapping to both an RBAC element (`SESSION` / `MODEL_DEPLOYMENT` / `ROUTING`) and a search scope; the actions extend `BaseBulkAction` and the processors run through `BulkActionProcessor` with `validators.rbac.bulk`.
- Previously these actions carried a `SearchScope` built straight from request input and exposed no RBAC element, so `validators.rbac` never ran — the repository turned the scope into a `WHERE` clause and `existence_checks` only confirmed the entity existed, not that the caller could read it. The routes papered over this by requiring super-admin, which made the scoped variants useless to the users who actually own the entities.
- **The three scoped routes now drop to `auth_required`.** With authorization enforced on the action path, a user can read the scheduling history of a session/deployment/route they hold, and gets `PermissionDeniedError` for one they do not — not a silently empty page. The unscoped admin searches stay `superadmin_required`.
- Handlers lose the `admin_` prefix, which no longer describes them.

`SchedulingHistoryProcessors.__init__` already accepted `ActionValidators` without using it; that was the wiring point. The repository now takes `Sequence[SearchScope]` so multiple authorized targets can be matched.

Independent of the BA-6889..BA-6893 kernel stack — branched from `main`, touches only the three pre-existing entities. `search_kernel_scoped_history` is deliberately untouched here (converted in #12869); doing it on both branches would conflict.

## Test plan

- [x] `pants check` clean (5251 source files)
- [x] `pants lint` clean over `src/ai/backend/::` and `tests/unit/manager/::`
- [x] Scheduling-history service tests updated to the target shape and passing — each asserts the forwarded scopes and `element_refs()`
- [x] Scheduling-history repository tests passing
- [x] Confirmed `BulkActionRBACValidator` short-circuits for super admins, so the unscoped admin paths are unaffected
- [ ] Verified against a live manager: a non-admin reading their own session's history succeeds, and another user's is denied

----
📚 Documentation preview 📚: https://sorna--12974.org.readthedocs.build/en/12974/

----
📚 Documentation preview 📚: https://sorna-ko--12974.org.readthedocs.build/ko/12974/


jopemachine:
I’ll work on this once the correct implementation for the BulkScope action is in place.

PR #12992, opened 2026-07-21
feat(BA-6828): remove update_scaling_group RPC and rewire ModifyAgent

fregataa:
## Summary
- Remove the manager→agent `update_scaling_group` RPC push: the `AgentRegistry` method, the agent client method, and the agent-side RPC handler including the `agent.toml` rewrite (+`.bak`) and in-memory config update. The RPC existed only to keep the next heartbeat from reverting a group change, which is obsolete now that the manager DB is the source of truth for an agent's resource group.
- Rewire the v1 `ModifyAgent` mutation to the conflicting-session cleanup service (`UpdateAgentResourceGroupAction`) with `policy=TERMINATE` and `force=False`, so the mutation fails with `AgentHasConflictingSessions` when the agent still has sessions under the old group — the admin drains them first.
- Drop the unit tests covering the removed `agent.toml` rewrite / in-memory update behavior.

## Test plan
- [ ] `pants test --changed-since=origin/main` passes in CI
- [ ] Live-server verification (agent group change via `ModifyAgent`, conflicting-session gating) is covered by BA-6831

Resolves BA-6828

### DIFF
--- a/changes/12992.feature.md
+++ b/changes/12992.feature.md
@@ -1,1 +1,1 @@
-Remove the manager-to-agent `update_scaling_group` RPC push; `ModifyAgent` now updates the manager DB as the source of truth and immediately terminates sessions conflicting with the agent's new resource group
+Remove the manager-to-agent `update_scaling_group` RPC push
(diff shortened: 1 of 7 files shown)


HyeockJinKim (approved):

HyeockJinKim:
I do find the "query" in the API a bit annoying, but since it's a minor issue, I'll let it slide.

PR #12996, opened 2026-07-21
feat(BA-6949): seed retention_policies as disabled by default (opt-in)

HyeockJinKim:
## Summary
- Retention swept automatically because the initial migration (`4a39641d0fc2`) seeded eight category rows without an explicit `enabled`, inheriting the column default `true`. This makes DB record retention **opt-in**.
- Follow-up migration `b3e8f1a24c76` flips the `enabled` server_default to `false` and disables the eight seeded rows; the row model's `server_default` is changed to match. Nothing is purged until a super-admin turns a category on.
- Adds `fixtures/manager/example-retention-policies.json` (8 categories, `enabled=false`, default periods) and wires it into `install-dev.sh`. Teaches `populate_fixture` to convert `sa.Interval` fixture values to `timedelta` (asyncpg encodes intervals only from `timedelta`).

## Test plan
- [x] `pants fmt/fix/lint/check` pass
- [x] `pants test` — retention row/repository/sweep + fixture loader tests pass
- [x] Migration upgrade/downgrade verified against local DB (seeded rows→false, new rows inherit false default, downgrade reverses only seeded categories)
- [x] Fixture loads end-to-end via real `populate_fixture` (8 rows, `enabled=false`, correct periods)

Resolves BA-6949

### DIFF
--- a/fixtures/manager/example-retention-policies.json
+++ b/fixtures/manager/example-retention-policies.json
@@ -1,0 +1,52 @@
+{
+  "retention_policies": [
+    {
+      "id": "b1e5f4a0-0001-4c10-8a01-000000000001",
+      "category": "logs",
+      "retention_period": {"days": 365},
+      "enabled": false
+    },
+    {
+      "id": "b1e5f4a0-0002-4c10-8a01-000000000002",
+      "category": "login",
+      "retention_period": {"days": 365},
+      "enabled": false
+    },
+    {
+      "id": "b1e5f4a0-0003-4c10-8a01-000000000003",
+      "category": "reconcile_history",
+      "retention_period": {"days": 365},
+      "enabled": false
+    },
+    {
+      "id": "b1e5f4a0-0004-4c10-8a01-000000000004",
+      "category": "roles_invitations",
+      "retention_period": {"days": 365},
+      "enabled": false
+    },
+    {
+      "id": "b1e5f4a0-0005-4c10-8a01-000000000005",
+      "category": "deployments",
+      "retention_period": {"days": 365},
+      "enabled": false
+    },
+    {
+      "id": "b1e5f4a0-0006-4c10-8a01-000000000006",
+      "category": "sessions",
+      "retention_period": {"days": 365},
+      "enabled": false
+    },
+    {
+      "id": "b1e5f4a0-0007-4c10-8a01-000000000007",
+      "category": "usage_records",
+      "retention_period": {"days": 90},
+      "enabled": false
+    },
+    {
+      "id": "b1e5f4a0-0008-4c10-8a01-000000000008",
+      "category": "usage_buckets",
+      "retention_period": {"days": 730},
+      "enabled": false
+    }
+  ]
+}
--- a/src/ai/backend/manager/models/alembic/versions/b3e8f1a24c76_seed_retention_policies_disabled.py
+++ b/src/ai/backend/manager/models/alembic/versions/b3e8f1a24c76_seed_retention_policies_disabled.py
@@ -1,0 +1,52 @@
+"""seed retention_policies disabled by default
+
+Make DB record retention opt-in (BEP-1063 follow-up). The initial table
+migration ``4a39641d0fc2`` seeded eight category rows without an explicit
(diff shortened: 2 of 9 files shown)


src/ai/backend/manager/models/alembic/versions/b3e8f1a24c76_seed_retention_policies_disabled.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+
+
+def downgrade() -> None:
+    op.alter_column("retention_policies", "enabled", server_default=sa.true())
+    op.execute(
+        sa.text(
+            "UPDATE retention_policies SET enabled = true WHERE category = ANY(:categories)"
+        ).bindparams(categories=SEED_CATEGORIES)
+    )
```
fregataa: This "UPDATE" seems not necessary in downgrade

HyeockJinKim: Since it didn't have a price anyway, it does seem pointless.

fixtures/manager/example-retention-policies.json
```diff
@@ -0,0 +1,52 @@
+{
+  "retention_policies": [
```
seedspirit, line 2: Shouldn't we also add fixtures to `src/ai/backend/install/fixtures`?

src/ai/backend/manager/models/alembic/versions/b3e8f1a24c76_seed_retention_policies_disabled.py
jopemachine, line 1: It seems that it could be confusing if the enabled value explicitly configured by an administrator later gets changed by a migration.

Since this feature hasn't been released yet, would it make sense to remove the migration script and only set the default value to False?

src/ai/backend/manager/models/alembic/versions/b3e8f1a24c76_seed_retention_policies_disabled.py
HyeockJinKim: Removed in 08d2a4c657 — downgrade now only restores the previous column default and leaves the operator-managed `enabled` values untouched.

fixtures/manager/example-retention-policies.json
HyeockJinKim: Added in 08d2a4c657 — copied the fixture into `src/ai/backend/install/fixtures/` and wired it into the installer's `load_fixtures()`.

src/ai/backend/manager/models/base.py
```diff
@@ -1072,6 +1073,19 @@ async def populate_fixture(
                                 del row[col.name]
                             else:
                                 row[col.name] = None
+                if isinstance(col.type, sa.Interval):
+                    # asyncpg encodes intervals only from datetime.timedelta.
+                    # Accept a kwargs dict (e.g. {"days": 365}) or a number of
+                    # seconds in the fixture JSON.
```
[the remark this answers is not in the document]
HyeockJinKim: Added real-DB `populate_fixture` regression tests in 08d2a4c657 (`tests/unit/manager/models/test_retention_policy_row.py`) covering both the `{"days": 90}` kwargs form and the numeric-seconds form.

src/ai/backend/manager/models/alembic/versions/b3e8f1a24c76_seed_retention_policies_disabled.py
HyeockJinKim: Agreed that migrations must not override operator-configured values — the downgrade UPDATE is removed in 08d2a4c657 for exactly that reason. I kept the upgrade migration though: any DB that already ran `4a39641d0fc2` (including dev/staging setups) has the eight seeded rows with `enabled=true` and the column's `server_default` set to `true` in the DB. Changing only the ORM default would leave that drift in place, and since nobody has explicitly configured these policies yet (the feature is unreleased), flipping the initially-seeded categories to disabled cannot clobber an admin's intent.

fregataa (approved):

PR #13000, opened 2026-07-21
fix(BA-6955): grant kernel scheduling-history read alongside session read

### LINKED ISSUE #12999: Fix kernel scheduling-history read permission migration granting nothing
## Objective

Migration a1c4e7b93f22 (add kernel history read permissions) grants no permission at all, so the scoped kernel scheduling-history endpoint POST /v2/scheduling-history/kernels/scoped/search is registered with auth_required but rejects every non-superadmin caller, including the owner of the kernel.

## Background

The migration mirrors existing grants onto the kernel:history entity, but has two defects:

1. The source predicate selects entity_type = kernel AND operation = read. No migration seeds kernel grants and the runtime never creates them, so the SELECT always returns zero rows. Kernel access is granted through the owning session.
1. The INSERT column list omits permissions.permission, which is NOT NULL. Defect 1 masks defect 2: because no row is ever selected, the NOT NULL violation never surfaces, so the migration passes silently in CI and in every dev environment.

The downgrade is also unscoped: it deletes every kernel:history read row regardless of which migration created it.

## Verification performed

Reproduced on a local stack: a session enqueued as a regular user produced a kernel whose owner received HTTP 403 (lacks permission read on kernel:history). Inserting the mirrored rows from session read grants, including the permission bits, changed the response to HTTP 200.

## Acceptance Criteria

- The upgrade mirrors session read grants and populates the NOT NULL permission column.
- The downgrade removes only the rows the upgrade created, leaving unrelated kernel:history grants intact.
- The upgrade stays idempotent for release-branch backports.
- The kernel owner can call the scoped kernel scheduling-history endpoint and receive HTTP 200.
- The resulting grant scope is reviewed and confirmed to match the intended policy.

## Notes

The migration has not shipped in a release yet, so it is corrected in place rather than superseded by a follow-up revision. Mirroring session read grants also matches domain-scoped session grants, so users other than the owner may gain read access; confirm this matches the intended policy for BA-6881.

JIRA Issue: BA-6955

jopemachine:
Resolves #12999 (BA-6955)

## Summary

`POST /v2/scheduling-history/kernels/scoped/search` (and the `scopedKernelSchedulingHistories` GQL query) is registered with `auth_required` and its action declares `kernel:history` as the entity being read, yet every non-superadmin caller — including the owner of the kernel — was rejected, because **nothing ever granted `kernel:history` read**:

- Migration `a1c4e7b93f22` mirrored grants from `entity_type = 'kernel'`, which nothing creates (kernels are reached through their owning session), so its `SELECT` matched zero rows. The empty result also masked the `INSERT` omitting the `NOT NULL` `permission` column.
- Runtime scope-role creation never included `kernel:history` either, so even a corrected backfill would have left users created afterwards without access.

Per the scope-chain model in `proposals/BEP-1048/permission-check-scope-chain-examples.md` §2, the subject of the check is the entity being read, with grants expected at ancestor scopes — so the fix supplies the grants rather than changing the action's entity type:

| Half | Change | Precedent |
|---|---|---|
| Existing roles | Rewrite the (unreleased) migration to mirror `session` READ grants, carrying the source `permission` bits | `21159a293dfb` and the other migrate-to-rbac backfills |
| New roles | Grant `kernel:history` READ natively in the user owner, project admin, project member, and domain admin scope roles | `*_admin_page` entries in the same `entity_operations()` methods |
| Downgrade | Documented no-op — a mirror-based DELETE cannot tell backfilled rows from natively granted ones | `f2b9a4c7e103` |

Reading a session is what carries kernel visibility (`search_kernel.py` authorizes via `session`), so kernel scheduling history follows the same boundary: it is exactly as visible as the session it belongs to, including domain-scoped `session` read grants.

## Test plan

Verified on a live stack (kernel `f53384a7…` owned by `user@lablup.com`):

- [x] Runtime half: a freshly created user's owner role receives `kernel:history read (permission=1)` natively, with no migration involved
- [x] Backfill half: the upgrade SQL inserted 16 mirrored rows; re-running inserted 0 (idempotent, `ON CONFLICT DO NOTHING`)
- [x] Owner passes the scoped query end-to-end (GQL v2, same adapter → action → RBAC path as REST)
- [x] Negative control: deleting the `kernel:history` rows makes the same owner query fail with `NotEnoughPermission`; restoring them makes it pass — the check is enforced by exactly these rows
- [x] Unit tests added for the four `entity_operations()` changes (`tests/unit/manager/data/{user,group,domain}/test_types.py`)
- [ ] Responses containing actual history rows — kernel scheduling-history rows are never written in this environment (session history recorded 4 rows for the same session while kernel history stayed empty); unrelated to authorization, tracked separately

### DIFF
--- a/changes/13000.fix.md
+++ b/changes/13000.fix.md
@@ -1,0 +1,1 @@
+Grant kernel scheduling-history read permission alongside session read — natively for newly created scope roles and via backfill for existing ones — so the owner of a kernel can query its scheduling history
--- a/src/ai/backend/manager/data/domain/types.py
+++ b/src/ai/backend/manager/data/domain/types.py
@@ -55,6 +55,7 @@ class DomainData:
             for entity in EntityType.admin_accessible_entity_types_in_domain()
         }
         operations[RBACElementType.DOMAIN_ADMIN_PAGE] = {OperationType.READ}
+        operations[RBACElementType.KERNEL_HISTORY] = {OperationType.READ}
         return operations
 
 
--- a/src/ai/backend/manager/data/group/types.py
+++ b/src/ai/backend/manager/data/group/types.py
@@ -72,6 +72,7 @@ class GroupData:
             for entity in EntityType.admin_accessible_entity_types_in_project()
         }
         operations[RBACElementType.PROJECT_ADMIN_PAGE] = {OperationType.READ}
+        operations[RBACElementType.KERNEL_HISTORY] = {OperationType.READ}
         return operations
 
 
@@ -99,10 +100,12 @@ class ProjectMemberRoleSpec:
         return f"project-{str(self.project_id)[:8]}-member"
 
     def entity_operations(self) -> Mapping[RBACElementType, Iterable[OperationType]]:
-        return {
+        operations: dict[RBACElementType, Iterable[OperationType]] = {
             entity.to_element(): OperationType.member_operations()
             for entity in EntityType.member_accessible_entity_types_in_project()
         }
+        operations[RBACElementType.KERNEL_HISTORY] = {OperationType.READ}
+        return operations
 
 
 @dataclass
--- a/src/ai/backend/manager/data/user/types.py
+++ b/src/ai/backend/manager/data/user/types.py
@@ -110,6 +110,7 @@ class UserData:
             entity.to_element(): OperationType.owner_operations()
             for entity in EntityType.owner_accessible_entity_types_in_user()
         }
+        resource_entity_permissions[RBACElementType.KERNEL_HISTORY] = {OperationType.READ}
         user_permissions = OperationType.owner_operations() - {OperationType.CREATE}
         return {RBACElementType.USER: user_permissions, **resource_entity_permissions}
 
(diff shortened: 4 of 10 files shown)


jopemachine:
Reopened with a reworked scope. The earlier close comment claimed the migration approach itself was wrong — re-verification against `BEP-1048/permission-check-scope-chain-examples.md` showed the opposite: the designed model checks the entity being read (`kernel:history`) at ancestor scopes, so the grant-seeding approach was correct and the `EntityType.SESSION` change attempted in #13003 violated the design.

What the earlier attempt was actually missing was the runtime half: scope roles created after the backfill also need the grant, following the `*_admin_page` precedent (`f2b9a4c7e103` + native `entity_operations()` entries). This PR now carries both halves, plus the precedent-mandated no-op downgrade.

jopemachine:
This will be addressed when we introduce virtual scopes and add the permission migration for kernel records. For now, I'll work around the issue by changing the entity type itself to `session`.

jopemachine:
For the record: the design owner confirmed that kernel-entity permission records are intentionally kept empty, so this PR's grant-seeding approach (backfill + native role-creation grants) contradicts the intended data model and stays closed. The kernel-owner 403 is fixed in #12866 by authorizing the scoped search through `session` read (`EntityType.SESSION`), which needs no permission rows.

PR #13003, opened 2026-07-21
fix(BA-6957): authorize kernel scoped scheduling-history via the owning session

jopemachine:
Resolves BA-6957

## Summary

`POST /v2/scheduling-history/kernels/scoped/search` (and the `scopedKernelSchedulingHistories` GQL query) is registered with `auth_required` and documented as running *under a non-admin scope*, but rejected every non-superadmin caller — including the owner of the kernel.

`SearchKernelScopedHistoryAction` declared `EntityType.KERNEL_HISTORY`, and no permission record for kernel entities exists anywhere — **which is intended**: the design owner confirmed kernel-entity permission records are deliberately kept empty. An action gated on them can therefore never pass for a regular user.

Kernel reads are authorized through the owning session, so the session becomes the subject, scope, and target of the RBAC check:

- The adapter (shared by REST v2 and GQL) resolves `kernel_id → session_id` up front through a new `ResolveKernelSessionAction` (repository lookup; `KernelNotFound` on a missing kernel) and dispatches the scoped search with both ids.
- `SearchKernelScopedHistoryAction` takes `kernel_id` and `session_id`: the session drives `entity_type` / `scope_type` / `target_element`, while `kernel_id` bounds the repository search scope (`KernelSchedulingHistorySearchScope`). Whoever may read the session may read its kernels' scheduling history — no permission rows needed, so users created later are covered too.
- A missing kernel now uniformly raises `KernelNotFound` at the resolve step, where it previously surfaced as 403 for regular users and 404 for admins.

Also empties migration `a1c4e7b93f22` into an explicit no-op (`pass` bodies). As shipped it was inert: it mirrored grants from a `kernel` entity that never exists, so its `SELECT` matched nothing — which also masked an `INSERT` omitting the `NOT NULL` `permission` column. The revision is kept only because it has already been applied.

## Test plan

Verified on a live stack with **zero** `kernel:history` permission records (the intended state), against real kernels created through the scheduler:

- [x] Owner's scoped query → **200**
- [x] Control: with `EntityType.KERNEL_HISTORY` the same call → **403** `lacks permission read on kernel:history`
- [x] Kernel narrowing with seeded rows: a kernel's scoped query excludes rows of **sibling kernels in the same session** as well as rows from other sessions; admin unscoped returns all rows
- [x] Another same-domain user → 200, matching `session` read semantics — kernel history is exactly as visible as its owning session
- [x] Missing kernel → `KernelNotFound` for both regular users and admins
- [x] Superadmin unaffected: scoped and unscoped both 200; regular user calling the admin-only unscoped search still → 403
- [x] Unit tests updated: resolve action + scoped action (SESSION subject/scope/target; kernel-bounded repository scope)
- [ ] Responses containing actual scheduler-written history rows — `kernel_scheduling_history` has no writer yet (BA-6852); unrelated to authorization

## Notes

- Sibling scoped-history endpoints (session, deployment, route) are all `superadmin_required`, so they bypass RBAC via the validator's superadmin early-return — the kernel one is the first non-admin history endpoint, which is why this went unnoticed.
- The SDK/CLI surface for these endpoints lands separately in #12866; this PR carries only the authorization change so either can merge first.

### DIFF
--- a/src/ai/backend/manager/repositories/scheduling_history/repository.py
+++ b/src/ai/backend/manager/repositories/scheduling_history/repository.py
@@ -11,6 +11,7 @@ from ai.backend.common.resilience import (
     RetryPolicy,
 )
 from ai.backend.common.resilience.policies.retry import BackoffStrategy
+from ai.backend.common.types import KernelId, SessionId
 from ai.backend.manager.data.deployment.types import (
     DeploymentHistoryListResult,
     RouteHistoryListResult,
@@ -93,6 +94,14 @@ class SchedulingHistoryRepository:
         return await self._db_source.search_kernel_history(querier)
 
     # ========== Kernel History (Scoped) ==========
+
+    @scheduling_history_repository_resilience.apply()
+    async def resolve_session_id(self, kernel_id: KernelId) -> SessionId:
+        """Return the id of the session owning ``kernel_id``.
+
+        Raises ``KernelNotFound`` when no such kernel exists.
+        """
+        return await self._db_source.resolve_session_id(kernel_id)
 
     @scheduling_history_repository_resilience.apply()
     async def search_kernel_scoped_history(
--- a/src/ai/backend/manager/services/scheduling_history/actions/search_kernel_scoped_history.py
+++ b/src/ai/backend/manager/services/scheduling_history/actions/search_kernel_scoped_history.py
@@ -4,7 +4,7 @@ from dataclasses import dataclass
 from typing import override
 
 from ai.backend.common.data.permission.types import EntityType, RBACElementType, ScopeType
-from ai.backend.common.types import KernelId
+from ai.backend.common.types import KernelId, SessionId
 from ai.backend.manager.actions.action.scope import BaseScopeAction, BaseScopeActionResult
 from ai.backend.manager.actions.types import ActionOperationType
 from ai.backend.manager.data.kernel.types import KernelSchedulingHistoryData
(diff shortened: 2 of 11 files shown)


jopemachine:
Closing — re-verification showed this change violates the designed authorization model. `BEP-1048/permission-check-scope-chain-examples.md` §2 specifies that the subject of a scope-chain check is the entity being read (`entity_type=kernel, op=read → MATCH ... if role has kernel:read at project scope`), so the scoped history action's `EntityType.KERNEL_HISTORY` declaration was correct all along; the missing piece was the grants. The complete fix (backfill migration + native grants in scope-role creation, per the `*_admin_page` precedent) lives in the reopened #13000.

jopemachine:
Reopened: the design owner confirmed kernel-entity permission records are intentionally kept empty, so this session-based authorization is the correct approach after all — the earlier close comment claiming it violated the design model was wrong. Rebased onto current main and reworded with the confirmed rationale; carried as its own PR instead of inside #12866.

src/ai/backend/manager/services/scheduling_history/actions/search_kernel_scoped_history.py
```diff
@@ -28,7 +27,7 @@ class SearchKernelScopedHistoryAction(BaseScopeAction):
     @override
     @classmethod
     def entity_type(cls) -> EntityType:
-        return EntityType.KERNEL_HISTORY
+        return EntityType.SESSION
```
HyeockJinKim, line 33: I don't think this will work. Although I have permissions for the session ID, I don't have any information about the session associated with the kernel ID, so if I'm going to set it temporarily, I'll probably have to run a session ID query to check.

HyeockJinKim (changes requested):

src/ai/backend/manager/repositories/scheduling_history/repository.py
```diff
@@ -94,6 +95,14 @@ async def search_kernel_history(
 
     # ========== Kernel History (Scoped) ==========
 
+    @scheduling_history_repository_resilience.apply()
+    async def kernel_session_id(self, kernel_id: KernelId) -> SessionId:
```
jopemachine: Rename this to `resolve_session_id`

HyeockJinKim (approved):

Issue #12929, opened 2026-07-16
[AppConfig] Fragment search REST v2 API (admin + scoped)


jopemachine:
## Objective

Expose the app_config_fragment search surface as REST v2, split out of the fragment write PR (BA-6921, #12928) to keep each PR review-sized. Two variants: system-wide admin search (superadmin) and principal-scoped search (auth).

## Background

BEP-1052. Stacks on BA-6921 (fragment write CRUD) and reuses its AppConfigFragmentNode DTO. The scoped search is the REST surface of the visible-fragments query (BA-6810, #12706).

## Scope

- Filter / order / scope DTOs (AppConfigFragmentFilter, AppConfigFragmentOrder, SearchAppConfigFragmentInput, AppConfigFragmentScope, ScopedSearchAppConfigFragmentInput) + types (order field, scope-type filter)
- adapter admin_search / scoped_search + filter/order converters
- REST routes: POST /search (superadmin), POST /scoped-search (auth)

## Acceptance Criteria

- /search superadmin-only (system-wide); /scoped-search auth-required (principal-visible)
- Both reflected in the OpenAPI dump; pants check / lint pass
- Verified against the live server (admin + non-admin)

## Related

- Split from BA-6921 (#12928). BEP-1052. Scoped search backs BA-6810 (#12706).

JIRA Issue: BA-6922

PR #12928, opened 2026-07-16
feat(BA-6921): AppConfig fragment write REST v2 API (CRUD)

jopemachine:
## 📚 Stacked PRs

Part of the AppConfigFragment / AppConfig stack under BEP-1052 (epic BA-5781). **Merge in order (bottom-up):**

1. ✅ #12306 — `feat(BA-6552)`: app_config_fragments DB model and Alembic migration
2. ✅ #12307 — `feat(BA-6553)`: repository layer
3. ✅ #12403 — `refactor(BA-6619)`: consolidate AppConfigScopeType into common.data
4. ✅ #12405 — `refactor(BA-6620)`: ExistsQuerier + AppConfigAllowList.exists
5. ✅ #12358 — `feat(BA-6554)`: AppConfigFragment service layer
6. ✅ #12516 — `feat(BA-6702)`: move fragment rank to the allow list with scope defaults
7. ✅ #12517 — `feat(BA-6701)`: expose allow-list rank on the v2 API surface
8. ✅ #12518 — `feat(BA-6704)`: cascade app config subtree deletion from the definition
9. ✅ #12426 — `feat(BA-6626)`: app_config_fragment bulk repository layer
10. ✅ #12401 — `feat(BA-6618)`: app_config_fragment bulk CRUD service layer
11. ✅ #12706 — `feat(BA-6810)`: AppConfigFragment visible-fragments query (repository layer)
12. ✅ #12826 — `feat(BA-6859)`: bind fragments to their RBAC scope on write (repository layer)
13. ✅ #12837 — `fix(BA-6872)`: seed `APP_CONFIG_FRAGMENT` permissions into the RBAC role fixture
14. ✅ #12839 — `chore(BA-6873)`: drop the dead APP_CONFIG RBAC entity type
15. ✅ #12359 — `feat(BA-6555)`: AppConfig merge engine + resolve service (service layer)
16. ✅ #12984 — `feat(BA-6948)`: store `app_config_fragments.scope_id` as a nullable UUID
17. 👉 **#12928 — `feat(BA-6921)`: AppConfig fragment write REST v2 API (CRUD)** ← you are here
18. #13020 — `feat(BA-6922)`: AppConfig fragment search REST v2 API (admin + scoped)
19. #12377 — `feat(BA-6556)`: merged AppConfig REST v2 read API

> **Fork at #12928**: the fragment search (#13020) and the merged-read (#12377) both stack on #12928 and can merge in either order once it lands.

## Summary

[...]

PR #13024, opened 2026-07-22
refactor(BA-6922): model fragment scoped search as a scope action

jopemachine:
## 📚 Stacked PRs

- #13020 — `BA-6922` feat: AppConfig fragment search REST v2 API (admin + scoped)
- this PR — `BA-6922` refactor: model the fragment scoped search as a scope action  ← you are here

Related: #12929 (BA-6922), resolved by #13020.

## Summary

- `ScopedSearchAppConfigFragmentAction` was a `BaseBulkAction[SearchableActionTarget]` holding a **list** of domain/user targets OR-combined into one query. It is now a `BaseScopeAction` acting at the single scope the request names, deriving its RBAC facets through `AppConfigScopeType.to_rbac_*` exactly as the sibling `CreateAppConfigFragmentAction` does.
- Wire it through `ScopeActionProcessor` with `validators.rbac.scope`. As a bulk action it reached the processor with **no RBAC validator at all**, so any authenticated caller could read any domain's or user's fragments by naming them in the request body.
- Repository: `DomainAppConfigFragmentSearchScope` / `UserAppConfigFragmentSearchScope` collapse into one `AppConfigFragmentSearchScope(scope_type, scope_id)` — the fragment row is `(scope_type, scope_id)`, so one scope expresses the query exactly. `scoped_search(querier, scope)` takes a single scope.
- DTO: `AppConfigFragmentScope` becomes `scope_type: AppConfigScopeType` + `scope_id: AppConfigScopeID`, mirroring `CreateAppConfigFragmentInput`. The route stays `POST /v2/app-config-fragments/scoped/search`.

**Why:** a search is not a bulk mutation. `BaseBulkAction` exists to authorize an operation per entity across a batch (`bulk_update` / `bulk_purge` over fragment ids); a scoped search acts at one RBAC scope and returns whatever lives under it.

**Tradeoff:** searching several scopes in one request is gone — that was the only thing the list bought, and it cannot be authorized as a single scope action.

**Open item, unchanged:** the DTO rejects `scope_type: public`. Public maps to the `GLOBAL` RBAC scope, which has no scope element to authorize against, so whether a scoped search should reach public fragments still needs its own decision. The repository scope models it (`scope_id=None`), so lifting the restriction is a validator change.

## Test plan

- [x] `pants fix fmt lint check` over the six touched packages — ruff, visibility, mypy all clean
- [x] `openapi.json` regenerated with `./backend.ai mgr api dump-openapi`; final delta vs #13020 is `/scoped-search` → `/scoped/search` plus the reshaped `AppConfigFragmentScope`
- [x] repository scoped-search tests parametrized over the scope (domain / user / public / unknown owner); the OR-across-scopes case dropped with the capability
- [x] `queried_refs` / `element_refs()` removal verified to have no consumer anywhere in the tree
- [ ] CI test run
- [ ] live-server verification of `/scoped/search` for admin and non-admin


jopemachine:
Superseded by #13032, which puts the same refactor on `main`. The manager-side scoped search (action, service, processor, repository scope) already lives on `main` via #12928, so this never needed to stack on #13020 — only the REST surface does. #13020 will be updated to build on the shape #13032 lands.

PR #13032, opened 2026-07-22
refactor(BA-6922): model the fragment scoped search as a scope action

jopemachine:
Related: #12929 (BA-6922)

## Summary

- `ScopedSearchAppConfigFragmentAction` was a `BaseBulkAction[SearchableActionTarget]` holding a **list** of domain/user targets that the repository OR-combined into one query. It is now a `BaseScopeAction` acting at the single scope the request names, deriving its RBAC facets through `AppConfigScopeType.to_rbac_*` exactly as the sibling `CreateAppConfigFragmentAction` does for a write.
- Wire it through `ScopeActionProcessor` with `validators.rbac.scope`. As a bulk action it reached the processor with **no RBAC validator at all** — `bulk_update` / `bulk_purge` pass `validators.rbac.bulk`, `scoped_search` passed nothing — so once an API reaches it, any authenticated caller could read any domain's or user's fragments by naming them in the request.
- The fragment row is `(scope_type, scope_id)`, so one scope expresses the query exactly: `DomainAppConfigFragmentSearchScope` / `UserAppConfigFragmentSearchScope` collapse into `AppConfigFragmentSearchScope` carrying `scope_type` + `scope_id`.
- `AppConfigFragmentSearchScope.existence_checks` no longer returns `()`. A search at a domain or user that does not exist came back as an empty page, making "no fragments here" indistinguishable from "no such domain". It now checks the scope owner the way every other scope does, so a missing owner is a 404. `public` keeps an empty check — the global scope has no owner row.

**Why not a bulk action:** `BaseBulkAction` exists to authorize an operation per entity across a batch. A scoped search has no per-entity target — it acts at one scope and returns whatever lives under it.

**Repository arity is unchanged.** `scoped_search(querier, scopes: Sequence[SearchScope])` still takes a sequence, mirroring `batch_query_with_scopes` / `BatchQuerier.scopes` underneath it and matching `audit_log`. The single-scope action wraps its scope at the call site; the read path is not pinned to one scope by its only caller.

**Tradeoff:** searching several scopes in one request is gone at the *action* level — it cannot be authorized as a single scope action. Nothing consumed the removed `queried_refs` / `element_refs()` on the result.

**Scope of this PR:** manager-side only (repository scope, action, service, processor). No REST or GraphQL surface reaches the scoped search on `main` yet, so this is self-contained. #13020 adds that surface and will build on this shape.

## Review notes

- The existence checks pull `DomainRow`, `UserRow` and their resource-policy / keypair dependencies into the repository test's `with_tables` set, plus a `scope_owners` fixture. That weight is the reason `audit_log` leaves its checks empty; the judgement here is that a superadmin typo'ing a scope id deserves a 404, since `ScopeActionRBACValidator` returns early for superadmins and when RBAC enforcement is disabled.
- Checks are aggregated across scopes and validated in a single query by `execute_batch_querier`, so this adds no round trip.

## Test plan

- [x] Formatter, linter, visibility and mypy clean over the touched service, repository and test packages
- [x] Repository scoped-search cases parametrized over the scope (domain / user / public), plus OR-combined scopes and a parametrized missing-owner case asserting `DomainNotFound` / `UserNotFound`
- [x] Service scoped-search cases parametrized over the scope, asserting the scope is passed through and the result reports the RBAC scope it was authorized at
- [x] `queried_refs` / `element_refs()` removal verified to have no consumer anywhere in the tree
- [ ] CI verification


changes/13032.enhance.md
jopemachine, line 1: enhance

src/ai/backend/manager/services/app_config_fragment/processors.py
```diff
@@ -84,9 +84,12 @@ def __init__(
             service.get, action_monitors, validators=[validators.rbac.single_entity]
         )
         # admin_search is system-wide (all scopes) — gated by superadmin_required at the API
-        # layer, so it carries no per-scope RBAC validator. Non-admins use scoped_search.
+        # layer, so it carries no per-scope RBAC validator. Non-admins use scoped_search,
+        # which is validated at the scope it names.
         self.admin_search = ScopeActionProcessor(service.admin_search, action_monitors)
```
jopemachine: GlobalActionProcessor

fregataa (approved):

PR #13055, opened 2026-07-23
feat(BA-6984): expose AppConfigFragment writes over GraphQL v2 (mutations)

jopemachine:
## 📚 Stacked PRs

- #13047 — `BA-6977` DTO + adapter + GraphQL: fragment read surface and batch load
- #13055 — `BA-6984` GraphQL: fragment write mutations  ← you are here

Resolves BA-6984

## Summary

The write half of the AppConfigFragment GraphQL surface, stacked on the read half (#13047). Writing a fragment is the same ordinary user work reading is — a user writes their own user-scope fragment, a domain admin their domain's — so a GraphQL client should not have to fall back to REST to change one.

| mutation | auth |
|---|---|
| `createAppConfigFragment(input:)` | auth, RBAC |
| `updateAppConfigFragment(id:, input:)` | auth, RBAC |
| `purgeAppConfigFragment(input:)` | auth, RBAC |
| `bulkUpdateAppConfigFragments(input:)` | auth, RBAC |
| `bulkPurgeAppConfigFragments(input:)` | auth, RBAC |

No resolver here calls `check_admin_only()` — authorization stays with the scope / single-entity / bulk RBAC validators the processors already run. The bulk mutations report per-item failures alongside successes, matching the REST payloads.

The purge takes an input rather than a bare `id` argument: an argument has nowhere to grow, so a later purge option would have to arrive as a second one. `PurgeAppConfigFragmentInput` is a new DTO used only at the GraphQL boundary — REST keeps naming the target in the path, and the adapter signature is unchanged.

`updateAppConfigFragment` still takes `id` alongside its input, matching the REST path + body split; `UpdateAppConfigFragmentInput` states that the target is named outside the body. Worth settling in review whether the id should move inside, which would mean splitting that DTO from the REST one.

## What's included

- **DTO**: `PurgeAppConfigFragmentInput` in `common/dto/manager/v2/app_config_fragment/request.py`.
- **GQL types**: create / update / purge / bulk-update / bulk-purge inputs, the bulk update item input, and the create / update / purge / bulk payloads including the per-item error type.
- **Resolvers**: the five mutations above.
- **Registration**: the fields on `Mutation` in `api/gql/schema.py`.
- **Regenerated** `docs/manager/graphql-reference/{v2-schema,supergraph}.graphql`.

## Test plan

- [x] `scripts/generate-graphql-schema.sh` regenerates cleanly; the five mutations land in `v2-schema.graphql` / `supergraph.graphql`.
- [ ] Live server, non-admin: create, update and purge a fragment at their own user scope; rejected at another principal's scope.
- [ ] Live server: a bulk update mixing a writable and a non-writable fragment reports the failure per item.

----
📚 Documentation preview 📚: https://sorna--13055.org.readthedocs.build/en/13055/

----
📚 Documentation preview 📚: https://sorna-ko--13055.org.readthedocs.build/ko/13055/


jopemachine:
I’ll address this in a separate PR stack based on #13113.

PR #13068, opened 2026-07-23
feat(BA-6974): add initial grace period source

seedspirit:
## Summary
- Add checker-owned `initial_grace_seconds` to each idle checker spec.
- Add the initial grace Source and repository query for `RUNNING` sessions with `NOT_CHECKED` rows.
- Return the checker grace value, grace start timestamp, and database current time without performing state transitions.

## Test plan
- [x] `pants fmt ::`
- [x] `pants fix ::`
- [x] Push-hook type checks and affected tests

Resolves BA-6974


src/ai/backend/common/data/idle_checker/types.py
```diff
@@ -58,6 +58,11 @@ class IdleCheckerSpec(BackendAISchema):
     """
 
     type: CheckerType = Field(description="Idle checker kind; selects the sub-config.")
+    initial_grace_period_seconds: int = Field(
+        default=0,
+        ge=0,
+        description="Delay before the first idle-check judgment.",
+    )
```
HyeockJinKim: Wouldn't it be better to add a column rather than storing it in JSONB?

src/ai/backend/manager/models/alembic/versions/39fadbbea696_add_initial_grace_period_to_idle_.py
```diff
@@ -0,0 +1,51 @@
...
+        ),
+    )
+    op.execute(
+        sa.text(
+            """
+            UPDATE idle_checkers
+            SET initial_grace_period_seconds = 0
+            """
+        )
+    )
```
HyeockJinKim: Is there a difference between setting the "initial grace period" to 0 seconds and setting it to None?

HyeockJinKim: Please enter a default value.

HyeockJinKim (approved):

PR #13071, opened 2026-07-23
feat(BA-6989): reshape KernelScope into scope-item lists and accept a session scope

jopemachine:
## Summary

Reshape the scoped kernel scheduling-history scope input from a bare `kernel_id: UUID` into category-keyed lists of scope items, mirroring `AuditLogScope`, and add a **session** category so the history of every kernel a session owns can be queried in one call.

| Layer | Before | After |
| --- | --- | --- |
| `KernelHistoryScopeDTO` | `kernel_id: UUID` | `kernel: list[UUIDScope] \| None`, `session: list[UUIDScope] \| None` |
| `KernelScope` (GraphQL) | `kernelId: UUID!` | `kernel: [UUIDScope!]`, `session: [UUIDScope!]` |
| `ScopedSearchKernelHistoriesInput` (REST v2) | `{"scope": {"kernel_id": "..."}}` | `{"scope": {"kernel": [{"value": "..."}]}}` or `{"scope": {"session": [{"value": "..."}]}}` |
| `SearchKernelScopedHistoryAction` | `kernel_id` + `session_id` | one `KernelHistoryTarget` |
| `SchedulingHistoryRepository.search_kernel_scoped_history` | `scope: KernelSchedulingHistorySearchScope` | `scopes: Sequence[SearchScope]` |

`KernelScope` is unreleased (`added_version` is the next release), so widening the field is not a breaking change.

## Scope items are `SearchableActionTarget`s

Each scope category maps to its own target, the same shape the audit-log scoped search uses (`EntityAuditLogTarget` / `TriggeredByAuditLogTarget`). A target derives both the row filter and its own RBAC element ref, so adding a category is adding a class — no branching in the service:

```
KernelHistoryTarget(SearchableActionTarget)
├── KernelKernelHistoryTarget(kernel_id)   → KernelKernelHistorySearchScope,   RBACElementRef(KERNEL, …)
└── SessionKernelHistoryTarget(session_id) → SessionKernelHistorySearchScope,  RBACElementRef(SESSION, …)
```

The repository takes `Sequence[SearchScope]` — the db source already wrapped the single scope into a one-element list for `execute_batch_querier` — so OR-ing several scope items later needs no signature change.

## What is temporary, and marked as such

Two current-implementation constraints are carried explicitly with `TODO`s rather than folded into the design:

- **Kernels hold no RBAC permission records.** `KernelKernelHistoryTarget` is therefore not dispatchable yet: the adapter converts a kernel scope item into a `SessionKernelHistoryTarget` on the owning session (via `ResolveKernelSessionAction`) and narrows the rows back down with a `kernel_id` query condition. Rows returned are identical to before. The kernel target and its search scope stay in the tree as the shape to switch to once virtual scopes land.
- **`SearchKernelScopedHistoryAction` is a single-target `BaseScopeAction`.** The scope input is already list-shaped and its items are meant to be OR'd, so the adapter rejects a scope carrying two or more items; both the rejection and the single `target` field go away when this becomes a bulk action.

Empty scopes are rejected by the DTO validator.

## Test plan
- [x] Service-layer unit test covers the session-scoped search (target → search scope → repository call, RBAC scope/target/entity type)
- [x] GraphQL v2 schema, supergraph, and OpenAPI regeneration succeed and show the new input shape
- [x] CI lint / typecheck / unit / component / integration green

### DIFF
--- a/src/ai/backend/manager/api/adapters/scheduling_history/adapter.py
+++ b/src/ai/backend/manager/api/adapters/scheduling_history/adapter.py
@@ -456,6 +456,10 @@ class SchedulingHistoryAdapter(BaseAdapter):
         )
         kernel_items = input.scope.kernel or []
         session_items = input.scope.session or []
+        # TODO: Drop this rejection once the scoped search becomes a bulk action.
+        # The scope input is already list-shaped and its items are meant to be
+        # OR'd, but SearchKernelScopedHistoryAction is a single-target
+        # BaseScopeAction, so only one item is dispatchable today.
         if len(kernel_items) + len(session_items) != 1:
             raise InvalidAPIParameters(
                 "Kernel scheduling history scope accepts exactly one scope item"
--- a/src/ai/backend/manager/api/adapters/scheduling_history/adapter.py
+++ b/src/ai/backend/manager/api/adapters/scheduling_history/adapter.py
@@ -30,7 +30,10 @@ from ai.backend.common.dto.manager.v2.scheduling_history.response import (
     SearchKernelHistoriesPayload,
     SessionHistoryNode,
 )
-from ai.backend.common.dto.manager.v2.scheduling_history.types import SubStepResultInfo
+from ai.backend.common.dto.manager.v2.scheduling_history.types import (
+    KernelHistoryScopeDTO,
+    SubStepResultInfo,
+)
 from ai.backend.common.identifier.kernel_scheduling_history import KernelSchedulingHistoryID
 from ai.backend.common.identifier.replica import ReplicaID
 from ai.backend.common.types import KernelId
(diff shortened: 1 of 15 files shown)


src/ai/backend/manager/api/adapters/scheduling_history/adapter.py
```diff
@@ -439,6 +442,35 @@ async def scoped_search_kernel_history(
...
+                )
+            )
+            session_id = resolve_result.session_id
+            conditions.append(
+                KernelSchedulingHistoryConditions.by_kernel_id_filter(
+                    UUIDEqualMatchSpec(value=kernel_id, negated=False)
+                )
+            )
+        else:
+            session_id = SessionId(session_items[0].value)
```
HyeockJinKim, line 473: This is a little bothersome, so I'll get it taken care of right away.

HyeockJinKim (approved):

PR #13078, opened 2026-07-23
feat(BA-6876): add virtual-scope-chain RBAC action validators

fregataa:
## Summary
- Implement the three RBAC action validators (scope / single-entity / bulk) on the pure-ABC action framework (`actions/{scope,single_entity,bulk}`), resolving effective permissions through the virtual-scope chain (`entity → entity_memberships → scope_bindings → scope`) instead of the recursive direct scope-walk.
- Add `subject_entity_type` to `VirtualScopePermissionCheckKey` so scope-action checks walk the scope-as-entity chain while matching permission rows on the acted-on entity type.
- Expose `check_permission_via_virtual_scope` / `check_bulk_permission_via_virtual_scope` on `PermissionControllerRepository` (previously db_source-only).

Cutover wiring behind a `rbac.resolution_backend` config is out of scope (BA-6875).

## Test plan
- [x] `pants test tests/unit/manager/actions/validators::` — permitted / denied / permission_cap clipping for all three validators, superadmin and enforcement-off short-circuits, multi-scope and bulk all-or-nothing semantics
- [x] `pants test tests/unit/manager/repositories/permission_controller::` — existing VS-chain checks stay green with the extended key

Resolves BA-6876

### DIFF
--- a/src/ai/backend/manager/actions/scope/validator/rbac.py
+++ b/src/ai/backend/manager/actions/scope/validator/rbac.py
@@ -19,7 +19,7 @@ class VirtualScopeScopeActionRBACValidator(ScopeActionValidator):
 
     Each target scope is checked as an entity (reachable through its own and its
     ancestors' virtual scopes), while permission rows are matched on the
-    acted-on entity type (``subject_entity_type``). Every target scope must be
+    acted-on entity type (``entity_type``). Every target scope must be
     authorized for the action to pass.
     """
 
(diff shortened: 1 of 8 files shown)


src/ai/backend/manager/actions/scope/validator/rbac.py
```diff
@@ -0,0 +1,69 @@
...
+
+        keys = [
+            VirtualScopePermissionCheckKey(
+                user_id=UserID(user.user_id),
+                entity=EntityRef(
+                    entity_type=EntityType(scope.scope_type),
+                    entity_id=scope.scope_id,
+                ),
+                subject_entity_type=action.entity_type(),
+            )
```
HyeockJinKim: I think it would be better to use a different name in `scope`. Also, this isn't a bulk permission check, is it?

fregataa: It is bulk permission check

HyeockJinKim (approved):

PR #13081, opened 2026-07-23
fix(BA-6994): split reserved start time from execution start time on sessions

HyeockJinKim:
## Summary
- Split the double meaning of `sessions.starts_at` (reserved start time at enqueue vs. actual execution start at the RUNNING transition) into a new `requested_starts_at` column, so the reservation is no longer lost when a batch session starts running.
- The reserved-batch validator now reads `requested_starts_at` and compares it against a DB-sourced snapshot timestamp (`SystemSnapshot.observed_at`) instead of per-server clocks; relative `starts_at` offsets are also resolved against DB time at enqueue (HA clock-skew safety).
- Alembic migration adds the column and backfills batch sessions that have not reached RUNNING yet (already-running/terminated sessions have lost the reservation, left NULL).

## Test plan
- [x] Reserved-batch validator hold/release scenarios with a fixed snapshot time, including the exact-boundary case
- [x] RUNNING transition writes `starts_at` without touching `requested_starts_at` (real-DB test)
- [x] Migration upgrade/downgrade verified against a scratch Postgres DB (9 cases across statuses/types)
- [x] Existing idle session-lifetime / fair-share tests unchanged and green

Resolves BA-6994

### DIFF
--- a/src/ai/backend/manager/repositories/scheduler/creators.py
+++ b/src/ai/backend/manager/repositories/scheduler/creators.py
@@ -133,7 +133,7 @@ class SessionRowFromSpec(CreatorSpec[SessionRow]):
     """Build one :class:`SessionRow` from a :class:`SessionSpec`.
 
     Session-level snapshot fields (``vfolder_mounts``, ``environ``,
-    ``startup_command``, ``bootstrap_script``, ``starts_at``,
+    ``startup_command``, ``bootstrap_script``, ``requested_starts_at``,
     ``batch_timeout``) are sourced from the main kernel's execution
     spec. ``images`` / ``image_ids`` are deduplicated from each
     kernel's resolved :class:`ImageInfo`, with the main kernel's image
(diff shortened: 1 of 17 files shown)


jopemachine:
News fragment is missing

src/ai/backend/manager/repositories/scheduler/creators.py
```diff
@@ -169,7 +169,7 @@ def build_row(self) -> SessionRow:
             )
 
         main_mounts = list(main_kernel.vfolder_mounts) if main_kernel else []
-        session_starts_at = main_kernel.execution_spec.starts_at if main_kernel else None
+        session_requested_starts_at = main_kernel.execution_spec.starts_at if main_kernel else None
```
seedspirit, line 172: Should this part also be changed to "session request starts at"? I was just wondering if "starts at" might be more accurate.

HyeockJinKim: The "starts at" value combines two different concepts—the time the current session is running and the time the batch session is scheduled to run—which causes confusion between the concepts of scheduling and execution time.
Therefore, for this part, we treat the value specified in `enqueue` as pertaining to the batch session and proceed accordingly.

seedspirit (approved):

PR #13082, opened 2026-07-23
feat(BA-6996): accept per-request agent selection policy in session enqueue

HyeockJinKim:
## Summary
- Expose an optional `agent_selection_policy` (`strict`/`preferred`) on `EnqueueSessionInput` for both REST v2 and GraphQL; an unset value keeps inheriting the resource group's `default_session_options.agent_selection_policy` through the existing merge rule.
- Thread the value through the session adapter and `SessionSchedulingSpec` into `SchedulingTargetDraft`, where the already-merged downstream path (merge rule, persistence, agent selector) consumes it. SDK v2 and the CLI enqueue command pass the DTO through unchanged, so the field is usable end-to-end without client code changes.
- Move `MountItemInput` to `dto/manager/v2/common.py` to break the circular import between the `session` and `session_options` request modules.

## Test plan
- [x] `test_enqueue_session_from_draft.py`: request-level STRICT overrides the RG default (PREFERRED) through the real preparer chain; unset inherits the RG default (STRICT).
- [x] `test_session_adapter.py`: DTO enum converts to the domain enum on the action; omitted stays `None`.
- [x] `test_session_v2_dto.py`: parses `strict`/`preferred`, defaults to `None`, rejects invalid values.
- [x] `pants fmt/fix/lint/check/test` pass for all affected targets (including direct dependents).

Resolves BA-6996

----
📚 Documentation preview 📚: https://sorna--13082.org.readthedocs.build/en/13082/

----
📚 Documentation preview 📚: https://sorna-ko--13082.org.readthedocs.build/ko/13082/

### DIFF
--- a/docs/manager/graphql-reference/v2-schema.graphql
+++ b/docs/manager/graphql-reference/v2-schema.graphql
@@ -4795,6 +4795,11 @@ input EnqueueSessionInput {
   """Designated agent IDs."""
   agentList: [String!] = null
 
+  """
+  Added in 26.8.0. How agent_list is enforced. null inherits the resource group default.
+  """
+  agentSelectionPolicy: SessionAgentSelectionPolicy = null
+
   """Network UUID to attach."""
   attachNetwork: ID = null
 
--- a/src/ai/backend/common/dto/manager/v2/common.py
+++ b/src/ai/backend/common/dto/manager/v2/common.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from decimal import Decimal
 from enum import StrEnum
 from functools import cached_property
+from uuid import UUID
 
 from pydantic import Field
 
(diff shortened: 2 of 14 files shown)


src/ai/backend/common/dto/manager/v2/common.py
```diff
@@ -123,6 +124,31 @@ class VFolderHostPermissionEntryInput(BaseRequestModel):
...
 
+class MountItemInput(BaseRequestModel):
+    """A single virtual folder mount specification.
+
+    Shared by the session enqueue input and the session-options kernel
+    execution spec (lives here to avoid a session <-> session_options
+    circular import).
+    """
+
+    vfolder_id: UUID = Field(description="Virtual folder UUID to mount.")
```
jopemachine, line 135: Nit: Let's use VFolderUUID

docs/manager/graphql-reference/v2-schema.graphql
```diff
@@ -4795,6 +4795,11 @@ input EnqueueSessionInput {
   """Designated agent IDs."""
   agentList: [String!] = null
 
+  """
+  Added in 26.8.0. How agent_list is enforced. null inherits the resource group default.
+  """
+  agentSelectionPolicy: SessionAgentSelectionPolicy = null
```
jopemachine, line 4801: I think it would be helpful if the schema description included a concrete explanation of how overrides behave in both PREFERRED and STRICT modes.

HyeockJinKim: I'll put this under "Follow-up Issues."

jopemachine (approved):

PR #13086, opened 2026-07-23
fix(BA-6995): reconcile stale uid/gid entries baked into committed images

fregataa:
## Summary
- The kernel container entrypoint silently failed on `groupadd`/`useradd` name collisions when a session was started from a committed (customized) image whose `/etc/passwd`/`/etc/group` already contained a `work` user/group created by a previous session with different ids. As a result, the requested `container_uid`/`container_main_gid` were not reflected in the container's user DB (e.g., `id work` kept showing the old gid), and a changed uid could break keyfile-based SSH logins against dropbear's owner checks.
- Renumber the existing `work` user/group with `usermod -u`/`groupmod -g` on name collisions instead of failing, and log a warning when reconciliation fails instead of ignoring errors.
- Skip the `usermod -l work` self-rename for already-normalized images (renaming a user to its own name makes the whole `usermod` fail, which also dropped the `-g` update), and re-apply the primary group instead.
- Always add the user to `ADDITIONAL_GIDS` groups even when the group pre-exists in the image, and guard the `grpread` group creation with an existence check.

## Test plan
- [x] Start a session with a fresh (non-committed) image and verify `id work` shows the configured `container_uid`/`container_main_gid` (no behavior change).
- [x] Commit a session image, change the user's `container_main_gid`, start a new session from the committed image, and verify `id work` reflects the new gid.
- [x] Commit a session image, change the user's `container_uid`, start a new session from the committed image, and verify `id work` reflects the new uid and keyfile SSH login via `id_container` succeeds.
- [x] Verify supplementary groups from `ADDITIONAL_GIDS` are applied to the `work` user on a committed image.

Resolves BA-6995

### DIFF
--- a/src/ai/backend/runner/entrypoint.sh
+++ b/src/ai/backend/runner/entrypoint.sh
@@ -155,6 +155,10 @@ else
     fi
     export SHELL=/bin/bash
   fi
+  if [ "$(getent passwd $USER_NAME | cut -d: -f3-4)" != "$USER_ID:$GROUP_ID" ]; then
+    echo "ERROR: /etc/passwd entry for '$USER_NAME' does not match $USER_ID:$GROUP_ID"
+    exit 1
+  fi
   export LD_LIBRARY_PATH="/opt/backend.ai/lib:$LD_LIBRARY_PATH"
   export HOME="/home/$USER_NAME"
 
--- a/src/ai/backend/runner/entrypoint.sh
+++ b/src/ai/backend/runner/entrypoint.sh
@@ -74,17 +74,40 @@ else
   echo "Setting up uid and gid: $USER_ID:$GROUP_ID"
   USER_NAME=$(getent passwd $USER_ID | cut -d: -f1)
   GROUP_NAME=$(getent group $GROUP_ID | cut -d: -f1)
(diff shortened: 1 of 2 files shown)


src/ai/backend/runner/entrypoint.sh
```diff
@@ -74,43 +74,83 @@ else
...
+  }
   if [ -f /bin/ash ]; then  # for alpine (busybox)
     if [ -z "$GROUP_NAME" ]; then
       GROUP_NAME=work
-      addgroup -g $GROUP_ID $GROUP_NAME
+      if getent group $GROUP_NAME > /dev/null 2>&1; then
+        renumber_group $GROUP_NAME $GROUP_ID
+      else
+        addgroup -g $GROUP_ID $GROUP_NAME \
+          || echo "WARNING: failed to create the group '$GROUP_NAME' with gid $GROUP_ID"
```
seedspirit, line 92: Is displaying a warning log enough? (Is it okay not to make failure?)

fregataa: `addgroup` sets values in `/etc/group` file which is used to map names and this file is not used for auth. I added "exit 1" when `/etc/passwd` is not updated properly

seedspirit (approved):

PR #13095, opened 2026-07-24
fix(BA-6986): harden keypair webapp auth (sToken + login verification)

seedspirit:
## Summary
- Remove `secret_key` from the sToken JWT payload — it was exposed via the redirect URL (a JWT payload is signed, not encrypted).
- Add `iat`/`exp` (60s TTL) with 10s clock-skew leeway so sTokens actually expire.
- Verify `secret_key` against the stored keypair (constant-time) at `/login`, closing an auth bypass where knowing an access key alone was enough to log in as its owner. Failures return a uniform 401 to avoid access-key enumeration.
- Mask bearer tokens in error logs.

## Test plan
- [x] `pants test tests/unit/manager/plugin/test_keypair_utils.py` — 5 sToken lifecycle regression tests (roundtrip, no secret_key in payload, iat/exp present, expiry, wrong secret)
- [x] `pants lint check` on changed files
- [ ] Manual: `/login` with correct vs incorrect secret_key, and an expired sToken

Resolves BA-6986

### DIFF
--- a/src/ai/backend/manager/plugin/keypair/webapp.py
+++ b/src/ai/backend/manager/plugin/keypair/webapp.py
@@ -1,9 +1,11 @@
 import json
 import logging
+import secrets
 from collections.abc import Mapping, Sequence
-from typing import Any, override
+from typing import Any, cast, override
 
 import aiohttp_cors
+import sqlalchemy as sa
 import yarl
 from aiohttp import web
 from pydantic import ValidationError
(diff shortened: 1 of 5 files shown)


src/ai/backend/manager/plugin/keypair/webapp.py
```diff
@@ -41,15 +46,31 @@ async def login(request: web.Request) -> web.Response:
             "Invalid login request data: {}",
             repr(e),
         )
-        raise web.HTTPBadRequest(reason="Invalid JSON data in request body.") from None
+        raise InvalidAuthParameters("Invalid JSON data in request body.") from None
+
+    db = cast(ExtendedAsyncSAEngine, root_app["_db"])
+    async with db.begin_readonly_session() as db_session:
```
fregataa: begin_readonly_session_read_committed

fregataa (approved):

PR #13096, opened 2026-07-24
feat(BA-7001): fail revision creation on malformed definition files

seedspirit:
## Summary
- Fail revision creation when `deployment-config.yaml` or `model-definition.yaml` exists but cannot be parsed or validated.
- Keep missing and empty definition files optional by preserving the existing `None` path.
- Surface a deployment-domain error containing the vfolder and underlying failure context.

## Root cause
`RevisionDraftReader` caught every exception raised while loading vfolder definition files and silently skipped the corresponding overlay. This allowed revisions to be created with unintended defaults when a user-provided file was malformed.

## Test plan
- [x] Verify missing and empty deployment config files return `None`.
- [x] Verify malformed model definitions fail with `DeploymentDefinitionFileReadError`.
- [x] Verify non-mapping and schema-invalid deployment configs fail with `DeploymentDefinitionFileReadError`.
- [x] Run repository push-hook lint, mypy, and affected tests.

Resolves BA-7001


src/ai/backend/manager/repositories/deployment/storage_source/storage_source.py
```diff
@@ -68,7 +70,14 @@ async def fetch_deployment_config(
         raw = await self._fetch_config_file_in_candidates(vfolder_location, candidates)
         if raw is None:
             return None
-        return DeploymentConfigInput.model_validate(dict(raw.payload))
+        try:
+            return DeploymentConfigInput.model_validate(dict(raw.payload))
+        except Exception as e:
```
fregataa, line 75: What kind of exceptions are raised here?

fregataa (approved):

HyeockJinKim (approved):

PR #13107, opened 2026-07-24
feat(BA-6876): add per-target virtual-scope permission check repository APIs

fregataa:
## Summary
- Expose the virtual-scope-chain permission checks on `PermissionControllerRepository`, split by action target: `check_single_entity_permission_via_virtual_scope`, `check_bulk_permission_via_virtual_scope`, and `check_scope_permission_via_virtual_scope`.
- Add `ScopePermissionCheckKey` (`user_id`, `scope`, `subject_entity_type`) so scope checks walk the scope as an entity while matching permission rows on the acted-on entity type, instead of reusing the entity-keyed bulk check.
- Extend the db_source batching key with `subject_entity_type` and add a scope-keyed resolution path sharing the same chain query.

Split out of #13078; the action validators wiring these APIs follow in that PR.

## Test plan
- [x] Existing VS-chain checks stay green with the renamed single-entity API (`tests/unit/manager/repositories/permission_controller`)
- [x] Scope-check tests (permitted / denied / cap clipping / subject-type matching) to be added

Resolves BA-6876

### DIFF
--- a/src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
@@ -59,7 +59,10 @@ from ai.backend.manager.data.permission.types import (
     ScopeListResult,
     ScopeType,
 )
-from ai.backend.manager.data.permission.virtual_scope import VirtualScopePermissionCheckKey
+from ai.backend.manager.data.permission.virtual_scope import (
+    ScopePermissionCheckKey,
+    VirtualScopePermissionCheckKey,
+)
 from ai.backend.manager.data.role_invitation.types import (
     RoleInvitationData,
     RoleInvitationState,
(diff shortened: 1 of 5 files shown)


src/ai/backend/manager/repositories/permission_controller/db_source/db_source.py
```diff
@@ -1215,54 +1220,102 @@ async def check_bulk_permission_via_virtual_scope(
...
+        if not keys:
+            return {}
+        groups: defaultdict[_VirtualScopePermissionGroupKey, list[ScopePermissionCheckKey]] = (
+            defaultdict(list)
+        )
+        for key in keys:
+            groups[
+                _VirtualScopePermissionGroupKey(
+                    user_id=key.user_id,
+                    entity_type=EntityType(key.scope.scope_type),
```
jopemachine, line 1294: If we can't guarantee that ScopeType will only contain values already present in EntityType, would it make sense to add a separate type conversion method?

Instead of raising a generic ValueError, adding a dedicated conversion method that can raise a more descriptive error.

fregataa: I replaced this to new entity type from legacy enum type

HyeockJinKim (approved):

PR #13110, opened 2026-07-24
feat(BA-7009): read app config fragments by config names over GraphQL

jopemachine:
## 📚 Stacked PRs

Part of the app config fragment upsert stack. **Merge in order (bottom-up):**

1. #13113 — `feat(BA-7010)`: add upsert_scoped RBAC write-op primitive
2. #13115 — `feat(BA-7011)`: upsert app config fragments over GraphQL
3. 👉 **#13110 — `feat(BA-7009)`: read app config fragments by config names over GraphQL** ← you are here
4. #13118 — `feat(BA-7012)`: expose fragment by-names read and upsert over REST v2
5. #13196 — `refactor(BA-7013)`: drop the fragment write paths the upsert replaces
6. #13122 — `test(BA-7014)`: cover the scoped app config fragment upsert

## Summary

- Add two GraphQL queries to read the current fragment values at a scope for given config names, so a client can inspect them before editing (upsert).
  - `appConfigFragmentsByNames(scopeType, scopeId, configNames)` — fragments at the given scope, RBAC-authorized at that scope.
  - `myAppConfigFragmentsByNames(configNames)` — the current user's own user-scope fragments (scope resolved via `current_user()`).
- Both reuse the internal `scoped_search` service action with a by-config-names condition — the action itself is unchanged.

## Notes / shape decisions (for review)

- Returns a plain `[AppConfigFragment!]` list (targeted read by names), not a paginated connection.
- `scopeId` is nullable (null for the public scope).
- GraphQL only; the REST equivalent is #13118, which also reshapes these adapter methods to take/return DTOs.

## Test plan
- [ ] `myAppConfigFragmentsByNames` returns the caller's user-scope fragments for the requested names.
- [ ] `appConfigFragmentsByNames` is RBAC-gated at the requested scope.

Resolves BA-7009

🤖 Generated with [Claude Code](https://claude.com/claude-code)

----
📚 Documentation preview 📚: https://sorna--13110.org.readthedocs.build/en/13110/

----
📚 Documentation preview 📚: https://sorna-ko--13110.org.readthedocs.build/ko/13110/

----
📚 Documentation preview 📚: https://sorna--13110.org.readthedocs.build/en/13110/

----
📚 Documentation preview 📚: https://sorna-ko--13110.org.readthedocs.build/ko/13110/

PR #13113, opened 2026-07-24
feat(BA-7010): add upsert_scoped RBAC write-op primitive

jopemachine:
## 📚 Stacked PRs

Part of the app config fragment upsert stack. **Merge in order (bottom-up):**

1. 👉 **#13113 — `feat(BA-7010)`: add upsert_scoped RBAC write-op primitive** ← you are here
2. #13115 — `feat(BA-7011)`: upsert app config fragments over GraphQL
3. #13110 — `feat(BA-7009)`: read app config fragments by config names over GraphQL
4. #13118 — `feat(BA-7012)`: expose fragment by-names read and upsert over REST v2
5. #13196 — `refactor(BA-7013)`: drop the fragment write paths the upsert replaces
6. #13122 — `test(BA-7014)`: cover the scoped app config fragment upsert

## Summary

- Add `RBACEntityUpserter` + `execute_rbac_entity_upserter` (in `repositories/base/rbac/entity_upserter.py`), the upsert counterpart of the scoped creator: `INSERT ... ON CONFLICT (index_elements) DO UPDATE`, then bind each scope via `AssociationScopesEntitiesRow` `ON CONFLICT DO NOTHING` — so a newly inserted row is bound and an updated row keeps its existing binding.
- Supports a partial-index conflict target via `index_where` (e.g. an index guarded by `scope_id IS NULL`).
- A GLOBAL upserter (`scope_ref=None`) binds to no scope.
- Integrity errors other than the conflict target (e.g. a FK gate) map to a domain error or re-raise, driven by the spec's `integrity_error_checks`.
- Expose it as `RBACWriteOps.upsert_scoped`, a one-line delegation like `create_scoped`.

Prerequisite for the app config fragment upsert feature (#13115).

## Contract for callers

- **The conflict target must identify the scope.** Alongside a `scope_ref`, `index_elements` has to include the row's scope-discriminating columns. The conflict path updates whichever row the target matches and then binds it to `scope_ref`, so a scope-blind target would let a caller overwrite another scope's row and pull it into their own.
- **`relation_type` applies only to a binding this upsert inserts.** The association insert ignores conflicts, so an already-bound entity keeps its recorded relation type.

## Notes
- Mirrors `execute_rbac_entity_creator` closely; reuses the base `UpserterSpec` for insert/update value building.
- No dedicated unit test here — the primitive is exercised end-to-end by #13115 and covered by #13122. Happy to add a standalone repository test if preferred.

## Test plan
- [ ] Insert path binds the new row to its scope (association present).
- [ ] Conflict path updates the row in place without duplicating the binding.
- [ ] FK/gate violation surfaces as the domain error.

Resolves BA-7010


HyeockJinKim (approved):

HyeockJinKim:
The implementation is a bit tricky, but I'll go ahead and apply it for now.

PR #13115, opened 2026-07-24
feat(BA-7011): upsert app config fragments over GraphQL

jopemachine:
## 📚 Stacked PRs

Part of the app config fragment upsert stack. **Merge in order (bottom-up):**

1. #13113 — `feat(BA-7010)`: add upsert_scoped RBAC write-op primitive
2. 👉 **#13115 — `feat(BA-7011)`: upsert app config fragments over GraphQL** ← you are here
3. #13110 — `feat(BA-7009)`: read app config fragments by config names over GraphQL
4. #13118 — `feat(BA-7012)`: expose fragment by-names read and upsert over REST v2
5. #13196 — `refactor(BA-7013)`: drop the fragment write paths the upsert replaces
6. #13122 — `test(BA-7014)`: cover the scoped app config fragment upsert

## Summary

- `upsertAppConfigFragments(scope_type, scope_id, items)` — upsert many `(config_name, config)` pairs at the named scope, RBAC-authorized there, all-or-nothing.
- `myUpsertAppConfigFragments(items)` — same at the current user's own user scope (resolved via `current_user()`).
- Built on the `upsert_scoped` primitive (#13113): each item inserts or replaces `config` on conflict; a new row binds to its scope; `public` conflicts on the partial index (`scope_id IS NULL`). The FK to `app_config_allow_list` stays the write gate.

Adds the DTOs, `AppConfigFragmentUpserterSpec`, repository/db_source `bulk_upsert`, the `UpsertAppConfigFragmentsAction` scope action + processor, adapter methods, and the GraphQL mutations.

## Notes
- GraphQL only; the REST equivalent is #13118.
- Verified with ruff (lint+format); mypy/tests via CI.

## Test plan
- [ ] `myUpsertAppConfigFragments` inserts new user-scope fragments and updates existing ones.
- [ ] `upsertAppConfigFragments` at a domain scope is RBAC-gated.
- [ ] Upserting a non-allow-listed config surfaces write-not-allowed.

Resolves BA-7011

🤖 Generated with [Claude Code](https://claude.com/claude-code)

----
📚 Documentation preview 📚: https://sorna--13115.org.readthedocs.build/en/13115/

----
📚 Documentation preview 📚: https://sorna-ko--13115.org.readthedocs.build/ko/13115/

----
📚 Documentation preview 📚: https://sorna--13115.org.readthedocs.build/en/13115/

----
📚 Documentation preview 📚: https://sorna-ko--13115.org.readthedocs.build/ko/13115/

PR #13120, opened 2026-07-25
feat(BA-7013): drop the id-addressed write and scoped-search REST endpoints

jopemachine:
## 📚 Stacked PRs

1. #13113 — feat(BA-7010): add upsert_scoped RBAC write-op primitive
2. #13115 — feat(BA-7011): upsert app config fragments over GraphQL
3. #13110 — feat(BA-7009): read app config fragments by config names over GraphQL
4. #13118 — feat(BA-7012): expose fragment by-names read and upsert over REST v2
5. **#13120 — feat(BA-7013): drop the id-addressed write and scoped-search REST endpoints** ← you are here

Merge in the order above.

## Summary

A fragment is addressed by `(scope, config_name)`, so `/upsert` supersedes the id-addressed writes. Removes:

| Removed | Superseded by |
|---|---|
| `POST /` (create) | `POST /upsert` |
| `PATCH /{fragment_id}` (update) | `POST /upsert` |
| `POST /bulk-update` | `POST /upsert` (upsert is inherently bulk) |
| `POST /scoped/search` | `POST /by-names` |

The `scoped_search` **service action stays** as the internal query path for `/by-names` and the Relay node loader — only its HTTP exposure is gone (the GraphQL `scopedAppConfigFragments` field was already removed).

Remaining routes: `/upsert`, `/by-names`, `/my/upsert`, `/my/by-names`, `/bulk-delete`, `/admin/search`, `GET`/`DELETE /{fragment_id}`.

## Notes for review

- **Scope call I made**: `bulk-update` was not named explicitly in the request, but it is update-by-id in bulk and `/upsert` covers it — say the word and I'll put it back.
- `get`, `bulk-delete` and `DELETE /{fragment_id}` were left as-is (purge stays user-facing).
- The adapter/service/action layers for create and update are left intact (internal, still unit-tested); only the HTTP surface is removed. They are now unused by any handler — happy to remove them in a follow-up if you'd rather not carry dead code.

## Test plan
- [ ] Removed routes return 404.
- [ ] `/upsert` and `/by-names` still work for a non-admin user.

Resolves BA-7013

### DIFF
--- a/changes/13120.feature.md
+++ b/changes/13120.feature.md
@@ -1,0 +1,1 @@
+Remove the id-addressed create, update and bulk-update app config fragment REST endpoints in favour of the scoped upsert, and stop exposing the paginated scoped search.
--- a/docs/manager/rest-reference/openapi.json
+++ b/docs/manager/rest-reference/openapi.json
@@ -590,101 +590,6 @@
         "title": "UpdateAgentResourceGroupBody",
         "type": "object"
       },
-      "AppConfigScopeType": {
-        "description": "Scope at which an app config fragment is written (BEP-1052).\n\nThe single definition shared across the data, DTO, GraphQL, and API layers.",
-        "enum": [
-          "public",
-          "domain",
-          "user"
-        ],
-        "title": "AppConfigScopeType",
-        "type": "string"
-      },
-      "CreateAppConfigFragmentInput": {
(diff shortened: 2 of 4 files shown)


jopemachine:
Folded into #13118 — the add and the removal are one reshaping of the same REST surface, so splitting them added no value.

PR #13132, opened 2026-07-27
feat(BA-6584): add session utilization metric queries

seedspirit:
## Summary

- Add the shared session-utilization batch action and threshold query contract.
- Batch sessions by metric name, kernel policy, and optional time window.
- Filter metrics by allocated resources and query Prometheus once per query shape.
- Support current-value queries and optional `max_over_time` windows per threshold.

Companion checker: #13133

## Test plan

- [x] Verify PromQL generation and Prometheus client requests.
- [x] Verify repository response parsing and failure handling.
- [x] Verify service batching, resource filtering, memory normalization, and query windows.
- [x] Run changed-source lint, MyPy, and tests through the pre-push hook.

Resolves BA-6584


src/ai/backend/common/data/idle_checker/types.py
```diff
@@ -24,6 +26,17 @@ class IdleCheckPhase(enum.StrEnum):
     IDLE_EXPIRED = "idle_expired"
 
 
+class UtilizationThresholdEntry(BackendAISchema):
+    """One preset-backed session utilization threshold."""
+
+    preset_id: UUID = Field(description="Prometheus query preset used to evaluate utilization.")
```
jopemachine: Isn't this value also supposed to use a typed ID?

src/ai/backend/manager/repositories/metric/repositories.py
```diff
@@ -10,10 +13,14 @@ class MetricRepositories:
     repository: MetricRepository
 
     @classmethod
-    def create(cls, args: RepositoryArgs) -> Self:
+    def create(
+        cls,
+        args: RepositoryArgs,
+        prometheus_query_preset_db_source: PrometheusQueryPresetDBSource,
```
jopemachine: Is there a reason why prometheus_query_preset_db_source is passed as a constructor argument instead of being instantiated in repository.py?

src/ai/backend/manager/services/metric/actions/session_utilization.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@dataclass(frozen=True)
+class SessionUtilizationObservation:
+    preset_id: UUID
+    value: Decimal
+
+
+@dataclass(frozen=True)
+class SessionUtilizationAction(QueryMetricAction):
```
jopemachine: Why is this named SessionUtilizationAction? Could you check whether a prefix like Search or Query is missing?

src/ai/backend/manager/services/metric/processors.py
```diff
@@ -25,6 +29,10 @@ class MetricProcessors(AbstractProcessorPackage):
     query_container_live_stat: ActionProcessor[
         ContainerLiveStatAction, ContainerLiveStatActionResult
     ]
+    query_session_utilization: ActionProcessor[
+        QuerySessionUtilizationAction,
+        QuerySessionUtilizationActionResult,
+    ]
```
HyeockJinKim: Please implement this here without using a service as well.

HyeockJinKim (approved):
I'm planning to modify it to use Jinja, so I'll merge this for now.

PR #13138, opened 2026-07-27
feat(BA-6658): apply the virtual-scope chain to the domain scope

fregataa:
## Summary
- Rekey domain-scoped RBAC rows (`association_scopes_entities`, `permissions`) from the domain name to the domain UUID (`domains.id`), and resolve domain ownership through the virtual-scope chain (scope → virtual_scope → entity).
- Domain creation now goes through `RBACWriteOps.create_scope` (virtual-scope node + self edges + SYSTEM/preset roles) and purge through `delete_scope`; project creation binds the domain scope into the project's virtual scope; user creation adds the user as an entity member of its domain's virtual scope.
- All writers that addressed the domain scope by name now pass the domain UUID (legacy GraphQL resource-group mutations resolve it via the domain processor), the scope-chain permission resolver canonicalizes DOMAIN keys that still arrive as names from request contexts, and scope search / RBAC GraphQL types emit and load domains by UUID.
- Ship an idempotent Alembic data migration (`857c4d02c4b9`) that converts existing name-keyed rows (with duplicate-safe guards) and backfills the virtual-scope chain for existing domains, projects, and users.

## Test plan
- [x] Migration verified against the local halfstack DB inside a rolled-back transaction: upgrade converts name-keyed rows and provisions the virtual-scope chain, re-running upgrade statements is a no-op, and downgrade restores name-keyed rows and drops domain virtual scopes.
- [x] CI unit tests pass (domain/group/user/scaling-group repository tests updated to UUID-keyed expectations).
- [x] Verify with the live server: domain create/purge provisions and removes the virtual scope, and domain-scoped permission checks pass for name- and UUID-addressed targets.

Resolves BA-6658


src/ai/backend/manager/api/adapters/project/adapter.py
```diff
@@ -271,8 +284,9 @@ async def search_by_domain(
             base_conditions=base_conditions,
         )
 
+        domain_id = await self._resolve_domain_id(scope.domain_name)
         action_result = await self._processors.group.search_projects_by_domain.wait_for_complete(
-            SearchProjectsByDomainAction(scope=scope, querier=querier)
+            SearchProjectsByDomainAction(scope=scope, querier=querier, _domain_id=domain_id)
```
HyeockJinKim: I don't think the "domain" needs to be included as an argument. It seems more appropriate for the service to provide only scope queries.

src/ai/backend/manager/api/gql_legacy/scaling_group.py
```diff
@@ -832,7 +838,7 @@ async def mutate(
                             domain=domain,
                         ),
                         entity_ref=RBACElementRef(RBACElementType.RESOURCE_GROUP, scaling_group),
-                        scope_ref=RBACElementRef(RBACElementType.DOMAIN, domain),
+                        scope_ref=RBACElementRef(RBACElementType.DOMAIN, str(domain_data.id)),
```
HyeockJinKim, line 841: Do I still need to convert the id value to a string?

fregataa: Yes — `RBACElementRef.element_id` is typed `str` since element ids are heterogeneous across element types. So the `str()` conversion is still needed.

HyeockJinKim (approved):

PR #13163, opened 2026-07-27
feat(BA-6660): apply the virtual-scope chain to the container registry scope

fregataa:
## Summary

- Route container registry ownership through the virtual-scope chain (`scope -> virtual_scope -> entity`): registry creation goes through `RBACWriteOps.create_scope` (registry row + virtual-scope node with self edges) and deletion through `delete_scope` (RBAC entries + virtual scope, edges via FK cascade).
- Express an allowed group as both virtual-scope edges — the registry becomes an entity member of the project's virtual scope, and the project scope is bound into the registry's virtual scope so project-scoped permissions reach the registry's images. Removal drops the N:N mapping row and both edges, replacing the legacy scope binder/unbinder (`scope_binders.py` deleted).
- Image rescan registers newly created images as entity members of the registry's virtual scope; `ensure_scope` backfills the node for registries created before the chain existed (full data backfill is BA-6664).
- The repository and the registry scanner hold an `RBACOpsProvider` and run each operation in a single `write_ops` transaction; a generic `ScopeEntityMember` is added for non-user entity memberships.

Stacked on #13138 (BA-6658, domain scope) — targets that branch and should be retargeted/rebased onto `main` once it lands.

## Test plan

- [ ] Create a registry with allowed groups → `virtual_scopes` node with self `scope_bindings`/`entity_memberships`, project→registry membership and binding rows exist
- [ ] Modify allowed groups (add/remove) and set `is_global=true` → edges added/removed accordingly
- [ ] Rescan a registry → new images appear as `entity_memberships` of the registry's virtual scope
- [ ] Delete a registry → registry row, RBAC entries, and virtual scope (with cascaded edges) removed
- [ ] Update/extend container registry repository unit tests

Resolves BA-6660


fregataa:
Superseded by #13406, rebuilt on the current RBAC ops infrastructure (post BA-7035/BA-7152).

PR #13211, opened 2026-07-29
fix(BA-7009): answer a by-names fragment read position by position

jopemachine:
Resolves BA-7009

## Summary

`scopedAppConfigFragmentsByNames` / `myAppConfigFragmentsByNames` (merged in #13110) dropped a requested name from the answer when the scope held no fragment for it, so a caller could not tell which of its names came back empty and could not read the answer by position.

Each name now holds its place:

| Requested | At the scope | Answer |
|---|---|---|
| `["theme", "menu"]` | `theme` only | `[theme, null]` |
| `["theme", "theme"]` | `theme` | `[theme, theme]` |
| `["a", "b"]` | neither | `[null, null]` |

- Schema: `[AppConfigFragment!]!` → `[AppConfigFragment]!` (the list stays non-null; its items may be null).
- The adapter answers `list[AppConfigFragmentNode | None]`, mapping each requested name through the fragments found.
- A scope with no fragment for a name is the normal state before the first write, so it stays a null rather than a not-found.

## Note for review

`api/gql/AGENTS.md` says nullable list items are the Relay node loader's shape (`resolve_nodes`). This query is the other case the rule did not anticipate — a positional batch read where absence is expected and the caller needs to match answers to its own request. Happy to write the exception into that document if you agree with the shape.

## Test plan
- [ ] Read two names where only one has a fragment at the scope — the missing one answers null in place.
- [ ] Read a name twice — it answers twice.

----
📚 Documentation preview 📚: https://sorna--13211.org.readthedocs.build/en/13211/

----
📚 Documentation preview 📚: https://sorna-ko--13211.org.readthedocs.build/ko/13211/

### DIFF
--- a/src/ai/backend/manager/api/gql/app_config_fragment/resolver.py
+++ b/src/ai/backend/manager/api/gql/app_config_fragment/resolver.py
@@ -145,8 +145,8 @@ async def my_upsert_app_config_fragments(
         added_version=NEXT_RELEASE_VERSION,
         description=(
             "Read the fragments written at one scope for the given config names — the current "
-            "values, to inspect before editing them. Answered in the order the names were "
-            "given, leaving out a name with no fragment there. RBAC-authorized at that scope."
+            "values, to inspect before editing them. Answered position by position, null "
+            "where the scope holds no fragment for that name. RBAC-authorized at that scope."
         ),
     )
 )  # type: ignore[misc]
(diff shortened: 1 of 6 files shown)


src/ai/backend/manager/api/gql/app_config_fragment/resolver.py
```diff
@@ -145,36 +145,40 @@ async def my_upsert_app_config_fragments(
...
     config_names: list[str],
-) -> list[AppConfigFragmentGQL]:
+) -> list[AppConfigFragmentGQL | None]:
     nodes = await info.context.adapters.app_config_fragment.my_app_config_fragments_by_names(
         MyAppConfigFragmentsByNamesInput(config_names=config_names)
     )
-    return [AppConfigFragmentGQL.from_pydantic(node) for node in nodes]
+    return [
+        AppConfigFragmentGQL.from_pydantic(node) if node is not None else None for node in nodes
+    ]
```
HyeockJinKim: It seems like this implementation would need to be in the adapter so that both the REST API and GraphQL return consistent values—is there a reason you only modified the GQL?

HyeockJinKim (changes requested):

fregataa (dismissed):

seedspirit (dismissed):

HyeockJinKim (approved):

PR #13222, opened 2026-07-29
feat(BA-7060): implement scope action monitors (audit log/reporter/Prometheus)

fregataa:
## Summary
- Add the monitor implementations bound to the pure-ABC `ScopeActionMonitor` base: `ScopeActionAuditLogMonitor` (one audit-log row per target scope, recorded as `"{scope_type}:{scope_id}"`), `ScopeActionReporterMonitor` (reports once per run with `entity_id=None` — reporter messages are per-run notifications, and the per-target trail belongs to the audit-log monitor), and `ScopeActionPrometheusMonitor` (one observation per run — the action metric is keyed by `(entity_type, operation, status)`).
- Introduce an `ActionMonitors` dataclass (mirroring `ActionValidators`) that groups the legacy monitor list with the per-type monitor lists, and wire the scope monitors through the DI composer and `create_processors`.
- Existing `BaseAction`-era processor packages keep receiving the legacy flat list unchanged. The `ActionMonitors` plumbing is intentionally identical to #13219 (BA-7059) and #13225 (BA-7061); whichever merges later only needs a trivial conflict resolution in the composer's per-type monitor lists.

## Test plan

Resolves BA-7060

### DIFF
--- a/changes/13222.feature.md
+++ b/changes/13222.feature.md
@@ -1,0 +1,1 @@
+Add scope action monitors (audit log, reporter, Prometheus) and wire them through the DI composer
--- a/src/ai/backend/manager/actions/monitors/__init__.py
+++ b/src/ai/backend/manager/actions/monitors/__init__.py
@@ -1,0 +1,22 @@
+from dataclasses import dataclass, field
+
+from ai.backend.manager.actions.bulk.monitor.base import BulkActionMonitor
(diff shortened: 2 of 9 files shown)


fregataa:
Superseded by #13255 — the work was re-sliced per the updated BA-7060 scope (scope action monitors covering all three backends).

PR #13224, opened 2026-07-29
feat(BA-7035): materialize the User scope's virtual scope on creation

fregataa:
## Summary

- Create every user (admin create, bulk create, signup) as a scope via `RBACWriteOps.create_scope`: the user row with its domain/project associations, its virtual scope (self membership + self scope binding), and its own-scope roles (SYSTEM + presets) in one transaction, per BEP-1062. The user's roles are granted through the new `assign_roles_to_user` op using the `auto_grant_role_ids` returned by `create_scope`.
- Consolidate RBAC writes inside `RBACWriteOps`: `add_users_to_scope` now fully delegates UUID-keyed enrollments to `add_entity_members` (name-keyed scopes such as domains keep the legacy path until they are UUID-keyed), auto-assign role granting is absorbed as an internal op instead of delegating to the deprecated `RoleManager`, and a user scope's system role is flagged `auto_assign` by the provider's scope-type rule.
- Convert the user/auth repositories off raw sessions onto provider-owned `write_ops()` transactions: user create/bulk-create/purge, signup (`insert_user_with_keypair` now uses creator specs — the signup user also gains its domain scope association), and `assign_project_membership`. User purge drops the virtual scope node together with the row via `batch_delete_scopes`.

Intentionally out of scope: `project -> VS_user` hierarchy bindings (a project must not reach a user's entities outside that project; only `domain -> VS_user` is justified and awaits UUID-keyed domains in BA-6658 — BEP-1062 3.(a)/(d) needs a follow-up amendment), keypair/session/vfolder entity memberships (separate creation paths; keypair is blocked by its string primary key), and the `update_user_validated` group-sync path (still session-based; mirrors after its ops conversion).

## Test plan

- [x] `pants test tests/unit/manager/repositories/ops/rbac::` — VS materialization for scope creation/ensure, UUID-keyed enrollment mirroring, name-keyed skip, legacy-scope removal tolerance
- [x] `pants test tests/unit/manager/repositories/user::` — user creation materializes the VS (self membership + self binding), purge removes it, auto_assign preset role granting
- [x] `pants test tests/unit/manager/repositories/group:: tests/unit/manager/repositories/permission_controller:: tests/unit/manager/services/auth::` — no regressions in project/role/signup flows

Resolves BA-7035


src/ai/backend/manager/repositories/auth/db_source/db_source.py
```diff
@@ -137,43 +140,36 @@ async def verify_email_exists(self, email: str) -> bool:
...
-                spec=UserRoleCreatorSpec(user_id=user_row.uuid, role_id=role.id)
+        project_ids: Collection[ProjectID],
+        *,
+        keypair_resource_policy: str,
+        keypair_rate_limit: int,
+    ) -> UserCreationData:
+        """Provision a signup user in one transaction: the row, its default keypair,
+        and its domain/project (model-store included) scope enrollments."""
+        async with self._rbac_ops_provider.write_ops() as w:
+            domain_name = user_data["domain_name"]
```
HyeockJinKim: Accessing the dictionary is making my head spin.

fregataa: it changed to use UserCreatorSpec

src/ai/backend/manager/repositories/ops/rbac/provider.py
```diff
@@ -670,6 +729,106 @@ async def _grant_auto_assign_roles(
...
+    ) -> list[ProjectID]:
+        """``project_ids`` narrowed to the domain's real projects, plus the domain's
+        model-store projects that every user joins."""
+        stmt = (
+            sa.select(GroupRow.id)
+            .join(DomainRow, DomainRow.name == GroupRow.domain_name)
+            .where(
+                DomainRow.id == domain_id,
+                sa.or_(GroupRow.id.in_(project_ids), GroupRow.type == ProjectType.MODEL_STORE),
+            )
```
HyeockJinKim, line 828: What kind of query is this?

fregataa: this is an existing step that adding a created user to model-store project

HyeockJinKim (approved):

PR #13248, opened 2026-07-29
refactor(BA-7074): consolidate user enrollment into RBAC entity-member ops

fregataa:
## Summary

- Remove `RBACWriteOps.add_users_to_scope`: user enrollment now goes through the virtual-scope path (`ensure_scope` + `add_entity_members`) with the new `ScopeUserMember`, a user member that always receives the scope's `auto_assign` roles. Domain-scope enrollment is dropped until domain RBAC rows are UUID-keyed (BA-6658); `assign_users_to_scope` loses its `domain_name` parameter across the repository/service layers.
- Drop the `RoleManager` dependency from `RBACWriteOps`: `auto_assign` role granting is absorbed as an internal op that resolves the scope's active `auto_assign` roles and the target users' existing grants in a single outer-join query, then bulk-inserts only the missing pairs.
- Promote the scope/entity type constants scattered as private module constants (`_PROJECT_SCOPE_TYPE`, `_USER_ENTITY_TYPE`) to public constants in `ai.backend.common.data.entity.types` (`DOMAIN/PROJECT/USER_SCOPE_TYPE`, `USER_ENTITY_TYPE`), derived from `RBACElementType` so the values stay in sync.

## Test plan

- [x] `pants test tests/unit/manager/repositories/ops/rbac::` — entity-member addition/removal, auto_assign granting via the internal op
- [x] `pants test tests/unit/manager/repositories/user:: tests/unit/manager/repositories/group:: tests/unit/manager/services/user::` — no regressions in user/project enrollment flows
- [x] `pants test tests/component/user::` — a new user receives the auto_assign roles of its initial project and model-store scopes

Resolves BA-7074

### DIFF
--- a/src/ai/backend/common/data/entity/types.py
+++ b/src/ai/backend/common/data/entity/types.py
@@ -7,6 +7,14 @@ from ai.backend.common.identifier.scope import ScopeID
 EntityType = NewType("EntityType", str)
 ScopeType = NewType("ScopeType", str)
 
+# Canonical scope/entity type values. Only types that already exist in
+# RBACElementType (i.e. are managed by RBAC) share its value; new types are
+# defined as raw strings without consulting RBACElementType.
+DOMAIN_SCOPE_TYPE = ScopeType("domain")
+PROJECT_SCOPE_TYPE = ScopeType("project")
+USER_SCOPE_TYPE = ScopeType("user")
+USER_ENTITY_TYPE = EntityType("user")
+
 
 @dataclass(frozen=True, slots=True)
 class ScopeRef:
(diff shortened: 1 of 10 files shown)


src/ai/backend/common/data/entity/types.py
```diff
@@ -7,6 +7,14 @@
 EntityType = NewType("EntityType", str)
 ScopeType = NewType("ScopeType", str)
 
+# Canonical scope/entity type values. Only types that already exist in
+# RBACElementType (i.e. are managed by RBAC) share its value; new types are
+# defined as raw strings without consulting RBACElementType.
+DOMAIN_SCOPE_TYPE = ScopeType("domain")
+PROJECT_SCOPE_TYPE = ScopeType("project")
+USER_SCOPE_TYPE = ScopeType("user")
+USER_ENTITY_TYPE = EntityType("user")
```
HyeockJinKim: If this was separated from the enum into a scope type, shouldn't it be placed in a separate package?

HyeockJinKim (approved):

PR #13254, opened 2026-07-29
feat(BA-7059): implement single-entity action monitors (audit log/reporter/Prometheus)

fregataa:
## Summary
- Add the monitor implementations bound to the pure-ABC `SingleEntityActionMonitor` base: `SingleEntityActionAuditLogMonitor` (one audit-log row per run, always recording the entity id), `SingleEntityActionReporterMonitor` (per-run report carrying the entity id), and `SingleEntityActionPrometheusMonitor` (one observation per run — the action metric is keyed by `(entity_type, operation, status)`).
- Introduce an `ActionMonitors` dataclass (mirroring `ActionValidators`) that groups the legacy monitor list with the per-type monitor lists, and wire the single-entity monitors through the DI composer and `create_processors`.
- Existing `BaseAction`-era processor packages keep receiving the legacy flat list unchanged. The `ActionMonitors` plumbing is intentionally identical to #13255 (BA-7060) and #13256 (BA-7061); whichever merges later only needs a trivial conflict resolution in the composer's per-type monitor lists.

## Test plan

Resolves BA-7059

### DIFF
--- a/changes/13254.feature.md
+++ b/changes/13254.feature.md
@@ -1,0 +1,1 @@
+Add single-entity action monitors (audit log, reporter, Prometheus) and wire them through the DI composer
--- a/src/ai/backend/manager/actions/monitors/__init__.py
+++ b/src/ai/backend/manager/actions/monitors/__init__.py
@@ -1,0 +1,22 @@
+from dataclasses import dataclass, field
+
+from ai.backend.manager.actions.bulk.monitor.base import BulkActionMonitor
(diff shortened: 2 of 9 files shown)


jopemachine:
How about explicitly renaming ActionMonitor to LegacyActionMonitor?

fregataa:
> How about explicitly renaming ActionMonitor to LegacyActionMonitor?

This change will touch more than 170 files. It will be better to open a separate PR for this rename work

HyeockJinKim (approved):

PR #13255, opened 2026-07-29
feat(BA-7060): implement scope action monitors (audit log/reporter/Prometheus)

fregataa:
## Summary
- Add the monitor implementations bound to the pure-ABC `ScopeActionMonitor` base: `ScopeActionAuditLogMonitor` (one audit-log row per target scope, recording the scope's id as `entity_id`), `ScopeActionReporterMonitor` (one reporter message per target scope with the scope's id as `entity_id`, matching the audit-log rows), and `ScopeActionPrometheusMonitor` (one observation per run — the action metric is keyed by `(entity_type, operation, status)`). Per-scope records share the `action_id`, which ties them back to the same run.
- Introduce an `ActionMonitors` dataclass (mirroring `ActionValidators`) that groups the legacy monitor list with the per-type monitor lists, and wire the scope monitors through the DI composer and `create_processors`.
- Existing `BaseAction`-era processor packages keep receiving the legacy flat list unchanged. The `ActionMonitors` plumbing is intentionally identical to #13254 (BA-7059) and #13256 (BA-7061); whichever merges later only needs a trivial conflict resolution in the composer's per-type monitor lists.

## Test plan

Resolves BA-7060

### DIFF
--- a/changes/13255.feature.md
+++ b/changes/13255.feature.md
@@ -1,0 +1,1 @@
+Add scope action monitors (audit log, reporter, Prometheus) and wire them through the DI composer
--- a/src/ai/backend/manager/actions/scope/monitor/audit_log.py
+++ b/src/ai/backend/manager/actions/scope/monitor/audit_log.py
@@ -1,0 +1,58 @@
+from __future__ import annotations
+
+from typing import override
+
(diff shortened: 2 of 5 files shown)


HyeockJinKim (approved):
I'll make some changes to the audit log section later.

PR #13265, opened 2026-07-29
chore: enable mypy truthy-bool error code

fregataa:
## Summary
- Enable the `truthy-bool` mypy error code in `pyproject.toml` and fix all 73 reported errors across 46 files.
- Most fixes either remove always-true dead checks on non-Optional values or convert intended None-checks to explicit `is not None` comparisons (adding `Optional` annotations where the value genuinely can be unset).
- The new error code surfaced four real runtime bugs, now fixed:
  - `common/redis_client.py`: `RedisConnection.disconnect()` raised `AttributeError` when called before/after a failed `connect()`; `reader`/`writer` are now declared Optional and guarded.
  - `test/testcases/session/vfolder_mount.py`: a failure in `_create_dummy_file()` caused `UnboundLocalError` in the `finally` block, masking the original exception.
  - Alembic `1d42c726d8a3`: the early-return checked a `CursorResult` (always truthy), so the migration created a local registry row and ran a no-op UPDATE even with nothing to migrate; it now checks fetched rows, which also improves idempotency on re-runs.
  - `manager/repositories/model_serving/repository.py`: `delete_auto_scaling_rule` caught SQLAlchemy's `NoResultFound` while the row loader raises `ObjectNotFound`, so a missing rule leaked a 404 instead of returning `False` per its contract.
- In `client/cli`, the flagged `Coroutine` truthiness checks are the known `@api_function` sync-Session typing limitation; they are handled with `cast(...)` following the existing precedent in `client/cli/image.py`.

## Test plan
- [x] `pants fmt` / `pants fix` / `pants lint` pass locally
- [x] CI typecheck (`pants check`) passes with the new error code enabled

### DIFF
--- a/src/ai/backend/common/redis_client.py
+++ b/src/ai/backend/common/redis_client.py
@@ -216,8 +216,8 @@ class RedisConnection(AbstractAsyncContextManager[RedisClient]):
     _socket_connect_timeout: float | None
     _keepalive_options: dict[int, int]
 
-    reader: asyncio.StreamReader | None
-    writer: asyncio.StreamWriter | None
+    _reader: asyncio.StreamReader | None
+    _writer: asyncio.StreamWriter | None
 
     def __init__(
         self,
@@ -230,8 +230,8 @@ class RedisConnection(AbstractAsyncContextManager[RedisClient]):
     ) -> None:
         self._redis_target = redis_target
         self._db = db
-        self.reader = None
-        self.writer = None
+        self._reader = None
+        self._writer = None
 
         self._socket_timeout = socket_timeout
         self._socket_connect_timeout = socket_connect_timeout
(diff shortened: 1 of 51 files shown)


src/ai/backend/common/redis_client.py
```diff
@@ -216,6 +216,9 @@ class RedisConnection(AbstractAsyncContextManager[RedisClient]):
     _socket_connect_timeout: float | None
     _keepalive_options: dict[int, int]
 
+    reader: asyncio.StreamReader | None
```
jopemachine: Nit: Let's keep it private (_reader, _writer)

jopemachine (dismissed):

jopemachine (approved):

PR #13283, opened 2026-07-30
feat(BA-7098): expose max_priority on the keypair resource policy API

jopemachine:
## 📚 Stacked PRs

- #13283 — `BA-7098` feat: expose max_priority on the keypair resource policy API  ← you are here
- #13290 — `BA-7101` fix: stop clearing GQL update fields that the client omitted

---

`max_priority` caps the scheduler priority a user may declare at enqueue time. The column exists with a range constraint, the scheduler loads it, and `PriorityLimitRule` enforces it — but the field reached no request DTO, no response node, and no CLI flag, so it stayed NULL and the rule returned immediately on every enqueue. The cap had never taken effect.

Carried through the same path as `max_pending_session_count`, which has the same shape (nullable int, `TriState` on update): create and update inputs, the policy node, `KeyPairResourcePolicyData`, the row conversion, the creator and updater specs, the adapter, and a `--max-priority` flag.

## Verified locally

```bash
./bai admin resource-policy keypair update default --max-priority 20
```

| step | result |
|---|---|
| CLI flag | `--max-priority INTEGER` now listed |
| set | stored as `20` |
| read back | `max_priority: 20` |
| enqueue with `priority: 50` | `400` — `The requested session priority 50 exceeds the allowed maximum priority of 20.` |
| enqueue with `priority: 10` | session created |

The last two are `PriorityLimitRule` firing for the first time.

## Notes

Existing tests needed the new required field on `KeyPairResourcePolicyData` and the creator spec; no new tests were added.

Legacy GraphQL is untouched — new surface goes to v2 per the repo guidelines.

Found while running CHECKLIST_26.8 item 4 (Scheduling Policy Controls), whose `max_priority` check could not be exercised without this.

## Related

- BA-7098
- BEP-1055 Preemption Scheduler Mechanics, section 2.4 — the keypair cap on the global `priority`

----
📚 Documentation preview 📚: https://sorna--13283.org.readthedocs.build/en/13283/

----
📚 Documentation preview 📚: https://sorna-ko--13283.org.readthedocs.build/ko/13283/


src/ai/backend/manager/api/gql/resource_policy_v2/types/mutations.py
```diff
@@ -129,6 +140,13 @@ class UpdateKeypairResourcePolicyInputGQL(PydanticInputMixin[UpdateKeypairResour
     max_pending_session_count: int | None = gql_field(
         default=UNSET, description="Updated max pending sessions."
     )
+    max_priority: int | None = gql_added_field(
+        BackendAIGQLMeta(
+            added_version=NEXT_RELEASE_VERSION,
+            description="Updated max scheduling priority. Set to null to clear (uncapped).",
+        ),
+        default=UNSET,
```
fregataa, line 148: Does this set max_priority to null if it is not set?

jopemachine: I will handle this in follow up PR

fregataa (approved):

PR #13287, opened 2026-07-30
fix(BA-7100): reject a half-specified revision resource config

jopemachine:
`add_revision` built the `ResourceSpec` only when `cluster_config` and `resource_config` were both present:

```python
resource_spec = None
if input.cluster_config is not None and input.resource_config is not None:
    resource_spec = ResourceSpec(...)
```

Supplying one silently discarded the other, and the request then failed on the slots it had actually been given:

```
InvalidAPIParameters: resource_slots is missing required resource slot(s): ['cpu', 'mem'].
```

Both fields are optional on `AddRevisionInput` because a revision preset may supply them — unlike `CreateRevisionInput`, where both are required. That makes silent discard the wrong answer for a half-specified pair, so it is rejected instead, the way `deployment_controller` already rejects a missing `image_id`, `mounts`, or `model_definition`.

## Verified locally

| request | before | after |
|---|---|---|
| `resource_config` only | `resource_slots is missing required resource slot(s): ['cpu','mem']` | `cluster_config is required alongside the one provided…` |
| `cluster_config` only | silently ignored | `resource_config is required alongside the one provided…` |
| both | revision created | revision created |
| neither | inherits from the preset | unchanged |

## How it was found

Running CHECKLIST_26.8 item 12 (Model Serving / Deployment) against 26.8.0rc1. It took six attempts to add a revision, because each rejection named a different field than the one actually missing.

## Related

- BA-7100

### DIFF
--- a/changes/13287.fix.md
+++ b/changes/13287.fix.md
@@ -1,0 +1,1 @@
+Reject a deployment revision that specifies only one of `cluster_config` / `resource_config`, instead of silently discarding both.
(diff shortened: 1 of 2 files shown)


jopemachine:

PR #13290, opened 2026-07-30
fix(BA-7101): stop clearing GQL update fields that the client omitted

jopemachine:
## 📚 Stacked PRs

- #13283 — `BA-7098` feat: expose max_priority on the keypair resource policy API ✅
- #13290 — `BA-7101` fix: stop clearing GQL update fields that the client omitted  ← you are here
- #13294 — `BA-7102` fix: stop declaring a clearable state the column cannot hold

Merge in order, bottom-up. #13283 has landed, so this PR is now based on `main`; each PR shows only its own layer.

---

An update mutation that mentions one field silently wipes others. Reported on #13283 ([review comment](https://github.com/lablup/backend.ai/pull/13283#discussion_r3680167313)) and split out here, since the defect predates that PR and spans four modules.

`updateKeypairResourcePolicy(name: "default", input: {idleTimeout: 1200})` also clears `max_priority`, `max_pending_session_count` and `max_pending_session_resource_slots`.

## The chain

`PydanticInputMixin.to_pydantic()` skips `strawberry.UNSET` and nothing else. Three modules shadow that name with a module-local `UNSET = None`, and `login_client_type` spells `default=None` per field. Either way the GQL field default is a null, which Strawberry renders into the SDL, so graphql-core fills every omitted field with it — and the adapter reads that null as "clear this column".

| step | before |
|---|---|
| SDL | `maxPriority: Int = null` |
| `coerce_input_value({}, UpdateKeypairResourcePolicyInput)` | `{'maxPriority': None, 'maxPendingSessionCount': None, …}` — 11 fields filled in |
| `to_pydantic()` | `max_priority=None` (not `SENTINEL`) |
| adapter | `TriState.nullify()` |
| updater | `to_update["max_priority"] = None` |

After the fix, `coerce_input_value({}, …)` returns `{}` and `to_pydantic()` yields `SENTINEL`, so the three input states stay distinct: **omitted keeps the value, explicit null clears it, a value sets it.** The `isinstance(dto.x, Sentinel)` branch already present in every adapter was dead code until now.

## Scope

15 fields in the codebase map a null to `TriState.nullify()`. Nine of them had a null GQL default:

| adapter | fields |
|---|---|
| `resource_policy` | `max_priority`, `max_pending_session_count`, `max_pending_session_resource_slots`, `max_concurrent_logins` |
| `domain` | `description`, `integration_name` |
| `project` | `description`, `integration_name` |
| `login_client_type` | `description` |

The other six (`user`, `object_storage`, `image`) already use `strawberry.UNSET` correctly and serve as the reference — this PR brings the rest in line with them.

Three of those modules keep their local `UNSET = None`; the nine fields spell out `strawberry.UNSET` instead, so the schema diff is exactly those nine and nothing else moves. Dropping the shim would have flipped every field in the module — including create inputs, where an omitted field has no stored value to destroy and the DTO default is null either way. The shim now carries a comment so the two spellings do not read as a typo.

Deliberately **not** in scope:

- Create inputs, and update fields the adapter maps to `OptionalState` — a null there is ignored, not acted on.
- Modules whose DTO fields accept `Sentinel` but whose adapters map a null to `OptionalState.nop()` (`model_card`, `resource_preset`, `runtime_variant_preset`, `deployment/revision_preset`). Nothing is destroyed there today.
- The `UNSET = None` shim itself. Removing it is a cleanup, not this fix.
- The same defect pointing the other way — `total_resource_slots`, `allowed_vfolder_hosts`, `group_ids` and five more declare a clearable state their column cannot hold. Handled in #13294.
- Fourteen fields that are optional in GQL but required on the DTO (e.g. `CreateModelCardInputGQL.framework`); switching those to `UNSET` would turn a silent null into a validation error.

## Schema impact

The `= null` default disappears from exactly the nine fields above. Client queries stay valid — the fields are nullable, so no default is needed to omit them. What changes is server-side behavior, which is the point. `schema.graphql` (v1) is untouched, so `graphql-inspector` sees nothing.

## Verified

- Each of the nine fields is asserted in all three states: omitted → `SENTINEL`, explicit null → `None`, value → the value.
- `pants fmt / fix / lint / check --changed-since=origin/main` — clean
- 52 tests pass

Resolves BA-7101


src/ai/backend/manager/api/gql/resource_policy_v2/types/mutations.py
```diff
@@ -262,13 +261,13 @@ class CreateProjectResourcePolicyInputGQL(PydanticInputMixin[CreateProjectResour
...
+        default=None, description="Updated max vfolder count."
     )
     max_quota_scope_size: BinarySizeInputGQL | None = gql_field(
-        default=UNSET, description="Updated max quota scope size."
+        default=None, description="Updated max quota scope size."
     )
     max_network_count: int | None = gql_field(
-        default=UNSET, description="Updated max network count."
+        default=None, description="Updated max network count."
     )
```
fregataa, line 271: Is None default value correct?

fregataa (approved):

PR #13320, opened 2026-07-31
feat(BA-7114): add role name template support to role presets

fregataa:
## Summary
- Add a nullable Jinja `role_name_template` column to role presets (e.g. `{{scope.type}}-{{scope.name}}-member`); when roles are instantiated from a preset at scope creation, the template is rendered with the target scope's attributes (`scope.id` / `scope.name` / `scope.type`) so per-scope roles get distinguishable names instead of the preset's fixed name.
- Rendering and validation live in `RBACWriteOps` (sandboxed Jinja + StrictUndefined): rendering is private to the ops, and `validate_role_name_template()` is a public ops method — no separate engine object or DI plumbing.
- `RolePresetDBSource` takes an `RBACOpsProvider` and validates `role_name_template` inside the create/update transaction, so bad templates (syntax errors, undefined variables, empty/oversized results) are rejected before the preset is stored; `RolePresetRepository` stays pure delegation.
- At scope creation, scope attributes are batch-resolved through the scope-row registry from #13390 (single UNION ALL query), and rendering falls back to `{scope_type}-{scope_id[:8]}-role` (plain string formatting, never the template engine) on unresolvable scopes or invalid templates — scope creation never fails on role naming.
- This PR covers only the repository/ops layers; exposing `role_name_template` over the v2 API stack (DTO/GQL/adapter/CLI) is a follow-up.

## Test plan
- [x] Unit tests for rendering/validation via `RBACWriteOps`: variable substitution, whitespace stripping, undefined-variable/syntax errors, empty/oversized results, sandbox escape blocking
- [x] `RoleManager.create_preset_roles` tests still pass; a new test documents that the legacy path keeps the preset's fixed name
- [x] CI: `pants check` / `pants test` across changed targets
- [x] Live verification: seed a preset with a name template, create a project, and confirm the instantiated role name is rendered

Resolves BA-7114

----
📚 Documentation preview 📚: https://sorna--13320.org.readthedocs.build/en/13320/

----
📚 Documentation preview 📚: https://sorna-ko--13320.org.readthedocs.build/ko/13320/

----
📚 Documentation preview 📚: https://sorna--13320.org.readthedocs.build/en/13320/

----
📚 Documentation preview 📚: https://sorna-ko--13320.org.readthedocs.build/ko/13320/


src/ai/backend/manager/repositories/group/repository.py
```diff
@@ -34,6 +34,9 @@
     GroupSearchResult,
     UserProjectSearchScope,
 )
+from ai.backend.manager.repositories.permission_controller.role_name_template import (
+    RoleNameTemplateEngine,
+)
```
HyeockJinKim: I don't quite understand why this implementation is here—could you explain it to me?

fregataa: I added this template engine to the RBACOpsProvider's init method

HyeockJinKim (approved):

PR #13328, opened 2026-07-31
feat(BA-7120): add session idle-check exclusion service, actions, and repository ops

seedspirit:
## Summary
- Implement the whole logic chain for session idle-check exclusion below the API adapter (the surface lives in #13682, stubbed at the adapter until this lands):
  - **Actions/service**: first production v2 bulk actions (`exclude_session_idle_checks` / `include_session_idle_checks`) wired through `ProcessorGroup.bulk()`, which attaches the virtual-scope RBAC bulk validator and per-entity audit fan-out. The action results carry the repository's `success`/`errors` partitions and derive per-entity audit fields via `ActionRunStatus`, so a bulk entity's error reads exactly like a single run's.
  - **Repository**: `batch_exclude_session_idle_checks` runs one savepoint-isolated `INSERT ... ON CONFLICT DO UPDATE` per pair (creating rows assignment-sync has not materialized yet); session existence is enforced per row by the `session_id` foreign key, mapped to `SessionNotFound` via the upserter spec's `integrity_error_checks` — no pre-validation query, no validation/write race. `batch_include_session_idle_checks` conditionally resets only `EXCLUDED` rows (idempotent; unknown sessions fail per entity). Both answer with `SessionIdleCheckBatchResult(success, errors)` carrying per-session exceptions; an unknown checker rejects the whole batch.
  - **Ops surface**: add `WriteOps.bulk_upsert` (single multi-VALUES statement) and `WriteOps.bulk_upsert_partial` (savepoint per row with `IntegrityError` → domain-error matching).
- Reworked from the original assignment-keyed design per the BA-7243 API review; `IdleCheckPhase.EXCLUDED` itself landed separately via BA-7242.

## Test plan
- [x] `pants test tests/unit/manager/repositories/idle_checker/test_repository.py` — exclude overwrites any phase / upserts missing pair rows / deduplicates repeated ids / reports unknown sessions as per-row `SessionNotFound` errors with the valid rows applied; include resets only `EXCLUDED` rows, is idempotent, no-ops on absent pairs, and fails unknown sessions per entity; an unknown checker rejects the whole batch
- [x] `pants test tests/unit/manager/repositories/base/test_upserter.py` — a failing row in `bulk_upsert_partial` rolls back only its own savepoint while the rest persist
- [x] `tests/unit/manager/actions/test_registry_catalog.py` covers the new v2 wiring
- [x] Live E2E on the dev stack (with the #13682 surface): mixed valid/unknown exclude and include return partial `success`/`failed` payloads, rows are upserted/reset accordingly, and audit rows carry per-entity success/error with the mapped reasons

## Relation to #13682
#13682 (BA-7243/BA-7244) carries the API surface (DTO/REST/GQL/SDK/CLI) with the adapter stubbed as `NotImplementedError`; the stubs are wired to this service when this PR lands.

Resolves BA-7120


src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
```diff
@@ -446,6 +453,71 @@ async def batch_update_session_idle_check_phase(
...
 
+    async def batch_exclude_session_idle_checks(
+        self,
+        checker_id: IdleCheckerID,
+        session_ids: Sequence[SessionID],
+    ) -> None:
+        async with self._ops.write_ops() as w:
+            await self._validate_checker_and_sessions(w, checker_id, session_ids)
+            # Deduplicate: a pair repeated in one INSERT..ON CONFLICT statement is an error.
+            for id_batch in batched(dict.fromkeys(session_ids), _IDLE_CHECK_UPDATE_BATCH_SIZE):
```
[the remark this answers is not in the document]
seedspirit: Added `test_exclude_deduplicates_repeated_session_ids` in 3c268571d — the same session id passed twice succeeds and leaves a single EXCLUDED row.

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
```diff
@@ -446,6 +453,71 @@ async def batch_update_session_idle_check_phase(
...
+    ) -> None:
+        """Require the checker and every session to exist; pair rows need not.
+
+        Exclusion upserts missing pair rows and assignment-sync reconciles strays,
+        so only the referenced entities themselves are validated.
+        """
+        checker = await w.query(Querier(row_class=IdleCheckerRow, pk_value=checker_id))
+        if checker is None:
+            raise IdleCheckerNotFound(str(checker_id))
+        query = sa.select(SessionRow.id).where(SessionRow.id.in_(session_ids))
```
fregataa: Is this SQLAlchemy query statement needed with BatchQuerier?

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
```diff
@@ -446,6 +453,71 @@ async def batch_update_session_idle_check_phase(
                     )
                 )
 
+    async def batch_exclude_session_idle_checks(
+        self,
+        checker_id: IdleCheckerID,
+        session_ids: Sequence[SessionID],
+    ) -> None:
+        async with self._ops.write_ops() as w:
+            await self._validate_checker_and_sessions(w, checker_id, session_ids)
```
fregataa: This blocks any update of session data if there is any session does not exist. How about changing this into bulk action?

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
```diff
@@ -446,6 +453,71 @@ async def batch_update_session_idle_check_phase(
...
+            # Only rows exclusion wrote are reset, so re-including twice is a
+            # no-op, pairs without a row stay absent, and other phases are untouched.
+            for id_batch in batched(session_ids, _IDLE_CHECK_UPDATE_BATCH_SIZE):
+                pair_values = [(SessionId(session_id), checker_id) for session_id in id_batch]
+                await w.batch_update(
```
fregataa, line 524: Same here

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
seedspirit: Moved the filter into the querier via `SessionConditions.by_ids` (ba73a8fe9 / 6e9e11935) — the base statement is now a bare `sa.select(SessionRow.id)`.

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
seedspirit: Changed in 6e9e11935: unknown session ids are skipped instead of failing the batch. Both ops act on the sessions that exist and return the acted-upon ids (request-ordered, deduplicated — ba73a8fe9) so the service layer reports per-entity SUCCESS/ERROR, matching the v2 bulk framework's partial-success model. The API-layer counterpart lands in #13682.

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
seedspirit: Same change applied to include in 6e9e11935 — unknown sessions are skipped per entity.

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
```diff
@@ -446,6 +456,89 @@ async def batch_update_session_idle_check_phase(
...
+                    specs=[
+                        SessionIdleCheckExcludeUpserterSpec(
+                            session_id=session_id,
+                            checker_id=checker_id,
+                        )
+                        for session_id in dict.fromkeys(session_ids)
+                    ]
+                ),
+                index_elements=["session_id", "idle_checker_id"],
+            )
```
HyeockJinKim, line 485: It doesn't seem like the upserter is being handled in the repositories; this value should have already been passed in, so it looks like a poorly designed implementation.

HyeockJinKim (approved):
First, let's merge them and tidy things up together.

PR #7218, opened 2025-12-11
refactor(BA-3320): Optimize CLI startup time by deferring heavy imports

HyeockJinKim:
## Results

  | Entry Point | Before | After | Improvement |
  |-------------|--------|-------|-------------|
  | `__total__` | ~2000-2400ms | ~370-520ms | ~80% faster |
  | `mgr` | ~342ms | ~3-5ms | ~99% faster |
  | `mgr.start-server` | ~570ms | ~0.5-0.7ms | ~99% faster |
  | `app-proxy-worker` | ~574ms | ~8-9ms | ~98% faster |
  | `install` | ~160-230ms | ~45-49ms | ~70-80% faster |

resolves #7217 (BA-3320)

**Checklist:** (if applicable)

- [ ] Milestone metadata specifying the target backport version
- [ ] Mention to the original issue
- [ ] Installer updates including:
  - Fixtures for db schema changes
  - New mandatory config options
- [ ] Update of end-to-end CLI integration tests in `ai.backend.test`
- [ ] API server-client counterparts (e.g., manager API -> client SDK)
- [ ] Test case(s) to:
  - Demonstrate the difference of before/after
  - Demonstrate the flow of abstract/conceptual models with a concrete implementation
- [ ] Documentation
  - Contents in the `docs` directory
  - docstrings in public interfaces and type annotations

Issue #1925, opened 2024-02-22
Make use of the output framework consistent across all client SDK's CLI commands


achimnol:
We should use the following patterns for the Client SDK's CLI commands, so that we can use `--output=json` option to parse and verify the command execution results programmatically for [our integration test suite](https://github.com/lablup/backend.ai/tree/main/src/ai/backend/test).

The output framework (https://github.com/lablup/backend.ai/tree/main/src/ai/backend/client/output) allows using different output implementations with a unified API to render the API responses.

We should **avoid**:
- Directly using `ai.backend.client.cli.pretty.print_error()`, `ai.backend.client.cli.pretty.print_fail()` or similar stuffs
- Directly using `logging` like `log.exception("An error has ocurred.")`

We shoud use:
- `ctx.output.print_error()`, `ctx.output.print_fail()`, ...
- `ctx.output.print_item()`, `ctx.output.print_paginated_list()`, ...

Currently there are many CLI modules that DOES NOT follow the output framework:
- [ ] `ai.backend.client.cli.image`
- [ ] `ai.backend.client.cli.server_log`
- [ ] `ai.backend.client.cli.dotfile`
- [ ] `ai.backend.client.cli.vfolder`
- [ ] `ai.backend.client.cli.session.lifecycle`
- [ ] `ai.backend.client.cli.admin.etcd`
- [ ] `ai.backend.client.cli.admin.license`
- [ ] `ai.backend.client.cli.admin.manager`
- [ ] `ai.backend.client.cli.admin.resource`
- [ ] We also need to update/adapt the existing integration test suite as well.

There are some cases that does not fit very well with the output framework, so let's exclude them:
- `ai.backend.client.cli.app`
- `ai.backend.client.cli.logs`
- `ai.backend.client.cli.session.app`
- `ai.backend.client.cli.session.execute`
- `ai.backend.client.cli.session.ssh`

Well-written examples:
- https://github.com/lablup/backend.ai/blob/main/src/ai/backend/client/cli/admin/keypair.py
- https://github.com/lablup/backend.ai/blob/main/src/ai/backend/client/cli/service.py

## The output framework patterns

These are the common patterns to use, but you may adapt it as long as the proper output framework APIs.

For paginated list queries:
```python
@parent_cmdgroup.command()
@pass_ctx_obj
@click.option("--filter", "filter_", default=None, help="Set the query filter expression.")
@click.option("--order", default=None, help="Set the query ordering expression.")
[...]

PR #13369, opened 2026-08-03
feat: Accept `--output` on client subcommands, not only at the root group

### LINKED ISSUE #13367: Accept `--output` on subcommands, not only at the root group
### Main idea

Let `--output` be accepted on the subcommand that actually produces the output, not only on the root group.

```console
$ ./backend.ai --output=json service list   # works
$ ./backend.ai service list --output=json   # Error: No such option: --output
```

The root-level-only form is not a bug — it is the design decided in #240 ("using a root-level option `--output=json`") and it is exactly how Click resolves options: an option is parsed only at the position of the group that declares it. But the trailing form is the better default UX, and today it is not merely unsupported — it is undiscoverable and inconsistent.

**Why the trailing form is more appropriate**

| | root-level only (today) | accepted on the leaf command |
|---|---|---|
| `backend.ai service list -h` shows the flag | ❌ never listed | ✅ |
| Appending to a recalled shell-history line | ❌ must edit mid-line | ✅ |
| Matches peer CLIs | — | `kubectl get pods -o json`, `gh pr list --json`, `docker ps --format`, `aws s3api list-buckets --output json` |
| Flag sits next to the thing it formats | ❌ | ✅ |
| One declaration site | ✅ | ✅ if injected by the group class (see below) |

clig.dev — the guide #240 itself cites — puts the discoverability of a command's own flags ahead of parser purity. A formatting option that cannot be seen or typed where the output is produced fails that test.

This is not hypothetical churn: the design has already broken down once. `session watch` re-declares its own leaf-level `--output` with the identical `Choice(["json", "console"])` (`src/ai/backend/client/cli/session/lifecycle.py:1381`) and hand-rolls `if output == "json"` branches instead of going through the output framework, because the root-level flag was not reachable where it was needed.

**Proposal**

Accept `--output` in both positions; the leaf value wins over the root value. Do not remove the root-level option — it is the correct place for a setting that applies to a whole invocation, and removing it breaks every existing script.

Implement it once, in `ExtendedCommandGroup` (`src/ai/backend/cli/extensions.py`), so no per-command decorator churn across the ~140 commands and 273 `ctx.output.*` call sites:

- when resolving a subcommand, recursively append the `--output` param to leaf commands and merge the parsed value into `CliContextInfo.info`;
- **skip any command that already declares an `--output` / `-o` param** — `admin export {users,sessions,...}` uses `-o/--output <PATH>` for the destination file in both v1 (`src/ai/backend/client/cli/admin/export.py`) and v2 (`src/ai/backend/client/cli/v2/admin/export.py`). A blanket injection would silently break them;
- the recursion has to reach through plain `click.Group` subgroups too — `service`, `vfolder`, etc. are declared with a bare `@click.group()`, not `ExtendedCommandGroup`.

Follow-ups this unblocks:
- delete the duplicated leaf-level `--output` in `session watch` and route it through the output framework (part of #1925);
- `src/ai/backend/client/cli/main.py:18-57` currently declares a second `--output` and builds a `CLIContext` that never runs — the client group is loaded under the `_` entrypoint key, which copies only `.commands` and discards the group callback. Worth deleting alongside, together with the now-dead `case CLIContext()` branch in `pass_ctx_obj` (`src/ai/backend/client/cli/extensions.py:46`).

### Alternative ideas

- **Name the leaf option `--format` instead**, leaving `--output` root-only. Sidesteps the `admin export -o PATH` collision entirely and reads correctly (`--format json`), at the cost of two names for one concept.
- **Hoist a trailing `--output` in `argv` before Click parses.** ~5 lines, but it would rewrite `admin export users --output report.csv` into a format flag. Rejected.
- **Do nothing, and document it.** The root-level form does work and is stable; the cost is the recurring "why doesn't this work" and copy-pasted leaf-level re-declarations like `session watch`.

### Anything else?

Not applicable to the v2 CLI (`./bai`) — v2 commands always emit JSON through `print_result()` and expose no `--output` at all. If v2 ever grows a console renderer, whatever is decided here should be the shape it adopts.

Related: #240 (root-level `--output` decision), #1925 (consistent use of the output framework).

### LINKED ISSUE #13368: Remove the dead group callback in `client/cli/main.py` and the stale `session/app.py` duplicate
### Main idea

Follow-up to #13367. While tracing where `--output` is consumed, `src/ai/backend/client/cli/main.py` turned out to declare a group whose options and callback never execute. The correct fix is a **partial** removal — the module and its entry point must stay — plus deletion of one genuinely dead 376-line file.

#### What is actually dead

`src/ai/backend/client/cli/main.py:18-57` — the `--skip-sslcert-validation` / `--output` / `--version` options and the callback that builds a `CLIContext`.

Why they never run:

1. `src/ai/backend/client/BUILD:34` registers the group under the `backendai_cli_v10` entry-point key `_`.
2. `src/ai/backend/cli/loader.py:19-21` handles `_` by copying **only** `cmd_group.commands` into the unified root group. The group's own params and callback are discarded.
3. The only module that ever registered commands on this group is `src/ai/backend/client/cli/session/app.py` (`@main.command()` at lines 268 and 330) — and that module is never imported. `src/ai/backend/client/cli/session/__init__.py` is a 0-byte file, and `client/cli/__init__.py`'s `from . import app, logs, proxy` resolves to the top-level `client/cli/app.py`, not `client/cli/session/app.py`.

So the group is empty, and step 2 copies zero commands. Two independent confirmations:

```console
$ ./py -m ai.backend.client.cli --help    # no "Commands:" section at all
$ ./backend.ai apps -h                    # metavar is SESSION_ID
```

`SESSION_ID` is `client/cli/app.py:331`; `client/cli/session/app.py:330` says `NAME`. The live `apps` command comes from the top-level module.

Every other client command group is registered directly on `ai.backend.cli.main:main` — `client/cli/__init__.py` imports it as `cli_main` and decorates with `@cli_main.group(cls=LazyGroup, ...)`.

#### What must NOT be removed

**The `_` entry point and the `ai.backend.client.cli.main` module.** Loading that entry point is what imports the `ai.backend.client.cli` package, and the package `__init__.py` import side-effect is the *only* thing that registers `service`, `admin`, `vfolder`, `session`, `v2`, … onto the unified root. Drop the entry point and `backend.ai` loses every client command. The loader also type-asserts the loaded object is a `click.Group`, so the name `main` has to keep resolving to one.

So: keep a minimal group as the entry-point anchor, delete the options and the callback body.

#### Proposed changes

| Target | Action |
|---|---|
| `src/ai/backend/client/cli/session/app.py` | **Delete.** Never imported; a stale near-duplicate of `client/cli/app.py` (`diff` shows only cosmetic drift plus `sys.exit(1)` vs `sys.exit(ExitCode.FAILURE)`). |
| `src/ai/backend/client/cli/main.py:18-57` | Reduce to a bare `@click.group(cls=ExtendedCommandGroup)` with a docstring stating it exists solely as the `backendai_cli_v10` `_` anchor. |
| `src/ai/backend/client/cli/extensions.py:46` | `case CLIContext():` in `pass_ctx_obj` becomes provably dead once the callback above is gone — `client/cli/main.py:53` is the only assignment of a client `CLIContext` to `ctx.obj` anywhere in the tree. Removable, though see below. |

#### Alternative ideas

- **Restore it instead of removing it** — i.e. make `python -m ai.backend.client.cli` a working client-only CLI again by registering the client groups on it and having the unified CLI mount that group. Rejected: it means every client group gains a second registration root, and `--output` would then need resolving in two places, which is the exact confusion #13367 is trying to remove. There is no shipped artifact that needs a client-only entry point — `pex_binary(name="backendai-client")` uses `common_scie_config`'s default `entry_point="ai.backend.cli.__main__"`, so even the standalone client scie goes through the unified root.
- **Keep `case CLIContext():`** as a deliberate extension point for third-party CLI plugins that build their own `CLIContext`. `src/ai/backend/plugin/entrypoint.py:88` does allow out-of-tree `backendai_cli_v10` providers, so this is a judgement call rather than a clear-cut deletion; if kept, it should carry a comment saying why.
- **Delete `client/cli/__main__.py` too.** It currently produces a CLI with zero commands. Left out of the proposal above because it is two lines and harmless; fold it in if the anchor group gets a clearer name.

### Anything else?

Regression coverage already has a home: `tests/unit/client/cli/conftest.py` exposes a `cli_entrypoint` fixture built from `load_entry_points(allowlist={"ai.backend.client.cli"})`. A test asserting that the loaded root still exposes `service` / `admin` / `vfolder` / `app` / `apps` would pin down the import side-effect that this cleanup must preserve — that is the one way this change can break silently.

rapsealk:
## Summary

- `./backend.ai service list --output=json` used to fail with `No such option: --output`, because `--output` was declared only on the root group and Click parses an option only at the position of the group that declares it. It is now accepted on the subcommand too, and a subcommand-level value wins over the root-level one. The root-level form is unchanged.
- The option is attached inside `pass_ctx_obj` rather than on each command. That decorator runs after every `@click.option` of a command but before `@group.command()` consumes `__click_params__`, so it is the single point that sees a command's full param list — which is what lets it skip commands that already own the name (`admin export -o/--output PATH` keeps its `Path` option). Keying on `pass_ctx_obj` also confines the flag to commands that actually use the output framework, so `backend.ai mgr …` is untouched.
- Subclass-based hooks were not an option: `LazyClickMixin` rebinds every `click.Group` method to delegate to `self._impl`, so a `get_command` override on `ExtendedCommandGroup` never fires for `backend.ai service list`. Walking the tree eagerly at load time would have force-imported every module and reverted the deferred-import startup work from #7218.
- Removes the dead group callback in `client/cli/main.py` and the stale `client/cli/session/app.py` (#13368). The `_` entry point copies only `.commands` off the loaded group, and the only module registering on that group was `session/app.py`, which nothing imports — `session/__init__.py` is empty and `client/cli/__init__.py`'s `from . import app` picks up the top-level `client/cli/app.py`. The module itself stays as the entry-point anchor: loading it is what imports `ai.backend.client.cli` and registers every client group onto `ai.backend.cli.main:main`.

## Test plan

`tests/unit/client/cli/test_output_option.py`:

- [x] `--output` at root, at the subcommand, and at both (subcommand wins); default when absent
- [x] a command owning `-o/--output PATH` keeps it, and does not gain the `[json|console]` one
- [x] real `service list --help` advertises `--output [json|console]`
- [x] real `admin export users --help` still shows `--output PATH` only
- [x] the client command groups stay registered — pins the entry-point import side effect that the `main.py` cleanup must preserve

Verified against a live server:

```console
$ ./backend.ai service list --output=json      # JSON
$ ./backend.ai --output=json service list      # JSON, unchanged
$ ./backend.ai --output=json service list --output=console   # console
$ ./backend.ai admin export users -h           # -o, --output PATH, single entry
$ ./backend.ai apps -h                         # still registered
```

## Notes for the reviewer

- Titled `feat:` rather than `fix:` so it does not auto-backport to every maintained version. Tell me if it should be backported and I will add the `Backport:` trailer.
- `case CLIContext():` was dropped from `pass_ctx_obj` along with the `client/cli/main.py` callback that was its only producer — `client/cli/main.py:53` was the sole assignment of a client `CLIContext` to `ctx.obj` in the tree. Out-of-tree CLI plugins are mounted on the unified root, whose `ctx.obj` is always a `CliContextInfo`, so the branch was unreachable for them as well. Say the word if you would rather keep it as a documented extension point.
- Middle-position (`backend.ai service --output=json list`) stays unsupported; groups do not go through `pass_ctx_obj`.
- `session watch` keeps its own hand-rolled `--output` — it does not use `pass_ctx_obj`. Cleaning that one up belongs with #1925.

Resolves #13367
Resolves #13368


rapsealk:
Split into two PRs for a cleaner review, per review feedback:

- #13370 — `--output` on subcommands (resolves #13367)
- #13371 — dead `client/cli/main.py` callback + stale `session/app.py` (resolves #13368)

They are independent; either can merge first.

PR #13370, opened 2026-08-03
feat: Accept `--output` on client subcommands, not only at the root group

### LINKED ISSUE #13367: Accept `--output` on subcommands, not only at the root group

rapsealk:
## Summary

`./backend.ai service list --output=json` failed with `No such option: --output`, because `--output` was declared only on the root group and Click parses an option only at the position of the group that declares it. It is now accepted on the command itself as well, and a command-level value wins over the root-level one. The root-level form is unchanged.

## The rule

> An option that formats **this command's own result** belongs on the command. An option that configures the session or the transport for the whole invocation stays at the root.

Every consequence below is deliberate:

| | Takes `--output`? | Why |
|---|---|---|
| Root group | ✅ (unchanged) | configures the whole invocation |
| Leaf command rendering through the output framework | ✅ (new) | formats its own result |
| Any group, at any nesting depth | ❌ | produces no result to format |
| `--skip-sslcert-validation` | root only | configures the transport, not a result |
| `--version` | root only | eager, exits |

Propagation is therefore **role-based, not depth-based**. "If groups nest three or more levels, should an n-level group option propagate to n+1?" does not arise — no group takes `--output` at any depth. There are exactly two valid positions, root and leaf, and both are visible in `--help`.

## Where the option is declared

Once, as `output_option` in `src/ai/backend/client/cli/extensions.py`, applied explicitly next to `pass_ctx_obj` on each of the 149 commands that take it:

```python
@service.command()
@pass_ctx_obj
@output_option
@click.option("--filter", ...)
def list_(ctx: CLIContext, ...): ...
```

An earlier revision had `pass_ctx_obj` splice a `click.Option` into the function's `__click_params__` so no command had to change. That was rejected on review as too magical: a command accepted a flag its own source never mentioned. The explicit form uses nothing but Click's public decorator API, and `pass_ctx_obj` goes back to doing one thing — supplying the `CLIContext`.

The five `admin export` commands deliberately do not get it; they declare `-o/--output PATH` for a destination file and the two would collide. That rule, and the reason groups never take the option, is recorded next to `output_option`.

**Verified equivalent to the injecting revision**: all 630 leaf command registrations carry identical `--output` option strings, types, and full parameter lists before and after.

Explicitness costs the two guarantees injection gave for free — the option can now be omitted on a new command, or applied to one that already declares its own `--output` (Click registers both silently and leaves one dead in its long-option table). `tests/unit/client/cli/test_output_option_surface.py` walks the real tree and asserts neither happens. Both failure modes are mutation-covered: dropping one `@output_option` reports `service list`, adding a colliding one reports `admin export list-reports`.

Click has no built-in for an option accepted at more than one position — an option is parsed only where it is declared. `Command.get_params(ctx)` is the public extension point, but using it requires controlling the `cls` of every command, and the leaf groups are plain `@click.group()` while the subgroups are `LazyGroup`, whose `LazyClickMixin` rebinds every `click.Group` method to `self._impl`.

## Coverage — please read before approving

Of the **268** leaf command registrations in the v1 tree, **154 gain `--output`**, **7 already own the name** (5 × `admin export … PATH`, plus `session watch` and its `watch` alias), and **107 gain nothing**: `session ls`, `vfolder ls`, `vfolder info`, `admin manager status`, `admin resource usage-per-month`, `dotfile list`, and so on.

Those 107 do not hold a `CLIContext` at all — they do not wear `pass_ctx_obj`, and they print with `tabulate`/`print()` directly — **they ignore `--output` today at the root position too.** The gap pre-dates this PR; it is #1925's scope, and I have left the inventory there.

**The 154 are not uniform either, and the flag does not promise as much as `pass_ctx_obj` suggests.** All 154 hold a `CLIContext`, so all 154 honor `--output` for *error* output. Only **71** also render their *results* through the handler. The other 83 use `ctx.output` solely for `print_error`/`print_fail` and hand-format success output — and about 40 of those are read commands with a result worth formatting (`deployment list`, all 17 `fair-share …`, `notification channel list`, `resource-usage … list`, `scheduling-history … list`, `admin prometheus-query-definition {info,search,execute}`, …). `backend.ai deployment list --output=json` prints the same plain text it does today, exit 0.

That is not a regression — those commands ignore the root-level `--output` today for exactly the same reason — but this PR does put the flag in their own `--help`, which is where this PR argues users look. Wearing `pass_ctx_obj` buys a `CLIContext`, not a guarantee that results are rendered through it. Narrowing it further has no clean key short of converting the commands, which is #1925.

Putting the option on all 268 would still be worse: it would advertise the flag on 107 commands that hold no `CLIContext` at all and honor it in no respect, not even errors.

## What this PR does not claim

Nothing is blocked by the current behaviour. All 49 `--output=json` call sites in `src/ai/backend/test/cli_integration/` use the root-level form and work today. This is an ergonomics change:

- the flag is listed in `backend.ai service list --help`, where a user looks for it;
- it can be appended to a recalled shell-history line instead of edited into the middle;
- it matches `kubectl get pods -o json`, `aws s3api list-buckets --output json`, `docker ps --format`;
- it removes the reason a future command would hand-roll its own leaf-level `--output`, as `session watch` did (`src/ai/backend/client/cli/session/lifecycle.py:1381`).

It does **not** fix `session watch` itself: that command does not wear `pass_ctx_obj`, and its own `--output` defaults to `"console"`, so `backend.ai --output=json session watch X` still prints console output today and after this PR. It is registered twice (`session watch` and the root `watch` alias) as two distinct `Command` objects. Fixing it belongs with #1925.

## An accepted risk

This PR makes `<command> --output=json` the advertised idiom on 154 commands, and 14 commands reinterpret the identical flag as a destination `Path` with no error: `admin export {audit-logs,keypairs,projects,sessions,users}`, the seven `v2 admin export …`, and `v2 my export {keypairs,sessions}`.

```console
$ backend.ai admin export users --output=json   # writes a CSV file named `json`; exit code 0
```

The behaviour is unchanged by this PR — `admin export` has always declared `-o/--output` as a `click.Path()` at leaf position (`src/ai/backend/client/cli/admin/export.py:59-64`), so that command already writes a file named `json` today. What changes is that the leaf-position form stops being a parse error elsewhere, so users have a reason to type it.

Accepted deliberately rather than papered over. `--format` for the injected option would stop this PR advertising the colliding spelling, but it fixes nothing in `admin export` and ships two names for one concept. Renaming `admin export`'s destination option would fix the collision at the source, but it breaks `-o out.csv` for existing scripts across 14 command declarations, and those commands do wear `pass_ctx_obj` — freeing the name means they immediately gain a `[json|console]` flag they ignore, which is the failure mode the Coverage section rejects.

## Considered and rejected

- **Injecting the option from inside `pass_ctx_obj`.** Zero churn, but a command accepted a flag its source never mentioned. Replaced with the explicit decorator after review.
- **A general root-option propagation mechanism.** The rule above closes the question instead of building machinery for it. Only one other root option is even a candidate (`--skip-sslcert-validation`), and by that rule it belongs at the root.
- **A `-o` short flag.** `-o` already means a destination `PATH` on `admin export`. One letter with two silently different meanings is worse than no short flag.
- **Naming the leaf option `--format`.** Sidesteps every collision and reads well, but gives one concept two names — the opposite of the consistency being asked for.
- **Putting the option on every leaf command.** See Coverage.
- **Hoisting a trailing `--output` in `argv` before Click parses.** ~5 lines, but it would rewrite `admin export users --output report.csv` into a format flag.

## Test plan

`tests/unit/client/cli/test_output_option.py`:

- [x] `--output` at root, at the subcommand, and at both (subcommand wins); default when absent
- [x] a command owning `-o/--output PATH` keeps it, and does not gain the `[json|console]` one
- [x] real `service list --help` advertises `--output [json|console]`
- [x] real `admin export users --help` still shows `--output PATH` only

`tests/unit/client/cli/test_output_option_surface.py`:

- [x] every command wearing `pass_ctx_obj` declares `--output` exactly once, across the real tree
- [x] a new `LazyGroup` appearing unaccounted for fails the suite, so the walk's coverage cannot shrink silently

Verified against a live server:

```console
$ ./backend.ai service list --output=json                     # JSON
$ ./backend.ai --output=json service list                     # JSON, unchanged
$ ./backend.ai --output=json service list --output=console    # console
$ ./backend.ai admin export users -h                          # -o, --output PATH, single entry
```

Under the earlier injecting revision such a test would have been tautological — the correspondence held by construction of the decorator. Explicit application is what makes it meaningful. Pants reaches only 194 of 630 leaves on its own, since dependency inference cannot see `LazyGroup`'s `import_name="…"` strings, so the test imports those subtrees for real rather than declaring them in `BUILD`.

## Notes for the reviewer

- Titled `feat:` rather than `fix:` so it does not auto-backport to every maintained version. Tell me if it should be backported and I will add the `Backport:` trailer.
- Middle-position (`backend.ai service --output=json list`) stays an error, by the rule above.
- **The v2 tree (`backend.ai v2 …`, 362 leaf registrations) is out of scope**, and `backend.ai v2 domain list --output=json` still errors. v2 has no `CLIContext` and no console renderer — `print_result()` (`client/cli/v2/helpers.py:214-221`) always serialises JSON to `sys.stdout` — so attaching the flag there would advertise a `console` mode that does not exist, which is what the Coverage section rejects. It is contained work, though: 338 of the 362 commands funnel through that one function. Tracked in #13431.
- Click has no built-in for an option accepted at more than one position; an option is parsed only where it is declared. `Command.get_params(ctx)` is the public extension point, but using it requires controlling the `cls` of every command, and the leaf groups are plain `@click.group()` while the subgroups are `LazyGroup`.
- `session watch`'s duplicated leaf-level `--output` is a symptom of the same root cause; cleaning it up belongs with #1925.
- The dead second `--output` in `src/ai/backend/client/cli/main.py:11-57` (a group callback that never runs) and the correspondingly dead `case CLIContext()` branch in `pass_ctx_obj` are tracked in #13368, deliberately kept out of this PR so it carries one behaviour change.

Resolves #13367


rapsealk:
Thanks for the review. Answering the three questions in order.

## 1. Does `info["output"] = value` on its own make the output conform?

**No.** That callback only writes into the `CliContextInfo.info` dict. What turns the value into behaviour is `set_client_config()` → `get_output_handler()`, called from `pass_ctx_obj` (`src/ai/backend/client/cli/extensions.py:15-36`). So it only means anything on commands wearing that decorator — which is exactly why the option is attached there and nowhere else.

Your question exposed an inaccuracy in the PR body and the docstring, now fixed. I had written that the flag is "offered only where it is honored". Measured against the real tree, **of the 154 commands that gain `--output`, only 71 render their results through the output handler.** The other 83 use `ctx.output` solely for `print_error`/`print_fail` and print their success output with `print()`/`tabulate` directly. The accurate statement is: all 154 honor `--output` for **error** output; 71 also honor it for **results**. The rest already ignore the root-level `--output` today for the same reason, so this is not a regression — but the flag does start appearing in their own `--help`. Converting them is #1925.

## 2. Shouldn't this apply to the v2 CLI as well?

Agreed in principle — `backend.ai v2 ...` is the same CLI, and "works on v1, errors on v2" is not explainable to a user. **It is deliberately not in this PR, though.**

As you noted, v2 writes to stdout directly:

```python
# src/ai/backend/client/cli/v2/helpers.py:214-221
def print_result(data: Any) -> None:
    """Print a Pydantic model or dict as formatted JSON."""
    ...
    sys.stdout.write(json_str + "\n")
```

`grep -c pass_ctx_obj src/ai/backend/client/cli/v2/` returns **0**. So attaching the option to v2 today would parse `--output=console` and silently discard it — advertising a mode that does not exist, which is the same failure this PR rejected blanket injection to avoid.

**The work is contained, though.** Of the 362 v2 leaf command registrations, **338 funnel through that single `print_result()`.** Giving that one function a console mode covers essentially the whole v2 tree. I filed #13431 with the details; the one thing to settle there first is whether v2 grows its own console renderer or adopts the output framework that v1 uses (`ctx.output.print_item` / `print_list`). The latter gives both trees one renderer and matches where #1925 is taking v1.

## 3. The added code looks hacky — is there a Click built-in for this?

I checked, and **there is not.** An option is parsed only at the position of the command that declares it; that is Click's parser design, not an oversight. There is no "global option" or "inherit the parent group's options" feature.

The public extension point is `Command.get_params(ctx)` (click 8.1.8, `core.py:1260`) — both `make_parser` and help rendering go through it, so appending the option there would be the clean way. But using it requires controlling the `cls` of **every** command, and we do not: the leaf groups are declared with a plain `@click.group()` and the intermediate ones are `LazyGroup`. Hooking `ExtendedCommandGroup` does not reach them either, because `LazyClickMixin` rebinds every `click.Group` method to delegate to `self._impl`, so an override on `ExtendedCommandGroup` never fires on the `backend.ai service list` path.

**The hacky part of the implementation is gone, though.** Building a `click.Option` by hand and splicing it into `__click_params__` is replaced by applying the public `click.option` decorator:

```python
_output_option = click.option(
    "--output",
    type=click.Choice([OutputMode.JSON.value, OutputMode.CONSOLE.value]),
    default=None,
    expose_value=False,
    callback=_override_output_mode,
    help="Set the output style of this command's result, overriding the root-level one.",
)

# in pass_ctx_obj
params = getattr(wrapped, "__click_params__", [])
if not any("--output" in param.opts for param in params):
    wrapped = _output_option(wrapped)
```

That removes the manual list manipulation and, with it, the `Any` cast that existed only to silence the resulting attribute error — effectively a `# type: ignore` in disguise. Behaviour is unchanged (all four precedence cases verified); the only difference is that `--output` now renders first in each command's `--help` instead of last.

Reading `__click_params__` for the skip check does remain. There is nowhere else the command's parameter list exists before `@group.command()` consumes it. I did change what that check keys on, from `param.name` to the option string: a declaration like `@click.option("--output", "output_path", ...)` is named `output_path`, so it would pass the old check, get a second `--output` injected beside it, and leave one of the two dead in Click's long-option table with no error.

src/ai/backend/client/cli/extensions.py
```diff
@@ -39,7 +40,33 @@ def set_client_config(info: Mapping[str, Any]) -> CLIContext:
...
 
 
+def _override_output_mode(
+    ctx: click.Context, _param: click.Parameter, value: str | None
+) -> str | None:
+    """Let a subcommand's own ``--output`` take precedence over the root-level one."""
+    if value is not None:
+        match ctx.find_root().obj:
+            case CliContextInfo(info=info):
+                info["output"] = value
```
HyeockJinKim, line 50: Just explain it to me once

rapsealk: `info` is the root group's own `**kwargs`, boxed once at parse time (`src/ai/backend/cli/main.py:37-39`):

```python
def main(ctx, /, **kwargs):   # {"skip_sslcert_validation": ..., "output": ...}
    ctx.obj = CliContextInfo(info=kwargs)
```

That box is the only channel to `set_client_config()`, which turns it into the `CLIContext` every command actually uses (`extensions.py:19,26-31`):

```python
output = info.get("output", "console")
cli_ctx.output = get_output_handler(cli_ctx, OutputMode(output))
```

So `info["output"]` **is** the output mode — nothing else reads the flag.

The leaf option is `expose_value=False`, so the parsed value never reaches the command callback. This callback is the only place it can go, and `info` is the only place `set_client_config` looks. Ordering holds without extra care: Click runs the root group callback (which creates the box) before the subcommand is parsed, and runs param callbacks before the command body. Last write wins — that is the "subcommand beats root" precedence.

Delete the assignment and the leaf `--output` still parses and validates, then is silently discarded.

Is the key mandatory? Two answers:

| | |
|---|---|
| the write on line 50 | **mandatory** — it is the whole mechanism |
| `"output"` present in `info` | **no** — the root declares `default="console"`, so a real invocation always has it; the `.get(..., "console")` fallback only covers a synthetic `CliContextInfo(info={})` |

One gap your question exposed: the `match` had no arm for a root object that is not a `CliContextInfo`, so in that case the flag was accepted and dropped with exit 0 — the silent failure this PR otherwise avoids. Pushing a `case obj: raise RuntimeError(...)`, mirroring what `pass_ctx_obj` already does a few lines down.

PR #13371, opened 2026-08-03
refactor: Drop the dead group callback in `client/cli/main.py` and the stale `session/app.py`

### LINKED ISSUE #13368: Remove the dead group callback in `client/cli/main.py` and the stale `session/app.py` duplicate

rapsealk:
## Summary

- `client/cli/main.py` declared `--skip-sslcert-validation` / `--output` / `--version` and a callback building a `CLIContext`, none of which ever ran. `src/ai/backend/client/BUILD` registers the group under the `backendai_cli_v10` `_` key, and `ai/backend/cli/loader.py` handles `_` by copying **only** `cmd_group.commands` into the unified root — the group's own params and callback are discarded.
- The group also holds no commands. The only module registering on it was `client/cli/session/app.py`, and nothing imports that file: `client/cli/session/__init__.py` is a 0-byte file, and `client/cli/__init__.py`'s `from . import app` resolves to the top-level `client/cli/app.py`, a near-duplicate that registers on `ai.backend.cli.main:main` like every other client group. `./backend.ai apps -h` prints metavar `SESSION_ID` (`client/cli/app.py`), not `NAME` (`session/app.py`), and `./py -m ai.backend.client.cli --help` prints no `Commands:` section at all. Deleted — 376 lines.
- **The module and its entry point stay.** Loading that entry point is what imports the `ai.backend.client.cli` package, and that import side effect is the *only* thing registering `service`, `admin`, `vfolder`, `session`, `v2`, … onto the unified root. Drop it and `backend.ai` silently loses every client command. `client/cli/main.py` is reduced to a bare group documenting that it exists as the anchor.
- `case CLIContext():` in `pass_ctx_obj` goes with it: `client/cli/main.py:53` was the only assignment of a client `CLIContext` to `ctx.obj` in the tree.

## Test plan

- [x] `tests/unit/client/cli/test_entrypoint_registration.py` asserts the loaded root still exposes `service` / `admin` / `vfolder` / `session` / `app` / `apps` — the one way this change can break silently
- [x] `./backend.ai -h` still lists all 47 command groups; `./backend.ai apps -h` and `./backend.ai admin export users -h` unchanged

## Notes for the reviewer

- Dropping `case CLIContext():` is the one judgement call here. Out-of-tree CLI plugins (`plugin/entrypoint.py` does allow them) are mounted on the unified root, whose `ctx.obj` is always a `CliContextInfo`, so the branch was unreachable for them too — but say the word if you would rather keep it as a documented extension point.
- `client/cli/__main__.py` is left alone. It now yields a CLI with zero commands, which it already did before this change; folding it in felt like scope creep.
- Independent of #13367 / its PR; either can merge first.

Resolves #13368


src/ai/backend/client/cli/main.py
```diff
@@ -1,57 +1,21 @@
...
-    cli_ctx = CLIContext(
-        api_config=config,
-        output_mode=output_mode,
-    )
-    cli_ctx.output = get_output_handler(cli_ctx, output_mode)
-    ctx.obj = cli_ctx
-
-    from .pretty import show_warning
-
-    warnings.showwarning = show_warning
```
HyeockJinKim, line 57: Could you please check if it's okay to leave all of them out?

PR #13372, opened 2026-08-03
fix: look up the cuda.device slot in cuda_open get_attached_devices

daemyung-lablup:
`CUDAPlugin.get_attached_devices()` in the open-source CUDA plugin (`cuda_open`)
looked up the slot name `cuda.devices` (plural), while the plugin registers and
allocates under `cuda.device` (singular). The plural key is never present in the
allocation mapping, so the method always returned an empty list.

**Impact is low.** The enterprise CUDA plugin is a separate implementation and is
unaffected. Within `cuda_open`, the wrong lookup only degrades the *reported*
attached-device info — scheduling, allocation, and the actual GPU assignment to
containers all go through `create_alloc_map()` / `generate_docker_args()`, which
use the correct slot name. Sessions still get their GPUs; the device details are
just missing from the reported container device info.

Fix: use the singular `cuda.device` so the lookup matches the slot the plugin
actually allocates. The other accelerator plugins were audited and are consistent.

### Test

Adds `tests/unit/accelerator/cuda_open/test_get_attached_devices.py` (6 cases),
verified to fail against the pre-fix code:

| Test | Pre-fix | Post-fix |
|---|---|---|
| `test_returns_devices_allocated_under_the_cuda_device_slot` | FAIL `assert [] == ['GPU-0']` | pass |
| `test_slot_name_matches_the_allocated_slot` | FAIL `assert [] == ['GPU-1']` | pass |
| `test_returns_every_allocated_device` | FAIL `assert [] == ['GPU-0','GPU-1']` | pass |
| `test_unallocated_devices_are_excluded` | pass | pass |
| `test_empty_allocation_returns_no_devices` | pass | pass |
| `test_unknown_device_id_is_ignored` | pass | pass |

The three negative cases pass in both directions by design — a method that always
returns nothing trivially satisfies "returns nothing here." They guard against an
over-broad fix rather than detecting this bug.

`test_slot_name_matches_the_allocated_slot` feeds in `slot_types[0][0]`, the same
name `create_alloc_map()` allocates under, so the lookup and the allocation stay
pinned together even if the literal changes — that decoupling is what caused this bug.

Also verified manually against a live session.

- [ ] Backport targets: a `fix:` PR goes to every version in `.github/maintained-versions.yml` automatically.
      A `Backport:` line here names the targets instead — some of them, other ones, or `none`.

Two notes on accuracy:

The low-impact claim is based on reading the code paths: create_alloc_map() (line 359) and generate_docker_args() both use cuda.device correctly, so allocation and container assignment are unaffected — only get_attached_devices() was wrong. What I can't verify from here is every downstream consumer of that reported info; if something in the enterprise stack keys off it, the user-visible impact could be larger than "cosmetic." Worth a reviewer's eye.

I dropped the unused checklist rows (installer, e2e CLI, API counterparts, docs) since none apply. Add them back if your template requires the full list.

### DIFF
--- a/changes/13372.fix.md
+++ b/changes/13372.fix.md
@@ -1,0 +1,1 @@
+Report attached GPU information for containers using the open-source CUDA plugin, which was always empty because the plugin looked up the wrong resource slot name (`cuda.devices` instead of `cuda.device`)
--- a/docs/manager/user-api/sessions.rst
+++ b/docs/manager/user-api/sessions.rst
@@ -114,7 +114,7 @@ Example:
        "resources": {
          "cpu": "2",
          "mem": "4g",
-         "cuda.devices": "1",
+         "cuda.device": "1",
        }
      },
      "tag": "example-tag"
--- a/src/ai/backend/accelerator/cuda_open/plugin.py
+++ b/src/ai/backend/accelerator/cuda_open/plugin.py
@@ -423,8 +423,8 @@ class CUDAPlugin(AbstractComputePlugin):
         device_alloc: Mapping[SlotName, Mapping[DeviceId, Decimal]],
     ) -> Sequence[DeviceModelInfo]:
         device_ids: list[DeviceId] = []
-        if SlotName("cuda.devices") in device_alloc:
-            device_ids.extend(device_alloc[SlotName("cuda.devices")].keys())
+        if SlotName("cuda.device") in device_alloc:
+            device_ids.extend(device_alloc[SlotName("cuda.device")].keys())
         available_devices = await self.list_devices()
         attached_devices: list[DeviceModelInfo] = []
         for device in available_devices:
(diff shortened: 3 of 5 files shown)


seedspirit (commented):
Could you also post the news fragment? Please refer to the link below for instructions on how to write it.
https://github.com/lablup/backend.ai/blob/074860dc4ae6b5d7ababa84a1ad7c78b57e898e4/.github/CONTRIBUTING.md#L151


And could you also please correct the other instances where "cuda.devices" appears elsewhere? https://github.com/lablup/backend.ai/blob/074860dc4ae6b5d7ababa84a1ad7c78b57e898e4/tests/unit/common/dto/manager/v2/agent/test_response.py#L330

seedspirit (approved):

PR #13388, opened 2026-08-03
feat(BA-7155): proxy anonymous requests through the webserver

jopemachine:
## Summary

The manager serves some endpoints without authentication so a client can read config before it holds credentials. A browser could reach none of them: the webserver answers **401 before the request ever leaves it** — `manager_web_handler` runs with `is_anonymous=False`, and `web_handler_with_jwt` demands a browser session.

| Client path | How it is opened |
|---|---|
| `POST /func/v2/{entity}/public/{operation}` | route pattern → `anon_web_handler`, ahead of the catch-all |
| `POST /func/admin/gql` | `apollo_router_handler` proxies a caller without a session anonymously |

## REST: a pattern, not one concrete path

Naming `v2/app-config/public/get` would force a webserver edit for every future public endpoint. The route matches `v2/{entity}/public/{operation}` — one entity segment, one operation segment — for every proxied HTTP method (v2 already uses GET, PATCH and DELETE elsewhere, so POST alone would be a trap).

What keeps that from widening the unauthenticated surface is that `public` now *means* "no authentication required" in a v2 path, and nothing else — a guardrail in `manager/api/rest/v2/AGENTS.md`, with the warning not to reuse the segment for a domain-level notion of public (`open_to_public`, `is_public`).

## GraphQL: anonymous through the router

`apollo_router_handler` wraps `web_handler_with_jwt`: with a session the JWT path is unchanged, without one the request is proxied anonymously through `_run_proxy_request` (whose body `web_handler_with_jwt` already duplicates apart from minting the token). Anonymous subscriptions are rejected explicitly — no subgraph serving them is reachable without credentials.

`PublicQueries` becomes a federation subgraph composed into the same supergraph as `graphene` and `strawberry`, serving `ping` for now. A public field must live on `PublicQueries` alone — declaring it on `Query` too lets the router resolve it against the authenticated subgraph, which rejects an anonymous caller. #13377 currently registers `publicAppConfigs` on both; one must go when the branches meet.

## Security posture

`/func/admin/gql` is now reachable without a web session, so the manager becomes the only gate, as it already is for REST. Private fields stay unreachable anonymously because they live on authenticated subgraphs, not because the webserver blocks the request.

## Route resolution

Checked against `aiohttp.UrlDispatcher` for all six proxied methods:

| Path | Before | After |
|---|---|---|
| `/func/v2/{any-entity}/public/{op}` | authenticated | **anonymous** |
| `/func/admin/gql` without a session | 401 at the webserver | **proxied anonymously** |
| `/func/admin/gql` with a session | JWT | JWT |
| `/func/v2/app-config/my/get` | authenticated | authenticated |
| `/func/v2/app-config/public/get/extra` | authenticated | authenticated (one operation segment only) |
| `/func/v2/a/b/public/get` | authenticated | authenticated (one entity segment only) |
| everything else under `/func/` | authenticated | authenticated |

## Test plan

- [x] Route resolution verified per method for the REST pattern, its authenticated siblings, and path-extension attempts
- [x] `rover supergraph compose` succeeds with the `public` subgraph enabled; committed supergraph is composed without it
- [x] `pants fmt lint check`
- [ ] Live: restart the webserver, then call `/func/v2/app-config/public/get` with no session and confirm it reaches the manager

----
📚 Documentation preview 📚: https://sorna--13388.org.readthedocs.build/en/13388/

----
📚 Documentation preview 📚: https://sorna-ko--13388.org.readthedocs.build/ko/13388/

----
📚 Documentation preview 📚: https://sorna--13388.org.readthedocs.build/en/13388/

----
📚 Documentation preview 📚: https://sorna-ko--13388.org.readthedocs.build/ko/13388/


src/ai/backend/web/server.py
```diff
@@ -1100,6 +1102,15 @@ async def webapp_ctx(
     )
     cors.add(app.router.add_route("POST", "/func/{path:auth/signup}", anon_web_plugin_handler))
     cors.add(app.router.add_route("POST", "/func/{path:auth/signout}", manager_web_handler))
+    # `manager_web_handler` answers 401 without a browser session, before reaching the manager.
+    # `public` is a reserved anonymous segment in v2 paths (manager `api/rest/v2/AGENTS.md`).
+    cors.add(
+        app.router.add_route("POST", "/func/{path:admin/gql/strawberry/public}", anon_web_handler)
+    )
```
HyeockJinKim: Doesn't this need to be registered with the Hive router instead of being put here?

scripts/generate-graphql-schema.sh
```diff
@@ -11,7 +11,7 @@ echo "Generating GraphQL v1 schema..."
 echo "Generating GraphQL v2 schema..."
 ./backend.ai mgr api dump-gql-schema --v2 --output docs/manager/graphql-reference/v2-schema.graphql
 
-# TODO(BA-6592): re-enable once a real public field exists (public endpoint is not registered yet).
```
HyeockJinKim, line 14: I think we should include this one this time, don't you think? If we don't include it, I don't think we'll be able to verify the GQL, so please add it, check it, and then upload it.

fregataa (approved):
Let's plan to implement global rate limit

Issue #13391, opened 2026-08-03
Deployment create from the model store can leave a revision-less, permanently stuck endpoint


rapsealk:
## Goal

Make "Deploy" from the model store either produce a deployment that actually rolls out, or produce nothing at all. Today it can produce a third thing: an endpoint row with no revision, stuck in `PENDING` forever.

## Background

`DeploymentService.create_deployment` (`services/deployment/service.py:449`) creates a deployment in two independently committed steps:

1. `DeploymentController.create_deployment` writes the endpoint with `revision=None` (`sokovan/deployment/deployment_controller.py:216`) and `lifecycle_stage=PENDING` (`repositories/deployment/creators/deployment.py:163`). Its session commits at `repositories/deployment/db_source/db_source.py:319`.
2. `DeploymentController.add_deployment_revision` (`deployment_controller.py:506`) resolves the revision from the preset chain, persists it, and calls `activate_revision`, which is the only writer that sets `deploying_revision_id` and moves the endpoint to `DEPLOYING` (`db_source.py:3025`).

There is no transaction spanning the two steps and no compensating delete. Any failure in step 2 leaves a durable endpoint with `current_revision_id = NULL`, `deploying_revision_id = NULL`, `lifecycle_stage = PENDING`. No lifecycle handler targets `PENDING`, so the coordinator never advances, retries, or reaps it — the row sits in the deployment list forever while the caller only saw a GraphQL error.

The model-store Deploy path (`deployModelCardV2` → `api/adapters/model_card/adapter.py:470`) is the most exposed caller because the WebUI sends only a preset id and a resource group. Image, resource slots, runtime variant, cluster config and model definition are all resolved *inside* step 2 from preset and runtime-variant data the caller never validated, so every "required X" check in `add_revision` (`deployment_controller.py:376-481`) fires after the endpoint is already committed.

Two of those checks can fail deterministically for data that the write APIs currently accept:

- a revision preset with no required resource slot rows (`sokovan/deployment/validators/required_resource_slot_rule.py:71`)
- a runtime variant whose `default_model_definition` yields no `models[].name` — the column is `NOT NULL` but `{}` is a legal value (`models/runtime_variant/row.py:33`), so it passes the DB check and fails at `to_resolved()` (`deployment_controller.py:436`)

## Acceptance Criteria

- A failed deploy leaves no endpoint row behind, on both the v2 and the legacy create paths.
[...]

PR #13395, opened 2026-08-03
fix: Destroy the endpoint when deployment create fails to add its first revision

### LINKED ISSUE #13392: [ModelDeploy] Destroy the endpoint when deployment create fails to add its first revision
## Goal

A deployment create that fails while building its first revision must leave no endpoint row behind.

## Background

`DeploymentService.create_deployment` (`services/deployment/service.py:449-474`):

```python
deployment_info = await self._deployment_controller.create_deployment(action.creator)   # commit #1
if action.creator.model_revision is not None:
    await self._deployment_controller.add_deployment_revision(...)                       # commit #2..N
```

Step 1 commits the endpoint as `PENDING` with no revision. Step 2 opens its own sessions and can raise from at least seven places:

| Raise site | Condition |
|---|---|
| `db_source.py:2593` | requester cannot reach the model vfolder via `query_accessible_vfolders` |
| `deployment_controller.py:408` / `:413` | preset repository unavailable, or neither request nor preset supplies `runtime_variant_id` |
| `deployment_controller.py:428` / `:430` | merge chain produced no `image_id` / no `mounts` |
| `deployment_controller.py:436` | `ModelDefinitionDraft.to_resolved()` rejects a definition with no `models[].name` |
| `deployment_controller.py:466` | `RequiredResourceSlotRule` finds a required slot missing or non-positive |
| `reader.py:170-184` | malformed `deployment-config.yaml` / `model-definition.yaml` on the model vfolder |
| `deployment_controller.py:568` | `activate_revision` fails after the revision row was already written |

The orphan is permanent: `EndpointLifecycle.PENDING` appears in no handler's `target_statuses()`, so the coordinator never sees the row. The caller received a GraphQL error and the WebUI never navigated, so nobody associates the leftover row with the failed click.

`create_legacy_deployment` (`services/deployment/service.py:476-506`) has the identical two-step shape and needs the same treatment.

## Acceptance Criteria

- When `add_deployment_revision` raises, the endpoint created moments earlier is destroyed and the original error is re-raised unchanged — callers keep seeing the real cause, not a cleanup error.
- A cleanup that itself fails is logged and does not mask the original exception.
- The same holds for `create_legacy_deployment`.
- Unit coverage: `create_deployment` with a revision step that raises leaves no endpoint; the raised error is the revision error.
- The stale `check_pending` reference in `deploying_initializing.py:30` is removed.

## Notes

The narrower fix is a compensating delete in the service method — one guard covering every caller (model-card deploy, vfolder deploy, deployment create v2, legacy create). Extending the endpoint-creation transaction to span the revision write is the stronger repair but reaches into the repository layer and the RBAC creator; if we take that route it should be scoped separately.

`createModelDeploymentV2` with `initial_revision: null` must keep working — it never enters the branch above.

JIRA Issue: BA-XXXX

rapsealk:
## Summary

- `create_deployment` and `create_legacy_deployment` commit the endpoint and its first revision separately. Only `activate_revision` moves an endpoint out of `PENDING`, and no lifecycle handler targets `PENDING` — so any failure in the revision step left an endpoint the coordinator never sees and never reaps, while the caller only saw an API error.
- Both paths now wrap the revision step in `_teardown_on_revision_failure`, which marks the endpoint `DESTROYING` and re-raises the original error unchanged. A cleanup that itself fails is logged and never masks the cause.
- Teardown goes through `destroy_deployment` rather than a hard delete: `deployment_revisions.endpoint` carries no FK, so deleting the row would orphan any revision written before the failure, and the destroying handler already ends at `DESTROYED` even with no revision and no routes.
- Drops the `check_pending` reference in `DeployingInitializingHandler`'s docstring — that stage no longer exists anywhere in the tree.

`createModelDeploymentV2 { initialRevision: null }` is untouched: it never enters the guarded branch, so a deliberately revision-less deployment stays supported.

## Test plan

- [x] `pants test tests/unit/manager/services/deployment/test_deployment_service.py`
- [ ] Deploy from the model store with a preset that cannot resolve a revision; confirm no endpoint row is left behind and the API surfaces the real error

Resolves #13392
Part of #13391

### DIFF
--- a/src/ai/backend/manager/services/deployment/service.py
+++ b/src/ai/backend/manager/services/deployment/service.py
@@ -1,6 +1,8 @@
 """Deployment service for managing model deployments."""
 
 import logging
+from collections.abc import AsyncIterator
+from contextlib import asynccontextmanager
 from datetime import UTC, datetime
 from typing import cast
 from uuid import UUID
@@ -446,6 +448,36 @@ class DeploymentService:
 
     # ========== Deployment CRUD ==========
 
+    @asynccontextmanager
+    async def _teardown_on_revision_failure(
+        self, deployment_id: DeploymentID
+    ) -> AsyncIterator[None]:
+        """Tear the just-created endpoint down when its first revision fails.
+
+        Endpoint creation and revision creation commit separately, and only
+        ``activate_revision`` moves the endpoint out of ``PENDING``. A failure
+        in between would otherwise leave a ``PENDING`` endpoint with no
+        revision, which no lifecycle handler targets — invisible to the
+        coordinator and stuck forever.
+
+        Marking the endpoint ``DESTROYING`` hands it to the destroying handler,
+        which always ends at ``DESTROYED`` even with no revision and no routes.
+        A failure to tear down is logged and never masks the original error,
+        which is what the caller needs to see.
+        """
+        try:
+            yield
+        except Exception:
+            try:
+                await self._deployment_controller.destroy_deployment(deployment_id)
+            except Exception:
+                log.exception(
+                    "Failed to tear down deployment {} after its revision could not be created; "
+                    "it is left behind with no revision",
+                    deployment_id,
+                )
+            raise
+
     async def create_deployment(
         self, action: CreateDeploymentAction
     ) -> CreateDeploymentActionResult:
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/services/deployment/service.py
```diff
@@ -460,12 +492,13 @@ async def create_deployment(
...
-                requester_id=action.creator.metadata.created_user,
-                auto_activate=action.auto_activate,
-            )
+            async with self._teardown_on_revision_failure(DeploymentID(deployment_info.id)):
+                await self._deployment_controller.add_deployment_revision(
+                    deployment_id=DeploymentID(deployment_info.id),
+                    revision=action.creator.model_revision,
+                    requester_id=action.creator.metadata.created_user,
+                    auto_activate=action.auto_activate,
+                )
```
HyeockJinKim, line 501: It seems like a situation where the behavior is different from what was intended, but I'm not sure if this is the right kind of issue to report as a fix.

PR #13396, opened 2026-08-03
fix: Default the model definition name in the revision merge baseline

### LINKED ISSUE #13393: [ModelDeploy] Runtime variants created through the API can never be deployed
## Goal

The revision merge chain must always be able to produce a resolvable model definition, so a preset-driven deploy cannot fail for want of a field no layer supplies.

## Background

*Rewritten after investigation — the original framing ("validate presets and runtime variants at write time") did not survive contact with the code. See the closing note.*

`ModelConfig` requires both `name` and `model_path` (`common/config.py:323`). The lowest-priority draft in the merge chain, `_model_mount_path_default_draft` (`sokovan/deployment/revision_draft/reader.py:145`), supplied only `model_path`. Everything above it is optional, so `name` had to come from the runtime variant's `default_model_definition`, the preset, a vfolder yaml, or the request.

For a model-store deploy none of those apply except the variant baseline — and **that baseline is empty for every runtime variant created through the API**:

- `RuntimeVariantCreatorSpec.build_row()` hardcodes `default_model_definition = ModelDefinitionDraft()` (`repositories/runtime_variant/creators.py:35`).
- `RuntimeVariantUpdaterSpec` exposes only `name` and `description` (`repositories/runtime_variant/updaters.py`).
- The field appears on the response DTO and the GQL node, on no input anywhere.

Only the alembic seed (b5c6d7e8f9a0) writes real values, so the built-in variants work and every API-created one is permanently undeployable: `to_resolved()` raises at `deployment_controller.py:436`, after the endpoint row is already committed.

The fix is one layer down from the symptom: the baseline draft now supplies `name` as well as `model_path`, so the chain always resolves regardless of what the variant carries. Any higher layer still overrides it.

## Acceptance Criteria

- A revision built on a runtime variant with an empty `default_model_definition` resolves, and names the model `model`.
- A `name` supplied by the variant baseline, the preset, a vfolder yaml, or the request still wins over the default.
- No change to `model_path` resolution.

## Dropped from the original scope

**Preset resource-slot validation at write time.** #13392 makes a preset that omits a required slot fail cleanly with a good error and no leftover endpoint. Duplicating `RequiredResourceSlotRule` at preset-write time would mean giving the preset service a dependency on `resource_slot_types.required` for defence in depth of marginal value.

## Follow-up worth filing separately

`runtime_variants.default_model_definition` is not writable through any API. This change makes such variants deployable, but a custom variant still cannot ship its own baseline model definition — a genuine API gap, and a feature rather than a fix.

JIRA Issue: BA-XXXX

rapsealk:
## Summary

- `ModelConfig` requires both `name` and `model_path`, but `_model_mount_path_default_draft` — the lowest-priority layer of the revision merge chain — supplied only `model_path`. Everything above it is optional, so `name` had to come from the runtime variant baseline, the preset, a vfolder yaml, or the request.
- That baseline is empty for **every runtime variant created through the API**: `RuntimeVariantCreatorSpec.build_row()` hardcodes `ModelDefinitionDraft()`, `RuntimeVariantUpdaterSpec` exposes only `name` and `description`, and no input DTO anywhere carries the field. Only the alembic seed (b5c6d7e8f9a0) writes real values.
- Result: a model-store deploy against a preset bound to an API-created variant always failed at `to_resolved()` — after the endpoint row was already committed. The baseline now supplies both fields; any higher layer still overrides them.

Follow-up, deliberately not in scope: `runtime_variants.default_model_definition` is still not writable through any API. This makes such variants deployable, but a custom variant still cannot ship its own baseline definition.

## Test plan

- [x] `pants test tests/unit/manager/sokovan/deployment:: tests/unit/manager/data/deployment::`
- [ ] Create a runtime variant through the API, bind a preset to it, deploy from the model store

Resolves #13393
Part of #13391

### DIFF
--- a/src/ai/backend/manager/sokovan/deployment/revision_draft/reader.py
+++ b/src/ai/backend/manager/sokovan/deployment/revision_draft/reader.py
@@ -37,6 +37,10 @@ if TYPE_CHECKING:
 
 __all__ = ("RevisionDraftReader",)
 
+# Fallback identifier for the single model entry when no layer in the merge
+# chain names one. Matches the conventional single-model definition.
+_DEFAULT_MODEL_NAME = "model"
+
 
 class RevisionDraftReader:
     """Fan out the DB + storage reads that feed the revision merge chain.
@@ -146,9 +150,23 @@ class RevisionDraftReader:
         self,
         mounts: MountMetadata,
     ) -> RevisionDraft:
-        """Build the lowest-priority model_path default draft."""
+        """Build the lowest-priority ``name`` / ``model_path`` default draft.
+
+        Both fields are required by the strict :class:`ModelConfig` that
+        ``to_resolved()`` produces at the persistence boundary, so this draft
+        supplies both. Without the ``name`` default, a variant whose
+        ``default_model_definition`` is empty — which is every variant created
+        through the API, since the write path does not expose the field — makes
+        every revision on it unresolvable. Any layer above (variant baseline,
+        preset, vfolder yaml, request) overrides these.
+        """
         model_definition = ModelDefinitionDraft(
-            models=[ModelConfigDraft(model_path=mounts.model_mount_destination)]
+            models=[
+                ModelConfigDraft(
+                    name=_DEFAULT_MODEL_NAME,
+                    model_path=mounts.model_mount_destination,
+                )
+            ]
         )
         return RevisionDraft(mounts=mounts, model_definition=model_definition)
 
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/sokovan/deployment/revision_draft/reader.py
```diff
@@ -146,9 +150,23 @@ def _model_mount_path_default_draft(
...
+        """
         model_definition = ModelDefinitionDraft(
-            models=[ModelConfigDraft(model_path=mounts.model_mount_destination)]
+            models=[
+                ModelConfigDraft(
+                    name=_DEFAULT_MODEL_NAME,
+                    model_path=mounts.model_mount_destination,
+                )
+            ]
         )
```
HyeockJinKim, line 170: Values must be entered into the custom fixture, and there should be no fixed values.

PR #13397, opened 2026-08-03
fix: Guard the preset-derived replica count and strategy fallbacks

### LINKED ISSUE #13394: [ModelDeploy] Guard the preset-derived replica count and strategy fallbacks
## Goal

Close two unguarded fallbacks in `DeploymentController._apply_deployment_level_preset` that are safe only by accident.

## Background

Both are in `sokovan/deployment/deployment_controller.py`, on the path every preset-driven deploy takes.

**Null replica count.** `:773-780` reads the preset's replica count with no null check:

```python
preset_replica = preset_data.replica_count if preset_data is not None else self._DEFAULT_REPLICA_COUNT
resolved_replica_spec = ReplicaSpec(replica_count=preset_replica)
```

The `open_to_public` fallback immediately below (`:782-789`) does guard with `is not None` and falls through to the system default. The replica branch only survives because `DeploymentRevisionPresetData.replica_count` is a non-optional `int` today; the asymmetry means the intent is not enforced, and a null would reach `endpoints.replicas` as an integrity error at insert time.

**Non-exhaustive strategy match.** `_build_policy_from_preset` (`:803-824`) matches `DeploymentStrategy.ROLLING` and `BLUE_GREEN` with no fallback arm, so `strategy_spec` would be unbound if `deployment_strategy` ever became nullable or gained a member. The failure mode is an `UnboundLocalError`, not a domain error.

Neither is reachable today. Both are one-line fixes and both sit on the code path this epic is already touching.

## Acceptance Criteria

- The replica fallback chain is explicit: caller input, then preset value when present, then `_DEFAULT_REPLICA_COUNT` — matching the shape of the `open_to_public` branch beside it.
- `_build_policy_from_preset` handles every `DeploymentStrategy` member, with any unhandled case surfacing as a domain error rather than an `UnboundLocalError`.

JIRA Issue: BA-XXXX

rapsealk:
## Summary

Two fallbacks in `_apply_deployment_level_preset` that are safe only by accident today:

- The preset replica count was read with no null check, unlike the `open_to_public` branch immediately below it. It survives only because `DeploymentRevisionPresetData.replica_count` is a non-optional `int`; a null would reach the `NOT NULL` `endpoints.replicas` column as an integrity error at insert time.
- `_build_policy_from_preset` matched `ROLLING` and `BLUE_GREEN` with no fallback arm, so a third `DeploymentStrategy` member would leave `strategy_spec` unbound — an `UnboundLocalError` rather than a domain error.

Neither is reachable through today's schema. Both are on the path every preset-driven deploy takes, and the tests pin the intent so a schema change surfaces as a domain error.

## Test plan

- [x] `pants test tests/unit/manager/sokovan/deployment/test_apply_deployment_level_preset.py`

Resolves #13394
Part of #13391

### DIFF
--- a/src/ai/backend/manager/sokovan/deployment/deployment_controller.py
+++ b/src/ai/backend/manager/sokovan/deployment/deployment_controller.py
@@ -774,7 +774,7 @@ class DeploymentController:
         if resolved_replica_spec is None:
             preset_replica = (
                 preset_data.replica_count
-                if preset_data is not None
+                if preset_data is not None and preset_data.replica_count is not None
                 else self._DEFAULT_REPLICA_COUNT
             )
             resolved_replica_spec = ReplicaSpec(replica_count=preset_replica)
@@ -804,7 +804,12 @@ class DeploymentController:
     def _build_policy_from_preset(
         preset_data: DeploymentRevisionPresetData,
     ) -> DeploymentPolicyConfig:
-        """Reconstruct a DeploymentPolicyConfig from preset-stored strategy fields."""
+        """Reconstruct a DeploymentPolicyConfig from preset-stored strategy fields.
+
+        Raises:
+            InvalidAPIParameters: If the preset carries a strategy this
+                method does not know how to build a spec for.
+        """
         spec_dict = preset_data.deployment_strategy_spec
         strategy_spec: RollingUpdateSpec | BlueGreenSpec
         match preset_data.deployment_strategy:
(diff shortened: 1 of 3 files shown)


src/ai/backend/manager/sokovan/deployment/deployment_controller.py
```diff
@@ -818,6 +823,13 @@ def _build_policy_from_preset(
                 strategy_spec = (
                     BlueGreenSpec.model_validate(spec_dict) if spec_dict else BlueGreenSpec()
                 )
+            case unsupported:
+                # Without this arm ``strategy_spec`` stays unbound and the
+                # caller gets an UnboundLocalError instead of a domain error.
+                raise InvalidAPIParameters(
+                    f"Revision preset {preset_data.id} carries an unsupported "
+                    f"deployment strategy: {unsupported}"
+                )
```
HyeockJinKim, line 832: This will probably get flagged by the type checker—is it really necessary?

src/ai/backend/manager/sokovan/deployment/deployment_controller.py
```diff
@@ -774,7 +774,7 @@ async def _apply_deployment_level_preset(
         if resolved_replica_spec is None:
             preset_replica = (
                 preset_data.replica_count
-                if preset_data is not None
+                if preset_data is not None and preset_data.replica_count is not None
```
HyeockJinKim, line 777: It doesn't look like it's nullable.

Issue #13401, opened 2026-08-03
[User] Key users to the domain by id instead of by name


jopemachine:
## Objective

Replace users.domain_name, a foreign key to domains.name, with users.domain_id keyed to domains.id, so a caller that holds a user no longer has to convert a domain name into a domain id.

## Background

users stores the domain by name while nearly everything that filters by domain keys off the id: RBAC scopes, app config fragment scopes, and session/deployment scheduling all take a domain id. Every path that starts from a user therefore performs a name-to-id conversion.

That conversion is already duplicated across nine files — get_domain_id_by_name exists separately in the scheduler, domain, and model_serving layers, plus a resolve_domain_id_by_name service action. It surfaced again in BA-6556, where the merged AppConfig read has to resolve the caller's domain before it can filter fragments at the domain scope, and where the alternative considered was adding a fourth join to the authentication query on every request.

## Scope

24 source files reference UserRow.domain_name or users.c.domain_name, so this is larger than a single issue and must be split. Suggested phases, each landing on its own:

1. Add users.domain_id with an Alembic migration, backfill it from domain_name, and keep both columns in sync on write. Nothing reads the new column yet.
1. Move readers over: carry domain_id on the session, and replace the name-to-id conversions with direct reads.
1. Retire the duplicated get_domain_id_by_name helpers once nothing calls them.
1. Drop users.domain_name, or keep it as a denormalized convenience if callers still want the name without a join — decide once phase 2 shows how often the name alone is enough.

## Acceptance Criteria (phase 1)

- users carries a domain_id column with a foreign key to domains.id, backfilled for every existing row.
- The migration is idempotent, so it can be backported to a release branch (see models/alembic/README.md).
- Creating or moving a user writes both columns consistently; a test covers that they cannot diverge.
- Phases 2 to 4 are filed as their own issues and linked here.

## Related

- BA-6556 — merged AppConfig REST v2 read, where the conversion surfaced most recently
- BA-5781 — App Config API epic

JIRA Issue: BA-7158

PR #12377, opened 2026-06-23
feat(BA-6556): merged AppConfig REST v2 read API

jopemachine:
## 📚 Stacked PRs

Part of the AppConfigFragment / AppConfig stack under BEP-1052 (epic BA-5781). **Merge in order (bottom-up):**

1. ✅ #12306 — `feat(BA-6552)`: app_config_fragments DB model and Alembic migration
2. ✅ #12307 — `feat(BA-6553)`: repository layer
3. ✅ #12403 — `refactor(BA-6619)`: consolidate AppConfigScopeType into common.data
4. ✅ #12405 — `refactor(BA-6620)`: ExistsQuerier + AppConfigAllowList.exists
5. ✅ #12358 — `feat(BA-6554)`: AppConfigFragment service layer
6. ✅ #12516 — `feat(BA-6702)`: move fragment rank to the allow list with scope defaults
7. ✅ #12517 — `feat(BA-6701)`: expose allow-list rank on the v2 API surface
8. ✅ #12518 — `feat(BA-6704)`: cascade app config subtree deletion from the definition
9. ✅ #12426 — `feat(BA-6626)`: app_config_fragment bulk repository layer
10. ✅ #12401 — `feat(BA-6618)`: app_config_fragment bulk CRUD service layer
11. ✅ #12706 — `feat(BA-6810)`: AppConfigFragment visible-fragments query (repository layer)
12. ✅ #12826 — `feat(BA-6859)`: bind fragments to their RBAC scope on write (repository layer)
13. ✅ #12837 — `fix(BA-6872)`: seed `APP_CONFIG_FRAGMENT` permissions into the RBAC role fixture
14. ✅ #12839 — `chore(BA-6873)`: drop the dead APP_CONFIG RBAC entity type
15. ✅ #12359 — `feat(BA-6555)`: AppConfig merge engine + resolve service (service layer)
16. ✅ #12984 — `feat(BA-6948)`: store `app_config_fragments.scope_id` as a nullable UUID
17. ✅ #12928 — `feat(BA-6921)`: AppConfig fragment write REST v2 API (CRUD)
18. #13020 — `feat(BA-6922)`: AppConfig fragment search REST v2 API (admin + scoped)
19. 👉 **#12377 — `feat(BA-6556)`: merged AppConfig REST v2 read API** ← you are here
20. #13377 — `feat(BA-7150)`: merged AppConfig GraphQL read (my + public)
21. #13378 — `feat(BA-7151)`: merged AppConfig SDK v2
[...]

PR #13402, opened 2026-08-03
refactor(BA-7158): key users to their domain by id as well as by name

jopemachine:
## 📚 Stacked PRs

Part of the BA-7158 user domain-id stack (issue #13401). **Merge in order (bottom-up):**

1. 👉 **#13402 — `refactor(BA-7158)`: key users to their domain by id as well as by name** ← you are here
2. #13407 — `feat(BA-7158)`: take domain_id on the user create and update APIs

Part of #13401 (BA-7158)

## Summary

`users` referenced its domain only by name, while nearly everything that filters by domain keys off the id — RBAC scopes, app config fragment scopes, session and deployment scheduling. Every path starting from a user therefore had to convert the name into an id, and that conversion is already duplicated across nine files (`get_domain_id_by_name` exists separately in the scheduler, domain and model_serving layers, plus a `resolve_domain_id_by_name` service action).

This adds `users.domain_id` beside `domain_name` and backfills it. It is phase 1 of BA-7158 — readers move over separately. Accepting `domain_id` on the user create and update APIs is split out into #13407, so this PR stays the column, the migration and the join it removes.

## What's included

| Layer | Change |
|---|---|
| Migration `c1a7d3f05e28` | `users.domain_id` + FK to `domains.id` + index, backfilled from `domain_name`. |
| `UserRow` | `domain_id` alongside `domain_name`; `domain_name` marked deprecated. |
| `UserCreatorSpec` | Carries the id resolved from the name **inside the same statement**, so a new row is written with both columns and they cannot drift. |
| `UserData` (common + manager) | Carries `domain_id`, read straight off the `users` row. Required on the auth context type, since the auth middleware always fills it. |
| Auth middleware | Drops the `domains` join it needed to label a `domain_id` onto the credential query, and the same join in the impersonation lookup — the column is on `users` now. |
| `DomainRow.users` | Names the column it joins on: a second foreign key into `domains` made the relationship ambiguous. |

## Why it matters beyond the column

A caller that needs the acting user's domain id can now read it off the session. Without this, the only ways to get it are a per-request join of `domains` into the authentication query, or a name-to-id lookup at each call site — both were considered in #12377 and neither is free. That PR keeps its own lookup for now and can drop it once this lands.

`domain_name` is retained deliberately. 24 files still read it, and whether it is worth keeping denormalized is a decision for phase 4, once phase 2 shows how often the name alone is enough.

## Test plan

- [x] `tests/unit/manager/repositories/user/test_user_repository.py` — creating a user writes a `domain_id` matching the named domain (real DB)
- [x] `tests/unit/manager/api/test_impersonation.py` and the context/message-queue suites — 28 passed
- [x] `pants fmt lint check` clean across `manager`, `common`, `tests/unit/manager`, `tests/unit/common`, `tests/component` — 5306 files
- [ ] Unit suites in CI
- [ ] Apply the migration against a populated database and confirm the backfill covers every row

## Known gap

Signup (`AuthService.signup`) and `fixtures/manager/example-users.json` write only `domain_name`, so users created through those paths get `domain_id = NULL` while `UserData.domain_id` is typed as required. Worth fixing before readers rely on the column.

## Upgrade note

`UserData.domain_id` is required, so a **serialized payload produced before this field existed fails to deserialize** — `MessageMetadata.deserialize` calls `UserData.from_dict`, which now raises on a payload without the key. Messages in flight across a rolling upgrade are affected. Either drain the queue before deploying, or add a compatibility shim in `from_dict` first.

## Follow-ups (BA-7158 phases 2–4)

1. Move readers onto `domain_id`, starting with the app config merged read.
2. Retire the duplicated `get_domain_id_by_name` helpers.
3. Decide whether `users.domain_name` is dropped or kept as a convenience.

----
📚 Documentation preview 📚: https://sorna--13402.org.readthedocs.build/en/13402/

----
📚 Documentation preview 📚: https://sorna-ko--13402.org.readthedocs.build/ko/13402/


src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -551,7 +542,6 @@ def _populate_auth_result(
         col.name: user_mapping[f"user_resource_policies_{col.name}"]
         for col in user_resource_policies.c
     }
-    auth_result["user"]["domain_id"] = user_mapping["domain_id"]
     auth_result["user"]["id"] = keypair_mapping["keypairs_user_id"]  # legacy
     auth_result["is_superadmin"] = auth_result["user"]["role"] == "superadmin"
```
HyeockJinKim, line 556: Why is this missing?

HyeockJinKim (approved):

PR #13405, opened 2026-08-03
feat(BA-7159): migrate resource-group RBAC keys from name to UUID

fregataa:
## Summary
- Re-key the `sgroups_for_domains` / `sgroups_for_groups` / `sgroups_for_keypairs` mapping tables from the scaling group name FK to a `scaling_group_id` UUID FK (and `sgroups_for_domains.domain` to a `domain_id` UUID FK, as a separate revision), with symmetric downgrades.
- Rewrite `association_scopes_entities` and `permissions` rows that key resource groups by name to the canonical string form of `scaling_groups.id`, dropping duplicates and dangling rows referencing deleted scaling groups.
- Make writers and readers UUID-keyed end to end: creator/purger specs and RBAC unbinders take UUIDs, GQL legacy mutations resolve the id from the name at the API boundary, name-based public APIs resolve/join internally, and the GQL RBAC resolvers load resource groups by id.

## Test plan
- [x] `pants fmt` / `fix` / `lint --changed-since=origin/main` pass
- [x] Both migrations verified on a local DB: upgrade converts name-keyed rows (dedup + dangling drop), downgrade restores name keys symmetrically
- [x] Live server: `resource-group allow-domains --add/--remove` writes UUID-keyed mapping/RBAC rows and returns names; GQL legacy `associate/disassociate_scaling_group_with_domain` round-trips; `scaling_groups_for_domain` query works
- [x] CI `pants check` / `pants test`

Resolves BA-7159


src/ai/backend/manager/api/gql_legacy/scaling_group.py
```diff
@@ -222,13 +244,9 @@ async def batch_load_by_domain(
...
-            ScalingGroupRow.name == ScalingGroupForDomainRow.scaling_group,
-        )
-        _stmt = (
-            sa.select(ScalingGroupRow)
-            .select_from(j)
-            .where(ScalingGroupForDomainRow.domain.in_(domain_names))
-        )
+            ScalingGroupRow.id == ScalingGroupForDomainRow.scaling_group_id,
+        ).join(DomainRow, DomainRow.id == ScalingGroupForDomainRow.domain_id)
+        _stmt = sa.select(ScalingGroupRow).select_from(j).where(DomainRow.name.in_(domain_names))
```
seedspirit: Is there a reason the query would change? Previously, there was no join on the Domain Row, so I'm wondering why this has changed.

src/ai/backend/manager/models/alembic/versions/9896475bc170_rekey_resource_group_rbac_rows_to_.py
```diff
@@ -0,0 +1,273 @@
...
+        """)
+    )
+    conn.execute(
+        sa.text("""
+            DELETE FROM permissions p
+            USING scaling_groups sg
+            WHERE p.scope_type = 'resource_group'
+              AND p.scope_id = sg.name
+        """)
+    )
```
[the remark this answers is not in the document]
seedspirit: Please Checkout this comment

src/ai/backend/manager/models/scaling_group/row.py
```diff
@@ -179,9 +180,9 @@ class ScalingGroupForProjectRow(Base):  # type: ignore[misc]
     id: Mapped[uuid.UUID] = mapped_column(
         "id", GUID, primary_key=True, server_default=sa.text("uuid_generate_v4()")
     )
-    scaling_group: Mapped[str] = mapped_column(
-        "scaling_group",
-        sa.ForeignKey("scaling_groups.name", onupdate="CASCADE", ondelete="CASCADE"),
+    scaling_group_id: Mapped[ResourceGroupID] = mapped_column(
+        "scaling_group_id",
```
seedspirit: Since we're making changes anyway, how about changing the field name to `resource_group_id`?

src/ai/backend/manager/models/scaling_group/row.py
```diff
@@ -215,9 +216,9 @@ class ScalingGroupForKeypairsRow(Base):  # type: ignore[misc]
...
```
seedspirit: Same here

src/ai/backend/manager/api/gql_legacy/scaling_group.py
fregataa: because this API allows domain names only but the DB table uses domain id only. The join is required to query with domain names.

HyeockJinKim (approved):

PR #13407, opened 2026-08-04
feat(BA-7158): take domain_id on the user create and update APIs

### LINKED ISSUE #13401: [User] Key users to the domain by id instead of by name

jopemachine:
## 📚 Stacked PRs

Part of the BA-7158 user domain-id stack (issue #13401). **Merge in order (bottom-up):**

1. #13402 — `refactor(BA-7158)`: key users to their domain by id as well as by name
2. 👉 **#13407 — `feat(BA-7158)`: take domain_id on the user create and update APIs** ← you are here

Resolves #13401 (BA-7158)

## Summary

- #13402 adds `users.domain_id`, but the only way to place a user in a domain through the API was still the domain name, which the manager resolved back into an id on every call. This lets user create and update take `domain_id` directly.
- `domainName` is now marked deprecated on the API surface — `@deprecated` in the GraphQL SDL and a `Deprecated since ...` note in the REST/DTO descriptions — following the `EnqueueSessionInput.resource_group` precedent.
- Split out of #13402 so the column and migration review separately from the API change.

| Layer | Change |
|---|---|
| DTO v2 + v1 | `domain_id` on the create and update inputs. Create requires exactly one of the two; update rejects both at once, since they may name different domains. |
| GraphQL | `domainId` on `CreateUserV2Input` / `UpdateUserV2Input`; `domainName` carries `@deprecated(reason: "Use domain_id instead.")`. |
| REST v2 adapter, v1 handler | Skip the name-to-id lookup when the request already names the id. |
| `UserCreatorSpec` / `UserUpdaterSpec` | `domain_name` becomes optional; `domain_id` added. |
| `UserDBSource` | Fills in whichever column the request did not name, so a row never carries one without the other. |

Two gaps are left for a follow-up: signup (`AuthService.signup`) and `fixtures/manager/example-users.json` still write only `domain_name`, so those rows get `domain_id = NULL`. Once both fill the pair, a composite FK — `UNIQUE (id, name)` on `domains` plus `FOREIGN KEY (domain_id, domain_name) REFERENCES domains (id, name)` on `users` — can make it impossible for the two columns to name different domains; a `CHECK` cannot express this in PostgreSQL.

## Test plan

- [x] `pants fmt lint check` across `manager`, `common`, `tests/unit/manager`, `tests/unit/common` — 5142 files clean
- [x] GraphQL schema regenerated; `@deprecated` survives Apollo Federation composition into `supergraph.graphql`
- [x] `tests/unit/manager/repositories/user/test_user_updaters.py` — leaving the domain alone touches neither column
- [ ] Unit suites in CI
- [ ] Live server: create and update a user by `domain_id` and by `domain_name`, as superadmin and as a domain admin

----
📚 Documentation preview 📚: https://sorna--13407.org.readthedocs.build/en/13407/

----
📚 Documentation preview 📚: https://sorna-ko--13407.org.readthedocs.build/ko/13407/


jopemachine:
Closed as not planned

Issue #13419, opened 2026-08-04
App Proxy client session pool can close sessions with in-flight requests


rapsealk:
Found while investigating #13408.

## Summary

`ClientPool._cleanup_loop` in `src/ai/backend/common/clients/http_client/client_pool.py` evicts and closes sessions based on `last_used`, but `last_used` is stamped only when a session is *acquired* — never while a request is in flight. A response that takes longer than `cleanup_interval_seconds` therefore has its `ClientSession` closed underneath it.

Confirmed present on `main`.

## Current behavior

`load_client_session` (line 123) stamps `last_used` at acquisition time and returns the raw session:

```python
def load_client_session(self, key: ClientKey) -> aiohttp.ClientSession:
    session = self._clients.get(key, None)
    now = time.perf_counter()
    if session is not None:
        session.last_used = now
        return session.session
    ...
```

`_cleanup_loop` (line 111) then closes anything idle by that stamp:

```python
async def _cleanup_loop(self, cleanup_interval_seconds: float) -> None:
    while True:
        await asyncio.sleep(cleanup_interval_seconds)
        now = time.perf_counter()
        for key, client in list(self._clients.items()):
            if now - client.last_used > cleanup_interval_seconds:
                del self._clients[key]
                await client.session.close()
```

The pool has no visibility into request completion, so a long-running request looks identical to an idle session. `session.close()` closes the underlying connector, which tears down acquired connections as well as idle ones, so the in-flight request fails rather than being allowed to finish.

## Impact

Worst on the App Proxy worker data path, where `client_pool_cleanup_interval` defaults to 60.0 (`worker/config.py`) and streamed LLM responses routinely exceed that. The failure requires the affected route to be otherwise idle for the whole window — any concurrent request on the same `ClientKey` refreshes the stamp and masks it — so it presents as an intermittent mid-stream disconnect on low-traffic deployments with long generations, which is a difficult shape to reproduce on demand.

The same pool type is constructed in eleven other places (manager `registry.py`, `sokovan.py`, `notification_center.py`, `prometheus_client.py`, appproxy `worker/server.py`, storage `hammerspace/client.py`), all with the 600s default, so the exposure is smaller but not zero for any long-poll or streaming call among them.

## Proposal

Skip eviction for sessions whose connector still has acquired connections, so cleanup reclaims genuinely idle sessions only. The pool keeps handing out raw sessions, so no call site changes.

[...]

PR #13420, opened 2026-08-04
fix: Make the App Proxy worker backend connect timeout configurable

### LINKED ISSUE #13408: Expose backend connection pool limit for App Proxy Worker
## Summary

`HTTPBackend` in App Proxy Worker creates its backend HTTP client without
specifying aiohttp's `limit` parameter, so the default of 100 concurrent
connections per session applies. This becomes a bottleneck under high
concurrency and is not adjustable via configuration.

Confirmed present on `main` at the time of writing.

## Current Behavior

In `src/ai/backend/appproxy/worker/proxy/backend/http.py`:

```python
self.client_pool = ClientPool(
    partial(
        tcp_client_session_factory,
        timeout=client_timeout,
        auto_decompress=False,
    ),
    cleanup_interval_seconds=cleanup_interval,
)
```

`tcp_client_session_factory` (in
`src/ai/backend/common/clients/http_client/client_pool.py`) defaults to
`limit: int = 100`, which is passed to `aiohttp.TCPConnector`.

Since `limit` is not specified at the call site, each `ClientSession` created
per `ClientKey` gets a `TCPConnector` capped at 100 connections. With N
replicas (distinct endpoints), the effective ceiling is 100×N, which is still
easily exceeded under bursty load.

`client_pool_cleanup_interval` is exposed in `worker/config.py`, but there is
no corresponding option for the pool size.

## Observation

Benchmarked with `vllm bench serve` (ShareGPT dataset, 500 prompts,
`--request-rate inf`, `--seed 42`) against a vLLM model service.

**Via App Proxy (2 replicas):**
- 55 / 500 requests failed
- `aiohttp.client_exceptions.ConnectionTimeoutError: Connection timeout to
  host http://<agent>:<port>/v1/completions`, surfaced to the client as
  Gateway Timeout
- Stack trace points to
  `aiohttp/connector.py:_wait_for_available_connection`, i.e. requests
  queueing for a free connection inside the proxy's connector rather than
  being rejected by the backend

**Directly to a single vLLM replica (bypassing App Proxy), identical
workload:**
- 0 / 500 requests failed
- Per-request latency was notably worse (TPOT p99 799ms vs 58ms via proxy),
  as expected from a single replica absorbing the entire load

With 2 replicas the effective connection ceiling would be ~200, well below
the 500 concurrent requests offered. The single replica sustained the full
500 without failures despite heavier per-replica load, which suggests the
failures originate from the proxy layer's connection pool rather than
backend saturation.

## Proposal

Expose the pool size as a configurable field (e.g.
`backend_connection_pool_limit`) in `worker/config.py`, following the same
pattern as `client_pool_cleanup_interval`, and pass it through to
`tcp_client_session_factory`.

I understand running multiple workers is the intended scaling path per the
App Proxy architecture, but a single tunable for the connection pool would
let operators absorb burst traffic without changing the deployment topology.
The change looks small in scope (one config field plus the call site).

### Alternative ideas

- Set `limit=0` (unlimited) at the call site by default, on the grounds that
  a reverse proxy should not impose its own connection cap below what the
  backends can handle. This requires no new config field, but removes the
  safety ceiling entirely, so an explicit opt-in config might be safer.
- Expose `limit_per_host` alongside `limit`. Since sessions are keyed per
  endpoint (`ClientKey`), `limit` is the effective knob today, but exposing
  both would match aiohttp's connector semantics and be more future-proof.

## Environment

- Backend.AI 26.4.4 (`backend.ai-appproxy-worker 26.4.4`)
- Installed via Docker Containers method
- Model service: vLLM 0.23.0, 2 replicas on a single agent node

I'd be glad to submit a PR if this direction sounds reasonable — please let
me know if you'd prefer a different approach.

rapsealk:
Resolves part of #13408.

## Summary

`HTTPBackend` built its `aiohttp.ClientTimeout` with a hardcoded `connect=10.0`. In aiohttp that budget covers *both* waiting for a free connector slot and establishing the connection — `BaseConnector.connect()` wraps `_wait_for_available_connection()` and `_create_connection()` in a single `ceil_timeout(timeout.connect, ...)`. With the connector limit at its default of 100 per route, any burst above that ceiling made queued requests fail with `ConnectionTimeoutError` after 10s, surfacing to the client as a Gateway Timeout — what #13408 reported.

This reads the value from a new `proxy_worker.backend_connect_timeout` config field instead. The default stays `10.0`, so existing deployments behave exactly as before. Setting it to `0` leaves the pool wait unbounded (aiohttp treats `0` as no timeout), while `sock_connect=10.0` still detects an unreachable upstream.

This does not raise the connection ceiling — it only lets operators stop saturation from being fatal. Making the ceiling configurable is a separate change.

## Test plan

- [x] `pants test tests/unit/appproxy/worker/test_backend_client_timeout.py` — with `limit=1` and a slow upstream, `connect=0` lets two concurrent requests both return 200, and a finite `connect` shorter than the upstream delay raises `ConnectionTimeoutError` on the queued request (proving the failure comes from pool queueing)
- [x] `pants check` / `pants fmt` / `pants fix` / `pants lint` clean on the changed files
- [x] Reporter re-ran `vllm bench serve` (ShareGPT, 500 prompts, `--request-rate inf`, `--seed 42`) with the pool wait unbounded: failures dropped from 55/500 to 0/500 (see comment below)

https://claude.ai/code/session_01KbQfet1ukwMiqPAgHbwZt1

### DIFF
--- a/src/ai/backend/appproxy/worker/proxy/backend/http.py
+++ b/src/ai/backend/appproxy/worker/proxy/backend/http.py
@@ -40,6 +40,17 @@ HOP_ONLY_HEADERS: Final[CIMultiDict[int]] = CIMultiDict([
     ("Upgrade", 1),
 ])
 
+BACKEND_CLIENT_TIMEOUT: Final[aiohttp.ClientTimeout] = aiohttp.ClientTimeout(
+    total=None,
+    # `connect` budgets waiting for a free pool slot *and* establishing the
+    # connection, so any finite value turns pool saturation into a gateway timeout
+    # once concurrency exceeds the connector limit. Leave queueing unbounded and
+    # cap the socket connect alone, which is what detects an unreachable upstream.
+    connect=None,
+    sock_connect=10.0,
+    sock_read=None,
+)
+
 
 class HTTPBackend(BaseBackend):
     """HTTP proxy backend that keeps routes in a health-checked pool.
@@ -56,17 +67,11 @@ class HTTPBackend(BaseBackend):
     def __init__(self, routes: list[RouteInfo], *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
         self._pool = RoutePool(initial_routes=routes)
-        client_timeout = aiohttp.ClientTimeout(
-            total=None,
-            connect=10.0,
-            sock_connect=10.0,
-            sock_read=None,
-        )
         cleanup_interval = self.root_context.local_config.proxy_worker.client_pool_cleanup_interval
         self.client_pool = ClientPool(
             partial(
                 tcp_client_session_factory,
-                timeout=client_timeout,
+                timeout=BACKEND_CLIENT_TIMEOUT,
                 auto_decompress=False,  # transparently pass the response body
             ),
             cleanup_interval_seconds=cleanup_interval,
--- a/src/ai/backend/appproxy/worker/proxy/backend/http.py
+++ b/src/ai/backend/appproxy/worker/proxy/backend/http.py
@@ -40,17 +40,6 @@ HOP_ONLY_HEADERS: Final[CIMultiDict[int]] = CIMultiDict([
     ("Upgrade", 1),
 ])
 
-BACKEND_CLIENT_TIMEOUT: Final[aiohttp.ClientTimeout] = aiohttp.ClientTimeout(
-    total=None,
-    # `connect` budgets waiting for a free pool slot *and* establishing the
-    # connection, so any finite value turns pool saturation into a gateway timeout
-    # once concurrency exceeds the connector limit. Leave queueing unbounded and
-    # cap the socket connect alone, which is what detects an unreachable upstream.
-    connect=None,
-    sock_connect=10.0,
-    sock_read=None,
-)
-
 
 class HTTPBackend(BaseBackend):
     """HTTP proxy backend that keeps routes in a health-checked pool.
@@ -67,11 +56,18 @@ class HTTPBackend(BaseBackend):
     def __init__(self, routes: list[RouteInfo], *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
         self._pool = RoutePool(initial_routes=routes)
-        cleanup_interval = self.root_context.local_config.proxy_worker.client_pool_cleanup_interval
+        proxy_worker_config = self.root_context.local_config.proxy_worker
+        client_timeout = aiohttp.ClientTimeout(
+            total=None,
+            connect=proxy_worker_config.backend_connect_timeout,
+            sock_connect=10.0,
+            sock_read=None,
+        )
+        cleanup_interval = proxy_worker_config.client_pool_cleanup_interval
         self.client_pool = ClientPool(
             partial(
                 tcp_client_session_factory,
-                timeout=BACKEND_CLIENT_TIMEOUT,
+                timeout=client_timeout,
                 auto_decompress=False,  # transparently pass the response body
             ),
             cleanup_interval_seconds=cleanup_interval,
--- a/tests/unit/appproxy/worker/test_backend_client_timeout.py
+++ b/tests/unit/appproxy/worker/test_backend_client_timeout.py
@@ -1,0 +1,100 @@
+from __future__ import annotations
+
+import asyncio
+from collections.abc import AsyncIterator
+from functools import partial
+
+import aiohttp
+import pytest
+from aiohttp import web
+
+from ai.backend.appproxy.worker.proxy.backend.http import BACKEND_CLIENT_TIMEOUT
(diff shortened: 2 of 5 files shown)


Joseng8908:
Re-ran the benchmark from #13408 against a patched worker as requested.

## Setup

- Same environment as the original report: Backend.AI 26.4.4 (Docker Containers), vLLM 0.23.0 model service, 2 replicas on a single agent node
- Same workload: `vllm bench serve` (ShareGPT, 500 prompts, `--request-rate inf`, `--seed 42`)
- Patch applied by editing the installed `http.py` in the worker container to match this PR's change (`connect=None`, `sock_connect=10.0` kept), then restarting the worker
- **Note:** verification runs used `client_pool_cleanup_interval = 600.0`. With the default 60.0, the first patched run hung at 432/500 — the run now outlives the cleanup interval and hits #13419 (deterministic repro details in a comment there). This doesn't affect the validity of this fix; it removes an independent failure mode from the measurement.

## Results

| | baseline (26.4.4) | this PR |
|---|---|---|
| Failed | 55 / 500 | **0 / 500** |
| Duration (s) | 38.43 | 116.03 |
| Peak concurrent | 445 | 500 |
| TTFT mean / p99 (ms) | 3136 / 10682 | 19253 / 57804 |
| TPOT p99 (ms) | 58.4 | 548.8 |
| Output throughput (tok/s) | 2485 | 914 |

The failure count drops to 0 as expected. TTFT rises steeply because the connection ceiling itself is unchanged — the ~300 requests above the ceiling now queue at the connector instead of dying, which is precisely the intent of this fix. (Baseline throughput looks higher only because 55 requests were dropped early; total completed work is not comparable.)

Result JSON archived; happy to share.

src/ai/backend/appproxy/worker/proxy/backend/http.py
```diff
@@ -40,6 +40,17 @@
     ("Upgrade", 1),
 ])
 
+BACKEND_CLIENT_TIMEOUT: Final[aiohttp.ClientTimeout] = aiohttp.ClientTimeout(
```
seedspirit, line 43: nitpick: I'm not really sure why it needs to be defined as a constant

tests/unit/appproxy/worker/test_backend_client_timeout.py
```diff
@@ -0,0 +1,100 @@
...
+            timeout=timeout,
+            auto_decompress=False,
+            ssl=False,
+            limit=1,  # saturate the connector with the second concurrent request
+        ),
+        cleanup_interval_seconds=600,
+    )
+
+
+async def _fetch_concurrently(pool: ClientPool, url: str, count: int) -> list[int]:
```
seedspirit, line 55: For the current tests, the standard convention is to group test methods within a class and to provide fixtures as methods within that class. Please make the necessary changes accordingly.

src/ai/backend/appproxy/worker/proxy/backend/http.py
```diff
@@ -40,6 +40,17 @@
...
+    total=None,
+    # `connect` budgets waiting for a free pool slot *and* establishing the
+    # connection, so any finite value turns pool saturation into a gateway timeout
+    # once concurrency exceeds the connector limit. Leave queueing unbounded and
+    # cap the socket connect alone, which is what detects an unreachable upstream.
+    connect=None,
```
HyeockJinKim, line 49: I'm not sure if the default setting should be "None," but if there's a request for it, it seems like the right approach would be to provide it through the config.

HyeockJinKim: Nothing is being done to address this—please look into it.

src/ai/backend/appproxy/worker/proxy/backend/http.py
rapsealk: Removed in 0cb3170502 — it only existed so the test could import the production value. The timeout is built inline again and now reads `proxy_worker.backend_connect_timeout`.

tests/unit/appproxy/worker/test_backend_client_timeout.py
rapsealk: Done in 0cb3170502 — grouped into `TestBackendConnectTimeout` with the upstream fixture and the fetch helper as methods.

src/ai/backend/appproxy/worker/proxy/backend/http.py
rapsealk: Addressed in 0cb3170502: added `proxy_worker.backend_connect_timeout`. The default stays `10.0`, so nothing changes unless an operator opts in. Setting it to `0` leaves the pool wait unbounded (TOML has no null, and aiohttp treats `0` as no timeout); `sock_connect=10.0` still catches unreachable upstreams. Sorry for the delay.

PR #13421, opened 2026-08-04
feat: Expose the App Proxy backend connection pool limit as a config field

### LINKED ISSUE #13408: Expose backend connection pool limit for App Proxy Worker

rapsealk:
Resolves the remainder of #13408. Builds on #13420, which should merge first — both touch the same constructor.

## Summary

`HTTPBackend` never passed `limit` to `tcp_client_session_factory`, so aiohttp's default of 100 applied. Because `ClientPool` keys a session per `ClientKey`, and `HTTPBackend` builds that key from `route_id`, 100 is the ceiling *per route* — a two-replica model service tops out around 200 concurrent upstream connections regardless of what the backends can absorb.

This adds `proxy_worker.backend_connection_pool_limit` alongside the existing `client_pool_cleanup_interval`, and passes it through.

Two deliberate choices:

- **Default stays 100**, so upgrading changes nothing. Operators opt in to a higher ceiling.
- **Lower bound of 1.** aiohttp reads `limit=0` as unlimited, which was raised as an alternative in #13408, but the per-route keying means an unbounded ceiling scales descriptor usage with the number of live routes. Exhausting the process descriptor limit fails every route at once, which is a worse failure than the one being fixed. Operators who want more can name a number.

`limit_per_host` is deliberately not exposed. Sessions are already keyed per endpoint and `limit_per_host` is 0, so it would be a knob that does nothing today.

## Notes for the reviewer

`configs/app-proxy-worker/sample.toml` is intentionally untouched. Per `configs/AGENTS.md` it is generated from the schema rather than hand-maintained, and the field's `BackendAIConfigMeta` description carries the documentation. Worth noting separately that `scripts/release.sh` regenerates the manager, agent, storage and webserver samples but not the two appproxy ones, so that file is already drifting from its schema — out of scope here, but probably worth a follow-up.

## Test plan

- [x] `pants test tests/unit/appproxy/worker/test_config.py` passes (5 tests)
- [x] Default resolves to 100 on a config that does not mention the field
- [x] An explicit 500 round-trips
- [x] 0 and -1 are rejected with a schema validation failure
- [x] `pants fmt` / `pants fix` / `pants lint --changed-since=origin/main` clean
- [ ] Reporter confirms a raised limit recovers throughput on the `vllm bench serve` workload from #13408

## Backport

Milestoned 26.4 per request. Note that this is a `feat:` PR, so it will not backport automatically — landing it on the 26.4 LTS line needs an explicit `Backport: 26.4` trailer, which is a release decision I have deliberately left to a maintainer. #13420 is the part that backports by default and is what stops the reported Gateway Timeouts.

### DIFF
--- a/src/ai/backend/appproxy/worker/config.py
+++ b/src/ai/backend/appproxy/worker/config.py
@@ -772,6 +772,24 @@ class ProxyWorkerConfig(BaseSchema):
         ),
     ]
 
+    backend_connection_pool_limit: Annotated[
+        int,
+        Field(default=100, ge=1),
+        BackendAIConfigMeta(
+            description=(
+                "The maximum number of concurrent connections the worker opens to a single "
+                "upstream route. The worker keeps one connection pool per route, so the "
+                "worker-wide ceiling is this value multiplied by the number of live routes, "
+                "not this value alone. Raise it for bursty high-concurrency workloads such as "
+                "model serving, and budget the result against the process file-descriptor "
+                "limit, since exhausting that affects every route rather than only the "
+                "saturated one."
+            ),
+            added_version="26.8.0",
+            example=ConfigExample(local="100", prod="500"),
+        ),
+    ]
+
     announce_addr: Annotated[
         HostPortPair | None,
         Field(default=None),
--- a/changes/13421.feature.md
+++ b/changes/13421.feature.md
@@ -1,0 +1,1 @@
+Add `proxy_worker.backend_connection_pool_limit` to tune the App Proxy worker per-route backend connection ceiling.
--- a/src/ai/backend/appproxy/worker/proxy/backend/http.py
+++ b/src/ai/backend/appproxy/worker/proxy/backend/http.py
@@ -62,14 +62,16 @@ class HTTPBackend(BaseBackend):
             sock_connect=10.0,
             sock_read=None,
         )
-        cleanup_interval = self.root_context.local_config.proxy_worker.client_pool_cleanup_interval
+        proxy_worker_config = self.root_context.local_config.proxy_worker
         self.client_pool = ClientPool(
             partial(
                 tcp_client_session_factory,
                 timeout=client_timeout,
                 auto_decompress=False,  # transparently pass the response body
+                # Applies per route, since each route gets its own ClientSession.
+                limit=proxy_worker_config.backend_connection_pool_limit,
             ),
-            cleanup_interval_seconds=cleanup_interval,
+            cleanup_interval_seconds=proxy_worker_config.client_pool_cleanup_interval,
         )
 
     @override
--- a/tests/unit/appproxy/worker/test_config.py
+++ b/tests/unit/appproxy/worker/test_config.py
@@ -8,7 +8,12 @@ import pytest
 from pydantic import TypeAdapter
 
 from ai.backend.appproxy.common.types import FrontendMode
-from ai.backend.appproxy.worker.config import TraefikConfig, TraefikPortProxyConfig
+from ai.backend.appproxy.worker.config import (
+    ProxyWorkerConfig,
+    TraefikConfig,
+    TraefikPortProxyConfig,
+)
+from ai.backend.common.exception import BackendAISchemaValidationFailed
 from ai.backend.common.typed_validators import AutoDirectoryPath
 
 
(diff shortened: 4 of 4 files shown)


Joseng8908:
Verified the effect of raising the pool limit, on top of #13420's fix.

## Setup

Same environment and workload as my comment on #13420. On top of the `connect=None` patch, I added `limit=500` at the `tcp_client_session_factory` call site in the worker container — simulating `backend_connection_pool_limit = 500` by hardcoding the pass-through this PR introduces, since the config field isn't present in the installed 26.4.4. `client_pool_cleanup_interval = 600.0` as noted there.

## Results

| | baseline (26.4.4) | #13420 only (limit=100) | #13420 + limit=500 |
|---|---|---|---|
| Failed | 55 / 500 | 0 / 500 | **0 / 500** |
| Duration (s) | 38.43 | 116.03 | 100.23 |
| TTFT mean (ms) | 3136 | 19253 | **6430** |
| TTFT p99 (ms) | 10682 | 57804 | **12542** |
| TPOT p99 (ms) | 58.4 | 548.8 | 1659 |
| Output throughput (tok/s) | 2485 | 914 | 1060 |

With the ceiling above the offered concurrency, TTFT p99 drops ~4.6× and throughput recovers — the backlog moves from the proxy's connector queue into vLLM's own scheduler, where remaining TTFT reflects backend capacity rather than proxy behavior. TPOT degrades correspondingly since per-replica decode batches grow ~2.5×; that's the expected trade-off of admitting the full burst, and it's now the operator's knob to turn, which is the point of this PR.

The two changes separate cleanly: #13420 stops saturation from being fatal, this PR removes the saturation.

src/ai/backend/appproxy/worker/config.py
```diff
@@ -772,6 +772,24 @@ class ProxyWorkerConfig(BaseSchema):
...
+                "worker-wide ceiling is this value multiplied by the number of live routes, "
+                "not this value alone. Raise it for bursty high-concurrency workloads such as "
+                "model serving, and budget the result against the process file-descriptor "
+                "limit, since exhausting that affects every route rather than only the "
+                "saturated one."
+            ),
+            added_version="26.8.0",
+            example=ConfigExample(local="100", prod="500"),
+        ),
+    ]
```
HyeockJinKim, line 791: I'm not sure if the pool can grow indefinitely. + At the very least, it should be possible to opt out of it. (I'm not sure if it's correct for the default value to be the limit.)

PR #13425, opened 2026-08-04
fix: Keep client pool cleanup from closing sessions mid-response

### LINKED ISSUE #13419: App Proxy client session pool can close sessions with in-flight requests

rapsealk:
Resolves #13419. Found while reviewing #13408; independent of it.

## Summary

`ClientPool._cleanup_loop` evicted sessions whose `last_used` was older than `cleanup_interval_seconds`. That timestamp is stamped only when a session is acquired in `load_client_session` and never refreshed while a request runs, so a response outliving the interval is indistinguishable from an idle session. Closing it tears down the connector, including connections a request still holds; the in-flight read then stalls silently rather than raising.

Worst on the App Proxy worker data path, where `client_pool_cleanup_interval` defaults to 60s and streamed model-serving responses routinely run longer. It only triggers when the `ClientKey` is otherwise idle for the whole window, so it presents as an intermittent mid-stream disconnect on low-traffic deployments.

## Change

| Where | What |
|---|---|
| `common/clients/http_client/client_pool.py` | `ClientPool(keep_inflight_sessions=False)`. When on, the cleanup loop skips a session whose connector still has acquired connections and refreshes `last_used`. Off by default, so the manager, webserver, and storage pools keep time-based eviction. |
| `appproxy/worker/config.py`, `configs/app-proxy-worker/sample.toml` | `proxy_worker.client_pool_keep_inflight_sessions` (bool, default `false`, experimental, added 26.9.0). |
| `appproxy/worker/proxy/backend/http.py` | Passes the flag to the backend `ClientPool`. |
| `tests/unit/common/clients/test_client_pool.py` | Streamed read survives cleanup passes with the flag on; idle sessions are still evicted; flag off keeps time-based eviction. |

## Notes

- aiohttp exposes no public in-flight count. The guard reads `connector._acquired` (the set the connector itself consults in `_available_connections`), via `getattr` so an aiohttp change degrades to time-based eviction with a warning instead of killing the cleanup task. Verified on the pinned 3.13.x; the attribute is unchanged across 3.12–3.14. A connection stays acquired until its body reaches EOF or, for websockets, until `ws.close()`.
- On the "hung upstream never clears" question: the previous eviction did not free a hung request either, it closed the connector under a live one and left the read hanging. With `sock_read=None` a hung upstream stalls the proxy request regardless of eviction; that is a `sock_read` timeout concern, not a cleanup one.
- The other pool with the same streaming pattern (webserver → manager) is tracked in #13963.

## Test plan

- [x] `pants test tests/unit/common/clients/test_client_pool.py`
- [x] Verified with `_has_inflight_requests` neutered: the positive test fails within the read deadline, the other two pass
- [x] Every read is bounded by `READ_DEADLINE`; the flag-off test asserts `session.closed` structurally (~0.4s) rather than waiting on a stalled read
- [x] `pants fmt` / `fix` / `lint` / `check --changed-since=origin/main` clean
- [x] `tests/unit/appproxy/worker/test_config.py`, `tests/unit/common/configs/test_config_field_validation.py` pass

The test starts a stub aiohttp server under `tests/unit/`; a real socket is the only way to exercise connector acquisition. Precedent: `tests/unit/common/health_checker/checkers/test_http.py`. Happy to move it to `tests/component/` if preferred.


src/ai/backend/common/clients/http_client/client_pool.py
```diff
@@ -51,6 +51,21 @@ def tcp_client_session_factory(
...
+
+    A connection stays acquired until its response body is fully read, so this stays
+    true for the whole of a long-lived streaming response.
+    """
+    connector = session.connector
+    if connector is None:
+        return False
+    # ponytail: aiohttp exposes no public in-flight count, and `_acquired` is the
+    # set the connector itself checks. Move to refcounting inside the pool if this
+    # ever breaks on an aiohttp upgrade — that needs an API change at every caller.
```
seedspirit: Please delete comments like this

tests/unit/common/clients/test_client_pool.py
```diff
@@ -0,0 +1,106 @@
...
+        partial(tcp_client_session_factory, ssl=False),
+        cleanup_interval_seconds=CLEANUP_INTERVAL,
+    )
+    try:
+        yield client_pool
+    finally:
+        await client_pool.close()
+
+
+async def test_streaming_response_survives_a_cleanup_pass(
```
seedspirit, line 91: Please follow the convention of grouping code into classes, writing test methods within them, and providing fixtures as methods as well.

src/ai/backend/common/clients/http_client/client_pool.py
```diff
@@ -113,12 +128,19 @@ async def _cleanup_loop(self, cleanup_interval_seconds: float) -> None:
...
-                    except Exception as e:
-                        log.exception("Error closing client session: {}", e)
+                if now - client.last_used <= cleanup_interval_seconds:
+                    continue
+                if _has_inflight_requests(client.session):
+                    # `last_used` is only stamped on acquisition, so a request that
+                    # outlives the interval looks idle. Closing the session here
+                    # would tear down its connections mid-response.
+                    client.last_used = now
+                    continue
```
HyeockJinKim: This shouldn't be applied between the web and manager. + Also, in this case, if the system occasionally hangs while still connected, it seems like an issue might arise where it never clears up.

kyujin-cho:
bump

HyeockJinKim (approved):

PR #13390, opened 2026-08-03
feat(BA-7152): add polymorphic ScopeRowMixin and RBAC ops scope-row registry

fregataa:
## Summary
- Add a non-mapped `ScopeRowMixin` declaring scope schema metadata (`__scope_type__`, `scope_id_expr()`, `scope_name_expr()`) and apply it to `DomainRow`, `GroupRow`, and `UserRow`, so scope-typed rows expose a uniform query interface (SQLAlchemy polymorphic association pattern; no schema change for these tables).
- Register the scope rows in `RBACWriteOps` as a class-var registry and resolve role-name-template scope values (`ScopeTemplateValue`, now in `data/permission/scope_template.py`) with a single UNION ALL query instead of per-scope-type branch queries.
- Make `users.username` non-nullable (existing NULLs backfilled with the email), removing the email fallback from the user scope display name.

## Test plan
- [x] `alembic upgrade`/`downgrade` verified on a local DB, including backfill of a NULL-username row
- [x] Registry-driven UNION ALL SQL compilation verified against the three scope tables
- [x] CI (lint, typecheck, tests)

Resolves BA-7152

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #13428, opened 2026-08-04
feat(BA-7170): add UPSERT operation and check permission masks as a subset

HyeockJinKim:
## Summary

- Add `ActionOperationType.UPSERT`, whose `to_permission()` returns the `CREATE | UPDATE` mask. An upsert may insert or overwrite, so requiring only `UPDATE` would let a subject without `CREATE` create rows through it, and requiring only `CREATE` would let a subject without `UPDATE` overwrite existing ones — the nine upsert actions on the legacy base are split 8/1 between exactly those two wrong answers today.
- Move the three virtual-scope permission checks from any-bit (`effective & permission != NONE`) to a subset test via the new `Permission.covers()`. Any-bit is weaker than a single bit once a mask arrives; the semantics have to move before the first mask does. **No behavior change today** — the three v2 RBAC validators are the only callers and all pass a single bit, where any-bit and all-bits agree.
- `to_permission_operation()` maps `UPSERT` to `OperationType.CREATE`: that legacy axis carries one value and cannot narrow a mask, and `CREATE` is both the stronger bit and what `BulkUpsertAppConfigFragmentsAction` already declares.

The nine existing upsert actions are deliberately **not** re-declared. They all sit on the legacy base, where re-declaring would silently strengthen enforcement from `UPDATE` to `CREATE` and break roles that hold `UPDATE` without `CREATE`. That needs a role-fixture and permission-preset review, and belongs in a separate issue once those actions move to the v2 bases.

The first commit is an unrelated fix, included because CI runs these files as direct dependents of this change: #13390 made `users.username` NOT NULL and left four permission-controller test files building `UserRow` without it, so every test in them errors at fixture setup on `main` today. Reviewable separately as `2b33b6d`.

## Test plan

- [x] `CREATE`-only and `UPDATE`-only subjects are both denied an UPSERT action; a subject holding both passes (validator end-to-end, real DB)
- [x] Single-bit operations keep their previous verdicts — asserted both at the `Permission.covers()` level against every atomic bit and through the validator
- [x] Subset semantics at the db_source level, including a cap that clips one bit of the mask, on both the single-entity and bulk paths
- [x] `pants fmt / fix / lint / check` clean; `pants test --changed-since=origin/main --changed-dependents=direct` fully green (the four fixture files included)

Resolves BA-7170

### DIFF
--- a/src/ai/backend/manager/actions/types.py
+++ b/src/ai/backend/manager/actions/types.py
@@ -38,6 +38,7 @@ class ActionOperationType(enum.StrEnum):
     SEARCH = "search"
     CREATE = "create"
     UPDATE = "update"
+    UPSERT = "upsert"
     DELETE = "delete"
     PURGE = "purge"
     RESTORE = "restore"
@@ -52,6 +53,11 @@ class ActionOperationType(enum.StrEnum):
         return frozenset({cls.GET, cls.SEARCH})
 
     def to_permission_operation(self) -> OperationType:
+        """The legacy single :class:`OperationType` this operation maps to.
+
+        ``UPSERT`` narrows to ``CREATE``: this axis carries one value and cannot
+        express the ``CREATE | UPDATE`` mask, so it keeps the stronger of the two.
+        """
         match self:
             case ActionOperationType.GET:
                 return OperationType.READ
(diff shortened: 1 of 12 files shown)


src/ai/backend/manager/actions/types.py
```diff
@@ -61,6 +67,8 @@ def to_permission_operation(self) -> OperationType:
                 return OperationType.CREATE
             case ActionOperationType.UPDATE:
                 return OperationType.UPDATE
+            case ActionOperationType.UPSERT:
+                return OperationType.CREATE
```
HyeockJinKim, line 71: We plan to replace all instances of "OperationType" and then remove them.

fregataa (approved):

PR #13436, opened 2026-08-04
refactor(BA-7171): extend execute_preset with label matchers and query time

seedspirit:
## Summary
- Generalize `PrometheusClient.execute_preset` to take `Mapping[str, LabelMatcher]` filter labels so callers can use exact and regex matchers alike; the str→exact conversion moves to the call sites (preset execute service, deployment autoscale executor)
- Add an optional `time` parameter to evaluate instant queries at a specific timestamp
- No behavior change for existing callers — the rendered PromQL is identical

## Test plan
- [x] `pants check` on changed sources
- [x] `pants test tests/unit/manager/clients/prometheus/test_client.py tests/unit/manager/services/prometheus_query_preset::`

Resolves BA-7171

### DIFF
--- a/src/ai/backend/manager/clients/prometheus/client.py
+++ b/src/ai/backend/manager/clients/prometheus/client.py
@@ -1,4 +1,4 @@
-from collections.abc import Mapping, Sequence
+from collections.abc import Sequence
 from http import HTTPMethod
 from typing import Any, cast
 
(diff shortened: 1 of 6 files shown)


src/ai/backend/manager/clients/prometheus/client.py
```diff
@@ -112,22 +112,20 @@ async def execute_preset(
         self,
         *,
         query_template: str,
-        filter_labels: Mapping[str, str],
+        filter_labels: Mapping[str, LabelMatcher],
         group_labels: Sequence[str],
         time_window: str,
         time_range: QueryTimeRange | None,
+        time: str | None = None,
```
fregataa: How about adding a typed argument class and use here?

fregataa (approved):

PR #13440, opened 2026-08-04
refactor(BA-7173): type the auth context and load only what it carries

### LINKED ISSUE #13439: [Auth] Inject a typed, minimal auth context instead of dumping whole table rows onto the request
## Objective

Replace the untyped dict dump that the REST auth middleware puts on the aiohttp request with dataclasses that carry only the fields callers actually use.

## Background

_populate_auth_result() in src/ai/backend/manager/api/rest/middleware/auth.py builds request["user"] and request["keypair"] by iterating over the table columns and copying every one of them into a plain dict:

```python
"user": {
    col.name: user_mapping[f"users_{col.name}"]
    for col in users.c
    if col.name not in ("password", "description", "created_at")
},
```

That injects 23 of the 26 users columns and 15 of the 16 keypairs columns. A survey of every reader under src/ shows only these keys are ever read:

- request["user"]: role (51 sites), domain_name (34), uuid (17), email (9), sudo_session_enabled (4), resource_policy (4), id (1, a legacy alias of the keypair's user id)
- request["keypair"]: access_key (35 sites), resource_policy (22), rate_limit (1)

Consequences: the credential query selects far more than needed on every authenticated request; nothing type-checks these accesses, so a renamed or dropped column surfaces as a KeyError at runtime; and the dict silently carries fields that are none of a handler's business (totp_key, status_info, integration_id, ssh_private_key, ...).

## Acceptance Criteria

- Frozen dataclasses define the authenticated user context and the keypair context, carrying only the fields listed above; id fields use the NewTypes in ai.backend.common.identifier (UserID, DomainID, AccessKey).
- _query_cred_by_access_key selects only the columns those dataclasses need, instead of whole tables.
- _populate_auth_result builds the dataclasses and puts them on the request; the dict form is gone.
- All readers (about 180 call sites, concentrated in api/rest/vfolder/, dto/context.py and plugin/totp/) move to attribute access and are clean under the type checker.
- The legacy request["user"]["id"] alias is either dropped or expressed as a named field, not a duplicate of uuid.
- The auth suites stay green, and a case asserts that a field outside the contract is not reachable from the request context.

## Notes

- The three authenticate paths (JWT, HMAC, hook) all funnel into _populate_auth_result, so there is a single write point to change.
- UserData in ai.backend.common.data.user.types already carries user_id, role, domain_name and domain_id for the same request; check whether the user half can reuse it rather than adding a parallel type.
- If the call-site sweep proves larger than expected, split it: (1) the types plus the narrowed query behind the existing keys, (2) migrating the readers.
- Related: BA-7158 (users.domain_id) removed the domains join this query used to carry, and is where the survey came from.

JIRA Issue: BA-7173

jopemachine:
## 📚 Stacked PRs

Auth-context typing stack. **Merge in order (bottom-up):**

1. 👉 **#13440 — `refactor(BA-7173)`: type the auth context and load only what it carries** ← you are here
2. #13470 — `refactor(BA-7189)`: move the REST request handlers onto the typed auth context
3. #13468 — `refactor(BA-7188)`: move the legacy GraphQL context onto the typed auth context

Resolves #13439 (BA-7173)

## Summary

- The REST auth middleware built `request["user"]` and `request["keypair"]` by copying **every** column of the `users` and `keypairs` rows into plain dicts — 23 of 26 and 15 of 16 columns. A survey of every reader under `src/` shows 8 user fields and 4 keypair fields are ever read, so each authenticated request paid for a wide select and carried fields no handler has any business seeing (`totp_key`, `status_info`, `integration_id`, `ssh_private_key`, …), with nothing type-checking the accesses.
- `AuthenticatedUser` and `AuthenticatedKeypair` now define that contract and the credential query loads exactly those columns through the ORM. To keep this PR to the middleware, it serializes the two dataclasses back into `request["user"]` and `request["keypair"]`, so no handler moves yet — #13470 removes that and migrates the readers.

| Layer | Change |
|---|---|
| `data/auth/types.py` | `AuthenticatedUser`, `AuthenticatedKeypair` — frozen, `UserID` / `DomainID` / `AccessKey` / `SecretKey` NewTypes for the ids, the existing `UserResourcePolicyData` / `KeyPairResourcePolicyData` for the policies |
| `_query_auth_context_by_access_key` | ORM query with `load_only` for the columns the contract carries and `joinedload` for the two policy rows, replacing the Core select over whole tables and its `LABEL_STYLE_TABLENAME_PLUS_COL` string lookups. Returns `_AuthContext`; the rows never leave the session |
| Auth flows (JWT / HMAC / hook) | Each returns the context it authenticated instead of mutating the request. The middleware writes the request state once, next to where it clears it |

Typing the fields surfaced nullability the `Mapping[str, Any]` hid. `users.domain_name` and `users.domain_id` are NOT NULL as of #13729, so the contract declares them non-optional and carries no runtime guard for them; an `allowed_client_ip` entry that did not parse into a network no longer takes part in the match.

## Test plan

- [x] `pants fmt lint check` across `manager`, `common`, `tests/unit/manager`, `tests/component` — clean
- [ ] Live server: authenticate over HMAC, JWT and the hook path, and exercise a vfolder call that reads the keypair resource policy

## Follow-ups

- The policy dataclasses are serialized with `dataclasses.asdict` at the ~24 call sites that hand them to code still typed `Mapping[str, Any]` (`manager/utils.py`, the vfolder host-permission chain, ~50 signatures). Typing that chain is its own change; the key sets are identical today, verified against both policy tables.
- `request` still carries `is_authorized` / `is_admin` / `is_superadmin` alongside the two contexts, and `is_admin` / `is_superadmin` duplicate what the contexts already say. Folding the five keys into one context object is worth doing separately (26 reader sites).
- The same identity exists a third time as the `UserData` the middleware pushes into contextvars.
- The readers move in #13470 (BA-7189) and #13468 (BA-7188); until then the middleware serializes the contexts back into the same request keys.

### DIFF
--- a/src/ai/backend/manager/api/rest/middleware/auth.py
+++ b/src/ai/backend/manager/api/rest/middleware/auth.py
@@ -58,7 +58,7 @@ from ai.backend.manager.errors.auth import (
     InvalidClientIPConfig,
     UserNotFound,
 )
-from ai.backend.manager.errors.common import GenericForbidden, InternalServerError, RejectedByHook
+from ai.backend.manager.errors.common import GenericForbidden, RejectedByHook
 from ai.backend.manager.models.keypair import KeyPairRow
 from ai.backend.manager.models.user import UserRow
 from ai.backend.manager.models.utils import execute_with_retry
@@ -644,19 +644,6 @@ async def _query_auth_context_by_access_key(
         )
         if user_row is None:
             return None
-        if user_row.domain_name is None or user_row.domain_id is None:
-            missing_fields = [
-                name
-                for name, value in (
-                    ("domain_name", user_row.domain_name),
-                    ("domain_id", user_row.domain_id),
-                )
-                if value is None
-            ]
-            raise InternalServerError(
-                f"The authenticated user has no {', '.join(missing_fields)} "
-                f"(user_id={user_row.uuid})"
-            )
 
         return _AuthContext(
             user=AuthenticatedUser(
--- a/src/ai/backend/manager/api/rest/middleware/auth.py
+++ b/src/ai/backend/manager/api/rest/middleware/auth.py
@@ -606,7 +606,7 @@ async def _query_auth_context_by_access_key(
 
     Only the columns that context carries are loaded, and the rows stay inside this session.
     """
-    async with db.begin_readonly_session() as sess:
+    async with db.begin_readonly_session_read_committed() as sess:
         keypair_row = await sess.scalar(
             sa.select(KeyPairRow)
             .options(
--- a/src/ai/backend/manager/api/rest/middleware/auth.py
+++ b/src/ai/backend/manager/api/rest/middleware/auth.py
@@ -672,7 +672,7 @@ async def _query_auth_context_by_access_key(
             ),
             keypair=AuthenticatedKeypair(
                 access_key=AccessKey(keypair_row.access_key),
-                secret_key=SecretKey(keypair_row.secret_key) if keypair_row.secret_key else None,
+                secret_key=SecretKey(keypair_row.secret_key),
                 is_admin=bool(keypair_row.is_admin),
                 rate_limit=keypair_row.rate_limit,
                 resource_policy=keypair_row.resource_policy_row.to_dataclass(),
@@ -701,9 +701,6 @@ async def _authenticate_via_jwt(
 
         if context is None:
             raise AuthorizationFailed("Access key not found in database")
-        if context.keypair.secret_key is None:
-            raise InternalServerError(f"The keypair has no secret key (access_key={access_key})")
-
         jwt_validator.validate_token(jwt_token, context.keypair.secret_key)
 
         log.trace("JWT authentication succeeded for access_key={}", access_key)
@@ -736,9 +733,6 @@ async def _authenticate_via_hmac(
 
     if context is None:
         raise AuthorizationFailed("Access key not found in HMAC")
-    if context.keypair.secret_key is None:
-        raise InternalServerError(f"The keypair has no secret key (access_key={access_key})")
-
     my_signature = await sign_request(sign_method, request, context.keypair.secret_key)
 
     if not secrets.compare_digest(my_signature, signature):
--- a/src/ai/backend/manager/api/rest/middleware/auth.py
+++ b/src/ai/backend/manager/api/rest/middleware/auth.py
@@ -36,7 +36,7 @@ from aiohttp import web
 from aiohttp.typedefs import Handler, Middleware
 from dateutil.parser import parse as dtparse
 from dateutil.tz import tzutc
-from sqlalchemy.orm import joinedload, load_only
+from sqlalchemy.orm import load_only
 
 from ai.backend.common.contexts.client_ip import with_client_ip
 from ai.backend.common.contexts.user import with_triggered_user, with_user
@@ -60,6 +60,10 @@ from ai.backend.manager.errors.auth import (
 )
 from ai.backend.manager.errors.common import GenericForbidden, RejectedByHook
 from ai.backend.manager.models.keypair import KeyPairRow
+from ai.backend.manager.models.resource_policy.row import (
+    KeyPairResourcePolicyRow,
+    UserResourcePolicyRow,
+)
 from ai.backend.manager.models.user import UserRow
 from ai.backend.manager.models.utils import execute_with_retry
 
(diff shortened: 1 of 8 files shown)


src/ai/backend/manager/services/user/service.py
```diff
@@ -236,6 +238,10 @@ async def purge_user(self, action: PurgeUserAction) -> PurgeUserActionResult:
 
         # Handle endpoint ownership delegation
         if action.delegate_endpoint_ownership.optional_value():
+            if action.user_info_ctx.main_access_key is None:
+                raise UserPurgeFailure(
+                    "Cannot delegate endpoint ownership to a user without a main access key.",
+                )
```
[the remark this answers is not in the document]
jopemachine: Fixed in 591bb8e304. The delegation target is now validated before any mutation runs, so a missing main access key can no longer leave shared vfolders migrated against a failed purge — both purge paths hoist the check above `migrate_shared_vfolders()`.

src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -855,14 +854,34 @@ async def _middleware(request: web.Request, handler: Handler) -> web.StreamRespo
...
+            context = await _authenticate_via_jwt(db, jwt_validator, valkey_stat, jwt_token)
         elif auth_header:
-            await _authenticate_via_hmac(request, db, valkey_stat)
+            context = await _authenticate_via_hmac(request, db, valkey_stat)
         else:
-            await _authenticate_via_hook(request, db, valkey_stat, hook_plugin_ctx)
+            context = await _authenticate_via_hook(request, db, valkey_stat, hook_plugin_ctx)
+
+        authenticated_user: UserData | None = None
+        if context is not None:
```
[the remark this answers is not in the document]
jopemachine: Fixed in 7ef0dd40f4, before this review landed. `test_impersonation.py` no longer imports `_authenticated_user`; it builds the caller's `UserData` directly, the way the middleware now does.

src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -584,97 +590,82 @@ def _set_unauthenticated_state(request: web.Request) -> None:
...
-        query = (
-            sa.select(keypairs, keypair_resource_policies)
-            .set_label_style(sa.LABEL_STYLE_TABLENAME_PLUS_COL)
-            .select_from(j)
+) -> _AuthContext | None:
+    """Resolve an access key into the context an authenticated request carries.
+
+    Only the columns that context carries are loaded, and the rows stay inside this session.
+    """
+    async with db.begin_readonly_session() as sess:
```
fregataa: Let's use `read_committed` session

src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -584,97 +590,82 @@ def _set_unauthenticated_state(request: web.Request) -> None:
...
+                    load_only(
+                        UserRow.uuid,
+                        UserRow.email,
+                        UserRow.role,
+                        UserRow.domain_name,
+                        UserRow.domain_id,
+                        UserRow.sudo_session_enabled,
+                        UserRow.allowed_client_ip,
+                    ),
+                    joinedload(UserRow.resource_policy_row),
```
fregataa: Let's use join or multiple queries rather than using SQLAlchemy orm APIs here

src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -584,97 +594,87 @@ def _set_unauthenticated_state(request: web.Request) -> None:
     request["user"] = None
 
 
-async def _query_cred_by_access_key(
+@dataclass(frozen=True)
+class _AuthContext:
+    """What an authenticated request carries about its caller."""
+
+    user: AuthenticatedUser
+    keypair: AuthenticatedKeypair
```
HyeockJinKim, line 602: The design of the data class seems off. Since it includes not only user information and key pair information but also resource policy information, please separate the user information, key pair information, and resource policy into distinct groups for management. (This is also to account for future database migrations.)

src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -584,97 +594,87 @@ def _set_unauthenticated_state(request: web.Request) -> None:
...
+                    UserRow.domain_id,
+                    UserRow.sudo_session_enabled,
+                    UserRow.allowed_client_ip,
+                ),
+            )
             .where(
-                (keypairs.c.access_key == access_key) & (keypairs.c.is_active.is_(True)),
+                (KeyPairRow.access_key == access_key) & (KeyPairRow.is_active.is_(True)),
             )
         )
```
HyeockJinKim, line 645: Could you ensure that the search is based on the default keypair, and if that doesn't exist, select and apply the first keypair based on the order? (This ordering method must work consistently across all queries.)
Also, it might be better to run separate queries for reading user/keypair information and for retrieving resource policies.

src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -973,14 +952,35 @@ async def _middleware(request: web.Request, handler: Handler) -> web.StreamRespo
...
+            })
+            authenticated_user = UserData(
+                user_id=context.user.uuid,
+                is_authorized=True,
+                is_admin=context.keypair.is_admin,
+                is_superadmin=is_superadmin,
+                role=context.user.role,
+                domain_name=context.user.domain_name,
+                domain_id=context.user.domain_id,
+            )
```
HyeockJinKim, line 975: It looks like you'll be working on this part next, so please make sure to maintain consistency here as well and clearly distinguish between user information (such as admin status) and the resource policy section.

src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -708,39 +706,37 @@ async def _authenticate_via_hmac(
...
-        return
+        return None
 
     sign_method, access_key, signature = params
 
-    user_row, keypair_row = await execute_with_retry(
-        functools.partial(_query_cred_by_access_key, db, access_key)
+    context = await execute_with_retry(
+        functools.partial(_query_auth_context_by_access_key, db, access_key)
     )
```
HyeockJinKim: Please remove `execute_with_retry`.

HyeockJinKim (approved):
This project seems like it's going to get way too big, so let's merge it for now, and please outline the next steps.

PR #13441, opened 2026-08-04
feat(BA-7172): add filter and group labels to utilization checker spec

seedspirit:
## Summary
- `UtilizationThresholdEntry` gains `filter_labels` / `group_labels`, so each utilization idle checker can parameterize its Prometheus preset query (e.g. pin `container_metric_name="cpu_util"`) instead of aggregating across all metric labels; defaults (`{}` / `["session_id"]`) keep existing stored specs behaving as before
- The fields propagate through the v2 DTOs, GraphQL types (schema-level defaults: `groupLabels: [String!]! = ["session_id"]`), and the idle-checker adapter
- `MetricRepository` now batches per `(preset, labels)` via `SessionUtilizationQuery`, validates spec labels against the preset's allow-lists at query time (invalid queries are skipped with an error log), scopes the query to the evaluated session batch when grouped by `session_id` without overriding a user-set filter, and folds multiple series per session with `max`
- Remove `PrometheusClient.fetch_session_utilization` in favor of the generalized `execute_preset` (BA-7171)

## Test plan
- [x] `pants fmt/fix/lint/check --changed-since` clean
- [x] `pants test` — client rendering, repository batching/scoping/validation/max-fold, checker batching, adapter round-trip

Resolves BA-7172

----
📚 Documentation preview 📚: https://sorna--13441.org.readthedocs.build/en/13441/

----
📚 Documentation preview 📚: https://sorna-ko--13441.org.readthedocs.build/ko/13441/

### DIFF
--- a/src/ai/backend/common/data/idle_checker/types.py
+++ b/src/ai/backend/common/data/idle_checker/types.py
@@ -2,12 +2,15 @@ from __future__ import annotations
 
 import enum
 from decimal import Decimal
-from typing import Self
+from typing import Final, Self
 
 from pydantic import Field, model_validator
 
 from ai.backend.common.identifier.prometheus_query_preset import PrometheusQueryPresetID
 from ai.backend.common.types import BackendAISchema
+
+# Metric label carrying the session UUID; grouping by it enables per-session mapping.
+SESSION_ID_LABEL: Final = "session_id"
 
 
 class CheckerType(enum.StrEnum):
@@ -34,6 +37,20 @@ class UtilizationThresholdEntry(BackendAISchema):
     )
     threshold: Decimal = Field(
         description="Underutilization threshold compared against the preset's query result.",
+    )
+    filter_labels: dict[str, str] = Field(
+        default_factory=dict,
+        description="Label filters injected into the preset's {labels} placeholder.",
+    )
+    group_labels: list[str] = Field(
+        default_factory=lambda: [SESSION_ID_LABEL],
+        description=(
+            "Labels injected into the preset's {group_by} placeholder. "
+            "Must include 'session_id' for per-session values to be mapped. "
+            "When 'session_id' is grouped, the checker adds a session_id filter that "
+            "limits the query to the sessions being evaluated; a user-provided "
+            "'session_id' entry in filter_labels takes precedence over it."
+        ),
     )
 
 
--- a/src/ai/backend/common/data/idle_checker/types.py
+++ b/src/ai/backend/common/data/idle_checker/types.py
@@ -29,6 +29,13 @@ class IdleCheckPhase(enum.StrEnum):
     IDLE_EXPIRED = "idle_expired"
 
 
+class MetricLabel(BackendAISchema):
+    """Single metric label key-value pair."""
+
+    key: str = Field(description="Label key.")
+    value: str = Field(description="Label value.")
+
+
 class UtilizationThresholdEntry(BackendAISchema):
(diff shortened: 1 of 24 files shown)


src/ai/backend/common/data/idle_checker/types.py
```diff
@@ -35,6 +38,20 @@ class UtilizationThresholdEntry(BackendAISchema):
     threshold: Decimal = Field(
         description="Underutilization threshold compared against the preset's query result.",
     )
+    filter_labels: dict[str, str] = Field(
+        default_factory=dict,
+        description="Label filters injected into the preset's {labels} placeholder.",
+    )
```
fregataa: `filter_labels` fields in DTOs are `list[MetricLabelEntryInfo]` but it is dict[str, str] here. is it intentional?

src/ai/backend/manager/api/adapters/idle_checker/adapter.py
```diff
@@ -231,6 +232,11 @@ def _spec_to_info(spec: IdleCheckerSpec) -> IdleCheckerSpecInfo:
                         threshold=UtilizationThresholdInfo(
                             preset_id=utilization.threshold.preset_id,
                             threshold=utilization.threshold.threshold,
+                            filter_labels=[
+                                MetricLabelEntryInfo(key=key, value=value)
+                                for key, value in utilization.threshold.filter_labels.items()
+                            ],
```
fregataa, line 239: What if the `utilization.threshold.filter_labels` has multiple same keys?

src/ai/backend/manager/repositories/metric/repository.py
```diff
@@ -90,59 +92,99 @@ async def query_container_live_stats(
...
-                query_template=preset.query_template,
-                time_window=preset.time_window or "",
-                session_ids=session_ids,
-                evaluation_time=evaluation_time.isoformat(),
+            response = await self._prometheus_client.execute_preset(
+                MetricPreset(
+                    template=preset.query_template,
+                    labels=filter_labels,
+                    group_by=set(query.group_labels),
+                    window=preset.time_window or "",
```
fregataa: window isn't applied to query when it is `""`?

src/ai/backend/manager/repositories/metric/repository.py
```diff
@@ -90,59 +92,99 @@ async def query_container_live_stats(
...
+        preset: PrometheusQueryPresetData,
+        query: SessionUtilizationQuery,
+    ) -> set[str]:
+        """Spec labels the preset does not declare as allowed (empty allow-list allows all)."""
+        invalid: set[str] = set()
+        if preset.filter_labels:
+            invalid |= {name for name, _ in query.filter_labels} - set(preset.filter_labels)
+        if preset.group_labels:
+            invalid |= set(query.group_labels) - set(preset.group_labels)
+        return invalid
```
fregataa, line 155: Q) Are filter/group labels validated when they are saved into utilization spec?

src/ai/backend/manager/api/adapters/idle_checker/adapter.py
seedspirit: I added validation to the API layer to prevent duplicate keys from being submitted.

src/ai/backend/manager/repositories/metric/repository.py
seedspirit: I added validation at the save point.

fregataa (approved):

PR #13444, opened 2026-08-04
feat(BA-7174): unify RBAC member ops with virtual-scope wiring and fix purge leaks

fregataa:
## Summary
- `add_bulk_members(_partial)` now wires every member fully: membership in the containing scope's virtual scope, the legacy scope association, and the containing scope's binding into the member's own virtual scope (`permission_cap` applies to both edges; the reverse binding is never written). Every entity doubles as a scope backed by its own virtual scope, so the separate scope-member ops are no longer needed. The three writes live in dedicated helpers shared by the bulk and partial variants.
- `remove_bulk_members` (née `remove_entity_members`) deletes the same three edges — including the binding previously leaked in the member's virtual scope — and never raises for missing virtual scopes, so legacy rows stay removable.
- Delete `bind_scope` / `unbind_scope` / `add_bulk_scope_members(_partial)` and the `bound_scope` parameter of `create_scope` / `ensure_scope`, all absorbed by the unified ops; project creation now joins its domain scope via `add_bulk_members`.
- Fix virtual-scope purge leaks: `delete_scope` / `batch_delete_scopes` also remove the edges a deleted scope left in other virtual scopes (its bindings as the reaching side and its entity memberships), which FK CASCADE on the scope's own node never covered.
- Express the edge writes as creator specs (`EntityMembershipCreatorSpec`, `ScopeBindingCreatorSpec`) inserted through spec-based `ON CONFLICT DO NOTHING` helpers.

## Test plan
- [x] `add_bulk_members` writes membership, association, and per-member binding with the given cap — never a reverse binding
- [x] Re-adding a member with a different cap is a no-op that keeps the original cap; an empty member collection is a no-op
- [x] A member without a virtual scope fails the bulk call with nothing written; the partial variant isolates it in `errors` while fully attaching the rest
- [x] `remove_bulk_members` deletes all three edges and still deletes membership/association when the member has no virtual scope
- [x] `delete_scope` / `batch_delete_scopes` remove the deleted scopes' bindings and memberships in surviving virtual scopes

Resolves BA-7174

### DIFF
--- a/src/ai/backend/manager/repositories/ops/rbac/provider.py
+++ b/src/ai/backend/manager/repositories/ops/rbac/provider.py
@@ -1161,7 +1161,7 @@ class RBACWriteOps(WriteOps):
         await self._bulk_create_dependent_ignore_conflicts(
             [
                 ScopeBindingCreatorSpec(
-                    owner=member_scope, scope=scope, permission_cap=permission_cap
+                    anchor_scope=member_scope, bound_scope=scope, permission_cap=permission_cap
                 )
                 for member_scope in member_scopes
             ],
--- a/src/ai/backend/manager/repositories/ops/rbac/provider.py
+++ b/src/ai/backend/manager/repositories/ops/rbac/provider.py
@@ -25,7 +25,7 @@ from ai.backend.common.data.entity.container_registry import CONTAINER_REGISTRY_
 from ai.backend.common.data.entity.domain import DOMAIN_SCOPE_TYPE
 from ai.backend.common.data.entity.project import PROJECT_SCOPE_TYPE
 from ai.backend.common.data.entity.resource_group import RESOURCE_GROUP_SCOPE_TYPE
-from ai.backend.common.data.entity.types import EntityRef, EntityType, ScopeRef, ScopeType
+from ai.backend.common.data.entity.types import EntityRef, ScopeRef, ScopeType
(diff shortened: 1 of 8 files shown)


seedspirit (dismissed):

src/ai/backend/manager/repositories/ops/rbac/provider.py
```diff
@@ -230,6 +236,36 @@ class EntityMembersAddition:
     members: Collection[ScopeMember]
 
 
+@dataclass
+class EntityMemberError:
```
jopemachine: Nit: How about EntityMemberCreationError?

src/ai/backend/manager/repositories/permission_controller/creators.py
```diff
@@ -141,3 +145,42 @@ def build_row(self) -> AssociationScopesEntitiesRow:
...
+
+
+@dataclass
+class ScopeBindingCreatorSpec(
+    DependentCreatorSpec[Mapping[ScopeRef, VirtualScopeID], ScopeBindingRow]
+):
+    """Binding of ``scope`` into ``owner``'s virtual scope; the owner→virtual-scope-id
+    mapping is resolved by the caller at execution time and passed as the dependency."""
+
+    owner: ScopeRef
```
seedspirit: I think it would be better to use a slightly more intuitive name.

seedspirit (approved):

PR #13464, opened 2026-08-05
feat(BA-7178): add devices and device_allocations tables

seedspirit:
## Summary
- Add the `devices` table (physical devices discovered on agents) and the `device_allocations` table, replacing the opaque `kernels.attached_devices` JSONB over the course of the epic.
- `devices` uses a surrogate `id` PK with `UNIQUE (agent_uuid, device_name, designator)`. `designator` is the plugin-assigned handle for the device (cpu core index `"0"`, `"root"`, a GPU UUID) — it is not a database identifier and is unique only within `(agent, device_name)`, hence the name. `model_name` lives here as a device attribute.
- `device_allocations` records the capacity **allocated to a kernel** as one row per capacity entry (`capacity_name`, `quantity Numeric(24,6)`), following the `resource_allocations` family pattern instead of a JSONB blob. Production data shows the same physical GPU carrying different `mem`/`smp` per kernel (fractional GPU shares), so capacity belongs to the attachment, not the device; `Numeric` also preserves fractional quantities.
- A NULL `quantity` row records a bare attachment for devices that report no capacity entries (e.g. the intrinsic `mem` device reports `data: {}`), so attachment facts are not lost by the per-entry decomposition.
- CASCADE FKs: `devices.agent_uuid → agents.uuid`, `device_allocations.kernel_id → kernels.id`, `device_allocations.device_id → devices.id` (+ index on `device_id` for the devices-side cascade). Rows are write-once (INSERT ON CONFLICT DO NOTHING at the RUNNING transition), removed only by FK cascade. The backfill from existing `kernels.attached_devices` is deferred to the write-path PR.

## Test plan
All verified against the live halfstack DB:
- [x] `alembic upgrade` creates both tables; `downgrade` drops them (roundtrip re-verified after each schema iteration)
- [x] `UNIQUE (agent_uuid, device_name, designator)` rejects duplicate device rows; `UNIQUE (kernel_id, device_id, capacity_name)` rejects duplicate allocation rows
- [x] Deleting a device cascades its `device_allocations` rows; agent and kernel FKs declared with the same CASCADE
- [x] NULL `quantity` (bare attachment) accepted; fractional (`0.5`) and >2^32 (`85899345920`) quantities preserved exactly
- [x] Single alembic head after repointing `down_revision` to the current main head (`2dccb3069031`)
- [x] `pants fmt/fix/lint/check` pass for affected packages

Resolves BA-7178


HyeockJinKim (changes requested):
This is a situation where the "devices" table must come first.

src/ai/backend/manager/models/device/row.py
```diff
@@ -0,0 +1,95 @@
...
+    device_id is agent-local (cpu "0", mem "root"), so it is unique only
+    within an agent and a device kind.
+    """
+
+    __tablename__ = "devices"
+
+    agent_id: Mapped[AgentId] = mapped_column("agent_id", sa.String(length=64), primary_key=True)
+    device_name: Mapped[DeviceName] = mapped_column(
+        "device_name", sa.String(length=64), primary_key=True
+    )
```
fregataa, line 42: Should device name be primary?

src/ai/backend/manager/models/device/row.py
```diff
@@ -0,0 +1,95 @@
...
+    device_id: Mapped[DeviceId] = mapped_column(
+        "device_id", sa.String(length=128), primary_key=True
+    )
+    data: Mapped[list[DeviceCapacityEntry]] = mapped_column(
+        "data",
+        PydanticListColumn(DeviceCapacityEntry),
+        nullable=False,
+        server_default=sa.text("'[]'::jsonb"),
+    )
```
fregataa: Did we decide to set resource allocation by JSON?

src/ai/backend/manager/models/device/row.py
seedspirit: I've combined `agent_id`, `device_name`, and `device_id` into a primary key, but would it be better to create a separate ID column and set a unique constraint on the first three columns?

HyeockJinKim: Shouldn't the "device" table handle only device information, rather than being linked to the "agent"?

src/ai/backend/manager/models/device/row.py
```diff
@@ -0,0 +1,108 @@
...
+        "id", GUID, primary_key=True, server_default=sa.text("uuid_generate_v4()")
+    )
+    agent_uuid: Mapped[AgentUUID] = mapped_column("agent_uuid", GUID(AgentUUID), nullable=False)
+    device_name: Mapped[DeviceName] = mapped_column(
+        "device_name", sa.String(length=64), nullable=False
+    )
+    designator: Mapped[DeviceId] = mapped_column(
+        "designator", sa.String(length=128), nullable=False
+    )
+    model_name: Mapped[str] = mapped_column("model_name", sa.String(length=255), nullable=False)
```
HyeockJinKim, line 46: It looks like we need to give the field names here a little more thought.

src/ai/backend/manager/models/device/row.py
```diff
@@ -0,0 +1,108 @@
...
+    __tablename__ = "device_allocations"
+
+    id: Mapped[UUID] = mapped_column(
+        "id", GUID, primary_key=True, server_default=sa.text("uuid_generate_v4()")
+    )
+    kernel_id: Mapped[KernelId] = mapped_column("kernel_id", KernelIDColumnType, nullable=False)
+    device_id: Mapped[UUID] = mapped_column("device_id", GUID, nullable=False)
+    capacity_name: Mapped[str] = mapped_column(
+        "capacity_name", sa.String(length=64), nullable=False
+    )
```
HyeockJinKim, line 82: Is "capacity" supposed to go here?

src/ai/backend/manager/models/device/row.py
```diff
@@ -0,0 +1,108 @@
...
+    set to the device name.
+    """
+
+    __tablename__ = "device_allocations"
+
+    id: Mapped[UUID] = mapped_column(
+        "id", GUID, primary_key=True, server_default=sa.text("uuid_generate_v4()")
+    )
+    kernel_id: Mapped[KernelId] = mapped_column("kernel_id", KernelIDColumnType, nullable=False)
+    device_id: Mapped[UUID] = mapped_column("device_id", GUID, nullable=False)
```
HyeockJinKim, line 79: Please use a typed ID here as well.

src/ai/backend/manager/models/device/row.py
```diff
@@ -0,0 +1,108 @@
...
+    capacity_name: Mapped[str] = mapped_column(
+        "capacity_name", sa.String(length=64), nullable=False
+    )
+    quantity: Mapped[Decimal | None] = mapped_column(
+        "quantity", sa.Numeric(precision=24, scale=6), nullable=True
+    )
```
HyeockJinKim, line 85: Since resources are managed separately, it seems like we need to organize them better. Is there any way to avoid making this optional?

PR #13472, opened 2026-08-05
fix(BA-7190): validate the delegation target before the purge mutates anything

jopemachine:
## 📚 Stacked PRs

Stands alone on `main`.

The auth-context typing stack that used to sit on top of this PR — #13440, #13470, #13468 — no longer depends on it: `users.main_access_key` is gone, so the auth context stopped carrying it. Those three branch from `main` themselves now.

Resolves BA-7190

## Summary

Purging a user with both `purge_shared_vfolders` and `delegate_endpoint_ownership` migrates the shared folders first and only then delegates. The delegation raises `KeyPairNotFound` when the target has no default keypair, so the purge aborts with the folders already moved and nothing to move them back.

This PR asks for the target's default keypair up front, before anything is mutated.

| Change | Detail |
|---|---|
| `repositories/user/db_source`, `repository.py` | `default_access_key(user_id)` — the access key of the user's default keypair, or None |
| `services/user/service.py` | Both purge paths reject a delegation target without one, ahead of the shared vfolder migration |

`delegate_endpoint_ownership` keeps its own lookup and its own `KeyPairNotFound`: it is the caller that uses the value, and the pre-check only fixes the ordering.

A user without a default keypair is reachable: `a2f6b90c41d7` marked keypairs from `users.main_access_key`, so anyone whose column was NULL came through the migration unmarked.

### What this PR no longer does

The original fix — `UserInfoContext.main_access_key` typed `AccessKey` while the column was nullable, with three sites papering over the gap as `AccessKey(value or "")`, and the empty string reaching delegated sessions — is obsolete. The BA-7193 stack removed the field from `UserInfoContext` entirely, and `delegate_endpoint_ownership` now resolves the key from `keypairs.is_default` instead of taking it from the caller. Only the ordering defect survived.

## Verification

- [x] `pants fmt lint check` clean
- [x] `tests/unit/manager/services/test_users.py` — 18 pass, including the rejection case asserting that `migrate_shared_vfolders`, `delegate_endpoint_ownership`, `delete_endpoints` and `purge_user` are all left uncalled
- [ ] CI


jopemachine:
We'll put this PR stack on hold for now, as we've decided to first proceed with the prerequisite work of removing the main_access_key concept altogether.

PR #13483, opened 2026-08-05
feat(BA-7194): mark the default keypair on the keypairs table

jopemachine:
## 📚 Stacked PRs

1. 👉 **`refactor(BA-7194)`: mark the default keypair on the keypairs table** ← you are here
3. #13489 — `refactor(BA-7196)`: serve main_access_key from the keypairs marker
4. #13496 — `refactor(BA-7205)`: move the legacy user GraphQL types off the column
5. #13531 — `feat(BA-7197)`: drop `users.main_access_key`
6. #13533 — `feat(BA-7198)`: make the always-populated keypairs columns NOT NULL
7. #13534 — `feat(BA-7199)`: make the always-populated users columns NOT NULL

Resolves BA-7194

## Summary

`users.main_access_key` records which keypair is a user's default one, and its foreign key is `ON DELETE SET NULL` — deleting *any* keypair silently clears the marker, and nothing restores it. A user whose marker was cleared cannot receive delegated endpoint ownership on purge, and the dev fixtures ship with the column null.

The fact being recorded ("this keypair is its user's default one") describes a keypair, so it moves onto `keypairs`, where deleting the row takes the fact with it instead of corrupting the user.

This is the **expand** step: the new marker is written everywhere the old column is written, nothing reads it yet, and `users.main_access_key` keeps working unchanged.

| Change | Detail |
|---|---|
| `models/keypair/row.py` | `is_default` (NOT NULL, default false) + partial unique index `uq_keypairs_is_default` on `("user") WHERE is_default` |
| migration `a2f6b90c41d7` | Adds the column, backfills from `users.main_access_key`, then creates the index |
| `repositories/keypair/creators.py` | `KeyPairCreatorSpec` carries the flag so the marker is written in the INSERT |
| `repositories/ops/rbac/provider.py` | `create_full_user` marks the default keypair (covers both admin creation and signup) |
| `repositories/user/db_source/db_source.py` | The admin `main_access_key` update and `switch_my_main_access_key` move the marker |
| `plugin/openid/webapp.py` | The keypair created for a new OpenID user is marked |
| `fixtures/manager/example-set-user-main-access-keys.json` | Marks the same five keypairs it points `users` at |

### Backfill

Only keypairs that **their own user** points at are marked:

```sql
UPDATE keypairs
SET is_default = true
FROM users
WHERE users.main_access_key = keypairs.access_key
  AND users.uuid = keypairs."user"
```

The foreign key enforced existence, not ownership, so a `main_access_key` naming another user's keypair is dropped rather than carried over — carrying it would record a false fact and collide with the new index.

Users whose `main_access_key` had already been cleared come out with no marked keypair. That is the state they are in today, and `get_default_keypair_row()` still promotes their first active keypair on the next login, so nothing changes for them here.

### Naming

`is_default`, not `is_main`, ahead of the same rename on `main_access_key` → `default_access_key`. The column this PR reads is still spelled `main_access_key` because it does not move until #13531.

### Migration rehearsal

Run against the dev database inside a rolled-back transaction, on top of the existing rows:

| Case | Result |
|---|---|
| User with two keypairs, `main_access_key` = the first | Only the first marked |
| `main_access_key` naming another user's keypair | Neither marked |
| `main_access_key` NULL | Not marked |
| Existing dev users (5) | All marked correctly |

`5 marked / 5 distinct users` afterwards — the partial unique index creates cleanly.

## Test plan

- [x] `pants fmt lint check` across `manager` and `tests/component` — clean
- [x] Migration rehearsed against the dev DB (table above)
- [x] Component suites green
- [ ] CI `alembic revision --autogenerate` check (model ↔ migration parity)

New coverage:

- `test_user_crud.py::test_s6_create_marks_the_default_keypair_as_main` — the auto-created keypair is the marked one and agrees with `users.main_access_key`
- `test_keypair_ops.py::test_switch_main_moves_the_marker` — switching moves the marker off the old keypair and keeps both sides in step


src/ai/backend/manager/models/alembic/versions/a2f6b90c41d7_add_is_main_to_keypairs.py
```diff
@@ -0,0 +1,52 @@
...
+    op.add_column(
+        "keypairs",
+        sa.Column("is_main", sa.Boolean(), nullable=False, server_default=sa.false()),
+    )
+    op.get_bind().execute(
+        sa.text("""
+            UPDATE keypairs k
+            SET is_main = true
+            FROM users u
+            WHERE u.main_access_key = k.access_key AND u.uuid = k."user"
```
jopemachine: Don't abbreviate

src/ai/backend/manager/models/keypair/row.py
```diff
@@ -53,6 +53,15 @@ def _get_session_row_join_condition() -> sa.ColumnElement[bool]:
...
     __tablename__ = "keypairs"
+    __table_args__ = (
+        # Partial unique index: at most one keypair per user may have is_main = true.
+        sa.Index(
+            "uq_keypairs_is_main",
+            "user",
+            unique=True,
+            postgresql_where=sa.text("is_main"),
+        ),
+    )
```
jopemachine: Can we enforce, as a constraint, that at least one main key pair must always exist?

jopemachine:
Rename title to refactor type

src/ai/backend/manager/models/alembic/versions/a2f6b90c41d7_add_is_main_to_keypairs.py
jopemachine: Spelled out — the backfill now reads `UPDATE keypairs … FROM users WHERE users.main_access_key = keypairs.access_key AND users.uuid = keypairs."user"`. Re-verified against the dev database inside a rolled-back transaction: 5 marked / 5 distinct users, index creates cleanly.

src/ai/backend/manager/models/keypair/row.py
jopemachine: Not as a table constraint. "At most one" is a single-column predicate the partial unique index can express; "at least one" is an assertion across rows of the same table, which PostgreSQL can only check with a `DEFERRABLE INITIALLY DEFERRED` constraint trigger firing on every `keypairs` insert/update/delete.

And today the property is not even true in the data: every user whose `main_access_key` was cleared by the old `ON DELETE SET NULL` foreign key has keypairs and nothing marking one of them, so such a trigger would reject any transaction touching them.

So I fixed the data instead of adding the trigger — the migration now has a second backfill that promotes the **oldest active** keypair of every user the first pass leaves unmarked, the same rule the original `d3f8c74bf148` migration used to populate `main_access_key`. Rehearsed on the dev database:

| Case | Result |
|---|---|
| 3 keypairs, no `main_access_key`, mixed active | Oldest **active** promoted, not the older inactive one |
| Only an inactive keypair | Promoted — a marker on an inactive keypair beats none |
| Already marked by pass 1 | Untouched |

Afterwards `users with keypairs and no main = 0`.

Keeping it that way is then a code guarantee, and the guards already exist: `revoke_my_keypair` and `admin_delete_keypair` both refuse to remove a marked keypair, and `create_full_user` / the OpenID signup path mark the keypair in the INSERT. The marker can only disappear together with the user's last keypair.

Happy to add the deferred constraint trigger on top if you want the database itself to hold the line — say the word and I will put it in BA-7198 (the keypairs NOT NULL pass), which is already touching this table.

jopemachine: Correction to my answer above: the promoting backfill is **gone**. It repaired data rather than moving the fact, so it is out of scope for this PR — the migration now only carries `main_access_key` over.

That means the property is still not true in the data: users whose `main_access_key` had already been cleared come out with no marked keypair, exactly as they are today. A deferred constraint trigger would still reject any transaction touching them, so "at least one" remains unenforceable at the database until someone decides to repair that data.

What this PR does enforce is "at most one", via the partial unique index. Everything else is a code guarantee: `revoke_my_keypair` and `admin_delete_keypair` refuse to remove a marked keypair, the creation paths mark one in the INSERT, and `get_default_keypair_row()` promotes the first active keypair on login for a user that has none.

fregataa (approved):

PR #13489, opened 2026-08-05
refactor(BA-7196): serve main_access_key from the keypairs marker

jopemachine:
## 📚 Stacked PRs

1. #13483 — `refactor(BA-7194)`: mark the default keypair on the keypairs table
3. 👉 **`refactor(BA-7196)`: serve main_access_key from the keypairs marker** ← you are here
4. #13496 — `refactor(BA-7205)`: move the legacy user GraphQL types off the column
5. #13531 — `feat(BA-7197)`: drop `users.main_access_key`
6. #13533 — `feat(BA-7198)`: make the always-populated keypairs columns NOT NULL
7. #13534 — `feat(BA-7199)`: make the always-populated users columns NOT NULL

Resolves BA-7196

## Summary

The value the API reports as `main_access_key` now comes from the keypair marked `is_default`, not from the `users` column. A user whose column was cleared by the old `ON DELETE SET NULL` foreign key stops reporting `null` while owning a perfectly good main keypair.

| Change | Detail |
|---|---|
| `models/user/row.py` | `UserRow.main_keypair_access_key` — a scalar-subquery column property over the marked keypair |
| `UserRow.to_data()` | sources `UserData.main_access_key` from it |
| `UserData.from_row()` | takes the access key as an argument; it accepts Core rows, which carry the column rather than the marker |
| `repositories/user`, `repositories/session` | pass it in — the two Core update paths query the marker after the update |
| `repositories/export/reports/user.py` | the main-keypair join follows the marker, and `main_access_key` is served by that join |
| `api/gql_legacy/user.py` | the `UserNode` filter and order specs carry the column property through `ORMFieldItem` |
| `models/minilang/` | `ORMFieldItem` — a spec that carries a column expression instead of a name — is now accepted by the ordering parser too, and its type widened to any `ColumnElement` |

`UserData` is what feeds the v2 DTOs, the REST responses, the Strawberry API and the two services that build a `UserInfoContext` (purge) or resolve a delegated session owner — all of them follow without further changes.

### Why the specs need `ORMFieldItem`

The filter and order parsers resolve a spec's string name against the statement's table, not against the ORM entity, so a column property is invisible to them — naming it produced `Unknown/unsupported field name` at parse time. Passing the attribute itself through `ORMFieldItem` sidesteps the lookup. The filter parser already understood that form; the ordering parser did not, so it does now.

### Why a column property and not the relationship

`to_data()` runs on rows fetched with no loader options at all, so anything lazy would raise `MissingGreenlet`. A column property loads with the row like an ordinary column, and the partial unique index added in #13483 (`("user") WHERE is_default`) is exactly the index the subquery needs.

Rendered:

```sql
SELECT (SELECT keypairs.access_key FROM keypairs
        WHERE keypairs."user" = users.uuid AND keypairs.is_default) AS anon_1,
       users.uuid, users.username, …
```

The physical column stays mapped and dual-written — dropping it from the ORM before the migration would make CI's `alembic revision --autogenerate` check fail.

### Deliberately not in this PR

The legacy graphene `User` type builds its rows with `sa.select(users)` (Core) and reads `row.main_access_key`, and its filter/order specs resolve against the `users` table. Serving those from the marker means reshaping the Core selects, which is a different kind of edit — **BA-7205**, landing before the column is dropped. Until then the dual write keeps that path correct.

## Verification

- [x] The column property renders as a correlated subquery and the mappers configure

New coverage in `tests/unit/manager/repositories/user/test_user_repository.py`, on the fixture whose `users.main_access_key` is deliberately unset:

- `test_user_data_reports_the_marked_keypair_as_main_access_key`
- `test_main_access_key_follows_a_switch`

`tests/unit/manager/repositories/export/reports/test_user.py` — `main_access_key` now asserts the keypair join instead of asserting there is none.

----
📚 Documentation preview 📚: https://sorna--13489.org.readthedocs.build/en/13489/

----
📚 Documentation preview 📚: https://sorna-ko--13489.org.readthedocs.build/ko/13489/


jopemachine:
Folded into #13484. Keeping the reader move split between "internal readers" and "API readers" was not a real review boundary — both are the same operation on the same column, and the split stranded `UserRow.default_keypair_access_key` one PR away from the callers that needed it. #13484 now carries the whole reader move, and #13496 is stacked directly on it.

PR #13496, opened 2026-08-05
fix(BA-7205): refuse to delete a keypair marked as its user's default

jopemachine:
## 📚 Stacked PRs

**Closed — no longer part of the stack.** Schema-truth stack (BA-7193), merge in order (bottom-up):

1. #13483 — `refactor(BA-7194)`: mark the default keypair on the keypairs table ✅
3. #13531 — `feat(BA-7197)`: drop `users.main_access_key`
4. #13533 — `feat(BA-7198)`: make the always-populated keypairs columns NOT NULL
5. #13534 — `feat(BA-7199)`: make the always-populated users columns NOT NULL

#13531 is based directly on #13484.

Relates to BA-7205

> **Rescoped.** BA-7205 was "move the legacy user GraphQL types off `users.main_access_key`". They are not being moved: `main_access_key` stays the column it is named after, and the legacy type leaves together with the column in BA-7197. What remains here is the one change in that PR that was never about the API field.

## Summary

`DeleteKeyPair` refuses to delete a keypair that is somebody's main one. It asked the question through `users.main_access_key`:

```python
sa.select(sa.func.count())
    .select_from(UserRow)
    .where(UserRow.main_access_key == access_key)
```

That counts users pointing at the key. The same question is now written on the keypair itself:

```python
sa.select(sa.func.count())
    .select_from(KeyPairRow)
    .where((KeyPairRow.access_key == access_key) & KeyPairRow.is_default)
```

Same answer today, because #13483 writes the marker everywhere the column is written. The difference is that the marker survives what the column does not — an unrelated keypair's deletion used to clear `users.main_access_key` through `ON DELETE SET NULL`, and this guard then waved the real main keypair through.

The v2 path (`admin_delete_keypair`) already reads `kp_row.is_default` as of #13484; this is its legacy GraphQL counterpart.

## Verification

- [x] `pants fmt lint check` — clean

No test changes: the guard's behaviour is unchanged for every state the dual write can produce.

----
📚 Documentation preview 📚: https://sorna--13496.org.readthedocs.build/en/13496/

----
📚 Documentation preview 📚: https://sorna-ko--13496.org.readthedocs.build/ko/13496/


jopemachine:
Closing: nothing left in this PR.

BA-7205 was "move the legacy user GraphQL types off `users.main_access_key`". They are not being moved — `main_access_key` stays the column it is named after, and the legacy types leave together with the column, deprecated in one step rather than rerouted first (see #13484).

The one change here that was not about the API field — the `DeleteKeyPair` guard — went to #13484 briefly and was reverted for the same reason: it is a legacy mutation, so it keeps reading `users.main_access_key` until the whole surface is deprecated together.

#13531 is now based directly on #13484.

PR #13504, opened 2026-08-05
feat(BA-7212): add a uuid column to the agents table

fregataa:
## Summary
- Add a `uuid` column (UUID type, UNIQUE, `server_default uuid_generate_v4()`) to the `agents` table, keeping the existing varchar(64) primary key as-is.
- Introduce the `AgentUUID` identifier type in `common/identifier/agent.py` and map the column as `GUID(AgentUUID)` on `AgentRow`.
- Add an Alembic migration that adds the column with the volatile server default, which backfills existing rows with distinct UUIDs.

## Test plan
- [x] `alembic upgrade head` on a local DB: existing agent rows are backfilled with distinct UUIDs.
- [x] Column definition verified: `uuid` NOT NULL, default `uuid_generate_v4()`, unique constraint `uq_agents_uuid`.
- [x] `alembic downgrade -1` → `upgrade head` round trip passes.

Resolves BA-7212

### DIFF
--- a/src/ai/backend/manager/models/agent/row.py
+++ b/src/ai/backend/manager/models/agent/row.py
@@ -20,6 +20,7 @@ from sqlalchemy.orm import (
 from sqlalchemy.sql.expression import false, true
 
 from ai.backend.common.auth import PublicKey
+from ai.backend.common.identifier.agent import AgentUUID
 from ai.backend.common.identifier.resource_group import ResourceGroupID
 from ai.backend.common.types import AccessKey, AgentId, ResourceSlot, SlotName, SlotTypes
 from ai.backend.manager.data.agent.types import (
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/models/agent/row.py
```diff
@@ -67,6 +68,13 @@
...
     __tablename__ = "agents"
 
+    uuid: Mapped[AgentUUID] = mapped_column(
+        "uuid",
+        GUID(AgentUUID),
+        unique=True,
+        nullable=False,
+        server_default=sa.text("uuid_generate_v4()"),
+    )
     id: Mapped[str] = mapped_column("id", sa.String(length=64), primary_key=True)
```
jopemachine: It looks better to update this here as well so that it uses AgentID.

jopemachine (approved):

PR #13520, opened 2026-08-06
refactor(BA-7216): make lifecycle timestamp columns non-nullable

fregataa:
## Summary
- Adopt `CreatedAtMixin` / `UpdatedAtMixin` (`LifecycleTimestampsMixin`) across 17 Row classes whose lifecycle timestamps (`created_at` / `modified_at` / `updated_at`) were declared nullable despite always being populated on insert; legacy `sa.Table` definitions (`session_templates`, `error_logs`) and `users.password_changed_at` are fixed in place. Columns that had `index=True` keep their index via an identically-named `sa.Index` in `__table_args__`.
- Rename the legacy `modified_at` columns of `users` / `keypairs` / `groups` / `vfolder_invitations` to `updated_at`. External contracts are unchanged — GraphQL fields, DTO/data-class fields, and export report keys keep the `modified_at` spelling and map to the renamed ORM attribute internally.
- Add migration `2dccb3069031`: renames the four columns, backfills NULLs from the sibling timestamp (falling back to `now()`), sets NOT NULL on 29 columns, and adds the missing `now()` server defaults to 9 columns (including `artifact_revisions` and six `updated_at`-with-`onupdate`-only tables where NULL previously meant "never updated").

## Test plan
- [x] Metadata introspection: all target columns report `nullable=False` and index names are unchanged (`ix_<table>_<column>`)
- [x] Local DB: seeded NULL timestamps on users/keypairs/groups/scaling_groups rows → `alembic upgrade` backfills them (both-NULL rows get `created_at == updated_at`), renames columns, applies NOT NULL and new server defaults
- [x] `alembic downgrade` restores nullable `modified_at` columns; re-upgrade succeeds
- [x] `pants fmt` / `fix` / `lint` pass

Resolves BA-7216


src/ai/backend/manager/models/artifact_revision/row.py
```diff
@@ -46,11 +46,13 @@ def _get_association_artifacts_storages_join_cond() -> sa.ColumnElement[bool]:
...
 
 
-class ArtifactRevisionRow(Base):  # type: ignore[misc]
+class ArtifactRevisionRow(LifecycleTimestampsMixin, Base):  # type: ignore[misc]
     __tablename__ = "artifact_revisions"
     __table_args__ = (
         # constraint
         sa.UniqueConstraint("artifact_id", "version", name="uq_artifact_id_version"),
+        sa.Index("ix_artifact_revisions_created_at", "created_at"),
+        sa.Index("ix_artifact_revisions_updated_at", "updated_at"),
```
seedspirit: nit: Is there a reason you're adding an index?

fregataa: they already exist. I just changed the syntax

src/ai/backend/manager/models/image/row.py
```diff
@@ -152,12 +153,13 @@ def _get_container_registry_join_condition() -> sa.sql.elements.ColumnElement[An
...
 
 
-class ImageRow(Base):  # type: ignore[misc]
+class ImageRow(CreatedAtMixin, Base):  # type: ignore[misc]
     __tablename__ = "images"
     __table_args__ = (
         sa.UniqueConstraint(
             "registry", "project", "name", "tag", "architecture", name="uq_image_identifier"
         ),
+        sa.Index("ix_images_created_at", "created_at"),
```
fregataa, line 162: Such index definitions are not added in this job. They already exist and only the syntax changed.

src/ai/backend/manager/models/image/row.py
```diff
@@ -409,7 +404,8 @@ def from_dataclass(cls, image_data: ImageData) -> Self:
             status=image_data.status,
         )
         image_row.id = image_data.id
-        image_row.created_at = image_data.created_at
+        if image_data.created_at is not None:
+            image_row.created_at = image_data.created_at
```
seedspirit, line 408: It would be good if modifying the data object were also addressed as a follow-up task.

seedspirit (approved):

PR #13523, opened 2026-08-06
feat(BA-7192): add create/update/delete API for resource slot types

HyeockJinKim:
## Summary

- The slot type catalog could only be changed by editing the database directly after the seeding migration. This opens a super-admin path to register, adjust and remove entries across REST v2, GraphQL, the SDK and the CLI.
- `enabled` and `required` are now returned by reads. `enabled` already drove the scheduler's `ImageSlotTypeRule` but appeared in no response, so a write could not be verified. This is what makes the `cuda.device` workaround (`UPDATE resource_slot_types SET enabled = false`) unnecessary on an fGPU cluster.
- `resource_slot_types` gains a `uuid` unique alternate key, following the composition of #13504. The `slot_name` primary key and the five referencing foreign keys are untouched — swapping the PK would drag five child tables and the scheduler hot path with it.
- Writes run through the generic ops path: the domain contributes creator/updater/purger specs and adds no repository or service method. Two symmetric bases (`UpdateGlobalOpsAction`, `PurgeGlobalOpsAction`) and their registry factories complete the existing global set.

### Contract

| | |
|---|---|
| Endpoints | `POST /v2/resource-slots/slot-types`, `PATCH|DELETE .../{slot_name}` |
| Mutations | `adminCreateResourceSlotType`, `adminUpdateResourceSlotType`, `adminPurgeResourceSlotType` |
| CLI | `./bai resource-slot slot-type create|update|delete` |
| Immutable | `slot_name`, `slot_type` |
| Create | fails on an existing name rather than overwriting |
| Delete | refused while `agent_resources`, `resource_allocations`, `model_card_resource_requirements`, `preset_resource_slots` or `deployment_revision_resource_slots` still names the slot |

All three writes are super-admin only at three layers: the REST middleware, `check_admin_only()` in the resolver, and the global action shape's `SuperAdminActionValidator`.

The GraphQL Node id stays `slot_name`; the new `uuid` column is not exposed.

## Test plan

- [x] `pants lint` / `pants check --changed-dependents=transitive`
- [x] `pants test --changed-dependents=direct` — 51 targets
- [x] Repository specs: create inserts and returns stored values, each row gets a distinct uuid, an existing name conflicts and leaves the row unchanged; update writes only the named fields, an unknown name matches nothing; purge removes an unreferenced slot type and is refused while `agent_resources` references it
- [x] Seeded uuids pinned in the fixture and backfilled by the migration; verified against a scratch PostgreSQL that seeded rows take the pinned values, an unlisted slot keeps its generated one, and the unique constraint holds
- [x] GraphQL schema dumps regenerated — additive only, 0 removals
- [x] Migration applied to a seeded database: 14 rows, every uuid matching the fixture
- [x] Live round-trip via `./bai` against a local stack (manager + agent):
  - reads report `enabled` / `required` / `uuid`
  - create inserts; a duplicate name returns the conflict and leaves the row untouched; the CLI rejects a `slot_type` outside the enum
  - update flips `enabled`, adjusts `number_format`, and returns not-found for an unknown name
  - delete removes an unreferenced slot and is refused for `cpu` while an agent reports it
  - all three GraphQL mutations round-trip
  - a non-superadmin is refused on all three REST endpoints, and the mutations are not exposed to them at all in the schema

This surfaced one defect that no unit test could: the domain errors named no
`web.HTTP*` base, so they serialized as `HTTP/1.1 -1` and a duplicate name came
back as a proxy 502. Fixed here, along with the four errors that predate this
branch and carried the same defect.

Resolves BA-7192

https://claude.ai/code/session_01M6UYi5DVfpZxt2LgmiE5Z3

----
📚 Documentation preview 📚: https://sorna--13523.org.readthedocs.build/en/13523/

----
📚 Documentation preview 📚: https://sorna-ko--13523.org.readthedocs.build/ko/13523/

----
📚 Documentation preview 📚: https://sorna--13523.org.readthedocs.build/en/13523/

----
📚 Documentation preview 📚: https://sorna-ko--13523.org.readthedocs.build/ko/13523/


src/ai/backend/manager/services/resource_slot/processors.py
```diff
@@ -55,12 +67,25 @@ class ResourceSlotProcessors(AbstractProcessorPackage):
...
+        PurgeGlobalOpsAction[Any, ResourceSlotTypeData],
+        EntityOpsResult[ResourceSlotTypeData],
+    ]
 
     def __init__(
         self,
         service: ResourceSlotService,
         action_monitors: list[ActionMonitor],
         validators: ActionValidators,
+        registry: ProcessorRegistry[ResourceSlotTypeData],
```
fregataa: it seems this should be group

src/ai/backend/manager/repositories/resource_slot/creators.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+
+    slot_name: str
+    slot_type: str
+    required: bool = False
+    enabled: bool = True
+    display_name: str = ""
+    description: str = ""
+    display_unit: str = ""
+    display_icon: str = ""
```
fregataa: Are those default values ok?

src/ai/backend/manager/repositories/resource_slot/creators.py
```diff
@@ -0,0 +1,67 @@
...
+class ResourceSlotTypeCreator(DataCreator[ResourceSlotTypeRow, ResourceSlotTypeData]):
+    """Creator for a resource slot type.
+
+    Registering a name that already exists fails rather than overwriting: the
+    catalog row is the FK target of five tables, so replacing one silently would
+    reinterpret every quota already expressed in that slot.
+    """
+
+    slot_name: str
+    slot_type: str
```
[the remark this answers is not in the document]
fregataa: Please also check this, SlotTypes can be used here?

fregataa (approved):

src/ai/backend/manager/api/rest/v2/resource_slot/handler.py
```diff
@@ -34,6 +38,34 @@ async def search_slot_types(
...
+        result = await self._adapter.admin_create_slot_type(body.parsed)
+        return APIResponse.build(status_code=HTTPStatus.CREATED, response_model=result)
+
+    async def admin_update_slot_type(
+        self,
+        path: PathParam[SlotNamePathParam],
+        body: BodyParam[UpdateResourceSlotTypeInput],
+    ) -> APIResponse:
+        """Update a resource slot type by slot name (superadmin only)."""
+        merged = body.parsed.model_copy(update={"slot_name": path.parsed.slot_name})
```
fregataa, line 55: Should we make separate `UpdateResourceSlotTypeInput` and `UpdateResourceSlotTypeInput` with any extra id field?

PR #13531, opened 2026-08-06
refactor(BA-7197): drop users.main_access_key

jopemachine:
## 📚 Stacked PRs

1. #13483 — `refactor(BA-7194)`: mark the default keypair on the keypairs table ✅
2. #13484 — `refactor(BA-7195)`: read the default keypair from the keypairs marker ✅
3. 👉 **`refactor(BA-7197)`: drop `users.main_access_key`** ← you are here

Split off: #13533 (`refactor(BA-7198)`) and #13534 (`feat(BA-7199)`) tighten `keypairs` and `users` columns that have nothing to do with `main_access_key`. Only the migration's `down_revision` tied them to this chain; they branch from `main` now.

Dropped: ~~#13496 — `refactor(BA-7205)`: move the legacy user GraphQL types off the column~~ and ~~#13611 — `feat(BA-7193)`: expose the keypair owner's email~~. Both assumed a column consolidation that is not happening.

Resolves BA-7197

## Summary

`keypairs.is_default` records which keypair is a user's main one. `users.main_access_key` recorded the same fact a second time, on the row that does not own it — two places to keep in step, and a partial unique index guards only one of them. The column goes; the marker stays.

**Nothing a client can see changes.** `main_access_key` keeps its name, its value, its update path and its filters — it just reads the marker now. Both GraphQL schema dumps regenerate byte-identical.

### The column and its writes

| Change | Detail |
|---|---|
| migration `c8d51e7a3b62` | Drops `users.main_access_key` and `fk_users_main_access_key_keypairs`; the downgrade restores both and refills them from the marker |
| `models/user/row.py` | The mapped column goes; the relationship joining on the marker is `default_keypair` |
| `repositories/ops/rbac/provider.py` | Provisioning stops writing the column — the keypair it creates is already `is_default=True` — and the created row is reloaded with its marker, since `to_data()` reads the relationship |
| `repositories/user/db_source/db_source.py` | Both `sa.update(users).values(main_access_key=…)` statements go |
| `repositories/user/updaters.py` | `main_access_key` leaves `UserUpdaterSpec` altogether: it is not a `users` column, so it no longer travels in the users UPDATE |
| `fixtures/manager/*.json` | `example-keypairs.json` carries `is_default`, and `example-set-user-main-access-keys.json` is deleted — the users↔keypairs cycle that forced a post-insert update fixture is gone with the column |

OpenID signup needed no change — `main` routes it through `RBACOpsProvider`, so it stops writing the column with every other provisioning path.

### The update path

`main_access_key` on the update APIs no longer rides the users UPDATE. Every caller updates the user first and moves the marker only if that succeeded:

| Caller | After |
|---|---|
| REST update handler | `SwitchDefaultAccessKeyAction(user_id, access_key)` after `ModifyUserAction` |
| v2 GQL `updateUser` (adapter) | the same action, after the update returns |
| v2 GQL bulk update | one switch per user, only for users whose own update went through; a switch that fails becomes that user's failure instead of aborting the batch |
| legacy graphene `ModifyUser` | the same action, after `modify_user` |
| `switchMyMainAccessKey` | unchanged externally; the action behind it is the same id-based one |

`switch_default_access_key(user_id: UserID, access_key: AccessKey)` is the single entry point — the email-based validate-and-update helper is gone, and there is no `require_active` switch: a missing, foreign or inactive keypair is rejected with `KeyPairNotFound` / `KeyPairForbidden`. An explicit `null` is ignored, and the request DTO says so now; clearing the marker would leave the user with no default keypair, which is also what the legacy field has always documented.

### Where the API value comes from now

The relationship is loaded where the value is wanted, and nowhere else — no column property, no expression on every `select(UserRow)`.

| Reader | After |
|---|---|
| `UserRow.to_data()` | `self.default_keypair.access_key` |
| the six ORM queries whose rows become `UserData` | `.options(joinedload(UserRow.default_keypair))` |
| `session/db_source.get_session_owner` | the same option, and `to_data()` instead of `from_row()` |
| `UserData.from_row()` | reads `default_access_key`; both remaining callers hold a Core `RETURNING` row that labels it |
| legacy graphene `User` — four loaders | select a correlated subquery labelled `main_access_key` |
| legacy `User` / `UserNode` filter + order specs | `ORMFieldItem` carrying that same subquery |
| `DeleteKeyPair` guard | `KeyPairRow.is_default` |
| `UserRow.get_main_keypair_row` | renamed `get_default_keypair_row` — it reads the relationship, not the column (folded in from the closed #13691) |

The legacy loaders carry a **subquery rather than a join** on purpose: a join replaces the statement's `FROM`, and minilang resolves its field specs against `select.froms[0]`. Compiled, `froms[0]` stays `Table users`.

`QueryOrderParser` gains the `ORMFieldItem` case the filter parser already had, and `ORMFieldItem` widens to any column expression — a scalar subquery is one. Without it, ordering by `main_access_key` would have had to be dropped, which would have removed working behaviour. The ordering column type widens with it (`OrderingColumn`, shared with the two legacy pagination helpers) instead of suppressing the assignment.

### Names

Everything a client sees keeps the `main_access_key` spelling; the internals follow the marker — `default_keypair`, `UserData.default_access_key`, `SwitchDefaultAccessKeyAction`, `switch_default_access_key`, and the errors that report it now say "default access key". The GQL resolver keeps the name `switch_my_main_access_key`, so the exposed field stays `switchMyMainAccessKey` with no explicit `name=`.

## Verification

- [x] Migration and its downgrade rehearsed against a throwaway database built by `mgr schema oneshot`
- [x] `select(UserRow)` keeps `users` as its sole `FROM`; the legacy loaders compile with the subquery labelled `main_access_key`
- [x] `main_access_key == "…"` and `-main_access_key` both reach `keypairs.is_default`, on the `User` and `UserNode` spec sets alike
- [x] Both GraphQL schema dumps regenerate byte-identical; `openapi.json` changes only the one description line
- [x] `pants fmt lint check` clean, mypy over the changed targets and their transitive dependents
- [ ] Unit, component and integration suites in CI

The `UserData` conversion tests move to `UserRow.to_data()` — `from_row()` now reads a label only the Core selects carry. `test_queryorder` gains a scalar-subquery ordering case. Test setups that seeded the column lose those lines, and two component assertions changed shape because they compared the two sides against each other: creating a user now asserts that the marked keypair is exactly the one it owns, and switching asserts the marker moved.

----
📚 Documentation preview 📚: https://sorna--13531.org.readthedocs.build/en/13531/

----
📚 Documentation preview 📚: https://sorna-ko--13531.org.readthedocs.build/ko/13531/


src/ai/backend/manager/api/gql_legacy/user.py
```diff
@@ -1129,6 +1157,13 @@ async def mutate(
         res: ModifyUserActionResult = await graph_ctx.processors.user.modify_user.wait_for_complete(
             action
         )
+        if props.main_access_key is not Undefined and props.main_access_key is not None:
+            await graph_ctx.processors.user.switch_default_access_key.wait_for_complete(
```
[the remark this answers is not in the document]
jopemachine: I don't like that this doesn't provide transactional guarantees, but since I'm planning to deprecate the field itself and introduce a separate mutation for updating the default keypair, I'll leave this as is for now.

src/ai/backend/manager/api/adapters/user/adapter.py
```diff
@@ -630,17 +629,34 @@ async def bulk_create_users_with_keypair(
...
             for error in result.data.failures
         ]
+        updated_users = []
+        for user in result.data.successes:
+            access_key = default_key_switches.get(UserID(user.id))
+            if access_key is not None:
+                try:
+                    await self.switch_default_access_key(UserID(user.id), access_key)
+                except Exception as e:
+                    failed.append(BulkUpdateUserV2Error(user_id=user.id, message=str(e)))
```
fregataa, line 657: 1. What kind of exception can be handled here?
2. Should the `modify_users()` adaptor method keep supporting setting default access key?

src/ai/backend/manager/repositories/ops/rbac/provider.py
```diff
@@ -1045,10 +1044,12 @@ async def create_full_user(
...
             )
 
-        # Flushing the main_access_key update expires server-onupdate columns; reload
-        # so callers can read the row without a sync-context lazy refresh.
+        # The insert leaves the server-default columns unloaded, and default_keypair is the
+        # keypair created just above; reload both so callers can read the row without a
+        # sync-context lazy refresh.
         await self._sess.flush()
         await self._sess.refresh(user_row)
+        await self._sess.refresh(user_row, ["default_keypair"])
```
fregataa, line 1052: Does this refresh `keypair_row`?

HyeockJinKim (approved):

HyeockJinKim (approved):

PR #13539, opened 2026-08-06
feat(BA-7233): require service port in runtime variant default model definition

seedspirit:
## Summary
- Add `DefaultModelServiceConfig` (required `port`) and make `DefaultModelConfig.service` mandatory, so the runtime variant baseline guarantees every revision resolves a service port the same way it guarantees the model name (BA-7209). The stored baseline is validated at the DB column boundary.
- The `command`/legacy `start_command` normalization validator carries over from `ModelServiceConfigDraft` — fixtures and raw inserts still supply list-form `start_command`.
- Alembic migration `55e0c3669e2e`: idempotent backfill that creates missing `service` blocks and fills absent/null ports with `8000` (every seeded baseline already ships a port; only rows written outside the seed chain can lack one). Non-object array elements are skipped; downgrade is a no-op.
- No fixture changes needed: all bundled runtime variants already carry `service.port`.

## Test plan
- [x] `pants lint/check` green including `--changed-dependents=direct` (pre-push scope); reader/adapter/repository test suites pass
- [x] Migration logic verified against a live DB across null shapes (missing/null service, scalar service, null port, string port, JSON-null models/elements, top-level null) — fills correctly, idempotent second pass selects nothing
- [x] Migration also executed via a real `alembic upgrade head` run on the local DB
- [x] Live: manager restart + `runtime-variant search` loads all variants through the new required-port validation

Resolves BA-7233

----
📚 Documentation preview 📚: https://sorna--13539.org.readthedocs.build/en/13539/

----
📚 Documentation preview 📚: https://sorna-ko--13539.org.readthedocs.build/ko/13539/

### DIFF
--- a/src/ai/backend/common/config.py
+++ b/src/ai/backend/common/config.py
@@ -753,8 +753,6 @@ class DefaultModelServiceConfig(BaseConfigModel):
     pre_start_actions: list[PreStartAction] | None = None
     start_command: str | None = None
     shell: str | None = None
-    # Mirrors the resolved ModelServiceConfig.port constraint so invalid ports
-    # fail at the DB boundary instead of at to_resolved().
     port: int = Field(gt=1)
     health_check: ModelHealthCheckDraft | None = None
 
--- a/src/ai/backend/common/config.py
+++ b/src/ai/backend/common/config.py
@@ -5,7 +5,7 @@ import shlex
 import sys
 from collections.abc import Mapping, MutableMapping
 from pathlib import Path
-from typing import Annotated, Any, override
+from typing import Any, override
 
 import humps
 import tomli
@@ -146,10 +146,6 @@ agent_selector_config_iv = t.Dict({}) | agent_selector_globalconfig_iv
 
 DEFAULT_SHELL = "/bin/bash"
 
-# Shared port constraint for every model-service config variant (strict,
-# draft, baseline) so the same range is enforced at every boundary.
-ModelServicePort = Annotated[int, Field(gt=1)]
-
 
 class PreStartAction(BaseConfigModel):
     action: str = Field(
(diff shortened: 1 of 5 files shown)


fregataa (dismissed):

jopemachine (dismissed):

src/ai/backend/common/config.py
```diff
@@ -746,15 +746,32 @@ def from_file_payload(cls, payload: Mapping[str, Any]) -> ModelDefinitionDraft:
...
 
 
+class DefaultModelServiceConfig(BaseConfigModel):
+    """Baseline model service config; ``port`` stays required so the baseline
+    guarantees every revision resolves a service port."""
+
+    pre_start_actions: list[PreStartAction] | None = None
+    start_command: str | None = None
+    shell: str | None = None
+    port: int = Field(gt=1)
```
fregataa, line 756: this is not the same as existing `ModelServiceConfigDraft` which doesn't have gt option. let's add or use port type field

src/ai/backend/manager/models/alembic/versions/55e0c3669e2e_backfill_missing_service_ports_in_runtime_variants.py
```diff
@@ -0,0 +1,65 @@
...
+    rows = bind.execute(
+        sa.text(
+            "SELECT id, default_model_definition FROM runtime_variants "
+            "WHERE jsonb_typeof(default_model_definition->'models') = 'array' "
+            "AND EXISTS ("
+            "  SELECT 1 FROM jsonb_array_elements(default_model_definition->'models') AS m"
+            "  WHERE jsonb_typeof(m) = 'object' AND (m->'service'->>'port') IS NULL"
+            ")"
+        )
+    ).fetchall()
```
fregataa, line 46: as the new default type require ports greater than 1, this revision should check the value

fregataa (approved):

PR #13556, opened 2026-08-06
refactor(BA-7201): route user project enrollment through the RBAC member ops

fregataa:
## Summary
- Consolidate the `modify_user` group-change path into a single `UserDBSource._sync_user_project_memberships` that runs through the RBAC member ops (`add_bulk_members` / `remove_bulk_members` + a project-role purger) in its own transaction, so the virtual-scope chain (`entity_memberships`, `scope_bindings`) and `auto_assign` role grants stay consistent with the modifyGroup path; the OpenID plugin's hand-rolled provisioning is replaced with `create_full_user`.
- Move all user↔scope membership reads off `association_scopes_entities` onto the virtual-scope chain via new `models/virtual_scope/queries.py` helpers — across the repositories layer, the legacy permission helpers (`models/vfolder`, `models/image`, `utils.py`), and the legacy GraphQL resolvers. Writes keep dual-writing the legacy association.
- Add an idempotent data migration that backfills `virtual_scopes`, `entity_memberships`, and `scope_bindings` from the existing DOMAIN/USER and PROJECT/USER associations, so the chain is complete before reads rely on it.

## Test plan
- [x] `pants test tests/component/openid:: tests/component/user:: tests/component/project:: tests/unit/manager/repositories::`
- [x] Migration verified locally: backfill correctness, ON CONFLICT idempotency (downgrade/re-upgrade), reuse of pre-existing chain rows
- [x] Live-server verification (halfstack, `./bai` + legacy REST/GQL):
  - Backfill: legacy PROJECT/USER association rows == virtual-scope chain rows; legacy GQL membership reads (`user.groups`, `users(group_id:)`, `keypair_list`, `vfolder_project_list`) intact
  - `modify_user` group change: dual-write consistency, `auto_assign` role grant on join, project-role purge on leave (model-store projects always kept, as designed)
  - `project assign-users` / `unassign-users`: membership dual-write and removal
  - Legacy `POST /session`: non-member group rejected (400), member group enqueued (201)
  - `POST /folders/{name}/share`: share to member succeeds, non-member email filtered out, user vfolder rejected
  - `GET /auth/role` / `POST /resource/check-presets`: member vs. non-member behavior correct
  - Regular-user scope boundary: `vfolder` / `model-card` project-search limited to member projects

Resolves BA-7201


src/ai/backend/manager/repositories/vfolder/repository.py
```diff
@@ -1821,22 +1823,16 @@ async def share_vfolder_with_users(
                 domain_name=domain_name,
             )
 
+            if vfolder_group is None:
+                # Sharing targets a group vfolder; a folder without an owning
+                # project cannot be shared with project members.
+                raise VFolderInvalidParameter("Only group vfolders can be shared with users.")
```
seedspirit, line 1829: Were you not performing this validation separately before?

src/ai/backend/manager/repositories/user/db_source/db_source.py
```diff
@@ -1169,18 +1091,7 @@ async def search_users_by_project(
...
-                .select_from(UserRow)
-                .join(
-                    AssociationScopesEntitiesRow,
-                    sa.and_(
-                        sa.cast(UserRow.uuid, sa.String) == AssociationScopesEntitiesRow.entity_id,
-                        AssociationScopesEntitiesRow.scope_type == ScopeType.PROJECT,
-                        AssociationScopesEntitiesRow.entity_type == EntityType.USER,
-                    ),
-                )
-            )
```
seedspirit, line 1183: Has logic for finding a user within a scope been added inside `execute_batch_querier`?

src/ai/backend/manager/repositories/vfolder/repository.py
fregataa: this method is only used in a legacy API handler and yeah it does not have such validation

src/ai/backend/manager/repositories/user/db_source/db_source.py
fregataa: The logic moved to `ProjectUserSearchScope.to_condition()`

seedspirit (approved):

PR #13572, opened 2026-08-06
feat(BA-7256): add a partial bulk upsert primitive to the RBAC write ops

jopemachine:
## 📚 Stacked PRs

1. 👉 **#13572** — `feat(BA-7256): add a partial bulk upsert primitive to the RBAC write ops` ← you are here
2. ⬇️ #13563 — `feat(BA-7250): support partial success in bulk app config fragment upsert`

Both PRs build and pass CI independently.

## Summary
- The RBAC write ops offer upsert only as the single-item `upsert_scoped`; the bulk partial primitives exist for create (`bulk_create_scoped_partial`) and purge (`bulk_purge_scoped_partial`) but not for upsert, so a bulk upsert caller has to loop the single op inside one transaction and gets all-or-nothing semantics.
- Add `bulk_upsert_scoped_partial`: each row and its scope association share one savepoint, so a rejected row rolls back alone and the rest of the batch lands. The result (`RBACBulkEntityUpserterResultWithFailures`) reports successes and per-index errors (`BulkUpserterError`), mirroring the create/purge partial primitives.

Resolves BA-7256.


src/ai/backend/manager/repositories/ops/rbac/provider.py
```diff
@@ -593,6 +595,30 @@ async def upsert_scoped[TRow: Base](
...
+            try:
+                async with self.savepoint():
+                    result = await self.upsert_scoped(upserter)
+                items.append(result.row)
+            except Exception as e:
+                # upsert_scoped maps the integrity errors its spec declares onto domain
+                # errors; whatever arrives here fails just this row.
+                failed.append(BulkUpserterError(spec=upserter.spec, exception=e, index=index))
+        return RBACBulkEntityUpserterResultWithFailures(items=items, failed=failed)
+
```
fregataa, line 621: Why is this implemented in rbac ops?

PR #13607, opened 2026-08-07
feat(BA-7193): expose the keypair's default marker and deprecate main_access_key

jopemachine:
Resolves #13609

Part of the schema-truth epic #13475 (BA-7193). This piece was missing from the epic's breakdown; #13609 tracks it. It has no JIRA key yet — the title carries the epic's.

## Summary

- #13483 added `keypairs.is_default` and backfilled it, but nothing exposes it. A client holding a user's keypairs still cannot tell which one is the default — the only answer available is `users.main_access_key`, which BA-7197 deletes.
- The marker therefore has to reach the API *before* `main_access_key` can be deprecated. This does both, in that order: `is_default` ships on all three surfaces, and `main_access_key` is deprecated pointing at it.
- No new queries. `KeyPairRow.is_default` already loads with the row, so every surface reads a column it had in hand — no join, no subquery.

| Surface | Change |
|---|---|
| `data/keypair/types.py`, `models/keypair/row.py` | `KeyPairData.is_default`, carried by `to_data()` |
| v2 DTO `KeypairNode` | `is_default: bool` |
| v2 GraphQL `KeyPairV2` | `isDefault: Boolean!` — `gql_added_field` with `NEXT_RELEASE_VERSION` |
| legacy GraphQL `KeyPair` | `is_default` — `Added in 26.9.0.` |
| v2 DTO / v2 GraphQL / legacy GraphQL `main_access_key` | deprecated |

```graphql
+  """Added in 26.9.0. Whether this is the owner's default keypair."""
+  isDefault: Boolean!

-  """Primary API access key for this user."""
-  mainAccessKey: String
+  mainAccessKey: String @deprecated(reason: "Deprecated since 26.9.0. Use the keypair's is_default field.")
```

`main_access_key` keeps its value throughout the epic — deprecated here, removed from the database only in BA-7197 — so clients have a full release to move.

Based on `main` and independent of the #13484 stack, which carried the deprecation until now and rebases onto this once it merges.

## Test plan

- [x] `pants fmt fix lint check` across `src/ai/backend` — clean
- [x] Schema dumps regenerated with `scripts/generate-graphql-schema.sh` and `dump-openapi`; `openapi.json` is unchanged because the REST surface exposes only the input/filter models
- [ ] Unit and component suites in CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

----
📚 Documentation preview 📚: https://sorna--13607.org.readthedocs.build/en/13607/

----
[...]

Issue #13491, opened 2026-08-05
[Schema] Make keypairs.user_id hold the user's uuid


jopemachine:
## Objective

Leave keypairs with a single owner column named user_id holding users.uuid — the value the code already means when it says user_id.

## Background

keypairs carries two owner columns. user_id is a String(256) holding the user's email, indexed and with no foreign key; user is a GUID with a real foreign key to users.uuid. The v2 stack has already settled the ambiguity in favour of the uuid: KeyPairRow.to_data() fills KeyPairData.user_id from the user column, and both KeyPairNode.user_id and the v2 keypair DTO document the field as the owner's uuid. The column actually named user_id is the one the API does not mean.

Nothing keeps the email copy in step with users.email either, so the readers that match on it — keypair deactivation in the auth db_source, the vfolder repository join, the legacy GraphQL keypair and resource-policy queries, the stats reporter group-by — join on a mutable string instead of the primary key.

## Acceptance Criteria

- keypairs ends with one owner column, named user_id, GUID, NOT NULL, foreign key to users.uuid; the email column and the separate user column are both gone.
- A migration performs the consolidation and is verified against a populated database; the downgrade restores the email column and backfills it from users.email.
- Every reader that matched keypairs by the user's email queries by uuid instead: the auth db_source deactivation, the vfolder repository join, the legacy GraphQL keypair and resource-policy queries, and the stats reporter group-by.
- KeyPairRow.user, the relationship's foreign_keys reference and the user/keypair join helpers are renamed; KeyPairRow.to_data() reads the column directly.
- No API field changes — KeyPairNode.user_id and the v2 DTO already carry the uuid.
- Tests and fixtures that seed or delete keypairs by email are moved onto the uuid.

## Notes

- Overlaps with BA-7198, which lists user_id among the columns to tighten to NOT NULL. Land this first, or drop user_id from that issue's list.
- Roughly 350 lines: migration, model, ~16 reader sites, tests.

JIRA Issue: BA-7206

Issue #13475, opened 2026-08-05
[Schema] Make the user and keypair schema say what is actually true


jopemachine:
## Objective

Bring the users and keypairs schema in line with what the code guarantees: move the main-keypair marker to the table it describes, name the keypair owner column after the value it holds, adopt the shared timestamp mixin, make columns that are always populated NOT NULL, and delete ORM relationships nobody loads.

## Background

Typing the auth context (BA-7173) kept running into columns whose declared shape does not match reality, and each one forced a decision at the call site instead of at the schema.

- users.main_access_key points at keypairs.access_key with ON DELETE SET NULL, so deleting a keypair silently clears which keypair is the main one. The fact belongs on keypairs, not on users. 150 references across 24 modules.
- 20 of 26 users columns and 11 of 16 keypairs columns are nullable, while a fresh database shows no NULLs in most of them. Every reader then has to handle a None that never occurs, or ignore it and fail later.
- keypairs carries two owner columns: user_id, a String(256) copy of the user's email with no foreign key, and user, the GUID that actually references users.uuid. The v2 data layer, the GraphQL node and the v2 DTO all already call the uuid user_id, so the column of that name is the one the API does not mean.
- users and keypairs predate models/mixins/timestamp.py and hand-roll created_at plus modified_at, while every table written since uses LifecycleTimestampsMixin with created_at and updated_at. Two spellings of the same fact leaks into every ordering enum, filter and report.
- Four relationships (UserRow.created_endpoints, UserRow.owned_endpoints, KeyPairRow.sgroup_for_keypairs_rows, DomainRow.sgroup_for_domains_rows) have no reader outside their own row module.

## Shape of the work

Each child issue is scoped to roughly 300–400 changed lines and can be reviewed and merged on its own. The main-keypair move is expand/migrate/contract: add the new marker and write both, move readers, then drop the old column.

Every column tightened to NOT NULL must be justified by counting NULLs on a populated database first, not by reading the model.

The modified_at rename stops at the database and the model; the API gains updated_at alongside a deprecated modified_at rather than breaking clients.

JIRA Issue: BA-7193

PR #13611, opened 2026-08-07
feat(BA-7193): expose the keypair owner's email and deprecate the legacy user_id field

### LINKED ISSUE #13610: [Schema] Expose the keypair owner's email and deprecate the legacy user_id field
## Objective

Give the legacy GraphQL `KeyPair` type an honest field for the owner's email and deprecate `user_id`, before BA-7206 changes what the column behind it holds.

## Background

BA-7206 consolidates the two owner columns on `keypairs` into one named `user_id` holding `users.uuid`. Its acceptance criteria say "No API field changes — KeyPairNode.user_id and the v2 DTO already carry the uuid." That is true of the v2 stack, and false of the legacy graphene type.

`gql_legacy/keypair.py` declares `user_id = graphene.String()` and `KeyPair.from_row` fills it straight from `row.user_id` — the `String(256)` column holding the user's **email**. The uuid is exposed separately as `user = graphene.UUID()`. So the legacy type already carries both values, under names that read backwards.

When BA-7206 repoints the column, `KeyPair.user_id` silently changes from an email address to a uuid. No deprecation, no replacement, no release where a client can see both.

The value is also stale by construction — nothing keeps `keypairs.user_id` in step with `users.email` — while the legacy keypair queries already join `users` and select `users.email`, so the correct value is in hand.

## Acceptance Criteria

- Legacy GraphQL `KeyPair` gains `user_email`, sourced from the joined `users.email` rather than the `keypairs` copy, with an "Added in" description.
- `KeyPair.user_id` is deprecated with a reason naming `user_email` for the email and `user` for the owner's uuid.
- The GraphQL schema dumps are regenerated.
- No behaviour change for existing clients: `user_id` keeps returning what it returns today.

## Notes

- Lands before BA-7206, which is then free to repoint the column.
- Same shape as the marker's API pairing: ship the replacement, deprecate the old field, change the schema in a later release.
- Roughly 30 lines plus the regenerated dumps.

jopemachine:
Resolves #13610

Part of the schema-truth epic #13475 (BA-7193). Like #13607, this is an API pairing the epic's breakdown missed; it has no JIRA key yet, so the title carries the epic's.

## Summary

- BA-7206 (#13491) consolidates the two owner columns on `keypairs` into one named `user_id` holding `users.uuid`. Its acceptance criteria say *"No API field changes"* — true of the v2 stack, false of the legacy graphene type.
- `gql_legacy/keypair.py` declares `user_id = graphene.String()` and fills it from `row.user_id`, the `String(256)` column holding the owner's **email**; the uuid is exposed separately as `user`. Repointing the column would silently turn `user_id` from an email into a uuid — no deprecation, no replacement, no release where a client sees both.
- So the replacement ships first: `user_email` carries the email, and `user_id` is deprecated naming both successors.

| Change | Detail |
|---|---|
| `KeyPair.user_email` | new, `Added in 26.9.0.`, read from the joined `users.email` |
| `KeyPair.user_id` | deprecated → `user_email` for the email, `user` for the uuid |
| `KeyPair.load_all` | now selects `users.email`; the other three loaders already did |

The email comes from the joined `users` row, not the `keypairs` copy — nothing keeps that copy in step with `users.email`, and the loaders already join `users` on `keypairs.user == users.uuid`.

```graphql
-  user_id: String
+  user_id: String @deprecated(reason: "Deprecated since 26.9.0. Holds the owner's email address; use user_email for the email or user for the owner's uuid.")
+
+  """Added in 26.9.0. The owner's email address, read from the users table."""
+  user_email: String
```

No behaviour change for existing clients: `user_id` keeps returning what it returns today. BA-7206 is then free to repoint the column.

## Test plan

- [x] `pants fmt fix lint check` across `src/ai/backend` — clean
- [x] `gql_legacy.keypair` imports; GraphQL schema dumps regenerated with `scripts/generate-graphql-schema.sh`
- [x] Checked every `KeyPair.from_row` caller for whether its query provides `users.email` — `load_all` did not, and now does

### DIFF
--- a/changes/13611.feature.md
+++ b/changes/13611.feature.md
@@ -1,0 +1,1 @@
+Report the keypair owner's email as `user_email` on the legacy GraphQL type and deprecate `user_id`, which currently holds the email under a name that means the owner's uuid everywhere else
--- a/docs/manager/graphql-reference/schema.graphql
+++ b/docs/manager/graphql-reference/schema.graphql
@@ -1202,7 +1202,10 @@ type UserList implements PaginatedList {
 
 type KeyPair implements Item {
   id: ID
-  user_id: String
+  user_id: String @deprecated(reason: "Deprecated since 26.9.0. Holds the owner's email address; use user_email for the email or user for the owner's uuid.")
+
+  """Added in 26.9.0. The owner's email address, read from the users table."""
+  user_email: String
   full_name: String
   access_key: String
   secret_key: String
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -9501,7 +9501,10 @@ type KeyPair implements Item
   @join__type(graph: GRAPHENE)
 {
   id: ID
-  user_id: String
+  user_id: String @deprecated(reason: "Deprecated since 26.9.0. Holds the owner's email address; use user_email for the email or user for the owner's uuid.")
+
+  """Added in 26.9.0. The owner's email address, read from the users table."""
+  user_email: String
   full_name: String
   access_key: String
   secret_key: String
--- a/src/ai/backend/manager/api/gql_legacy/keypair.py
+++ b/src/ai/backend/manager/api/gql_legacy/keypair.py
@@ -97,7 +97,12 @@ class KeyPair(graphene.ObjectType):  # type: ignore[misc]
     class Meta:
         interfaces = (Item,)
 
-    user_id = graphene.String()
+    user_id = graphene.String(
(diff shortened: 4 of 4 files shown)


jopemachine:
Closing — not planned.

The premise was that consolidating the two owner columns would silently turn the legacy `KeyPair.user_id` from an email into a uuid, so a replacement had to ship first. That consolidation is not needed: `keypairs."user"` already holds the owner's uuid, so `user_id` can stay the email column it is, and the legacy field keeps meaning what it always meant.

No deprecation, no `user_email` successor, no release window required.

PR #13615, opened 2026-08-07
feat(BA-7283): expose resource group is_default through the create/update API

fregataa:
## Summary
- `scaling_groups.is_default` landed schema-only in BA-6824, so the default resource group could only be changed by writing to the database directly. This wires the flag through the write paths: the v2 create/update DTOs, the adapter, `ScalingGroupCreatorSpec` / `ScalingGroupStatusUpdaterSpec`, the GraphQL create/update inputs, and the admin CLI (`create --is-default` plus a new `update` command).
- The partial unique index still permits a single default, so promoting a group while another one holds the flag is rejected — the default is switched by clearing the old one first. `ScalingGroupUpdaterSpec` now carries an `integrity_error_checks` entry that maps the `uq_scaling_groups_is_default` violation to a 400 (`DefaultScalingGroupAlreadyExists`) instead of leaking the raw unique-violation error.
- The CLI `update` command targets `PATCH /v2/resource-groups/{name}/config` rather than the plain update endpoint, because the latter treats an omitted `description` as "clear it" (see the note below).

## Test plan
- [x] `pants test tests/unit/manager/repositories/scaling_group::` — four real-DB cases covering `is_default` true/false against another default existing/not existing
- [x] `pants test tests/unit/common/dto/manager/v2/resource_group::`
- [x] Verify against a live server: `./bai admin resource-group update <name> --is-default`, then confirm the second promotion returns 400 until the first is cleared

## Note for reviewers
Two pre-existing issues were left untouched here:
- `ResourceGroupAdapter.update()` inverts the SENTINEL convention for `description` (SENTINEL clears, `None` keeps), the opposite of every other adapter. Combined with the SDK's `exclude_unset` serialization, a PATCH that omits `description` wipes it.
- `ScalingGroupCreatorSpec.integrity_error_checks` matches any unique violation without filtering on the constraint name, so creating a group with `is_default=true` while another default exists reports "Duplicate scaling group name".

Resolves BA-7283

----
📚 Documentation preview 📚: https://sorna--13615.org.readthedocs.build/en/13615/

----
📚 Documentation preview 📚: https://sorna-ko--13615.org.readthedocs.build/ko/13615/

### DIFF
--- a/src/ai/backend/client/cli/v2/admin/resource_group.py
+++ b/src/ai/backend/client/cli/v2/admin/resource_group.py
@@ -114,7 +114,8 @@ def get(name: str) -> None:
 @click.option("--domain-name", required=True, help="Domain name.")
 @click.option("--description", default=None, help="Description.")
 @click.option(
-    "--is-default/--no-is-default",
+    "--set-default/--unset-default",
+    "is_default",
     default=False,
     help="Make this the default resource group. Clear the current default first.",
 )
@@ -154,7 +155,8 @@ def create(name: str, domain_name: str, description: str | None, is_default: boo
     help="Whether the resource group is visible to all users.",
 )
 @click.option(
-    "--is-default/--no-is-default",
+    "--set-default/--unset-default",
+    "is_default",
     default=None,
     help="Make this the default resource group. Clear the current default first.",
 )
--- a/src/ai/backend/client/cli/v2/admin/resource_group.py
+++ b/src/ai/backend/client/cli/v2/admin/resource_group.py
@@ -113,7 +113,12 @@ def get(name: str) -> None:
 @click.option("--name", required=True, help="Resource group name.")
 @click.option("--domain-name", required=True, help="Domain name.")
 @click.option("--description", default=None, help="Description.")
-def create(name: str, domain_name: str, description: str | None) -> None:
+@click.option(
+    "--is-default/--no-is-default",
+    default=False,
+    help="Make this the default resource group. Clear the current default first.",
+)
+def create(name: str, domain_name: str, description: str | None, is_default: bool) -> None:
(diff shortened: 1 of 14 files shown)


src/ai/backend/client/cli/v2/admin/resource_group.py
```diff
@@ -113,7 +113,12 @@ async def _run() -> None:
 @click.option("--name", required=True, help="Resource group name.")
 @click.option("--domain-name", required=True, help="Domain name.")
 @click.option("--description", default=None, help="Description.")
-def create(name: str, domain_name: str, description: str | None) -> None:
+@click.option(
+    "--is-default/--no-is-default",
```
HyeockJinKim: no is default is so wierd

src/ai/backend/client/cli/v2/admin/resource_group.py
```diff
@@ -125,6 +131,58 @@ async def _run() -> None:
...
+@click.option(
+    "--is-active/--no-is-active",
+    default=None,
+    help="Whether the resource group accepts new sessions.",
+)
+@click.option(
+    "--is-public/--no-is-public",
+    default=None,
+    help="Whether the resource group is visible to all users.",
+)
```
seedspirit, line 158: Is prefix seems it returns bool result. Are there any other good ways to say this?

HyeockJinKim: I think we should keep it simple here, with options like "set active," "set inactive," "set public," and "unset public."

fregataa: I changed them to set-active/unset-active, set-public/unset-public

HyeockJinKim (approved):

PR #13654, opened 2026-08-09
feat(BA-7308): add configurable preemption victim scope to resource groups

HyeockJinKim:
## Summary
- Add `PreemptionVictimScope` (`user` / `project` / `domain` / `resource-group`, default `user`) to the resource group preemption config, letting admins choose which sessions may become preemption victims for a pending session. The JSON `scheduler_opts` column needs no migration and existing groups keep the current per-user behavior.
- Generalize the victim-candidate snapshot from per-user to per-scope-key: `WorkloadMeta.preemption_scope_key` is the single key derivation shared by the loader's grouping and the provisioner's lookup, and the candidate query matches/groups on the scope's session column (`RESOURCE_GROUP` degenerates to one group-wide key with no matching column). The strict `job_priority` comparison is unchanged and applies within every scope.
- Expose `victim_scope` through the v2 DTOs, GraphQL types (marked `Added in 26.8.1.`), the adapter, and the scheduler-config updater. `gql_pydantic_type` now restores field descriptions declared with `gql_field`/`gql_added_field` over the DTO descriptions, so field-level version markers render in the SDL.

## Test plan
- [x] Scope scenarios against a real DB: unset/user default, project (same project other users in, other projects out), domain (cross-project in, other domains out), resource-group (whole group in), plus the equal-or-higher-priority exclusion in every scope (`test_preemption_candidates.py`)
- [x] Preemption config update round-trip including `victim_scope` (`test_scaling_group_repository.py`, `test_updaters.py`)
- [x] Selection-layer behavior unchanged under the renamed snapshot types (`test_preemption_selection.py`)

Backport: 26.8

Resolves BA-7308

----
📚 Documentation preview 📚: https://sorna--13654.org.readthedocs.build/en/13654/

----
📚 Documentation preview 📚: https://sorna-ko--13654.org.readthedocs.build/ko/13654/

### DIFF
--- a/src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
@@ -43,6 +43,7 @@ from ai.backend.common.types import (
     ClusterMode,
     KernelId,
     PreemptionMode,
+    PreemptionVictimScope,
     ResourceSlot,
     SessionId,
     SessionTypes,
@@ -173,14 +174,15 @@ from ai.backend.manager.views.sokovan.snapshot import (
     ResourceLimit,
     ResourceOccupancySnapshot,
     ResourcePolicySnapshot,
+    ScopeVictimCandidates,
     SessionDependencySnapshot,
     SlotAllocation,
     UserResourceAllocation,
     UserResourceLimit,
-    UserVictimCandidates,
 )
 from ai.backend.manager.views.sokovan.workload import (
     KernelWorkload,
+    PreemptionScopeKey,
     ResourceRequest,
     SessionDependencyInfo,
     SessionGroupPolicy,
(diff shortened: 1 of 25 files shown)


src/ai/backend/manager/repositories/scheduler/db_source/db_source.py
```diff
@@ -1076,10 +1096,58 @@ async def _fetch_preemption_candidates(
...
+            )
+        return _PreemptionScopeMatch(
+            scope_value_columns=[scope_column.label("scope_value")],
+            candidate_filter=sa.or_(
+                *(
+                    sa.and_(scope_column == key, s.c.job_priority < threshold)
+                    for key, threshold in max_pending_priority.items()
+                )
+            ),
+            group_wide_key=None,
```
fregataa, line 1150: It seems possible to remove this `if` branch

HyeockJinKim: I made this change because I thought making an explicit distinction would improve visibility—what other approaches are possible?

fregataa (approved):

seedspirit (approved):

PR #13661, opened 2026-08-10
feat(BA-7274): emit notification event before idle checkers reclaim a session

seedspirit:
## Summary
- Add a `session.idle-timeout.warning` notification rule type: when a session's remaining time before idle-based termination (network_timeout / session_lifetime checkers) drops below `idle.warning_threshold` (seconds, default 300, 0 disables), the legacy `IdleCheckerHost` emits a `NotificationTriggeredEvent` so admins can route warnings through existing notification rules/channels (webhook, email) in the Control Panel.
- Warn once per approach to the deadline: a Valkey warned-marker (`session.{id}.idle_warning_sent`, managed via new `ValkeyLiveClient` methods) dedupes across check ticks and is cleared when user activity resets the timer, re-arming the warning.
- The message payload carries session name/id, owner uuid/email, the triggering checker, remaining seconds, and the expected termination timestamp as Jinja2 template variables; utilization (grace-period type) remainings are excluded since they carry no firm deadline.

## Test plan
- [x] Unit tests: warning fired below threshold (payload verified), deduped while the marker is set, marker cleared on recovery, grace-period remainings ignored (`tests/unit/manager/test_idle_check_host.py`)
- [x] Live E2E on local dev: created an email rule/channel via admin GQL, session with `max_session_lifetime=120s` → exactly one rendered warning mail received over SMTP, session reclaimed on schedule
- [x] `pants fmt/lint/check/test` on changed files; notification service/adapter/repository/DTO suites pass

Resolves BA-7274

----
📚 Documentation preview 📚: https://sorna--13661.org.readthedocs.build/en/13661/

----
📚 Documentation preview 📚: https://sorna-ko--13661.org.readthedocs.build/ko/13661/


configs/manager/sample.toml
```diff
@@ -692,6 +692,14 @@
...
   # Added in 25.8.0
   checkers = {}
+  # Seconds before an idle checker reclaims a session to emit a 'session.idle-
+  # timeout.warning' notification event. The event fires once per session when
+  # its remaining time first drops below this threshold, and again if user
+  # activity resets the timer and it later drops below the threshold.
+  # Administrators subscribe to the event via notification rules. 0 disables the
+  # warning.
+  # Added in 26.9.0
+  warning_threshold = 600
```
HyeockJinKim, line 702: I have no intention of adding a config file.

PR #13663, opened 2026-08-10
fix(BA-7252): honor X-Forwarded-URL only from trusted proxies

fregataa:
## Summary

- `sign_request()` replaced the HMAC signature host and path with the client-supplied `X-Forwarded-URL` header without any trust check, so a client could choose the host its signature is bound to (defeating endpoint binding) and have the signature cover a path other than the one actually routed. The header is now honored only when the connection peer matches `manager.trusted-proxies`; an untrusted origin is ignored rather than rejected, so a misconfigured proxy cannot turn otherwise valid requests into hard failures.
- Enabling `manager.trusted-proxies` was itself unusable, which made that gate unreachable. `XForwardedStrict([trusted_proxies])` declares exactly **one** trusted hop regardless of how many entries the setting holds, and `remote_ip()` rejects any request whose hop count differs — so every request without exactly one `X-Forwarded-For` entry answered `400`, including direct calls and any deployment with two or more proxies in front. The middleware was also registered with `app.middlewares.append()`, placing it *inside* the auth middleware, so `request.remote` was still the peer address while authentication ran and `allowed_client_ip` was matched against the proxy.
- The middleware is dropped and the address is resolved where it is consumed: walk the `X-Forwarded-For` chain inwards from the connection peer and take the first address that is not a trusted proxy. Any hop count works, a request without the header keeps the peer address, and a client that forges the header while connecting directly is caught by its own peer address ending the walk.
- When `manager.trusted-proxies` is empty the previous behavior is kept byte-for-byte and a deprecation warning is logged once per process, so this is safe to backport.

## Test plan

- [x] `pants test tests/unit/manager/api/test_auth_trusted_proxy.py` — 19 unit tests: CIDR/wildcard parsing, trusted-peer decision, signature equivalence for honored vs ignored header, once-only deprecation warning, and client-IP resolution over 0/1/N hops including a forged header from an untrusted peer
- [x] `pants test tests/unit/manager/api/test_auth_params.py tests/unit/manager/api/auth::` — existing auth tests still pass
- [x] `pants fmt`, `pants lint`, `pants check`
- [x] Live verification against a local manager, HMAC-signed requests over three configurations:

  | request | `trusted-proxies` unset | peer **not** trusted | peer trusted |
  |---|---|---|---|
  | correctly signed, no forwarded headers | 200 | 200 | 200 |
  | injected `X-Forwarded-URL`, signed with the forwarded host+path | 200 (legacy, warns once) | **401** | 200 |
  | `./bai admin domain search` | 200 | 200 | 200 |

- [x] Client IP resolution against the live manager (`GET /auth/my-ip`, trusted `["127.0.0.1/32", "10.0.0.0/8"]`, peer `127.0.0.1`):

  | `X-Forwarded-For` | resolved |
  |---|---|
  | (none) | `127.0.0.1` |
  | `203.0.113.9` | `203.0.113.9` |
  | `203.0.113.9, 10.0.0.5, 10.0.0.6` | `203.0.113.9` |
  | `203.0.113.9, 198.51.100.4, 10.0.0.5` | `198.51.100.4` |
  | `1.2.3.4` forged from an untrusted peer | peer address, header ignored |

  On `main` the same configuration answered `400` to every one of these except the single-entry case.

Resolves BA-7252

### DIFF
--- a/src/ai/backend/manager/api/rest/middleware/auth.py
+++ b/src/ai/backend/manager/api/rest/middleware/auth.py
@@ -22,7 +22,7 @@ import ipaddress
 import logging
 import secrets
 import uuid
-from collections.abc import Awaitable, Callable, Mapping
+from collections.abc import Awaitable, Callable, Iterable, Mapping
 from contextlib import ExitStack
 from datetime import datetime, timedelta
 from typing import TYPE_CHECKING, Any, Final
@@ -69,6 +69,9 @@ if TYPE_CHECKING:
     from ai.backend.manager.models.utils import ExtendedAsyncSAEngine
 
 log: Final = BraceStyleAdapter(logging.getLogger(__spec__.name))
+
+TRUSTED_PROXY_NETWORKS_KEY: Final = "_trusted_proxy_networks"
+FORWARDED_URL_HEADER: Final = "X-Forwarded-URL"
 
 _whois_timezone_info: Final = {
     "A": 1 * 3600,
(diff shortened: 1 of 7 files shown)


src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -350,6 +353,93 @@ def _extract_auth_params(request: web.Request) -> tuple[str, str, str] | None:
...
+    An untrusted origin is ignored rather than rejected so that a misconfigured
+    proxy cannot turn otherwise valid requests into hard failures.
+    """
+    upstream_url = request.headers.get(FORWARDED_URL_HEADER)
+    if upstream_url is None:
+        return None
+    if not request.config_dict.get(TRUSTED_PROXY_NETWORKS_KEY):
+        _warn_forwarded_url_without_trusted_proxies()
+        return upstream_url
+    if not is_from_trusted_proxy(request):
```
seedspirit, line 434: If it's result is None, will the request still go through but fail at the authentication stage?

fregataa: `None` result means that the header of the request cannot be overridden. And if the request has not trusted host path, manager will raise 400 error

seedspirit (approved):

PR #13664, opened 2026-08-10
release: 26.8.1

fregataa:
## Summary

Cut the `26.8.1` patch release on the `26.8` branch.

- Bump `VERSION` to `26.8.1`. WebUI is not re-bundled (stays at the `26.8.0` bundle).
- Consume the 9 `changes/` news fragments accumulated since `26.8.0` — 1 feature, 8 fixes.
- Regenerate the OpenAPI reference (version string only). Sample configs and the GraphQL schema regenerate byte-identical, so they carry no diff.
- Update the external tool binaries (`bssh`), committed separately.

## Changelog organization

Flat bullet list under `## 26.8.1`, matching the format of previous patch releases (`26.4.1`, `26.4.2`, `26.4.3`). No RC was cut for `26.8.1`, so there is nothing to consolidate and no `####` subsection grouping is warranted at this entry count.

## Notes

- `NEXT_RELEASE_VERSION` handling is a no-op here: the placeholder has no usages outside `common/meta/`, and the auto-advance step is skipped for patch releases (patch != 0), so `meta.py` still points at `26.9.0`.
- `scripts/release.sh` has no `set -e`/`pipefail`, so when the exported virtualenv is missing it silently skips towncrier, sample-config generation, the OpenAPI dump, and the GraphQL schema generation while still exiting `0`. That happened on the first run here; the steps were re-run after `pants export` and folded into the release commit. Hardening the script is left as a follow-up on `main`.

**Checklist:**

- [x] Documentation (CHANGELOG, regenerated API references)


fregataa:
Closing: the 26.8.1 release needs to bundle a WebUI release as well. Will re-cut the release branch with the WebUI version included.

PR #13675, opened 2026-08-10
feat(BA-7317): render prometheus query preset templates from named placeholders

seedspirit:
## Summary

- Render `query_template` with a sandboxed Jinja environment whose variable delimiters are `${{` `}}` instead of `{{` `}}`; placeholders are `${{labels}}`, `${{window}}`, `${{group_by}}`, and every other character is literal PromQL — no escaping. PromQL uses braces for label selectors, so the default delimiters collide with the query language; `${` alone would collide too (a `$` regex anchor before a `{n}` quantifier), while `${{` cannot occur in literal PromQL. Replaces the `str.format` engine and its `escape_non_placeholders` regex heuristic.
- Rendering moves off `MetricPreset.render()` into `PromQLTemplateRenderer` (compile cache of 256 templates), which also owns validation: an AST whitelist (literal text + variable substitution only — no filters, attribute access, statements) plus `StrictUndefined`, and an explicit rejection of legacy `{labels}` / `{{{` syntax with a message pointing at the new form. `MetricPreset` is now pure data. The renderer is injected into `PrometheusQueryPresetService` via `services/factory.py`, and constructed by `PrometheusClient`.
- Validation runs in the service layer (create / modify / preview) instead of in five duplicated DTO validators across v1 and v2; `dto/manager/v2/prometheus_query_preset/validators.py` is deleted. It is server-side only now — the client SDK no longer rejects templates locally.
- Migration `4b8e2f7a91d3` rewrites every stored preset, seeded defaults included. It keeps a frozen copy of the legacy parsing logic: plain substitution breaks on legacy-valid shapes such as `metric{{mode!="idle",{labels}}}`. Idempotent — already-converted rows are skipped; `downgrade()` is a no-op.
- The 12 code-embedded fixed templates in `fixed_query_builder.py` (container metric / live stat) and the example fixture JSON are converted to the new syntax; the idle-checker config schema and GQL/REST descriptions are updated with the regenerated schema and OpenAPI docs.

Known limitation: a template escaped past validity under the old engine, e.g. `m{{{{job="x"}}}}`, converts to `m{{job="x"}}` — the same broken PromQL the old engine rendered. Such rows were never valid.

## Test plan

- [x] `pants fmt`/`fix`/`lint`/`check`, and `pants test --changed-since=origin/main --changed-dependents=direct` (27 targets, component tests included)
- [x] Migration against a live PostgreSQL: 31 legacy shapes inserted as real rows — seeded defaults, bare `{labels}`, escaped literals, static-and-injected selectors, regex quantifiers, `$` anchors, repeated placeholders — every converted row renders byte-identical PromQL to the legacy engine; re-running leaves converted rows untouched
- [x] `./bai`: legacy syntax, unknown placeholder names, filters and statements each rejected with a 400; `execute` on a migrated preset returns data
- [x] All 12 fixed live-stat/container-metric queries accepted by a real Prometheus

Resolves BA-7317

----
📚 Documentation preview 📚: https://sorna--13675.org.readthedocs.build/en/13675/

----
📚 Documentation preview 📚: https://sorna-ko--13675.org.readthedocs.build/ko/13675/


fregataa (commented):
let's add breaking changelog

src/ai/backend/common/dto/manager/v2/prometheus_query_preset/validators.py
```diff
@@ -1,66 +1,70 @@
...
+            "Legacy str.format template syntax is no longer supported; "
+            f"use {{{{ labels }}}}, {{{{ window }}}}, {{{{ group_by }}}}: {template!r}"
+        )
+    try:
+        ast = PROMQL_TEMPLATE_ENV.parse(template)
+    except TemplateSyntaxError as e:
+        raise InvalidMetricPresetTemplate(f"Invalid template syntax ({e}): {template!r}") from e
+    for node in _walk(ast):
+        if not isinstance(node, _ALLOWED_NODE_TYPES):
+            raise InvalidMetricPresetTemplate(
```
HyeockJinKim: Why is this implementation here? + Since the implementation places `jinja` in the `global` scope, the location of the implementation seems off—please adjust it properly. Even in the Notification Center and other places, values for the implementation are provided in fields, so please do not place variables in the `global` scope.

HyeockJinKim (changes requested):

src/ai/backend/common/dto/manager/v2/prometheus_query_preset/validators.py
HyeockJinKim: The DTO package should not contain any business logic.

src/ai/backend/manager/models/alembic/versions/4b8e2f7a91d3_convert_prometheus_query_preset_templates_to_jinja.py
```diff
@@ -0,0 +1,105 @@
...
+        return template
+    has_placeholder = False
+    for _literal, field, _spec, _conv in parsed:
+        if field in ("labels", "window", "group_by"):
+            has_placeholder = True
+            break
+    if not has_placeholder:
+        try:
+            jinja2.Environment().parse(template)
+            return template
```
HyeockJinKim: I'm not sure if I need iLogic for this—can't I just run the format command and replace the values with {{...}}?

seedspirit: These helpers are a frozen copy of the removed legacy renderer (`escape_non_placeholders(template).format(...)`).

Plain substitution breaks on legacy-valid templates like `metric{{mode!="idle",{labels}}}`: replacing only `{labels}` leaves `{{mode!="idle"` behind, which Jinja rejects — the legacy `{{`/`}}` escapes must also be unfolded into single braces. `string.Formatter().parse()` does exactly that, with the same parser the legacy renderer used.

Verified the converted output renders identically to the legacy renderer for all seeded presets and the shape above.

fixtures/manager/example-prometheus-query-presets.json
```diff
@@ -7,7 +7,7 @@
       "rank": 100,
       "category_name": "container",
       "metric_name": "backendai_container_utilization",
-      "query_template": "sum by ({group_by})(backendai_container_utilization{{{labels}}})",
+      "query_template": "sum by (${group_by})(backendai_container_utilization{${labels}})",
```
HyeockJinKim: If we're going to do this, wouldn't it be better to just use ${{, just like on GitHub?

HyeockJinKim (approved):

PR #13687, opened 2026-08-11
refactor(BA-7314): model message and event payloads with Pydantic

fregataa:
## Summary

Prepares the message payloads for the msgpack-to-JSON event serialization cutover. Nothing on the wire changes, so this is safe to deploy in any order.

- **`AnycastPayload` / `BroadcastPayload` are now independent models.** The shared `MessagePayload` base is gone — the two differ in transport, wire encoding, and retry state, so the base was hiding more than it saved.
- **New `payload` field** carrying the serialized event arguments in the form each transport takes them (`bytes` for a stream field, `str` for a JSON message). Producers do not fill it yet, so it stays `None`; giving it a wire key is the next step.
- **`source` / `body` renamed to `legacy_source` / `legacy_body`**, keeping their original keys via `validation_alias`. Every producer and decoder keeps constructing them as `source=` / `body=` unchanged.
- **`MessageMetadata` is a frozen Pydantic model** instead of a dataclass. Field validation replaces the hand-written decoder, and `extra="forbid"` keeps the strictness that `cls(**result)` used to give. The legacy top-level `user_id` is declared as an excluded field so an older producer's message still parses.

## Test plan

- [x] `pants check ::` — full run, since a symbol was removed and fields renamed
- [x] `pants test tests/unit/common/message_queue:: tests/unit/common/events::` (CI)
- [x] Anycast and broadcast round-trips unchanged: `to_stream_fields()` / `from_stream_fields()` and `to_json()` / `from_json()`
- [x] Metadata round-trip, legacy `user_id` accepted and never written back, unknown keys rejected as `InvalidMessagePayloadError`

Resolves BA-7314


src/ai/backend/appproxy/common/events.py
```diff
@@ -6,18 +6,62 @@
...
+class CheckWorkerLostEventPayload(AnycastEventPayload):
+    """The worker-lost check carries no arguments."""
+
+
+class CheckUnusedPortEventPayload(AnycastEventPayload):
+    """The unused-port check carries no arguments."""
+
+
+class ReconcileTraefikRoutesEventPayload(AnycastEventPayload):
+    """The traefik reconcile trigger carries no arguments."""
```
HyeockJinKim: It makes sense to separate the values read from the message queue from those read from the event dispatcher, but it seems odd that the values are being declared redundantly. It seems strange that there’s something like `to_payload`.

src/ai/backend/common/events/AGENTS.md
```diff
@@ -5,8 +5,10 @@ Event classes live in `event_types/{domain}/{anycast|broadcast}.py`; base types
 ## Define an event
 
 - Subclass `AbstractAnycastEvent` (one consumer) or `AbstractBroadcastEvent` (all subscribers) from `types.py`. The base class — not a flag — decides the delivery pattern.
+- An event is a Pydantic model: declare its body as annotated fields, and construct it by keyword. Do NOT use `@dataclass`.
 - Implement `event_name()` (unique snake_case), `event_domain()` (an `EventDomain` value), and `serialize()` / `deserialize()`.
 - **`serialize()` and `deserialize()` must use the same tuple order** — a mismatch silently swaps fields.
+- Do NOT write `to_message()` / `from_message()`. `AbstractEvent` derives both from the model's fields, so the body an event carries is its field set and nothing else. Override them only to refuse — an event that must never cross the queue raises there, as `BgtaskAlreadyDoneEvent` does.
```
HyeockJinKim, line 11: I think it would be better to just add "final" to this.
Instead of saying "Do NOT use `@dataclass`," a guideline simply stating "Use PyDantic" would suffice.

HyeockJinKim (approved):
1. First, remove the source and context from the event dispatcher.
2. Implement the registration of event handlers by group (refer to the processor group for guidance).
3. Please ensure that only one item is registered, even if multiple items are registered in the event dispatcher.
---
It seems you should proceed with the work in that order.

PR #13691, opened 2026-08-11
refactor(BA-7193): name get_default_keypair_row after the relationship it reads

jopemachine:
## Summary

`UserRow.get_main_keypair_row()` reads `UserRow.default_keypair` — the relationship joined on `keypairs.is_default`, renamed in #13484. The method name still says `main_keypair`, so it points at `users.main_access_key`, which it has not read since that PR.

```
UserRow.get_main_keypair_row  ->  get_default_keypair_row
```

Its one caller (`services/auth/service.py`) and that caller's local follow.

## What this deliberately does not touch

The first draft of this PR renamed the whole `main_access_key` vocabulary and that was wrong. Anything carrying `users.main_access_key` keeps its name, because renaming it makes the name claim the marker while the code still reads the legacy column:

| Kept | Why |
|---|---|
| `UserData.main_access_key` | reads `users.main_access_key`, feeds the API field of the same name |
| `SwitchMyMainAccessKeyAction`, `switch_my_main_access_key` (service, processor, adapter, repository, db_source) | back the live `switchMyMainAccessKey` operation one-for-one |
| `_validate_and_update_main_access_key` | writes that column, from an input of that name |
| `UserRow.main_access_key` | names the column |

A `switch_my_default_access_key` would be *added* next to the existing operation rather than renamed into it, and belongs with the API operation it backs — not here. The column and everything named after it retire together in BA-7197.

## Verification

- [x] `pants fmt lint` — clean
- [x] `pants check` across the change and its transitive dependents — 2549 files, no issues
- [x] Nothing under `docs/` changes: the GraphQL dumps and `openapi.json` are untouched
- [ ] Unit suites in CI


jopemachine:
Folded into #13531 (12c6560) rather than shipping separately — the method reads the relationship that PR's column removal is about, so the rename belongs with it.

PR #13695, opened 2026-08-11
fix(BA-7323): define the roles table locally in RBAC role migrations

fregataa:
## Summary

- `alembic upgrade` aborted at revision `09206ac04fd3` with `CompileError: Unconsumed column names: status`, leaving any database below that revision unable to upgrade at all.
- The shared frozen `get_roles_table()` never declared `status`, while `RoleCreateInput.to_dict()` supplies it. The mismatch was masked because the migration helpers shared the production metadata with `extend_existing=True` and thus handed back the production `roles` table; commit 213ad6856b isolated that metadata and the borrowed column disappeared.
- Each affected revision now snapshots the `roles` columns it touches on a file-local `sa.MetaData()`, so the definition matches the schema at that point in the chain and cannot drift with the production model again. `09206ac04fd3` also gets a local get-or-create in place of `PermissionUpdateUtil.get_or_create_role()`.
- `2c9000848b6e` raised the same error via `insert_skip_on_conflict()`; `e3fb172166dc` did not raise but silently discarded `status` through the executemany form, which drops unconsumed keys.

## Test plan

- [x] Reproduced the reported failure on a scratch database stamped at `de1032a11cca` — same `CompileError`, transaction rolled back, revision unchanged.
- [x] With the change, `upgrade 09206ac04fd3` succeeds and writes `role_superadmin` / `role_monitor` with `source=system`, `status=active`, plus the expected `user_roles` mappings and 44 permission rows.
- [x] `2c9000848b6e` single-step run creates the missing user system role with its permissions and user-role mapping.
- [x] `e3fb172166dc` single-step run creates one `project-{id8}-member` role per project with its scope binding and 11 permissions each.
- [x] Re-running each revision after re-stamping its parent produces no duplicate rows.
- [x] `e3fb172166dc` legacy path: a project holding a legacy `role_project_{id8}_member` with `source=custom` has that row promoted to `system` in place with no second member role created, while a project without one still gets a fresh member role and its 11 permissions.
- [x] `2c9000848b6e` across pages: 254 users backfilled in a single run (`page_size=100`) — 254 roles, 254 mappings, 1778 permission rows, no duplicate role names, no user left without a system role.
- [x] Confirmed `permissions.permission` (NOT NULL, no default) is introduced after both revisions by `c6648c039bd4`, so the verification schema matched the chain at that point instead of being more permissive.
- [x] `pants fmt` / `pants fix` / `pants lint --changed-since=origin/main` pass.

Resolves BA-7323

### DIFF
--- a/src/ai/backend/manager/models/alembic/versions/e3fb172166dc_backfill_project_member_roles.py
+++ b/src/ai/backend/manager/models/alembic/versions/e3fb172166dc_backfill_project_member_roles.py
@@ -39,12 +39,10 @@ from ai.backend.manager.models.rbac_models.migration.enums import (
 )
 from ai.backend.manager.models.rbac_models.migration.models import (
     get_association_scopes_entities_table,
-    get_roles_table,
     mapper_registry,
 )
 from ai.backend.manager.models.rbac_models.migration.utils import (
     insert_skip_on_conflict,
-    query_role_rows_by_name,
 )
 
 # revision identifiers, used by Alembic.
@@ -53,6 +51,10 @@ down_revision = "d8e4f2a1b3c7"
 # Part of: 26.5.0
 branch_labels = None
 depends_on = None
+
+# Metadata local to this revision so that the table snapshots below cannot be
+# altered by definitions living in other revisions.
+_local_metadata = sa.MetaData()
 
 
 # Snapshot of MEMBER_ACCESSIBLE_ENTITY_TYPES_IN_PROJECT at the time of writing.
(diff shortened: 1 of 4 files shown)


src/ai/backend/manager/models/alembic/versions/e3fb172166dc_backfill_project_member_roles.py
```diff
@@ -167,7 +182,7 @@ def _create_member_roles_for_projects(
     if not project_ids:
         return {}
 
-    roles_table = get_roles_table()
```
seedspirit, line 170: Will the legacy `get_roles_table` be left in place? Since there are no longer any references to it, it seems safer to just delete it.

fregataa: it is still used somewhere else. but I dont want to replace all of them right now

seedspirit (approved):

PR #13708, opened 2026-08-11
refactor(BA-7297): migrate the manager's pass-through actions to ops-generic wiring

HyeockJinKim:
## Summary

Moves the manager's pass-through actions onto ops-generic wiring: an operation whose
service only forwarded a spec to a repository that only forwarded it to db_source now
carries the spec on the action and runs against the generic ops services, and the two
forwarding layers go. Fifteen domains, ninety-six actions.

The sweep kept turning up things that had to be settled before the next domain could
move, so the branch carries those too:

- **Framework** — a global get factory, a global create composed with the field rows it
  owns, RESTORE wired through the v2 bases, and the many-row writes renamed for their
  failure mode (`atomic_*` raises, `partial_bulk_*` answers per item).
- **Declarations** — an operation is named for what it does to the row. A write that
  sets a lifecycle flag is a delete; one that removes the row is a purge. Fourteen
  actions were declaring the wrong one, including four that removed rows while claiming
  a soft delete. `Modify` becomes `Update` everywhere, matching the operation the
  action already declared.
- **Projections** — a `data/` type that mirrors one row may not nest another entity.
  Notification rules name their channel by id; model cards no longer carry their
  resource requirements. Both compose one layer up instead, which is what freed those
  domains for the v2 specs at all.
- **Two security fixes** — a resource policy node reached the keypairs sharing it, so
  any authenticated user could enumerate other users; and the legacy schema loaded any
  named policy with no role check. Neither was introduced here.
- **Tests** — eleven files referenced modules the migrations had deleted, hidden because
  `pants check --changed-since` never reached files this branch had not edited. Repo-wide
  `pants check ::` finds them; that is the gate this branch used from then on.

## Test plan

- [ ] `pants check ::` and `pants lint ::` pass repo-wide
- [ ] `test_registry_catalog.py` — every v2 action is wired, names are unique and snake_case
- [ ] Component suites for the migrated domains (storage, object storage, vfs storage,
      notification, resource policy, app config, prometheus preset, error log)
- [ ] Model card repository suite — requirements read and write through their own table
- [ ] Verify against a live server that the migrated reads and writes still answer, and
      that `minResource` resolves on demand in GraphQL

Resolves BA-7297

----
📚 Documentation preview 📚: https://sorna--13708.org.readthedocs.build/en/13708/

----
📚 Documentation preview 📚: https://sorna-ko--13708.org.readthedocs.build/ko/13708/


tests/component/operations/test_operations.py
```diff
@@ -41,21 +43,20 @@ async def test_admin_appends_error_log(
...
-                context_env="{}",
+        with pytest.raises(PermissionDeniedError):
+            await user_registry.operations.append_error_log(
+                AppendErrorLogRequest(
+                    severity=ErrorLogSeverity.WARNING,
+                    source="test-user",
+                    message="Test warning from user",
+                    context_lang="python",
+                    context_env="{}",
+                )
```
fregataa: Is the change of results of the same action okay?

Issue #13582, opened 2026-08-07
Publish Backend.AI service Docker images via GitHub Actions with Docker Hub OIDC authentication


kyujin-cho:
## Objective

Automatically bake and publish the Backend.AI service Docker images from `docker/*.dockerfile` to the `lablup` Docker Hub organization on every release tag, authenticating with **Docker Hub OIDC connections** (no stored PAT/password secrets in GitHub).

Reference: [Docker OIDC connections for GitHub Actions](https://www.docker.com/blog/docker-oidc-connections-for-github-actions-available-for-docker-orgs/) / [Docker docs](https://docs.docker.com/enterprise/security/oidc-connections/create-manage/)

## Scope

|Dockerfile|Image|
|---|---|
|`docker/backend.ai-manager.Dockerfile`|`lablup/backend.ai-manager`|
|`docker/backend.ai-agent.dockerfile` (new, BA-7271)|`lablup/backend.ai-agent`|
|`docker/backend.ai-appproxy-coordinator.dockerfile`|`lablup/backend.ai-appproxy-coordinator`|
|`docker/backend.ai-appproxy-worker.dockerfile`|`lablup/backend.ai-appproxy-worker`|
|`docker/backend.ai-webserver.dockerfile`|`lablup/backend.ai-webserver`|
|`docker/backend.ai-storage-proxy.dockerfile`|`lablup/backend.ai-storage-proxy`|
|`docker/backend.ai-client.dockerfile`|`lablup/backend.ai-client`|

Out of scope:

- Infra images (`krunner-extractor`, `socket-relay`, `linuxkit-nsenter`) — self-contained in backend.ai-krunner, not published here.
- Refactoring agent `docker load`/`docker exec` subprocess calls to aiodocker (backlog follow-up noted in BA-7271).
- Not related to the isla-sorna pipeline; that dispatch stays as is.

## Design decisions (confirmed)

- **Auth**: Docker Hub OIDC connection (lablup org is a Team account with the feature available). `docker/login-action@v4` (≥ 4.5.0), `username: lablup`, no password, `DOCKERHUB_OIDC_CONNECTIONID` env from a GitHub Actions repo variable, `permissions: id-token: write`.
- **Ruleset subject claim**: `repo:lablup/backend.ai:environment:deploy-to-dockerhub` — push jobs run inside a protected GitHub environment (mirrors the existing `deploy-to-pypi` pattern used for PyPI trusted publishing).
[...]

PR #13711, opened 2026-08-11
fix(BA-7321): mount only per-service configs into the non-elevated DOCKER-mode containers

kyujin-cho:
## Summary
- The DOCKER install mode mounted the whole install directory into **every** service container, so a compromised webserver or app-proxy container could read `manager.toml` (DB credentials) and every other secret in the tree. The parity mount is only genuinely needed where paths must resolve on the host or bootstrap state is shared.
- `docker-compose.services.yml`: the webserver, appproxy-coordinator, appproxy-worker, and appproxy-worker-tcp services now mount **only their own config file, read-only**; the base_path parity mount remains on manager, manager-cli, agent, and storage-proxy (agent/storage-proxy hand paths under it to the host Docker daemon; manager/manager-cli exchange the RPC keypair fixtures through it). Host networking and the `command:` override are unchanged, so the generated configs still work verbatim — their working-dir-relative state (`ipc/`, `var/`, logs) simply becomes container-local, which nothing external consumes (the file log driver is inactive; logs go to the console).
- `DockerContext._run_service_cli` gains `parity_mounted`: for the config-only coordinator service, one-off `docker compose run` arguments under the install directory (e.g. the copied `alembic-appproxy.ini` for `schema oneshot`) now get the same per-file read-only bind mounts as outside paths.
- Tests: the compose-rendering suite now asserts the parity/config-only split exactly — config-only services carry precisely one volume (their own config, `:ro`) matching the file their command reads.
- `docker/README.md` DOCKER-mode contract updated accordingly.

Deliberately NOT adopted from the hand-maintained `dockerized/` layout: bridge networking with published ports — that would require the config generators to emit container-DNS addressing and keep compose port maps in lockstep with the tomls, forking the PACKAGE-mode code path this mode exists to reuse.

**Needs a live re-verification** of the DOCKER install flow on a real Docker host before undrafting (webserver/app-proxy startup and the appproxy `schema oneshot` one-off are the affected paths) — part of the BA-7270 E2E scope.

**Merge order note:** stacked on the bake CI PR — last PR of the BA-7264 chain; retarget to `main` after the base PR merges.

Part of the BA-7264 epic (#13582).

### DIFF
--- a/changes/docker-minimal-service-mounts.fix.md
+++ b/changes/docker-minimal-service-mounts.fix.md
@@ -1,0 +1,1 @@
+Mount only each service's own config file (read-only) into the DOCKER install-mode webserver and app-proxy containers instead of the whole install directory, so they cannot read the credentials in the other services' configs
--- a/docker/README.md
+++ b/docker/README.md
@@ -181,9 +181,14 @@ the **DOCKER install mode** of `backend.ai-installer` generates at
 
 - Every service runs on the host network, so the generated configs use the
   same `127.0.0.1` addressing as a package-based install.
-- The install directory is bind-mounted into every container at the identical
-  absolute path, and each service's `command:` reads its config from there —
-  no `/etc/backend.ai` mounts.
+- Each service's `command:` reads its config from the install directory — no
+  `/etc/backend.ai` mounts. The install directory is bind-mounted at the
+  identical absolute path (path parity) only into the services whose files
+  must resolve on the host or that share bootstrap state: agent and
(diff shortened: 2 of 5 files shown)


kyujin-cho:
Closing per design discussion: identical host↔container path mounts are cheap (Docker auto-creates missing bind sources), so the uniform install-directory parity mount stays for simplicity. Effort redirected to keeping the reference compose and the installer-generated compose in contract parity.

PR #13717, opened 2026-08-12
fix(BA-7328): stop scheduling past a resource-exhausted session (#13707)

lablup-octodog:
This is an auto-generated backport of #13707 to the 26.4 release.

### DIFF
--- a/changes/13707.fix.md
+++ b/changes/13707.fix.md
@@ -1,0 +1,1 @@
+Stop scheduling sessions behind one whose resource group ran out of resources, so lower-priority sessions no longer take what a blocked higher-priority session is waiting for
--- a/src/ai/backend/manager/models/scheduling_history/row.py
+++ b/src/ai/backend/manager/models/scheduling_history/row.py
@@ -77,17 +77,28 @@ class SessionSchedulingHistoryRow(Base):  # type: ignore[misc]
         onupdate=sa.func.now(),
     )
 
+    def records_an_attempt(self) -> bool:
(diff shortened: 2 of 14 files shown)


HyeockJinKim:
Closing this auto-backport — sokovan diverged too far from main between 2026-08-12 (when this backport opened) and now.

The conflict markers left uncommitted by the auto-backport aren't a simple 3-way resolve here:
- `results.py` / `coordinator.py` / `sokovan/data/lifecycle.py`: mostly additive and resolvable, but only because unrelated preemption/reservation fields (`PreemptionPlanEntry`, `reserved_session_ids`, `preemption_plan`) that exist on main don't exist on 26.4 at all — that feature was never backported here.
- `provisioner.py` is the real blocker: main distinguishes "resource group ran out of resources" (which should block sessions queued behind it) from other per-session failures (architecture mismatch, quota, etc.) by catching two specific exception types (`BatchAgentSelectionFailedError`, `NoAvailableAgentError`). On 26.4, `BatchAgentSelectionFailedError` doesn't exist, and `NoAvailableAgentError` is a single exception that aggregates architecture mismatch, insufficient resources, and container-limit-exceeded together (`agent_errors: Mapping[AgentId, TrackerCompatibilityError]`). Faithfully reproducing the original PR's skip-vs-fail semantics on 26.4 would require writing new classification logic to inspect `agent_errors`' value types — a design decision the original PR never had to make, and not something that should be decided without the original author.

Recommend re-evaluating whether/how this fix should land on 26.4 directly with sokovan-team review, rather than via auto-backport.

PR #13682, opened 2026-08-11
feat(BA-7243): expose session idle-check exclusion and inclusion APIs

seedspirit:
## Summary
- Expose the session idle-check exclusion feature across the full API surface (BA-7243/BA-7244): shared v2 DTOs with identifier types (`IdleCheckerID`/`SessionID`), `POST /v2/sessions/exclude-idle-checks` / `/include-idle-checks`, the `excludeSessionIdleChecks` / `includeSessionIdleChecks` GQL mutations with wrapped inputs, `V2SessionClient` methods, and the `bai session exclude-idle-checks` / `include-idle-checks` CLI commands.
- Both operations take **(checker, session) pair targets** (`SessionIdleCheckTargetInput`) and are partial-success: the payloads carry `success` (the pairs actually processed) and `failed` (per-verb FailureInfo with the reason per pair), mirroring the rbac bulk assign/revoke pattern.
- **The PR boundary is the adapter**: `SessionAdapter.exclude_idle_checks` / `include_idle_checks` raise `NotImplementedError` here. The pair-upsert rework below the adapter lives in #13743 (BA-7348), targeting main independently; the stubs are wired to those processors once both land.

## Test plan
- [x] Logic-layer tests (actions/service/repository) live in #13743 alongside the implementation
- [ ] Live E2E on the dev stack with the #13743 layer applied and the stubs wired: mixed valid/unknown exclude and include pairs return partial `success`/`failed` payloads over REST/GQL/CLI, rows are upserted accordingly with the managing writer recorded, and audit rows carry per-entity success/error with `ActionRunStatus`-derived reasons

## Dependencies
- #13328 (BA-7120) landed the initial actions/service/repository chain on main; #13743 (BA-7348) reworks it to pair-based manual upserts. This PR's stubs are wired once both land.
- Non-superadmin callers currently receive 403 on a stock dev stack: sessions are not yet materialized into the v2 virtual-scope tables (`entity_memberships` has no session rows — the scheduler writes only the legacy `association_scopes_entities`). The role/permission seeds are already correct: materializing the missing rows by hand produced the intended matrix (owner and project-admin allowed; project member and monitor denied). Session virtual-scope materialization needs a separate ticket.

Resolves BA-7243 (also covers BA-7244)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

----
📚 Documentation preview 📚: https://sorna--13682.org.readthedocs.build/en/13682/

----
[...]

PR #13718, opened 2026-08-12
feat(BA-7339): record the managing writer on session idle-check rows

seedspirit:
## Summary
- Add manual-write provenance to `session_idle_checks`: `is_manual` (bool) and `manually_triggered_by` (FK to `users.uuid`, `ON DELETE SET NULL` — the flag outlives user deletion). Existing rows and assignment-sync writes stay non-manual.
- Assignment sync's cleanup now skips manual rows, so manually written exclusions/inclusions are not silently removed when scope assignments change.
- Nothing writes manual rows yet — the exclusion/inclusion APIs that stamp these columns land in BA-7243 (#13682); this PR carries only the schema and the sync guard.
- A check constraint keeps the pair consistent: `manually_triggered_by IS NULL OR is_manual`.

## Test plan
- [x] Repository unit test: sync deletion spares a manual row while removing the non-manual one; assignment-sync-created rows assert `is_manual == False`
- [x] `alembic upgrade head` and a downgrade→upgrade round-trip verified on the local dev stack

Resolves BA-7339

### DIFF
--- a/src/ai/backend/manager/repositories/idle_checker/purgers.py
+++ b/src/ai/backend/manager/repositories/idle_checker/purgers.py
@@ -68,7 +68,7 @@ class IdleCheckerPurgerSpec(PurgerSpec[IdleCheckerRow]):
 
 
 @dataclass
-class SessionIdleCheckBatchPurgerSpec(BatchPurgerSpec[SessionIdleCheckRow]):
+class SessionIdleCheckSyncPurgerSpec(BatchPurgerSpec[SessionIdleCheckRow]):
     pairs: Sequence[SessionIdleCheckPair]
 
     @override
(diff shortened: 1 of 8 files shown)


fregataa (dismissed):

src/ai/backend/manager/repositories/idle_checker/purgers.py
```diff
@@ -80,6 +80,7 @@ def build_subquery(self) -> sa.sql.Select[tuple[SessionIdleCheckRow]]:
                 SessionIdleCheckRow.idle_checker_id,
             ).in_(pair_values),
             SessionIdleCheckRow.last_status != IdleCheckPhase.IDLE_EXPIRED,
+            sa.not_(SessionIdleCheckRow.is_manual),
```
fregataa, line 83: Why does this purger skip deleting user created spec?

fregataa (approved):

PR #13726, opened 2026-08-12
feat(BA-7343): backfill missing idle_checkers.target_session_types

seedspirit:
## Summary

- 26.7.0 released migration `d3f8a1c45e9b` without `target_session_types`; #12398 later added the column by editing that migration in place, so databases migrated in that window never received it and the idle-check queries fail with `UndefinedColumnError`.
- Adds a single idempotent repair migration (`f2d658cac56b`) at the chain head: `ADD COLUMN IF NOT EXISTS ... NOT NULL DEFAULT '{}'` followed by `DROP DEFAULT`, so it is a no-op on databases that already have the column.
- Pre-existing rows are backfilled with an empty array so checkers created before the column existed stay inert until an admin assigns target types.

26.8 does not take a schema repair; it unregisters the idle-check reconciler stages instead (#13734, BA-7344). A 26.7.x database upgrading through 26.8 to a main-based release is still covered — this migration sits on the main chain it will pass through.

## Test plan

- [x] Repair path: cloned schema without the column + a legacy row at `c8d51e7a3b62` → `upgrade head` restores the column, backfills `{}`, NOT NULL without default (matches the model)
- [x] No-op path: re-applying on a column-present DB passes without touching data
- [x] `downgrade -1` → `upgrade head` round trip
- [x] Single alembic head (`f2d658cac56b`)

Resolves BA-7343


jopemachine:
CI is failing

HyeockJinKim (approved):

PR #13727, opened 2026-08-12
feat(BA-7343): backfill missing idle_checkers.target_session_types (26.8)

seedspirit:
## Summary

Manual backport of #13726 to 26.8.

- 26.7.0 released migration `d3f8a1c45e9b` without `target_session_types`; #12398 later added the column by editing that migration in place, so databases migrated on 26.7.x never received it and fail the `idle_check_judgment` lifecycle on 26.8.x with `UndefinedColumnError`.
- Adds the idempotent repair migration `f2d658cac56b` on top of the 26.8 head (`b93d1c47af52`) — the same revision id as the copy inserted into the main chain by #13726, so 26.8 databases can later upgrade to main-based releases without a missing-revision error.
- Pre-existing rows are backfilled with an empty array so checkers created before the column existed stay inert until an admin assigns target types.

## Test plan

- [x] Repair path verified on a cloned schema without the column: column restored, legacy row backfilled with `{}`, NOT NULL without default (matches the model)
- [x] `downgrade -1` → `upgrade head` round trip
- [x] 26.8 chain resolves to the single head `f2d658cac56b`

Resolves BA-7343

### DIFF
--- a/changes/13727.fix.md
+++ b/changes/13727.fix.md
@@ -1,0 +1,1 @@
+Fix `UndefinedColumnError` on `idle_checkers.target_session_types` for databases migrated on 26.7.x by re-adding the column idempotently with an empty-array backfill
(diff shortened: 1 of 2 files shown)


seedspirit:
Superseded: instead of backporting the schema repair, 26.8 will unregister the idle-check reconciler stages entirely since the feature is not actually enabled there (BA-7344 follow-up PR).

Issue #11716, opened 2026-05-21
[Domain] Move users.domain_name FK to users.domain_id


fregataa:
Goal

Move the foreign key reference from ``users.domain_name`` (string, → ``domains.name``) to a new ``users.domain_id`` (UUID, → ``domains.id``). The legacy ``domain_name`` string column is retained without a FK constraint; its removal is deferred to BA-6122. ``domains`` itself keeps ``name`` as PK in this story — the PK swap on ``domains`` is BA-6046.

Scope

1. Alembic: single revision, parent ``down_revision = fdc9d6ac49b4``. Add ``users.domain_id`` UUID column (NULLABLE — ``users.domain_name`` is nullable). Backfill ``domain_id = domains.id`` via join on ``domains.name``. Drop the existing ``users.domain_name → domains.name`` FK. Create new ``users.domain_id → domains.id`` FK. Provide downgrade.

2. ORM: add ``domain_id: Mapped[DomainID \| None]`` on ``UserRow``; keep ``domain_name`` column (drop only its FK clause); update ``_get_domain_join_condition`` if needed so the ``users → DomainRow`` relationship joins via ``domain_id``.

3. Create / update specs: add ``domain_id: DomainID`` field to ``UserCreatorSpec``; add ``domain_id: OptionalState[DomainID]`` to ``UserUpdaterSpec``. Make ``build_row`` / ``build_values`` populate both legacy and new columns.

4. API entry points that build the spec from client input: convert ``domain_name`` → ``domain_id`` via ``ResolveDomainIDByNameAction`` (already exists). Affected: ``api/rest/user/handler.py``, ``api/adapters/user/adapter.py``, ``api/gql/user/resolver/mutation.py``, ``api/gql_legacy/user.py``.

5. Tests: update fixtures that construct ``UserRow`` directly so ``domain_id`` is populated alongside ``domain_name``.

Parallelism

This story does NOT depend on the other 9 referencing-table stories; they may all proceed in parallel. The 10 stories will each emit an Alembic revision pointing at the same ``down_revision``, producing multiple heads. The heads are consolidated by an ``alembic merge`` revision under BA-6046 before the ``domains`` PK swap lands.

Out of scope

- ``domains`` PK swap — BA-6046.

- Legacy ``domain_name`` column drop and search-filter / propagation cleanup — BA-6122.

- RBAC scope migration — BA-6047.

JIRA Issue: BA-6123

PR #13729, opened 2026-08-12
refactor(BA-7346): make the users domain columns NOT NULL

jopemachine:
## 📚 Stacked PRs

**Merge in order (bottom-up):**

1. ~~#13713 — `refactor(BA-7329)`: make the always-populated users columns NOT NULL~~ (merged)
2. 👉 **#13729 — `refactor(BA-7346)`: make the users domain columns NOT NULL** ← you are here

Resolves BA-7346

## Summary

`users.domain_id` and `users.domain_name` are written on every insert — both creation paths resolve the domain by name first and raise `UserCreationBadRequest` when it does not exist — and no update path can write NULL. The two columns follow the four in #13713.

They are here rather than there because they are the ones the test fixtures do not already fill: `DomainRow.id` is server-generated, so a fixture that seeds a domain has no id to hand the user it seeds next.

| Change | Detail |
|---|---|
| `models/user/row.py` | `domain_name: Mapped[str]` and `domain_id: Mapped[DomainID]`, both `nullable=False` |
| migration `e7b2c9f04d31` | backfills each half from the other, then places whatever is still missing both in the default domain, then two `alter_column`s |
| `data/user/types.py` | `UserData.domain_name` / `.domain_id` follow the columns |
| `api/rest/middleware/auth.py`, `models/rbac/__init__.py`, `repositories/vfolder/repository.py` | drop the NULL guards the columns now make impossible |
| 116 test files | the fixture that seeds the domain hands back its id with its name |

### What the fixtures look like now

The fixture generates the id where it generates every other seed value, and returns both halves through the existing `DomainFixtureData`:

```python
@pytest.fixture
async def test_domain(self, db_with_cleanup: ExtendedAsyncSAEngine) -> DomainFixtureData:
    domain_id = DomainID(uuid.uuid4())
    domain_name = f"test-domain-{uuid.uuid4().hex[:8]}"
    ...
    return DomainFixtureData(domain_name=DomainName(domain_name), domain_id=domain_id)
```

Consumers read `test_domain.domain_name` and `test_domain.domain_id`. Fixtures that seed both rows in one function keep the id local, the way they already keep `user_uuid` and `group_id`. Only the fixtures that were named `test_domain_name` are renamed; the rest keep their name and just carry the id too.

## Test plan

- [ ] `check-alembic-migrations` — the migration is chained onto main's head `c4d1a7e58b32`, single head

## Note for the reviewer

`domain_id` is written on create and never on update: `UserUpdaterSpec` has no such field, and the update path only checks that the new `domain_name` exists. A user who changes domain keeps the old id. NOT NULL guarantees a value, not the right value — the drift belongs to the contract-phase work on #11716.


tests/unit/manager/repositories/scaling_group/test_scaling_group_repository.py
```diff
@@ -339,6 +345,7 @@ async def scaling_group_for_update(
     async def test_user_domain_group(
         self,
         db_with_cleanup: ExtendedAsyncSAEngine,
+        domain_id: DomainID,
```
jopemachine: Don't inject this

tests/unit/manager/repositories/scheduler/test_preemption_candidates.py
```diff
@@ -41,10 +41,16 @@
 from ai.backend.manager.models.utils import ExtendedAsyncSAEngine
 from ai.backend.manager.repositories.scheduler.db_source.db_source import ScheduleDBSource
 from ai.backend.manager.views.sokovan.workload import PreemptionScopeKey
+from ai.backend.testutils.fixtures import DomainFixtureData
 
 from .conftest import create_pending_session_with_kernels
 
 
+@pytest.fixture
+def domain_id() -> DomainID:
```
jopemachine: Don't inject this

tests/unit/manager/repositories/scheduler/test_scheduling_history_recording.py
jopemachine, line 117: Why id is not used?

tests/unit/manager/repositories/session/test_session_repository.py
```diff
@@ -69,6 +69,11 @@ class SessionTestData:
     access_key: AccessKey
 
 
+@pytest.fixture
+def domain_id() -> DomainID:
+    return DomainID(uuid.uuid4())
```
jopemachine: Use existing test_domain_id

tests/unit/manager/test_idle_check_host.py
```diff
@@ -112,6 +112,11 @@
 ]
 
 
+@pytest.fixture
+def domain_id() -> DomainID:
```
jopemachine: Don't make this

fregataa (dismissed):

src/ai/backend/manager/api/rest/middleware/auth.py
```diff
@@ -782,8 +782,6 @@ async def _query() -> UserRow | None:
     row = await execute_with_retry(_query)
     if row is None:
         raise UserNotFound("Impersonation target user not found")
-    if row.role is None or row.domain_name is None or row.domain_id is None:
-        raise InternalServerError(f"Impersonation target user is misconfigured (user_id={user_id})")
```
HyeockJinKim, line 786: 👍

HyeockJinKim (approved):

PR #13734, opened 2026-08-12
feat(BA-7344): unregister idle-check reconciler stages on 26.8

seedspirit:
## Summary

- The first-class idle checker is not enabled in 26.8, and databases migrated on 26.7.x lack `idle_checkers.target_session_types` (released migration `d3f8a1c45e9b` was amended in place by #12398), so the `idle_check_judgment` lifecycle fails with `UndefinedColumnError`.
- Drops the four idle-check stage registrations (assignment sync, initial grace period, judgment, sweep) from the reconciler coordinator factory, so neither the stages nor their leader event tasks run on 26.8. Stale queued reconcile events are ignored safely (`stages.get()`).
- The legacy `IdleCheckerHost` (etcd-config) path is unaffected.
- Replaces #13727: no schema repair on 26.8 — main keeps the column via #13726.

## Test plan

- [x] `tests/unit/manager/sokovan/idle_check/` — 52 tests pass against this branch; the factory registration test now guards that the stages stay **unregistered**

Resolves BA-7344

### DIFF
--- a/tests/unit/manager/sokovan/idle_check/test_factory.py
+++ b/tests/unit/manager/sokovan/idle_check/test_factory.py
@@ -1,31 +1,23 @@
 from __future__ import annotations
 
-from unittest.mock import MagicMock, patch
+from unittest.mock import MagicMock
 
 from ai.backend.manager.sokovan.stages.factory import build_reconciler_coordinator
 
 
 class TestFactoryRegistration:
-    def test_idle_check_stages_are_registered(self) -> None:
-        metric_repository = MagicMock()
-        with patch(
-            "ai.backend.manager.sokovan.stages.idle_check_judgment.UtilizationChecker"
-        ) as utilization_checker:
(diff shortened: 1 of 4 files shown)


tests/unit/manager/sokovan/idle_check/test_factory.py
```diff
@@ -1,31 +1,23 @@
...
-            _, task_specs = build_reconciler_coordinator(
-                replica_group_repository=MagicMock(),
-                idle_checker_repository=MagicMock(),
-                metric_repository=metric_repository,
-                scheduling_controller=MagicMock(),
-                valkey_live=MagicMock(),
-                valkey_schedule=MagicMock(),
-                lock_factory=MagicMock(),
-                config_provider=MagicMock(),
-            )
```
fregataa, line 23: Seems you can set `xfail` annotation here

fregataa (approved):

PR #13735, opened 2026-08-12
feat: make the agent heartbeat event serializable as JSON

fregataa:
## Summary

`AgentHeartbeatEvent.to_message()` raised and its JSON schema could not be built, because `AgentInfo` carried three values Pydantic cannot render as JSON:

| Field | Why it could not be rendered | Resolution |
|---|---|---|
| `available_resource_slots` | `ResourceSlot` is a `UserDict`, with no Pydantic serializer | becomes `list[ResourceSlotEntry]` |
| `slot_key_and_units` | keyed by `SlotName`, a `UserString`, likewise | keyed by `ResourceSlotName` |
| `images` | zlib-compressed msgpack; never valid UTF-8, so it cannot ride in JSON as the text Pydantic decodes `bytes` into | removed — nothing read it |

The first two move the wire type onto Pydantic-native forms rather than teaching the legacy types to serialize. Conversion happens at the two boundaries: the agent projects its `ResourceSlot` when building the heartbeat, and `AgentHeartbeatUpsert.from_agent_info` collapses the entry list back, since the `agents` table still stores a `ResourceSlot`. Both are declared `when_used="json"`, so the Python-mode dump that the msgpack path packs is unchanged and the event still round-trips through `serialize()` / `deserialize()` exactly as before.

`images` turned out to be dead weight rather than something to fix. It was rebuilt and compressed on every heartbeat, and no consumer read it: `handle_heartbeat` touches `addr`, `public_key`, `scaling_group` and the resource slots, and `AgentHeartbeatUpsert.from_agent_info` uses neither it nor `images_opts`. Installed images reach the manager by another route — the agent pushes them to Valkey immediately after emitting the heartbeat, and the manager reads them back via `AgentRepository.sync_installed_images`. Both fields are dropped.

An earlier commit on this branch added `__get_pydantic_core_schema__` to `ResourceSlot` and `SlotName` to reach the same goal; the field-type change makes those hooks callerless, so a later commit reverts them. What remains of that work is hoisting the `pydantic_core` import to the top of `common/types.py`, which `QuotaScopeID` and `VFolderID` now use.

### Compatibility

The heartbeat wire form changes, so a manager must not receive heartbeats from an agent running an older release. Roll the manager out ahead of the agents. In the other direction an older agent that still sends `images` and `images_opts` is harmless, since the event body ignores unknown fields.

## Test plan

- [x] `tests/unit/common/events/test_agent_heartbeat_event.py` — slot keys round-trip whether given as `str`, `SlotName`, or a mix
- [x] Manual verification: the msgpack path is byte-identical to before, the JSON path round-trips with values and types preserved, `model_json_schema()` builds, and a payload still carrying `images` validates to the same event
- [x] `pants check` / `pants lint` on the touched files; affected unit tests pass
- [ ] CI full type check and test suite


seedspirit (dismissed):

src/ai/backend/common/data/agent/types.py
```diff
@@ -1,11 +1,39 @@
...
+# leaving the msgpack path to pack the raw bytes as before.
+PackedImages = Annotated[
+    bytes,
+    BeforeValidator(_decode_images),
+    PlainSerializer(
+        lambda value: base64.b64encode(value).decode("ascii"),
+        return_type=str,
+        when_used="json",
+    ),
+]
```
HyeockJinKim: What kind of logic is this?

fregataa: Removed the `AgentInfo.images` field

HyeockJinKim (approved):

PR #13736, opened 2026-08-12
feat(BA-7248): add --dst-dir option to vfolder upload

fregataa:
## Summary

- Add `-d/--dst-dir` to `backend.ai vfolder upload`, exposing the SDK's existing `dst_dir` argument so files land in a subdirectory of the virtual folder instead of always at its root.
- Stringify `dst_dir` before putting it into the storage proxy query string. Passing a `Path` raised `TypeError: Invalid variable type: value should be str, int or float, got PosixPath(...)` from yarl, which broke `backend.ai model upload` / `model download` — the only existing callers of `dst_dir`.

No server-side change: the storage proxy already handles the `dst_dir` query parameter for both upload and download.

## Test plan

- [x] `backend.ai vfolder upload <folder> -d data/inputs sample.csv` places the file at `data/inputs/sample.csv`
- [x] `backend.ai vfolder upload <folder> sample.csv` (no `--dst-dir`) still uploads to the vfolder root

## Known gaps (out of scope, filed for follow-up)

- `--recursive` combined with `--dst-dir` leaves empty directories at the vfolder root: `_upload_recursively` calls `_mkdir()` without the `dst_dir` prefix. Files themselves land correctly because the storage proxy creates parent directories.
- The storage proxy's `tus_upload_part` has no `relative_to(vfpath)` traversal check, unlike `download`.

Resolves BA-7248

### DIFF
--- a/changes/13736.fix.md
+++ b/changes/13736.fix.md
@@ -1,0 +1,1 @@
+Fix a `TypeError` raised when `dst_dir` is given as a `Path`, which broke `backend.ai model upload` and `model download`.
--- a/changes/13736.fix.md
+++ b/changes/13736.fix.md
@@ -1,1 +1,0 @@
-Fix a `TypeError` raised when `dst_dir` is given as a `Path`, which broke `backend.ai model upload` and `model download`.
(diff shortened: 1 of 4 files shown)


changes/13736.fix.md
jopemachine, line 1: Is this file correct?

jopemachine (approved):

PR #13739, opened 2026-08-12
feat(BA-7345): limit the size of an app config fragment

### LINKED ISSUE #13730: Bound the size of an app config fragment
## Objective

Put an upper bound on how large a single app config fragment may be. Today no layer bounds it, so any principal allowed to write a fragment can store a document of practically unlimited size. Decide the limit, enforce it on every write path, and reject oversize documents with a typed error.

## Background

- `AppConfigFragmentRow.config` is a plain `JSONB` column with no check constraint (`manager/models/app_config_fragment/row.py`). Postgres stops only at the ~1 GB field ceiling.
- `AppConfigFragmentUpsertItem.config` is typed `dict[str, Any]` with no `Field()` constraint (`common/dto/manager/v2/app_config_fragment/request.py`), and `AppConfigFragmentService.bulk_upsert` adds no size check of its own.
- The bulk upsert inputs (`ScopedUpsertAppConfigFragmentsInput.items`, `MyUpsertAppConfigFragmentsInput.items`) declare `min_length=1` but no maximum, so one request may also carry an unbounded number of items.
- The `my` upsert lets an ordinary user write fragments at their own user scope, so the growth path is reachable without any admin role.
- Every app config read merges the fragments for the requested names (`AppConfigService.get_app_configs`), so an oversize fragment is paid for on each read, not just once at write time.

## Scope

- Decide the limit and how it is measured (serialized UTF-8 JSON bytes is the obvious unit).
- Enforce it in the service layer so GraphQL and REST v2 are both covered, and raise a `BackendAIError` subclass added to `manager/errors/app_config.py` rather than a bare validation error.
- Keep the per-item partial-success contract from BA-7250: in a bulk upsert an oversize item fails on its own and does not sink the rest of the batch.
- Decide the maximum item count per bulk upsert request, or split it off explicitly if it is not taken here.
- State the limit in the DTO field description so it shows up in the generated API reference.

## Open questions

- The value itself is undecided — the candidates are on the order of 64 KiB / 256 KiB / 1 MiB per fragment.
- Fixed constant, or configurable through the manager config? A per-config-name limit carried on the definition or the allow-list entry is a third option.
- Whether the merged result of all fragments for one config name needs its own cap on top of the per-fragment one.

## Acceptance Criteria

- A per-fragment size limit is chosen, documented, and enforced on both the scoped and the `my` upsert paths.
- An oversize fragment is rejected with a typed app config error carrying a distinguishable error code, not a generic 400.
- In a bulk upsert, an oversize item rejects the whole batch. Per-item partial reporting waits until the repository supports partial success — the upsert's failed list is never populated today.
- A decision on the per-request item count limit is recorded — implemented, or deferred to a linked issue.
- Tests cover a document just under the limit (accepted), one just over (rejected), and the mixed bulk batch.

## Related

- BA-7230 — App Config follow-up issues (parent epic).
- BA-7250 — Support partial success in bulk app config fragment upsert.

JIRA Issue: BA-7345

jopemachine:
Resolves #13730 (BA-7345)

## Summary
- Cap one app config fragment at 64 KiB of compact JSON (`MAX_APP_CONFIG_FRAGMENT_BYTES`), checked in `AppConfigFragmentService.bulk_upsert` before the batch reaches the repository, so the GraphQL and the REST v2 write paths are both covered. Nothing bounded the document until now — the DTO accepted any dict and the JSONB column stopped only at Postgres' field ceiling.
- Reject an oversize document with a new `AppConfigFragmentTooLarge` error under a new `ErrorDomain.APP_CONFIG_FRAGMENT`. One oversize item rejects the whole batch, as the allow-list write gate already does.
- State the limit in the upsert input's field description, on the v2 DTO and on the GraphQL input.

The value is arbitrary for this PR. BA-7345 keeps the open questions: the number itself, whether it belongs in the manager config or per config name on the definition / allow-list entry, whether the merged result of all fragments for one config name needs its own cap, and whether the per-request item count should be bounded too.

## Test plan
- [x] Unit tests for the service: a document exactly at the limit is written, and a batch holding one oversize item is rejected without reaching the repository
- [ ] CI green — the suite was not exercised locally

### DIFF
--- a/src/ai/backend/common/data/app_config/types.py
+++ b/src/ai/backend/common/data/app_config/types.py
@@ -3,11 +3,18 @@
 from __future__ import annotations
 
 import enum
+from typing import Final
 
 from ai.backend.common.data.permission.types import RBACElementType, ScopeType
 from ai.backend.common.identifier.app_config import AppConfigScopeID
 
-__all__ = ("AppConfigScopeType",)
+__all__ = (
+    "MAX_APP_CONFIG_FRAGMENT_BYTES",
+    "AppConfigScopeType",
+)
+
+#: Upper bound on one fragment's config document, measured on its compact JSON encoding.
+MAX_APP_CONFIG_FRAGMENT_BYTES: Final[int] = 64 * 1024
 
 
 class AppConfigScopeType(enum.StrEnum):
--- a/changes/13739.feature.md
+++ b/changes/13739.feature.md
@@ -1,0 +1,1 @@
+Limit one app config fragment to 64 KiB of JSON; a larger document is now rejected at upsert time instead of being stored unbounded.
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -1254,7 +1254,9 @@ input AppConfigFragmentUpsertItem
   """Registered config name."""
   configName: String!
 
-  """The fragment's JSON config document."""
+  """
+  The fragment's JSON config document, at most 65536 bytes once serialized.
+  """
   config: JSON!
 }
 
--- a/docs/manager/graphql-reference/v2-schema.graphql
+++ b/docs/manager/graphql-reference/v2-schema.graphql
@@ -928,7 +928,9 @@ input AppConfigFragmentUpsertItem {
   """Registered config name."""
   configName: String!
 
-  """The fragment's JSON config document."""
+  """
+  The fragment's JSON config document, at most 65536 bytes once serialized.
+  """
   config: JSON!
 }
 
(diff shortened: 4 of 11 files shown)


src/ai/backend/common/data/app_config/types.py
```diff
@@ -3,11 +3,18 @@
...
 from ai.backend.common.identifier.app_config import AppConfigScopeID
 
-__all__ = ("AppConfigScopeType",)
+__all__ = (
+    "MAX_APP_CONFIG_FRAGMENT_BYTES",
+    "AppConfigScopeType",
+)
+
+#: Upper bound on one fragment's config document, measured on its compact JSON encoding.
+MAX_APP_CONFIG_FRAGMENT_BYTES: Final[int] = 64 * 1024
```
jopemachine, line 17: Temp value FYI

PR #13743, opened 2026-08-12
feat(BA-7348): apply session idle-check exclusion and inclusion as manual pair upserts

seedspirit:
## Summary
- Rework the session idle-check exclusion/inclusion chain below the adapter (landed by BA-7120) to operate on (checker, session) pair targets applied as manual upserts.
- The service owns upserter assembly: it deduplicates the pair targets and builds the `SessionIdleCheckExclude/IncludeUpserterSpec`s (`is_manual=True`, `manually_triggered_by=<requesting user>`, BA-7339); the repository receives only the assembled `BulkUpserter` and executes it with savepoint-isolated partial success.
- Inclusion becomes an upsert mirroring exclusion: any phase resets to `NOT_CHECKED` (grace period restarts) and missing pair rows are created.
- The checker pre-check is dropped — the foreign keys map an unknown session/checker to `SessionNotFound` / `IdleCheckerNotFound` for that pair only, instead of failing the whole batch.
- **The PR boundary is the adapter**: no API-layer changes here. The BA-7243 adapter stubs (#13682) keep raising `NotImplementedError` and are wired to these processors once both PRs land.

## Behavior changes
| Before | After |
|--------|-------|
| Unknown checker fails the whole batch | Fails only the pairs referencing it |
| Include updates only `EXCLUDED` rows | Include resets any phase and creates missing rows |
| Rows written without provenance | Both verbs set `is_manual` + `manually_triggered_by`, so assignment sync spares them |

## Test plan
- [x] Repository tests (real DB): per-pair upsert semantics, partial failure mapping (`SessionNotFound` / `IdleCheckerNotFound`), writer stamping, assignment sync sparing manual rows
- [x] Service tests (mocked repository): pair dedup and upserter spec assembly
- [ ] Live E2E on the dev stack once wired through #13682: mixed valid/unknown pairs over REST/GQL/CLI return partial `success`/`failed` payloads and per-pair audit records

## Dependencies
- Independent of #13682 (BA-7243) at the code level; the two meet when the adapter stubs are wired after both land.

Resolves BA-7348

### DIFF
--- a/src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
+++ b/src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
@@ -18,7 +18,6 @@ from ai.backend.common.data.permission.types import RBACElementType, ScopeType
 from ai.backend.common.identifier.domain import DomainID
 from ai.backend.common.identifier.idle_checker import IdleCheckerID
 from ai.backend.common.identifier.resource_group import ResourceGroupID
-from ai.backend.common.identifier.session import SessionID
 from ai.backend.common.types import SessionId, SessionTypes
 from ai.backend.manager.data.common.types import SearchResult
 from ai.backend.manager.data.idle_checker.types import (
@@ -33,7 +32,6 @@ from ai.backend.manager.errors.idle_checker import (
     IdleCheckerAssignmentScopeNotFound,
     IdleCheckerNotFound,
 )
-from ai.backend.manager.errors.kernel import SessionNotFound
 from ai.backend.manager.models.domain.conditions import DomainConditions
 from ai.backend.manager.models.domain.row import DomainRow
 from ai.backend.manager.models.group.row import GroupRow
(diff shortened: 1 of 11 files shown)


src/ai/backend/manager/repositories/idle_checker/upserters.py
```diff
@@ -59,4 +68,64 @@ def build_update_values(self) -> dict[str, Any]:
             "last_status": IdleCheckPhase.EXCLUDED,
             "expire_at": None,
             "last_message": "Excluded from idle checks.",
+            "is_manual": True,
+            "manually_triggered_by": self.user_id,
+            "updated_at": datetime.now(UTC),
```
fregataa: this updated_at is not needed

src/ai/backend/manager/repositories/idle_checker/db_source/db_source.py
```diff
@@ -458,86 +456,69 @@ async def batch_update_session_idle_check_phase(
...
-            if checker is None:
-                raise IdleCheckerNotFound(str(checker_id))
-            querier = BatchQuerier(
-                pagination=NoPagination(),
-                conditions=[
-                    SessionConditions.by_ids([SessionId(session_id) for session_id in session_ids])
-                ],
+            result = await w.bulk_upsert_partial(
+                upserter,
+                index_elements=["session_id", "idle_checker_id"],
```
fregataa, line 503: Is this index_elements needed?

seedspirit: Yes, it is required. `bulk_upsert_partial` takes `index_elements` as a mandatory parameter and forwards it to `pg_insert(...).on_conflict_do_update(index_elements=...)`, and PostgreSQL's `ON CONFLICT DO UPDATE` cannot omit the conflict target (only `DO NOTHING` can). `(session_id, idle_checker_id)` is the composite PK of `session_idle_checks`, which is the conflict target here.

HyeockJinKim (approved):

HyeockJinKim:
The approach to applying the UPSERT is a little different, but I'll merge it for now and make the necessary adjustments later.

PR #13746, opened 2026-08-12
refactor(BA-7315): carry event payloads as JSON instead of msgpack

fregataa:
## Summary

- Publish every event body as JSON, derived from the event's own Pydantic fields via the `@final` `to_message()` / `from_message()` pair, and remove the 130 hand-written `serialize()` / `deserialize()` implementations plus the msgpack wire format beneath them. A positional tuple can no longer disagree with the field order it is built from.
- Replace `creation_info: Mapping[str, Any]` with a typed `KernelCreationInfo` whose leaves are JSON-representable. This was the one payload relying on pickled `ResourceSlot` / `Decimal` / `Path` values carried in msgpack ext types, and it is now declared only on the started events — the only ones that ever carried it. The manager's untyped duplicate view of the same data is gone, and its allocation aggregation is now `KernelResourceSpecData.to_resource_slot()`.
- Raise `EventPayloadEncodingError` / `EventPayloadDecodingError` from the conversion pair instead of letting Pydantic errors escape, and ack a message whose body cannot be decoded rather than leaving it to be redelivered until the retry limit discards it. Also drops `args_matcher`, which matched against the positional tuple and had no callers.

Every component reads and writes the same single stream (`events`) and channel (`events_all`), so this requires the cluster to run one version across the upgrade. In-flight msgpack entries are logged and skipped as malformed.

This folds BA-7316 (removing the msgpack fallback) into this change: with the write path on JSON only, there is no fallback left to keep.

## Test plan

- [x] `pants test tests/unit/common/events:: tests/unit/common/message_queue::`
- [x] `pants test tests/unit/common/test_events.py tests/unit/common/test_distributed.py`
- [x] `pants test tests/unit/manager/repositories/scheduler::`
- [x] Verify event flows end to end on a live server: session create/terminate (kernel lifecycle), agent heartbeat, and image scan

Resolves BA-7315

### DIFF
--- a/src/ai/backend/manager/views/sokovan/lifecycle.py
+++ b/src/ai/backend/manager/views/sokovan/lifecycle.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field
 from datetime import datetime
-from decimal import Decimal
 from typing import Any
 from uuid import UUID
 
@@ -12,7 +11,6 @@ from ai.backend.common.identifier.architecture import ArchName
 from ai.backend.common.types import (
     AccessKey,
     AgentId,
-    BinarySize,
     ClusterMode,
     KernelId,
     ResourceSlot,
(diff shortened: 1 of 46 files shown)


src/ai/backend/manager/views/sokovan/lifecycle.py
```diff
@@ -179,111 +177,6 @@ class PreparedSessionsWithImages:
...
-    container_id: str | None = None
-    resource_spec: dict[str, Any] | None = None
-    attached_devices: dict[str, Any] = field(default_factory=dict)
-    repl_in_port: int | None = None
-    repl_out_port: int | None = None
-    stdin_port: int | None = None
-    stdout_port: int | None = None
-    service_ports: list[int] = field(default_factory=list)
-    kernel_host: str | None = None
-
```
fregataa, line 195: Replaced by `ai.backend.common.events.event_types.kernel.types.KernelCreationInfo`

HyeockJinKim (approved):

PR #13747, opened 2026-08-13
refactor(BA-7351): remove the HiRedisQueue message queue implementation

fregataa:
## Summary

- Delete `HiRedisQueue` and `common/redis_client.py` (a hand-rolled RESP client on raw `hiredis` whose only importer was `HiRedisQueue`), leaving `RedisQueue` as the only message queue transport. The flag has been off by default since 25.12.0, and a second transport meant every message queue change had to be written and verified twice.
- Use `RedisQueue.create()` unconditionally in the manager, agent, storage-proxy, app-proxy-coordinator and app-proxy-worker, and drop `use-experimental-redis-event-dispatcher` from the five unified configs, the manager legacy trafaret and the sample configs.
- **Breaking config change**: a config still setting the option to `true` now fails validation with an explicit message, rather than silently switching transport. `false` stays accepted and ignored. The wire format is unchanged, so no message is lost across the switch.
- Drop the direct `hiredis` requirement (regenerated `python.lock`); `redis[hiredis]` still pulls it in as a parser extra.

## Test plan

- [x] `pants fmt` / `pants fix` / `pants lint --changed-since=origin/main` pass
- [x] All five config classes verified: `true` rejected for both the kebab and snake alias, `false` ignored, absent loads normally
- [x] Regression tests added for the manager and agent config paths; the `HiRedisQueue` branch test removed
- [x] CI type check and tests
- [x] Live server: every component starts and event flows work end to end

Resolves BA-7351


src/ai/backend/manager/config/unified.py
```diff
@@ -1294,6 +1279,12 @@ def _parse_rpc_auth_manager_keypair(cls, v: str) -> str:
             )
         return v
 
+    @model_validator(mode="before")
+    @classmethod
+    def _reject_experimental_redis_event_dispatcher(cls, values: Any) -> Any:
```
jopemachine, line 1284: Do we really need to keep implementations like reject_experimental_redis_event_dispatcher around?

It doesn't seem like a good idea to maintain something like this for each config we've removed.

src/ai/backend/common/config.py
```diff
@@ -39,6 +40,27 @@
...
+    for alias in (
+        "use-experimental-redis-event-dispatcher",
+        "use_experimental_redis_event_dispatcher",
+    ):
+        if values.get(alias):
+            raise ValueError(
+                f"The '{alias}' option has been removed. "
+                "RedisQueue is now the only message queue implementation. "
+                "Delete the option from the configuration to keep using RedisQueue."
+            )
```
HyeockJinKim, line 61: Instead of checking a specific field, is there a way to apply this more generally?

HyeockJinKim (approved):

PR #13751, opened 2026-08-13
refactor(BA-7352): adopt `LifecycleTimestampsMixin` on domains

### LINKED ISSUE #13748: [Schema] Adopt LifecycleTimestampsMixin on domains and retire the hand-rolled modified_at column
## Objective

Give domains the audit-timestamp columns every table written since the mixins landed already has: created_at and updated_at from LifecycleTimestampsMixin.

## Background

models/mixins/timestamp.py provides CreatedAtMixin, UpdatedAtMixin and LifecycleTimestampsMixin, and users, keypairs, groups and vfolder_invitations have already moved onto them (BA-7207). domains was left behind: DomainRow inherits only CreatedAtMixin and hand-rolls modified_at with onupdate=current_timestamp. Two spellings of the same fact means every ordering enum and filter has to know which table uses which.

A fresh database is built from the ORM metadata (mgr schema oneshot calls metadata.create_all), so there both timestamps are already NOT NULL. A database that came up through the 2019 migration bae1a7326e8a still allows NULLs, and the NOT NULL sweep in BA-7216 left domains out, so this issue also closes that gap.

## Acceptance Criteria

- A migration renames domains.modified_at to updated_at, backfills NULLs in created_at and updated_at, and sets both NOT NULL.
- DomainRow inherits LifecycleTimestampsMixin and drops its own modified_at declaration.
- DomainOrders, DomainConditions and the legacy GraphQL field specs read updated_at while keeping their public modified_at names.
- DomainData and the v2 DTOs / GraphQL types are unchanged: modified_at stays the API name and carries the value of the renamed column, so no response or schema changes ship.

## Notes

- Renaming the API field is deliberately out of scope - the column name and the wire name are allowed to differ.
- login_client_types is the remaining table that still hand-rolls modified_at; out of scope here.

JIRA Issue: BA-7352

jopemachine:
Resolves #13748 (BA-7352)

## Summary
- Rename `domains.modified_at` to `updated_at` and let `DomainRow` inherit `LifecycleTimestampsMixin` instead of hand-rolling the column, the move `users` / `keypairs` / `groups` / `vfolder_invitations` already made.
- Backfill and set both domain timestamps NOT NULL where they still allow NULLs. A database built by `create_all` already has them, but one that came up through `bae1a7326e8a` does not — #13520 normalized the other tables and left `domains` out. The migration reflects the column first, so a tightened database runs neither the backfill nor the `ALTER`.
- `DomainData`, `DomainOrders` and `DomainConditions` follow the column and take the name `updated_at`.
- The API keeps `modified_at` everywhere it is the wire name — the v1 and v2 DTOs, the GraphQL types, the order enums — and the adapters translate. No response or schema changes ship.

## Test plan
- [x] `ruff format` / `ruff check` (pre-commit)
- [ ] lint, type check and the unit suite on CI
- [ ] `alembic upgrade` / `downgrade` against a local DB — not run locally, no halfstack up

----
📚 Documentation preview 📚: https://sorna-ko--13751.org.readthedocs.build/ko/13751/

----
📚 Documentation preview 📚: https://sorna--13751.org.readthedocs.build/en/13751/


src/ai/backend/manager/models/alembic/versions/c4d1a7e58b32_rename_domains_modified_at_to_updated_at.py
```diff
@@ -0,0 +1,48 @@
...
+branch_labels = None
+depends_on = None
+
+
+def _backfill_and_set_not_null(column: str, backfill: str) -> None:
+    """Tighten ``domains.{column}``, skipping a database that already has it NOT NULL."""
+    inspector = sa.inspect(op.get_bind())
+    reflected = {col["name"]: col for col in inspector.get_columns("domains")}
+    if not reflected[column]["nullable"]:
+        return
```
fregataa, line 31: is this check logic needed?

fregataa (commented):
please align the alembic revision to the current main one

fregataa (approved):

PR #13752, opened 2026-08-13
fix(BA-7350): skip etcd puts for keys whose value is unchanged

jopemachine:
Resolves BA-7350

Backport: 26.8, 26.4

## Summary
- `AsyncEtcd.atomic_replace_prefixes` read the existing subtree only to compute the stale keys, then put every new key back — so re-publishing an identical subtree still issued one put per key. etcd bumps a key's revision on every put even when the value is byte-identical.
- The AppProxy coordinator's `reconcile_traefik_routes` cycle and the manager's `APPPROXY_SYNC` long cycle each re-publish every circuit's routes every 30 seconds, so an idle cluster kept appending revisions for keys nobody changed. On the reported cluster, which runs etcd without auto-compaction, that grew the backend past its 2 GB quota until every write failed with `mvcc: database space exceeded` and app connections broke.
- Current values are now compared against the new ones and only differing keys are put; replacing an unchanged subtree issues no transaction at all. The existing `get_prefix` call already returned the values, so this costs no extra round trip.

Measured against etcd 3.4.22 with a 200-key Traefik-shaped subtree rewritten unchanged: the backend grew ~180 bytes per put, linearly and without bound (14.4 MB after 80k puts). With the values compared first, no transaction is issued and both the database size and the revision stay frozen.

Note that auto-compaction bounds this growth at roughly `write rate × retention` on its own, so a cluster that has it configured would not have hit the quota. This change removes the write amplification itself, which also drops the raft replication, fsync and Traefik watcher wakeups that idle circuits were causing; configuring compaction on deployed etcd clusters remains worthwhile independently.

## Test plan
- [x] `pants fmt lint check` on the changed files
- [ ] `tests/unit/common/test_etcd.py` — new `test_atomic_replace_prefixes_skips_unchanged_keys` asserts the key version stays at 1 after an identical replace and moves to 2 after a changed one (CI)

### DIFF
--- a/src/ai/backend/common/etcd.py
+++ b/src/ai/backend/common/etcd.py
@@ -503,22 +503,10 @@ class AsyncEtcd(AbstractKVStore):
     ) -> None:
         """Replace each subtree under the given prefixes in a single etcd transaction.
 
-        For every ``(prefix, dict_obj)`` pair the subtree rooted at ``prefix`` is
-        replaced by the flattened ``dict_obj``: keys present under the prefix but
-        absent from the new contents are deleted, and keys whose value changes are
-        put. The current-value reads and the delete/put operations are committed as
-        one transaction so watchers (e.g. Traefik) never observe a partially-applied
-        state where a router's backing service has briefly vanished.
-
-        A key already holding the new value is left out of the transaction: etcd
-        bumps the revision on every put even when the value is identical, so
-        re-publishing an unchanged subtree on a periodic reconcile would grow the
-        mvcc history until the backend quota is exceeded. Replacing an unchanged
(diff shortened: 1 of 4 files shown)


src/ai/backend/common/etcd.py
```diff
@@ -522,21 +528,24 @@ async def atomic_replace_prefixes(
         scope_prefix = self._merge_scope_prefix_map(scope_prefix_map)[scope]
```
HyeockJinKim, line 516: The comments here are ridiculously long. Please delete everything except the essential parts.

jopemachine:
Closed as we plan to move these impl to redis

PR #13753, opened 2026-08-13
refactor(BA-7354): adopt the timestamp mixin on the account manager keypairs

jopemachine:
Resolves BA-7354

## Summary

- `KeypairRow` inherits `LifecycleTimestampsMixin` instead of hand-rolling nullable `created_at` plus `modified_at`, matching every manager table since BA-7207
- `account_manager/models/mixins/timestamp.py` mirrors the manager's module — `ai.backend.common` carries no SQLAlchemy dependency, so the two components cannot share one
- A migration renames `keypairs.modified_at` to `updated_at`, backfills NULLs from the sibling timestamp, and sets both columns NOT NULL
- Nothing outside the row declaration read `modified_at` in this component, so no service, handler or response shape changes
- `user_profiles` carries the same hand-rolled pair and moves separately

## Test plan

- [x] `pants lint check` on the changed files
- [x] `KeypairRow` maps to `created_at` / `updated_at`, NOT NULL, with `onupdate` on `updated_at`
- [ ] `alembic -c alembic-accountmgr.ini upgrade head` then `downgrade` round-trips on a populated account manager database

### DIFF
--- a/changes/13753.enhance.md
+++ b/changes/13753.enhance.md
@@ -1,0 +1,1 @@
+Rename the account manager's `keypairs.modified_at` and `user_profiles.modified_at` columns to `updated_at` and make both lifecycle timestamps non-nullable, matching the manager's schema.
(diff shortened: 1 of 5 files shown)


jopemachine:
Closing: this was account_manager work that was not asked for. The manager-side keypair change lives in #13686. BA-7354 stays filed as a backlog item.

PR #13754, opened 2026-08-13
feat(BA-7353): add the kernel creation payload type and event conversion errors

fregataa:
## Summary

Groundwork split out of #13746 so the JSON cutover can be reviewed without the type definitions attached to it. **Nothing changes on the wire here** — the new types have no caller yet, and the existing `serialize()` / `deserialize()` msgpack path is untouched.

- Add `KernelCreationInfo` to `events/event_types/kernel/types.py`, beside the events that carry it. It is the contract with the manager, not a rendering of the agent's resource spec: it carries only what `update_kernel_status_running()` reads — `container_id`, `kernel_host`, the two repl ports, `service_ports` and `occupied_devices`.
- Describe a held device once. `attached_devices` and the per-slot occupancy used to describe the same devices from two sides, both keyed by device name and then by device, so a unit metered along two axes — a `cuda` device is, with `cuda.device` and `cuda.shares` — appeared twice with its attributes duplicated. `OccupiedDevices.units` is keyed by `DeviceName` and then `DeviceId`, and each `OccupiedDevice` carries every slot it supplies.
- Carry the allocation in both units it is expressed in. `allocated` is in the scheduler's — keyed by `ResourceSlotName`, and what `slot_totals` sums into the kernel's occupancy. `processing_units` and `memory_size` are the same allocation in the device's own, mirroring `AbstractComputeDevice`. They cannot share one mapping: `smp` is not a slot, and a GPU's `mem` is not the host memory slot of the same name, so folding them together would put both into the resource accounting.
- Raise `EventPayloadEncodingError` / `EventPayloadDecodingError` from `AbstractEvent.to_message()` / `from_message()` so a caller sees one event-layer error rather than a Pydantic one. An event deriving `BackendAISchema` maps `ValidationError` to a `BackendAIError` of its own, so both forms are funnelled in.

Every field is JSON-representable and round-trips unchanged: amounts stay `Decimal`, mappings are declared `Mapping` so a consumer editing a payload it received is caught by the type checker, and there are no field defaults — a producer states the whole payload.

A non-finite amount is rejected at decode. `Infinity` belongs to a resource policy with `DefaultForUnspecified.UNLIMITED`, not to an allocation — a device supplies a finite share of what it has, and `alloc_map` never produces one.

## Notes for the reviewer

- **Where the device fields come from.** The plugins fill `DeviceModelInfo.data` freely — its declared `ComputedDeviceCapacity` has never matched what anyone writes. Both accelerator plugins in reach (the enterprise CUDA one and the mock) write exactly `smp` and `mem`, which are `AbstractComputeDevice.processing_units` / `.memory_size` projected onto the kernel's share; the intrinsic cpu plugin writes `cores`, which is the kernel's total core count copied onto every device and which nothing reads. Hence two named fields rather than an open mapping, and `cores` dropped.
- **`model_name` is kept** even though it is not part of the creation contract: the GPU usage stats aggregate the device models a kernel ran on.
- **`slot_totals` is a plain property, not a `computed_field`**: a computed field is written into the payload beside the allocations it is derived from, where nothing reads it — a receiver recomputes it — and it can disagree with the value next to it.
- **The middle commits move the payload into a `common/interchange/` package and then back out again.** That package was the answer to "where does an inter-component payload live"; defining it beside its events made it unnecessary. Kept as separate commits so the review discussion stays readable.

## Follow-up for #13746

The manager writes `kernels.attached_devices` as `{"cuda": [{"model_name", "data": {"smp", "mem"}}]}`, and the GPU usage stats read it back through that shape. Wiring this payload up has to reshape it, or those stats go to zero.

## Test plan

- [x] `tests/unit/common/events/` — 41 passed
- [ ] `pants test tests/unit/common/events::`
- [ ] `pants check ::` — the new types are unused by production code in this PR, so type checking is the main guard

Resolves BA-7353


src/ai/backend/common/data/kernel/types.py
```diff
@@ -0,0 +1,123 @@
...
+        on.
+        """
+        slots = ResourceSlot()
+        for alloc_map in self.allocations.values():
+            for slot_name, allocation_by_device in alloc_map.items():
+                if not allocation_by_device:
+                    continue
+                total = sum(allocation_by_device.values(), Decimal(0))
+                slots[SlotName(slot_name)] = str(total)
+        return slots
```
HyeockJinKim: If possible, could you change it from `ResourceSlot` to `ResourceSlotEntry` wherever it appears?

src/ai/backend/common/data/kernel/types.py
```diff
@@ -0,0 +1,123 @@
...
+
+class KernelResourceSpecData(BackendAISchema):
+    """
+    The agent's resource spec for one kernel.
+    """
+
+    slots: list[ResourceSlotEntry] = Field(default_factory=list)
+    allocations: dict[DeviceName, dict[ResourceSlotName, dict[DeviceId, AllocationAmount]]] = Field(
+        default_factory=dict
+    )
```
HyeockJinKim: If possible, it would be best to manage the `Dict` by separating the types.
A three-level dictionary just doesn't seem to be very readable.

src/ai/backend/common/data/kernel/types.py
HyeockJinKim, line 1: I think “schema” might be a better fit than “data,” but even that’s a bit ambiguous.
In my opinion, it would be best to separate only the values that are actually needed so that the values stored in “data” within the Agent are distinct from those passed via the DTO. Could you expand the scope of this task a bit to address this?

HyeockJinKim (changes requested):

src/ai/backend/common/KNOWLEDGE.md
HyeockJinKim, line 1: We don't provide information about sub-packages, though.

src/ai/backend/common/interchange/__init__.py
HyeockJinKim, line 1: In my opinion, the `interchange` package doesn't seem necessary; I think the best approach would be to define a separate type in the `event` section and implement the conversion logic there.

src/ai/backend/agent/resources.py
```diff
@@ -123,7 +123,7 @@ class KernelResourceSpec:
     slots: ResourceSlot
     """Stores the original user-requested resource slots."""
 
-    allocations: MutableMapping[DeviceName, Mapping[SlotName, Mapping[DeviceId, Decimal]]]
+    allocations: MutableMapping[DeviceName, DeviceAllocation]
```
HyeockJinKim: On the event side, it looks like "ResourceSlotName" should be the key.

src/ai/backend/common/events/event_types/kernel/types.py
```diff
@@ -43,3 +55,74 @@ def from_value(cls, value: str | None) -> Self | None:
...
+
+class OccupiedDevices(BackendAISchema):
+    """
+    The devices the kernel occupies.
+
+    Keyed by device name (`cuda`) and then by unit (`0`): `DeviceName` names a kind of
+    device, `DeviceId` one of its units.
+    """
+
+    units: Mapping[DeviceName, Mapping[DeviceId, OccupiedDevice]]
```
HyeockJinKim: Is there no other name besides “Occupied”? It feels like a name we’d previously removed for resource-related reasons is coming back. (We’d decided to stick to using only the names “used,” “free,” and “capacity.” Although “reserved,” “requested,” and “prereserved” have been added now.)

fregataa: I renamed them to "Used..."

HyeockJinKim (approved):

PR #13769, opened 2026-08-13
feat(BA-7363): publish per-user rate limit values to Redis on authorize

jopemachine:
## Summary
- Add `set_user_rate_limit` / `get_user_rate_limit` to `ValkeyRateLimitClient` so the per-user limit is published under a dedicated key namespace (`user-rate-limit.<user_id>`) in the shared rate-limit Redis DB, separate from the access-key rolling counters.
- Publish the limit from `AuthService._create_login_session` on every successful authorize, with the TTL bound to the login session lifetime; a publish failure is logged and does not fail the login.
- Thread `ValkeyRateLimitClient` through the DI chain (`ValkeyDependency` → `ProcessingInput` → `ServiceArgs` → `AuthService`).
- The published value is a fixed default (`10000`) for now; it will be read from the user resource policy once the policy field lands (BA-7362), and the webserver-side enforcement follows in BA-7365.

Resolves BA-7363.


jopemachine:
Closing: the webserver-side limiter (#13771) now uses a fixed default limit and keys the existing rolling counter by the session user ID directly, so publishing per-user limit values to Redis is unnecessary at this stage. When BA-7362 adds the per-user limit to the user resource policy, we will revisit how to deliver the value to the webserver (this PR can be reopened then if Redis publication is still the chosen mechanism).

PR #13770, opened 2026-08-13
feat(BA-7364): deliver the user ID to the webserver session at login

jopemachine:
## 📚 Stacked PRs

1. 👉 **#13770** — `feat(BA-7364): deliver the user ID to the webserver session at login` ← you are here
2. ⬇️ #13771 — `feat(BA-7365): add a per-user rate limit middleware to the webserver`

## Summary
- Add an optional `user_id` field (`UserID` identifier type) to `AuthSuccessResponse`; optional keeps the wire format compatible with managers that do not send it yet.
- Fill it in the manager REST `authorize` handler from `AuthorizationResult.user_id`, which the auth service already returns.
- Store `user_id` in the webserver session token in both login paths (credential login and token login), so the webserver-side rate limiter (#13771) can key per-user counters on it.

Resolves BA-7364.

### DIFF
--- a/changes/13770.feature.md
+++ b/changes/13770.feature.md
@@ -1,0 +1,1 @@
+Include the user ID in the login authorize response and keep it in the webserver session so that webserver-side features can identify the user without extra manager queries.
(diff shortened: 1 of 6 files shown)


jopemachine:
Folded into #13771 — the two changes are small and only meaningful together, so they now ship as a single PR.

PR #13777, opened 2026-08-14
feat(BA-7364): deliver the login user id in the auth response

jopemachine:
## 📚 Stacked PRs

This PR is part of a 5-PR stack. **Merge in order:**

1. 👉 **#13777** — `feat(BA-7364): deliver the login user id in the auth response` ← you are here
2. ⬇️ #13778 — `feat(BA-7365): key the rate limit counter by user id`
3. ⬇️ #13779 — `feat(BA-7362): move the per-user API rate limit to the user resource policy`
4. ⬇️ #13784 — `feat(BA-7363): publish the per-user rate limit to the shared Redis DB`
5. ⬇️ #13771 — `feat(BA-7365): add a per-user rate limit middleware to the web server`

## Summary
- Add a required `user_id` (`UserID`) field to `AuthSuccessResponse`, filled in the manager REST `authorize` handler, and store it in the web server session token in both login paths. The web server needs the login user's identity to key a per-user rate limit counter without asking the manager again.
- Type the identifiers the auth result already carried: `AuthorizationResult` and `AuthSuccessResponse` now use `UserID`, `AccessKey`, `SecretKey` and `UserRole` instead of bare `uuid.UUID` / `str`.
- Drop the dead `or ""` fallbacks on `keypairs.secret_key`, which is NOT NULL.

The rate limit value is deliberately not part of this response (see #13777 review): it is a resource policy value rather than part of who authenticated, and a session snapshot would go stale — the manager reads the policy per request, while a web server session would hold the login-time value for up to a week. The web server will read it from the shared rate limit Redis DB instead (BA-7363).

Resolves BA-7364.

### DIFF
--- a/src/ai/backend/common/dto/manager/auth/types.py
+++ b/src/ai/backend/common/dto/manager/auth/types.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 from enum import StrEnum
 from typing import Any, Self
 
+from ai.backend.common.identifier.user import UserID
 from ai.backend.common.types import BackendAISchema
 
 __all__ = (
@@ -55,9 +56,13 @@ class AuthSuccessResponse(AuthResponse):
     role: str
     status: str
     session_token: str
+    # Optional for wire compatibility with managers that do not send them yet.
+    user_id: UserID | None = None
+    # The keypair rate limit (requests per window); None means unlimited.
+    rate_limit: int | None = None
     type: AuthTokenType = AuthTokenType.KEYPAIR
 
-    def to_dict(self) -> dict[str, str]:
+    def to_dict(self) -> dict[str, Any]:
         return self.model_dump(mode="json")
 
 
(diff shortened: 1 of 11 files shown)


seedspirit (dismissed):

src/ai/backend/common/dto/manager/auth/types.py
```diff
@@ -50,14 +52,16 @@ def parse(cls, data: dict[str, Any]) -> Self:
...
-    access_key: str
-    secret_key: str
-    role: str
+    access_key: AccessKey
+    secret_key: SecretKey
+    role: UserRole
     status: str
     session_token: str
+    user_id: UserID
+    rate_limit: int | None
```
HyeockJinKim: Adding a rate limit to the response seems unusual. If we need to pass rate limit or current counter information, we could use headers or similar methods, and using Redis to share data between the web server and the manager seems like the appropriate approach. Including this value would actually make it harder to modify later, so I’ll submit a request for changes.

HyeockJinKim (changes requested):

fregataa (approved):

seedspirit (approved):

HyeockJinKim (approved):
I think it would be possible to assign a user ID for the user's information, but I don't think the web server should be implemented in a way that relies on that value.

PR #13779, opened 2026-08-14
feat(BA-7362): move the per-user API rate limit to the user resource policy

jopemachine:
## 📚 Stacked PRs

This stack is 3 open PRs. **Merge in order:**

1. ✅ #13777 — `feat(BA-7364): deliver the login user id in the auth response` (merged)
2. 👉 **#13779** — `feat(BA-7362): move the per-user API rate limit to the user resource policy` ← you are here
3. ⬇️ #13778 — `feat(BA-7365): key the rate limit counter and the published limit by user id`
4. ⬇️ #13771 — `feat(BA-7365): add a per-user rate limit middleware to the web server`

## Summary
- Add a nullable `max_api_requests_per_window` column to `user_resource_policies` (NULL means unlimited) and plumb it through the v2 stack: row, data type, creator/updater specs, filters and orders, v2 DTOs, GQL v2 node/mutation/filter types, the adapter, and the admin CLI.
- Nothing enforces the new field yet: the rate limiter still reads `keypairs.rate_limit` at this point of the stack. #13778 switches it to the policy field in the same step that keys the counter by user, so the ceiling and the counter move to the user together.
- `keypairs.rate_limit` is deprecated, not removed: the column and its API fields are still written and returned. The legacy GQL field carries a `deprecation_reason` and the v2 keypair DTO fields are marked `deprecated=True`; their texts describe the end state of the stack.

Existing deployments start with the new column NULL, which means unlimited — an operator must set the policy value to restore a limit.

Resolves BA-7362.

----
📚 Documentation preview 📚: https://sorna--13779.org.readthedocs.build/en/13779/

----
📚 Documentation preview 📚: https://sorna-ko--13779.org.readthedocs.build/ko/13779/

### DIFF
--- a/changes/13779.feature.md
+++ b/changes/13779.feature.md
@@ -1,0 +1,1 @@
+Move the per-user API rate limit from the keypair to the user resource policy field max_api_requests_per_window; the keypair rate limit is deprecated and no longer enforced.
--- a/docs/manager/graphql-reference/schema.graphql
+++ b/docs/manager/graphql-reference/schema.graphql
@@ -1216,7 +1216,6 @@ type KeyPair implements Item {
   resource_policy: String
   created_at: DateTime
   last_used: DateTime
-  rate_limit: Int
   num_queries: Int
 
   """Added in 24.09.0."""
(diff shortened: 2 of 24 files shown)


jopemachine:
Closing as not planned for now: moving the limit to the user resource policy is deferred, and #13778 / #13771 proceed with the limit staying on the keypair. The branch is kept for when this is picked up again.

PR #13788, opened 2026-08-17
fix(manager): correct check-presets per-sgroup remaining for agent-less resource groups

yomybaby:
Fixes the per-scaling-group remaining that the WebUI dashboard derives from `POST /resource/check-presets`.

## Agent-less resource groups reported policy headroom (often `Infinity`) as free capacity

`_fetch_all_check_presets_data` clamps each scaling group's agent-derived remaining with `min_quantities(final_remaining, sg_remaining)`. `min_quantities` keys off its **first** list and substitutes that list's own value for slots missing from later lists (`repositories/resource_slot/types.py`). Since `_get_agent_available_resources` returned an **empty** list for a scaling group with zero ALIVE+schedulable agents (and omitted any slot type no agent provides), the clamp inverted: `scaling_groups[rg].remaining` degenerated to `final_remaining` — pure policy headroom, serialized as `"Infinity"` under default unlimited policies.

User-visible effect: the WebUI "My Resources in Resource Group" panel showed **Free: Unlimited** (or the full policy remaining, e.g. "108 cores free") for a resource group whose agents are all down or unschedulable — when nothing can actually be scheduled there. The same leak inflated `scaling_group_remaining`.

Fix: zero-fill `per_sgroup_remaining` with the known slot types before returning, mirroring the existing zero-fill of per-sgroup *occupancy* (`_get_user_session_occupancy_per_sgroup`). A slot no alive schedulable agent provides now clamps to 0, as the panel semantics require.

No response schema change: the fix only corrects values in existing fields.

> An earlier revision of this PR also added a `domain_limits` field to the response. It was dropped because it bypassed the `api.resources.group_resource_visibility` gate that hides project limits from regular users; a visibility policy for domain limits needs to be decided first.

## Verification

- CI: full `pants` gate (lint / check / unit / component / integration) green on the previous revision; re-running for this one.
- `ruff format --check` / `ruff check` (0.14.0, the pants-pinned version) on the changed file: clean.
- `python3 -m py_compile` on the changed file: OK.
- No repository test asserts the old fallback behavior (checked `tests/unit/manager/repositories/resource_preset/test_check_presets.py`; its assertions look up specific slots and are unaffected by the zero-fill entries).

- [x] Backport targets: none — behavior fix paired with a WebUI change; Backport: none
- [ ] Installer updates
- [x] API server-client counterparts — no schema change; the WebUI counterpart is lablup/backend.ai-webui#8830.
- [ ] Test case(s)
- [ ] Documentation


yomybaby:
FYI: the `copilot-pull-request-reviewer` check failure on `a21fbb6` is not caused by this PR — the job aborted before reviewing with HTTP 402 `You have exceeded your monthly quota` from the Copilot API (org-level Copilot code-review quota exhausted). All other check suites on this head are green. The check can be re-run once the quota resets; re-running now would hit the same error.

---
_Generated by [Claude Code](https://claude.ai/code)_

HyeockJinKim (changes requested):
If a user accesses the "domain limits," they would be able to view the limit information for that domain, and I'm not sure if that's appropriate.
+ We need to take a permission-based approach that accounts for RBAC in the new query, but since the current legacy implementations bypass permissions, it feels a bit awkward to add new ones as well.

yomybaby:
@HyeockJinKim Thanks for the review — addressed in 35f13d8b.

**On exposing domain limits to users:** two mitigating facts, and a change:

- The response's `keypair_remaining` is already `min(keypair, group, domain remaining)`, so whenever the domain cap actually binds, a user can already observe it indirectly — `domain_limits` mostly makes the limit that is *enforced on them* displayable (the WebUI tooltip promises "the most restrictive of keypair/project/domain limits" but couldn't show a consistent total).
- This endpoint already exposes project-wide `group_limits` to every member, gated by the `config/api/resources/group_resource_visibility` etcd flag (NaN-masked when off).

The pushed commit now masks `domain_limits` under that **same existing flag**: deployments that consider shared limits sensitive hide domain limits together with group limits, with no new knob. The WebUI counterpart already treats NaN as absent, so it degrades to the current behavior.

**On RBAC in legacy:** agreed — that's why the PR deliberately adds no new query or permission path. The domain usage was already fetched internally to compute `final_remaining`; the change only surfaces an already-computed value under the endpoint's existing visibility gate. A properly RBAC-scoped exposure of scope limits belongs to the v2 API surface, and I'd keep it out of this legacy endpoint.

Happy to drop the field to admin-only or remove it entirely if you still prefer — but then the WebUI total falls back to overstating the allocatable amount whenever a domain quota binds (the bug this pair of PRs fixes).

---
_Generated by [Claude Code](https://claude.ai/code)_

yomybaby:
@HyeockJinKim Thanks for the review. Agreed on both points, so I dropped `domain_limits` from this PR entirely rather than masking it (7659fc32).

The PR now only fixes the per-scaling-group remaining for resource groups with no alive schedulable agent (`Infinity` → `0`); the response schema is unchanged. Exposing domain limits to users will be revisited separately, on the v2 API surface with proper RBAC scoping.

yomybaby:
FYI: the `labeler` check failure on `7659fc3` is a GitHub API flake, not this PR — `size-label-action` got an HTTP 500 (`Unexpected end of JSON input`) POSTing to the labels endpoint, and the `size:S` label it was applying is already on the PR. Safe to ignore or re-run.

---
_Generated by [Claude Code](https://claude.ai/code)_

HyeockJinKim (approved):
I think we need to change the part where the "known slot types" are set, but since that doesn't seem to fall within the scope of this task, I'll merge it.

PR #13796, opened 2026-08-18
fix(BA-7371): resolve container cgroup paths from /proc/<pid>/cgroup

fregataa:
## Summary

- The agent assembled a container's cgroup path from the runtime's cgroup driver, assuming Docker's layout — `docker/<id>` for `cgroupfs`, `system.slice/docker-<id>.scope` for `systemd`. Rootful Podman uses `machine.slice/libpod-<id>.scope`, so the path never existed and container-level metrics were silently never collected.
- The same Docker-shaped assumption was baked into `get_container_pids()` and `get_container_id_of_cgroup()`, breaking the container-PID ↔ host-PID mapping in both directions.
- The path is now resolved from the cgroup the container's main process actually belongs to (`/proc/<pid>/cgroup`), which holds for every runtime and layout observed. `get_cgroup_path()` becomes async and caches the resolution per container, so the extra inspect costs one call per container lifetime.
- Hardcoded layout knowledge survives only in `get_container_id_of_cgroup()`, which must parse a cgroup path back into an identifier. It now recognizes the Docker, Podman and cri-o forms plus the interposed `container` / `conmon` leaves.
- `controller` becomes a `CgroupController` enum and `container_id` a `ContainerId`.

Observed layouts, all read from `/proc/<pid>/cgroup` on a Podman testbed:

```
docker, cgroupfs        docker/<id>
docker, systemd         system.slice/docker-<id>.scope
podman rootful          machine.slice/libpod-<id>.scope
podman (reporter env)   machine.slice/libpod-<id>.scope/container
podman --cgroups=split  user.slice/.../session-N.scope/libpod-payload-<id>
```

Because the layout varies with runtime *and* configuration — not just the Podman major version — adding another hardcoded branch would break on the next variant.

## Compatibility

`get_cgroup_path()` changes signature and becomes async. It has no callers outside `agent/docker/intrinsic.py`; no accelerator plugin in this repository or in `plugins/` (including cuda-enterprise) references it, and nothing outside the agent subclasses `AbstractAgent`. The compute-plugin interface `gather_container_measures(container_ids: Sequence[str])` is deliberately left untouched, so plugins are unaffected.

`CgroupResolutionFailed` now derives from `BackendAIError` rather than `RuntimeError`; the one call site that caught `RuntimeError` around a cgroup lookup (`vendor/linux.py`) was updated to keep its fallback path working.

## Test plan

- [x] `tests/unit/common/test_cgroup.py` — 22 new tests: identifier extraction across the Docker/Podman/cri-o layouts above and their leaf variants, non-container cgroups returning `None`, path composition, controller propagation, failure propagation, and the v1 `tasks` / v2 `cgroup.procs` split.
- [x] `tests/unit/agent/test_docker_intrinsic.py` — existing tests pass with the mocks updated for the async signature.
- [x] Verify container-level metrics are actually collected on a Podman-backed agent.
- [x] Regression check on a Docker-backed agent.

## Known gap

`vendor/linux.py` still derives the agent-level cpuset from a hardcoded `docker` / `system.slice` parent. It reads the wrong cgroup on Podman, but degrades through an existing fallback to the scheduler affinity mask rather than failing, so it is left for a separate change.

Resolves BA-7371

### DIFF
--- a/src/ai/backend/agent/docker/agent.py
+++ b/src/ai/backend/agent/docker/agent.py
@@ -109,7 +109,11 @@ from ai.backend.agent.utils import (
     update_nested_dict,
 )
 from ai.backend.common.asyncio import current_loop
-from ai.backend.common.cgroup import CgroupController, get_container_cgroup_path
+from ai.backend.common.cgroup import (
+    CgroupController,
+    get_cgroup_path_of_pid,
+    get_container_main_pid,
+)
 from ai.backend.common.data.image.types import InstalledImageInfo
 from ai.backend.common.docker import (
     MAX_KERNELSPEC,
@@ -176,7 +180,14 @@ _SECCOMP_PROFILE_FILENAME: Final[str] = "seccomp.json"
 # document, others open it as a path, and neither accepts the other form. These engine
 # components, as reported by the version API, are the ones that require a path.
 _SECCOMP_PATH_ENGINES: Final[frozenset[str]] = frozenset({"Podman Engine"})
-_CGROUP_PATH_CACHE_SIZE: Final[int] = 256
+# Cached per container, so the capacity is the number of containers an agent may host.
+_CGROUP_PATH_CACHE_SIZE: Final[int] = 2048
+# Controllers the intrinsic plugins read, resolved together from a single inspect.
+_TRACKED_CGROUP_CONTROLLERS: Final[tuple[CgroupController, ...]] = (
+    CgroupController.CPUACCT,
+    CgroupController.MEMORY,
+    CgroupController.BLKIO,
+)
 
 known_glibc_distros: Final[dict[float, str]] = {
     2.17: "centos7.6",
@@ -1497,7 +1508,7 @@ class DockerAgent(AbstractAgent[DockerKernel, DockerKernelCreationContext]):
     gwbridge_subnet: str | None
     checked_invalid_images: set[str]
     _seccomp_profile_as_path: bool
-    _cgroup_path_cache: LRUCache[tuple[CgroupController, ContainerId], Path]
+    _cgroup_path_cache: LRUCache[ContainerId, dict[CgroupController, Path]]
 
     network_plugin_ctx: NetworkPluginContext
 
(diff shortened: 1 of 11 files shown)


fregataa:
Superseded by #13798. The change is a refactor of how cgroup paths are resolved rather than a fix for BA-7371, so it was re-filed under BA-7378. BA-7371 stays open: its default `stats-type = "docker"` path never touches cgroup paths and still needs its own Podman compatibility work.

src/ai/backend/common/cgroup.py
```diff
@@ -39,65 +82,94 @@ async def get_docker_cgroup_version() -> CgroupVersion:
...
     cgroup_version = await get_docker_cgroup_version()
     log.debug("Cgroup version: {}, {}", cgroup_version.version, cgroup_version.driver)
-    tasks_path: Path
-    match (cgroup_version.version, cgroup_version.driver):
-        case ("2", "systemd"):
-            tasks_path = Path(f"/sys/fs/cgroup/system.slice/docker-{cid}.scope/cgroup.procs")
-        case ("2", "cgroupfs"):
-            tasks_path = Path(f"/sys/fs/cgroup/docker/{cid}/cgroup.procs")
-        case ("1", _):
-            tasks_path = Path(f"/sys/fs/cgroup/pids/docker/{cid}/tasks")
```
HyeockJinKim, line 101: Will this logic become completely unnecessary? When considering only the container scenario, I’m actually starting to wonder if it was ever necessary in the first place, but could you please check it for me?

src/ai/backend/agent/docker/agent.py
```diff
@@ -1655,19 +1671,28 @@ async def _write_kernel_registry_to_recovery(
...
-        mount_point = get_cgroup_mount_point(version, controller)
-        # See https://docs.docker.com/config/containers/runmetrics/#find-the-cgroup-for-a-given-container
-        match driver:
-            case "cgroupfs":
-                cgroup = f"docker/{container_id}"
-            case "systemd":
-                cgroup = f"system.slice/docker-{container_id}.scope"
-            case _:
-                raise ValueError(f"Unsupported cgroup driver: {driver!r}")
-        return mount_point / cgroup
```
HyeockJinKim, line 1700: Please take a look at this as well.

HyeockJinKim (dismissed):

HyeockJinKim (approved):

PR #13798, opened 2026-08-18
refactor(BA-7378): resolve container cgroup paths from /proc/<pid>/cgroup

fregataa:
## Summary

- The agent assembled a container's cgroup path from the runtime's cgroup driver, which encodes one runtime's layout. Every runtime or configuration that lays cgroups out differently needs another hardcoded branch.
- The path is now read from the cgroup the container's main process actually belongs to, so a single lookup covers all layouts without the agent knowing which runtime it talks to.
- `get_cgroup_path()` becomes async and caches the resolution per container, resolving every controller the intrinsic plugins read from one inspect and dropping the entry when the kernel is cleaned up. `get_container_pids()` reuses the same resolution.
- Layout knowledge survives only in `get_container_id_of_cgroup()`, which maps a cgroup path back to an identifier and therefore cannot delegate. It now parses the Docker, Podman and cri-o forms plus the interposed `container` / `conmon` leaves.
- `controller` becomes a `CgroupController` enum and `container_id` a `ContainerId`.

Five distinct layouts, each read back from `/proc/<pid>/cgroup` on a real host:

```
docker, cgroupfs        docker/<id>
docker, systemd         system.slice/docker-<id>.scope
podman rootful          machine.slice/libpod-<id>.scope
podman (other setups)   machine.slice/libpod-<id>.scope/container
podman --cgroups=split  user.slice/.../session-N.scope/libpod-payload-<id>
```

## Compatibility

`get_cgroup_path()` changes signature and becomes async. It has no callers outside `agent/docker/intrinsic.py`; no accelerator plugin in this repository or in `plugins/` (including cuda-enterprise) references it, and nothing outside the agent subclasses `AbstractAgent`. The compute-plugin interface `gather_container_measures(container_ids: Sequence[str])` is deliberately left untouched, so plugins are unaffected.

`CgroupResolutionFailed` derives from `BackendAIError` rather than `RuntimeError`; the one call site that caught `RuntimeError` around a cgroup lookup (`vendor/linux.py`) was updated so its fallback still works.

## Test plan

- [x] `tests/unit/common/test_cgroup.py` — 22 tests covering identifier extraction across the layouts above and their leaf variants, non-container cgroups returning `None`, path composition, controller propagation, failure propagation, and the v1 `tasks` / v2 `cgroup.procs` split.
- [x] `tests/unit/agent/test_docker_intrinsic.py` — passes with mocks updated for the async signature.
- [x] Verified end to end on a rootful Podman host (Podman 4.9.3, agent as root, `stats-type = "cgroup"`). Running the previous logic on the same agent and container produced 4 `sysfs unreadable` warnings and 0 `backendai_container_utilization` samples; the new resolution produced 10 samples with live values (`cpu_util 2843.95`, `mem 1.097e8` of `2.147e9`, plus io and net counters) and no warnings.
- [ ] Regression check on a Docker-backed agent in `stats-type = "cgroup"` mode.

## Scope

This changes how a cgroup path is resolved. It is not a Podman-compatibility fix for the agent as a whole:

- The default `stats-type` is `docker`, which reads the container statistics API and never touches cgroup paths. Podman compatibility for that mode is separate work.
- `agent/vendor/linux.py` still derives the agent-level cpuset from a hardcoded `docker` / `system.slice` parent. On Podman it does not fail — it silently reads the host's `system.slice` cpuset, which is the wrong cgroup.
- Some accelerator plugins iterate `/sys/fs/cgroup/pids/docker` directly.

Relates to BA-7378

### DIFF
--- a/changes/13798.enhance.md
+++ b/changes/13798.enhance.md
@@ -1,0 +1,1 @@
+Resolve a container's cgroup path from the cgroup its main process belongs to, so container statistics and PID mapping work across container runtimes instead of only the Docker cgroup layout.
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -87,6 +87,7 @@ from ai.backend.agent.tasks import (
 )
 from ai.backend.common.asyncio import cancel_tasks, current_loop
 from ai.backend.common.bgtask.bgtask import BackgroundTaskManager, BackgroundTaskManagerArgs
+from ai.backend.common.cgroup import CgroupController
 from ai.backend.common.clients.valkey_client.valkey_bgtask.client import ValkeyBgtaskClient
 from ai.backend.common.clients.valkey_client.valkey_container_log.client import (
     ValkeyContainerLogClient,
@@ -1944,7 +1945,9 @@ class AbstractAgent[
         )
 
     @abstractmethod
-    def get_cgroup_path(self, controller: str, container_id: str) -> Path:
+    async def get_cgroup_path(
+        self, controller: CgroupController, container_id: ContainerId
+    ) -> Path:
         """
         Get the cgroup path for the given controller and container ID.
         This is used to read/write cgroup files for resource management.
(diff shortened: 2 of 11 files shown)


fregataa:
Reverted to #13796. The change fixes the environment BA-7371 was reported from (rootful Podman with the agent running as root and `stats-type = "cgroup"`), so it belongs to BA-7371 as a fix rather than a separate refactor. BA-7378 is closed as not planned.

PR #13817, opened 2026-08-19
fix(BA-7377): take the kernel container log driver from configuration

fregataa:
## Summary

- The kernel container config hardcoded `LogConfig.Type = "local"`, a driver only Docker implements. Runtimes serving the Docker-compatible API without it reject the create request, so every compute session fails to start.
- Adds `container-logs.driver` to the agent configuration (`local` | `json-file`), defaulting to `local` so existing Docker deployments are unaffected.
- The payload is now built through a typed `LogConfig` model and dumped to a plain dict at the point it enters the container creation request, so `update_nested_dict` merging and JSON encoding are unchanged.

## Verification matrix

| Driver | Docker | Podman |
|---|---|---|
| `local` (default) | ✅ session runs — `{"Type":"local","Config":{"max-size":"2m","max-file":"5","compress":"false"}}`, identical to the previous hardcoded payload | ❌ rejected at create — `HTTP 500 invalid log driver: invalid argument` (the bug this PR fixes) |
| `json-file` | ✅ session runs — `{"Type":"json-file","Config":{"max-size":"2m","max-file":"5","compress":"false"}}`; `session logs` returns kernel output | ⚠️ accepted at create (`HTTP 201`); full session not re-run here |

### How the rotation options land

| | `max-size` (from `max-length`) | `max-file` | `compress` |
|---|---|---|---|
| Docker, `local` / `json-file` | applied | applied (5 files) | applied |
| Podman, `json-file` | applied (`Size: "2MB"`) | no equivalent | no equivalent |

Podman keeps a single `ctr.log` per container and caps it by size only, so `max-length` takes
effect but the "split 10M across 5 files" half of the intent does not. This is documented on the
`container-logs.driver` config description.

**How each cell was produced**

- **Docker** — Backend.AI agent 26.8.0rc1 on OrbStack (kernel 7.0.14, aarch64). Sessions created through `./bai session enqueue` against a live local stack, then `docker inspect <kernel container> --format '{{json .HostConfig.LogConfig}}'`. `agent.toml` was restored and the agent restarted afterwards.
- **Podman** — the `HTTP 500` / `HTTP 201` results are the create-time driver-acceptance measurements from the BA-7377 report (Podman 4.9.3 via `podman.socket`, Ubuntu 24.04, x86_64). They were not re-run as full sessions for this PR.
- `max-size: 2m` is the `container-logs.max-length` default (10M) split across 5 files.

`journald` was considered and dropped: it takes no size-based rotation options at all, which would
have required the log config to carry an optional-and-usually-empty options block for a driver we
could not verify end to end.

## Test plan

- [x] `pants test tests/unit/agent/test_docker_agent.py::TestBuildLogConfig` — per-driver payload and dump keys
- [x] Docker agent, default config — `LogConfig` unchanged from before this PR
- [x] Docker agent, `driver = "json-file"` — session starts and `./bai session logs` works
- [ ] Podman-backed agent, `driver = "json-file"` — end-to-end session run

Resolves BA-7377

### DIFF
--- a/src/ai/backend/agent/config/unified.py
+++ b/src/ai/backend/agent/config/unified.py
@@ -78,7 +78,6 @@ class ContainerSandboxType(enum.StrEnum):
 class ContainerLogDriver(enum.StrEnum):
     LOCAL = "local"
     JSON_FILE = "json-file"
-    JOURNALD = "journald"
 
 
 class ScratchType(enum.StrEnum):
@@ -1933,9 +1932,7 @@ class ContainerLogsConfig(BaseConfigSchema):
                 "'local' is Docker-only and stores logs in an efficient binary format. "
                 "'json-file' is accepted by both Docker and Podman, but Podman keeps a "
                 "single size-capped log per container, so only 'max-length' takes effect "
-                "there while file count and compression have no equivalent. "
-                "'journald' does not support size-based rotation, "
-                "so 'max-length' is not applied to it."
+                "there while file count and compression have no equivalent."
             ),
             added_version="26.4.10",
             example=ConfigExample(local="local", prod="local"),
--- a/src/ai/backend/agent/config/unified.py
+++ b/src/ai/backend/agent/config/unified.py
@@ -75,6 +75,12 @@ class ContainerSandboxType(enum.StrEnum):
     JAIL = "jail"
 
 
+class ContainerLogDriver(enum.StrEnum):
+    LOCAL = "local"
+    JSON_FILE = "json-file"
+    JOURNALD = "journald"
+
+
 class ScratchType(enum.StrEnum):
     HOSTDIR = "hostdir"
     HOSTFILE = "hostfile"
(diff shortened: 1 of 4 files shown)


src/ai/backend/agent/config/unified.py
```diff
@@ -75,6 +75,12 @@ class ContainerSandboxType(enum.StrEnum):
     JAIL = "jail"
 
 
+class ContainerLogDriver(enum.StrEnum):
+    LOCAL = "local"
+    JSON_FILE = "json-file"
+    JOURNALD = "journald"
```
HyeockJinKim: Since `journald` doesn't seem like a particularly good option for production use, I'll remove it.

src/ai/backend/agent/config/unified.py
```diff
@@ -1916,6 +1922,25 @@ def _parse_affinity_policy(cls, v: Any) -> AffinityPolicy:
...
+        ),
+        BackendAIConfigMeta(
+            description=(
+                "Log driver used for kernel containers. "
+                "'local' is Docker-only and stores logs in an efficient binary format. "
+                "'json-file' is accepted by both Docker and Podman, but Podman keeps a "
+                "single size-capped log per container, so only 'max-length' takes effect "
+                "there while file count and compression have no equivalent. "
+                "'journald' does not support size-based rotation, "
+                "so 'max-length' is not applied to it."
```
HyeockJinKim: It is said that `journald` does not handle size individually but instead integrates it into `docker.service`.

HyeockJinKim (approved):

PR #13829, opened 2026-08-20
fix(BA-7394): attach GPUs through CDI device requests on Podman

fregataa:
## Summary

- Podman's Docker-compatible API honours neither `HostConfig.Runtime` nor nvidia device requests, so the GPU was never attached — the container started, the `cuda.device` slot stayed charged, and no error surfaced.
- Detect the engine from the version API component list instead of parsing the advertised server version. Podman fills that field with its own version (`4.9.3`, `5.4.0`), which always fell below the Docker 19.3 threshold and pinned the plugin to the pre-19.3 branch on every Podman release.
- An engine that attaches devices only through CDI now emits a CDI device request, and disables the plugin below the version whose compatible API honours it (Podman 5.4.0, containers/podman#25171). Reporting a `cuda.device` capacity of 0 keeps GPU sessions off the node while CPU workloads keep running.
- Introduce `DeviceAttachMechanism` so each mechanism owns how it names the allocated devices. CDI resolves the device UUID because the CUDA runtime index and the index `nvidia-ctk` assigns are not guaranteed to agree on a multi-GPU host.

## Test plan

Unit:

- [x] `pants test tests/unit/accelerator/cuda_open/test_device_attach_mechanism.py`

On a live agent — RTX 5090, driver 580.173.02 open modules, nvidia-container-toolkit 1.20.0, Ubuntu 24.04. Halfstack and the manager stayed on Podman 4.9.3; only the agent was pointed at the engine under test. The 5.4.0 engine deliberately does **not** register the nvidia runtime, so `/info` never advertises it.

Podman 5.4.0, the declared minimum:

- [x] The plugin selects CDI and enables itself without the nvidia runtime being registered — `attaching GPUs via the cdi mechanism`
- [x] The agent reports the slot — `Resource slots: {'cuda.device': 1, ...}`
- [x] A GPU session reaches RUNNING and the engine records 7 devices on the kernel container
- [x] The nvidia device nodes are present inside the container
- [x] `nvidia-smi` reports the GPU, and a driver-API workload allocates device memory, memsets it and reads it back
- [x] The CDI hooks ran: `libcuda.so.1` inside the container is a symlink to `libcuda.so.580.173.02`, while the bind mounts land only on the versioned filenames — the engine applied the spec rather than the plugin emulating it
- [x] No `NVIDIA_*` environment variable reaches the container
- [x] The allocation is released and reusable — a second GPU session is scheduled and sees the GPU after the first terminates

Podman 4.9.3, below the minimum:

- [x] The plugin refuses the engine — `Podman Engine 4.9.3 ignores CDI device requests; 5.4.0 or later is required`
- [x] The agent stays up and reports `cuda.device: 0`, so no GPU session is scheduled to it

Docker 29.7.2 with the nvidia runtime registered, on a separate host (same GPU model, driver 580.178.04):

- [x] The plugin selects the nvidia device driver — `attaching GPUs via the nvidia-driver mechanism`
- [x] The emitted request is unchanged from before this PR — `DeviceRequests: [{"Driver": "nvidia", "DeviceIDs": ["0"], "Capabilities": [["utility", "compute", "video", "graphics", "display"]]}]`, keyed by the runtime index rather than the UUID, with no `Env` and no explicit device mounts
- [x] A GPU session reaches RUNNING; the nvidia device nodes are present, `nvidia-smi` reports the GPU and the driver-API workload passes
- [x] The allocation is released and reusable

## Notes

- Backport target not decided yet — this is a `fix:` PR, so without a `Backport:` trailer it would go to every maintained version (26.8, 26.4).

Resolves BA-7394

### DIFF
--- a/src/ai/backend/accelerator/cuda_open/plugin.py
+++ b/src/ai/backend/accelerator/cuda_open/plugin.py
@@ -1,18 +1,25 @@
 import logging
 import re
 import uuid
+from abc import ABC, abstractmethod
 from collections.abc import Collection, Mapping, MutableMapping, Sequence
 from decimal import Decimal
 from pathlib import Path
 from pprint import pformat
 from typing import (
     Any,
+    Final,
+    final,
+    override,
 )
 
 import aiodocker
+import yaml
 from aiodocker.exceptions import DockerError
 from aiotools import closing_async
+from pydantic import BaseModel, Field
 
+from ai.backend.agent.errors.resources import InvalidResourceArgument
 from ai.backend.agent.resources import (
     AbstractAllocMap,
     AbstractComputeDevice,
@@ -20,7 +27,6 @@ from ai.backend.agent.resources import (
     DeviceSlotInfo,
     DiscretePropertyAllocMap,
 )
-from ai.backend.agent.utils import update_nested_dict
 from ai.backend.logging import BraceStyleAdapter
 
 try:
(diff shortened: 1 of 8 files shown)


HyeockJinKim (commented):
I'll add a note about Podman compatibility to README.md.

src/ai/backend/accelerator/cuda_open/plugin.py
```diff
@@ -92,38 +272,28 @@ class CUDAPlugin(AbstractComputePlugin):
...
-            return
-
-        rx_triple_version = re.compile(r"(\d+\.\d+\.\d+)")
-        m = rx_triple_version.search(docker_info["ServerVersion"])
-        if m:
-            self.docker_version = tuple(map(int, m.group(1).split(".")))
-        else:
-            log.error("could not detect docker version!")
+        attach_mechanism = self._detect_attach_mechanism(docker_info, version_info)
+        if attach_mechanism is None:
```
HyeockJinKim: No matter how much I think about it, "attach" is definitely a verb, but I can't help wondering if "attached" might be better.

src/ai/backend/accelerator/cuda_open/plugin.py
```diff
@@ -84,6 +128,142 @@ def __repr__(self) -> str:
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def build_device_config(
+        self,
+        device_alloc: Mapping[SlotName, Mapping[DeviceId, Decimal]],
+        devices: Collection[CUDADevice],
+    ) -> Mapping[str, Any]:
+        """The container creation config that attaches the allocated devices."""
+        raise NotImplementedError
```
HyeockJinKim, line 185: Wouldn't it be better not to use mapping for the new section we're working on?

HyeockJinKim (changes requested):

HyeockJinKim (approved):

PR #13835, opened 2026-08-20
fix(BA-7401): seed the virtual-scope chain in the example fixtures

fregataa:
## Missing data

The example fixtures declare domains, projects, users and roles but never the virtual-scope chain, so a fresh install leaves `virtual_scopes`, `entity_memberships` and `scope_bindings` empty. Nothing derives them while the fixtures load, and the backfill migration never runs on a fresh install — `mgr schema oneshot` creates the tables and stamps head.

Rows added, mirroring what the runtime write paths produce:

| Relation | Rows | Mirrors |
|---|---|---|
| virtual scope per domain / project / user, each with its self membership and self binding | 8 VS, 8 + 8 edges | `RBACWriteOps._insert_virtual_scopes` |
| projects and users as inheriting members of their domain | 7 memberships, 7 bindings | `GroupDBSource.create`, `RBACWriteOps.create_full_user` |
| users as ordinary members of their projects, no binding back | 10 memberships | `add_bulk_members` |
| roles enrolled in their scope's virtual scope | 11 memberships | `V2EntityWriteOps._create_preset_roles` |

Virtual scope ids are derived as `uuid5` over the scope type and id, so they are reproducible and cross-referenceable between `example-users.json` and `example-roles.json`.

Also rekeys the two domain-scoped role associations in `example-roles.json` from the domain name (`"default"`) to the domain uuid, matching the 155 domain `permissions` rows in the same file.

## Verification

Ran `scripts/install-dev.sh` on a clean clone (ports 8301/8311/8321). All twelve fixtures populated without error; the chain holds 8 / 36 / 15 rows.

`POST /session` against the manager started from that install:

| chain | admin | domain-admin | user |
|---|---|---|---|
| emptied | `Invalid group (default)` | — | `Invalid group (default)` |
| populated | created | created | created |
| emptied again | `Invalid group (default)` | — | `Invalid group (default)` |

Sessions created carry `group_id` = the default project and reach `PREPARING`.

Creating a new user through `UserDBSource.create_user_validated` grants all three `auto_assign` roles; before the domain rekey, `role_domain_default_member` was not granted.

Two notes on scope. The membership check lives on `POST /session` (v1, `create_from_params` → `query_userinfo`); v2 `session enqueue` does not perform it. Reaching that check needs an agent serving resource slots, so a synthetic `agents` row was inserted into the throwaway database — the session therefore stops at `PREPARING` rather than running.

Backport: 26.8

Resolves BA-7401

### DIFF
--- a/changes/13835.fix.md
+++ b/changes/13835.fix.md
@@ -1,0 +1,1 @@
+Seed the virtual-scope chain in the example fixtures so session creation works on a fresh development install.
--- a/fixtures/manager/example-roles.json
+++ b/fixtures/manager/example-roles.json
@@ -6987,7 +6987,7 @@
         {
             "id": "0acb8b75-c8bb-4227-88a6-f35b784e37ee",
             "scope_type": "domain",
-            "scope_id": "default",
+            "scope_id": "b9e676dc-e23e-54c6-b350-e83fba4127e9",
             "entity_type": "role",
             "entity_id": "416fca68-f247-457b-ac83-32a0a45bb7d1",
             "registered_at": "2025-09-12 09:51:58.265582+00",
@@ -6996,7 +6996,7 @@
         {
             "id": "40bae437-988a-477e-89ce-d600842a127b",
             "scope_type": "domain",
-            "scope_id": "default",
+            "scope_id": "b9e676dc-e23e-54c6-b350-e83fba4127e9",
             "entity_type": "role",
             "entity_id": "bfb69730-1374-4ba4-bc26-0ea0b8dd2708",
             "registered_at": "2025-09-12 09:51:58.265582+00",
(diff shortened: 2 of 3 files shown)


fregataa:
## Verification on a clean database

Reproduced the fresh-install sequence on a scratch database — `mgr schema oneshot`, `alembic upgrade head`, then `mgr fixture populate` over all twelve fixtures in the order `scripts/install-dev.sh` uses — and ran the real read paths against it. The same run was done twice: once with the fixtures on `main`, once with this branch.

`mgr schema oneshot` reports `Detected a fresh new database` and stamps head, so the subsequent `alembic upgrade head` is a no-op. The backfill migration never executes on a fresh install at all.

### Chain contents

| | `main` | this branch |
|---|---|---|
| `virtual_scopes` | 0 | 8 |
| `entity_memberships` | 0 | 36 |
| `scope_bindings` | 0 | 15 |

Shape on this branch, matching what the runtime write paths produce:

```
 memberships: scope_type | entity_type | n     bindings: anchor | bound   | n
              domain     | domain      | 1               domain | domain  | 1
              domain     | project     | 2               project| domain  | 2
              domain     | role        | 2               project| project | 2
              domain     | user        | 5               user   | domain  | 5
              project    | project     | 2               user   | user    | 5
              project    | role        | 4
              project    | user        | 10
              user       | role        | 5
              user       | user        | 5
```

No user-virtual-scope-to-project binding, as intended.

### Read paths

**Session-creation group resolution** — the USER-path query from `query_userinfo` (`manager/utils.py:213`), which own-session creation always takes regardless of the requester's role:

| user | `main` | this branch |
|---|---|---|
| admin@lablup.com (superadmin) | no group | resolves |
| domain-admin@lablup.com | no group | resolves |
| user@lablup.com | no group | resolves |
| user2@lablup.com | no group | resolves |
| monitor@lablup.com | no group | resolves |

On `main` every user hits the `Invalid group (default)` path.

**auto_assign role lookup** (`_grant_auto_assign_roles`) — on `main` no role is reachable; on this branch all three resolve:

```
domain:b9e676dc…  -> role_domain_default_member
project:2de2b969… -> role_project_2de2b969_member
project:8e32dd28… -> role_project_8e32dd28_member
```

**Permission resolution** (`_resolve_permissions_for_virtual_scope_group`, the `entity -> entity_memberships -> scope_bindings -> permissions` walk) returns nothing at all on `main`. On this branch it resolves with sensible RBAC:

```
admin@lablup.com        self READ|UPDATE|SOFT_DELETE|HARD_DELETE, others same
domain-admin@lablup.com self and others READ|UPDATE|CREATE|SOFT_DELETE|HARD_DELETE
user@lablup.com         self READ|UPDATE|SOFT_DELETE|HARD_DELETE, others READ
monitor@lablup.com      self only
```

The project-scope leg was exercised separately with `vfolder` as the subject type — project- and domain-scoped roles do grant it — and reaches all five users. Resolving `project` as a subject type returns nothing on both branches simply because the fixtures contain no `permissions` row with `entity_type = 'project'`; that is unrelated to the chain.

### Not covered

- No full `scripts/install-dev.sh` run on a clean host — only its database provisioning sequence.
- Session creation was verified at the query level, not through a live API call against a running manager.
- `role_superadmin` and `role_monitor` remain enrolled in no scope. They carry no scope association in the fixtures to begin with and this branch does not change that; role enrollment only feeds the auto_assign lookup, so it does not affect their grants.

HyeockJinKim (approved):

PR #13842, opened 2026-08-20
fix(BA-7392): omit a wildcard HostIp from kernel port bindings

fregataa:
## Summary

- Every published kernel service port was unreachable on Podman: the host socket listened, the service answered on the container IP, but nothing was forwarded and connections piled up unaccepted in the backlog. No error surfaced anywhere — the container started, the ports appeared published, and the failure showed up only as client-side timeouts, so app proxying failed for every session on the node.
- The agent always sent an explicit `HostIp` in `HostConfig.PortBindings`, taken from `container.bind-host`. netavark reads `"0.0.0.0"` as a literal destination and emits a DNAT rule constrained with `-d 0.0.0.0/32`, which never matches real traffic. Docker's iptables backend accepts the same value, which is why this never showed up before.
- Leaving `container.bind-host` empty does not avoid it: the agent auto-fills an empty value from `agent.rpc-listen-addr.host`, which is `0.0.0.0` whenever the agent listens on all interfaces, as it does in a multi-node deployment. The value reaches the container config even though the operator never wrote it.
- A wildcard is now expressed by leaving `HostIp` out of the payload. A concrete address is still sent verbatim, so the loopback-only repl and protected service ports keep their binding.

Measured on the reporter's environment (Podman 4.9.3, rootful, netavark, Ubuntu 24.04):

| `HostIp` | via loopback | via node IP | generated DNAT rule |
|---|---|---|---|
| omitted | 200 | 200 | no `-d` constraint |
| `""` | 200 | 200 | no `-d` constraint |
| `"0.0.0.0"` | 000 | 000 | `-d 0.0.0.0/32 ...` |
| `"<node IP>"` | 000 | 200 | `-d <node IP>/32 ...` |

## Test plan

Unit — `tests/unit/agent/test_docker_agent.py::TestMakePortBinding`:

- [x] A wildcard host IP (`""`, `0.0.0.0`, `::`, `[::]`) drops `HostIp`
- [x] A concrete address (`127.0.0.1`, `10.0.1.5`, `::1`) is kept
- [x] The serialized payload uses the Docker API keys and omits `HostIp` only for a wildcard

Integration — live agent on the Docker backend (OrbStack, Docker Engine 29.4.0), varying `container.bind-host` and reading both `HostConfig.PortBindings` (what the agent sent) and `NetworkSettings.Ports` (what the engine published), then reading the sshd banner off the published port. Reading the banner rather than opening a bare TCP connection is deliberate: the failure leaves connections unaccepted in the backlog, where a connect alone can still look like success.

| `bind-host` | `HostIp` sent | via loopback | via node IP |
|---|---|---|---|
| `127.0.0.1` | `127.0.0.1` | OK | fails, as intended |
| `0.0.0.0` | omitted | OK | OK |
| `::` | omitted | OK | OK |
| `192.168.139.2` (concrete) | `192.168.139.2` | fails | OK ¹ |

¹ Probed from inside the engine host's network namespace — that address is an interface of the Docker VM, not of the macOS host. This reproduces the `"<node IP>"` row of the table above, confirming a concrete address keeps its binding semantics.

- [x] Reverting to the pre-fix payload on the same setup sends `HostIp: "0.0.0.0"` again, so the live path is what changed. Both are reachable under Docker, matching the 200/200 row above — the difference is the payload, and that is where netavark diverges.
- [x] The repl ports and the protected service ports stay bound to `127.0.0.1` in every case
- [ ] Verify on Podman/netavark — not reproducible on the Docker backend, so the final check belongs on the reporter's environment

## Notes

- Backport target not decided yet — this is a `fix:` PR, so without a `Backport:` trailer it would go to every maintained version (26.8, 26.4).

Resolves BA-7392

### DIFF
--- a/changes/.fix.md
+++ b/changes/.fix.md
@@ -1,0 +1,1 @@
+Omit `HostIp` from kernel port bindings when `container.bind-host` is a wildcard address, which netavark otherwise takes literally and turns into a DNAT rule that never matches, leaving every service port unreachable on Podman.
--- a/src/ai/backend/agent/docker/agent.py
+++ b/src/ai/backend/agent/docker/agent.py
@@ -191,6 +191,8 @@ _CGROUP_PATH_CACHE_SIZE: Final[int] = 2048
 
 # Docker splits the configured total container-log size across this many files.
 _CONTAINER_LOG_FILE_COUNT: Final[int] = 5
+# Addresses that mean "bind to every interface" rather than a concrete host address.
+_WILDCARD_BIND_HOSTS: Final[frozenset[str]] = frozenset({"", "0.0.0.0", "::", "[::]"})
 # Controllers the intrinsic plugins read, resolved together from a single inspect.
 _TRACKED_CGROUP_CONTROLLERS: Final[tuple[CgroupController, ...]] = (
     CgroupController.CPUACCT,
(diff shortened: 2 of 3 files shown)


fregataa:
Closed this PR since the issue is not reproduced

PR #13942, opened 2026-08-24
docs(BA-7466): add BEP-1073 entity labels

fregataa:
## Summary

- Add BEP-1073, a design for generalized `key=value` labels that can be put on any entity, searched on their own, and used as a filter over the entities that carry them.
- Treats a label as a field of the entity it is attached to: one table holding `(entity, key, value)`, a polymorphic entity reference so no labelable table gains a column, and no owner dimension of its own.
- Permission checks route through that entity, so labels add no permission surface — a label is readable exactly when its entity is, which is also what bounds label search.
- Reuses existing conventions rather than inventing: the `some`/`every`/`none` nested filter from BEP-1060, and the filter and page/order shape from `audit_log`.

Resolves BA-7466

### DIFF
--- a/changes/13942.doc.md
+++ b/changes/13942.doc.md
@@ -1,0 +1,1 @@
+Add BEP-1073, a design for generalized key=value entity labels with label search and label-based entity filtering.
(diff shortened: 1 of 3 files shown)


HyeockJinKim (approved):
LGTM

PR #13948, opened 2026-08-24
refactor(BA-7462): relocate the container registry and image write specs to the v2 bases

HyeockJinKim:
## Summary

- The eight write specs of the container registry and image domains move to `models/` beside their rows and onto the `models/specs` lineage — the registry as a `GlobalEntityCreator` / `EntityPurger` / `DataUpdater`, its project associations as a `FieldCreator` / `FieldBatchPurger`, the image as an `EntityCreator` and its alias as a `FieldCreator`. The legacy `ScopeCreation` adapter goes away: `create_global_entity` provisions the registry's virtual scope itself.
- The registry-project reach edges are a primitive only this domain has, so they sit on `ContainerRegistryWriteOps` behind `ContainerRegistryOpsProvider`. `ContainerRegistryRepository` and `ImageRepository` take a provider rather than reaching the legacy `RBACOpsProvider`.
- An image update names its row by an id resolved from a canonical name or an alias, which is not known where the values are built, so the values are carried apart from the updater as `ImageUpdate`.

## Where the target could not be met

| Item | Why |
|---|---|
| `repositories.base` imports are not zero in the two directories | `BatchQuerier` / `execute_batch_querier` remain. Those are the legacy read path; this change covers the write specs, and moving the search axis is separate work. |

## Behavior changes to know about

| Change | Detail |
|---|---|
| Legacy scope associations are no longer written | The v2 create paths record virtual-scope membership instead of `association_scopes_entities`. Deleting a registry created before this change leaves that table's older rows behind. |
| A rescan writes in two transactions | Updating the images that already exist and inserting the new ones no longer share one transaction. The two touch disjoint rows and a rescan is re-runnable. |
| A customized image needs its owner's virtual scope | Membership resolves-or-fails, so an owner without a virtual scope fails the whole insert batch, where the legacy association was written unvalidated. The registry itself is provisioned first. |

## Test plan

- [x] `pants fmt` / `pants fix` / `pants lint`
- [x] `pants check` — no issues across 3691 source files
- [x] `pants test --changed-since --changed-dependents=direct` — passing; `tests/component/image/test_image_query.py` and `tests/unit/manager/repositories/container_registry/test_container_registry_repository.py` re-confirmed after raising the local process timeout (the former runs ~230s here)
- [ ] CI runs the component targets that exceed the local timeout

Resolves BA-7462


src/ai/backend/manager/repositories/ops/v2/container_registry/write.py
```diff
@@ -0,0 +1,72 @@
...
+                ScopeBindingRow.virtual_scope_id.in_(
+                    sa.select(VirtualScopeRow.id).where(
+                        VirtualScopeRow.scope_type == registry_id.entity_type(),
+                        VirtualScopeRow.scope_id == registry_id,
+                    )
+                ),
+                ScopeBindingRow.scope_type == project_id.entity_type(),
+                ScopeBindingRow.scope_id == project_id,
+            )
+        )
```
fregataa, line 72: This function should be implemented in a general writer ops, like `RBACWriteOps.remove_bulk_members()` method.

fregataa (approved):

PR #13950, opened 2026-08-24
refactor(BA-7464): move the user and project purge specs to the v2 lineage

HyeockJinKim:
## Summary

- Move twelve of the user and project domains' legacy specs from
  `repositories/base` to the `models/specs` v2 lineage, declared next to their
  rows in `models/{user,project}/purgers.py`.
- Switch `purge_user_by_uuid` and `purge_group` from the RBAC ops provider to
  `V2DBOpsProvider`.
- Model sessions, session groups and deployments as entity batch purges, so each
  deleted row's RBAC graph goes with it.

## What is left on the legacy path, and why

- **`_UserRowInsert` / `UserScopeCreation`** stay. A user's owner role over its
  own scope comes from `UserSystemRoleSpec`; the v2 role-managed create
  provisions roles from role presets only, and no migration seeds a `user` scope
  preset. Moving the creation now would leave every new user without permissions
  on its own resources. `create_full_user` also writes the row, the default
  keypair and the domain/project enrollments in one transaction, which the v2 ops
  have no primitive for.
- **The membership paths** (`update_members`, `assign_users_to_project`,
  `unassign_users_from_project`, `_sync_user_project_memberships`) stay on
  `RBACOpsProvider`: writing an association row has no v2 primitive.
- **Two transaction boundaries split.** `purge_group` reads its preconditions
  (vfolder mounts, routed sessions, purgable vfolders) in a read transaction and
  performs every write in one v2 transaction.
  `_sync_user_project_memberships` removes memberships on the legacy ops and
  purges the left projects' role mappings in a v2 transaction of its own.

## Test plan

- [ ] `pants lint --changed-since=origin/main`
- [ ] `pants check --changed-since=origin/main`
- [ ] `pants test --changed-since=origin/main --changed-dependents=direct`
- [ ] Purge a user and confirm its error logs, keypairs, vfolder permissions,
      group associations, placement groups and scope associations are gone
- [ ] Purge a project with deployments and sessions and confirm the routed
      sessions, kernels and scope associations are gone

Resolves BA-7464


src/ai/backend/manager/models/user/purgers.py
```diff
@@ -0,0 +1,259 @@
...
+    def conflict_checks(self) -> Sequence[ConflictCheck]:
+        return ()
+
+    @override
+    def to_data(self, row: AssociationScopesEntitiesRow) -> UUID:
+        return row.id
+
+
+@dataclass
+class UserProjectRolePurger(FieldBatchPurger[UserID, UserRoleRow, RoleID]):
```
fregataa, line 150: This should remove all role assignment data rather than purging UserProjectRole

HyeockJinKim: I plan to remove the parts of the existing implementation that handle RBAC all at once later. (For now, I’m focusing on migrating the code, and since these are field values, I don’t think they’ll have much impact.)
The permissions assigned to a user will be removed along with the user entity via a teardown by EntityPurger when the user entity is deleted anyway.

fregataa (approved):

PR #13946, opened 2026-08-24
feat(BA-7474): add the entity label storage and repository ops

fregataa:
## Summary

- Adds the `labels` table: one `key=value` per row, the labeled entity held as a polymorphic `(entity_type, entity_id)` pair with no foreign key, so a label goes on any type.
- Adds the field-lineage specs — create, purge, search, owner lookup — plus the `some`/`every`/`none` correlated `EXISTS` builder. Wiring it into each entity's filter is not part of this PR.
- `FieldOwnerLookup` now selects the owner's type beside its id, so a polymorphic owner can be rebuilt for the RBAC check. `dangling_lookup_field_group` is the entry point for a field kind whose owner type is not fixed.
- Purge propagates through `_teardown_entity`, which already keys on the same pair.
- Corrects the field-row criterion in `models/specs/KNOWLEDGE.md`, which named a cascading FK as part of the test.

Design: `proposals/BEP-1073-entity-labels.md`.

## Test plan

- [x] `alembic upgrade head` creates `labels` with `uq_labels_label`, `ix_labels_entity`, `ix_labels_pair`; `downgrade` reverses it

Resolves BA-7474

PR #13954, opened 2026-08-24
feat(BA-7475): expose entity labels through REST v2, GraphQL, the SDK, and the CLI

fregataa:
## Summary

- Adds the label operations to REST v2: `PUT /v2/entity-labels` sets a key on an entity, `DELETE /v2/entity-labels/{id}` takes one off, `POST /v2/entity-labels/search` reads a page. One surface, gated by the labeled entity's own permission rather than by an admin flag.
- Adds `labels` to the filter of session, deployment, vfolder, and agent, with the `some` / `every` / `none` relations of BEP-1060. Each relation is one correlated `EXISTS`, so a relation's key and value narrow the same label.
- Adds the GraphQL inputs `EntityLabelFilter` and `EntityLabelNestedFilter`, and regenerates the supergraph.
- Adds the SDK client and `./bai entity-label {put,remove,search}`, plus `--label KEY[=VALUE]` on every search of a labelable entity.

Design: `proposals/BEP-1073-entity-labels.md`. Depends on #13946.

## Test plan

- [x] `./bai entity-label put` twice on one key replaces the value in place and moves `updated_at`
- [x] `--label gpu=reserved` selects the labeled agent, `--label gpu=nope` selects none, two `--label`s require both
- [x] A non-admin is refused `put` and `remove` on an entity they cannot write, and a search naming an entity they cannot read is refused rather than answered with an empty page
- [x] The four filters carry `labels: EntityLabelNestedFilter` in the regenerated supergraph

Resolves BA-7475

----
📚 Documentation preview 📚: https://sorna--13954.org.readthedocs.build/en/13954/

----
📚 Documentation preview 📚: https://sorna-ko--13954.org.readthedocs.build/ko/13954/

### DIFF
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -417,6 +417,9 @@ input AgentFilter
   status: AgentStatusFilter = null
   schedulable: Boolean = null
   scalingGroup: StringFilter = null
+
+  """Added in 26.9.0. Select entities by the labels on them."""
+  labels: EntityLabelNestedFilter = null
   AND: [AgentFilter!] = null
   OR: [AgentFilter!] = null
   NOT: [AgentFilter!] = null
@@ -6164,6 +6167,9 @@ input DeploymentFilter
 
   """Added in 26.8.0. Filter by conditions on deployment replicas."""
   replicas: ReplicaNestedFilter = null
+
+  """Added in 26.9.0. Select entities by the labels on them."""
+  labels: EntityLabelNestedFilter = null
   AND: [DeploymentFilter!] = null
   OR: [DeploymentFilter!] = null
   NOT: [DeploymentFilter!] = null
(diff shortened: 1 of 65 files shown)


docs/manager/graphql-reference/v2-schema.graphql
HyeockJinKim, line 1: Why isn't there a "mutation" option?

src/ai/backend/manager/api/gql/session/types.py
```diff
@@ -183,6 +184,14 @@ class SessionV2FilterGQL(PydanticInputMixin[SessionFilter]):
...
     user_uuid: UUIDFilter | None = None
 
+    labels: EntityLabelNestedFilterGQL | None = gql_added_field(
+        BackendAIGQLMeta(
+            added_version=NEXT_RELEASE_VERSION,
+            description="Select entities by the labels on them.",
+        ),
+        default=None,
+    )
+
```
HyeockJinKim, line 202: It's not enough to just add a filter; a labels resolver based on `async def` must also be attached for each type.

HyeockJinKim (commented):
It looks like there's a missing implementation of "labels" in the GQL side
please check it and update it.

docs/manager/graphql-reference/supergraph.graphql
```diff
@@ -7649,6 +7663,101 @@ input EntityFilter
...
+  key: String!
+
+  """Label value."""
+  value: String!
+
+  """When the label was first put on the entity."""
+  createdAt: DateTime!
+
+  """When the label's value was last replaced."""
+  updatedAt: DateTime!
```
HyeockJinKim, line 7841: I think this was optional in the database...

HyeockJinKim (dismissed):

docs/manager/graphql-reference/supergraph.graphql
HyeockJinKim: I was mistaken about that. I ended up doing a bulk replacement later.

src/ai/backend/client/cli/v2/helpers.py
```diff
@@ -219,3 +221,69 @@ def print_result(data: Any) -> None:
...
+        """
+        merged = base.model_copy(
+            update={"AND": [*(getattr(base, "AND", None) or []), *self._as_filters(type(base))]}
+            if self.terms
+            else {}
+        )
+        return merged if merged.model_dump(exclude_none=True) else None
+
+    def _as_filters(self, filter_cls: type[Any]) -> list[Any]:
+        return [filter_cls(labels=term) for term in self.terms]
```
HyeockJinKim: This implementation seems a bit hacky—what is this?

HyeockJinKim (approved):
Let's just get through this for now and sort it out later.

PR #13957, opened 2026-08-24
feat(BA-7472): generalize the RBAC invitation to carry an entity grant

HyeockJinKim:
## Summary

- Adds `entity_invitations`: an offer of one existing entity to one email address, settled by the invitee's answer. The target is a polymorphic `(target_entity_type, target_entity_id)` pair, so any entity type is offerable without the schema knowing it, and `permission_cap` bounds what acceptance grants.
- Accepting settles the invitation and grants its entity in one write. The grant **widens** rather than states: an offer is somebody else's, and taking it must not cost the invitee access they already had. Setting a cap exactly stays on `grant_entities`, which an invitation never reaches.
- The invitee is an email, not a user id, so an invitation may name someone with no account yet. Their two answers are authorized by a guard matching that address, carried in the statement so the read and the write cannot part ways.
- One scoped search over the sides a read comes in through — addressed to me, sent by me, offering one entity — each declaring the scope it is answered for beside the rows it selects.
- Reachable over REST v2, GraphQL, the SDK and the CLI, all through one adapter.

Scope is the RBAC-side generalization. Replacing the vfolder invitation with it is BA-7473.

## Known gaps

- **No permission definition yet.** Nothing grants `entity_invitation` in the RBAC graph, so only a superadmin passes today. That belongs to the separate role-preset work.
- **Purging a target leaves its invitations.** `target_entity_id` is polymorphic and carries no foreign key, so nothing cascades; the rows go when the retention sweep reaches them.
- **BA-7473 inherits** the vfolder cutover, the issuing-side permissions per target type, and replacing `VFolderGrantAlreadyExists` with a read of the sharing list — an invitation by email cannot reliably answer "do they already have it", and a failed re-invite is the wrong place to learn it.

## Test plan

- [x] `pants fmt` / `fix` / `lint` / `check` over the changed set
- [x] 13 real-DB tests covering what the guards decide and what acceptance writes into the graph (widening, an absent ceiling staying absent, somebody else's invitation, an already-answered one, the open-offer conflict, re-inviting after a rejection)
- [x] `pants test --changed-dependents=direct` — 38 targets, no failures
- [ ] Exercise the CLI against a live server once a role preset grants `entity_invitation`

Resolves BA-7472

----
📚 Documentation preview 📚: https://sorna--13957.org.readthedocs.build/en/13957/

----
📚 Documentation preview 📚: https://sorna-ko--13957.org.readthedocs.build/ko/13957/


src/ai/backend/manager/api/gql/entity_invitation/resolver.py
```diff
@@ -0,0 +1,152 @@
...
+    EntityInvitationEdge,
+    EntityInvitationFilterGQL,
+    EntityInvitationGQL,
+    EntityInvitationOrderByGQL,
+    EntityInvitationPayloadGQL,
+    EntityInvitationScopeGQL,
+)
+from ai.backend.manager.api.gql.types import StrawberryGQLContext
+
+_ADDED = "26.8.3"
```
fregataa: Will this be backported or this code just adds version notations?

HyeockJinKim: No. I haven't seen this one. It should be NEXT_RELEASE_VERSION.

fregataa (approved):

PR #13961, opened 2026-08-25
feat(BA-7481): name an entity type by what the wiring registers

fregataa:
## Summary

- An entity reference at the API boundary is typed by `RBACElementTypeDTO`, a closed enum. An entity type is not an enum: it is an `EntityType` declared where its operations are wired, so a new one arrives without the enum knowing, and one the enum spells differently — `model_deployment` against the `deployment` every row, filter and RBAC action carries — names something no read finds.
- `WiredEntityTypes` reads those names once from the processor registry the wiring records and refuses one it does not hold. `entityTypes` returns them, so a caller reads what it may name rather than guessing. `EntityTarget` is the shape that names an entity by one of them together with a UUID.
- The registry reaches the adapters through the processor bundle the dependency stage already builds, so the `Processors` contract stays as it was.

## Notes

`input EntityTarget` is not in the SDL yet: Strawberry emits an input type only once a field takes it, and the first consumer is the entity-label target in BA-7475, which lands on top of this.

## Test plan

- [x] `entityTypes` returns 46 names against a running manager — `agent`, `deployment`, `session`, `vfolder` present, `global` excluded, `model_deployment` absent
- [x] `WiredEntityTypes.resolve("deployment")` resolves; `resolve("model_deployment")` raises `InvalidAPIParameters`
- [x] `pants check --changed-since=HEAD~1` (mypy, 19 files)
- [x] CI

Resolves BA-7481

----
📚 Documentation preview 📚: https://sorna--13961.org.readthedocs.build/en/13961/

----
📚 Documentation preview 📚: https://sorna-ko--13961.org.readthedocs.build/ko/13961/


HyeockJinKim (commented):
you've only added gql.
please add REST, the client SDK, and the CLI as well.

HyeockJinKim (approved):

PR #13970, opened 2026-08-25
misc: resolve installer client env configs by keypair user id

devyubin:
### Summary
The TUI installer aborted partway through with `KeyError: 'user_id'`, and reported success anyway.

### Changes

**Client environ configs are generated again**
- `configure_client` now maps each keypair's `user` UUID through the users fixture to get the account name, instead of reading a `user_id` field and parsing the name out of the email.
- Reason: the installer's fixtures are symlinks to `fixtures/manager/`, so the manager's schema changes reach it directly. `example-keypairs.json` carried both `user` (UUID) and a redundant `user_id` (email) until bd94f48db dropped the duplicate, and the installer was the only remaining reader of it. The install died before Harbor setup, image population, and `INSTALL-INFO`, leaving a half-configured tree.
- Reading `user` as if it were the email is not a fix: it is a UUID, so the old `^(\w+)@` match fails for every row and the step silently writes zero files instead of raising.
- Side effect: `domain-admin` gets its env files now. `\w` excludes the hyphen, so the email regex skipped that account.

**A failed install exits non-zero**
- Install failures are recorded on the app, returned from `action_shutdown`, and propagated by the CLI via `app.return_code`.
- Reason: both setup widgets log the exception and fall through to the normal quit path, and `main()` discarded `app.run()`'s result, so every install exited 0. Our automation recorded this broken run as COMPLETED, which is how the `KeyError` went unnoticed until someone opened the install log.
- Package mode also ran its target-path checks outside the `try` block, where a failure disappeared with the task and never reached the log. Those checks moved inside.

**Regression coverage**
- `tests/unit/install/test_client_env.py` runs `configure_client` against the real fixtures and asserts the env files it writes.
- Reason: asserting that specific keys exist would not have caught this, since the breaking change removed a duplicate field rather than renaming one. Driving the actual function fails on both the original `KeyError` and the zero-files variant.


fregataa (commented):
Please avoid using "fix" for a PR title prefix or set backport version to null in the PR description

seedspirit (approved):

PR #13973, opened 2026-08-25
feat(BA-7483): declare the scope types a scope action accepts

fregataa:
## Summary

- `BaseScopeAction` gains `available_scope_types()`, declared beside the run-time `scope_targets()` it bounds. Every wired scope action implements it — 82 classes across 68 declaring bases. `global` stands for a scope type the caller names, which no fixed set bounds; the empty sequence is an action that names no scope at all.
- The declaration is served from what already lists the wiring, rather than a new endpoint:
  - **`backend.ai mgr ops list`** gains a `scope_types` column, one value per operation. `-` where the shape is not scope, `(none)` where it is and names no scope at all.
  - **`entityTypes` / `GET /v2/entity-types`** gain `scope_types`, the union of what that entity type's scope operations declare, in name order.
- `VALID_SCOPE_ENTITY_COMBINATIONS` and the `rbacScopeEntityCombinations` resolver are untouched. Replacing them is a follow-up once a caller reads the new field.

Two of the 82 declare `global`: an entity invitation names its target by a caller-supplied entity type, so no fixed set bounds it. One declares the empty sequence: `anonymous_search_app_configs` targets no scope. The other 79 declare a fixed scope type or the pair an input branches over.

A read whose wiring fixes no owner is recorded under `global`, which the entity type listing leaves out by definition. That is the three dangling usage-bucket reads (`search_{domain,project,user}_usage_buckets`, all declaring `resource_group`); they are named by `mgr ops list` alone. Folding them into their `entity_type()` would claim that a permission on `domain` may be configured at a resource-group scope, when what the operation reads is bucket rows.

## Test plan

- [x] `pants fmt / fix / lint / check` pass (mypy over 5297 source files including direct dependents).
- [x] `test_scope_actions_declare_the_scope_types_they_accept` — no scope type listed twice, and `global` is never listed beside a concrete one.
- [x] `tests/unit/manager/cli/test_ops.py` — only a scope-shaped operation declares scope types, and a declared set is never `global` plus something else.
- [x] `tests/unit/manager/api/adapters/test_entity_adapter.py` — an entity type carries what its scope operations declare, merged in name order; one nothing scopes carries none; an operation the wiring gives no entity type is left out.
- [x] Live manager: `backend.ai mgr ops list` shows the column, with `-` / `(none)` / `global` / `resource_group` all rendering as intended.
- [x] Live manager: `./bai entity-type list` and `entityTypes` over GraphQL both return 48 entity types, 21 of them carrying scope types; `global` is not among them. An unauthenticated `GET /v2/entity-types` is 401.

Resolves BA-7483

----
📚 Documentation preview 📚: https://sorna--13973.org.readthedocs.build/en/13973/

----
📚 Documentation preview 📚: https://sorna-ko--13973.org.readthedocs.build/ko/13973/


HyeockJinKim (changes requested):
Instead of adding a new API, wouldn't it be better to just include it as part of the existing API?

HyeockJinKim (changes requested):
The request was to display "scope action" alongside the "ops list" rather than showing it separately.

docs/manager/graphql-reference/supergraph.graphql
```diff
@@ -8028,6 +8028,11 @@ type EntityType implements Node
 
   """The entity type, as a request names it."""
   name: String!
+
+  """
+  Added in 26.9.0. The scope types an operation on this entity type may be targeted at; `global` where the caller names the scope type, empty where no operation on it names a scope.
+  """
+  scopeTypes: [String!]!
```
HyeockJinKim, line 8035: I don't quite understand—what were you trying to say?

PR #13974, opened 2026-08-25
doc(BA-7486): record the v2 action wiring audit as a summary and a per-concern table

fregataa:
## Summary

- Adds `docs/reports/v2-action-audit/`: a `README.md` summarising the audit and `01-wiring.md` holding the static wiring investigation. The audit is split into three — the wiring read, a runtime test, and a permission-logic review — and the summary names all three so the later two land beside this one.
- `01-wiring.md` carries every wiring the processor assembly makes, one row per catalog row, split into failed/clean tables and sectioned by concern. Columns are those of `backend.ai mgr ops list` minus `concern` (which heads each section), plus a verdict and a reason.
- 653 rows — the 652 catalog rows plus `import_artifact_batch`, which is defined but never wired. 66 rows carry 69 findings under four verdicts: `W` dead wiring 28, `O` wrong operation declared 21, `R` record defect 19, `G` gate justification 1. Each verdict is counted per sub-case as well.
- Every count and cited line was re-derived against the tree at `b2b200c0da` by importing all of `ai.backend.manager`, walking `__subclasses__()` from the nine v2 bases, and diffing that against `load_wiring_catalog()`. One correction came out of it: the claim that only the GraphiQL page and the public GraphQL subgraph carry no auth middleware is wrong — `/auth/authorize`, `/auth/signup`, `/auth/update-password-no-auth`, `/app-config/public/get` and `/container-registries/webhook/harbor` carry none either. All of them serve anonymous-gated actions, so no permission-gated action sits on an unauthenticated route.

Documentation only — no code, no behavior change. The fixes are filed separately; this records the audit.

## Test plan

- [ ] The per-concern totals add up to 653 rows (66 failed / 587 clean)
- [ ] The sub-case counts sum to the verdict totals, and those sum to 69
- [ ] Every table row renders with all nine columns, and both internal links resolve
- [ ] The row set matches `backend.ai mgr ops list` at `b2b200c0da`

Resolves BA-7486

----
📚 Documentation preview 📚: https://sorna--13974.org.readthedocs.build/en/13974/

----
📚 Documentation preview 📚: https://sorna-ko--13974.org.readthedocs.build/ko/13974/


HyeockJinKim (changes requested):
To prevent incorrect merging.

PR #13988, opened 2026-08-25
refactor(BA-7498): move the permissions table off the legacy type enums

fregataa:
## Summary

- `permissions.entity_type` and `permissions.scope_type` move from `StrEnumType` over the closed `common.data.permission.types` enums to `sa.String(32)` with the open `common.data.entity.types` ones, matching how `entity_memberships` and `scope_bindings` already declare theirs. This closes the seam the virtual-scope chain spanned: the chain query bridged it by binding the open string as a plain VARCHAR, which was fail-closed — an entity type absent from the legacy enum matched no permission row — and writing was blocked from the other side, since the column bound through `value.value`. That bridge is gone.
- **No alembic migration.** `StrEnumType.impl` is `sa.VARCHAR`, so the DDL was already `VARCHAR(32)`; the live columns are `character varying(32)` with no CHECK constraints, and the original create used `sa.VARCHAR(length=32)`. Stored values were already plain strings, so no data migration either. Alembic autogenerate against a live database reports 0 diffs both before and after the change.
- `PermissionCreatorSpec` / `PermissionUpdaterSpec` carry the open types, dropping the `to_entity_type()` / `to_scope_type()` bridge from the permission write path; callers convert at their own boundary. `PermissionDTO.entity_type` widens with them, since a response holding an out-of-enum value would otherwise fail serialization.

### Scope decisions

- **`association_scopes_entities` stays on the legacy enums.** It is the association table the virtual-scope chain replaces rather than a member of it — the chain query joins `entity_memberships`, `scope_bindings`, `permissions`, `roles`, `user_roles` only — and its whole read/write surface is keyed on the closed `RBACElementType`, so widening the column alone would change no behavior.
- **Request DTOs keep their closed enums.** Responses widen because reads would otherwise break, but `CreatePermissionRequest` and the GraphQL `PermissionNode` keep validating against their enum and the adapters convert. Widening those would drop REST input validation and change the GraphQL supergraph.
- **Known remaining limit:** the preset and system-role seeding path is still keyed on `RBACElementType` and `RolePermissionPresetRow.entity_type`, both still closed. The `permissions` table itself is open now, but grants for entity types missing from those two vocabularies still cannot be authored. Confirming that every affected operation resolves a real permission needs that path opened as well.

## Test plan

- [x] `pants fmt` / `fix` / `lint` / `check` pass on the changed files
- [x] All 30 RBAC unit targets pass, including the virtual-scope chain suites (`test_check_permission_via_virtual_scope`, `test_resolve_effective_permissions`, `test_check_bulk_permission_with_scope_chain`)
- [x] `tests/component/rbac/test_rbac_permission.py` passes
- [x] Regression tests added over `role_preset`, a wired entity type absent from the legacy enum: `test_grant_over_unmapped_entity_type_resolves` authors a grant on it and resolves it through the virtual-scope chain, and `test_stored_unmapped_entity_type_reads_back` stores it directly and reads the row back through the ORM. Verified by A/B — reverting `src/` to `main` and rerunning the same two tests gives `2 failed`, with `AttributeError: 'EntityType' object has no attribute 'value'` on the write and `ValueError: 'role_preset' is not a valid EntityType` on the read, both at `models/base.py:327`; on this branch they give `2 passed`.
- [x] Verify the RBAC admin surfaces against a running server, admin and non-admin

Resolves BA-7498

https://claude.ai/code/session_01FBFw11xQvmhWuP2qfrkmjU

### DIFF
--- a/src/ai/backend/manager/models/rbac_models/permission/permission.py
+++ b/src/ai/backend/manager/models/rbac_models/permission/permission.py
@@ -7,7 +7,7 @@ import sqlalchemy as sa
 from sqlalchemy.orm import Mapped, mapped_column
 
 from ai.backend.common.data.entity.role import RoleID
-from ai.backend.common.data.entity.types import EntityType, ScopeType
+from ai.backend.common.data.entity.types import EntityType
 from ai.backend.manager.data.permission.permission import PermissionCreator, PermissionData
 from ai.backend.manager.data.permission.types import (
     OperationType,
@@ -52,7 +52,7 @@ class PermissionRow(CreatedAtMixin, Base):
         sa.ForeignKey("roles.id", ondelete="CASCADE"),
         nullable=False,
     )
-    scope_type: Mapped[ScopeType] = mapped_column(
+    scope_type: Mapped[EntityType] = mapped_column(
         "scope_type", sa.String(length=32), nullable=False
     )
     scope_id: Mapped[str] = mapped_column("scope_id", sa.String(64), nullable=False)
--- a/src/ai/backend/manager/models/rbac_models/permission/permission.py
+++ b/src/ai/backend/manager/models/rbac_models/permission/permission.py
@@ -7,12 +7,11 @@ import sqlalchemy as sa
 from sqlalchemy.orm import Mapped, mapped_column
 
 from ai.backend.common.data.entity.role import RoleID
+from ai.backend.common.data.entity.types import EntityType, ScopeType
 from ai.backend.manager.data.permission.permission import PermissionCreator, PermissionData
 from ai.backend.manager.data.permission.types import (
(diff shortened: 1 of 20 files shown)


src/ai/backend/manager/models/rbac_models/permission/permission.py
```diff
@@ -54,11 +53,11 @@ class PermissionRow(CreatedAtMixin, Base):
         nullable=False,
     )
     scope_type: Mapped[ScopeType] = mapped_column(
```
HyeockJinKim: Could you change this to an EntityType as well? We no longer need the Scope Type.

HyeockJinKim (approved):

PR #14008, opened 2026-08-26
fix(BA-7397): release glide client resources when creation fails

fregataa:
## Summary

- `GlideClient.create()` returns the instance only on success. When it raises at `_set_connection_configurations()` — the failure mode seen in production — the UDS stream, the reader task, and the glide-core client it already built are left unreachable to the caller, so nothing can close them. The reader task keeps a strong reference to the abandoned instance, so it is never garbage collected either.
- Each failed reconnect therefore leaked 2 file descriptors and 1 asyncio task, and glide-core kept its Valkey connection open because the UDS stream was never closed. On the server this piled up as thousands of `cmd=NULL` entries in `CLIENT LIST`.
- `create()` exposes no handle to the half-built instance, so this captures it through a `GlideClient` subclass — `create()` calls `cls(config)` synchronously before its first `await`, so a ContextVar slot set by the caller is safe against interleaving. On failure the captured instance is closed, which shuts the UDS stream, terminates the reader loop, and lets glide-core drop its connection.

This is an upstream defect. valkey-glide 2.5.0 removed the UDS transport entirely (valkey-io/valkey-glide#5637), so the workaround should be dropped once the pin moves to 2.5.2 or later — the docstring says so. 2.5.2 is not yet published as a stable release, and 2.5.1 still carries a per-command payload leak (valkey-io/valkey-glide#6740), so 2.3.1 remains vulnerable in the meantime.

## Test plan

Verified against the local halfstack Valkey with `ClosingError` injected at `_set_connection_configurations()` — the exact failure point in the reported traceback — over 20 failed creates:

| | `close()` calls | leaked fds | leaked asyncio tasks |
|---|---|---|---|
| before (`GlideClient.create`) | 0/20 | 40 | 20 |
| after (`_create_glide_client`) | 20/20 | 0 | 0 |

- [x] Failed creation releases the UDS stream and reader task
- [x] Success path unchanged (identical fd/task counts before and after)
- [x] `pants fmt` / `fix` / `lint` / `check` pass

Resolves BA-7397

### DIFF
--- a/changes/14008.fix.md
+++ b/changes/14008.fix.md
@@ -1,0 +1,1 @@
+Release glide client resources when `GlideClient.create()` fails, so failed Valkey reconnects no longer leak sockets and connections.
--- a/src/ai/backend/common/clients/valkey_client/client.py
+++ b/src/ai/backend/common/clients/valkey_client/client.py
@@ -3,6 +3,7 @@ import logging
 from abc import ABC, abstractmethod
 from collections.abc import AsyncIterator, Iterable
 from contextlib import AbstractAsyncContextManager, asynccontextmanager
+from contextvars import ContextVar
 from dataclasses import dataclass
 from typing import Any, Final, Self, override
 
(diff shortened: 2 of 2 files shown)


HyeockJinKim (approved):
Since this code is a temporary implementation, I'll go ahead and merge it for now.

PR #14010, opened 2026-08-26
feat(BA-7504): add the abstract authentication plugin and call it from authorize

fregataa:
## Summary

- Adds `AbstractAuthPlugin` and the `backendai_auth_v1` entry-point group, so an authentication integration declares one contract instead of reaching into the manager's internals. At most one plugin may be loaded; the conflict is caught at discovery, before `init()`, and the manager refuses to start naming both plugins and the setting that disables one.
- Calls the plugin from `authorize`: it names the account a request belongs to, and the manager performs the lookup, the status checks and everything downstream. Answering with nothing means the request carries no credential this plugin handles, so the existing hook and password paths run unchanged.
- On a failed lookup the plugin's callback decides what happens next — returning lets the manager retry up to `lookup_retry_count`, raising aborts at once. An exhausted lookup fails with the generic credential error and never names the account.
- Removes SQLAlchemy row mappings from the authorize flow, which now carries `UserData` end to end; the legacy `AUTHORIZE` hook's row is converted at the boundary. The request reaches the plugin as `HTTPRequestData`, built in the handler and carried on the action, with headers and query parameters kept in multi-dict form so repeated keys and case-insensitive header lookup survive.

## Test plan

- [x] `pants lint check ::` and `pants test ::` in CI
- [x] Unit tests cover the single-plugin rule, the retry-count bound, the lifecycle defaults, and lookup data that names no account

Resolves BA-7504

https://claude.ai/code/session_01RJSXWXGpBjxa7dE6Y7Q1ox


src/ai/backend/manager/data/auth/lookup.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+
+    user_id: UserID | None = None
+    email: str | None = None
+    username: str | None = None
+    access_key: AccessKey | None = None
+
+    def __post_init__(self) -> None:
+        if not any((self.user_id, self.email, self.username, self.access_key)):
+            raise InvalidUserLookupData("The lookup data names no account.")
```
HyeockJinKim: When we talked yesterday, I did suggest managing the values in the `data` package, but on second thought, since the `data` package is likely to be affected by changes, should we manage them in a different package, such as `dto` or `schema` instead?

HyeockJinKim: Since it's not a storage operation, the schema is a bit ambiguous.

src/ai/backend/manager/plugin/auth.py
```diff
@@ -0,0 +1,99 @@
...
+    async def init(self, context: Any | None = None) -> None:
+        pass
+
+    @override
+    async def cleanup(self) -> None:
+        pass
+
+    @override
+    async def update_plugin_config(self, plugin_config: Mapping[str, Any]) -> None:
+        self.plugin_config = plugin_config
```
HyeockJinKim: There seem to be a lot of plugin configurations now—are all of these really necessary? I think it should be possible to override the "lookup retry count" in the plugin, but I don't quite understand the "init" and "plugin config" parts.

src/ai/backend/manager/config/unified.py
```diff
@@ -1995,6 +1997,18 @@ class PluginsConfig(BaseConfigSchema):
             added_version="25.8.0",
         ),
     ]
+    auth: Annotated[
+        dict[PluginName, AuthPluginConfig],
```
HyeockJinKim, line 2001: Mapping

src/ai/backend/manager/plugins/auth.py
```diff
@@ -0,0 +1,58 @@
...
+
+        The manager resolves the returned key without verifying the credential itself.
+        Return ``None`` when the request carries no credential this plugin handles.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    async def on_user_lookup_success(self, user: UserData) -> None:
+        """Called once the manager has resolved the account the lookup data named."""
+        raise NotImplementedError
```
HyeockJinKim, line 49: It still says "data class," doesn't it?

src/ai/backend/manager/models/user/row.py
```diff
@@ -31,6 +31,7 @@
 from ai.backend.common.types import ReadableCIDR
 from ai.backend.logging import BraceStyleAdapter
 from ai.backend.manager.data.auth.hash import PasswordHashAlgorithm
+from ai.backend.manager.data.auth.types import UserData as AuthUserData
```
HyeockJinKim, line 34: ??

src/ai/backend/manager/plugins/plugins.py
```diff
@@ -0,0 +1,12 @@
...
+from dataclasses import dataclass
+
+from ai.backend.manager.plugins.auth import AuthPlugin
+
+
+@dataclass
+class ManagerPlugins:
+    """The plugins the manager loaded at startup, registered once and injected from here."""
+
+    auth_plugin: AuthPlugin | None
```
HyeockJinKim, line 12: Please set the default value.

src/ai/backend/manager/repositories/auth/db_source/db_source.py
```diff
@@ -248,27 +264,7 @@ async def modify_ssh_keypair(self, access_key: str, public_key: str, private_key
...
-            status_info=row.status_info,
-            created_at=row.created_at,
-            modified_at=row.updated_at,
-            password_changed_at=row.password_changed_at,
-            domain_name=row.domain_name or "",
-            role=row.role or UserRole.USER,
-            integration_name=row.integration_id,  # DB column is integration_id
-            resource_policy=row.resource_policy,
-            sudo_session_enabled=row.sudo_session_enabled,
-        )
```
HyeockJinKim, line 271: I'm not sure why this is defined separately elsewhere and done this way, but please make it a method.

src/ai/backend/manager/repositories/auth/db_source/db_source.py
```diff
@@ -389,6 +385,37 @@ async def record_login_history(
...
+    async def fetch_user_by_email(self, email: str) -> UserData:
+        return await self._fetch_user_by_lookup(UserEmailLookup(email=email))
+
+    @auth_db_source_resilience.apply()
+    async def fetch_user_by_username(self, username: str) -> UserData:
+        return await self._fetch_user_by_lookup(UserNameLookup(username=username))
+
+    @auth_db_source_resilience.apply()
+    async def fetch_user_by_access_key(self, access_key: AccessKey) -> UserData:
+        return await self._fetch_user_by_lookup(KeypairAccessKeyUserLookup(access_key=access_key))
```
HyeockJinKim, line 417: I wanted it to provide only the lookup method and pass in different values from the parent class, but it seems to have been implemented incorrectly.

src/ai/backend/manager/plugins/auth.py
```diff
@@ -0,0 +1,58 @@
...
+class AuthPluginConfig(BasePluginConfig):
+    """The `plugin-config.auth.<name>` section an authentication plugin is created from."""
+
+
+class AuthPlugin(BasePlugin[AuthPluginConfig], metaclass=ABCMeta):
+    """The contract an authentication integration implements.
+
+    The plugin names the account a request belongs to; the manager performs the
+    lookup, the status checks and everything downstream.
+    """
```
HyeockJinKim, line 26: Let's also add`on_server_verified`, `on_login_success` and `on_login_failed`.

PR #14022, opened 2026-08-27
chore(BA-7508): read the audit-log and login-history client IP as a string

fregataa:
## Summary

- `audit_logs.client_ip` and `login_history.client_ip` are `inet`, so the driver decodes them into `ipaddress` objects while every layer above the row declares `str | None`. Reading an audit log therefore failed pydantic validation and `GET /v2/audit-logs`, `adminAuditLogsV2` and `scopedAuditLogsV2` all answered 500.
- Add `IPAddressColumn`, a `TypeDecorator` over `INET` that keeps the value a string on both sides, and use it for both columns. The DDL is unchanged, so there is no migration and rows written before this read back the same.
- `login_history.client_ip` had the same defect and is fixed with it; only the audit-log surfaces were reported.

## Test plan

- [x] `pants test tests/unit/manager/repositories/audit_log/test_client_ip_column.py` — round-trip of IPv4 / IPv6 / masked / NULL values, a row written as a bare `inet` before this change, a prefixed value, and update-then-read (replace and clear).
- [x] `pants test tests/unit/manager/repositories/auth/test_login_history_client_ip.py` — the same round-trip for `login_history`.
- [x] Verify `./bai audit-log search --limit 2` against a running manager.

Resolves BA-7508


changes/14022.misc.md
HyeockJinKim, line 1: This is supposed to be generated by CI, but it looks like it was skipped. Could you please check it?

HyeockJinKim, line 1: This is supposed to be generated by CI, but it looks like it was skipped. Could you please check it?

HyeockJinKim (approved):
Thank you. Defining a column is a better approach. 👍

PR #14025, opened 2026-08-27
fix(BA-7510): source known resource slots from the slot-type registry

fregataa:
## Summary

- `ResourceSlot.normalize_slots()` reads its known-slot set from the `current_resource_slots` ContextVar, which is declared with no default. `ResourcePresetRepository.list_presets()` returns early on a preset-list cache hit, skipping the only call that populated it, so every cache-hit `GET /resource/presets` raised `LookupError` and returned 500. Reproduced against a running manager: 200 on the first request after restart, 500 on all 9 that followed.
- Adds `ResourceSlot.normalize_slots_by_known_slots(known_slots, *, ignore_unknown)` and uses it at the manager's four call sites with the known slots read from the `resource_slot_types` table. `normalize_slots()` stays for the agent, whose four call sites set the ContextVar on the preceding line.
- Same defect as BA-1459, whose one-line fix was removed by the cache layer in BA-2293. Sourcing the registry explicitly removes the implicit dependency that neither the type checker nor review caught.
- **Behaviour change:** preset responses now carry every enabled registry slot instead of whatever etcd happened to hold, so unset accelerator slots appear as `"0"` (2 slots → 14 on the deployment under test). Allocatability is unaffected — `quantities_ge` treats a missing slot as 0, so `0 >= 0` never flips a preset from allocatable to not.
- Drops the `current_resource_slots.set(...)` workarounds in `tests/component/conftest.py` and the resource-preset service unit test, which had been masking this defect.

## Test plan

Verified against a running manager, admin and plain user:

- [x] `GET /resource/presets` — 10/10 200 (admin), 6/6 200 (user); was 200 then 500×5 before the fix
- [x] `POST /resource/check-presets` — 200 on both cache miss and hit, `allocatable` unchanged
- [x] `./bai admin resource-preset check-availability` — 3/3 OK, `available` unchanged
- [x] `pants test tests/unit/manager/services/resource_preset:: tests/unit/manager/repositories/resource_preset:: tests/component/resource_preset:: tests/component/resource_allocation:: tests/component/infra::`

Resolves BA-7510

https://claude.ai/code/session_0174YdjLSbhCawNoariv4bab


HyeockJinKim (approved):
The implementation isn't quite right, but since this is a fix, I'll just accept it with a simple tweak.

PR #14029, opened 2026-08-27
fix(BA-7512): bind the halfstack GraphQL gateway to all interfaces

studioego:
Resolves BA-7512

The halfstack GraphQL gateway (`backendai-half-apollo-router`, running
`ghcr.io/graphql-hive/gateway:2.1.12`) is unreachable on WSL2, so web login fails
with `ManagerConnectionUnavailable` on `POST /func/admin/gql`.

Hive Gateway picks its default bind address from the kernel release string and
chooses `127.0.0.1` when it contains `"microsoft"`. A container shares the host
kernel, so inside the container `os.release()` is still
`...-microsoft-standard-WSL2`; the gateway concludes it is running on a WSL host
and listens on loopback only, leaving its published port 4000 unreachable. The
webserver's `HealthyEndpointPool` probe then marks every upstream unhealthy.

The `HOST=0.0.0.0` env var already present in the compose file is inert -- the
CLI binds an env var to `--port` but not to `--host` -- so the bind address is
now passed on the command line instead.

This is a no-op on Linux, where the same default branch already yields
`0.0.0.0`, and it does not widen host-level exposure, which the unchanged
`ports: - "4000:4000"` line governs.

### Verification (WSL2, Ubuntu 24.04, kernel 6.18.33.2-microsoft-standard-WSL2)

Container listening socket, before and after (`/proc/net/tcp`, state `0A`):

```
before: 0100007F:0FA0   -> 127.0.0.1:4000
after:  00000000:0FA0   -> 0.0.0.0:4000
```

```
$ curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:4000/readiness
200

$ curl -s -X POST http://127.0.0.1:8090/server/login -H 'Content-Type: application/json' \
    -d '{"username":"admin@lablup.com","password":"..."}'
{"authenticated": true, "data": {"access_key": "AKIAIOSFODNN7EXAMPLE", "role": "superadmin", "status": "active"}}

$ curl -s -b cookies.txt -X POST http://127.0.0.1:8090/func/admin/gql -H 'Content-Type: application/json' \
    -d '{"query":"query { agent_list(limit:5, offset:0) { total_count items { id status } } }"}'
{"data":{"agent_list":{"total_count":1,"items":[{"id":"i-...","status":"ALIVE"}]}}}
```

### Affected versions

`26.8` and `26.4` carry the same image tag and the same bug. `25.15` is
unaffected -- it still runs the Rust Apollo Router
(`ghcr.io/apollographql/router:v1.61.9`), which has no such heuristic.

No test accompanies this change: it is a single compose command-line argument
with no Python target to exercise. Verified manually as shown above.


docker-compose.halfstack-ha.yml
```diff
@@ -387,7 +387,10 @@ services:
       - half
     ports:
       - "4000:4000"
-    command: ["supergraph"]
+    # Hive Gateway defaults --host to 127.0.0.1 when os.release() contains "microsoft",
+    # which is true inside a container on a WSL2 kernel. Bind explicitly; the HOST env
+    # var is not read by the CLI (only PORT is).
```
fregataa: Remove or write comments concisely

docker-compose.halfstack-ha.yml
```diff
@@ -387,7 +387,10 @@ services:
...
+    command: ["supergraph", "--host", "0.0.0.0"]
     environment:
       - HOST=0.0.0.0
```
fregataa: If this env is not used at all, how about removing them?

fregataa (commented):
Please remove `Backport: none` in PR description if this work should be backported

docker-compose.halfstack-ha.yml
studioego: Okay. I'll remove the comment.
Also, I checked the Hive Gateway CLI code.
https://github.com/graphql-hive/gateway/blob/17d76ab6/packages/gateway/src/cli.ts#L292-L317
`--port` is bound to the `PORT` env var via `.env('PORT')`, but `--host` has no env binding, so `HOST=0.0.0.0` is silently ignored.

Based on this, I'm also planning to open an upstream PR to Hive Gateway that binds the `HOST` environment variable to `--host` in the CLI.

docker-compose.halfstack-ha.yml
studioego: Okay.

studioego:
I submitted Issue https://github.com/graphql-hive/gateway/issues/2547 for Hive gateway cli.

checked the Hive Gateway CLI code.
https://github.com/graphql-hive/gateway/blob/17d76ab6/packages/gateway/src/cli.ts#L292-L317
--port is bound to the PORT env var via .env('PORT'), but --host has no env binding, so HOST=0.0.0.0 is silently ignored.

As a result, I request to add `HOST` env binding code.

fregataa (commented):
Is this needed to be backported or not??
As the description, 26.4 and 26.8 has the same issue but you marked `Backport: none`. Is there any reason?

studioego:
there was no technical reason. `26.8` and `26.4` pin the same gateway image tag and
carry the identical bug, so the fix applies to both unchanged. 
Now, I remove `Backport: none`.

fregataa (approved):

PR #14030, opened 2026-08-27
refactor(BA-7513): route plugin database access through the auth repository

fregataa:
## Summary
- `keypairs.secret_key` is stored encrypted at rest, so a plugin that queries the table on its own reads ciphertext. `AuthRepository` is now the single entry point plugins reach the database through, and decryption happens inside it — no plugin code holds `KeyProviderPool`, `KEYPAIR_SECRET_KEY_CONTEXT`, or a SQLAlchemy query any more.
- `server.py` exposes `AuthRepository` alone on `root_app`, not the whole repository container. The repository takes v2 specs — `lookup(DataLookup)` and `query_user_data(DataQuerier[UserRow, …])` — so callers name a declared spec instead of building SQL, and the only hand-written methods left are the ones that decrypt or write several rows.
- The authorize hooks return the new `AuthorizingUser` DTO instead of a raw `sa.RowMapping`. It carries `totp_activated` / `totp_key` / `username`, which the POST_AUTHORIZE two-factor plugins read and no existing `data/` type had.

## Test plan
- [x] `tests/component/openid` passes (hook returns the DTO, `create_user_if_not_exists` returns `UserID`)
- [x] `tests/unit/manager/repositories/auth` passes with the new `V2DBOpsProvider` argument
- [x] Keypair sToken login and signature-based authorization both succeed against an encrypted `keypairs.secret_key`
- [x] Keypair authorization still succeeds against a legacy plaintext row
- [ ] OpenID first login provisions the user and its default keypair; a second login reuses them

Resolves BA-7513

https://claude.ai/code/session_01RJSXWXGpBjxa7dE6Y7Q1ox

### DIFF
--- a/changes/14030.enhance.md
+++ b/changes/14030.enhance.md
@@ -1,0 +1,1 @@
+Plugins now reach the database only through `AuthRepository`, so keypair authentication keeps working with `keypairs.secret_key` encrypted at rest.
--- a/src/ai/backend/manager/data/auth/types.py
+++ b/src/ai/backend/manager/data/auth/types.py
@@ -82,6 +82,33 @@ class AuthenticatedUser:
 
 
 @dataclass(frozen=True)
+class KeyPairSigningMaterial:
+    """The owner and decrypted secret key of a keypair, as a signature check needs them."""
+
+    user_id: UserID
+    secret_key: str
+
+
+@dataclass(frozen=True)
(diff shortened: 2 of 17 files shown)


HyeockJinKim (approved):
Logically, it should be "approve," but please check it and let me know.

PR #14040, opened 2026-08-28
feat(BA-7511): answer per id in bulk user purge and vfolder bulk mutations

fregataa:
## Summary

Three bulk surfaces reported a count where they knew the ids, so a caller could not tell which of the entities it named were acted on. Found in the BA-7489 runtime audit, verdict `E-BULK` (`docs/reports/v2-action-audit/03-runtime.md` rows 444, 802, 803).

- **`global_purge_users`** — the service already collected `purged_user_ids` in request order (`services/user/service.py:271`); the adapter dropped them for `purged_count()`. Carried into the payload.
- **`vfolder bulk-delete`** — the loop had no per-item guard, so the first failure escaped while the ids before it stayed soft-deleted, and `deleted_count` was computed from `len(input.ids)` rather than from what happened. Each id is now handled independently and the payload reports `deleted_ids` and `failed`.
- **`vfolder bulk-purge`** — the loop was already per item; added `purged_ids` so the answer maps back to the request.

Every surface keeps its existing count and failure fields. The id lists are additive, and together with `failed` they answer for each requested id exactly once, in request order — duplicates are not collapsed. New GQL fields use `NEXT_RELEASE_VERSION`.

Not in this PR: the other six `E-BULK` rows (`delete_artifacts`, `restore_artifacts`, `assign_users_to_project`, the three `bulk_upsert_*_fair_share_weights`) need repository-level changes and are left for a follow-up. Distinguishing a denial from a miss by machine still needs an `error_code` on the bulk error DTOs, which is a cross-cutting decision across every bulk surface.

## Test plan

- [x] `bulk-purge-users` with a mix of existing and nonexistent user ids returns every id once across `purged_user_ids` and `failed`
- [x] `vfolder bulk-delete` where one id is denied: the permitted ids appear in `deleted_ids`, the denied one in `failed`, and the call no longer returns a bare 403
- [x] `vfolder bulk-delete` `deleted_count` matches `len(deleted_ids)` rather than the input length
- [x] `vfolder bulk-purge` with a duplicated id answers twice — once success, once failure
- [x] GraphQL schema exposes `deletedIds`, `failed` on `BulkDeleteVFoldersV2Payload`, `purgedIds` on `BulkPurgeVFoldersV2Payload`, `purgedUserIds` on `BulkPurgeUsersV2Payload`

Resolves BA-7511

https://claude.ai/code/session_013i38G9Y7aVGAArqrEFDKuN

----
📚 Documentation preview 📚: https://sorna--14040.org.readthedocs.build/en/14040/

----
📚 Documentation preview 📚: https://sorna-ko--14040.org.readthedocs.build/ko/14040/

### DIFF
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -2560,14 +2560,14 @@ type BulkDeleteVFoldersV2Payload
   @join__type(graph: STRAWBERRY)
 {
   """
-  Added in 26.9.0. UUIDs of the vfolders that were soft-deleted, in the order they were requested. Together with `failed` this answers for every requested vfolder exactly once.
+  Added in 26.9.0. The vfolders that were soft-deleted, in the order they were requested. A soft-deleted vfolder is still addressable, so it is returned whole. Together with `failed` this answers for every requested vfolder exactly once.
   """
-  successes: [UUID!]!
+  items: [VFolder!]!
 
   """
   Added in 26.4.2. Number of virtual folders successfully soft-deleted. Deprecated since 26.9.0.
   """
-  deletedCount: Int! @deprecated(reason: "Use the length of successes.")
+  deletedCount: Int! @deprecated(reason: "Use the length of items.")
 
   """
   Added in 26.9.0. List of errors for vfolders that failed to be soft-deleted.
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -2559,8 +2559,33 @@ input BulkDeleteVFoldersV2Input
 type BulkDeleteVFoldersV2Payload
   @join__type(graph: STRAWBERRY)
 {
-  """Number of virtual folders successfully soft-deleted."""
-  deletedCount: Int!
+  """
+  Added in 26.9.0. UUIDs of the vfolders that were soft-deleted, in the order they were requested. Together with `failed` this answers for every requested vfolder exactly once.
+  """
+  successes: [UUID!]!
+
+  """
+  Added in 26.4.2. Number of virtual folders successfully soft-deleted. Deprecated since 26.9.0.
+  """
+  deletedCount: Int! @deprecated(reason: "Use the length of successes.")
+
+  """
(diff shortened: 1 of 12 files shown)


docs/manager/graphql-reference/supergraph.graphql
```diff
@@ -2559,6 +2559,29 @@ type BulkDeleteVFoldersV2Payload
...
+  """
+  Added in 26.9.0. UUIDs of the vfolders that were soft-deleted, in the order they were requested. Together with `failed` this answers for every requested vfolder exactly once.
+  """
+  deletedIds: [UUID!]!
+
+  """
+  Added in 26.9.0. List of errors for vfolders that failed to be soft-deleted.
+  """
+  failed: [BulkDeleteVFolderV2Error!]!
+}
```
HyeockJinKim, line 2576: For bulk mutations, let's settle on two payload shapes:

Non-purge operations — the affected objects are still accessible, so return them:
items: [Node!]!
failed: [FailedItem!]!

Bulk purge — the objects no longer exist, so return IDs only:
successes: [UUID!]!
failed: [FailedItem!]!

Could you also document this convention in common/dto/AGENTS.md?

HyeockJinKim: "Failed item" simply recommends defining individual failure type.

docs/manager/graphql-reference/supergraph.graphql
```diff
@@ -2559,8 +2559,33 @@ input BulkDeleteVFoldersV2Input
 type BulkDeleteVFoldersV2Payload
   @join__type(graph: STRAWBERRY)
 {
-  """Number of virtual folders successfully soft-deleted."""
-  deletedCount: Int!
+  """
+  Added in 26.9.0. UUIDs of the vfolders that were soft-deleted, in the order they were requested. Together with `failed` this answers for every requested vfolder exactly once.
+  """
+  successes: [UUID!]!
```
HyeockJinKim: If it were a "soft delete," wouldn't it make more sense to provide the `Items`? The reason `purge` only provides the ID is that the node can no longer be referenced, but a "soft delete" is a different case.

HyeockJinKim (approved):

PR #14066, opened 2026-09-01
feat(BA-7526): read project-owned entities from the virtual scope chain

fregataa:
## Summary
- Generalize the virtual-scope membership predicate into `scope_membership_exists`, which takes the entity type; `user_scope_membership_exists` becomes a thin wrapper over it.
- Convert `ProjectVFolderOperationScope`, `ProjectDeploymentOperationScope` and `ProjectModelCardOperationScope` to resolve ownership through `entity_memberships` joined to `virtual_scopes` instead of the owner column on the row.
- Backfill the project membership of existing vfolders, endpoints and model cards. The earlier chain backfill (`3ebcf2c3c959`) covered `entity_type = 'user'` only, so without this the converted reads would answer nothing for rows created before the chain.

## Test plan
- [x] `tests/unit/manager/repositories/model_card/test_model_card_search_in_project.py` (new) — project A returns only its enrolled cards, project B only its own, a card with the project column but no membership stays out, and an unknown project raises `ProjectNotFound`
- [x] `tests/unit/manager/repositories/vfolder/test_vfolder_search_in_project.py` — fixture now seeds the chain
- [x] `tests/unit/manager/repositories/deployment/test_deployment_search_in_project.py` — fixture now seeds the chain
- [x] `pants check` — `testutils/virtual_scope.py` passes `common.data.permission.types.EntityType` (a `StrEnum`) where the column is typed `common.data.entity.types.EntityType` (a `str` subclass); the values match at runtime but mypy may object
- [x] Component suites that create through the API (`tests/component/model_card`, `tests/component/vfolder`) — the v2 create ops already write the membership, so these should be unaffected

## Notes
The backfill is limited to the `entity_memberships` edge the converted reads consult. The full provisioning the create path performs — the entity's own virtual scope, its self membership and self binding — is left to BA-6657.

Resolves BA-7526
Resolves BA-7529
Resolves BA-7530


src/ai/backend/testutils/virtual_scope.py
```diff
@@ -60,20 +60,34 @@ async def seed_user_scope(self, sess: AsyncSession, user_id: uuid.UUID) -> None:
...
     async def enroll_user_in_project(
         self, sess: AsyncSession, group_id: uuid.UUID, user_id: uuid.UUID
     ) -> None:
         """Write the chain rows the enrollment path creates for a user-project
         membership: the user joins the project's virtual scope. The project is not bound
         into the user's own virtual scope — a member does not hand the project its
         personal entities."""
-        project_scope_id = await self.get_or_create_scope(sess, ScopeType.PROJECT, group_id)
-        sess.add(
-            EntityMembershipRow(
```
HyeockJinKim: It keeps getting added to `testutils`; I think we should put this in `AGENTS.md`. If we leave it like this, `testutils` will keep contaminating `manager`.

src/ai/backend/manager/models/alembic/versions/b41d7c92e58a_backfill_project_entity_memberships_for_vfolders_endpoints_model_cards.py
```diff
@@ -0,0 +1,77 @@
...
+down_revision = "c8e4b1a09d37"
+branch_labels = None
+depends_on = None
+
+
+_OWNED_BY_PROJECT = (
+    ("vfolder", "vfolders", "group"),
+    ("deployment", "endpoints", "project"),
+    ("model_card", "model_cards", "project"),
+)
```
HyeockJinKim: It looks like only this part will be migrated separately this time, so I'll make sure to add the migration step after finishing the invitation task.

PR #14142, opened 2026-09-02
feat(BA-7608): send the rate limit headers on every response of a rate-limited request

### LINKED ISSUE #14141: Manager 429 rate limit response lacks the X-RateLimit headers
## Objective

Make the manager's HTTP 429 response from the rate limit middleware carry X-RateLimit-Limit, X-RateLimit-Remaining, and X-RateLimit-Window, the same headers it sets on every allowed response.

## Background

The rate limit middleware (manager/api/rest/ratelimit/handler.py) raises RateLimitExceeded before the lines that set the headers, so only allowed responses carry them. A client that reads the headers on 429 to back off gets nothing. The web server proxy copies upstream headers as-is, so the gap is also visible through the web server (observed on a live cluster while verifying BA-7603).

## Acceptance Criteria

- A request rejected by the manager rate limiter returns 429 with X-RateLimit-Limit, X-RateLimit-Remaining (0), and X-RateLimit-Window.
- Allowed responses keep the same headers as before.
- The unit tests for the middleware assert the headers on the raised RateLimitExceeded.

## References

Parent epic BA-7361. Surfaced while verifying BA-7603 (PR #13771).

JIRA Issue: BA-7608

jopemachine:
Resolves #14141 (BA-7608)

## Summary
- The manager's rate limit middleware attached `X-RateLimit-Limit`, `X-RateLimit-Remaining` and `X-RateLimit-Window` only to the response a handler returned. The 429 it raised itself, errors raised inside handlers, and debug-mode error responses carried none of them.
- The middleware now only computes the quota and stores it on the request as a frozen `RateLimitQuota` dataclass. A new `on_response_prepare` hook, `apply_rlim_headers`, registered next to the middleware in `mount_registries`, turns that quota into the three headers on whichever response finally goes out. aiohttp runs the hook right before writing the headers for every response, whether a handler returned it, an `HTTPException` was raised, or the exception middleware built a new one, so all of them now carry the headers. The 429 reports `X-RateLimit-Remaining: 0`.
- The web server proxy forwards upstream headers as-is, so a 429 the manager returns for a proxied request now reaches the browser with the headers too (observed missing while verifying #13771 on a live cluster).

## Test plan
- [x] `tests/unit/manager/api/test_ratelimit.py`: each case runs the middleware, then the hook, and asserts the three headers on the response for anonymous, within-limit and exceeded requests; a request without a quota checks the hook leaves the response alone.
- [x] Live check against halfstack with a keypair limit of 3: requests 1–3 return 200 with `Remaining` 2, 1, 0; requests 4–5 return 429 with `X-RateLimit-Limit: 3`, `X-RateLimit-Remaining: 0`, `X-RateLimit-Window: 900`.

https://claude.ai/code/session_01EdrsTvazb2QprMmS8DpEDM


src/ai/backend/manager/api/rest/ratelimit/handler.py
```diff
@@ -40,21 +48,19 @@ async def rlim_middleware(
             rate_limit = request["keypair"]["rate_limit"]
             rolling_count = await valkey_client.execute_rate_limit_logic(
                 user_id=request["user"]["uuid"],
-                window=_rlim_window,
+                window=_RATELIMIT_WINDOW,
             )
             if rate_limit is not None and rolling_count > rate_limit:
-                raise RateLimitExceeded
+                raise RateLimitExceeded(headers=_rlim_headers(rate_limit, 0))
```
HyeockJinKim: I don't quite understand this part. When it's "raised," how is the header inserted?

src/ai/backend/manager/api/rest/ratelimit/handler.py
```diff
@@ -22,7 +25,19 @@
...
+@dataclass(frozen=True)
+class RateLimitQuota:
+    limit: int | None
+    remaining: int
+    window: int = _RATELIMIT_WINDOW
+
+    def apply_to(self, headers: CIMultiDict[str]) -> None:
+        headers["X-RateLimit-Limit"] = str(self.limit)
+        headers["X-RateLimit-Remaining"] = str(self.remaining)
+        headers["X-RateLimit-Window"] = str(self.window)
```
HyeockJinKim, line 40: I do wonder if ABC might be better.

HyeockJinKim (approved):

PR #14145, opened 2026-09-02
fix(BA-7617): terminate a route whose bound session is terminal

HyeockJinKim:
## Summary

- A replica whose session died inside the health check `initial_delay` stayed in `PROVISIONING` + `WARMING_UP` until the warming-up timeout expired, because the session liveness check only ran on `RUNNING` routes.
- The running route handler now also picks up `PROVISIONING` routes, with no sub-status or health filter, so a terminal session sends the route straight to `TERMINATING` + `DRAINING`.
- A route with no `session_id` yet passes the check untouched. The status map cannot tell "not enqueued yet" apart from "session row gone", so the split is made on `RouteData.session_id`; without it every pre-enqueue `PENDING` route would be terminated.

## Test plan

- [x] A `PROVISIONING` + `WARMING_UP` route whose session is terminal is classified as a failure before the initial delay elapses.
- [x] The same holds for a `PROVISIONING` + `STARTING` route.
- [x] A `PROVISIONING` + `PENDING` route with `session_id` unset is not transitioned.
- [x] Existing `RUNNING` route behavior is unchanged.
- [x] The handler targets `[PROVISIONING, RUNNING]` and keeps the `TERMINATING` + `DRAINING` + `NOT_CHECKED` failure transition.

Resolves BA-7617

https://claude.ai/code/session_01Dq47RTG5REBoXjgJG3cPRj

### DIFF
--- a/src/ai/backend/manager/sokovan/deployment/route/executor.py
+++ b/src/ai/backend/manager/sokovan/deployment/route/executor.py
@@ -339,13 +339,13 @@ class RouteExecutor:
         )
 
     async def check_running_routes(self, routes: Sequence[RouteData]) -> RouteExecutionResult:
-        """Check health status of running routes.
+        """Check that each route's bound session is still alive.
 
         Args:
-            routes: Routes to check (should be HEALTHY or UNHEALTHY status)
+            routes: PROVISIONING or RUNNING routes to check
 
         Returns:
-            Result containing routes with updated health status
+            Result whose errors are the routes whose session is gone or terminal
         """
         # Phase 1: Load status
         with RouteRecorderContext.shared_phase("load_status"):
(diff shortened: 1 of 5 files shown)


src/ai/backend/manager/sokovan/deployment/route/executor.py
```diff
@@ -1182,12 +1182,18 @@ def _verify_route_session_status(
...
+        A route with no bound session yet passes; the status map cannot tell that
+        case apart from a missing session row.
+        """
         pool = RouteRecorderContext.current_pool()
         recorder = pool.recorder(route.route_id)
 
         with recorder.phase("verify_session"):
             with recorder.step("check_session_exists"):
+                if route.session_id is None:
+                    return
```
seedspirit, line 1196: Is it okay to treat `_verify_route_session_status` as a success when there is no session ID because it now also verifies PROVISIONING status?

HyeockJinKim: This is an action taken to clean up when a session has ended; it does not account for scenarios where no session is registered.

PR #2415, opened 2024-07-10
fix: Utilize `ExtendedJSONEncoder` for error logging to handle `UUID` extra data

rapsealk:
This pull request fixes potential errors caused by invalid attempt to serialize `UUID` object.

| Before | After |
| ------- | ----- |
|  | 스크린샷 2024-07-10 오후 4 38 20 |

**Checklist:** (if applicable)

- [x] Milestone metadata specifying the target backport version

----
📚 Documentation preview 📚: https://sorna--2415.org.readthedocs.build/en/2415/

----
📚 Documentation preview 📚: https://sorna-ko--2415.org.readthedocs.build/ko/2415/

PR #14175, opened 2026-09-03
fix(BA-7631): raise HTTP exceptions instead of returning them in web and appproxy

### LINKED ISSUE #14173: Webserver and appproxy handlers return HTTPException objects instead of raising them
## Objective

Raise HTTP exceptions from the webserver and appproxy aiohttp handlers instead of returning the exception object as the response. Follow-up to BA-7625 / #14164, which did the same for the manager.

## Background

Returning an `HTTPException` is deprecated since aiohttp 3.0 (#2415) and breaks on aiohttp 4 master, where `HTTPException` no longer subclasses `Response`: the server fails with `AttributeError: ... object has no attribute 'prepare'` and answers 500.

Remaining sites: `web/server.py` (10), `web/proxy.py` (8), `appproxy/coordinator/api/proxy.py` (1), `appproxy/worker/api/setup.py` (1). The appproxy exception middlewares rewrite every non-`BackendAIError` `HTTPException` other than 404/405 into `GenericBadRequest`, so raised redirects need to pass through them, as done for the manager in #14166.

## Acceptance Criteria

- No aiohttp handler or middleware in `web/` or `appproxy/` returns an `HTTPException` instance.
- Responses seen by clients (status, headers, body) are unchanged.
- Raised redirects in appproxy reach the client as redirects, not as a generic 400.

## Related

- Jira: BA-7631
- BA-7625 / #14164 / #14166 (manager)

Story points: 1

jopemachine:
Resolves #14173 (BA-7631)

## Summary
- Follow-up to #14166 for the remaining components. Returning an `HTTPException` object from a handler is deprecated since aiohttp 3.0 (`returning HTTPException object is deprecated (#2415)`) and fails on aiohttp 4 master with `AttributeError: ... object has no attribute 'prepare'` and a 500. Every site in `web/server.py` (10), `web/proxy.py` (8), `appproxy/coordinator/api/proxy.py` (1) and `appproxy/worker/api/setup.py` (2, one of them an `HTTPFound` built into a variable, given a cookie, and then returned) now raises. Sites inside `except` blocks chain the cause with `from e`.
- `web/server.py` `_check_params` returned the `HTTPBadRequest` for its caller to return; it now raises it and the caller no longer inspects a result.
- The appproxy coordinator and worker exception middlewares rewrote every non-`BackendAIError` HTTP exception other than 404/405 into `GenericBadRequest`, which would have turned the raised redirects into a 400. They now pass `HTTPSuccessful` and `HTTPRedirection` through untouched and apply the rewrite to `HTTPError` only, as the manager does since #14166.
- The webserver's `security_policy_middleware` applied its response policies (CSP, `X-Content-Type-Options`) only to the response the handler returned. It now applies them to a raised `HTTPException` as well before re-raising it, so the converted sites keep those headers.

## Behavior notes
- The webserver's `general_exception_middleware` re-raises every `HTTPException` untouched, but logs 4xx as a warning and 5xx with a traceback. The converted 4xx/5xx sites now produce those log lines; the returned responses did not.
- `logout_handler` raises `web.HTTPOk()`; the response is the same 200.

## Test plan
- [x] `tests/unit/webserver/test_security_policy.py`: a handler that raises `HTTPFound` gets the response policy headers on the exception. The file's 23 tests pass.
- [x] Live check on halfstack, webserver from `main` vs this branch, same requests: `/server/update-password-no-auth` and `/server/token-login` with missing fields (400), `/server/logout` (200), and with the manager stopped `/server/login`, `/server/token-login`, `/server/update-password-no-auth`, `/func/custom-auth/login` (502 / the pre-existing 500). Status, headers and bodies are identical. The webserver log now shows the `client error raised inside handlers` warning for the 4xx cases.
- [x] Same live check, `GET /..%2F..%2Fetc%2Fpasswd` (404 raised by `console_handler`): identical except the response now also carries `X-Content-Type-Options: nosniff`, which is the security policy middleware applying to raised exceptions.
- [x] Live check on halfstack, appproxy coordinator and worker from `main` vs this branch: with a running session, a token from `POST /v2/conf` and a circuit from `/v2/proxy/{token}/{session}/add`, `GET /v2/proxy/auth` (308 to the worker) and the worker's `GET /setup` (302 with the `appproxy_permit` cookie and CORS headers) are identical in status, headers and body.
- [ ] CI

https://claude.ai/code/session_01EdrsTvazb2QprMmS8DpEDM

### DIFF
--- a/changes/14175.fix.md
+++ b/changes/14175.fix.md
@@ -1,0 +1,1 @@
+Raise HTTP exceptions instead of returning them from the webserver and appproxy handlers, keeping the appproxy redirects and the webserver's security response headers intact.
--- a/src/ai/backend/appproxy/coordinator/api/proxy.py
+++ b/src/ai/backend/appproxy/coordinator/api/proxy.py
@@ -221,7 +221,7 @@ async def proxy(
             ),
             headers={"Access-Control-Allow-Origin": "*", "Access-Control-Expose-Headers": "*"},
         )
-    return web.HTTPPermanentRedirect(app_url)
+    raise web.HTTPPermanentRedirect(app_url)
 
 
 async def init(_app: web.Application) -> None:
--- a/src/ai/backend/appproxy/coordinator/server.py
+++ b/src/ai/backend/appproxy/coordinator/server.py
@@ -242,7 +242,9 @@ async def exception_middleware(
             else:
                 log.warning("Client error: {0!r}", ex)
             raise
-        except web.HTTPException as ex:
+        except (web.HTTPSuccessful, web.HTTPRedirection):
+            raise
+        except web.HTTPError as ex:
             if ex.status_code == 404:
                 raise URLNotFound(extra_data=request.path) from ex
             if ex.status_code == 405:
--- a/src/ai/backend/appproxy/worker/api/setup.py
+++ b/src/ai/backend/appproxy/worker/api/setup.py
@@ -182,7 +182,7 @@ async def setup(
                 samesite="Lax",
                 max_age=604800,  # 7 days
             )
-            return response
+            raise response
         case ProxyProtocol.TCP:
             protocol = "tcp"
             queryparams = {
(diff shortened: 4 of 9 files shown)


jopemachine:
Folded into #14166 so the return→raise change lands in one PR across manager, agent, webserver and appproxy.

PR #14209, opened 2026-09-04
feat(BA-7375): upgrade valkey-glide to 2.5.2

fregataa:
## Summary

- Raise the `valkey-glide` pin from `~=2.3.1` to `~=2.5.2`. 2.5.x replaces the Unix domain socket transport with FFI, so no glide socket file is created under `/tmp` and `GlideClient.create()` no longer leaks the UDS stream and reader task when it fails partway through.
- Drop the partial-client cleanup workaround (`_TrackedGlideClient`, `_create_glide_client`, `_close_partial_glide_client`, `_pending_glide_clients`) added for BA-7397 and call `GlideClient.create()` directly, matching `redis_helper`. The `_reader_task` attribute it reached for no longer exists.
- Import exceptions from the top-level `glide` package. `glide.exceptions` is a deprecation shim that never resolved for the type checker, so both `# type: ignore[import-not-found]` comments go away.
- Waive mypy's `no-untyped-call` for `ai.backend.common.clients.valkey_client.*` in `pyproject.toml`: 2.5.2 moved `Script` into the `py.typed` `glide` package with an unannotated `__init__`, which fails that check at every call site.
- Add a regression test for BA-7397 and re-enable `test_global_timer_etcdlock`.

### Verification against 2.5.2

Measured on a local Valkey 9.1.1 with an isolated py3.13.7 environment:

| Success criterion | Result |
|---|---|
| Every module importing from `glide` resolves | All 24 names used across 23 source and 3 test modules resolve at the top level; config kwargs unchanged |
| No glide socket under `/tmp` after startup | 0 new socket files (2.3.1 had left 679 stale ones on the dev box) |
| Repeated GET of a large value keeps RSS flat | +1.6 MB over 2000x 1 MiB GET (2 GiB transferred); valkey-io/valkey-glide#6740 does not reproduce, fixed in 2.5.1 |
| A forked child can create a client and run a command | Child `PING` succeeds with the parent holding a live client, exit 0. No sync glide client is used anywhere, so valkey-io/valkey-glide#6672 does not apply |

### BA-7397 regression test

`tests/unit/common/clients/valkey_client/test_connection_leak.py` injects the production fingerprint without docker or a real Valkey: a listener that completes the TCP handshake and then never speaks the Valkey protocol (`cmd=NULL`, empty `lib-name`, `resp=2`). Each round runs six concurrent `connect()` calls, mirroring a reconnect cycle of three `MonitoringValkeyClient`s.

| glide version | Result | Sockets left open per round |
|---|---|---|
| 2.3.1 | fails | `[0, 0, 12]` — 6 clients x 2, the accumulation shape reported in BA-7397 |
| 2.5.2 | passes (3/3 runs) | `[0, 0, 0]` |

The second leak path in BA-7397 — `disconnect()` against an unresponsive server failing to release the server-side connection — needs a real Valkey and stays in the repro script.

## Known regression in 2.5.2, not addressed here

`GlideClient.create()` blocks the asyncio event loop for the entire connection attempt. `glide/glide_client.py:570` calls `self._lib.create_client(...)`, a synchronous cffi call with no `await`; 2.3.1 awaited this over the UDS pipe.

Loop stall measured with a 10 ms heartbeat, connecting through an out-of-process delay proxy to a healthy Valkey:

| Connect duration | 2.3.1 loop stall | 2.5.2 loop stall |
|---|---|---|
| 0.03s (direct) | 0.033s | 0.034s |
| 0.51s | 0.029s | 0.521s |
| 1.01s | 0.031s | 1.017s |
| 2.00s (refused port, fails) | 0.030s | 2.006s |

The stall equals the connect duration whether the connect succeeds or fails. Command execution is unaffected — 500 sequential and 500 concurrent GETs stall the loop 0.011-0.015s on both versions.

Impact: `MonitoringValkeyClient._reconnect()` recreates the operation and monitor clients, so during a Valkey outage a worker's event loop freezes for roughly `2 x connection_timeout` (4s at the default) per reconnect cycle, repeating with no backoff. Six concurrent creates serialize into 12s of frozen loop.

This is why the new regression test takes ~36s on 2.5.2 and 8s on 2.3.1.

No upstream issue exists for it yet. Bounding it would mean passing a shorter `connection_timeout` in `AdvancedGlideClientConfiguration`, but at `connection_timeout=300` the BA-7397 leak stops reproducing even on 2.3.1, so that setting would also blind the regression test. Left for a follow-up.

## Test plan

- [x] `pants test tests/unit/common/clients/valkey_client:: tests/unit/common/test_distributed.py` after `pants export --resolve=python-default`
- [x] `test_global_timer_etcdlock`, re-enabled here, spawns 7 processes against etcd and redis containers. It was disabled for a multiprocessing incompatibility with valkey-glide and was also stale (it called `EtcdLockContext(addr=...)` against an `addrs: list[HostPortPair]` field). Needs a run on Linux CI, where `multiprocessing` defaults to fork rather than spawn.
- [x] Verify a manager and an agent start against the halfstack Valkey and leave no socket file under `/tmp`.

Resolves BA-7375


src/ai/backend/common/clients/valkey_client/script.py
```diff
@@ -0,0 +1,13 @@
...
+from glide import Script
+
+# glide ships `py.typed` but leaves `Script.__init__` unannotated, so calling the class
+# directly fails the `no-untyped-call` check. Drop this once upstream annotates it.
+_new_script: Final[Callable[[str], Script]] = cast(Callable[[str], Script], Script)
+
+
+def create_script(code: str) -> Script:
+    """Compile a Lua script for `invoke_script`."""
+    return _new_script(code)
```
HyeockJinKim: Is this true??

HyeockJinKim (dismissed):

src/ai/backend/common/clients/valkey_client/script.py
fregataa: yes, but I reverted this change and added mypy rule exception on pyproject

tests/unit/common/test_distributed.py
```diff
@@ -421,68 +423,66 @@ async def test_global_timer_redlock(
...
+    for timer_process in processes:
+        timer_process.join()
+    print("checking records")
+    event_records: list[float] = []
+    while not event_records_queue.empty():
+        event_records.append(event_records_queue.get())
+    print(event_records)
+    num_records = len(event_records)
+    print(f"{num_records=}")
+    assert target_count - 2 <= num_records <= target_count + 2
```
HyeockJinKim: Is there a reason this was added back in?

HyeockJinKim: Actually, this test seems like it should be geared toward erasing things.

HyeockJinKim (approved):

PR #14211, opened 2026-09-04
feat(BA-7648): count requests in a per-user fixed window instead of a sorted set

### LINKED ISSUE #14210: Replace the sliding-log rate limit counter with a per-user fixed-window counter
## Objective

Replace the manager rate limit counter, one sorted set member per request, with one hash per user holding a count and the limit fixed for the current window, so memory per user stays constant regardless of request volume.

## Background

The current limiter adds one member to user.<uuid> (sorted set) per authorized request and returns ZCARD after pruning members older than 15 minutes. Measured cost is about 56 B per member (57.5 / 54.7 / 56.0 B at 1,000 / 10,000 / 35,000 members). The 429 decision is made after the member is added, so rejected requests are recorded as well: the set size is bounded by the client's sustained request rate, not by rate_limit. A client that ignores 429 at 1,000 rps holds about 900,000 members, about 50 MB, for one user.

## Proposed structure

- One hash per user with fields count and limit. Each request HINCRBYs count; when the result is 1 the request opens the window, stores the current rate_limit in limit and sets the TTL to the window. Later requests only increment count and return the stored limit. The hash expires with the window.
- The manager rejects with 429 when count > limit. The web server reads both fields in one call and applies the same rule; a missing key passes through.
- The limit stored for a window is the user's default keypair limit, whichever keypair opened the window.
- X-RateLimit-Remaining resets at once when the TTL expires instead of recovering gradually; X-RateLimit-Reset carries the seconds until then.
- A fixed window lets up to 2x the limit through around a window boundary. Accepted for abuse control.

## Acceptance Criteria

- The first request of a window creates the hash with count, limit and the TTL; later requests leave limit unchanged; a new window after expiry picks up the current limit.
- A keypair without a limit is counted without storing a limit and is never rejected.
- The manager returns 429 with X-RateLimit-Limit, X-RateLimit-Remaining and X-RateLimit-Reset once count exceeds the limit.
- MEMORY USAGE of the per-user key stays near 100 B after sending more requests than the limit, and does not grow after the first 429.
- The sorted set script, the global __request_id counter and the sorted-set-only client methods are removed.

## Related

- lablup/backend.ai#13771 (web server middleware; this change is stacked on it)
- BA-7365, BA-7603

JIRA Issue: BA-7648

jopemachine:
## 📚 Stacked PRs

Part of the web server per-user rate limiting stack (epic BA-7361). **Merge in order (bottom-up):**

1. ✅ #13777 — `feat(BA-7364)`: deliver the login user id in the auth response
2. ✅ #13778 — `feat(BA-7365)`: key the rate limit counter by user id
3. 👉 **#14211 — `feat(BA-7648)`: count requests in a per-user fixed window instead of a sorted set** ← you are here
4. #13771 — `feat(BA-7603)`: add a per-user rate limit middleware to the web server

Resolves #14210 (BA-7648)

## Summary

- Replace the sliding-log counter (one sorted set member per request) with one hash per user (`user.<user_id>`) holding the `count` of the current window and the `limit` fixed for it. One atomic batch per request (`HINCRBY`, `HSETNX limit`, `EXPIRE NX`, `HGET`, `TTL`): the first request creates the hash and sets the 15-minute TTL, the first request carrying a limit stores it, later requests only increment `count`. No Lua script. Memory per user is 72 B regardless of request volume, against about 56 B per request before (1,960,000 B measured at 35,000 members), rejected requests included.
- The limit fixed for a window is the user's **default keypair** limit, whichever keypair signed the request. The manager counts first; when the window carries no limit yet it resolves that limit once through a new public auth action (`public_resolve_default_keypair_rate_limit`, backed by the existing `AuthRepository.default_keypair()`) and fixes it with `HSETNX` (`store_limit()`). Later requests in the window read the stored limit and never touch the DB. The manager rejects with 429 when the count exceeds the fixed limit, so every keypair of the user is judged by one rule and a reader of the window (`get_state()`: the legacy GQL `rolling_count` field, the web server in #13771) applies the same one. A user without a default keypair is counted and never rejected.
- Every rate-limited response carries `X-RateLimit-Reset` (seconds until the window ends) next to `Limit`, `Remaining` and `Window`. The API docs describe the fixed window and reinstate the `Reset` header.
- Removed: the Lua script, the global `__request_id` counter, and the sorted-set-only client methods.

## Trade-offs of the fixed window

- **One limit per user.** The manager judges by the default keypair limit fixed when the window opened, not by each request's own keypair limit. The counter has been shared per user since #13778, so a per-keypair limit was already only half in effect; this makes the rule explicit and deterministic (no dependence on which keypair happened to open the window). A changed limit takes effect from the next window.
- **Boundary burst.** A client can pass up to 2× the limit around a window boundary (the last seconds of one window plus the first of the next). With the default 30,000 per 15 minutes that is 60,000 requests in a few seconds. The sliding window log did not allow this; the fixed window trades that exactness for constant memory. Acceptable for abuse control; see the follow-ups below if it is not for a deployment.
- **15 minutes is long for a fixed window.** Fixed-window limiters usually run on 1-second to 1-minute windows (GitHub's 1 hour is the outlier), because a shorter window shrinks both the burst (2 × 2,000 on a 1-minute window with the same rate) and the longest wait after a 429 (1 minute instead of 15). The 15-minute window comes from the sliding log, where its length cost nothing. It stays in this PR because changing it changes what `rate_limit` means: the stored keypair values are per 15 minutes and would need converting.
- **`X-RateLimit-Remaining` is a sawtooth**: it sits at 0 until the window expires, then jumps back to the limit, instead of recovering gradually. `X-RateLimit-Reset` tells clients when.
- **Windows are per user, opened by the first request**, not aligned to the clock, so users do not all reset at the same moment. The TTL lives in Redis, so the limiter no longer depends on the manager clocks agreeing (the sliding log used `time.time()` of the manager as the score).
- The counter key keeps its name; the manager flushes the rate limit DB on shutdown, so no sorted set from the previous version survives a restart.

Follow-ups, either of which removes the boundary burst while keeping constant memory:

- Shorten the window (e.g. 1 minute) and convert the stored `rate_limit` values accordingly.
- Sliding window counter: clock-aligned windows, keep the previous window's count and weight it by the overlap. Two keys per user.

Also noticed: the v2 keypair DTOs (`common/dto/manager/v2/keypair/request.py`, `response.py`) describe `rate_limit` as "requests per minute" while the window is 15 minutes. That was already wrong before this PR.

## Test plan

- [x] Unit tests for the Valkey client: the first request opens the window without a limit, later requests keep the TTL, `store_limit()` fixes the limit and a second call keeps the first, count grows past the limit, a new window starts without a limit, keyed by user
- [x] Unit tests for the manager middleware: headers including `X-RateLimit-Reset`, 429 on `count > limit`, a window opened without a limit resolves the default keypair limit once and stores it, a window with a limit does not resolve it
- [x] Batch exercised against a local Redis (valkey 9.1): TTL unchanged by later requests, hash of 72 B, listpack encoding
- [ ] Live check with a keypair limit of 3 against the manager

https://claude.ai/code/session_01BZD8hsYKVkwmiAsrX8pus2

----
📚 Documentation preview 📚: https://sorna--14211.org.readthedocs.build/en/14211/

----
📚 Documentation preview 📚: https://sorna-ko--14211.org.readthedocs.build/ko/14211/


docs/manager/common-api/auth.rst
```diff
@@ -219,7 +220,8 @@ Failure example with a missing authorization header
    Content-Length: 139
    X-RateLimit-Limit: 2000
    X-RateLimit-Remaining: 1998
-   X-RateLimit-Reset: 834821
+   X-RateLimit-Reset: 834
```
HyeockJinKim: The meaning of "reset" hasn't changed, has it?

tests/unit/common/clients/valkey_client/test_valkey_rate_limit_client.py
```diff
@@ -1,33 +1,54 @@
...
+async def test_later_requests_keep_the_window(
+    test_valkey_rate_limit: ValkeyRateLimitClient,
+) -> None:
+    user_id = UserID(uuid.uuid4())
+    await test_valkey_rate_limit.consume(user_id, window=60)
+    await asyncio.sleep(1.1)
+
+    state = await test_valkey_rate_limit.consume(user_id, window=60)
+
+    assert state == RateLimitState(count=2, reset=59)
```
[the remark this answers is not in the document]
jopemachine: Fixed in 7029f10: the test now asserts `0 < state.reset < 60`, i.e. the TTL decreased and was not reset, instead of an exact second. The same assertion in #13771 got the same change.

docs/manager/common-api/auth.rst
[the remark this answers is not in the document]
jopemachine: Fixed in 7029f10: the unauthenticated example now shows the non-authorized branch values (Limit 1000, Remaining 1000, Reset 900, Window 900).

src/ai/backend/common/clients/valkey_client/valkey_rate_limit/client.py
```diff
@@ -109,110 +94,50 @@ async def close(self) -> None:
...
-            results = await conn.exec(tx, raise_on_error=True)
-        # Handle the result properly by extracting the first result
-        if results and len(results) > 0:
-            return cast(int, results[0])
-        return 0
+            results = await conn.exec(batch, raise_on_error=True)
+        if results is None:
+            raise UnreachableError("an atomic batch without WATCH cannot be aborted")
+        count, _, reset = results
+        return RateLimitState(count=cast(int, count), reset=cast(int, reset))
```
HyeockJinKim: Why isn't the "limit" value being saved?

src/ai/backend/manager/api/rest/ratelimit/handler.py
```diff
@@ -50,21 +52,45 @@ async def rlim_middleware(
...
+            remaining = rate_limit - state.count if rate_limit is not None else state.count
+            reserve_response_headers(
+                request,
+                RateLimitQuota(
+                    limit=rate_limit,
+                    remaining=remaining,
+                    reset=state.reset,
+                    window=_RATELIMIT_WINDOW,
+                ),
+            )
```
HyeockJinKim: This seems to be working differently from what we discussed earlier—could you please check it? The rate limit isn't being enforced per window right now, is it?

HyeockJinKim (changes requested):

src/ai/backend/manager/api/rest/ratelimit/handler.py
jopemachine: The original intention was to add this in a follow-up PR where limit is used, but I'll add it to this PR as well.

docs/manager/common-api/auth.rst
jopemachine: Reset was not originally present in the header.

This was an update to outdated information left over from the fixed-window rate limiter used in a very old implementation. The reason for the change is that the unit was changed from milliseconds to seconds.

src/ai/backend/common/clients/valkey_client/valkey_rate_limit/client.py
```diff
@@ -109,110 +95,83 @@ async def close(self) -> None:
...
-        :param key: The key to get.
-        :return: The rate limit data or None if not found.
+        :param user_id: The user the counter is keyed by.
+        :param rate_limit: The limit to fix for the window.
+        :return: The count, the limit fixed for the window and the seconds until it ends.
         """
+        key = f"user.{user_id}"
+        batch = Batch(is_atomic=True)
+        batch.hsetnx(key, "limit", str(rate_limit))
+        batch.hget(key, "count")
```
HyeockJinKim: Since this probably won't guarantee proper operation, is there a reason you're changing the implementation?

src/ai/backend/common/clients/valkey_client/valkey_rate_limit/client.py
```diff
@@ -40,25 +39,12 @@
...
-end
-redis.call('ZADD', key, now, tostring(request_id))
-redis.call('EXPIRE', key, window)
-return redis.call('ZCARD', key)
-"""
+@dataclass(frozen=True)
+class RateLimitState:
+    count: int
+    limit: int
+    reset: int
```
HyeockJinKim: The verb "reset" doesn't really resonate with me—wouldn't another name be better? It doesn't seem necessary to keep the header and name consistent within the internal values.

src/ai/backend/manager/api/rest/ratelimit/handler.py
```diff
@@ -50,21 +53,44 @@ async def rlim_middleware(
...
-        reserve_response_headers(request, RateLimitQuota(limit=1000, remaining=1000))
+        reserve_response_headers(
+            request,
+            RateLimitQuota(
+                limit=_ANONYMOUS_RATELIMIT,
+                remaining=_ANONYMOUS_RATELIMIT,
+                reset=_RATELIMIT_WINDOW,
+                window=_RATELIMIT_WINDOW,
+            ),
+        )
```
HyeockJinKim, line 93: ?? What is this behavior? I think we need to fix this right away. Unauthenticated requests require rate limiting based on IP address.

src/ai/backend/common/clients/valkey_client/valkey_rate_limit/client.py
```diff
@@ -109,110 +95,61 @@ async def close(self) -> None:
...
+    async def consume_rate_limit(self, user_id: UserID, window: int, limit: int) -> RateLimitState:
+        """
+        Consume one request of the user's current window and return its state.
+
+        :param user_id: The user the counter is keyed by.
+        :param window: The window length in seconds, applied when the request opens a window.
+        :param limit: The limit to fix for the window, taken only when the request opens one.
+        :return: The count, the limit fixed for the window, and the seconds until it ends.
+        """
+        key = f"user.{user_id}"
```
HyeockJinKim: This key is intended solely for rate limiting, but the name seems too generic. Please make it clear what action it represents based on the key value stored in the database.

src/ai/backend/manager/api/rest/ratelimit/handler.py
jopemachine: I will handle this in a follow up PR

HyeockJinKim (approved):
Please handle the IP-based unauthorized request rate limit as a follow-up task.

PR #14212, opened 2026-09-04
chore(BA-7488): fix model card creation broken by a resolver field shadowing its DTO field

fregataa:
## Summary

- `adminCreateModelCardV2` failed with `ModelCardGQL.__init__() got an unexpected keyword argument 'min_resource'`. The record was written and only the response payload failed to serialize, so the card was created but never returned — and `modelCardV2`, every model card search connection, and `VFolderGQL.modelCards` failed the same way.
- Cause: `ModelCardGQL.min_resource` was turned into a field resolver earlier in this cycle while `ModelCardNode.min_resource` stayed on the DTO. `from_pydantic` maps every DTO attribute whose name matches a dataclass field onto the constructor, and a resolver field is `init=False`. Restored as a DTO-backed field.
- The resolver's stated purpose — reading the requirements table on demand — never took effect, because the adapter fills `min_resource` unconditionally via `_nodes_with_min_resources()` on all six node-producing paths; the resolver only issued a second read of data already in hand. REST v2 returns the same DTO and depends on that field, so the DTO side stays.

Generated SDL keeps the same signature: `minResource: [ModelCardV2ResourceSlotEntry!]`.

## Test plan

Verified against a local manager over `./bai gql --v2` (direct API mode). REST v2 does not exercise this path — it returns the DTO as-is and was never broken.

- [x] The reported mutation succeeds: `adminCreateModelCardV2` returns the created `modelCard`, `errors: null`
- [x] The other paths through the same `from_pydantic` call work: `modelCardV2` (get) and `adminModelCardsV2` (search connection)
- [x] `minResource` carries real values end to end — a card created over REST with `cpu: 2` / `mem: 4294967296` reads back through GraphQL as `[{resourceType: "cpu", quantity: "2"}, {resourceType: "mem", quantity: "4294967296"}]`; a card with no requirements reads back as `null`, matching the resolver's old empty-list behavior
- [x] Negative control: restoring the resolver and restarting reproduces the reported error verbatim, including `path: ["adminCreateModelCardV2"]` and `code: backendai_generic_internal-error`
- [x] Under that failure the record is still written — the card appears in `model-card search` even though the mutation returned `null`, confirming creation succeeded and only the response failed
- [x] Full GQL schema imports cleanly (1265 types)
- [x] `pants fmt` / `pants fix` / `pants lint` / `pants check --changed-since=origin/main`

Not backported: both maintained release branches (26.8, 26.4) still carry `min_resource` as a plain field, so the buggy resolver never shipped.

Follow-up, not in this PR: `CreateModelCardV2Input` and `UpdateModelCardV2Input` expose no `minResource` field, so minimum resource requirements can only be set over REST v2 — the GraphQL clients cannot set them at all.

Resolves BA-7488

### DIFF
--- a/src/ai/backend/manager/api/gql/pydantic_compat.py
+++ b/src/ai/backend/manager/api/gql/pydantic_compat.py
@@ -60,55 +60,6 @@ from strawberry import UNSET
 from strawberry.relay import Node
 from strawberry.types.base import StrawberryObjectDefinition
 
-from ai.backend.manager.errors.api import ResolverFieldShadowsDTOFieldError
-
-
-def _dto_attribute_names(dto_cls: type[BaseModel]) -> frozenset[str]:
-    names = set(dto_cls.model_fields) | set(dto_cls.model_computed_fields)
-    for klass in dto_cls.__mro__:
-        for attr_name, attr in vars(klass).items():
-            if isinstance(attr, property):
-                names.add(attr_name)
-    return frozenset(names)
-
-
-def node_dto_type(cls: type) -> type[BaseModel] | None:
-    """The DTO a Node type is parameterized with, or ``None`` if it is not concrete."""
-    for klass in cls.__mro__:
-        for base in getattr(klass, "__orig_bases__", ()):
-            if get_origin(base) is not PydanticNodeMixin:
-                continue
-            args = get_args(base)
-            if args and isinstance(args[0], type) and issubclass(args[0], BaseModel):
-                return args[0]
-    return None
-
-
-def check_no_resolver_field_shadowing(cls: type, dto_cls: type[BaseModel]) -> None:
-    """Reject a GQL type whose resolver field names collide with the DTO's.
-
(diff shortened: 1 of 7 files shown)


src/ai/backend/manager/api/gql/pydantic_compat.py
```diff
@@ -60,6 +60,55 @@ class CreateFooInputGQL(PydanticInputMixin[CreateFooInputDTO]):
...
+    )
+    if not shadowed:
+        return
+    raise ResolverFieldShadowsDTOFieldError(
+        extra_msg=(
+            f"{cls.__name__} defines resolver field(s) {shadowed} that also exist on "
+            f"{dto_cls.__name__}. Either drop the resolver and let from_pydantic() carry "
+            f"the DTO value, or remove the field from the DTO."
+        )
+    )
```
HyeockJinKim: What is this used for?

HyeockJinKim (approved):

PR #14214, opened 2026-09-04
chore: add a `--recreate` option to refresh GQL schemas when restarting dev setup

achimnol:
## Summary
- `supergraph.graphql` and `gateway.config.ts` are mounted as docker configs, which are materialized at container creation time — `docker compose restart` reuses the container, so a refreshed schema never reached the gateway.
- Add `-r, --recreate` to `scripts/refresh-graphql-gateway.sh`, which reloads via `up -d --force-recreate --wait`. The default stays `restart` and now prints a note pointing at the flag.
- `/halfstack` and `/local-dev` skill docs use the recreate form in their schema-refresh examples; `scripts/README.md` documents the flag.

## Test plan
- [x] `./scripts/refresh-graphql-gateway.sh --recreate` equivalent verified manually: after `up -d --force-recreate --wait backendai-half-apollo-router`, the in-container `/gateway/supergraph.graphql` checksum matches the regenerated root file (plain `restart` and `up -d` left the old content).
- [x] `bash -n scripts/refresh-graphql-gateway.sh`
- [ ] `./scripts/refresh-graphql-gateway.sh -h` prints usage

https://claude.ai/code/session_01GhRvnrrskHeM6ybsbHmFhF

### DIFF
--- a/.claude/skills/halfstack/SKILL.md
+++ b/.claude/skills/halfstack/SKILL.md
@@ -211,9 +211,13 @@ The Hive Gateway serves the federated GraphQL schema. Regenerate when:
 cp docs/manager/graphql-reference/supergraph.graphql ./supergraph.graphql
 cp configs/graphql/gateway.config.ts ./gateway.config.ts
 
-# 3. Restart the gateway
-docker compose -f docker-compose.halfstack.current.yml restart backendai-half-apollo-router
+# 3. Recreate the gateway (docker configs are baked in at container creation,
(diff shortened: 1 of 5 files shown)


fregataa (commented):
This seems more like `feature` or `chore` rather than `fix`. How do you think?

achimnol:
Changed the title prefix from "fix:" to "chore:".

HyeockJinKim (approved):

PR #14216, opened 2026-09-04
feat(BA-7626): add UNDEFINED sentinel and drop it from update DTO dump and schema

### LINKED ISSUE #14167: Implement `UNDEFINED` handling
## Description

Define `UNDEFINED`, the value meaning "this field was not provided" in update DTOs, and add an update-only base `BaseUpdateRequestModel` that drops `UNDEFINED`-valued fields from serialization and the JSON schema. This reproduces what pydantic's `MISSING` does, in a form mypy accepts.

Why a separate base instead of `BaseRequestModel`: the convention applies to update only, and the other 1,300-odd DTOs (create, search, ...) must stay untouched.

## Tasks

- In `common/api_handlers.py`, define `Undefined` / `UNDEFINED` and keep `Sentinel = Undefined`, `SENTINEL = UNDEFINED` as deprecated aliases. No existing import site changes. The branching logic stays as is.
- Add `BaseUpdateRequestModel(BaseRequestModel)`. It excludes fields whose value is `UNDEFINED` from `model_dump()` and `model_json_schema()`.

## Out of scope

- Replacing `SENTINEL` imports across the codebase. Update DTOs switch in Story 2; the remaining sites and the alias removal are a separate cleanup.

## Acceptance Criteria

- An omitted field has no key in `model_dump()`. `None` serializes as `null`, a value as the value.
- `Undefined` does not appear in the OpenAPI schema.
- Dump and schema of other DTOs (create, etc.) are unchanged.
- `pants check` passes.

JIRA Issue: BA-7626

wngh4628:
resolves #14167 (BA-7626)

## Summary
- Define `Undefined` / `UNDEFINED` in `common/api_handlers.py` as the "field not provided" marker for update DTOs (`UNDEFINED` = no change, `None` = clear, value = update).
- `BaseRequestModel` drops `UNDEFINED`-valued fields from `model_dump()` / `model_dump_json()` and removes every `Undefined` trace from `model_json_schema()`, so the token never reaches the wire or the OpenAPI spec. The behavior is attached at the core-schema level so the serialization-mode JSON schema keeps its fields and a before/after/wrap `@model_validator` on a DTO does not disable it.
- `Undefined` validates only as the enum instance itself; a wire `1` / `true` / `1.0` on a `T | Undefined | None` field is rejected instead of being coerced into "not provided".
- Add `TriState.from_input()` / `OptionalState.from_input()` for converting a v2 update DTO field into the partial-update wrappers.

Existing DTOs are unaffected: no DTO uses `Undefined` yet, and dump / validation-schema output of all `common/dto/manager/v2` request models was compared before and after with no difference.

## Test plan
- [x] `pants fmt/fix/lint/check` on changed files pass
- [x] `pants test tests/unit/common/test_undefined.py tests/unit/manager/test_types.py tests/unit/common/test_api_handlers.py tests/unit/common/dto/manager/v2::` pass (116 targets)
- [x] Local manager restarted; admin login, domain create → update → get → delete → purge verified with `./bai`

https://claude.ai/code/session_01CQ8Hz7ixJ7AZNFkLUHqakB

### DIFF
--- a/changes/14216.feature.md
+++ b/changes/14216.feature.md
@@ -1,0 +1,1 @@
+Add the `UNDEFINED` sentinel for update DTOs: `BaseRequestModel` drops `UNDEFINED`-valued fields from `model_dump()` and from the OpenAPI schema, rejects wire values that would coerce into it, and `TriState.from_input()` / `OptionalState.from_input()` convert such fields into partial-update wrappers (BA-7626).
--- a/changes/14216.feature.md
+++ b/changes/14216.feature.md
@@ -1,1 +1,1 @@
-Add the `UNDEFINED` sentinel for update DTOs: `BaseRequestModel` drops `UNDEFINED`-valued fields from `model_dump()` and from the OpenAPI schema, rejects wire values that would coerce into it, and `TriState.from_input()` / `OptionalState.from_input()` convert such fields into partial-update wrappers (BA-7626).
+Add the `UNDEFINED` sentinel for update DTOs so that omitted fields are dropped from `model_dump()` and the OpenAPI schema instead of being sent as a value.
--- a/src/ai/backend/common/api_handlers.py
+++ b/src/ai/backend/common/api_handlers.py
@@ -16,13 +16,22 @@ from typing import (
     cast,
     get_args,
     get_origin,
+    override,
 )
 
 from aiohttp import web
 from aiohttp.web_urldispatcher import UrlMappingMatchInfo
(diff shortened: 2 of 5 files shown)


HyeockJinKim (changes requested):
Rather than implementing it this way, I think mypy will handle it anyway, so we can just manage the pydantic SENTINEL and MISSING errors by defining type aliases on our end and use those instead.

PR #14221, opened 2026-09-04
feat(BA-7653): add the storage volume and backend schema

fregataa:
## Summary

- Add the storage backend and volume tables BEP-1078 defines — `storage_backends`, `service_storage_backends`, `storage_volumes`, `service_storage_volumes`, `resource_group_storage_volumes` — plus the nullable `vfolders.storage_volume_id` reference. Additive only: no legacy column is dropped and nothing changes behaviour until the heartbeat and mount resolution work lands.
- A backend carries no status and a volume carries neither a path nor a status. Only the services that mount a volume can reach it, each over its own network path, so the observation lives on the service relationship and there is no aggregate verdict. Nothing is hard-deleted — a relationship no recent heartbeat names is marked `stale`, so a returning service reattaches.
- Built-in backend types are seeded when the schema is created, following the resource slot type precedent. `storage_volumes.is_default` names the volume a folder lands on when no host is given, held to at most one row by a partial unique index.
- The storage backend implementations the storage proxy registers become a `StorageBackendType` enum in `common`, which both the volume classes and the `storage_backends.type` column are typed by.

## Not in this PR

The session vfolder mount table is left out. Mounts are resolved per kernel group, not per session — `SchedulerRepository._resolve_vfolder_mounts_by_role` returns `dict[role, tuple[VFolderMount, ...]]`, and destination-path uniqueness is deliberately scoped per kernel (`MountNameValidationRule`), so two kernels may mount the same vfolder at the same path. A mount table therefore needs a kernel group to key on, and kernel groups still live only inside the `SessionRow.options` JSONB. Normalizing them is tracked as BA-7663; the mount table follows it.

Nothing reads the new tables yet: the heartbeat that registers backends and volumes, and the resolution that fills `vfolders.storage_volume_id`, are separate issues.

## Test plan

- [x] `alembic upgrade head` → `downgrade -1` → `upgrade head` against a populated local database, on the current `main` chain
- [x] Single alembic head after rebase
- [x] Live schema compared against the model metadata column by column, including the composite primary keys and the partial unique index on `storage_volumes.is_default`
- [x] A second `is_default = true` insert is rejected by `uq_storage_volumes_is_default`
- [x] Every manager model module imports cleanly; the storage backend registry still resolves each volume class by name
- [x] CI: `pants check` and `pants test`

Resolves BA-7653


src/ai/backend/common/data/storage/types.py
```diff
@@ -26,6 +28,29 @@ class NamedStorageTarget(BackendAISchema):
...
+    WEKA = "weka"
+    GPFS = "gpfs"
+    SPECTRUMSCALE = "spectrumscale"
+    DELLEMC_ONEFS = "dellemc-onefs"
+    VAST = "vast"
+    EXASCALER = "exascaler"
+    HAMMERSPACE = "hammerspace"
+    HAMMERSPACE_BASE = "hammerspace-base"
+    NOOP = "noop"
+
```
HyeockJinKim: 1. Shouldn't this be recorded in the database?
2. We're considering a way to add storage backends and other components on a per-plugin basis, so that we don't have to include them in the core system. Please refer to the resource slot type section.

src/ai/backend/manager/models/storage_backend/row.py
```diff
@@ -0,0 +1,102 @@
...
+    supports_fast_size: Mapped[bool] = mapped_column(
+        "supports_fast_size", sa.Boolean, nullable=False, server_default=sa.false()
+    )
+    # How long this may go without a fresh check before its status counts as stale.
+    status_stale_after: Mapped[timedelta] = mapped_column(
+        "status_stale_after",
+        sa.Interval,
+        nullable=False,
+        server_default=sa.text(f"'{int(DEFAULT_STATUS_STALE_AFTER.total_seconds())} seconds'"),
+    )
```
HyeockJinKim: More columns have appeared than the ones you mentioned. Could you please explain where they came from?
I believe that the values stored in the Manager by default are not intended for use by the storage proxy. If these values are included in the user metadata, we need to decide whether to display them based on the columns or using `pydanticColumn`.

src/ai/backend/manager/models/storage_volume/row.py
```diff
@@ -0,0 +1,130 @@
...
+        GUID(StorageBackendID),
+        sa.ForeignKey("storage_backends.id", ondelete="RESTRICT"),
+        nullable=False,
+    )
+    # At most one volume may be the default at a time, enforced by the partial unique
+    # index in ``__table_args__`` (a minimum of one is NOT guaranteed). To switch the
+    # default, clear the previous one before setting the new one in the same transaction.
+    is_default: Mapped[bool] = mapped_column(
+        "is_default", sa.Boolean, nullable=False, server_default=sa.false()
+    )
```
HyeockJinKim: Does "is_default" refer to the default setting for the storage backend?

src/ai/backend/manager/models/vfolder/row.py
```diff
@@ -303,6 +304,19 @@ class VFolderRow(LifecycleTimestampsMixin, Base):
...
+        "storage_volume_id",
+        GUID(StorageVolumeID),
+        sa.ForeignKey(
+            "storage_volumes.id",
+            ondelete="SET NULL",
+            use_alter=True,
+            name="fk_vfolders_storage_volume_id_storage_volumes",
+        ),
+        nullable=True,
+    )
```
HyeockJinKim, line 319: Logically, the storage volume in vfolder shouldn’t be nullable, so it does seem odd that it’s defined as nullable.
+ Since the volume is something that needs to be deleted when the vfolder is removed, please consider that scenario first before using `SET NULL`.
It appears that the choice between CASCADE and RESTRICT will depend on the specific behavior.

HyeockJinKim (changes requested):

src/ai/backend/manager/models/vfolder/row.py
fregataa: It is null only when the folder is an unmanaged one. I will add the constraint after data migration is done

PR #14271, opened 2026-09-05
fix: prevent concurrent "session ssh" runs from racing on the downloaded key

hoyajigi:
`backend.ai session ssh` cannot be run twice concurrently from the same working
directory. The second invocation dies with:

    FileNotFoundError: 2 No such file or directory
      File ".../client/cli/session/ssh.py", line 40, in container_ssh_ctx
        Path(key_filename).rename(key_path)

## Cause

`container_ssh_ctx()` shells out to `session download <session> id_container`
without `--dest`. `session download` writes the file under its container-side
basename into the destination directory, which defaults to `.`, so the key
lands at `./id_container` in the caller's cwd; the next line renames it from
that same relative path into `~/.ssh/id_<random>`.

The *destination* is already unique per invocation. The *source* is not. Two
invocations sharing a cwd interleave as: A downloads, B downloads (overwriting),
A renames, B renames a file that is no longer there.

Distinct `-p` ports do not avoid this — the port only picks the sshd proxy's
bind address, not the key path.

## Reproduction

    # one running session with the sshd service
    cd /tmp/anywhere
    backend.ai session ssh -p 9922 mysession &
    backend.ai session ssh -p 9923 mysession &
    wait
    # one of the two fails with FileNotFoundError on ssh.py:40

Observed with three agents driving one session from a single machine, each with
its own port.

## Changes

- Download the key into a per-invocation `tempfile.TemporaryDirectory()` passed
  as `--dest`, and rename from there. No two invocations share a source path.
- The temporary directory is created inside `~/.ssh` on purpose: `Path.rename()`
  is `os.rename()` and fails with `EXDEV` across filesystems, and `/tmp` is
  frequently a different filesystem from `$HOME`. This also avoids copying a
  private key across filesystems.
- Two side effects, both improvements: the key no longer touches the user's
  working directory, and a failed download no longer leaves `./id_container`
  behind (the `sys.exit()` now unwinds through the `with`).

## Tests

`tests/unit/client/cli/session/test_ssh.py`:

- concurrent invocations from one cwd — two threads held at a barrier between
  download and rename. Fails on `main` with the exact `FileNotFoundError` above,
  passes with this change.
- failed download leaves nothing in the cwd or in `~/.ssh`. Fails on `main`,
  passes with this change.

- [x] Backport targets: `26.8` and `26.4` both carry the identical code and both
      already support `session download --dest`, so the automatic `fix:` backport
      is correct and no `Backport:` line is needed.


hoyajigi:
포크에서 올라온 PR이라 `ci.yml`의 labeler job이 항상 403("Resource not accessible by integration")으로 실패했습니다. 포크발 `pull_request` 이벤트에서는 GitHub이 `GITHUB_TOKEN`을 워크플로의 `permissions: pull-requests: write` 선언과 무관하게 읽기 전용으로 강등하기 때문이라, 재실행으로는 풀리지 않습니다.

같은 커밋을 리포지터리 내부 브랜치로 옮겨 #14464로 다시 열었습니다. 코드 변경은 없고 news fragment 번호만 새 PR 번호로 바꿨습니다. 이어지는 리뷰는 #14464에서 부탁드립니다.

PR #14307, opened 2026-09-07
feat(BA-127): rate limit unauthenticated requests by client address

jopemachine:
## 📚 Stacked PRs

1. ✅ #13777 — `feat(BA-7364)`: deliver the login user id in the auth response
2. ✅ #13778 — `feat(BA-7365)`: key the rate limit counter by user id
3. ✅ #14211 — `feat(BA-7648)`: count requests in a per-user fixed window instead of a sorted set
4. 👉 **this PR — `feat(BA-127)`: rate limit unauthenticated requests by client address** ← you are here
5. #13771 — `feat(BA-7603)`: add a per-user rate limit middleware to the web server

⛔ #13779 — `feat(BA-7362): move the per-user API rate limit to the user resource policy` is closed (not planned); the limit stays on the keypair.

Addresses the IP-based half of BA-127, following up https://github.com/lablup/backend.ai/pull/14211#discussion_r3946538051.

## Summary

An unauthenticated request took a fixed `1000/1000` quota header and no window at all — the headers announced a limit while nothing counted the requests. `docs/manager/common-api/ratelimit.rst` has described address-based limiting for these routes all along; this makes the server do it.

- The caller's address keys a window of its own, `ip.<client_ip>.rate_limit_window`, counted and judged exactly as a user's is: 1000 requests per 15-minute window, then 429.
- The two paths now differ only in what identifies the caller, so `consume_user_rate_limit()` and `consume_ip_rate_limit()` share one atomic batch and differ in the key they build.

## `trusted-proxies` decides what the window keys on

Unchanged by this PR, but it decides how well the limiter holds:

| `trusted-proxies` | keyed by |
|---|---|
| configured | the first address outside the named proxies — a forged `X-Forwarded-For` changes nothing |
| empty (default) | the first `X-Forwarded-For` entry, else the peer — the caller writes that header, so rotating it draws a fresh window per request |

The webserver forwards `X-Forwarded-For` to the manager (`web/auth.py:121`), so a deployment that wants this limiter to hold must name its proxies. `configs/manager/sample.toml` already recommends it and marks the empty fallback deprecated. Closing that case also decides the `allowed_client_ip` check and the audit log record, so it needs its own change.

## Test plan

- [x] Unit — the window a caller is judged by, the quota headers, and the 429, each across both callers
- [x] Live — anonymous requests count down with `Reset` fixed, per-address windows isolate, the 1001st gets 429 with `Remaining: 0`, rejected requests still count, and an authenticated call opens `user.<uuid>.rate_limit_window` at 30000 without touching the IP window
- [x] Live with `trusted-proxies = ["127.0.0.1"]` — forwarded addresses key separate windows

https://claude.ai/code/session_01WEsyc7pNsff3fa9j73vS54

----
📚 Documentation preview 📚: https://sorna--14307.org.readthedocs.build/en/14307/

----
📚 Documentation preview 📚: https://sorna-ko--14307.org.readthedocs.build/ko/14307/

----
📚 Documentation preview 📚: https://sorna--14307.org.readthedocs.build/en/14307/

----
📚 Documentation preview 📚: https://sorna-ko--14307.org.readthedocs.build/ko/14307/


tests/unit/manager/api/test_ratelimit.py
```diff
@@ -67,67 +71,61 @@ def mock_handler(self) -> AsyncMock:
...
-        """Mock request carrying the rate limit the auth middleware injected."""
+    def authorized_caller(self, mock_valkey_client: MagicMock) -> MockCaller:
+        """A caller the auth middleware resolved, carrying the rate limit it injected."""
         request = make_mocked_request("GET", "/")
         request["is_authorized"] = True
         request["user"] = {"uuid": _USER_ID, "rate_limit": _RATE_LIMIT}
-        return request
+        return MockCaller(request=request, consumer=mock_valkey_client.consume_user_rate_limit)
 
-    async def test_anonymous_query_returns_default_headers(
```
jopemachine, line 84: Removed: this asserted the fixed `1000/1000` headers and that the client was never called, which is the behaviour this PR changes. An unauthenticated request now opens a window and counts against it, covered by `test_a_query_is_judged_by_its_own_window[anonymous_caller]` and the anonymous half of every other case.

tests/unit/manager/api/test_ratelimit.py
```diff
@@ -67,67 +71,61 @@ def mock_handler(self) -> AsyncMock:
...
-
-        # Valkey should not be called for anonymous requests
-        mock_valkey_client.consume_rate_limit.assert_not_called()
+    @pytest.fixture
+    def caller(self, request: pytest.FixtureRequest) -> MockCaller:
+        """The caller named by the parametrized fixture."""
+        caller: MockCaller = request.getfixturevalue(request.param)
+        return caller
 
-    async def test_authorized_query_counts_against_the_injected_limit(
```
jopemachine, line 106: Removed: this asserted the arguments passed to the client, which restated the line above it in the middleware and left the types to the type checker.

Replaced by `test_a_query_is_judged_by_its_own_window`, which gives the two windows different limits so the limit in the response says which one governed — an outcome rather than a call. Mutating the branch to always take the anonymous path fails it.

tests/unit/manager/api/test_ratelimit.py
```diff
@@ -143,113 +141,53 @@ async def test_authorized_query_counts_against_the_injected_limit(
                 expected_remaining="0",
                 description="exactly at limit",
             ),
-            RateLimitSuccessCase(
-                limit=0,
-                count=0,
-                expected_remaining="0",
-                description="zero limit before any request lands",
```
jopemachine, line 150: Removed: `consume_*` increments before it reads, so a window it returns holds at least one request. A count of nought never reaches the middleware, so this case described a state that cannot occur.

tests/unit/manager/api/test_ratelimit.py
```diff
@@ -143,113 +141,53 @@ async def test_authorized_query_counts_against_the_injected_limit(
...
         assert response.headers["X-RateLimit-Limit"] == str(case.limit)
         assert response.headers["X-RateLimit-Remaining"] == case.expected_remaining
         assert response.headers["X-RateLimit-Reset"] == str(_RESET_AFTER_SECONDS)
         assert response.headers["X-RateLimit-Window"] == str(_RATELIMIT_WINDOW_SECONDS)
+        mock_handler.assert_called_once_with(caller.request)
 
-        # Handler should be called
-        mock_handler.assert_called_once_with(mock_request_authorized)
-
-    async def test_the_limit_fixed_for_the_window_outranks_the_injected_one(
```
jopemachine, line 184: Removed: it claimed a limit changed after the window opened does not reach it, but it changed no limit — it stood a mock up to return one and read the header back.

That behaviour needs a window to change it on, and `test_the_limit_of_the_open_user_window_stands` / `test_the_limit_of_the_open_ip_window_stands` do it against a real valkey. What this did check — the header reports the limit the window holds, not the one asked for — the remaining cases already cover, since the anonymous caller asks for a limit none of them return.

tests/unit/manager/api/test_ratelimit.py
```diff
@@ -143,113 +141,53 @@ async def test_authorized_query_counts_against_the_injected_limit(
...
-        [
-            RateLimitExceedCase(
-                limit=_RATE_LIMIT,
-                count=_RATE_LIMIT + 1,
-                description="exceeds by 1",
-            ),
-            RateLimitExceedCase(
-                limit=_RATE_LIMIT,
-                count=50000,
-                description="far exceeds limit",
```
jopemachine, line 218: Removed: the middleware compares once, so a count one past the limit settles what a count far past it would. `exceeds by 1` remains.

tests/unit/manager/api/test_ratelimit.py
```diff
@@ -143,113 +141,53 @@ async def test_authorized_query_counts_against_the_injected_limit(
...
-            ),
-            RateLimitExceedCase(
-                limit=0,
-                count=1,
-                description="zero limit always exceeds",
```
jopemachine, line 223: Removed: with a limit of nought, a count of one is again one past the limit — the same relation `exceeds by 1` already covers. The middleware compares `count > limit` once and reads the magnitude of neither, so a zero limit takes the path every other limit takes.

src/ai/backend/common/clients/valkey_client/valkey_rate_limit/client.py
```diff
@@ -109,7 +112,26 @@ async def consume_rate_limit(
...
+        Consume one request of the address's current window and return its state.
+
+        Unauthenticated requests are counted here: they name no keypair, so the address
+        the server sees is the only caller identity there is.
+
+        :param client_ip: The address the counter is keyed by.
+        :param window_seconds: The window length, applied when the request opens a window.
+        :param limit: The limit to fix for the window, taken only when the request opens one.
+        :return: The count, the limit fixed for the window, and the seconds until it ends.
+        """
```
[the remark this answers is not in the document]
jopemachine: Done in 63b5acd8c8 — the paragraph on why unauthenticated requests are keyed by address is gone, leaving the contract and matching `consume_user_rate_limit()` beside it.

tests/unit/manager/api/test_ratelimit.py
```diff
@@ -67,66 +62,67 @@ def mock_handler(self) -> AsyncMock:
...
+        mock_valkey_client.consume_user_rate_limit.return_value = RateLimitState(
+            count=1, limit=_RATE_LIMIT, reset_after_seconds=_RESET_AFTER_SECONDS
+        )
+
         # Act
         response = await middleware(mock_request_anonymous, mock_handler)
         await apply_reserved_response_headers(mock_request_anonymous, response)
 
         # Assert
         assert response.headers["X-RateLimit-Limit"] == str(_ANONYMOUS_RATELIMIT)
```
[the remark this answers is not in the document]
jopemachine: Fixed in 63b5acd8c8, though not by asserting the arguments: the stand-in now answers by the address it is asked for, the way the client does, so a wrong address changes the limit in the response and fails the case. Verified by passing a constant address from the middleware — the case fails.

Argument assertions were taken out of this file earlier in review: they restate the line above them and leave the types to the type checker, so the wiring is pinned through what the caller receives instead.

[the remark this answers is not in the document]
jopemachine: Reverted in 0ac11b8800 — the stand-in answers with one state again, so this point stands open rather than fixed. Reopening the thread.

src/ai/backend/manager/api/rest/ratelimit/handler.py
```diff
@@ -54,43 +59,33 @@ async def rlim_middleware(
...
-                RateLimitQuota(
-                    limit=state.limit,
-                    remaining=state.limit - state.count,
-                    reset_after_seconds=state.reset_after_seconds,
-                    window_seconds=_RATELIMIT_WINDOW_SECONDS,
-                ),
+        else:
+            client_ip = current_client_ip()
+            if client_ip is None:
+                raise UnreachableError("a request over a socket always has a peer address")
```
seedspirit, line 72: Does rate limit always have to be based on IP? I'm wondering if it's correct to always treat requests as "unreachable" and fail them when the IP cannot be collected.

HyeockJinKim, line 72: Is it okay if this is "unreachable"? What should we do in a scenario where the client IP isn't collected?

HyeockJinKim (approved):
Since it doesn't seem like it'll break, I'll go ahead and merge it.

PR #14308, opened 2026-09-07
chore: revert the alembic empty-database replay fix (#14135)

seedspirit:
## Summary
- Reverts PR #14135 (`6a2a7696c5`) in full — the manager alembic chain replay fix.
- The change lives only on `main`: every touched migration file on `26.8.3` / `26.4.10` and on the `26.8` / `26.4` branch tips is byte-identical to the pre-#14135 state, so no released version is affected.
- No commit on `main` has touched those 8 migration files since the merge, so the revert restores them exactly to their prior content.

## Test plan
- [x] `git revert` applied without conflicts
- [x] Reverted tree diffs empty against `6a2a7696c5^` for all 8 migration files
- [ ] CI green


HyeockJinKim (dismissed):

changes/14308.misc.md
HyeockJinKim, line 1: Let's not add "fragment," shall we?

HyeockJinKim (approved):

PR #14311, opened 2026-09-07
fix(BA-6451): cancel the session when kernel resource allocation fails

fregataa:
## Summary

- The agent raised out of `create_kernel()` on a resource allocation error without emitting any lifecycle event. The manager saw only an RPC error, which `start_sessions` discards, so the session stayed in `CREATING` indefinitely with no reason recorded in `status_info`.
- Emit `KernelCancelledAnycastEvent`/`KernelCancelledBroadcastEvent` from the `ResourceError` handler in `create_kernel()`. The existing manager path (`handle_kernel_cancelled` → `mark_kernel_cancelled`) then cancels the kernel with the failure reason, frees its allocations and cancels the session.
- Allow cancelling a kernel in `PREPARED` in `update_kernel_status_cancelled`. A session reaches `CREATING` while its kernels are still `PREPARED`, and the previous status list made the cancel a no-op in exactly that window.

## Test plan

Verified against a local stack by injecting `FractionalResourceFragmented` at the allocation site (no GPU required).

- [x] Session reaches `CANCELLED` with `status_info = "All kernels cancelled"`; the kernel reaches `CANCELLED` with `status_info = "Resource allocation failed: FractionalResourceFragmented: ..."`, ~7s end to end instead of hanging
- [x] A normal session creation is unaffected and reaches `RUNNING`
- [x] `pants fmt` / `fix` / `lint` pass

## Known follow-up (not fixed here)

The cancel races `start_sessions`. Because `launcher.py` swallows the agent RPC error and reports success, the coordinator still applies its `success → CREATING` transition; in one run that landed ~50ms *after* the cancel and overwrote the terminal `CANCELLED`, letting the session drift to `RUNNING` → `TERMINATING` → `TERMINATED`. Two candidate fixes, both out of scope for this change:

- have `launcher.py` report failed sessions so `start_sessions` classifies them instead of reporting success
- guard session status transitions so a non-terminal status is never written over a terminal one

Resolves BA-6451

https://claude.ai/code/session_0128bKAiecc6WfVK9VBRjxBv

### DIFF
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -135,6 +135,7 @@ from ai.backend.common.events.event_types.agent.anycast import (
 )
 from ai.backend.common.events.event_types.kernel.anycast import (
     DoSyncKernelLogsEvent,
+    KernelCancelledAnycastEvent,
     KernelCreatingAnycastEvent,
     KernelPreparingAnycastEvent,
     KernelPullingAnycastEvent,
@@ -142,6 +143,7 @@ from ai.backend.common.events.event_types.kernel.anycast import (
     KernelTerminatedAnycastEvent,
 )
 from ai.backend.common.events.event_types.kernel.broadcast import (
+    KernelCancelledBroadcastEvent,
     KernelCreatingBroadcastEvent,
     KernelPreparingBroadcastEvent,
     KernelPullingBroadcastEvent,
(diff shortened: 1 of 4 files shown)


src/ai/backend/agent/agent.py
```diff
@@ -1218,6 +1220,25 @@ async def anycast_and_broadcast_event(
...
+                session_id=session_id,
+                reason=reason,
+            ),
+            KernelCancelledBroadcastEvent(
+                kernel_id=kernel_id,
+                session_id=session_id,
+                reason=reason,
+            ),
+        )
+
```
HyeockJinKim, line 1241: It seems like this might involve creating it on a different node, excluding the agent, or adjusting the GPU settings—should this be canceled?

fregataa:
We decided to add "creation failure" status and implement handlers

PR #14313, opened 2026-09-07
feat(BA-7721): filter runtime variant presets by runtime version

seedspirit:
## Summary

A preset is not valid across every version of the runtime it configures: a vLLM flag introduced in 0.9.0 should not be offered at 0.8.5, and one dropped in 0.11.0 should disappear once the user moves past it. Every preset was offered unconditionally.

- **The range.** `added_version` and `deprecated_version` on `runtime_variant_presets` form a half-open range, `added_version <= v < deprecated_version`. Both are NULL for all 57 seeded rows — a NULL bound is unbounded on that side — so a search without the new filter returns exactly what it did before and no backfill is needed.
- **The filter.** `runtime_version` on `RuntimeVariantPresetFilter`, shared by GraphQL and REST v2, composing with the existing AND/OR/NOT tree. The client picks the version; the server does not infer it from the image. Both v2 CLI search commands take `--runtime-version`, and either bound can be ordered by.
- **Compared in SQL, not in Python.** The filter is a DB-side `QueryCondition`, and Relay's pagination and `total_count` ride on the same query — filtering in application code would have made `has_next_page` and the count disagree with the rows returned.
- **Parsed on write.** Each bound is split into generated `major`/`minor`/`patch` columns, read from its leading numeric prefix and zero-filled. So a read compares integers instead of parsing text per row, `1`, `1.0` and `1.0.0` name one version, and `0.9.0rc1` compares as `0.9.0`. Nine digits per segment keeps one inside the int it is cast to; a longer run reads as nothing rather than being truncated to a different number.

## Where validation lives

The database takes whatever it is given. What has no numeric prefix leaves the segments NULL, and the filter tells that apart from an absent bound by testing the text column — so an unreadable bound matches no version rather than every one.

Requiring a version to be numeric is the API's job alone: a Pydantic `pattern` on create, update and the filter. That is the one constraint to lift the day prerelease versions need to be stored, and storage, filtering and ordering already handle them.

## Test plan

- [x] `pants lint` / `pants check` clean
- [x] 24 unit tests, 18 of them against a real Postgres:
  - inclusive lower bound, exclusive upper bound
  - a preset added at 0.9.0 is included at 0.10.0 — catches a regression to text comparison
  - `1`, `1.0` and `1.0.0` are one version; a range bounded by `1.0` and `1.2.3` has no gap
  - a suffixed bound compares on its numeric prefix; one with no numeric prefix matches no version
  - `total_count` / `has_next_page` follow the version-filtered set, on a page of one so the assertion can fail
  - create / update / filter DTOs reject a non-numeric version, a segment past the int, and a fourth segment
- [x] Migration `c8d5b2740f1e` upgrade and downgrade verified on a scratch database, including what each input parses to
- [x] Verified end to end against the local stack over GraphQL and the CLI: boundaries, pagination, and rejection at the API
- [ ] CI green

## Notes

`0.9.0rc1` and `0.9.0` compare equal — dropping the suffix means a prerelease cannot sort before its release. Expressing that would need a different ordering rule, not a wider pattern.

`deprecated_version` hides the preset rather than warning about it. The name follows the precedent in `ai/backend/common/meta/meta.py`.

The `OR` sub-filter in the adapter flattens a child's implicit AND. It predates this branch, is not specific to `runtime_version`, and the same block is copied across the other entity adapters, so it is left to one pass over all of them.

Resolves BA-7721

----
📚 Documentation preview 📚: https://sorna--14313.org.readthedocs.build/en/14313/

----
📚 Documentation preview 📚: https://sorna-ko--14313.org.readthedocs.build/ko/14313/

----
📚 Documentation preview 📚: https://sorna--14313.org.readthedocs.build/en/14313/

----
📚 Documentation preview 📚: https://sorna-ko--14313.org.readthedocs.build/ko/14313/


src/ai/backend/manager/models/runtime_variant_preset/conditions.py
```diff
@@ -20,6 +22,11 @@
 __all__ = ("RuntimeVariantPresetConditions",)
 
 
+def _as_int_array(expr: Any) -> sa.sql.expression.ColumnElement[Any]:
+    """Postgres orders int arrays element-wise then by length, so '0.9.0' < '0.10.0'."""
+    return sa.func.string_to_array(expr, ".").cast(postgresql.ARRAY(sa.Integer))
```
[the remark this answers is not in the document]
seedspirit: Confirmed. `2147483648` passed both the pattern and the CHECK while overflowing the `::int[]` cast, so a stored bound could have failed every version-filtered search. A segment now caps at nine digits — f6a29c4603.

docs/manager/rest-reference/openapi.json
```diff
@@ -25910,6 +25952,36 @@
             "description": "Toggle required flag; None = no change.",
             "title": "Required"
           },
+          "added_version": {
+            "anyOf": [
+              {
+                "type": "string"
```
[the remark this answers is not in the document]
seedspirit: Fixed in f6a29c4603. The constraint rides on the string member via `Annotated`, so create, update and filter all publish the pattern now.

src/ai/backend/common/dto/manager/v2/runtime_variant_preset/request.py
```diff
@@ -115,6 +130,14 @@ def validate_default_value(self) -> Self:
...
     name: StringFilter | None = Field(default=None)
     runtime_variant_id: UUIDFilter | None = Field(default=None)
+    runtime_version: str | None = Field(
+        default=None,
+        pattern=VERSION_PATTERN,
+        description=(
+            "Keep only presets valid at this runtime version "
+            "(added_version <= version < deprecated_version)."
+        ),
+    )
```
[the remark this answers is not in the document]
seedspirit: Added in 118ebdc27f. Both v2 search commands take `--runtime-version`.

src/ai/backend/manager/api/adapters/runtime_variant_preset/adapter.py
```diff
@@ -267,6 +269,10 @@ def _convert_filter(self, filter_: RuntimeVariantPresetFilter) -> list[QueryCond
             )
             if cond:
                 conditions.append(cond)
+        if filter_.runtime_version is not None:
+            conditions.append(
+                RuntimeVariantPresetConditions.by_valid_at_version(filter_.runtime_version)
+            )
```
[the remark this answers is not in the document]
seedspirit: Real, and it predates this branch: `OR: [{name: ..., runtimeVariantId: ...}]` is wrong on main the same way, and the same OR/NOT block is copied across the other entity adapters. Leaving it to one pass over all of them rather than this one file.

src/ai/backend/manager/models/runtime_variant_preset/conditions.py
```diff
@@ -109,6 +116,28 @@ def inner() -> sa.sql.expression.ColumnElement[bool]:
...
+
+        A NULL bound is unconstrained on that side.
+        """
+
+        def inner() -> sa.sql.expression.ColumnElement[bool]:
+            target = _as_int_array(sa.literal(version, sa.Text))
+            return sa.and_(
+                sa.or_(
+                    RuntimeVariantPresetRow.added_version.is_(None),
+                    _as_int_array(RuntimeVariantPresetRow.added_version) <= target,
```
HyeockJinKim: I'm a bit concerned about whether this will perform well. Also, it seems like this should be applied to "orders" as well, so please take a look at it.

seedspirit: Both addressed.

**Ordering** — added in e77c3de34a. You were right that it needed the same treatment; otherwise the text comparison this PR avoids would have come back through the ordering.

**Performance** — the bounds are now parsed on write into generated `major`/`minor`/`patch` columns, so a read compares integers instead of casting per row (0eea75fd83, 0af31e0727).

src/ai/backend/manager/models/runtime_variant_preset/version_expr.py
```diff
@@ -0,0 +1,19 @@
...
+from typing import Any
+
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+__all__ = ("as_int_array",)
+
+
+def as_int_array(expr: Any) -> sa.sql.expression.ColumnElement[Any]:
+    """Postgres orders int arrays element-wise then by length, so '0.9.0' < '0.10.0'."""
```
fregataa: Does this distinguish `1.0.0`, `0.10.0` and `10.0` ?

src/ai/backend/manager/models/runtime_variant_preset/row.py
```diff
@@ -59,6 +69,13 @@ class RuntimeVariantPresetRow(LifecycleTimestampsMixin, Base):
         "required", sa.Boolean, nullable=False, server_default=sa.false()
     )
 
+    # Half-open validity range: added_version <= runtime version < deprecated_version.
+    # NULL added_version reads as always present, NULL deprecated_version as not yet removed.
+    added_version: Mapped[str | None] = mapped_column("added_version", sa.Text, nullable=True)
+    deprecated_version: Mapped[str | None] = mapped_column(
+        "deprecated_version", sa.Text, nullable=True
+    )
```
HyeockJinKim: It seems better to extract values such as major, minor, and patch based on sa.Computed. In a scenario with fewer writes and more reads, the current implementation doesn't seem to perform very well in terms of read performance.

HyeockJinKim (changes requested):

src/ai/backend/common/dto/manager/v2/runtime_variant_preset/request.py
```diff
@@ -115,6 +136,14 @@ def validate_default_value(self) -> Self:
...
```
HyeockJinKim: Please create a `SemverVersion` type so that the string doesn't get corrupted when it's parsed at the point where the runtime version is retrieved. It just needs to be validatable, regardless of where it is. It's currently labeled `_VersionStr`, but it seems better to keep it as a public implementation. Please ensure consistency.

HyeockJinKim (approved):

PR #14320, opened 2026-09-07
fix: Keep project membership within a single domain

### LINKED ISSUE #14315: Project membership can be written across domains, breaking the same-domain invariant
## Summary

Three of the five paths that write a user's project membership enforce "a user's projects are in the user's domain". Two do not. The result is that `AssociationScopesEntitiesRow` can hold cross-domain rows, and `user { groups }` will return them.

Found while debugging a downstream consumer (Backend.AI FastTrack) that relies on this invariant to attribute a project to a domain.

## The invariant, where it holds

| Path | Enforcement |
| --- | --- |
| `modifyUser(group_ids:)` | `_get_project_scope_ids_for_user` filters `GroupRow.domain_name == domain_name` — `repositories/user/db_source/db_source.py:940-958` |
| `modifyGroup(user_update_mode: add)` | `_users_addable_to_project` filters `UserRow.domain_name == project_domain_subq` — `repositories/group/db_source/db_source.py:228-256` |
| signup | `project_id_by_name_in_domain(...)` — `services/auth/service.py:521-527` |

## Where it does not

**1. `adminAssignRole` / `adminBulkAssignRole` write membership with no domain check.**

`PermissionControllerService.assign_role` binds the user to the project as a side effect of granting a project-scoped role:

- `services/permission_contoller/service.py:253-256` (and `:287-292` for the bulk variant) call `group_repository.bind_user_to_project(user_id, project_id)`.
- `bind_user_to_project` (`repositories/group/db_source/db_source.py:721-733`) calls `add_entity_members` with no filter — unlike its two siblings in the same file.
- `add_entity_members` (`repositories/ops/rbac/provider.py:665-678`) inserts `AssociationScopesEntitiesRow(scope_type=PROJECT, entity_type=USER, relation_type=AUTO)`.
- Nothing validates the domain above it either: `api/adapters/rbac/adapter.py:997-1010` passes `input.project_id` straight through, and the resolvers (`api/gql/rbac/resolver/role.py:313,339`) only call `check_admin_only()`.

So a superadmin granting user U (domain `acme`) a project-scoped role on project P (domain `beta`) makes U a member of P.

**2. A domain move does not re-sync existing memberships.**

`repositories/user/db_source/db_source.py:531-535` (and `:641-645`) only re-sync when `group_ids` is supplied:

```python
group_ids = updater_spec.group_ids_value
if group_ids is not None:
    await self._update_user_groups(session, updated_user.uuid, updated_user.domain_name, group_ids)
```

`UserUpdaterSpec.domain_name` and `.group_ids` are independent (`repositories/user/updaters.py:30,41`), and `ModifyUserInput` exposes them independently (`api/gql_legacy/user.py:1021,1023`). `modifyUser(email:…, domain_name:"beta")` alone moves the user and leaves every `acme` project association in place.

## Why it matters

`user { groups }` reads `AssociationScopesEntitiesRow` with no `relation_type` or domain filter (`api/gql_legacy/user.py:532-561`), so both cases surface as a user apparently belonging to another domain's project. Any consumer that treats project membership as domain-scoped — including core's own domain-scoped admin views — will disagree with the data.

## Suggested direction

The first question is which way the invariant should go, and that is a maintainer call:

- **If the invariant is real**, `bind_user_to_project` should carry the same `UserRow.domain_name == project_domain_subq` filter its siblings do, and a domain change should re-sync memberships (or reject the move while cross-domain memberships exist).
- **If cross-domain project membership is intentional for RBAC**, then the other three paths are over-restrictive and the inconsistency should be resolved the other way — in which case it would help downstream if the constraint were documented.

Either way the two sites above should agree with the other three. Happy to send a PR once the intended direction is settled.

rapsealk:
Resolves #14315.

## Problem

Three of the five paths that write a user's project membership enforced "a user's projects are in the user's domain". Two did not, so `AssociationScopesEntitiesRow` could hold cross-domain rows and `user { groups }` returned them — breaking any consumer (including core's own domain-scoped admin views) that treats project membership as domain-scoped.

| Path | Before |
| --- | --- |
| `modifyUser(group_ids:)` | filters on the user's domain |
| `modifyGroup(user_update_mode: add)` | filters on the project's domain |
| signup | resolves the project by name **in the domain** |
| **`adminAssignRole` / `adminBulkAssignRole`** | **no domain check** |
| **`modifyUser(domain_name:)`** | **no re-sync of existing memberships** |

## What changed

**1. `ProjectDBSource.bind_user_to_project` refuses a cross-domain user.**

This is the membership write that `PermissionControllerService.assign_role` / `bulk_assign_role` perform as a side effect of granting a project-scoped role. It now applies the same `UserRow.domain_name == project_domain_subq` constraint its sibling `_users_addable_to_project` already applies, following the shape of the `_refuse_personal_project` guard sitting right above it. Raising rather than silently skipping means the role assignment that follows does not happen either, so no user ends up holding a role in a project it is not a member of. Both the single and bulk callers route through this one function.

**2. A domain move re-syncs the user's project memberships.**

`UserDBSource._update_single_user_validated` re-synced only when `group_ids` was supplied, but `domain_name` and `group_ids` are independent inputs — `modifyUser(email:…, domain_name:"beta")` alone moved the user and left every `acme` project association in place. The re-sync now also runs on a domain move, so the user leaves the projects it can no longer belong to. This resolves the invariant question in the issue in the direction the majority of paths already took: **project membership is domain-scoped.**

**3. Adjacent, in the same lines: the move now carries `domain_id`.**

`UserUpdater.build_values()` writes only `domain_name` (marked *deprecated: use `domain_id`* on `UserRow`), so a domain move left `domain_id` pointing at the old domain — the user was in two domains at once by different columns, which would have made the re-sync above only half-correct. The existence check is folded into the id lookup, so this costs no extra query and `_check_domain_exists` (no other callers) is gone.

## Tests

- `tests/unit/manager/repositories/permission_controller/test_role_assignment.py` — `bind_user_to_project` refuses a user from another domain and writes no membership row.
- `tests/unit/manager/repositories/user/test_user_domain_move.py` (new) — moving a user leaves its origin-domain projects; the move carries `domain_id` over.

All three were confirmed to fail against `main` and pass with this change. Related suites (`repositories/{group,permission_controller,user,domain}`, `repositories/ops/v2/user`, `services/permission_controller`, `services/test_users.py`, `models/user`) are green, as are `pants fmt`/`fix`/`lint`/`check`.

## Known remaining

A user's **personal** project stays in the origin domain across a move: `replace_user_projects` deliberately excludes personal projects from both sides of the diff, and moving the project row itself raises separate questions (name collisions in the destination domain, resources it holds). Left alone rather than guessed at.

## Relation to BEP-1076 / BEP-1077 — @HyeockJinKim please weigh in

Neither draft retires the invariant this PR enforces; both restate it. BEP-1077 §5.1: *"Projects and users are created in their domain"*, *"A user's virtual entity is governed by its domain only"*, and in the summary table *"direct registration is a domain-admin operation (BEP-1076)"*. BEP-1076: *"a domain is not invited into. A user is created under a domain, and moving between domains is an edit of the user, not a membership operation."* So the direction this PR picks is the one the plan already assumes.

Where the three changes sit against the plan:

| Change | Against the plan |
| --- | --- |
| **1.** cross-domain guard on `bind_user_to_project` | The mechanism is slated for rewrite — BEP-1076 splits membership into relation + role grant and undoes the `ScopeUserMember` fusion; BEP-1077 makes a roster entry a `replace_share_fields(project, user, …)`. The **constraint** survives that rewrite, so this is a guard on code that is wrong today and will be re-expressed, not contradicted. Does not touch the personal-project path, which joins through `ops/v2/user/write.py::_join` rather than `ProjectDBSource`. |
| **3.** carry `domain_id` on a move | Anticipated by BEP-1076's own open question — *"the identifier becomes `domain_id` (resolved from a name by a lookup at the edge)"*. This resolves it in the db_source rather than at the edge; same outcome, one layer lower. |
| **2.** re-sync memberships on a domain move | **This is the one that needs a call.** See below. |

### Change 2 answers a question BEP-1076 lists as open

> `UserUpdater.domain_name` stays an edit of the user, but it is a move: the identifier becomes `domain_id` …, and what else the move has to carry — the graph membership, the projects of the old domain, **the domain copied onto sessions and vfolders** — is not settled here

This PR answers two thirds of that (graph membership, the projects of the old domain) and leaves the third untouched. BEP-1076 also lists *"Take `group_ids` off `UserUpdater`"* under Migration — the gate this PR widens is scheduled for removal.

And the "known remaining" caveat above gets sharper under BEP-1077, not softer: §5.1 puts **every individually-owned resource entity in the personal project**, and personal projects are excluded from the membership diff on both sides. Today a personal project left behind in the origin domain is a loose end; once personal folders and sessions hang off it, it is the user's whole resource footprint sitting in the domain they left.

**So, two options, and it is your call:**

1. **Keep change 2 as an interim fix**, with this note standing so the partial answer is on the record when BEP-1076 settles the question properly.
2. **Drop change 2 from this PR** and take it to BEP-1076 as the answer to that open question, covering sessions, vfolders and the personal project in one go.

Changes 1 and 3 stand either way. I lean toward (1) — the current behavior is plainly broken, and re-syncing is the same semantics `modifyUser(group_ids: [])` already has — but pre-committing a design decision in a bugfix PR is exactly the kind of thing worth asking about first.

### One thing neither BEP appears to settle

BEP-1076's invitation branch is addressed **by email to someone who need not have an account**. If that email already belongs to a user in another domain, which rule wins — the invitation, or the same-domain constraint? Might be worth an Open Question entry.

https://claude.ai/code/session_01HV6Q8x9RvpfCNREMvQSDme


HyeockJinKim (changes requested):
We are currently migrating all scope operations to the virtual entity and applying them to all parts that haven't been updated yet. There are some parts that aren't working correctly right now, and even this current implementation is flawed.
If any issues arise after today's migration related to the virtual entity, please let us know at that time.

PR #14322, opened 2026-09-07
refactor(BA-7708): judge each model card of a bulk delete on its own

HyeockJinKim:
## Summary
- `bulk_delete_model_card` was wired to `global_scope`, so the super-admin check was the only gate it passed and no individual card was judged. It is now a partial bulk action whose run names the cards it deletes, and the bulk RBAC validator asks for the permission on every one of them.
- Both model card delete actions declare `PURGE`, the operation they actually perform — the row is removed rather than marked deleted, which is also what the `delete_associated_vfolder` cascade acts on.
- A card's failure is carried as the exception it was raised as instead of a flattened string, so the audit row keeps its error code. The GraphQL payload still reports the message text.
- The route, the `admin_bulk_delete_model_cards_v2` super-admin contract and the `successes` / `failed` payload are unchanged.

## Test plan
- [ ] Bulk-deleting several model cards as a super admin still reports every deleted id in `successes`
- [ ] A batch mixing a deletable card with a missing one, and with one whose VFolder is mounted, still reports the rest as deleted and the failure in `failed` with its message
- [ ] `delete_associated_vfolder: true` still moves the VFolder to trash and purges the sibling cards on it
- [ ] Deleting one card through the single-card mutation is unchanged

Resolves BA-7708

https://claude.ai/code/session_01HnDWcwrgQsmcAQbAd56kfh

### DIFF
--- a/src/ai/backend/manager/services/model_card/service.py
+++ b/src/ai/backend/manager/services/model_card/service.py
@@ -4,11 +4,18 @@ from uuid import UUID
 from ruamel.yaml import YAML
 
 from ai.backend.common.config import ModelDefinition
+from ai.backend.common.data.entity.model_card import ModelCardID
 from ai.backend.common.types import VFolderID
 from ai.backend.logging.utils import BraceStyleAdapter
+from ai.backend.manager.actions.v2.bulk.result import (
+    PartialBulkEntityResult,
+    PartialBulkResult,
+)
 from ai.backend.manager.clients.storage_proxy.session_manager import StorageSessionManager
 from ai.backend.manager.data.model_card.types import ResourceRequirementEntry, VFolderScanData
+from ai.backend.manager.errors.resource import ModelCardNotFound
 from ai.backend.manager.errors.storage import ModelCardParseError
+from ai.backend.manager.models.model_card.purgers import ModelCardPurger
 from ai.backend.manager.models.model_card.upserters import ModelCardScanUpserter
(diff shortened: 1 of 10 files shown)


src/ai/backend/manager/services/model_card/service.py
```diff
@@ -61,9 +67,21 @@ async def delete(self, action: DeleteModelCardAction) -> DeleteModelCardActionRe
...
-        data = await self._repository.bulk_delete(action.purgers, action.options)
-        return BulkDeleteModelCardActionResult(data=data)
+    ) -> PartialBulkResult[ModelCardID]:
+        purgers = [ModelCardPurger(card_id=card_id) for card_id in action.ids]
+        data = await self._repository.bulk_delete(purgers, action.options)
+        deleted = {ModelCardID(card_id) for card_id in data.successes}
+        errors = {ModelCardID(failure.card_id): failure.error for failure in data.failures}
+        return PartialBulkResult(
+            items=[
+                PartialBulkEntityResult[ModelCardID].succeeded(card_id, card_id)
```
seedspirit, line 77: Is it right to put PartialBulkEntityResult value with `card_id`?

HyeockJinKim: This is the part where values are sent; I'm only passing the card ID right now because I'm not actually reading any values.
Since we've decided to use the `PartialBulkEntityResult` object for all bulk operations, this is the correct behavior.

seedspirit (approved):

PR #14324, opened 2026-09-07
refactor(BA-7698): add the v2 specs for roles and their permission fields

HyeockJinKim:
Roles become an entity of the v2 lineage and permissions a field row of the role holding them, so BA-7699 / BA-7700 / BA-7701 can wire the role actions without the `permission_controller` creators.

## Specs

| Package | Specs |
|---|---|
| `models/rbac_models/role/` | `RoleCreator` (created in its scope), `RoleQuerier`, `RoleSearcher`, `RoleUpdater` / `RoleSoftDeleteUpdater` / `RoleRestoreUpdater`, `RolePurger` |
| `models/rbac_models/permission/` | `RolePermissionCreator` (field of the role), `RolePermissionPurger`, `RolePermissionOwnerLookup`, `RolePermissionSearcher` |

Field-scoped permission entries stay on the existing `set` / `widen` / `revoke` ops: `permission_fields` has no id column of its own, so it cannot be a field row of the role.

`PermissionID` becomes a `FieldIdentifier`, `RoleData` an `EntityData` and `PermissionData` a `FieldData`.

## Scope reads follow the graph

`ScopedRoleOperationScope` and `RoleConditions.by_mapped_scope` read `scope -> virtual entity -> entity` through the new `owning_scope_exists` / `OwningScopeConditions`. Neither reads `association_scopes_entities` any more, per the rule already stated in `models/virtual_entity/queries.py`.

**A role created before the v2 write path has no virtual entity node and is absent from a scope read until BA-7571 backfills one.** No migration here by design; the legacy role-creation entry point (`ops/rbac/provider.py::create_scope`) has no production caller, so newly created roles are unaffected.

Scope types and ids are now the open `EntityType` / `EntityID` / `EntityIdentifier` rather than `RBACElementType` and stringified uuids, so the deprecated `to_scope_type()` / `to_entity_type()` bridges are gone from these paths.

## Preset roles

`_create_preset_roles` uses `_created_in` like every other entity create, so a scope governs the roles it provisions as well as owning them. `_own` had no caller left and is removed; `models/specs/AGENTS.md` and `KNOWLEDGE.md` are updated in the same change.

## Verification

`pants fmt` / `fix` / `lint` clean; `pants check --changed-dependents=transitive` clean over 4,656 files; `pants test --changed-dependents=direct` 68/68.

New spec unit tests under `tests/unit/manager/models/rbac_models/`. The three fixtures that registered a role through the association table now seed a virtual entity node and an ownership edge instead.

https://claude.ai/code/session_01WVQz8KSLQ9RdP3v5cU599y


src/ai/backend/manager/models/rbac_models/role/creators.py
```diff
@@ -0,0 +1,56 @@
...
+    The scope owns and governs the new role, the same pair every other entity's
+    ``created_in`` writes; the roles a preset instantiates take the same path.
+    """
+
+    name: str
+    scope: EntityIdentifier
+    source: RoleSource = RoleSource.CUSTOM
+    status: RoleStatus = RoleStatus.ACTIVE
+    description: str | None = None
+    auto_assign: bool = False
```
seedspirit, line 30: Is there reason you set default value in Creator object?

HyeockJinKim: I'll fix this later. The workflow of moving items from "Creator" to "Row" is correct, but since there are still so many items in "Creator," I couldn't get the flow right while working on this task.

PR #14326, opened 2026-09-07
fix(BA-7726): grant every user's self role the keypair permissions it owns

fregataa:
## Summary

- A user created after the keypair RBAC migration (`21159a293dfb`) gets a self role covering only the resource entity types, which leave out `keypair`. It therefore holds no permission on its own keypairs and is refused on `/auth/ssh-keypair` with `NotEnoughPermission(403)`. Neither the user-scope permission presets nor the project member roles cover the gap.
- Backfills the missing permission rows at both maintained release heads, together with the keypair AUTO edges a keypair created outside the RBAC creator never got, so the scope chain resolves a keypair to its owner. Both migrations are idempotent via `ON CONFLICT DO NOTHING` and leave `downgrade()` empty.
- No repeat follows on the main head. There a keypair is a field of its owner (`KeyPairID` is a `FieldIdentifier`), so the permission an operation on it reads is the one on the `user` entity, which the self role already holds — the rows would be inert.
- The roles fixture gains the same data: the keypair AUTO edges the runtime creator writes, and `keypair` on the user-scope permission preset.

## Migration chain

```
… → bd1bf0524350 → e1b7f3d20a94 → 2d6443ac0d4a → …     (26.4 head)
… → b93d1c47af52 → f4a2b7c9e105 → 857c4d02c4b9 → …     (26.8 head)
… → b4c2e7f19a30                                        (main head, unchanged)
```

Both splice points sit before `5c1e7a2d9b40` drops `permissions.operation` and before `b44c240c7680` adds `all_fields`, and after `c6648c039bd4` adds `permissions.permission`, so both migrations write the release-branch column set.

## Backport notes

Default `fix:` backporting already targets exactly `26.8` and `26.4`, so no `Backport:` trailer is set.

- **26.8** cherry-picks cleanly: `bd1bf0524350 → e1b7f3d20a94 → 2d6443ac0d4a → … → b93d1c47af52 → f4a2b7c9e105`, single head, the second migration a no-op after the first.
- **26.4** will conflict — `2d6443ac0d4a`, `857c4d02c4b9` and `b93d1c47af52` are all past its head. Keep `e1b7f3d20a94` and the fixture change; drop `f4a2b7c9e105` and both `down_revision` repoints.

## Test plan

- [x] Ran both statements against a local halfstack DB inside a transaction, rolled back: 26.x variant produced 200 user-scope keypair rows (20 self roles × 10 operations), main-schema variant 100 (20 × 5 bits); rerun of each was `INSERT 0 0`.
- [x] The ASE backfill inserted 3 missing keypair edges on that DB — keypairs issued after `21159a293dfb` that had no edge.
- [x] Alembic graph resolves to a single head with no dangling `down_revision`; column availability verified at both splice positions.
- [x] `pants fmt` / `fix` / `lint` / `check` on every touched file.
- [x] CI migration replay on a fresh database.

Resolves BA-7726

https://claude.ai/code/session_01JN3jEjobnMrFzdAjrtnPz5

### DIFF
--- a/src/ai/backend/manager/models/alembic/versions/e1b7f3d20a94_grant_self_roles_keypair_permissions.py
+++ b/src/ai/backend/manager/models/alembic/versions/e1b7f3d20a94_grant_self_roles_keypair_permissions.py
@@ -1,0 +1,74 @@
+"""grant every user's self role the permissions on the keypairs it owns
+
+The self role a user creation provisions covers only the resource entity types, which
+leave out ``keypair``, so a user created after 21159a293dfb holds no permission on its
+own keypairs and is refused on ``/auth/ssh-keypair``. Backfills those rows, together
+with the keypair AUTO edges a keypair created outside the RBAC creator never got, so
+the scope chain resolves a keypair to its owner.
+
+Spliced in at the 26.4 release head; f4a2b7c9e105 repeats it at the 26.8 one. No repeat
+follows on the main head: there a keypair is a field of its owner, so the permission an
+operation on it reads is the one on the ``user`` entity, which the self role already
+holds. Idempotent via ``ON CONFLICT DO NOTHING``.
+
+Revision ID: e1b7f3d20a94
+Revises: bd1bf0524350
+Create Date: 2026-09-07 22:30:00.000000
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "e1b7f3d20a94"
+down_revision = "bd1bf0524350"
+# Part of: NEXT_RELEASE_VERSION
+branch_labels = None
+depends_on = None
+
+# A role its holder carries on that same holder's user scope: the self role. Derived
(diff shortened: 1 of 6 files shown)


src/ai/backend/manager/models/alembic/versions/e1b7f3d20a94_grant_self_roles_keypair_permissions.py
```diff
@@ -0,0 +1,74 @@
...
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "e1b7f3d20a94"
+down_revision = "bd1bf0524350"
+# Part of: NEXT_RELEASE_VERSION
```
HyeockJinKim: Is it correct to include "NEXT_RELEASE_VERSION"? It appears this will be filled in differently for each version.

HyeockJinKim (approved):

PR #14327, opened 2026-09-07
fix(BA-6697): report GPU allocation as a per-device ratio

fregataa:
## Summary

- The agent summed the raw per-slot allocations recorded in each kernel's `resource.txt` without dividing them by the device capacity. With a fractional GPU (`cuda.shares`), one physical device holds ~13.5 shares, so several sessions sharing a device reported a total of 2, 3, … instead of a 0–1 ratio.
- `scan_gpu_alloc_map()` now normalizes every allocation by the capacity the compute plugin registered for that device (`AbstractAllocMap.device_slots[...].amount`), and the RPC handler passes that capacity map in via the new `collect_device_capacities()`. `resource.txt` is only read, never rewritten.
- The hardcoded `cuda.shares` / `cuda.device` branches are replaced by an iteration over every slot under the `cuda` device, each divided by its own capacity. This also fixes the latent unit-mixing bug where a share count and a device count were added into one number, and it picks up MIG-specific slots for free.

### Notes for review

- **Value semantics change.** `gpu_alloc_map` goes from an absolute share amount to a 0–1 ratio. The manager resolver is a pass-through, so no manager change is needed, but during a rolling upgrade older agents keep reporting the old absolute values on the same field.
- **Consumer-side display** lives in `lablup/backend.ai-webui` (`AgentResources.tsx` renders the number verbatim) and in the control panel. Those need a matching percentage-formatting change; they are **not** in this PR.
- When a device is no longer present in the alloc map (removed or masked) the capacity is unknown, so the raw amount is reported and a warning is logged rather than dropping the device from the map.

Resolves BA-6697

https://claude.ai/code/session_01QfgPQeoBrYhq68TpQCXJw1

### DIFF
--- a/src/ai/backend/agent/resources.py
+++ b/src/ai/backend/agent/resources.py
@@ -930,11 +930,11 @@ def collect_device_capacities(
     Collect the allocatable capacity of each device in its own slot unit,
     which is the denominator to normalize recorded allocations into ratios.
     """
-    return {
-        (slot_info.slot_name, device_id): slot_info.amount
-        for computer_ctx in computers.values()
-        for device_id, slot_info in computer_ctx.alloc_map.device_slots.items()
-    }
+    capacities: dict[tuple[SlotName, DeviceId], Decimal] = {}
+    for computer_ctx in computers.values():
+        for device_id, slot_info in computer_ctx.alloc_map.device_slots.items():
+            capacities[(slot_info.slot_name, device_id)] = slot_info.amount
+    return capacities
 
 
 def _normalize_device_alloc(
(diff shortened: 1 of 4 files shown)


src/ai/backend/agent/resources.py
```diff
@@ -921,11 +923,52 @@ async def _wrap_future(fut: asyncio.Future[Any]) -> None:
...
+    computers: Mapping[DeviceName, ComputerContext],
+) -> DeviceCapacityMap:
+    """
+    Collect the allocatable capacity of each device in its own slot unit,
+    which is the denominator to normalize recorded allocations into ratios.
+    """
+    return {
+        (slot_info.slot_name, device_id): slot_info.amount
+        for computer_ctx in computers.values()
+        for device_id, slot_info in computer_ctx.alloc_map.device_slots.items()
```
seedspirit: Nitpick: If you put the nested for loop on a single line, it seems to reduce readability. Could you break it out a bit more?

src/ai/backend/agent/resources.py
```diff
@@ -921,11 +923,52 @@ async def _wrap_future(fut: asyncio.Future[Any]) -> None:
...
+    if capacity is None or capacity <= 0:
+        # The device is no longer in the alloc map (e.g. removed or masked),
+        # so the ratio is unknown and the raw amount is reported as-is.
+        log.warning(
+            "scan_gpu_alloc_map(): no capacity for {}:{} (capacity:{!r}), "
+            "reporting the raw allocation {}",
+            slot_name,
+            device_id,
+            capacity,
+            alloc,
```
seedspirit: for structured logging, how about injecting slot_name, device_id etc to last rather than in the middle?

HyeockJinKim (approved):

PR #14348, opened 2026-09-07
refactor(BA-7704): write the idle checker bindings as a relation

HyeockJinKim:
## Summary
- The idle checker binding links a scope to an idle checker, so its writes now take the relation shape: create, enable, disable and purge run through the relation group and answer for both sides. The id the API names is resolved into the pair by a lookup; the reads name the checker the relation reaches.
- `RelationCreator` gains `precondition_checks`, run before the insert, so the scope side of a binding is refused by the spec when its row is missing.
- Drops the legacy single-entity and bulk RBAC validators, the `RBACEntityCreator` / `Purger` / `Upserter` set, and the scope creation family on the RBAC write ops, none of which had callers left.

## Test plan
- [ ] `tests/unit/manager/repositories/idle_checker/test_assignment_crud.py`: link, switch, unlink, and the graph edges each writes
- [ ] `tests/component/idle_checker_assignment/test_scoped_search_permissions.py`: superadmin, scope reader, both-sided switch permission, denied cases
- [ ] `tests/unit/manager/actions/test_registry_catalog.py`: the assignment wiring is in the catalog
- [ ] `tests/unit/manager/repositories/ops/rbac/test_provider.py`: membership writes still pass after the removal

Resolves BA-7704

https://claude.ai/code/session_01XfXv86ECU1Um7Uzd1897Zh


src/ai/backend/manager/models/alembic/versions/e3fb172166dc_backfill_project_member_roles.py
```diff
@@ -118,7 +118,7 @@ def _get_permissions_table() -> sa.Table:
 
 
 def _member_role_name(project_id: uuid.UUID) -> str:
-    """Match ProjectMemberRoleSpec.role_name() used by the runtime create path."""
+    """The member role name the runtime create path produced at the time."""
     return f"project-{str(project_id)[:8]}-member"
```
HyeockJinKim, line 122: Removed implementation details described in the comments.

PR #4103, opened 2025-04-07
feat(BA-1099): Align action architecture code

HyeockJinKim:
resolves #4102 (BA-1099)

**Checklist:** (if applicable)

- [ ] Milestone metadata specifying the target backport version
- [ ] Mention to the original issue
- [ ] Installer updates including:
  - Fixtures for db schema changes
  - New mandatory config options
- [ ] Update of end-to-end CLI integration tests in `ai.backend.test`
- [ ] API server-client counterparts (e.g., manager API -> client SDK)
- [ ] Test case(s) to:
  - Demonstrate the difference of before/after
  - Demonstrate the flow of abstract/conceptual models with a concrete implementation
- [ ] Documentation
  - Contents in the `docs` directory
  - docstrings in public interfaces and type annotations

PR #14367, opened 2026-09-08
feat(BA-7733): add delete_shared_vfolders to the user purge APIs

fregataa:
## Summary

- `purge_shared_vfolders` on the user purge APIs does the opposite of its name: since 25.6.0, `true` keeps the purged user's shared virtual folders (migrating them to the requesting admin) and `false` deletes them. The inversion came from `3c97066da2` (BA-1099, #4103), which dropped a `not` while realigning the action architecture; no changelog acknowledged a behavior change.
- Rather than flipping the polarity under existing callers, this deprecates `purge_shared_vfolders`, **ignores its value**, and adds `delete_shared_vfolders`, whose name matches what it does: `true` deletes the shared folders, `false`/unset transfers their ownership to the requesting admin.
- Callers that sent `purge_shared_vfolders=true` keep the behavior they had. Callers that sent `false` now preserve the folders instead of destroying them, so the only behavior change moves away from irreversible deletion.
- The action field is a plain `bool` instead of `OptionalState[bool]`. The `if x else OptionalState.nop()` gates in the adapter, REST handler and GraphQL resolver collapsed `False` into "unset", which is what made the inverted flag hard to read; a flag with a default has no third state to carry.

Applied across the whole surface: v2 DTO + Strawberry GQL inputs, v1 REST DTO + handler, `gql_legacy` input, adapter, service, action, and the client SDK/CLI (`--purge-shared-vfolders` → `--delete-shared-vfolders`).

## Points for review

- **`gql_legacy` gained a new field.** `AGENTS.md` says not to add to `gql_legacy`, but the client SDK's `User.purge()` calls the legacy `purge_user` mutation, so without it the CLI could no longer express "delete the shared folders" at all. Happy to drop it if the capability loss is acceptable.
- **The CLI flag was renamed, not aliased.** Scripts passing `--purge-shared-vfolders` will now fail rather than silently do the wrong thing. Say the word if a hidden alias is preferred.
- **WebUI is a separate repo.** `PurgeUsersModal.tsx` still sends `purge_shared_vfolders`; its "Delete shared virtual folders as well?" checkbox only becomes truthful once lablup/backend.ai-webui switches to `delete_shared_vfolders`. Until then that checkbox is inert, which is the safe failure mode. Blocks lablup/backend.ai-webui#9508.

## Not addressed here

Two defects found while tracing `_migrate_shared_vfolders`, both independent of the flag and unchanged by this PR:

- The `vfolder_permissions` join has no `DISTINCT`, so a folder shared with N users produces N `UPDATE`s; the collision suffix is regenerated per iteration, so the folder keeps the last random name and the returned rowcount over-counts.
- Only folders with a per-user `vfolder_permissions` row migrate. A user-owned folder shared through a project/group grant may fall outside that join and be deleted even when preservation was requested. Not traced to a conclusion — worth its own issue.

## Test plan

- [x] `pants fmt ::`, `pants fix ::`, `pants lint --changed-since=origin/main`
- [x] `pants check --changed-since=origin/main` (14 files, mypy clean)
- [x] `pants test` in CI
- [x] GraphQL schema regeneration

Resolves BA-7733

----
📚 Documentation preview 📚: https://sorna--14367.org.readthedocs.build/en/14367/

----
📚 Documentation preview 📚: https://sorna-ko--14367.org.readthedocs.build/ko/14367/

### DIFF
--- a/docs/manager/graphql-reference/schema.graphql
+++ b/docs/manager/graphql-reference/schema.graphql
@@ -2872,7 +2872,12 @@ type PurgeUser {
 }
 
 input PurgeUserInput {
-  purge_shared_vfolders: Boolean
+  purge_shared_vfolders: Boolean @deprecated(reason: "Deprecated since 26.9.0 and ignored: it did the opposite of its name. Use `delete_shared_vfolders`.")
+
+  """
+  Added in 26.9.0. The default value is `false`. If true, the user's virtual folders shared with other users are deleted along with the rest; otherwise their ownership goes to the requester.
+  """
+  delete_shared_vfolders: Boolean
 
   """
   Added in 25.4.0. The default value is `false`. Indicates whether the user's existing endpoints are delegated to the requester.
--- a/changes/14367.feature.md
+++ b/changes/14367.feature.md
@@ -1,0 +1,1 @@
+Add `delete_shared_vfolders` to the user purge APIs. When set, the purged user's virtual folders that are shared with other users are deleted along with the rest; otherwise their ownership is transferred to the requesting admin.
--- a/docs/manager/graphql-reference/supergraph.graphql
+++ b/docs/manager/graphql-reference/supergraph.graphql
@@ -2627,9 +2627,14 @@ input BulkPurgeUsersV2Options
   @join__type(graph: STRAWBERRY)
 {
   """
-  If true, migrate shared virtual folders to the admin user before purging.
+  Added in 26.3.0. Whether to purge the user's shared virtual folders. Deprecated since 26.9.0.
   """
-  purgeSharedVfolders: Boolean! = false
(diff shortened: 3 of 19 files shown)


docs/manager/graphql-reference/schema.graphql
```diff
@@ -2872,7 +2872,12 @@ type PurgeUser {
 }
 
 input PurgeUserInput {
-  purge_shared_vfolders: Boolean
+  purge_shared_vfolders: Boolean @deprecated(reason: "Deprecated since 26.9.0 and ignored: it did the opposite of its name. Use `delete_shared_vfolders`.")
+
+  """
+  Added in 26.9.0. The default value is `false`. If true, the user's virtual folders shared with other users are deleted along with the rest; otherwise their ownership goes to the requester.
+  """
+  delete_shared_vfolders: Boolean
```
HyeockJinKim, line 2880: Since “personal projects” have been introduced, I don’t think this part is relevant anymore—what do you think?

PR #14399, opened 2026-09-08
fix(BA-7772): open the image reads that a wiring swap can fix

jopemachine:
## 📚 Stacked PRs

1. 👉 **#14399** — `fix(BA-7772): open the image reads that a wiring swap can fix` ← you are here
2. ⬇️ #14413 — `fix(BA-7773): open the last two image read paths that still require super-admin`

## Summary
- Four image reads — `get_all_images`, `get_image_by_id`, `get_image_by_identifier`, `get_images_by_canonicals` — were wired through `ProcessorGroup.global_scope`. `GlobalActionProcessor` always prepends `SuperAdminActionValidator`, so the legacy `images` and `image` queries and `image_node` answered `user_auth_forbidden` to a non-superadmin, and the WebUI session launcher could not list environments.
- The layer above already decides what each role may read: `resolve_images` narrows ADMIN/USER through `Image.filter_allowed()` (the domain's allowed registries) and `matches_filter()` (customized images by owner), and all three `get_*` service methods read the caller's role to set `hide_agents`. That last one is a degradation path no non-superadmin could reach while the gate stood in front of it.
- Wire the four through `ProcessorGroup.public` / `PublicActionProcessor`, which replaces the SUPERADMIN gate with an authentication check and accepts read operations only. Naming follows the public shape rule in `services/AGENTS.md`.
- Drop the `# Actions without RBAC validation` comment in `services/image/processors.py`: `global_scope` gates on SUPERADMIN, so the line described neither half of the block it sat above.
- Remove `@pytest.mark.timeout(10)` from `tests/component/image/test_image_registry.py`. It sits on the file's first test, whose setup is what first starts the session-scoped postgres, redis and etcd containers; pytest-timeout covers setup, so three container starts had a ten-second budget. The test is `xfail`, which swallowed the setup error, and the session fixture was then cached as failed, erroring the ten DB-backed tests below it. The run takes ~42s and passes once the marker is gone.

### Verified against a running server
A regular user (`role=user`) was used throughout. Reproduced on the merge base first: `images`, `image(id:)` and `image(reference:, architecture:)` each returned `Insufficient privilege. (This operation requires super-admin privileges.)` / `user_auth_forbidden`. After the change all three return data, and a session created by that user reaches RUNNING with its container up. Superadmin behaviour is unchanged.

`get_images_by_canonicals` is the exception: `Image.batch_load_by_image_ref` is its only route into the GQL layer and nothing calls that, so it had no reachable caller to break. It moves with the other three for consistency rather than to fix an observed failure. Session creation through REST v2 `enqueue` was never affected either — it takes `image_id` directly and touches none of these actions.

### Left alone here
- `get_image_installed_agents` — moved in BA-7773 instead, so this PR stays one reviewable unit.
- `search_images` / `search_aliases` — one action backs both `admin_search*` (superadmin by design, per `api/AGENTS.md`) and the DataLoader `batch_load_by_ids` used across `gql/data_loader/data_loaders.py`. One action cannot carry two gates; splitting it is BA-7774.

Backport: none

The gate does not exist on either maintained version: on 26.8 and 26.4 these actions are wired as `ActionProcessor(service.…, action_monitors)` with no validators, and `resolve_images` filters ADMIN/USER the same way it does here. It arrived on `main` with the v2 processor migration, so there is nothing to backport.

Resolves BA-7772.

### DIFF
--- a/src/ai/backend/manager/api/gql_legacy/image.py
+++ b/src/ai/backend/manager/api/gql_legacy/image.py
@@ -75,7 +75,7 @@ from ai.backend.manager.services.image.actions.forget_image import (
     ForgetImageAction,
     ForgetImageByIdAction,
 )
-from ai.backend.manager.services.image.actions.get_all_images import GetAllImagesAction
+from ai.backend.manager.services.image.actions.get_all_images import PublicGetAllImagesAction
 from ai.backend.manager.services.image.actions.get_image_installed_agents import (
     GetImageInstalledAgentsAction,
 )
@@ -307,8 +307,8 @@ class Image(graphene.ObjectType):  # type: ignore[misc]
             filter_by_statuses = [ImageStatus.ALIVE]
         if types is None:
             types = set()
-        result = await ctx.processors.image.get_all_images.run(
-            GetAllImagesAction(status_filter=filter_by_statuses)
+        result = await ctx.processors.image.public_get_all_images.run(
+            PublicGetAllImagesAction(status_filter=filter_by_statuses)
         )
         all_items = [cls.from_image_with_agent_install_status(img) for img in result.data.values()]
         return [item for item in all_items if item.matches_filter(ctx, types)]
--- a/src/ai/backend/manager/api/gql_legacy/image.py
+++ b/src/ai/backend/manager/api/gql_legacy/image.py
@@ -80,9 +80,9 @@ from ai.backend.manager.services.image.actions.get_image_installed_agents import
     GetImageInstalledAgentsAction,
 )
 from ai.backend.manager.services.image.actions.get_images import (
-    GetImageByIdAction,
-    GetImageByIdentifierAction,
-    GetImagesByCanonicalsAction,
+    PublicGetImageByIdAction,
+    PublicGetImageByIdentifierAction,
+    PublicGetImagesByCanonicalsAction,
 )
 from ai.backend.manager.services.image.actions.purge_images import PurgeImageByIdAction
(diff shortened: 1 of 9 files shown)


src/ai/backend/manager/services/image/actions/get_all_images.py
```diff
@@ -12,7 +12,7 @@
 
 
 @dataclass
-class GetAllImagesAction(ImageAction):
+class PublicGetAllImagesAction(ImageAction):
```
fregataa, line 15: Should action type name be changed?

src/ai/backend/manager/api/gql_legacy/image.py
```diff
@@ -278,8 +269,8 @@ async def load_item_by_id(
     ) -> Image:
         if filter_by_statuses is None:
             filter_by_statuses = [ImageStatus.ALIVE]
-        result = await ctx.processors.image.get_image_by_id.run(
-            GetImageByIdAction(
+        result = await ctx.processors.image.public_get_image_by_id.run(
+            PublicGetImageByIdAction(
```
HyeockJinKim, line 273: ?? In the case of "get image," I think it should be based on image access permissions. It seems like it should function as a single entity.

HyeockJinKim (changes requested):

HyeockJinKim (approved):

HyeockJinKim:
This isn't the right approach from an implementation standpoint, but I'll leave it open temporarily to allow access and then clean up the actions.

PR #14400, opened 2026-09-08
refactor(BA-7746): switch object_storage update DTO to Unset

wngh4628:
Switch the `object_storage` update DTO from the legacy `Sentinel` marker to `Unset` (omitted = no change, `null` = clear).

- `common/dto/manager/v2/object_storage/request.py`: `UpdateObjectStorageInput.region` changes from `str | Sentinel | None = SENTINEL` to `str | None | Unset = UNSET`; the field description now reads "Omit to leave unchanged; null clears."
- `manager/api/adapters/object_storage/adapter.py`: the `Sentinel` branch for `region` is replaced with `TriState.from_unset(input.region)`.
- `client/cli/v2/object_storage/commands.py`: an omitted `--region` now defaults to `UNSET` instead of `SENTINEL`.
- `tests/unit/common/dto/manager/v2/object_storage/test_request.py`: `SENTINEL`/`Sentinel` assertions become `UNSET`/`Unset`.

----
📚 Documentation preview 📚: https://sorna--14400.org.readthedocs.build/en/14400/

----
📚 Documentation preview 📚: https://sorna-ko--14400.org.readthedocs.build/ko/14400/


src/ai/backend/client/cli/v2/object_storage/commands.py
```diff
@@ -94,13 +94,13 @@ def update(
...
 
-    # SENTINEL means "no change", None means "clear the field".
-    # When the CLI user does not pass --region, keep SENTINEL (no change).
+    # UNSET means "no change", None means "clear the field".
+    # When the CLI user does not pass --region, keep UNSET (no change).
     # When they pass an empty string, interpret as None (clear).
-    region_value: str | Sentinel | None = SENTINEL
+    region_value: str | None | Unset = UNSET
     if region is not None:
         region_value = region if region else None
```
fregataa, line 116: We decided to add nullify flag for tristate values. How about `set-null-region`?

wngh4628: --set-null-region` clears the region and is mutually exclusive with `--region`; passing an empty string no longer clears it.
  Options the user does not pass now reach the DTO as `UNSET` instead of `None`.

seedspirit (approved):

PR #14402, opened 2026-09-08
test(BA-7736): keep one test container per execution slot across pytest processes

HyeockJinKim:
## Summary
- `etcd_container`, `redis_container` and `postgres_container` take their container from a pool with one container per (role, Pants execution slot). A process reuses what the previous process in its slot left running instead of starting three containers per test file. Containers are found by label and their port read from `inspect`; no fixed names or ports. The pool is keyed by checkout path, so worktrees do not share it.
- A detached reaper removes the pool once no process has held it (flock) for 20 s, so a process killed by the Pants timeout is handled too. `BACKEND_TEST_SHARE_CONTAINERS=0` restores per-process containers.
- `database_connection` is defined once in `ai.backend.testutils.db_fixtures` and creates a per-process database, replacing six copies that all used the fixed database `testing`. Component/integration teardown now terminates only its own database's connections. The pooled postgres has its own credentials (`bai_testpool`), defined once in `bootstrap.py`.

## Measurements
50 component test files (first 50 of `pants list tests/component::`), `--process-execution-local-parallelism=4`, same machine, base commit vs this branch, 1-minute load 6-10 during both runs:

| | before (9ed5ea3f31) | after |
|---|---|---|
| wall clock | 300 s | 225 s |
| per-file wall (mean / max) | 21.1 s / 55.7 s | 15.3 s / 52.0 s |
| containers started | 150 (3 per file) | 12 (3 per slot) |
| first-test setup per file | 3.7-4.8 s | 0.8-0.9 s on reuse, 3.7 s on create |

The saving is a fixed cost per file, so it grows with the number of selected files. The remaining ~15 s per file is Pants process, pex and import cost.

Full `tests/component::` plus the unit repository/model tests at parallelism 12: 326 passed; the pool held 36 containers during the run and was empty 40 s after it.

## Test plan
- [x] `pants test tests/component::` and `tests/unit/manager/{repositories,models}::` at parallelism 12 pass against the pool
- [x] `pants test --changed-since=HEAD --changed-dependents=direct` passes (704 targets; the one failure, `test_etcd_config_repo.py`, is a pre-existing mock-only failure)
- [x] pool is removed after the run; a process killed mid-run leaves no held lock
- [x] `BACKEND_TEST_SHARE_CONTAINERS=0` path passes

Resolves BA-7736

https://claude.ai/code/session_01TT4bCWvpdneBvnHziEQHp6


src/ai/backend/testutils/db_fixtures.py
```diff
@@ -0,0 +1,99 @@
...
+    return (
+        f"postgresql+asyncpg://{POSTGRES_USER}:{POSTGRES_PASSWORD}@{addr.host}:{addr.port}/{dbname}"
+    )
+
+
+async def _create_database(addr: HostPortPairModel, dbname: str) -> None:
+    engine = create_async_engine(
+        _url(addr, POSTGRES_MAINTENANCE_DB),
+        connect_args=pgsql_connect_opts,
+        isolation_level="AUTOCOMMIT",
```
fregataa, line 44: Should the level be AUTOCOMMIT?

HyeockJinKim: It was a situation where compatibility with existing operations had been ensured.

src/ai/backend/testutils/db_fixtures.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+    try:
+        while True:
+            try:
+                async with engine.connect() as conn:
+                    await conn.execute(sa.text(f'CREATE DATABASE "{dbname}";'))
+            except (ConnectionError, OSError):
+                await asyncio.sleep(0.1)
+                continue
+            break
```
fregataa, line 60: Does this not require any timeout? Or a caller has a timeout?

fregataa (dismissed):

PR #14408, opened 2026-09-08
refactor(BA-7751): switch rbac update DTO to Unset

wngh4628:
Switch the `rbac` role update DTOs from the legacy `Sentinel` marker to `Unset` (omitted = no change, `null` = clear).

- `common/dto/manager/v2/rbac/request.py`: `UpdateRoleInput.description` changes from `str | Sentinel | None = SENTINEL` to `str | None | Unset = UNSET`; the field description now reads "Omit to leave unchanged; null clears."
- `common/dto/manager/rbac/request.py`: `UpdateRoleRequest.description` gets the same change, since the REST role adapter builds its updater from it.
- `manager/api/adapters/rbac/adapter.py` and `manager/api/rest/rbac/role_adapter.py`: the `SENTINEL` branches for `description` are replaced with `TriState.from_unset(...)`, and the now-redundant `TriState.nop()` pre-assignments are removed.
- `tests/unit/common/dto/manager/v2/rbac/test_request.py`, `tests/unit/manager/api/gql/rbac/test_role_input.py`: `SENTINEL`/`Sentinel` assertions become `UNSET`/`Unset`.

----
📚 Documentation preview 📚: https://sorna--14408.org.readthedocs.build/en/14408/

----
📚 Documentation preview 📚: https://sorna-ko--14408.org.readthedocs.build/ko/14408/


src/ai/backend/manager/api/adapters/rbac/adapter.py
```diff
@@ -811,14 +810,10 @@ async def update_permission(self, input: UpdatePermissionInputDTO) -> Permission
...
-                else OptionalState.nop()
-            ),
+            entity_type=OptionalState.from_unset(input.entity_type),
             permission=(
-                OptionalState.update(single_bit(input.permission_bit()))
-                if input.permission is not None
+                OptionalState.update(single_bit(input.permission.to_permission()))
+                if isinstance(input.permission, PermissionBitDTO)
                 else OptionalState.nop()
             ),
```
fregataa, line 818: Could you check how `single_bit()` works? I think It can be placed in DTO function

fregataa (commented):
Please check CI failure

src/ai/backend/manager/api/adapters/rbac/adapter.py
wngh4628: PermissionBitDTO is a StrEnum whose members mirror Permission by name, 
so to_permission() (Permission[self.name]) always yields exactly one bit, 
and RolePermissionUpdater/RolePermissionCreator already call  single_bit() right before the write. 
The adapter call is redundant; 
I'll drop it and use input.permission.to_permission() directly.

wngh4628: Cleaned up the messy branching by moving it into UpdatePermissionInput. As for single_bit(): PermissionBitDTO.to_permission() always returns
  exactly one bit, and RolePermissionUpdater.build_values() (updaters.py:46) calls single_bit() again right before the write, so the adapter
  doesn't strictly need it.

fregataa (approved):

PR #14416, opened 2026-09-08
refactor(BA-7758): switch domain update DTO to Unset

wngh4628:
Switch the `domain` update DTO from the legacy `Sentinel` marker to `Unset` (omitted = no change, `null` = clear).

- `common/dto/manager/v2/domain/request.py`: `UpdateDomainInput.description`, `allowed_docker_registries`, and `integration_name` change from `X | Sentinel | None = SENTINEL` to `X | None | Unset = UNSET`; the field descriptions now read "Omit to leave unchanged; null clears." (`allowed_docker_registries` is non-nullable, so its description stops at "Omit to leave unchanged.").
- `manager/api/adapters/domain/adapter.py`: the three `Sentinel` branches are replaced with `TriState.from_unset(...)` for `description` and `integration_name`, and `OptionalState.from_unset(...)` for `allowed_docker_registries`.
- `tests/unit/common/dto/manager/v2/domain/test_request.py`: `SENTINEL`/`Sentinel` assertions become `UNSET`/`Unset`.

----
📚 Documentation preview 📚: https://sorna--14416.org.readthedocs.build/en/14416/

----
📚 Documentation preview 📚: https://sorna-ko--14416.org.readthedocs.build/ko/14416/


src/ai/backend/common/dto/manager/v2/domain/request.py
```diff
@@ -63,21 +64,21 @@ class UpdateDomainInput(BaseRequestModel):
...
-        default=SENTINEL,
-        description="New domain description. Set to null to clear.",
+    description: str | None | Unset = Field(
+        default=UNSET,
+        description="Updated domain description. Omit to leave unchanged; null clears.",
     )
     is_active: bool | None = Field(
         default=None,
         description="Updated active status.",
     )
```
HyeockJinKim, line 76: I thought we agreed that optional values in both GQL and REST should be left unset, so why are they still treated as simply optional here?

wngh4628: Fixed in 22d3203ef. `name` and `is_active` were still defaulting to `None` on the DTO, and the GQL input had `None` defaults on `name`, `is_active`, and
  `allowed_docker_registries` as well. All five update fields now default to `UNSET` on both the DTO and the GQL input, and the adapter converts `is_active` with
  `OptionalState.from_unset`, so the unset/null distinction is decided only below the adapter.

fregataa (commented):
Good! no need to change CLI?

fregataa (approved):

PR #14418, opened 2026-09-08
refactor(BA-7755): switch notification update DTO to Unset

wngh4628:
Switch the `notification` update DTOs from the legacy `Sentinel` marker to `Unset` (omitted = no change, `null` = clear).

- `common/dto/manager/v2/notification/request.py`: `UpdateNotificationChannelInput.description` and `UpdateNotificationRuleInput.description` change from `str | Sentinel | None = SENTINEL` to `str | None | Unset = UNSET`; the field descriptions now read "Omit to leave unchanged; null clears."
- `manager/api/adapters/notification/adapter.py`: the two `Sentinel` branches in the channel and rule updater builders are replaced with `TriState.from_unset(input.description)`.
- `tests/unit/common/dto/manager/v2/notification/test_request.py`: `SENTINEL`/`Sentinel` assertions become `UNSET`/`Unset`.

----
📚 Documentation preview 📚: https://sorna--14418.org.readthedocs.build/en/14418/

----
📚 Documentation preview 📚: https://sorna-ko--14418.org.readthedocs.build/ko/14418/


src/ai/backend/common/dto/manager/v2/notification/request.py
```diff
@@ -165,9 +166,9 @@ class UpdateNotificationRuleInput(BaseRequestModel):
...
     name: str | None = Field(default=None, description="Updated rule name")
-    description: str | Sentinel | None = Field(
-        default=SENTINEL,
-        description="Updated rule description. Use SENTINEL to clear.",
+    description: str | None | Unset = Field(
+        default=UNSET,
+        description="Updated rule description. Omit to leave unchanged; null clears.",
     )
     message_template: str | None = Field(
         default=None,
```
HyeockJinKim: We agreed to set everything—including GQL and pydantic—to "Unset" and to distinguish behavior based on `OptionalState` and `TriState` only within the internal logic, but it seems none of that has been implemented.

wngh4628: Fixed in 0a102158e. The previous commit had only switched `description` to Unset; `name`, `spec`, `enabled`, and `message_template` still defaulted to `None`,
  so the adapter was branching on `is not None` in the DTO layer. Both update DTOs now declare every field as `X | None | Unset = UNSET`, and the adapter
  converts each with `OptionalState.from_unset` / `TriState.from_unset`, so the null/unset distinction lives only below the adapter. GQL input types were already
  all `UNSET`, so nothing changed there.

fregataa (approved):

PR #14421, opened 2026-09-08
refactor(BA-7763): switch deployment_revision_preset update DTO to Unset

wngh4628:
Switch the `deployment_revision_preset` update DTO from the legacy `Sentinel` marker to `Unset` (omitted = no change, `null` = clear).

- `common/dto/manager/v2/deployment_revision_preset/request.py`: the nine `UpdateDeploymentRevisionPresetInput` fields (`description`, `image_id`, `model_definition`, `startup_command`, `bootstrap_script`, `open_to_public`, `replica_count`, `revision_history_limit`, `deployment_strategy`) change from `X | Sentinel | None = SENTINEL` to `X | None | Unset = UNSET`.
- `manager/api/adapters/deployment_revision_preset/adapter.py`: every `SENTINEL` branch is replaced with `TriState.from_unset(...)`; `model_definition` and `deployment_strategy` / `deployment_strategy_spec` use `.map(...)` for conversion; the local helpers `_convert_model_definition_state`, `_convert_tri_state`, `_convert_strategy_input`, `_convert_strategy_update_state`, and `_convert_strategy_spec_update_state` are deleted.

----
📚 Documentation preview 📚: https://sorna--14421.org.readthedocs.build/en/14421/

----
📚 Documentation preview 📚: https://sorna-ko--14421.org.readthedocs.build/ko/14421/


src/ai/backend/manager/api/adapters/deployment_revision_preset/adapter.py
```diff
@@ -325,69 +317,31 @@ async def update(
...
-            environ=environ_state,
-            runtime_variant_preset_values=preset_values_state,
-            open_to_public=self._convert_tri_state(input.open_to_public),
-            replica_count=self._convert_tri_state(input.replica_count),
-            revision_history_limit=self._convert_tri_state(input.revision_history_limit),
-            deployment_strategy=self._convert_strategy_update_state(input.deployment_strategy),
-            deployment_strategy_spec=self._convert_strategy_spec_update_state(
-                input.deployment_strategy
+            deployment_strategy_spec=TriState.from_unset(input.deployment_strategy).map(
+                lambda si: self._convert_required_strategy_input(si)[1]
```
fregataa: How about not using map here and just assigning values here?

src/ai/backend/manager/api/adapters/deployment_revision_preset/adapter.py
```diff
@@ -295,28 +295,20 @@ async def update(
         input: UpdateDeploymentRevisionPresetInput,
     ) -> UpdateDeploymentRevisionPresetPayload:
         # A model_definition patch merges onto the currently stored preset, so fetch it
-        # upfront when it's being patched (SENTINEL/None need no current value).
+        # upfront when it's being patched (unset/None need no current value).
         current: DeploymentRevisionPresetNode | None = None
-        if input.model_definition not in (SENTINEL, None):
+        if isinstance(input.model_definition, UpdatePresetModelDefinitionInput):
```
seedspirit, line 301: Why is the `isinstance` check necessary here?

src/ai/backend/manager/api/adapters/deployment_revision_preset/adapter.py
wngh4628: The helper now returns only the spec dict and the enum comes from`input.deployment_strategy.type`, so the tuple indexing is gone; create() was adjusted slightly to match.

src/ai/backend/manager/api/adapters/deployment_revision_preset/adapter.py
wngh4628: model_definition is typed `UpdatePresetModelDefinitionInput | None | Unset`, so anything that is not an instance is either UNSET or None. 

Only an actual instance carries a patch and needs the current preset loaded, so instead of comparing against UNSET and None separately, a single is instance check selects exactly that case.

src/ai/backend/manager/api/adapters/deployment_revision_preset/adapter.py
```diff
@@ -574,91 +529,29 @@ def _convert_preset_values_input(
...
-                rolling = strategy_input.rolling_update
-                spec_dict: dict[str, Any] = (
-                    rolling.model_dump(mode="json") if rolling is not None else {}
-                )
-                return DeploymentStrategy.ROLLING, spec_dict
-            case DeploymentStrategy.BLUE_GREEN:
-                bg = strategy_input.blue_green
-                spec_dict = bg.model_dump(mode="json") if bg is not None else {}
-                return DeploymentStrategy.BLUE_GREEN, spec_dict
+        return TriState.from_unset(value).map(_merge)
```
fregataa: it seems this `_merge()` should be done before `map()`.

wngh4628: the merge now runs first and the result is wrapped with TriState.update(), so map() and the closure are gone.

src/ai/backend/manager/api/adapters/deployment_revision_preset/adapter.py
```diff
@@ -295,28 +296,20 @@ async def update(
...
-            if input.preset_values is not None
-            else OptionalState.nop()
+            OptionalState.from_unset(input.resource_slots)
+            .map(
+                lambda slots: [
+                    PresetResourceSlotCreator(entry=entry)
+                    for entry in self._convert_resource_slots_input(slots)
+                ]
+            )
+            .optional_value()
```
fregataa: same here

wngh4628: slot_creators is now built in a plain branch like create() does, without map() or the lambda.

src/ai/backend/manager/api/adapters/deployment_revision_preset/adapter.py
```diff
@@ -301,16 +301,10 @@ async def update(
...
-            .map(
-                lambda slots: [
-                    PresetResourceSlotCreator(entry=entry)
-                    for entry in self._convert_resource_slots_input(slots)
-                ]
-            )
-            .optional_value()
-        )
+        slot_creators: list[PresetResourceSlotCreator] | None = None
+        if isinstance(input.resource_slots, list):
```
fregataa, line 305: This condition check seems not necessary

wngh4628: It is needed: the check keeps UNSET and None out of the conversion. 
Without it, UNSET raises TypeError inside `_convert_resource_slots_input` (it isn't iterable), and None becomes `[]`, which the repository treats as "replace with nothing" and wipes the slots. 
Only a real list should produce slot creators; anything else must stay None so the slots are left alone.

fregataa (approved):

PR #14425, opened 2026-09-08
refactor(BA-7759): switch scaling_group update DTO to Unset

wngh4628:
Switch the `scaling_group` update DTO from the legacy `Sentinel` marker to `Unset` (omitted = no change, `null` = clear).

- `common/dto/manager/v2/scaling_group/request.py`: `UpdateScalingGroupInput.description`, `wsproxy_addr`, and `wsproxy_api_token` change from `str | Sentinel | None = SENTINEL` to `str | None | Unset = UNSET`; the field descriptions now read "Omit to leave unchanged; null clears."
- `tests/unit/common/dto/manager/v2/scaling_group/test_request.py`: `SENTINEL`/`Sentinel` assertions become `UNSET`/`Unset`.


src/ai/backend/common/dto/manager/v2/scaling_group/request.py
```diff
@@ -66,19 +67,17 @@ class UpdateScalingGroupInput(BaseRequestModel):
...
+    wsproxy_addr: str | None | Unset = Field(
+        default=UNSET,
+        description="WebSocket proxy address. Omit to leave unchanged; null clears.",
     )
-    wsproxy_api_token: str | Sentinel | None = Field(
-        default=SENTINEL,
-        description="WebSocket proxy API token. Use SENTINEL to clear, null to keep existing value.",
+    wsproxy_api_token: str | None | Unset = Field(
+        default=UNSET,
+        description="WebSocket proxy API token. Omit to leave unchanged; null clears.",
```
HyeockJinKim: It seems like this type isn't used at all—have you checked to see if that's the case?

wngh4628: You're right, it's a dead type. Apart from the __init__.py re-export and its unit tests, nothing references it: no REST v2 handler, adapter, or GQL type. Since
  it's only used by the unit tests, would it be fine to delete it?

wngh4628:
Removed the unused files in 8118d1de5 — the whole `dto/manager/v2/scaling_group` package and its unit tests.
Nothing referenced them; `dto/manager/v2/resource_group` carries the live equivalents. If you'd prefer this as a separate PR or want it rolled back, let me know.

fregataa (approved):

PR #14444, opened 2026-09-09
refactor(BA-7740): route every VE-entity create/delete through ops write paths

HyeockJinKim:
An entity row that carries a virtual entity node was being inserted or deleted with a bare `sa.insert` / `sa.delete` / `session.add` / `session.delete` in nine places. The insert paths gave the row no node at all; the delete paths left the node behind. Each is routed through an `EntityCreator` / `EntityPurger` and the v2 write ops, so the node is provisioned and torn down with the row.

Every consumer of the deprecated write specs under `repositories/base/` is moved onto the `models/specs/` lineage as well, and the specs and the provider that ran them are deleted.

### VE-entity write paths

| Key | What |
|---|---|
| 결함-1 | `DeploymentPurger` — the deployment delete in the deployment db source |
| 결함-2 | `UserEndpointPurger` — a user's deployments in the user db source |
| 결함-3 | `ImagePurger` — the image purge in the image db source and the manager CLI |
| 결함-4 | `VFolderInvitationCreator` — the invitation insert |
| 결함-4 | `ModelCardPurger` — the cards a vfolder purge cascades to |
| 결함-5 | `VFolderInviteeInvitationBatchPurger` — the three invitation deletes |
| 결함-6 | container registry create and delete in the deprecated `gql_legacy` mutations |
| 결함-7 | `NetworkPurger` — the network delete in `gql_legacy` |
| 결함-9 | `ResourcePresetPurger` — the preset delete |
| 결함-10 | `KernelUsageRecordCreator` — usage slices as nested fields of their kernel |
| 삭제-1 | the entity deletes nothing calls, removed |
| — | comments marking the batch purges a user operation must scope |

### Legacy write spec removal

| What |
|---|
| `DeploymentPolicyPurger`, `RolePermissionCreator` / `RolePermissionUpdater` — the last field-row consumers |
| `replace_permissions` on the permission ops, `grant_roles` on the entity write ops |
| the permission and permission controller specs nothing calls, removed |
| `repositories/base/{creator,updater,purger,upserter}.py`, removed |
| `WriteOps` and `DBOpsProvider` in `ops/base/`, removed — their last callers only read |

## Behavior notes

- Explicit child deletes are dropped where the FK already carries `ondelete="CASCADE"`: routings, deployment policies, endpoint tokens, image aliases. `ProjectEndpointPurger` already relies on the same cascade.
- The model card purge no longer shares the transaction that marks the vfolders `DELETE_ONGOING`. Ops are handed out session-bound and a repository must not construct them, so the cards go first; a failure in between leaves the folders alive and card-less, which a retry completes.
- The deprecated container registry mutations now run through the processors their `*Node` twins use, which carry the same `SUPERADMIN` gate. The hostname names a row per project, so the delete resolves them all and purges each, as the name-keyed statement did.
- The resource preset delete takes the preset id alone. Its callers already resolve a name to an id, so the read that preceded the delete is gone.
- `replace_role_permissions` keeps its GraphQL `failed` field, always empty and carrying a deprecation reason: v2 refuses partial failure on a replace.
- The session and idle checker db sources read joined queries off the engine directly. A spec owns a single table, so these queries were never ops material; they now run the batch querier on their own session, as the other joined reads in the same files already do.

## Excluded

- Field rows and rows with no node type, including 32 of the 33 raw writes in the deployment db source. Verified: endpoint tokens, auto scaling rules, event logs, login sessions and history, keypairs, image aliases.
- Every `sa.update`.
- Entity types with no v2 spec: `session_template`, `service_catalog`, `agent`.
- `RoleCreatorSpec`, dead and handled separately.

## Follow-up

- Ten `batch_purge_entities_in_global` call sites reached from user operations should be scoped instead. Converting them needs an operation scope resolved through the virtual-entity own edges, and those edges are not backfilled for these entity types yet, so each site is marked with a comment rather than changed here.
- `BatchQuerier` and the rest of the read side in `repositories/base/` move to `Searcher`.

https://claude.ai/code/session_01QhG4NeG6XT1zgDvjzethic

----
📚 Documentation preview 📚: https://sorna--14444.org.readthedocs.build/en/14444/

----
📚 Documentation preview 📚: https://sorna-ko--14444.org.readthedocs.build/ko/14444/


src/ai/backend/manager/models/endpoint/purgers.py
```diff
@@ -0,0 +1,75 @@
...
+    def conflict_checks(self) -> Sequence[ConflictCheck]:
+        return ()
+
+    @override
+    def to_data(self, row: EndpointRow) -> DeploymentInfo:
+        return row.to_bare_deployment_info()
+
+
+@dataclass
+class UserEndpointPurger(EntityBatchPurger[EndpointRow, DeploymentID]):
```
jopemachine, line 51: Nit: How about UserDeploymentPurger like DeploymentPurger above?

src/ai/backend/manager/models/endpoint/purgers.py
```diff
@@ -0,0 +1,75 @@
...
+            EndpointRow.session_owner == self.user_id,
+            EndpointRow.lifecycle_stage.in_(self.lifecycle_stages),
+        )
+
+    @override
+    def conflict_checks(self) -> Sequence[ConflictCheck]:
+        return ()
+
+    @override
+    def to_data(self, row: EndpointRow) -> DeploymentID:
```
jopemachine, line 74: It seems incorrect for to_data to return an ID type.

PR #14447, opened 2026-09-09
refactor(BA-7775): declare entity and field types as per-kind classes

HyeockJinKim:
## Summary

- Each entity and field type is now a class that answers its own `name()` and `description()`, replacing the 53 `*_ENTITY_TYPE` and 32 `*_FIELD_TYPE` module constants. A value is built at its use site as `AgentEntityType()`.
- A field kind answers the entity kind owning its rows through `owner_type()`. A kind whose owner is a value on the row rather than one fixed kind is a `DanglingFieldType`, which covers audit log, audit log scope, label and the three usage buckets.
- A value rebuilt from a string stays a bare base instance and answers no metadata. That is reserved for a boundary reading one, such as a row column, a legacy action or an RBAC element, and the rule is written in `common/data/entity/KNOWLEDGE.md`.
- Adds `backend.ai mgr ops types`, which lists every declared type by the ownership it declares, and a guard comparing the owner a field group is wired under against the owner its field type declares.

The three usage bucket kinds were declaring `domain`, `project` and `user` as owners while being wired as dangling field groups. The new guard caught that, and they are now declared dangling to match.

## Test plan

- [ ] `pants lint` and `pants check` pass over the changed set
- [ ] `pants test --changed-since=origin/main --changed-dependents=direct`
- [ ] `tests/unit/common/data/entity/test_types.py` sweeps every kind for name, description, owner and round-tripping
- [ ] `tests/unit/manager/actions/test_field_owner_wiring.py` cross-checks the declarations against the production wiring
- [ ] `backend.ai mgr ops types` prints the catalog

## Not in this change

- `secret` and the fair share kinds should become dangling field types, but their operations are service-backed and `FieldGroup` offers only ops-spec wiring, so moving them needs a framework addition.
- `auth`, `etcd_config`, `export` and `manager_admin` name no row and are left as they are.

Resolves BA-7775

https://claude.ai/code/session_0145xGLj7NBzqsNjxrGMpo2p

### DIFF
--- a/src/ai/backend/common/data/entity/KNOWLEDGE.md
+++ b/src/ai/backend/common/data/entity/KNOWLEDGE.md
@@ -3,7 +3,7 @@ name: entity-type-catalog
 type: reference
 description: entity classification (Entity/Public entity/Global/Field/Virtual), ownership and reference relations, the foreign-key rule for assignments
 scope: src/ai/backend/common/data/entity
-keywords: [EntityType, ENTITY_TYPE, EntityCreator, GlobalEntityCreator, FieldEntityCreator, member_of, virtual entity, ownership, entity relation]
+keywords: [EntityType, FieldType, DanglingFieldType, EntityCreator, GlobalEntityCreator, FieldEntityCreator, member_of, virtual entity, ownership, entity relation]
 sources:
   - src/ai/backend/common/data/entity
   - src/ai/backend/manager/models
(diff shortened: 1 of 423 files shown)


src/ai/backend/common/data/entity/app_config_definition.py
```diff
@@ -3,16 +3,24 @@
...
+class AppConfigDefinitionEntityType(EntityType):
+    @override
+    @classmethod
+    def name(cls) -> str:
+        return "app_config_definition"
+
+    @override
+    @classmethod
+    def description(cls) -> str:
+        return "The declaration of an app config key and its value type."
```
jopemachine, line 20: The value type of an app config is not information held by the app config definition, and the backend does not need to be concerned with it.

src/ai/backend/common/data/entity/KNOWLEDGE.md
```diff
@@ -15,13 +15,26 @@ status: draft
...
-what the entity is, what it owns and what it is tied to is recorded here. Everything is
-stated at the conceptual level; which table holds it belongs to the `models` documents.
-
-A tree states ownership; a table states fields and references. A field is authorized
-through the entity that owns it, while a reference is not ownership and therefore never
-becomes an authorization path.
+이 패키지는 종류마다 `EntityType` 또는 `FieldType` 서브클래스를 하나 둔다. 클래스는
+`name()`과 `description()`을 답하고, 필드 종류는 행을 소유하는 엔티티 종류를
+`owner_type()`으로 답한다. 소유자가 행의 값이라 고정되지 않는 필드 종류(audit_log,
+audit_log_scope, label)는 `DanglingFieldType`을 상속해 `None`을 답한다. 무엇을 소유하고
```
seedspirit, line 21: It seems some doc changed to Korean

PR #14460, opened 2026-09-09
refactor(BA-7789): answer for every auth operation with the user it reaches

jopemachine:
## Summary

- The `auth` entity kind named no row, while every action wired under it operated on a user or on a row a user owns. It is removed, and the seven wirings move onto the `user` kind that already existed.
- `authorize`, `update_password_no_auth`, `global_unblock_user` and `global_revoke_login_session` stay global — they run before or beside a principal, so they name no id — and declare the user kind, the way `signup` already did. `public_get_role` declares it on the same public global read.
- `public_resolve_user_scope` and `public_resolve_access_key_scope` become lookups: each turns an email, an access key, or nothing at all into the user the request acts for, and answers with that user's id so the audit row names it. The keypair path reads the owner's id alongside the domain and role it already read. The REST responses are unchanged.
- The two delegation reads on `AuthRepository` answer with a `DelegationTargetUser` (`user_id`, `role`, `domain_name`) instead of positional tuples, and the access-key read takes an `AccessKey`.
- `ProcessorGroup.public_lookup` is added for a domain-run lookup every authenticated caller may resolve — the service decides the caller's reach, so a post-validator would ask the same question twice.

## Test plan

- [ ] `tests/unit/manager/actions/test_registry_catalog.py` — the catalog sweep still matches, and a new pin holds the seven wirings to `(user, kind, gate)`.
- [ ] `tests/unit/manager/services/auth/test_resolve_scope.py` — both resolutions answer with the user they resolved, self and delegated alike.
- [ ] `tests/unit/manager/api/service/test_handler.py` — the scope result the service handler reads is unchanged in shape.
- [ ] CI type check and the component suite.

Resolves BA-7789

https://claude.ai/code/session_01Su1Tz6odc3MoQHBUgNFJuH


jopemachine:
Closing without merging. Superseded by #14651 (BA-7908), which already removed `AuthEntityType`; BA-7789 is closed as Not Planned.

PR #14465, opened 2026-09-09
refactor(BA-7791): classify exports by their target entities

seedspirit:
## Summary
- Remove `ExportEntityType` and declare user, session, or project targets on CSV actions. Wire keypair exports as user-owned fields and audit-log export as a dangling field; keep report definitions global.
- Use `SEARCH` for CSV operations so scoped authorization requires read permission. Preserve domain, project, and user scopes and query filtering.
- Keep `BaseGlobalAction`, global audit monitoring, and `AuditLogPolicy` unchanged.

## Validation
- Local formatting, lint, type checks, and tests run before submission.
- Live server: all nine CSV exports and both report-definition endpoints succeeded as superadmin; own-keypair export returned only the caller's row.
- Live server: regular-user access to unscoped exports and report-definition endpoints returned 403.
- Scoped exports for regular users are blocked by the existing `GetReportAction` superadmin gate before scope authorization. Session CSV results were empty, so session row isolation was not verified live.
- The local DB is behind the migration head; a control keypair search returned 500 because `permissions.all_fields` is missing.

## Follow-up
- Make report-definition lookup usable from scoped export routes so regular-user read permissions can be verified end to end.
- Update the local DB schema and repeat non-admin scope checks.
- Decide whether successful CSV exports require unconditional audit recording. With the existing policy, successful `SEARCH` operations are not recorded by default.
- Record declared targets in global audit logs; the existing global monitor still uses `GlobalEntityType`.

Refs BA-7791


src/ai/backend/manager/api/rest/export/handler.py
```diff
@@ -72,10 +72,17 @@
...
+    def field_names(self) -> list[str]: ...
+
+    @property
+    def row_iterator(self) -> AsyncIterator[Sequence[Sequence[Any]]]: ...
+
+    @property
+    def encoding(self) -> str: ...
+
+    @property
+    def filename(self) -> str: ...
```
HyeockJinKim, line 85: It doesn't seem like there's any reason for this to change—is there a specific reason you made this change?

HyeockJinKim (approved):
There is a minor issue, but I'll just overlook it for now.

PR #14468, opened 2026-09-09
refactor(BA-7793): compose the storage proxy startup dependencies

fregataa:
## Summary

- Drive the storage proxy server startup from `StorageDependencyComposer` instead of the hand-rolled `AsyncExitStack` in `server_main`, so the server, `storage dependencies verify` and `storage health check` build the same resources through the same path.
- Compose `StoragePluginContext` once in a new plugins stage and inject the merged backend map into both `VolumePool` and `RootContext`; the volume registry no longer owns a plugin context as a side effect of being created, and the plugin group is no longer loaded twice.
- Extend the composer with the plugins, messaging and system stages, move the remaining hand-built resources (volume pool, background task manager, watcher, volume stats, manager client pool, health probe, service discovery, redis config, tus/volume-stats clients) into providers, and collapse the two divergent `DEFAULT_BACKENDS` copies into `volumes/backends.py`.

Stage order: bootstrap → infrastructure → plugins → messaging → storage → system.

### Behavior notes

- Workers now receive the config path and log level, and the composer loads the configuration; `main()` resolves the path before spawning so the loader's `chdir` cannot change how a relative path resolves.
- The artifact Valkey client is created once (the server used to open a second one), and the tus/artifact/volume-stats clients now resolve through their role profile target, so `override-configs` applies to them.
- The background task manager shares the composed bgtask client, removing the private `_valkey_client` access in the health probe wiring.
- The metric registry is now `CommonMetricRegistry.instance()`, the same registry the metrics endpoint exports.
- Service discovery is registered during composition, i.e. slightly before the API sites start listening — this matches the manager's ordering.

`RootContext` is unchanged and still assembled in `server.py`; breaking it up is separate work. The artifact verifier plugin context is untouched.

## Test plan

- [x] `pants fmt` / `pants fix` / `pants lint` on the changed files
- [x] `pants check` / `pants test` (CI)
- [x] Storage proxy starts and shuts down cleanly against the local halfstack: plugin context initialized once, volume stats runner and health probe started, service registered, `Started the storage-proxy service.`, clean teardown with no traceback
- [x] `backend.ai storage dependencies verify` / `backend.ai storage health check`
- [x] Startup with a storage backend plugin installed (no plugin was present in the local run)

Resolves BA-7793

### DIFF
--- a/src/ai/backend/storage/dependencies/infrastructure/redis_config.py
+++ b/src/ai/backend/storage/dependencies/infrastructure/redis_config.py
@@ -1,0 +1,32 @@
+from __future__ import annotations
+
+import logging
+from collections.abc import AsyncIterator
+from contextlib import asynccontextmanager
+from typing import override
+
+from ai.backend.common.configs.redis import RedisConfig
+from ai.backend.common.dependencies import NonMonitorableDependencyProvider
+from ai.backend.common.etcd import AsyncEtcd
+from ai.backend.common.types import safe_print_redis_config
+from ai.backend.logging import BraceStyleAdapter
+
+log = BraceStyleAdapter(logging.getLogger(__spec__.name))
+
+
+class RedisConfigProvider(NonMonitorableDependencyProvider[AsyncEtcd, RedisConfig]):
+    """Provider for the Redis configuration stored in etcd."""
+
+    @property
+    @override
+    def stage_name(self) -> str:
+        return "redis-config"
+
+    @asynccontextmanager
+    @override
+    async def provide(self, setup_input: AsyncEtcd) -> AsyncIterator[RedisConfig]:
+        # TODO: Override UnifiedConfig with etcd values
+        raw_redis_config = await setup_input.get_prefix("config/redis")
+        redis_config = RedisConfig.model_validate(raw_redis_config)
+        log.info("configured redis_config: {0}", safe_print_redis_config(redis_config))
+        yield redis_config
--- a/src/ai/backend/storage/volumes/backends.py
+++ b/src/ai/backend/storage/volumes/backends.py
@@ -1,0 +1,37 @@
+from __future__ import annotations
+
+from collections.abc import Mapping
+
+from .abc import AbstractVolume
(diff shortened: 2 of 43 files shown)


src/ai/backend/storage/dependencies/infrastructure/redis_config.py
```diff
@@ -0,0 +1,32 @@
...
+        return "redis-config"
+
+    @asynccontextmanager
+    @override
+    async def provide(self, setup_input: AsyncEtcd) -> AsyncIterator[RedisConfig]:
+        # TODO: Override UnifiedConfig with etcd values
+        raw_redis_config = await setup_input.get_prefix("config/redis")
+        redis_config = RedisConfig.model_validate(raw_redis_config)
+        log.info("configured redis_config: {0}", safe_print_redis_config(redis_config))
+        yield redis_config
```
HyeockJinKim, line 32: In theory, we should only need to create the config provider for this, but I'll pass it on for now.

src/ai/backend/storage/volumes/backends.py
```diff
@@ -0,0 +1,37 @@
...
+    WekaVolume.name: WekaVolume,
+    GPFSVolume.name: GPFSVolume,  # IBM SpectrumScale or GPFS
+    "spectrumscale": GPFSVolume,  # IBM SpectrumScale or GPFS
+    CephFSVolume.name: CephFSVolume,
+    VASTVolume.name: VASTVolume,
+    EXAScalerFSVolume.name: EXAScalerFSVolume,
+    NoopVolume.name: NoopVolume,
+    HammerspaceVolume.name: HammerspaceVolume,
+    BaseHammerspaceVolume.name: BaseHammerspaceVolume,
+}
```
HyeockJinKim, line 37: I believe this is how it was implemented previously, but we plan to improve this later as well. (I'm writing this for context.)

HyeockJinKim (approved):
Since these are just minor issues, I'll merge the changes for now.

PR #14479, opened 2026-09-10
chore(BA-7769): expect the keypair lookup's error under the row's own type

jopemachine:
## Summary

- #14471 gave `KeypairAccessKeyLookup` a `field_type()`, so an access key naming no keypair now reports `user-keypair_read_not-found` — the row's owning entity and its own field type — instead of the repository's `database_access_not-found`.
- `tests/component/user/test_keypair_ops.py` kept the old expectation, so `TestUpdateMyKeypair::test_update_nonexistent_access_key_raises_not_found` fails.
- The component suite does not run on a push to `main` — that job executes no pytest — which is why `main` stayed green while every pull request whose merge base carries #14471 fails on it. #14460 and #14478 both hit it, with the same assertion.

## Test plan

- [ ] `tests/component/user/test_keypair_ops.py::TestUpdateMyKeypair::test_update_nonexistent_access_key_raises_not_found` passes in `test-component (1/6)`.
- [ ] The rest of that file's `keypair_read_forbidden` assertions are untouched and still pass.

Follow-up to BA-7769, whose change this test was left behind by.


jopemachine:
Superseded by #14478, which merged the same fix (`_GQL_ERR_NO_ROW = "user-keypair_read_not-found"`) as part of its own change. Nothing left to do here.

PR #14483, opened 2026-09-10
feat(BA-7810): cap GraphQL query depth and alias count

fregataa:
## Summary

- Both Strawberry schemas — the authenticated one and the anonymous `public` one — accepted queries of arbitrary depth and width, so a single request had no cost ceiling. Add the library's `QueryDepthLimiter` and `MaxAliasesLimiter` to both.
- Limits are constants (`MAX_QUERY_DEPTH = 10`, `MAX_ALIAS_COUNT = 10`) in `schema.py`. Exceeding either fails validation and the query is not executed.
- The alias cap is not redundant with the depth cap: repeating one field under thousands of aliases keeps a query shallow while multiplying the work, and it applies to the `public` schema too, where the small field count offers no protection.

Measured maxima across WebUI queries are depth 9 and 6 aliases, both of which still pass.

## Test plan

- [x] Verified the thresholds against the library validators directly: depth 9 and 10 pass, 11 is blocked; 6 and 10 aliases pass, 11 is blocked.
- [x] `pants fix` / `pants lint --changed-since=origin/main`
- [x] `pants lint` / `check` / `test --changed-dependents=direct` via the pre-push hook
- [ ] Confirm no real WebUI query exceeds depth 10 — the measurement covered the WebUI bundle only, not external tools or plugins.

Backport: 26.4

Resolves BA-7810

https://claude.ai/code/session_015d4RfANyMk9v2RgtWUozpc

### DIFF
--- a/src/ai/backend/manager/api/gql/schema.py
+++ b/src/ai/backend/manager/api/gql/schema.py
@@ -4,6 +4,7 @@ from typing import override
 import strawberry
 from graphql import GraphQLError
 from graphql.pyutils.undefined import Undefined as GraphQLUndefined
+from strawberry.extensions import MaxAliasesLimiter, QueryDepthLimiter
 from strawberry.federation import Schema
 from strawberry.schema.config import StrawberryConfig
 from strawberry.types import ExecutionContext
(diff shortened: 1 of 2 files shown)


src/ai/backend/manager/api/gql/schema.py
```diff
@@ -1085,6 +1086,11 @@ class Subscription:
     background_task_events = background_task_events
 
 
+# Per-request cost ceilings applied to every schema below.
+MAX_QUERY_DEPTH = 10
+MAX_ALIAS_COUNT = 10
```
HyeockJinKim: I've seen the depth go as high as 9 in the web UI—wouldn't it be better to guarantee at least double that? (Around 20?)

HyeockJinKim (approved):

PR #14498, opened 2026-09-10
refactor(BA-7817): move the storage errors onto EntityError

fregataa:
## Summary
- Eleven errors naming a vfolder, a vfolder invitation, a vfolder permission, an object storage or a VFS storage now inherit `EntityError` or `FieldError` and report the row's type as the code domain. `VFolderPermissionFieldType.owner_type()` is `VFolderEntityType`, so the grant conflict is a field row.
- Storage is mostly traffic with an external service, so twenty errors stay on `BackendAIError`.
- `errors/KNOWLEDGE.md` now states the three tests an error passes to move, and lists every error that fails one — the permission-domain list from BA-7821 folded into the same table.

### The three tests

1. It states something about one **row** — its existence, its state, a conflict with it. An error stating that a *call* failed, that an *input* did not parse, or that a *configuration* offers no such thing names no row.
2. `data/entity/` declares that row's kind, as an `EntityType` or a `FieldType`.
3. `ActionOperationType` can carry the operation it reports (`generic`, `access`, `grant`, `request`, `start`, `execute` have no value there).

Test 1 does most of the work here: `VFolderGone` states the folder's state and moves, while `VFolderCreationFailure` states that the storage proxy refused the call and stays, though both surface during a vfolder operation.

### Outward code changes

| Error | Before | After |
|---|---|---|
| `VFolderGrantAlreadyExists` | `vfolder_grant_already-exists` | `vfolder-vfolder-permission_create_already-exists` |

Granting is creating a `vfolder_permissions` row, which is what the already-merged `PermissionAlreadyGranted` reports as `role-permission_create_already-exists`. The other ten moved errors keep their codes: every `ErrorDomain` value in play (`vfolder`, `vfolder-invitation`, `object-storage`, `vfs-storage`) already matched its entity type's name.

`VFolderOwnerNotFound` now answers **500** instead of 404. It fires from `VFolderCreator.created_in()` after the insert settled, so a folder that landed in no project is a broken account, not a missing resource. Its code is unchanged.

### Errors left on `BackendAIError`

| Error | Which test it fails |
|---|---|
| `QuotaScopeNotFoundError`, `StorageProxyNotFound` | 2 — neither declares a row kind |
| the four `Dotfile*` errors | 2 — dotfiles are a msgpack blob in a `keypairs` column, not rows |
| `VFolderBadRequest`, `VFolderOperationFailed`, `VFolderCreationFailure` | 1 and 3 — the storage-proxy call, under `generic` |
| `StorageProxyConnectionError`, `StorageProxyTimeoutError`, `UnexpectedStorageProxyResponseError` | 1, 2 and 3 — the external call itself, under `request` |
| `VFolderPermissionError`, `VFolderInvalidParameter`, `InsufficientStoragePermission` | 1 — an authorization denial about the caller, or a parameter refused before a row is reached |
| `ModelCardParseError` | 1 — a model-definition file that does not parse, before any model card row exists |
| `VFolderFilterStatusNotAvailable` | 1 — the status-set alias names no entry in a constant map; the row-status half is `VFolderFilterStatusFailed` |
| `UnsupportedStorageTypeError`, `ObjectStorageOperationNotSupported` | 1 — a requested storage type, or a configured one offering no such operation |

## Test plan
- [x] `pants lint` / `check` / `test` in CI
- [x] Constructed every moved error and asserted its code, HTTP status and title
- [x] Checked WebUI: it branches on three vfolder codes — `vfolder_create_already-exists`, `vfolder_access_forbidden`, `vfolder_read_not-found` — plus `storage_access_forbidden`. The first and third belong to moved errors whose codes are unchanged; the other two belong to errors that stay. `vfolder_grant_already-exists` appears nowhere
- [x] Confirmed no error in these three modules is unreferenced, so nothing is deleted here

Resolves BA-7817

https://claude.ai/code/session_0128bKAiecc6WfVK9VBRjxBv

### DIFF
--- a/src/ai/backend/manager/errors/storage.py
+++ b/src/ai/backend/manager/errors/storage.py
@@ -10,6 +10,8 @@ from typing import Any, override
 from aiohttp import web
 
 from ai.backend.common.data.entity.vfolder import VFolderEntityType
+from ai.backend.common.data.entity.vfolder_invitation import VFolderInvitationEntityType
+from ai.backend.common.data.entity.vfolder_permission import VFolderPermissionFieldType
 from ai.backend.common.exception import (
     BackendAIError,
     ErrorCode,
@@ -19,21 +21,18 @@ from ai.backend.common.exception import (
 )
 from ai.backend.manager.actions.types import ActionOperationType
 from ai.backend.manager.errors.base.entity import EntityError, EntityErrorCode
+from ai.backend.manager.errors.base.field import FieldError, FieldErrorCode
 
 from .common import ObjectNotFound
 
 
-class TooManyVFoldersFound(BackendAIError, web.HTTPNotFound):
+class TooManyVFoldersFound(EntityError, web.HTTPNotFound):
     error_type = "https://api.backend.ai/probs/too-many-vfolders"
     error_title = "Multiple vfolders found for the operation for a single vfolder."
 
     @override
-    def error_code(self) -> ErrorCode:
-        return ErrorCode(
-            domain=ErrorDomain.VFOLDER,
-            operation=ErrorOperation.READ,
-            error_detail=ErrorDetail.CONFLICT,
-        )
+    def entity_error_code(self) -> EntityErrorCode:
+        return EntityErrorCode(VFolderEntityType(), ActionOperationType.GET, ErrorDetail.CONFLICT)
 
     def __init__(self, matched_rows: Sequence[Mapping[str, Any]]) -> None:
         serialized_matches = [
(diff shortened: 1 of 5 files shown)


src/ai/backend/manager/errors/storage.py
```diff
@@ -19,21 +21,18 @@
 )
 from ai.backend.manager.actions.types import ActionOperationType
 from ai.backend.manager.errors.base.entity import EntityError, EntityErrorCode
+from ai.backend.manager.errors.base.field import FieldError, FieldErrorCode
 
 from .common import ObjectNotFound
 
 
-class TooManyVFoldersFound(BackendAIError, web.HTTPNotFound):
+class TooManyVFoldersFound(EntityError, web.HTTPNotFound):
```
jopemachine: It seems it is not HTTPNotFound

HyeockJinKim (approved):

PR #14503, opened 2026-09-11
doc(BA-7803): fill the container registry and image adapter scenarios

jopemachine:
## Summary
- Replace the TODO stubs in `api/adapters/container_registry/KNOWLEDGE.md` and `api/adapters/image/KNOWLEDGE.md` with the scenarios each adapter guarantees, following the shape `api/adapters/domain/KNOWLEDGE.md` set in #14448.
- Every public adapter call carries at least one scenario — 6 calls / 33 rows for the container registry, 11 calls / 40 rows for images — so the tests under `tests/scenario` can be written one for one against them.
- Every way a call can be refused gets its own row: the superadmin role, the entity gate, the ownership check `ImageService` runs after the gate, request validation, and the rbac relation gate behind the registry's allowed-project list.

Two asymmetries worth a reviewer's eye, both stated in the documents:
- A registry URL is checked on update but not on create, so an `ftp://` registry can be created and only refuses later when edited.
- An image read answers differently depending on who asks: `hide_agents` blanks the installed-agent information for a non-superadmin.

## Test plan
- [ ] `python3 scripts/knowledge/check.py` passes
- [ ] Scenario tests under `tests/scenario` are written against these rows (follow-up)

Resolves BA-7803

https://claude.ai/code/session_013TYyokzEfpr9AF2NnHtyXB


jopemachine:
Split into #14519 (container registry, BA-7803) and #14520 (image, BA-7837). The concern holds two entities with an adapter each, and one PR of 4.7k lines was too much to review in one sitting.

PR #14504, opened 2026-09-11
fix: drop the scope-entity rows left in the seed fixtures

devyubin:
### Summary
Every fresh install stops at `mgr fixture populate`, because the seed fixtures still seed a table that was dropped.

### Changes

**The fixtures no longer seed `association_scopes_entities`**
- The key and its rows come out of `fixtures/manager/example-users.json` (10 rows) and `fixtures/manager/example-roles.json` (27 rows).
- Reason: #14478 removed the table from the models and added `b3c81f0a49d5` to drop it, but did not touch `fixtures/`. The fixture loader resolves each top-level key against the live metadata, so both files now fail with `Table association_scopes_entities not found in metadata` and take the installer's fixture-loading step down with them.
- Nothing is lost. The dropping migration states that the table "has no reader left" and that "nothing in the association table is unique to it, so the rows are dropped rather than moved" — the scope-entity relation is answered from the virtual entity graph. Both fixtures have carried that graph's `virtual_entities`, `entity_memberships` and `scope_bindings` rows since #14127, and no code outside old migrations references the table.

### Verification
Ran a full develop-mode install against a wiped halfstack, with #14502 applied on top so the earlier app-proxy CLI failure did not mask this one. The install ran to completion, `INSTALL-INFO` included:

```
🟢 Loading fixtures...
🟢 Configuring app-proxy...
🟢 Generating client environ configs...
🟢 Scanning and pulling configured Backend.AI container images...
🟢 Dumping the installation configs...
```

The RBAC seed data landed in full, so removing the rows costs nothing:

| table | rows |
|---|---|
| roles | 13 |
| user_roles | 16 |
| permissions | 730 |
| virtual_entities | 19 |
| entity_memberships | 47 |
| scope_bindings | 26 |

`to_regclass('association_scopes_entities')` returns NULL on that database, confirming the table is genuinely gone rather than merely unreferenced. Populating `main`'s version of either fixture against the same database still raises the error.

### Ordering
Independent of #14502, but an install needs both: that one fails at the app-proxy schema step, which comes first.


devyubin:
Superseded by #14609 and #14657: the former dropped `association_scopes_entities` from `example-users.json`, the latter removed `example-roles.json` entirely in favour of provisioning preset roles from a command. No fixture on main seeds the table any more, so this branch has nothing left to change.

PR #14501, opened 2026-09-10
test(BA-7806): add scenario tests for the deployment adapter's own calls

fregataa:
## Summary
- Write `adapters/deployment/KNOWLEDGE.md` for the deployment's own calls (create, read, search, update, options, sync, delete) and transcribe it into `tests/scenario/bai_scenario/manager/deployment/` — 32 rows over 7 modules.
- Add `SeedDeployment` (lays a deployment through `DeploymentCreator`, no revision), fakes for the schedule client and the coordinator's handler registry, and deployment components.
- Pin behaviours found while transcribing:
  - an unknown id is refused as `NotEnoughPermission` for a plain user; only a superadmin sees `EndpointNotFound`
  - `my_search` is guarded by READ at the caller's own scope, not just narrowed to it
  - `replace_options` replaces the options whole — an empty handler list drops the inherited entries

Not covered here:
| Rows | Why |
|---|---|
| create with an initial revision, current revision, activate / refresh revisions | need revision, vfolder, runtime-variant and policy seeds (overlaps BA-7825) |
| schedule marks left by delete / sync_replicas | `run_scenario` does not hand fakes to `then` (BA-7830) |

## Test plan
- [x] `pants test tests/scenario/bai_scenario/manager/deployment::`
- [x] Generate `report.md` with `scripts/scenario-report.py --split` and commit it

Resolves BA-7806

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015d4RfANyMk9v2RgtWUozpc

PR #14506, opened 2026-09-11
test(BA-7807): add scenario tests for the audit log adapter

jopemachine:
## Summary
- Replace the TODO stub in `api/adapters/audit_log/KNOWLEDGE.md` with the scenarios the adapter guarantees, following the shape `api/adapters/domain/KNOWLEDGE.md` set in #14448 and `api/adapters/deployment/KNOWLEDGE.md` in #14501, then transcribe them into `tests/scenario/bai_scenario/manager/audit_log/`.
- Three reads / 28 rows over three modules: the whole-table `admin_search` behind the superadmin role (a monitor passes, since it is a read), the by-id `batch_load_by_ids` behind the per-entity read permission answered slot by slot (node, refusal, or a gap for an id nothing answers to), and the `scoped_search` behind the same permission that refuses the whole page when one named entity is unreadable.
- Add `SeedAuditRecord`, which lays a record under the entity it is about through the field creator the monitors write with (no adapter call makes one), setting `triggered_by` from a user an earlier step laid.
- Each returned node is compared whole (ids skipped, owner and actor read as coming from the laid rows), not by operation alone.

Behaviours pinned while transcribing:
- A monitor passes the whole-table search but neither the by-id read nor the scoped search — those ask the permission graph, where a monitor holds nothing it was not granted.
- There is no `my_` path: a user reading the records they themselves triggered, or the records about themselves, still needs read permission on their own user entity.
- The scoped read refuses the whole page as `NotEnoughPermission` when any named entity is unreadable; the by-id read refuses only that slot and answers the rest.
- Naming an entity id nothing answers to is refused as `NotEnoughPermission` for a superadmin too — the own check holds no bits on an entity without a node — never a not-found. An absent id in the by-id read is a gap for anyone, since no entity answers for it.

Not covered here:
| Rows | Why |
|---|---|
| A by-id read of a record naming no entity (a global action's record, answered for by the global entity) | no seed lays an ownerless record — the seeder has no dangling-field entry; left as a gap rather than adding one here |
| CSV export of audit logs (and the other visibility wirings) | they live in `api/rest/export/handler.py`, which calls processors directly — not an `api/adapters/` adapter |

## Test plan
- [x] The scenario suite under `tests/scenario/bai_scenario/manager/audit_log` runs green in CI
- [x] Generate `report.md` from the CI run's `scenarios.jsonl` with `scripts/scenario-report.py --split` and commit it

Resolves BA-7807

https://claude.ai/code/session_01QM7T3buh1thJkiXFELHRgE


src/ai/backend/manager/api/adapters/audit_log/KNOWLEDGE.md
```diff
@@ -1,26 +1,114 @@
...
+|---|---|---|---|
+| 읽기 권한을 받은 사용자가 엔티티를 지정해 검색한다 | 엔티티 둘에 기록이 하나씩, 한쪽에만 읽기 권한 있음 | 권한 있는 엔티티를 지정해 검색 | 그 엔티티의 기록만 반환된다 |
+| 지정한 엔티티 안에서 실행된 동작의 기록도 반환된다 | 프로젝트 하나, 그 프로젝트를 스코프로 실행된 동작의 기록, 프로젝트에 읽기 권한 있음 | 그 프로젝트를 지정해 검색 | 그 기록이 반환된다 |
+| `triggered_by` 사용자로 지정해 검색한다 | 두 사용자가 남긴 기록이 하나씩, 한 사용자에 읽기 권한 있음 | 그 사용자를 `triggered_by` 사용자로 지정해 검색 | 대상 엔티티가 무엇이든 그 사용자가 실행한 기록만 반환된다 |
+| 엔티티 여럿을 지정하면 합쳐서 반환된다 | 엔티티 둘에 기록이 하나씩, 둘 다 읽기 권한 있음 | 둘을 함께 지정해 검색 | 두 기록이 다 반환되고, 최근 것이 먼저다 |
+| 지정한 범위 안에서 필터가 적용된다 | 한 엔티티에 `success` 기록과 `denied` 기록, 읽기 권한 있음 | 그 엔티티를 지정하고 상태 필터로 검색 | 지정한 상태의 기록만 반환된다 |
+| 페이지 크기를 생략한다 | 한 엔티티에 기록 11개, 읽기 권한 있음 | 페이지 크기 없이 검색 | 10건이 반환되고 다음 페이지가 있다고 응답한다 |
+| 볼 수 없는 엔티티가 하나라도 섞이면 전부 거부된다 | 엔티티 둘에 기록이 하나씩, 한쪽에만 읽기 권한 있음 | 둘을 함께 지정해 검색 | 권한 부족으로 거부 |
+| 읽기 권한이 없는 사용자가 지정해 검색한다 | 엔티티 하나에 기록 하나, 권한 없음 | 그 엔티티를 지정해 검색 | 권한 부족으로 거부 |
+| 모니터 역할 사용자가 권한 없이 지정해 검색한다 | 같은 기록, 모니터 역할, 권한 없음 | 그 엔티티를 지정해 검색 | 권한 부족으로 거부 |
```
HyeockJinKim, line 81: In `monitor`, all read operations should be possible even without access permissions, but it looks like this isn't working as intended. We'll need to fix this later.

jopemachine: Filed as BA-7952, fixed in #14713: the permission read ops now give a monitor the read bit on every entity and scope that has a node, next to the superadmin's every bit. Once that merges, this row and the by-id one above ("모니터 역할 사용자가 권한 없이 조회한다") flip to the superadmin's answer, along with `TheMonitorRoleGetsNoScopeForFree` and its by-id counterpart.

HyeockJinKim (approved):
I'll merge them for now.

PR #14507, opened 2026-09-11
test(BA-7809): add scenario tests for the resource policy adapter

jopemachine:
Resolves BA-7809

## Summary
- Write `api/adapters/resource_policy/KNOWLEDGE.md` for the keypair, user and project policies and transcribe it into `tests/scenario/bai_scenario/manager/resource_policy/` — 83 rows over 6 modules, every public adapter call (17 calls, 20 wirings) reached.
- One adapter carries three policies of the same shape, so `components/resource_policy.py` holds one class per policy family — the seed that lays one, the calls that read and write one, the places its node has — and each row is written once and run against every family it applies to. Rows a single family owns (unlimited slot, priority cap, keypair-owner filter, reading one's own policy) name it.
- Seeds added: a project policy under a name of its own (the pinned one is what every personal project looks for), a keypair laid under an existing user, and the nullable limits the editing rows clear.

Behaviours the rows pin, worth a reviewer's eye:
- A name nobody has is refused as *cannot resolve*, not *not found* — for the superadmin too. The lookup hides whether the name exists from everyone, so this adapter has no not-found row.
- Clearing a nullable limit clears it; clearing a non-nullable one is read as omitted and leaves the value as it was.
- Purge has no conflict check; a policy still referenced by a keypair, user or project is stopped by the foreign key, so the refusal is the repository's constraint-violation error, as is a duplicate name or an out-of-range priority cap on create.
- Reading one's own policy is guarded by READ at the caller's own scope, not merely narrowed to it. Among several keys the default-marked one names the keypair policy; an inactive default key yields to an active one; no active key finds nothing.

## Test plan
- [x] `test-scenario` on CI: 83 scenarios, 0 failing
- [x] `report.md` generated from the CI run's `scenarios.jsonl` with `scripts/scenario-report.py --split` and committed


jopemachine:
Split into a stack of three, one per policy: #14515 (project) → #14516 (keypair) → #14517 (user). Same content, same rows.

PR #14508, opened 2026-09-11
test(BA-7805): add scenario tests for the prometheus query preset adapters

jopemachine:
Resolves BA-7805

## Summary
- Replace the TODO stubs in `api/adapters/prometheus_query_preset/KNOWLEDGE.md` and `api/adapters/prometheus_query_preset_category/KNOWLEDGE.md` with the scenarios each adapter guarantees — 8 calls / 55 rows for the preset, 5 calls / 19 rows for the category — in the shape `api/adapters/domain/KNOWLEDGE.md` and `api/adapters/deployment/KNOWLEDGE.md` (#14501) set.
- Transcribe them into `tests/scenario/bai_scenario/manager/prometheus_query_preset/` (45 rows over 8 modules) and `.../prometheus_query_preset_category/` (18 rows over 5 modules). Every public call of both adapters is called by at least one row.
- Add `SeedPreset` / `SeedPresetIn` / `SeedCategory` (rows laid through the creators alone, so a template the renderer would refuse can still be stored) and `FakePrometheusClient` (answers one sample for every query and keeps what it was asked).

- Both entities are global, so no role ever reaches a single preset or category: the entity-gated calls pass only for a superadmin, or for anyone once enforcement is switched off.
- Reading a preset is open to every authenticated caller while executing it is entity-gated — a user who can read a definition cannot run it.
- Preview is the only role-gated read here, so the monitor role passes it and is refused on create.
- Template validation runs only when the request carries a template: a stored template the renderer refuses does not block renaming.
- Changing only the filter labels keeps the group labels, though both share one column.
- A preset name is not unique while a category name is; neither the unique nor the foreign-key violation is mapped to a domain error, so the rows expect the storage-level errors.
- The three container-metric wirings of the metric concern (`global_search_container_metrics`, `public_search_container_metric_metadata`, `search_user_container_metrics`) have no adapter and are consumed by legacy GraphQL only, so they are named in the preset document and kept out of the tables.

Not covered here:
| Rows | Why |
|---|---|
| execute: instant vs range query, the window fallback (request → preset → server default), the allowed labels reaching the query, unrestricted labels; preview: the server default window | read what the stand-in received — `run_scenario` does not hand fakes to `then` (BA-7830) |
| execute / preview: Prometheus refusing the query | need the stand-in scripted per row, which the runner has no channel for yet |
| category: purging a category empties the category of the presets pointing at it | `then` cannot read the presets back after the call |

## Test plan
- [x] `pants lint` / `pants check` on `tests/scenario::`
- [x] The scenario job of CI runs both components green (45 + 18 rows)
- [x] Generate the two `report.md` with `scripts/scenario-report.py --split` from that run and commit them


jopemachine:
Split by entity into #14514 (prometheus_query_preset_category, BA-7835) and #14518 (prometheus_query_preset, BA-7836), stacked in that order.

PR #14509, opened 2026-09-11
test(BA-7804): add scenario tests for the merged app config read

jopemachine:
## Summary
- Replace the TODO stub in `api/adapters/app_config/KNOWLEDGE.md` — the merged read, an entity with no row of its own — with the scenarios the adapter guarantees, following the shape `api/adapters/domain/KNOWLEDGE.md` set in #14448 and the sibling #14503.
- Transcribe them into `tests/scenario/bai_scenario/manager/app_config/` — 20 rows over 2 modules (reading_mine, reading_public), both public calls exercised — and commit the `report.md` the run produces.
- The app config seeds lay a definition, an allow-list entry and a fragment through the manager's own write specs, and `components/app_config.py` holds the design nest and the users a row calls as. The sibling PRs #14511 (definition), #14512 (allow-list) and #14513 (fragment) carry byte-identical copies of these files, so the four merge in any order; whichever lands first brings them and the rest apply as no-ops. The seeder's upserter support this relies on is already in `main` (#14529); nothing here touches the framework.

The merged read's rows are of a different shape from every other adapter's: seed the three tables, read, compare the merge whole. There is no create-read-delete row; what a purge does to the merge is answered by the fragment and allow-list purge sections.

What the sibling PRs carry a copy of:
| Piece | What |
|---|---|
| `seeds/app_config/` | `SeedDefinition`, `SeedAllowListEntry`, `SeedFragmentOf`, `SeedPublicFragment`; a fragment seed reads the allow-list entry, not the definition, so the report shows the write gate it rests on |
| `components/app_config.py` | `ADesignOf` (definition + entries), and the users a row calls as — granted on their own scope, or on a domain |

Open questions of the BA-7804 research note this read settles, stated in the document:

| Question | What the code says |
|---|---|
| how the merge treats lists and nulls | lists replace wholesale, an explicit null overwrites |
| whether the anonymous read reveals which names exist | no — a missing definition and a missing fragment answer the same empty config |
| duplicate allow-list entries and rank ties | `(config_name, scope_type)` is unique; equal ranks have no tiebreaker — flagged as a decision |

## Test plan
- [x] `pytest tests/scenario/bai_scenario/manager/app_config` — 20 passed
- [x] `scripts/scenario-report.py --split` committed; `--verify` level across two runs
- [x] mypy and ruff on the changed targets

Resolves BA-7804

https://claude.ai/code/session_011BnvoxSA2RSKgBkJ1yvC9P


HyeockJinKim (approved):
I'll merge them for now.

PR #14511, opened 2026-09-11
test(BA-7832): add scenario tests for the app config definition adapter

jopemachine:
## Summary
- Replace the TODO stub in `api/adapters/app_config_definition/KNOWLEDGE.md` with the scenarios the adapter guarantees, following the shape `api/adapters/domain/KNOWLEDGE.md` set in #14448.
- Transcribe them into `tests/scenario/bai_scenario/manager/app_config_definition/` — 18 rows over 5 modules (registering, reading, reading_many, searching, purging), all 5 public calls exercised — and commit the `report.md` the run produces.
- Carries byte-identical copies of `seeds/app_config/` and `components/app_config.py` — the same files as #14509, #14511, #14512 and #14513 — so the four PRs are independent and merge in any order; whichever lands first brings the files and the rest apply as no-ops. Adds nothing to the framework.

Worth a reviewer's eye:
- A definition belongs to no scope, so no role reaches one: every door reads as the superadmin passing and anyone else refused, the way the domain document does it. The batch read refuses a plain user per id — the missing id included — and hands the empty slot only to the superadmin.
- Registering a duplicate name is refused as a raw constraint violation rather than a domain error; the document flags it as a decision to make.
- Search keeps only the superadmin check and the unfiltered listing. Filtering, ordering and pagination are shared search contracts, not this adapter's, so they have no scenarios here; the document says so.
- Purging a definition leaves the RBAC nodes of its cascaded allow-list entries and fragments behind (BA-7879). The document records it as a limitation; the scenarios check the adapter's answer only.

## Test plan
- [x] `pytest tests/scenario/bai_scenario/manager/app_config_definition` — 18 passed
- [x] `scripts/scenario-report.py --split` committed; `--verify` level across two runs

Resolves BA-7832

https://claude.ai/code/session_011BnvoxSA2RSKgBkJ1yvC9P


HyeockJinKim (approved):
I'll merge them for now.

PR #14514, opened 2026-09-11
test(BA-7835): add scenario tests for the prometheus query preset category adapter

jopemachine:
Resolves BA-7835

## Summary
- Replace the TODO stub in `api/adapters/prometheus_query_preset_category/KNOWLEDGE.md` with the scenarios the adapter guarantees — 5 calls / 19 rows — in the shape `api/adapters/domain/KNOWLEDGE.md` and `api/adapters/deployment/KNOWLEDGE.md` (#14501) set.
- Transcribe 18 of them into `tests/scenario/bai_scenario/manager/prometheus_query_preset_category/` (5 modules) and add the `report.md` split from the CI run. Every public call of the adapter is called by at least one row.
- Add `SeedCategory` and the caller helper (`lay_someone`). #14518 carries its own copies of both so the two PRs land independently; whichever merges second can drop its copy.

- The category is global, so no role ever reaches one: purge and create pass only for a superadmin.
- Reads are open to every authenticated caller, so an unknown id is refused as not found for anyone.
- A category name is unique but the violation is not mapped to a domain error, so the row expects the storage-level unique violation.

Not covered here:
| Row | Why |
|---|---|
| purging a category empties the category of the presets pointing at it | `then` sees only the answer, and the category adapter cannot read presets; the seeder lays rows through creators only, so a preset table cannot lay a purged category either |

## Judgment calls

Decisions a reviewer may want to overturn; none of them is settled by a rule I could find.

| Where | What I decided | The alternative |
|---|---|---|
| `test_reading.py`, `test_searching.py` | Two rows call the adapter with no user context and expect the authentication refusal (`UserNotFound`). No sibling table has such rows. | Drop them as a validator concern rather than an adapter guarantee. |
| `test_creating.py` | The duplicate-name row expects the storage-level `UniqueConstraintViolationError`, because nothing maps it to a domain error. | Map it to a domain error first (a `Creator.integrity_error_checks` entry) and expect that. |
| `components/prometheus_query_preset_category.py` | `lay_someone` (a caller in a domain, with no entity of its own) lives in this component's module, duplicated in #14518. | `components/domain.py`, which already owns `SomeoneOf`, shared by both. |
| KNOWLEDGE.md, "아직 적지 않은 것" | Names `batch_load_fields` (inherited from `BaseAdapter`, no field rows here) as the one call the run leaves alone, since the report lists it. The deployment document does not mention it. | Leave it out as not this adapter's own call. |

Split out of #14508. Independent of #14518 (prometheus_query_preset), which covers the sibling entity.


HyeockJinKim (approved):
I'll merge them for now.

PR #14519, opened 2026-09-11
test(BA-7803): write the container registry adapter scenarios and their tests

jopemachine:
Split out of #14503, which held both entities of this concern. The image adapter is #14520.

## Summary
- Replace the TODO stub in `api/adapters/container_registry/KNOWLEDGE.md` with what the adapter guarantees, in the shape `adapters/domain/KNOWLEDGE.md` set in #14448, and transcribe it into `tests/scenario/bai_scenario/manager/container_registry/` — 34 rows over six modules, one per call.
- Add `AllowProject`, the write spec that links a project to a registry, so the rows about removing a link start from one that is really there.
- The given hands a caller either scope of the allowed-project pair, or neither: the relation gate asks for both, so one-sided is a row of its own.

## Coverage
| Call | Rows |
|---|---:|
| `admin_create` | 4 |
| `admin_search` | 4 |
| `batch_load_by_ids` | 3 |
| `admin_update` | 7 |
| `apply_allowed_groups` | 12 |
| `admin_delete` | 4 |

## What a reviewer should look at
The document is the claim; the tests are its transcription, and CI checks the transcription against a real database. Reading `KNOWLEDGE.md` (118 lines) plus the coverage table above is the review. Two rows say something surprising and are worth a second look:

- A registry URL is checked on update but never on create, so an address the update path refuses still creates a registry.
- That check looks only for a host. A scheme it does not know has `http://` prepended before parsing, so `ftp://reg.example` passes, as does a string with spaces. Only an empty host is refused. The document asks whether that should stay.

## Test plan
- [ ] The scenario suite for this entity is green in CI
- [ ] `report.md` generated from the run and committed

Resolves BA-7803

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013TYyokzEfpr9AF2NnHtyXB

Issue #10, opened 2017-09-06
Rename to "backend.ai"


achimnol:

achimnol:
Now all done, except internal naming of sorna-media Javascript library. I will revisit it later.

PR #14520, opened 2026-09-11
test(BA-7837): write the image adapter scenarios and their tests

jopemachine:
Split out of #14503, which held both entities of this concern. The container registry adapter is #14519.

## Summary
- Replace the TODO stub in `api/adapters/image/KNOWLEDGE.md` with what the adapter guarantees, in the shape `adapters/domain/KNOWLEDGE.md` set in #14448, and transcribe it into `tests/scenario/bai_scenario/manager/image/` — 54 rows over six modules for eleven of the twelve calls it offers; `scoped_search` has no row yet.
- Extend `SeedImage` with the status and the customized-owner pair, and add `SeedAlias`, so a scenario can lay an image its caller owns.
- The role that reaches an image sits on its registry, because that is the scope an image is created under. The agent registry and the image cache stay unwired: none of the twelve calls read them.

## Coverage
| Call | Rows |
|---|---:|
| `admin_search`, `admin_search_images_gql`, `admin_search_image_aliases` | 15 |
| `batch_load_by_ids`, `batch_load_aliases_by_ids` | 7 |
| `admin_forget`, `admin_restore` | 11 |
| `admin_purge` | 6 |
| `admin_alias`, `admin_dealias` | 7 |
| `admin_update` | 8 |

## What a reviewer should look at
The document is the claim; the tests are its transcription, and CI checks the transcription against a real database. Reading `KNOWLEDGE.md` (140 lines) plus the coverage table above is the review.

This adapter splits its calls across three gates, and the third is the one worth checking:

- The superadmin role guards searching, aliasing and editing.
- The entity permission guards forgetting, restoring and purging.
- An ownership check runs *after* that gate. An image is owned only when it is customized and the caller made it, so a shipped image refuses every caller but a superadmin — even one holding permission on it.

Two rows say what the run does rather than what the code looked like it would do, and both look like bugs worth their own issue rather than something to fix here:

- Restoring cannot reach a forgotten image. It is looked up where only a live image is visible, so `admin_restore` answers that the image is not there. Only an image that was never forgotten can be "restored", which changes nothing.
- Forgetting cannot reach an image a purge is working through, for the same reason, so the check that refuses a status write mid-purge is unreachable through this adapter.

## Test plan
- [ ] The scenario suite for this entity is green in CI
- [ ] `report.md` generated from the run and committed

Resolves BA-7837

https://claude.ai/code/session_013TYyokzEfpr9AF2NnHtyXB


changes/14519.test.md
```diff
@@ -0,0 +1 @@
+Document what the container registry adapter guarantees, as scenario tables, and add the scenario tests that transcribe them.
```
jopemachine: Delete this

src/ai/backend/manager/api/adapters/image/KNOWLEDGE.md
```diff
@@ -1,26 +1,175 @@
...
+| 삭제하지 않은 것을 복원한다 | 살아 있는 이미지, 호출자는 슈퍼관리자 | 복원 | 살아 있는 상태 그대로 |
+| 존재하지 않는 ID를 소프트 삭제한다 | 다른 이미지만 있음, 호출자는 슈퍼관리자 | 소프트 삭제 | 대상을 찾을 수 없어 거부 |
+| 존재하지 않는 ID를 복원한다 | 다른 이미지만 있음, 호출자는 슈퍼관리자 | 복원 | 대상을 찾을 수 없어 거부 |
+| 아무 권한도 받지 않은 사용자가 소프트 삭제한다 | 커스터마이즈되지 않은 이미지, 아무 권한도 없음 | 소프트 삭제 | 권한 부족으로 거부 |
+| 아무 권한도 받지 않은 사용자가 복원한다 | 커스터마이즈되지 않은 이미지, 아무 권한도 없음 | 복원 | 권한 부족으로 거부 |
+| 일반 이미지는 권한을 받은 사용자도 삭제하지 못한다 | 커스텀 이미지가 아님, 호출자에게 이미지 삭제 권한 있음 | 소프트 삭제 | 커스텀 이미지 작성자 검사에서 거부 |
+| RBAC 권한 검사를 꺼도 작성자 검사는 남는다 | RBAC 권한 검사 비활성화, 커스텀 이미지가 아님 | 소프트 삭제 | 커스텀 이미지 작성자 검사에서 거부 |
+| 완전 삭제 중인 이미지를 소프트 삭제한다 | 완전 삭제 중인 이미지, 호출자는 슈퍼관리자 | 소프트 삭제 | 대상을 찾을 수 없어 거부 |
+
+9번과 10번은 이미지 권한 검사와 커스텀 이미지 작성자 검사가 별개임을 확인한다.
```
jopemachine, line 98: Referring to them simply as “#9” and “#10” still makes it difficult for the reader to understand what they mean.

jopemachine: Replaced every row-number reference in the prose with the scenario name it points at (1efe3a5933).

HyeockJinKim (approved):

PR #14510, opened 2026-09-11
test(BA-7802): add the shared components the system concern scenarios use

jopemachine:
## 📚 Stacked PRs

Part of the system concern scenario stack (BA-7802). **Merge in order (bottom-up):**

1. 👉 **#14510 — `test(BA-7802)`: shared components the system concern scenarios use** ← you are here
2. #14521 — `test(BA-7802)`: runtime variant
3. #14522 — `test(BA-7802)`: runtime variant preset
4. #14523 — `test(BA-7802)`: resource slot type
5. #14524 — `test(BA-7802)`: retention policy
6. #14525 — `test(BA-7802)`: login client type
7. #14526 — `test(BA-7802)`: client IP masking
8. #14527 — `test(BA-7802)`: secret

> **Fork at #14510**: #14521 → #14522 is ordered; #14523–#14527 each sit on #14510 alone and may merge in any order after it.

Part of BA-7802

## Summary
- Add `tests/scenario/bai_scenario/components/system.py`, the one piece every system-concern scenario table shares: a caller of a given role and nothing else (`SomeoneAlone`, `lay_a_caller`), the enforcement switch the entity gate reads (`ENFORCEMENT`), and the marker an edit uses for a place it must leave as laid (`KEPT`).
- A system entity is created in no scope, so its tables lay no place — only the caller and the rows the call reads. That is why this is all the stack shares; each entity PR above brings its own seed, components, `KNOWLEDGE.md`, tests and `report.md`.

Beside the stack, #14528 fills `service_catalog/KNOWLEDGE.md` alone: no write spec lays a catalog row (the registration event handler inserts it directly), so it has no tests yet and sits on `main` by itself.

The twelve `global` wirings of the concern (`EtcdConfigProcessors`, `ManagerAdminProcessors`) have no adapter package and therefore no `KNOWLEDGE.md` slot; they are out of this stack.

## Test plan
- [x] `pants check lint` on the new module
- [ ] CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Snp9H2jEVbB2s1C6ixUHGn

PR #14528, opened 2026-09-11
doc(BA-7802): fill the service catalog adapter scenarios

jopemachine:
Part of BA-7802

## Summary
- Fill `api/adapters/service_catalog/KNOWLEDGE.md` with the scenarios its one call guarantees: a search behind the superadmin role that the monitor role also passes, with group and status filters and the default page size.
- No tests yet: no write spec lays a catalog row — the registration event handler inserts it directly — and `tests/scenario` lays rows through write specs only. The document says so, and the rows wait for a way to lay one.

Independent of the stack under #14510; sits on `main` by itself.

## Test plan

### DIFF
--- a/changes/14528.doc.md
+++ b/changes/14528.doc.md
@@ -1,0 +1,1 @@
+Document what the service catalog adapter guarantees, as a scenario table the integration tests are written against one for one.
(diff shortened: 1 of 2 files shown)


jopemachine:
Superseded: the stack was dissolved into independent PRs, one issue each — #14535 (BA-7843 runtime variant), #14536 (BA-7844 preset), #14538 (BA-7845 resource slot type), #14542 (BA-7846 retention policy), #14543 (BA-7847 login client type), #14544 (BA-7848 client IP masking), #14545 (BA-7849 secret), #14546 (BA-7850 service catalog).

PR #14535, opened 2026-09-11
test(BA-7843): add scenario tests for the runtime variant adapter

jopemachine:
Resolves BA-7843

## Summary
- Fill `api/adapters/runtime_variant/KNOWLEDGE.md` with the scenarios the adapter guarantees, following the shape `api/adapters/domain/KNOWLEDGE.md` set in #14448, and transcribe it into `tests/scenario/bai_scenario/manager/runtime_variant/` — 29 rows over five modules (creating, reading, searching, editing, retiring) — with `seeds/runtime_variant/runtime_variant.py` and `components/runtime_variant.py`.
- All eight calls are exercised, `resolve_by_name` and `batch_load_by_ids` included.

What the rows pin:
- The reads (by id, by name, by many ids) ask for authentication alone; a user granted nothing reads, and an unknown name resolves to not-found because no permission check follows the lookup.
- Create is behind the superadmin role; update and delete are behind the entity gate, which no role can reach since the variant is created in no scope.
- Renaming to a taken name is refused with the raw unique-violation error, not the conflict error the create path maps it to.
- `bulk_delete` is a loop, not a transaction: an unknown id behind a known one is refused as not-found after the known one is gone.

Deferred, stated in the document: deleting a variant with presets attached — the preset's variant column has a foreign key in the alembic migration but not in the ORM, so the result differs between a migrated database and one built by `schema oneshot`.

`tests/scenario/bai_scenario/components/system.py` — the caller-only situation every system-concern table shares — travels in each of the eight sibling PRs byte for byte (#14535, #14536, #14538, #14542, #14543, #14544, #14545). Whichever merges first, the rest rebase clean.

## Test plan
- [x] The rows passed in CI run 34558080395 of the unsplit branch (0 failing); `report.md` is generated from that run's artifacts with `scripts/scenario-report.py --split`


src/ai/backend/manager/api/adapters/runtime_variant/KNOWLEDGE.md
```diff
@@ -1,26 +1,132 @@
...
+|---|---|---|---|
+| 슈퍼관리자가 삭제한다 | 변형 하나, 전역 역할 있음 | 삭제 | 삭제한 id를 담은 응답 |
+| 존재하지 않는 id를 삭제한다 | 다른 변형만 있음, 전역 역할 있음 | 삭제 | 대상을 찾을 수 없어 거부 |
+| 아무 권한도 없는 사용자가 삭제한다 | 같은 변형, 아무 권한도 없음 | 삭제 | 권한 부족으로 거부 |
+| 권한 검사를 끄면 권한 없이도 삭제된다 | 권한 검사 비활성화, 아무 권한도 없음 | 삭제 | 삭제한 id를 담은 응답 |
+| 여럿을 한 번에 삭제한다 | 변형 둘, 전역 역할 있음 | 두 id를 한 번에 | 요청한 id의 수를 담은 응답 |
+| 없는 id가 섞인 목록을 삭제한다 | 변형 하나, 전역 역할 있음 | 있는 id 뒤에 없는 id를 붙여 한 번에 | 대상을 찾을 수 없어 거부 |
+| 아무 권한도 없는 사용자가 여럿을 삭제한다 | 변형 둘, 아무 권한도 없음 | 두 id를 한 번에 | 권한 부족으로 거부 |
+
+일괄 삭제는 단건 삭제를 id마다 반복하는 것이다. 한 트랜잭션이 아니어서 중간에 거부되면 앞의
```
jopemachine, line 107: This part seems to need improvement.

src/ai/backend/manager/api/adapters/runtime_variant/KNOWLEDGE.md
```diff
@@ -1,26 +1,132 @@
...
+것은 삭제된 채 남는다. 여섯 번째 시나리오의 거부가 그 경우이고, 남은 것은 응답에 담기지
+않으므로 시나리오는 거부만 확인한다. 응답의 수는 삭제한 수가 아니라 요청한 id의 수인데, 끝까지
+성공했을 때만 응답이 오므로 둘이 같다.
+
+preset이 딸린 변형을 삭제하는 시나리오는 아직 두지 않는다. preset의 변형 컬럼에 외래 키가
+선언된 곳과 선언되지 않은 곳이 갈려 있어, 마이그레이션으로 만든 데이터베이스에서는 preset이
+함께 사라지고 스키마를 한 번에 만든 데이터베이스에서는 남는다. 어느 쪽이 맞는지 정한 뒤
+시나리오를 적는다.
```
jopemachine: This part also seems to need some improvement.

jopemachine: Rewritten in 4837537453. The purge now clears the presets before the variant, so the paragraph states that and points at the scenario, which is checked in commented out in `test_retiring.py` behind BA-7931 (ORM column lacks the FK the migration declares) and the preset seed from #14536.

src/ai/backend/manager/api/adapters/runtime_variant/KNOWLEDGE.md
jopemachine: Filed as BA-7937: wire bulk_delete as a partial bulk purge (per-id answers, `items`/`failed` like role_preset's bulk_purge) instead of the loop over the single purge. The paragraph points at it since 1648882ec7; the three bulk rows stay as they are until then.

src/ai/backend/manager/api/adapters/runtime_variant/KNOWLEDGE.md
jopemachine: BA-7931 is now a 작업 (not a bug): declare the FK on the ORM column that the migration already has.

src/ai/backend/manager/api/adapters/runtime_variant/KNOWLEDGE.md
```diff
@@ -1,26 +1,127 @@
...
+생성한다" 시나리오가 둘을 함께 검사한다.
+
+이름이 비어 있거나 길이 제한을 넘는 요청은 시나리오로 두지 않는다. 요청 타입이 이미 막으므로
+어댑터가 보장하는 것이 아니다.
+
+## 조회
+
+| 시나리오 | 상황 | 요청 | 결과 |
+|---|---|---|---|
+| 아무 권한도 없는 사용자가 id로 조회한다 | 변형 하나, 아무 권한도 없음 | id로 조회 | 그 변형 전체 |
```
seedspirit: It seems there are some parts where the terminology sounds a bit awkward, so could you please take a look?

HyeockJinKim (approved):

PR #14536, opened 2026-09-11
test(BA-7844): add scenario tests for the runtime variant preset adapter

jopemachine:
Resolves BA-7844

## Summary
- Fill `api/adapters/runtime_variant_preset/KNOWLEDGE.md` and transcribe it into `tests/scenario/bai_scenario/manager/runtime_variant_preset/` — 34 rows over five modules — with `seeds/runtime_variant/preset.py` and `components/runtime_variant_preset.py`.
- A preset is laid under a variant, so `seeds/runtime_variant/runtime_variant.py` and `components/runtime_variant.py` come along; #14535 carries the same two files byte for byte.
- All six calls are exercised; the five value types each get a create row, and the version filter is laid with all four bound combinations.

What the rows pin:
- The rank is filled in by the insert: the first preset of a variant is 100 and the next is 100 higher.
- The value-type rules are checked by the request type on create and by the service against the stored row on update, so the update rows refuse with `InvalidAPIParameters` where the create rows are not rows at all.
- `batch_load_by_ids` is a search narrowed to the ids, so an unknown id is an empty slot rather than a refusal.

Deferred, stated in the document: a preset for an unknown variant (the ORM/migration foreign-key split), and changing only the value type — the service re-checks the default only when a default is sent, so this leaves a mismatched default in place.

`tests/scenario/bai_scenario/components/system.py` — the caller-only situation every system-concern table shares — travels in each of the eight sibling PRs byte for byte (#14535, #14536, #14538, #14542, #14543, #14544, #14545). Whichever merges first, the rest rebase clean.

## Test plan
- [x] The rows passed in CI run 34558080395 of the unsplit branch (0 failing); `report.md` is generated from that run's artifacts with `scripts/scenario-report.py --split`


src/ai/backend/manager/api/adapters/runtime_variant_preset/KNOWLEDGE.md
```diff
@@ -1,26 +1,123 @@
...
+## 수정
+
+| 시나리오 | 상황 | 요청 | 결과 |
+|---|---|---|---|
+| 슈퍼관리자가 이름만 바꾼다 | 프리셋 하나, 전역 역할 보유 | 이름 수정 | 이름만 변경되고 나머지는 유지됨 |
+| 설명을 지운다 | 설명이 있는 프리셋, 전역 역할 보유 | 설명을 비우도록 수정 | 설명이 없어짐 |
+| 순위를 바꾼다 | 프리셋 하나, 전역 역할 보유 | 순위 수정 | 순위가 변경된 노드 |
+| 변경할 값을 지정하지 않는다 | 프리셋 하나, 전역 역할 보유 | 빈 수정 요청 | 아무것도 바뀌지 않은 노드 |
+| 값 종류만 `flag`로 바꾼다 | 대상이 `env`인 프리셋, 전역 역할 보유 | 값 종류를 `flag`로 수정 | 잘못된 입력으로 요청이 거부됨 |
+| 기본값만 바꿔 값 종류와 어긋나게 한다 | 값 종류가 정수인 프리셋, 전역 역할 보유 | 기본값을 숫자가 아닌 문자열로 수정 | 잘못된 입력으로 요청이 거부됨 |
```
seedspirit: Since this content is similar to the updater test, it seems unnecessary for this particular scenario test.

jopemachine: This is different from an update test succeeding, because this case intentionally changes the value type to make the test fail.

src/ai/backend/manager/api/adapters/runtime_variant_preset/KNOWLEDGE.md
```diff
@@ -1,26 +1,130 @@
...
+`runtime_variant` 컬럼에 외래 키가 마이그레이션에는 있지만 ORM에는 없다. 따라서 마이그레이션으로
+만든 데이터베이스에서는 요청이 거부되지만 스키마를 한 번에 만든 데이터베이스에서는 생성된다. 올바른
+동작을 정한 뒤 시나리오를 작성한다.
+
+## 조회
+
+| 시나리오 | 상황 | 요청 | 결과 |
+|---|---|---|---|
+| 아무 권한도 없는 사용자가 ID로 조회한다 | 프리셋 하나, 아무 권한도 없음 | ID로 조회 | 해당 프리셋 전체 |
+| 존재하지 않는 ID로 조회한다 | 다른 프리셋만 있음 | ID로 조회 | 대상을 찾을 수 없어 거부 |
```
fregataa, line 59: Let's describe what the code of rejection response is - such as not enough permission(forbidden) or not found

src/ai/backend/manager/api/adapters/runtime_variant_preset/KNOWLEDGE.md
jopemachine: Agreed after all — both update rows (flag on an env preset, a default that does not fit the stored type) are decided in the service alone and `tests/unit/manager/services/runtime_variant_preset` already covers them. Dropped in f78611e255, along with the per-value_type create rows (moved to the DTO unit test).

HyeockJinKim (approved):

PR #14547, opened 2026-09-11
test(BA-7854): let the user policy seed carry a concurrent-login cap

jopemachine:
Resolves BA-7854

## Summary
- `SeedUserPolicy` takes a `max_concurrent_logins` and says so in its detail. It is the only change to a shared scenario seed the resource policy scenarios need, so it goes in on its own.
- The cap is the only value of a user policy an update may clear, so the row that clears it (BA-7853) needs a seeded policy holding one. The default stays `None`, so every row laid so far — and every existing `report.md` — reads the same.

## Test plan
- [ ] `test-scenario` on CI for every component that lays a user policy passes with its `report.md` unchanged

### DIFF
--- a/changes/14547.test.md
+++ b/changes/14547.test.md
@@ -1,0 +1,1 @@
+Let the user policy scenario seed carry a concurrent-login cap.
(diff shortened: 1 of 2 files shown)


jopemachine:
Not needed on its own: the seed change rides inside #14550, the one PR that uses it.

PR #14550, opened 2026-09-11
test(BA-7853): add scenario tests for the user resource policy

jopemachine:
Resolves BA-7853

## Summary
- Replace the TODO stub in `api/adapters/resource_policy/KNOWLEDGE.md` with the intro every policy shares and the user policy's chapter, and transcribe that chapter into `tests/scenario/bai_scenario/manager/resource_policy`: 27 rows over 6 modules, every one of its 6 adapter calls reached. Reading one's own user policy goes straight off the user's row, so it has no not-found row.
- `components/resource_policy.py` holds the frame the three policies share and the user family (an `OwnFamily`, like the keypair one). The project (#14548) and keypair (#14549) PRs carry the same frame so each runs on its own; whichever merges first keeps it. Nothing under `runner/`, `seeder.py` or `components/answers.py` changes.
- `SeedUserPolicy` takes a `max_concurrent_logins`, for the row that clears it. `None` stays the default, so every row laid so far — and every existing `report.md` — reads the same.

Behaviours the rows pin, worth a reviewer's eye:
- Reading one's own user policy is guarded by READ at the caller's own scope, not merely narrowed to it.
- Clearing the nullable concurrent-login cap clears it; clearing the non-nullable folder count leaves it as it was.
- A duplicate name on create and a purge of a policy a user still references are stopped by the database, so the refusal is the repository's (BA-7839, BA-7840). A user the permission graph does not reach is refused as *cannot resolve* whether the name exists or not; the superadmin's unknown name is refused as *not found* (BA-7913, #14660).

## Test plan
- [x] `test-scenario` on CI: 27 scenarios, 0 failing
- [x] `report.md` generated from the CI run's `scenarios.jsonl` with `scripts/scenario-report.py --split`; the project and keypair calls are listed as not exercised until their PRs land


src/ai/backend/manager/api/adapters/resource_policy/KNOWLEDGE.md
```diff
@@ -1,26 +1,111 @@
...
+그것뿐이다. 자기 정책 조회는 사용자 행이 가리키는 정책을 그대로 읽는다.
+
+### 생성
+
+| 시나리오 | 상황 | 요청 | 결과 |
+|---|---|---|---|
+| 슈퍼관리자가 모든 값을 지정해 생성한다 | 정책 없음 | 생성 | 지정한 값이 그대로 담긴 노드 전체 |
+| `max_concurrent_logins` 값을 생략하고 생성한다 | 정책 없음 | `max_concurrent_logins` 없이 생성 | 그 필드가 비어 있는 노드 전체 |
+| 이미 사용 중인 이름으로 생성한다 | 그 이름의 정책이 있음 | 생성 | 이름 중복으로 거부 |
+| 슈퍼관리자가 아닌 사용자가 생성한다 | 권한 없음 | 생성 | 역할 부족으로 거부 |
```
seedspirit, line 206: Wouldn't it be better to describe it as a lack of previledge rather than a lack of roles?

jopemachine: The two phrases are deliberate and paired across these docs: "역할 부족으로 거부" is the superadmin check (`users.role`, raises `InsufficientPrivilege`), "권한 부족으로 거부" is the permission-graph check (`NotEnoughPermission`). Using "권한 부족" for both would hide which gate refused the call, and the tables rely on that pair to show over-granting.

HyeockJinKim (approved):

PR #14551, opened 2026-09-11
fix(agent): stop the periodic reconcilers destroying a kernel that is still being created

daemyung-lablup:
## What this fixes

The agent sometimes destroys a kernel it has just created. The session dies with
`self-terminated` and `exit_code: -1`, which reads as if the kernel had crashed on its own.

## The cause

The agent asks two questions a moment apart, and treats the difference between the answers as a
fact:

1. "docker, which containers are
2.    ... kernel C's container starts here ...
3. "which kernels does my registry hold?"     →  [A, B, C]     C is already RUNNING
4. difference = {C}  →  "C's contt up

C is alive. It was simply born between the two questions.

Two things make step 3 include a kernel that step 1 could not: the registry marks a kernel
`RUNNING` the moment its containeernel` is still running — and the
first answer is a snapshot that stops being true as soon as it is taken.

### DIFF
--- a/src/ai/backend/agent/agent.py
+++ b/src/ai/backend/agent/agent.py
@@ -1594,12 +1594,12 @@ class AbstractAgent[
             try:
                 alive_containers = await self.enumerate_containers()
                 alive_kernel_ids = {kid for kid, _ in alive_containers}
-                registered_kernel_ids = {
-                    kid
+                registered_kernels = {
+                    kid: kernel
                     for kid, kernel in self.kernel_registry.items()
                     if kernel.state == KernelLifecycleStatus.RUNNING and self._is_reconcilable(kid)
                 }
-                dangling_kernel_ids = registered_kernel_ids - alive_kernel_ids
+                dangling_kernel_ids = set(registered_kernels) - alive_kernel_ids
                 if dangling_kernel_ids:
                     # Confirmed against a fresh listing before anything is closed: this cleanup
                     # unregisters the kernel, and the lifecycle sync then destroys the container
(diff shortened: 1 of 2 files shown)


src/ai/backend/agent/docker/agent.py
```diff
@@ -1783,49 +1784,68 @@ async def extract_image_command(self, image: str) -> list[str] | None:
...
+                    kernel_id = await get_kernel_id_from_container(container)
+                    if kernel_id is None:
+                        return
+                    kernel_id_str = str(kernel_id)
+                    if container["State"]["Status"] in status_filter:
+                        owner_id = AgentId(
+                            container["Config"]["Labels"].get(LabelName.OWNER_AGENT, "")
                         )
+                        if self.id == owner_id:
+                            await container.show()
```
fregataa, line 1806: This `show()` is duplicate since `get_kernel_id_from_container()` calls show()

src/ai/backend/agent/docker/agent.py
```diff
@@ -1783,49 +1784,68 @@ async def extract_image_command(self, image: str) -> list[str] | None:
...
+                    raise
 
-                fetch_tasks.append(_fetch_container_info(container))
-
-            await asyncio.gather(*fetch_tasks, return_exceptions=True)
-        return result
+            try:
+                async with asyncio.TaskGroup() as task_group:
+                    for container in containers:
+                        task_group.create_task(_fetch_container_info(container))
```
fregataa, line 1821: There is no timeout and it can cause a single child task can stop all other tasks

PR #14562, opened 2026-09-11
fix(BA-7857): write a registry's allowed projects in the update's transaction

jopemachine:
Found while writing the container registry scenarios (#14519, [review thread](https://github.com/lablup/backend.ai/pull/14519#discussion_r4002727482)).

## Summary
- `admin_update` wrote the allowed-project links through the rbac relation calls **before** it ran the update action, each in a transaction of its own. The superadmin gate lives on the update's global processor, so a non-superadmin holding the two relation permissions changed the links and then got `InsufficientPrivilege`; likewise an update refused by validation (URL without a host, harbor without a project) left the links changed.
- The links now ride with the update: `UpdateContainerRegistryAction` carries what to allow and stop allowing, and the repository writes the row and the links on **one session**, so the gate runs before any write and a refusal rolls back both.

## Why this needed no new plumbing
`ContainerRegistryRepository` already takes a `RelationOpsProvider`, and `V2RelationWriteOps` already extends the general write ops. One `write_ops()` can do `update_data` and `create_relations` / `purge_relations` together, which is what `modify_registry` now does.

## Tests
`tests/unit/manager/repositories/container_registry/test_container_registry_repository.py`
- `test_modify_registry_writes_the_links_with_the_row` — one `modify_registry` call edits the row, links one project and unlinks another.
- `test_modify_registry_refused_leaves_the_row_and_the_links_alone` — parametrized over the two places a refusal can come from: the row (`InvalidContainerRegistryURL`) and the links (`ContainerRegistryGroupsAssociationNotFound`). Either way the row's fields and the linked set are exactly what the fixture seeded.

The adapter-level scenario (non-superadmin refused, no relation change) belongs to #14519's scenario set and is not added here.

## What is not touched
- `apply_allowed_groups` stays as it was, for the create path and its four other callers, where the link change is the whole of what the call does.
- The relation permission gate does not apply on this path any more, and nothing is lost by that: it only refuses a caller who is not a superadmin, and the update itself is behind the superadmin role, so no caller reaches one check without the other.
- Removing is still refused only when none of the named pairs was linked, and adding a pair that is already linked is still not an error. Both answers moved with the code rather than changing.

Resolves BA-7857

----
📚 Documentation preview 📚: https://sorna--14562.org.readthedocs.build/en/14562/

----
📚 Documentation preview 📚: https://sorna-ko--14562.org.readthedocs.build/ko/14562/


jopemachine:
We plan to keep this implementation only for legacy support. For future APIs, rather than exposing APIs that handle cross-cutting concerns, we intend to separate this functionality into its own service.

Therefore, I’m closing this PR as **not planned**.

PR #14579, opened 2026-09-12
fix(accelerator.furiosa): correct the RNGD plugin entry point class name

hoyajigi:
## Problem

The `rngd` entry point in the Furiosa accelerator package declares

```python
"rngd": "ai.backend.accelerator.furiosa.rngd.plugin:RNGDPlugin",
```

but the class is actually named `RngdPlugin` (`src/ai/backend/accelerator/furiosa/rngd/plugin.py:68`). The entry point has never resolved — it was introduced this way in #6916 when the accelerator plugins were open-sourced.

## Impact is wider than the rngd plugin

`BasePluginContext.discover_plugins()` calls `entrypoint.load()` with no exception handling:

```python
# src/ai/backend/common/plugin/__init__.py:134-135
log.info("loading plugin (group:{}): {}", plugin_group, entrypoint.name)
yield entrypoint.name, entrypoint.load()
```

The resulting `AttributeError` propagates out of the generator and aborts `init()` (`:148-153`). So on any host where the `backend.ai-accelerator-furiosa` wheel is installed, the agent fails to load **every** accelerator plugin — CUDA included — not just rngd.

The sibling `warboy` entry point is unaffected because its class name matches.

## Fix

One line in `src/ai/backend/accelerator/furiosa/BUILD`.

Fixing BUILD rather than renaming the class keeps the blast radius minimal: nothing in the tree imports `RngdPlugin`, whereas renaming it would also drag in the adjacent `RngdDevice` for consistency. I'm happy to flip it the other way if maintainers prefer the uppercase-acronym convention used by `CUDAPlugin` / `IPUPlugin` / `TPUPlugin` / `LPUPlugin`.

## Verification

I checked every accelerator entry point by parsing each `BUILD` and confirming the target class is actually defined in the referenced module (AST-level, so it needs no imports or Pants):

| | broken entry points |
|---|---|
| before | **1 of 14** — `furiosa/rngd → RNGDPlugin` (actual: `RngdPlugin`) |
| after | **0 of 14** |

All of `cuda`, `warboy`, `rngd`, `gaudi2`, `gaudi3`, `lpu`, `ipu`, `mock`, `atom`, `atom_plus`, `atom_max`, `rocm`, `tt-n300`, `tpu` now resolve.

## Suggested follow-up

`tests/unit/plugin/test_entrypoint.py` exercises `extract_entrypoints_from_buildscript()` but only checks that entry points can be *extracted*, never that they can be *loaded* — which is why this slipped through. Adding a loadability check (or the AST-level check above, which is cheap and needs no plugin dependencies installed) would prevent a recurrence. Happy to add it here or in a separate PR.


hoyajigi:
Closing in favour of #14581.

I opened this against a base six months behind `main` and did not check whether the bug was still present. The rngd entry point is **already fixed on current `main`** (`RngdPlugin`), so the fix here is a no-op and the branch conflicts.

The part still worth having is the regression guard, which #14581 carries on top of current `main`: a static `ast` check that every `backendai_accelerator_v21` entry point resolves to a class that exists. All 14 pass today.

PR #14589, opened 2026-09-14
refactor(BA-7798): settle the three shapes a scope reads its condition from

HyeockJinKim:
## Summary

- **A scope now reads its condition the way a permission check does.** `scope_membership_exists` walked one edge — the scope's own virtual entity holds the row — while authorization walks two: the row is held by a virtual entity, and that virtual entity is governed by the scope. A domain therefore saw only what it held directly, never what its projects hold. Every node carries a self-govern edge, so a project and a user reach exactly what they did before.
- **Naming a user as a scope now stands for the project that is theirs alone.** What a user holds is created in their personal project (BEP-1077), so a scope walking from the user alone found none of it. The vfolder scope carried that hop inline; the session, deployment and model card scopes did not. `user_scope_reaches` puts the rule in one place.
- **The three shapes a scope reads from are written down**, in `models/KNOWLEDGE.md`: an entity row answers through the membership graph, a field row through the column naming its owner, a dangling field row through the type and id written on it. `scopes.py` keeps the contract and points at the document.
- **Entities gained the axes they were missing**, each reached through one scoped-search entry (`POST /scoped/search`, `scoped*` in GraphQL) that takes a list per scope kind and ORs them: a domain for deployments, vfolders and sessions; a domain and a user for model cards; a user for sessions. Existing per-axis routes keep their shapes and build a single scope item for the same action.
- **The vfolder and session scoped searches moved onto ops**, replacing one action per scope kind — each carrying its scope as a field, reaching rows through a service and repository method of its own — with one action carrying scope items.
- **What is not a scope stopped being one:** a role is a filter on the global user search, an offer's sender is a filter within the ownership scope, and an offer is reached through the entity it is attached to rather than the id pair on its row. An entity's audit records and the records of runs that named it as their scope are now asked for separately.

## Left open

| Item | Why |
|---|---|
| REST's project-scoped session search authorizes as a general session search plus a `group_id` filter, where the GraphQL one authorizes against the project | Closing it narrows authorization; it is a decision, not a refactor |
| `endpoints.created_user` still stands as the user term | It records who created a deployment, not who owns it, but dropping it now hides every pre-backfill row from the user axis (BA-7571) |
| `EntityShareRecipientProjectScope` still reads the table | The recipient side stays table-backed because an address invitation has no node; the project recipient could move, but the two belong together |
| 212 actions are wired global (`backend.ai mgr ops list`) | A separate question from what a scope reads — worth its own pass |

## Not in scope

- **Fair share and usage bucket** — a superadmin feature, and their reads enumerate every domain by design rather than bounding by ownership.
- **Missing entity scopes** (project resource policy, session template, container registry, image) **and relation reads** — each needs an API entry to be reachable, and this change does not add API surface beyond the scoped-search entries above.

## Test plan

- [ ] `pants test ::` green in CI
- [ ] A domain-scoped read returns the rows its projects hold
- [ ] A project-scoped and a user-scoped read return what they returned before
- [ ] `my` and project searches for vfolders, sessions and deployments are unchanged
- [ ] An entity's audit records still come back for a request naming it once

Resolves BA-7798

----
📚 Documentation preview 📚: https://sorna--14589.org.readthedocs.build/en/14589/

----
📚 Documentation preview 📚: https://sorna-ko--14589.org.readthedocs.build/ko/14589/


fregataa (commented):
please check the CI failure

PR #14606, opened 2026-09-14
fix(BA-7878): build the cursor keyset condition in PaginationSpec

jopemachine:
## 📚 Stacked PRs

1. 👉 **#14606** — `fix(BA-7878): build the cursor keyset condition in PaginationSpec` ← you are here
2. ⬇️ #14671 — `fix(BA-7878): move every entity onto the PaginationSpec cursor condition`

## Summary
Two defects in cursor pagination, fixed in the shared pagination layer:

1. **Tied sort values were skipped.** Every entity's `by_cursor_forward` / `by_cursor_backward` compared the sort column alone (`created_at < cursor row's created_at`), so rows sharing the cursor row's value — every row inserted in one transaction, since `created_at` defaults to `now()` — fell outside the page.
2. **The tiebreaker was not reversed for `last`/`before`.** `_build_querier` and `BaseGQLAdapter.build_querier` appended `tiebreaker_order` as-is, so the backward page order was `(created_at ASC, id ASC)` rather than the mirror of `(created_at DESC, id ASC)`. Within a tie group larger than `last`, backward pages picked the wrong rows even with (1) fixed.

`PaginationSpec` now derives the backward orders by reversing `forward_order` / `tiebreaker_order`, and `build_cursor_condition(cursor_id, backward=)` builds the keyset predicate `sort <op> sort_at_cursor OR (sort == sort_at_cursor AND tiebreaker <op> cursor)` from those two orders. The tiebreaker column is the row identity a cursor encodes, so it must be unique; a GUID tiebreaker rejects a non-UUID cursor with `InvalidCursor`. Both querier builders append the reversed tiebreaker for `last`/`before`.

A spec that still passes `backward_order` and the condition factories keeps using them, so every other entity is unchanged here; app config definitions are the first entity moved onto the shared condition. #14671 moves the rest and deletes the legacy fields.

Test: `tests/unit/manager/api/adapter_options/pagination/test_pagination_spec.py` pages a tie group forward and backward through `build_querier` + `execute_batch_querier`; four of its six cases fail on the previous behaviour.

## Reproduction

Against a manager without this fix, insert several rows in one statement so they share `created_at`, then page with `first: 1`:

```sql
INSERT INTO app_config_definitions (config_name) VALUES ('tie-a'), ('tie-b'), ('tie-c');
-- all three: created_at = 2026-09-15 08:45:54.486859+00
```

```graphql
{ adminAppConfigDefinitions(first: 1) { count pageInfo { hasNextPage } edges { cursor node { configName } } } }
# count: 3, hasNextPage: true, edges: [tie-a]

{ adminAppConfigDefinitions(first: 1, after: "<cursor of tie-a>") { count pageInfo { hasNextPage } edges { node { configName } } } }
# count: 3, hasNextPage: false, edges: []   ← tie-b and tie-c are never reachable
```

Carved out of #14511, where the fix first landed inside the scenario test PR.

Resolves BA-7878.


HyeockJinKim (changes requested):
return PaginationSpec(
 forward_order=DEFAULT_FORWARD_ORDER,
 backward_order=DEFAULT_BACKWARD_ORDER,
        forward_condition_factory=ContainerRegistryConditions.by_cursor_forward,
 backward_condition_factory=ContainerRegistryConditions.by_cursor_backward,
 tiebreaker_order=TIEBREAKER_ORDER,
    )
 ----
I assume I set `tiebreaker_order` to ensure this, but it’s behaving strangely—all queries are now triggering the tiebreaker.

jopemachine:
Superseded by #14734 (BA-7979), which builds the cursor keyset condition in PaginationSpec for every entity. Closing BA-7878 as done.

PR #14610, opened 2026-09-14
chore(BA-7879): tear down the RBAC nodes a definition purge cascades to

jopemachine:
## Summary
- The database cascades an app config definition delete to its allow-list entries and their fragments, but the generic purge (`V2WriteOps.purge_entity` → `_teardown`) removes only the definition's own virtual entity, so the cascaded rows left their RBAC nodes behind.
- `definition_purge` now runs `AppConfigService.purge_definition`: a new `AppConfigDefinitionRepository.purge` deletes the fragments (`DefinitionFragmentPurger`), then the entries (`DefinitionAllowListPurger`), then the definition in one `write_ops()` transaction, each with the graph it left — the way `ProjectRepository.purge_group` clears a project's endpoints and sessions. `PurgeAppConfigDefinitionAction` becomes a plain single-entity action with its own result.
- A repository test seeds the three rows and their virtual entities directly and checks that none survives the purge.

The allow-list purge has the same gap for the fragments it cascades to; #14614 (BA-7882) closes it the same way. Both add a repository and a constructor argument to `AppConfigService`; whichever merges second rebases over the other and folds the two `KNOWLEDGE.md` sections together. `chore` rather than `fix`: the v2 definition purge is not on 26.8 or 26.4.

Resolves BA-7879.


jopemachine:
Superseded by #14652, which landed the same teardown: `AppConfigRepository.purge_definition` batch-purges the fragments and allow-list entries with their nodes before the definition, and `TestPurgeDefinition.test_takes_the_allow_list_and_fragments_with_their_nodes` covers it. Not planned.

PR #14614, opened 2026-09-14
chore(BA-7882): tear down the RBAC nodes an allow-list purge cascades to

jopemachine:
## Summary
- The database cascades an app config allow-list entry delete to the fragments under its `(config_name, scope_type)`, but the generic purge (`V2WriteOps.purge_entity` → `_teardown`) removes only the entry's own virtual entity, so the cascaded fragments left their RBAC nodes behind.
- `allow_list_purge` now runs `AppConfigService.purge_allow_list_entry`: a new `AppConfigAllowListRepository.purge` deletes the fragments (`AllowListFragmentPurger`), then the entry, in one `write_ops()` transaction, each with the graph it left. `PurgeAppConfigAllowListAction` becomes a plain single-entity action carrying the entry id, with its own result.
- A repository test seeds the entry, the fragment and their virtual entities directly and checks that none survives the purge.

Sibling of #14610 (BA-7879, the definition purge). Both add a repository and a constructor argument to `AppConfigService`; whichever merges second rebases over the other. `chore` rather than `fix`: the v2 allow-list purge is not on 26.8 or 26.4.

Resolves BA-7882.


jopemachine:
Superseded by #14652, which landed the same teardown: `AppConfigRepository.purge_allow_list` batch-purges the fragments with their nodes before the entry, and `TestPurgeAllowList.test_takes_the_fragments_and_keeps_the_definition` covers it. Not planned.

PR #14671, opened 2026-09-16
fix(BA-7878): move every entity onto the PaginationSpec cursor condition

jopemachine:
## 📚 Stacked PRs

1. ⬇️ #14606 — `fix(BA-7878): build the cursor keyset condition in PaginationSpec`
2. 👉 **#14671** — `fix(BA-7878): move every entity onto the PaginationSpec cursor condition` ← you are here

## Summary
- Every remaining entity moves onto the keyset condition `PaginationSpec` gained in #14606: `backward_order` and the two condition factories are dropped from the 61 other `PaginationSpec` constructions, and the legacy fields and factory branch are removed from `PaginationSpec` itself.
- The per-entity `by_cursor_forward` / `by_cursor_backward` conditions (132 methods across 44 modules), `CursorConditionFactory`, and the `*_DEFAULT_BACKWARD_ORDER` constants they served are deleted. `api/rest/deployment/adapter.py` builds its route cursor condition from `get_route_pagination_spec()`.
- `ResourceGroup`, `RuntimeVariant` and `ResourcePreset` specs tiebreak on `id` instead of `name` — the cursor already encoded the id, and `ResourcePresetRow.name` is not unique. The deployment policy spec (previously reusing `EndpointRow` conditions) and the role assignment spec (previously comparing `id` under a `granted_at` order) come out correct by construction.

Not covered: the two allocated-resource-slot specs sort on the joined `ResourceSlotTypeRow.rank` and tiebreak on `slot_name`; their search payloads carry no cursors, so their cursor branch is unreachable today, and a join-aware condition would be needed before one is issued.

Resolves BA-7878.


jopemachine:
Superseded by #14734 (BA-7979), which moved every entity onto the PaginationSpec cursor condition. Closing BA-7878 as done.

PR #14672, opened 2026-09-16
fix(BA-7923): let a superadmin's partial bulk read report a missing id as a miss

jopemachine:
## What

A superadmin's `batch_load_by_ids` (and every other `partial_bulk_get_ops` read) answered a missing id with `NotEnoughPermission` instead of `None`.

`VirtualEntityPartialBulkActionRBACValidator` asked `BulkOwnCheck` for a superadmin too. `owned_permissions` deliberately holds every bit on an entity that has a graph node and none on one that has not, so an id matching no row came back `Permission.NONE` and was denied before `PartialBulkGetService` could report it as `EntityNotFoundError`. `BaseAdapter.batch_load_failure` only maps `NotFoundError` to `None`, so the DataLoader was handed a 403 for a plain miss — contradicting its own docstring ("An id matching no row stays `None`").

## Fix

The validator now passes a superadmin without asking the own check, as the single-entity, scope and relation RBAC validators already do. The read then runs against every id the superadmin named and the miss comes back as the miss it is.

- `actions/v2/bulk/validator/rbac.py` — superadmin short-circuit in `VirtualEntityPartialBulkActionRBACValidator.validate`.
- `tests/unit/manager/actions/validators/test_virtual_entity_rbac_validators.py` — the partial validator had no tests; adds the superadmin-without-a-node, missing-user and denied-item cases.
- `tests/scenario/bai_scenario/manager/user/test_reading.py` — `the-superadmin-batch-load-leaves-a-missing-id-empty` now asserts the `None` it already described instead of skipping the slot; `user/report.md` regenerated.

The atomic bulk validator is left as is: `test_superadmin_is_refused_an_entity_without_a_node` pins that a superadmin's bulk *write* naming an entity without a node is refused as a whole, and this change does not touch that path.

## Issue

[BA-7923](https://lablup.atlassian.net/browse/BA-7923)

## Found by

Writing the image adapter scenarios (#14520): `loading-many-image-ids-keeps-the-order-and-leaves-a-hole` got `[ImageNode, NotEnoughPermission, ImageNode]` for a superadmin. The user scenario on `main` had already sidestepped the same slot with a `Skipped`.

## Checklist

- [x] `pants fmt / fix / lint / check --changed-since=origin/main`
- [x] `pants test --changed-since=origin/main --changed-dependents=direct`, plus `tests/component/idle_checker_assignment::` and `tests/component/session/permission::`
- [x] `scripts/scenario-report.py --verify` level with the run

[BA-7923]: https://lablup.atlassian.net/browse/BA-7923?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jopemachine:
Superseded by #14675 — the head branch was renamed to carry the issue key (BA-7923) and GitHub closed this one with it.

PR #14544, opened 2026-09-11
test(BA-7848): add scenario tests for the client IP masking adapter

jopemachine:
Resolves BA-7848

## Summary
- Fill `api/adapters/client_ip_masking/KNOWLEDGE.md` and transcribe it into `tests/scenario/bai_scenario/manager/client_ip_masking/` — 23 rows over three modules (upserting, searching, retiring) — with `seeds/client_ip_masking/policy.py` and `components/client_ip_masking.py`.
- A masking policy has no creator, only an upserter, so `Seeder` learns to write a `GlobalEntityUpserter` through `upsert_global_entity`.

What the rows pin:
- Upsert replaces the row whole: putting a policy on a target that has one rewrites the same id, and omitting the prefixes clears the stored ones.
- Each of the three targets and three modes gets a row.
- Upsert and search are behind the superadmin role (the monitor passes the search); purge is behind the entity gate, which no role can reach since the policy is created in no scope.

`tests/scenario/bai_scenario/components/system.py` — the caller-only situation every system-concern table shares — travels in each of the eight sibling PRs byte for byte (#14535, #14536, #14538, #14542, #14543, #14544, #14545). Whichever merges first, the rest rebase clean.

## Test plan
- [x] `python3 scripts/knowledge/check.py` passes
- [x] The rows passed in CI run 34558080395 of the unsplit branch (0 failing); `report.md` is generated from that run's artifacts with `scripts/scenario-report.py --split`
- [ ] `test-scenario` and `verify-scenario-report` pass on this PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Snp9H2jEVbB2s1C6ixUHGn

PR #14543, opened 2026-09-11
test(BA-7847): add scenario tests for the login client type adapter

jopemachine:
Resolves BA-7847

## Summary
- Fill `api/adapters/login_client_type/KNOWLEDGE.md` and transcribe it into `tests/scenario/bai_scenario/manager/login_client_type/` — 22 rows over five modules — with `seeds/login_client_type/login_client_type.py` and `components/login_client_type.py`.
- All five calls are exercised.

What the rows pin:
- The reads ask for authentication alone; the default page size is fifty, not the ten of the other catalogs.
- Create is behind the superadmin role; update and delete are behind the entity gate, which no role can reach since the type is created in no scope.
- Renaming to a taken name is refused with the raw unique-violation error, not the conflict error the create path maps it to.

`tests/scenario/bai_scenario/components/system.py` — the caller-only situation every system-concern table shares — travels in each of the eight sibling PRs byte for byte (#14535, #14536, #14538, #14542, #14543, #14544, #14545). Whichever merges first, the rest rebase clean.

## Test plan
- [x] `python3 scripts/knowledge/check.py` passes
- [x] The rows passed in CI run 34558080395 of the unsplit branch (0 failing); `report.md` is generated from that run's artifacts with `scripts/scenario-report.py --split`
- [ ] `test-scenario` and `verify-scenario-report` pass on this PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Snp9H2jEVbB2s1C6ixUHGn

PR #14701, opened 2026-09-17
test(BA-7940): add scenario tests for the resource group adapter

jopemachine:
Resolves BA-7940

## Summary
- Fill `api/adapters/resource_group/KNOWLEDGE.md` and transcribe it into `tests/scenario/bai_scenario/manager/resource_group/` — 84 rows over eight modules (creating, reading, searching, editing, retiring, allowing, fair share, options), replacing the three PoC rows in `test_resource_group.py` — with `components/resource_group.py`, a `FakeScheduleCoordinator` in `fakes/scheduler.py` beside the existing `FakeDeploymentCoordinator`, and `description`/`is_active`/`is_default` on `SeedResourceGroup`.
- All 22 calls are exercised; `report.md` no longer lists any of them as unexercised.

What the rows pin:
- A resource group is created in no scope and is its own scope, so a custom role laid on the group (`SeedRole(scope=ResourceGroupID)`) with `READ`/`UPDATE`/`HARD_DELETE` on `resource_group` is what lets a user other than the superadmin read, edit, purge, read the resource info and the allowed domains/projects, change the fair share spec and replace the default deployment options. Every entity-gated call has that granted/ungranted pair.
- The name lookup is public: a name nothing answers to is not found for the superadmin and for a user granted nothing alike; `get` refuses with `ResourceGroupNotFound`, the lookup-based calls with `EntityNotFoundError`.
- `get_fair_share_spec` finds the group through the global search, so a user granted `READ` on the group is refused for the superadmin role there while `update_fair_share_spec` lets a user granted `UPDATE` through. With no agents the capacity is empty, so any `resource_weights` entry is refused.
- Scoped search and `get_allowed_resource_groups_for_domain` pass a user holding `resource_group:READ` in the domain scope and answer only the groups linked to that domain, the superadmin included.
- The create path reports a second default group as a name conflict (`ResourceGroupConflict`); the update path reports it as `DefaultResourceGroupAlreadyExists`.
- `update_allowed_domains_for_resource_group` with an unknown name in `remove` is refused with `EntityNotFoundError` rather than skipped, contrary to the adapter's docstring ("unknown names to remove are skipped"): the lookup raises `EntityNotFoundError`, and the adapter only catches `DomainNotFound`. The row pins the current behavior; the fix is a one-line change in `_resolve_allowed_domain_ids` and is left out of this test PR.
- Both default-options replacements refuse a handler name the coordinator has not registered.

`tests/scenario/bai_scenario/components/system.py` — the caller-only situation every global-catalog table shares — travels here byte for byte as in #14535, #14536, #14538, #14542, #14543, #14544, #14545, #14700. Whichever merges first, the rest rebase clean.

## Test plan
- [x] All 84 rows pass locally (`pants test --force tests/scenario/bai_scenario/manager/resource_group::`, 0 failing); `report.md` is generated from that run with `scripts/scenario-report.py --split` and `--verify` is level with it


src/ai/backend/manager/api/adapters/resource_group/KNOWLEDGE.md
```diff
@@ -1,43 +1,214 @@
...
+것. 허용 목록 호출들은 그중 도메인과 프로젝트에 거는 관계를 쓰고 읽는다.
+
+## 생성
+
+| 시나리오 | 상황 | 요청 | 결과 |
+|---|---|---|---|
+| 슈퍼관리자가 이름만 주고 만든다 | 슈퍼관리자 | 생성 | `is_active`와 `is_public` 값은 true, `is_default` 값은 false, `description` 값은 비어 있고 `fifo` 스케줄러를 쓰는 노드 전체 |
+| `is_default` 값을 true로 지정해 만든다 | 기본 그룹 없음, 슈퍼관리자 | `is_default` 값을 true로 지정해 생성 | `is_default` 값이 true인 노드 |
+| 이미 있는 이름으로 만든다 | 그 이름의 그룹 있음, 슈퍼관리자 | 생성 | 이름 중복으로 거부 |
+| 기본 그룹이 있는데 또 기본으로 만든다 | `is_default` 값이 true인 그룹 있음, 슈퍼관리자 | `is_default` 값을 true로 지정해 생성 | 이름 중복으로 거부 |
```
seedspirit: In this scenario, is it normal for the request to be rejected due to a duplicate name?

jopemachine: No, it is not — the row pinned a defect. `ResourceGroupCreator.integrity_error_checks` matches every `UniqueConstraintViolationError` with no `constraint_name`, so the `uq_scaling_groups_is_default` partial index violation is reported as a name conflict; `ResourceGroupUpdater` already tells the two apart and raises `DefaultResourceGroupAlreadyExists`.

Filed as BA-7946. In 1938ce7029 the scenario expects `DefaultResourceGroupAlreadyExists` (fails on current code) and is commented out of `SCENARIOS` behind `TODO(BA-7946)`; the table row and the paragraph under it say so.

jopemachine: It is what the code does today: the create path maps every unique-constraint violation to the same name-duplicate error, so the partial unique index on `is_default` surfaces as "이름 중복". The update path does distinguish it (`DefaultResourceGroupAlreadyExists`), which is why the update table says "기본 그룹이 이미 있다는 이유로 거부". The prose under the table records the difference; making create raise the dedicated error is a product change outside this test PR.

HyeockJinKim (approved):

PR #14713, opened 2026-09-17
feat(BA-7952): let the monitor role read every provisioned entity without a grant

jopemachine:
## Summary

- The permission read ops (`repositories/ops/v2/permission/read.py`) answered a superadmin's own and govern checks from the stored role but a monitor's only through the graph, so every RBAC-gated read refused a monitor holding no grant. The scoped audit log search is where it surfaced: [PR #14506 review](https://github.com/lablup/backend.ai/pull/14506/changes#diff-45afd36b7f004c2284a1f7c724823bc92c07e9bd3af5fbb58cc24a5adf5c01cdR81) — "In `monitor`, all read operations should be possible even without access permissions."
- `owned_permissions` / `governed_permissions` now give a monitor the read bit on every entity and scope that has a node, next to the superadmin's every bit. Nothing changes for an entity without a node, for writes, or for any other role.
- Tests: `held_permissions` asserts the exact bits a monitor holds on an entity with a node (`READ`) and without one (`NONE`), and the scope validator passes a monitor's read within a provisioned scope. The two cover both changed functions; the bulk validators only turn those bits into pass/refuse and are unchanged.
- Scenarios: the three rows that asserted the old refusal flip to the answer a superadmin gets — audit log by-id read and scoped search, notification channel read by id — in `KNOWLEDGE.md`, the test, and `report.md`. The user adapter gains a monitor row on the by-id read and on the domain search.

## Test plan

- [x] `pants test tests/unit/manager/actions/validators:: tests/unit/manager/repositories/permission_controller:: tests/unit/manager/actions/test_relation_processor.py tests/unit/manager/api/adapters/test_rbac_adapter.py`
- [x] `pants test tests/scenario/bai_scenario/setup:: tests/scenario/bai_scenario/manager/audit_log:: tests/scenario/bai_scenario/manager/notification:: tests/scenario/bai_scenario/manager/user::` with `BACKEND_SCENARIO_LOG`, then `scripts/scenario-report.py --split` / `--verify` (3 reports level with the run)
- [x] `pants check` / `lint` / `fmt` on the changed files
- [ ] Live: a monitor-role user runs the scoped audit log search (`scopedAuditLogs`) on an entity it holds no grant on and gets the same page a superadmin gets

Resolves BA-7952


jopemachine:
Will be addressed in a separate task, So I will close this PR as not planned

PR #14761, opened 2026-09-18
refactor(BA-7989): consolidate manager/service/ into services/

jopemachine:
## Summary

Resolves BA-7989. Supersedes #12926 (closed unmerged; its branch fell 144 commits behind `main`).

- Delete the singular `manager/service/` package. Its per-project container registry quota logic is folded into `ContainerRegistryService`, which now takes `PerProjectRegistryQuotaRepository` and `PerProjectContainerRegistryQuotaClientPool` and performs the four quota operations itself (`_registry_quota_target` resolves the registry row → client → Harbor project/auth once per call).
- Drop `ServicesContext` and the `AbstractPerProjectContainerRegistryQuotaService` / `PerProjectContainerRegistryQuotaService` pair. Nothing is threaded through the DI stages any more: `DomainResources`, `ProcessingInput`, `ServiceArgs`, `GQLContextDeps` and `GraphQueryContext` lose their quota-service field, and `dependencies/domain/services.py` (`ServicesContextDependency`) is removed. `services/factory.py` builds the service from `repositories.container_registry_quota.repository` (the existing `PerProjectRegistryQuotaRepositories` bundle, now wired into `Repositories`) and a fresh client pool.
- `PerProjectContainerRegistryQuotaClientPool` moves next to the Harbor client it dispatches to (`clients/container_registry/harbor.py`).
- **Behavior change (gql_legacy):** `create/update/delete_container_registry_quota` and `GroupNode.registry_quota` now run through `processors.container_registry.*_registry_quota` instead of calling the service directly, so the global-scope `SuperAdminActionValidator` applies — the same gate REST `/group/registry-quota` already has. Domain admins lose the three mutations (`allowed_roles` still lists ADMIN but the processor refuses); `registry_quota` reads are superadmin/monitor only.
- Also removes an unused `GlobalEntityUpserter` import in `actions/v2/ops/base.py` — `pants lint` fails on the `manager:src` target without it.

| Before | After |
|---|---|
| `manager/service/container_registry/harbor.py` (abstract service, client pool, service) | client pool → `clients/container_registry/harbor.py`; service logic → `ContainerRegistryService` |
| `manager/service/base.py` (`ServicesContext`) | removed |
| `dependencies/domain/services.py` (`ServicesContextDependency`, `ServicesInput`) | removed |
| `DomainResources.services_ctx` / `GQLContextDeps.services_ctx` / `GraphQueryContext.services_ctx` | removed |
| `ProcessingInput.registry_quota_service` / `ServiceArgs.registry_quota_service` (`… \| None`) | removed |
| `ContainerRegistryService(db, repo, quota_service=None)` + `_ensure_quota_service()` | `ContainerRegistryService(db, repo, quota_repository, quota_client_pool)` |
| `graph_ctx.services_ctx.per_project_container_registries_quota.X(...)` | `graph_ctx.processors.container_registry.X_registry_quota.run(...)` |

Tests: `tests/component/group/conftest.py` seeds a HARBOR2 `ContainerRegistryRow` bound to `target_group` and fakes only the Harbor client through the pool; the other `ContainerRegistryService` construction sites pass the real repository and pool.

The `preset_factory` (`tests/component/resource_preset/conftest.py`) and `resource_preset_fixture` (`tests/component/infra/conftest.py`) fixtures now call `valkey_clients.stat.invalidate_all_resource_presets()` after their direct DB insert and on teardown. `ResourcePresetRepository.list_presets` reads a 60s valkey cache first, so when the two files land in one pytest batch (this PR's changed set puts them there) the second file's `list_presets` returned the first file's stale list — `test_admin_lists_presets_with_fixture` failed in the pre-push hook. Pre-existing isolation gap, surfaced by the batch composition.

## Test plan

- [x] `pants lint` / `pants check` on `--changed-since=origin/main` pass
- [x] `pants test tests/unit/manager/dependencies:: tests/unit/manager/services/container_registry:: tests/unit/manager/repositories/container_registry_quota:: tests/unit/manager/actions/test_registry_catalog.py`
- [x] `pants test tests/unit/manager/api/group:: tests/unit/manager/api/test_gql_legacy_*.py`
- [x] `pants test tests/component/group:: tests/component/container_registry:: tests/component/infra:: tests/component/resource_preset:: tests/component/etcd_config:: tests/scenario/bai_scenario/manager/container_registry::`
- [x] `ai.backend.manager.api.rest.setup`, `dependencies.composer`, `services.factory` import cleanly; `ai.backend.manager.service` no longer resolves
- [ ] CI


tests/component/container_registry/conftest.py
```diff
@@ -50,7 +56,12 @@ def container_registry_processors(
     processor_registry: ProcessorRegistry[Any],
 ) -> ContainerRegistryProcessors:
     repo = ContainerRegistryRepository(database_engine, RelationOpsProvider(database_engine))
-    service = ContainerRegistryService(database_engine, repo)
+    service = ContainerRegistryService(
+        database_engine,
+        repo,
+        PerProjectRegistryQuotaRepository(database_engine),
+        PerProjectContainerRegistryQuotaClientPool(),
```
jopemachine: Inject this as separate fixtures

jopemachine: Split into `registry_quota_repository` / `registry_quota_client_pool` fixtures in e8e721d3a7 (applied to the same fixture in infra, etcd_config, resource_preset and group as well; group overrides the pool fixture with `InMemoryQuotaClientPool`).

tests/component/group/conftest.py
```diff
@@ -17,64 +18,107 @@
...
-    the abstract read_quota() returns int, but the API response model
+    Does not inherit AbstractPerProjectRegistryQuotaClient because the abstract
+    read_quota() returns int, but the API response model
     (ReadRegistryQuotaResponse.result) is int | None. This fixture mirrors
     the expected API behaviour: None when no quota is configured.
     """
 
     def __init__(self) -> None:
-        self._store: dict[uuid.UUID, int] = {}
+        self._store: dict[str, int] = {}
```
jopemachine, line 64: Key type changed, why?

src/ai/backend/manager/repositories/container_registry/repository.py
```diff
@@ -246,6 +251,57 @@ async def get_registry_by_url_and_project(
             result = await session.execute(stmt)
             return result.scalars().one_or_none()
 
+    @container_registry_repository_resilience.apply()
+    async def get_by_project_scope(self, scope_id: ProjectScope) -> PerProjectContainerRegistryInfo:
```
jopemachine: Wrong method name

src/ai/backend/manager/repositories/container_registry/repository.py
```diff
@@ -246,6 +251,57 @@ async def get_registry_by_url_and_project(
...
+                )
+
+            return PerProjectContainerRegistryInfo(
+                id=registry_row.id,
+                url=registry_row.url,
+                registry_name=registry_row.registry_name,
+                type=registry_row.type,
+                project=registry_row.project or "",
+                username=registry_row.username or "",
+                password=registry_row.password or "",
```
jopemachine: Why is this needed?

src/ai/backend/manager/repositories/container_registry/repository.py
jopemachine: Renamed to `get_project_registry` in 23eaa45063 — it resolves the registry the project's `container_registry` config points at, not a registry filtered by project.

src/ai/backend/manager/repositories/container_registry/repository.py
jopemachine: The placeholders came verbatim from the deleted `container_registry_quota` db_source. Dropped in eea97070a4: `get_project_registry` now returns `ContainerRegistryData` as-is and the service raises `ContainerRegistryQuotaNotConfigurable` when the registry has no project, username or password.

src/ai/backend/manager/repositories/container_registry/repository.py
```diff
@@ -249,6 +254,46 @@ async def get_registry_by_url_and_project(
             result = await session.execute(stmt)
             return result.scalars().one_or_none()
 
+    @container_registry_repository_resilience.apply()
+    async def get_project_registry(self, scope_id: ProjectScope) -> ContainerRegistryData:
```
seedspirit, line 258: Would be better to introduce new container registry dbsource instead of accessing db in repository layer.

seedspirit (commented):
There are wide changes to the PR, but it looks like they can be broken down into about two parts. Could you please take a look at it?

src/ai/backend/manager/repositories/container_registry/repository.py
jopemachine: Since `db_source` was not distinguished separately, it makes sense to handle that as a separate task.

jopemachine:
> There are wide changes to the PR, but it looks like they can be broken down into about two parts. Could you please take a look at it?

I've already been asked not to split PRs too finely, so how would you like this PR to be split? @seedspirit 
If there are specific changes you'd like to see included in this PR, please list those as well.

Aside from the test code, most of the changes in this PR are either code moves or renames, so I don't think there are any particularly difficult parts.

HyeockJinKim (approved):

PR #5644, opened 2025-08-26
feat: Add `reservoirs` DB table

jopemachine:
resolves #NNN (BA-MMM)

**Checklist:** (if applicable)

- [ ] Milestone metadata specifying the target backport version
- [ ] Mention to the original issue
- [ ] Installer updates including:
  - Fixtures for db schema changes
  - New mandatory config options
- [ ] Update of end-to-end CLI integration tests in `ai.backend.test`
- [ ] API server-client counterparts (e.g., manager API -> client SDK)
- [ ] Test case(s) to:
  - Demonstrate the difference of before/after
  - Demonstrate the flow of abstract/conceptual models with a concrete implementation
- [ ] Documentation
  - Contents in the `docs` directory
  - docstrings in public interfaces and type annotations

PR #14843, opened 2026-09-21
refactor(BA-8001): remove the object_storage_row relationship from StorageNamespaceRow

wngh4628:
## Summary
- `models/storage_namespace/row.py`: remove the `object_storage_row` relationship and its join condition `_get_storage_namespace_join_cond`, along with the `relationship`/`foreign`/`TYPE_CHECKING` imports they needed. The row now declares its three columns only (its `to_dataclass()` already moved to `StorageNamespaceSearchableFields` in BA-8062).
- `ObjectStorageDBSource.get_by_namespace_id`, the one place that loaded the relationship, joins `object_storages` on `storage_namespace.storage_id` directly instead of `selectinload`, and converts through `ObjectStorageSearchableFields.own.to_data(row)` as the other reads do. Its two not-found branches collapse into one `ObjectStorageNotFoundError`, since a single join answers both.

## Open question
`ObjectStorageRepository.get_by_namespace_id` (repository.py:49) has no caller anywhere in `src` or `tests`. Its four callers were removed in #5644 (BA-2192), and #6490 (BA-2901) moved the uncalled wrapper and db_source method as they were. If nothing plans to use it, both `get_by_namespace_id` methods should go instead of being rewritten, which would make this PR a pure deletion. Please confirm whether it is planned work or dead code.

Resolves BA-8001


fregataa (commented):
Good. Please check merge conflict

fregataa (commented):
This looks good to me. Do you need to do more task here?

wngh4628:
No, nothing more to do here. The change is complete and the PR is ready to merge.

fregataa (approved):

PR #14850, opened 2026-09-21
refactor(BA-8002): remove the meta relationships between artifact registry rows

wngh4628:
## Summary
- `models/huggingface_registry/row.py`, `models/reservoir_registry/row.py`: remove the `meta` relationship and its join condition. `models/artifact_registries/row.py`: remove their `back_populates` pair (`huggingface_registries`, `reservoir_registries`) and the two join conditions. The registry rows keep their columns and inherit `RegistryNameMixin` (`models/mixins/registry_name.py`), a `query_expression` slot the reader fills with the name from `artifact_registries`.
- The name is read by join instead: the two db_sources and the searchers join `artifact_registries` and load its name onto `registry_name` with `with_expression`, and the registry write ops sets `registry_name` after create and update instead of `refresh(row, ["meta"])`. `{HuggingFace,Reservoir}RegistrySearchableFields.own` declares `name` as that derived field and `to_data` reads it instead of `row.meta`, so the conversion path from BA-8062 stays the only one.
- Four commits, each green on its own: add the slot, switch the huggingface readers, switch the reservoir readers, delete the relationships.

Resolves BA-8002


src/ai/backend/manager/models/huggingface_registry/searchable_fields.py
```diff
@@ -36,15 +40,17 @@ class _HuggingFaceRegistryOwnFields(
...
+    name = SearchableField(HuggingFaceRegistryRow.registry_name, None, None)
+    """Derived: the artifact_registries row's name, filled by the reader's join."""
 
     @override
     def to_data(self, row: HuggingFaceRegistryRow) -> HuggingFaceRegistryData:
-        meta = row.meta
-        if meta is None:
+        name = self.name.read(row)
+        if name is None:
             raise RelationNotLoadedError()
```
fregataa: Let's change this exception since it is not about relation loading anymore

src/ai/backend/manager/models/huggingface_registry/searchers.py
```diff
@@ -18,11 +19,20 @@
...
     def build_select(self) -> sa.sql.Select[Any]:
-        return sa.select(HuggingFaceRegistryRow).options(selectinload(HuggingFaceRegistryRow.meta))
+        return (
+            sa.select(HuggingFaceRegistryRow)
+            .join(
+                ArtifactRegistryRow,
+                ArtifactRegistryRow.registry_id == HuggingFaceRegistryRow.id,
+            )
+            .options(
+                with_expression(HuggingFaceRegistryRow.registry_name, ArtifactRegistryRow.name)
```
fregataa, line 33: Could you explain why this `with_expression()` is needed?

src/ai/backend/manager/models/mixins/registry_name.py
```diff
@@ -0,0 +1,16 @@
...
+
+from __future__ import annotations
+
+from sqlalchemy.orm import Mapped, declared_attr, query_expression
+
+
+class RegistryNameMixin:
+    @declared_attr
+    def registry_name(cls) -> Mapped[str | None]:
+        return query_expression()
```
seedspirit, line 16: Can you explain why Mixin is implemented instead of using individual field?

src/ai/backend/manager/models/huggingface_registry/searchers.py
wngh4628: with_expression is what puts artifact_registries.name into the select, so the query becomes:

  SELECT artifact_registries.name,
         huggingface_registries.id,
         huggingface_registries.url,
         huggingface_registries.token
  FROM huggingface_registries
  JOIN artifact_registries
    ON artifact_registries.registry_id = huggingface_registries.id

  artifact_registries.name exists only in artifact_registries, and HuggingFaceRegistrySearchableFields.own.to_data used to read it through the meta relationship (row.meta). To remove meta while changing as little of the existing code as possible, I kept to_data(row) and the returned data as they were and read artifact_registries.name through the join instead. A join by itself doesn't return artifact_registries.name, since the select lists only huggingface_registries columns. with_expression adds it and loads it into the row's registry_name, so to_data(row) can keep taking just the row.

src/ai/backend/manager/models/huggingface_registry/searchable_fields.py
wngh4628: Added a separate ArtifactRegistryNameNotLoadedError and replaced it in both registries.

src/ai/backend/manager/models/mixins/registry_name.py
wngh4628: Removing the meta relationship (Mapped[ArtifactRegistryRow | None]) left nowhere on the row for the name, and I wanted to keep to_data(row) and the returned data unchanged. So the query loads the joined name into a registry_name slot on the row:

SELECT artifact_registries.name,   -- loaded into row.registry_name
     reservoir_registries.id,
     reservoir_registries.endpoint,
     reservoir_registries.access_key,
     reservoir_registries.secret_key,
     reservoir_registries.api_version
FROM reservoir_registries
JOIN artifact_registries
ON artifact_registries.registry_id = reservoir_registries.id

Both registry rows need the same slot, and the registry write ops fills it for either one after create/update. Declaring it once in a mixin lets the ops set it with a single cast(RegistryNameMixin, row) instead of an isinstance branch per registry type.

PR #14881, opened 2026-09-21
refactor(BA-8003): remove the endpoints relationship from ServiceCatalogRow

wngh4628:
## Summary
- `models/service_catalog/row.py`: remove the `endpoints` relationship (`Mapped[list[ServiceCatalogEndpointRow]]`). The row keeps its columns and gains `endpoint_rows`, a `query_expression` slot the searcher fills with the service's endpoint rows as JSON.
- `ServiceCatalogSearcher`, the only reader of the relationship, loads the endpoints with a correlated `json_agg(row_to_json(service_catalog_endpoint))` subquery through `with_expression(endpoint_rows, ...)` instead of `selectinload`. A plain join can't be used here: it would repeat each service once per endpoint, and since the ops layer applies the limit, offset and count to that result, pages and `total_count` would be counted per endpoint instead of per service. The subquery keeps one row per service, so filters, count and pages stay the same, and `ServiceCatalogSearchableFields.own.to_data` unfolds the JSON into `ServiceCatalogEndpointData`, so the returned data is unchanged.

Resolves BA-8003


src/ai/backend/manager/models/service_catalog/types.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import annotations
+
+from typing import Any, TypedDict
+
+__all__ = ("ServiceCatalogEndpointRowJson",)
+
+
+class ServiceCatalogEndpointRowJson(TypedDict):
```
seedspirit: Is using `TypedDict` required? Using `Dataclass` instead of `TypedDict` seems more in line with existing conventions. If conversion to a dictionary is necessary, it might be a good idea to include a method like `to_dict` inside the class.

wngh4628: A dataclass is a good option too, but I think it's inefficient here. The driver decodes the `json_agg` result into a list of dicts for `endpoint_rows`, so a dataclass would need a hand-written `from_dict`, followed by another conversion into `ServiceCatalogEndpointData`, which adds one more step. The direction is also dict → object, so a `to_dict` wouldn't be used.

That said, I agree that matching the team convention is worthwhile. For JSON values read from the DB in the models layer, inheriting `BackendAISchema` fits the convention better (see `NumberFormat`, `RuntimeVariantPresetValueEntry`, and `VerificationStepResult.model_validate`). So I changed `ServiceCatalogEndpointRowJson` to a `BackendAISchema`, and the subquery result is now coerced with `PydanticListColumn`, so `endpoint_rows` holds validated models directly.

src/ai/backend/manager/models/service_catalog/row.py
```diff
@@ -70,10 +71,9 @@ class ServiceCatalogRow(Base):
         "config_hash", sa.String(length=128), nullable=False, server_default=sa.text("''")
     )
 
-    endpoints: Mapped[list[ServiceCatalogEndpointRow]] = relationship(
-        "ServiceCatalogEndpointRow",
-        cascade="all, delete-orphan",
-    )
+    # Not a column: the endpoint rows as JSON objects, filled by the searcher's
+    # with_expression(endpoint_rows, <json_agg subquery>). None when not loaded.
+    endpoint_rows: Mapped[list[ServiceCatalogEndpointRowJson] | None] = query_expression()
```
fregataa, line 76: How does this work if you write a new test file with a test setup that does not import ServiceCatalogEndpointRowJson?

wngh4628: `ServiceCatalogEndpointRowJson` is a pydantic model holding JSON, not a Row with a table, so it is not something that goes into a `with_tables` list. And SQLAlchemy resolves the name in `Mapped[list[ServiceCatalogEndpointRowJson] | None]` in `row.py`, not in the test file, so the module-level import in `row.py` is enough and a test file doesn't need to know this type.

I confirmed that three tests which don't import this type pass on this branch:
- `tests/unit/manager/repositories/service_catalog/test_service_catalog_repository.py` (`with_tables` lists only `ServiceCatalogRow` and `ServiceCatalogEndpointRow`)
- `tests/unit/manager/repositories/service_catalog/test_cursor_pagination.py` (same setup)
- `tests/scenario/bai_scenario/manager/service_catalog/test_searching.py` (checks that each service comes back with its endpoints)

PR #14906, opened 2026-09-22
doc: add BEP-1079 for the VXLAN overlay cluster network

daemyung-lablup:
Reserves 1079 and lands the draft: the manager allocates each cluster session a subnet and a VNI, each agent builds its node's half of a VXLAN overlay through a runtime-neutral seam, with an optional privileged helper so the agent itself needs no CAP_NET_ADMIN. Docker is the first backend.

resolves #NNN (BA-MMM)

- [ ] Backport targets: a `fix:` PR goes to every version in `.github/maintained-versions.yml` automatically.
      A `Backport:` line here names the targets instead — some of them, other ones, or `none`.
- [ ] Test case(s) to:
- [ ] Documentation


fregataa (commented):
Looks good!
1. Please add a news fragment file (changelog)
2. Let's add a mention about BEP-1028 which describes k8s bridge network to understand why k8s is out of scope in this doc

daemyung-lablup:
@fregataa I added changelog, but BEP-1028 is somewhat missing. so I don't refer it.

fregataa (approved):

PR #14927, opened 2026-09-22
refactor(BA-8005): remove the runtime_variant_row relationship from DeploymentRevisionRow

wngh4628:
## Summary
- `models/deployment_revision/row.py`: remove the `runtime_variant_row` relationship and its join condition from `DeploymentRevisionRow`. `image_row` stays with a comment, since `api/gql_legacy/endpoint.py` still reads it; `resource_slot_rows` is out of scope (`to_data` and the creator depend on it).
- `repositories/deployment/db_source/db_source.py`: the deployment context read gets the revision with `session.get` and then its `ImageRow` with `session.get` instead of `selectinload(image_row)`.  Same number of selects as before, and the returned data is unchanged.

Resolves BA-8005


src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -1993,18 +1994,14 @@ async def fetch_deployment_context(
...
 
-            revision_query = (
-                sa.select(DeploymentRevisionRow)
-                .where(DeploymentRevisionRow.id == revision_id)
-                .options(selectinload(DeploymentRevisionRow.image_row))
-            )
-            revision_result = await db_sess.execute(revision_query)
-            revision_row = revision_result.scalar_one_or_none()
-            if revision_row is None or revision_row.image_row is None:
+            revision_row = await db_sess.get(DeploymentRevisionRow, revision_id)
```
fregataa, line 1997: Seems we don't need to change the original query since image row relationship is not removed

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
```diff
@@ -2477,11 +2474,7 @@ async def _fetch_latest_revision_row(
                 .where(DeploymentRevisionRow.endpoint == endpoint_id)
                 .order_by(DeploymentRevisionRow.revision_number.desc())
                 .limit(1)
-                .options(
-                    selectinload(DeploymentRevisionRow.resource_slot_rows),
-                    selectinload(DeploymentRevisionRow.runtime_variant_row),
-                    selectinload(DeploymentRevisionRow.image_row),
-                )
+                .options(selectinload(DeploymentRevisionRow.resource_slot_rows))
```
fregataa, line 2477: Is `resource_slot_rows` relationship is also still in use?

wngh4628: Yes, it is still in use — the creator writes the child rows through it and both DeploymentRevisionRow and EndpointRow to_data read the slot quantities from it. Removing it touches more than this PR should, so I decided to split it into a separate PR. This is noted in the summary as well.

src/ai/backend/manager/repositories/deployment/db_source/db_source.py
wngh4628: The `image_row` relationship is still read by gql_legacy (`api/gql_legacy/endpoint.py`), and there is no date for when that goes away. So the relationship itself stays, but the v2 readers are moved off it now, so that less work is left when the legacy code is finally removed. This PR handles the two readers in the deployment domain (this spot and `_fetch_latest_revision_row`); the remaining non-legacy readers (endpoint row, model_serving) will be cleaned up in their own domain PRs in turn.

